PR #1183, opened 2022-11-17
SOLR-16555: SolrIndexSearcher - FilterCache intersections/andNot should not clone bitsets repeatedly

risdenk:
https://issues.apache.org/jira/browse/SOLR-16555

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
+++ b/solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
@@ -1211,17 +1211,42 @@ public class SolrIndexSearcher extends IndexSearcher implements Closeable, SolrI
     }
 
     // Are all of our normal cached filters negative?
-    if (end > 0 && answer == null) {
-      answer = getLiveDocSet();
-    }
+    if (end > 0) {
+      if (answer == null) {
+        answer = getLiveDocSet();
+      }
 
-    // do negative queries first to shrink set size
-    for (int i = 0; i < end; i++) {
-      if (neg[i]) answer = answer.andNot(sets[i]);
-    }
+      // min size must be maxDoc based on BitDocSet conversion at the end
+      FixedBitSet tempAnswer = FixedBitSet.ensureCapacity(answer.getFixedBitSetClone(), maxDoc());
 
-    for (int i = 0; i < end; i++) {
-      if (!neg[i] && i != smallestIndex) answer = answer.intersection(sets[i]);
+      // do negative queries first to shrink set size
+      for (int i = 0; i < end; i++) {
+        if (neg[i]) {
+          // Clear any documents that are in the negative filter query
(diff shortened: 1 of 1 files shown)


solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1211,17 +1211,42 @@ public ProcessedFilter getProcessedFilter(DocSet setFilter, List<Query> queries)
...
-    for (int i = 0; i < end; i++) {
-      if (neg[i]) answer = answer.andNot(sets[i]);
-    }
+      // min size must be maxDoc based on BitDocSet conversion at the end
+      FixedBitSet tempAnswer = FixedBitSet.ensureCapacity(answer.getFixedBitSetClone(), maxDoc());
+
+      // do negative queries first to shrink set size
+      for (int i = 0; i < end; i++) {
+        if (neg[i]) {
+          // Clear any documents that are in the negative filter query
```
risdenk, line 1225: I need to add a comment around why not using `answer.andNot(sets[i])`

risdenk (commented):
I'm running through some performance tests with this over the next day or so. Will be interesting to see what shakes out.

dsmiley (changes requested):
Your first draft of this isn't going to be performant.  Instead of bulk bitSet operations, you're iterating doc-by-doc!  For small sets (SortedIntDocSet), this is fine, but for bigger ones -- no.

dsmiley:
Just to be clear, this is a highly performance critical code path.  Sadly we have no benchmarks of this that I know of but at least there's peer review.  Once the solution looks more work-able, I'll at-mention some more eye-balls.  Maybe you might consider adding to /solr/benchmark.

risdenk:
> Instead of bulk bitSet operations, you're iterating doc-by-doc! For small sets (SortedIntDocSet), this is fine, but for bigger ones -- no.

@dsmiley yea not saying this is the final version at all. I'm actually curious on testing this with some benchmarking stuff to see how much the memory allocations slowed this down. This approach requiring more CPU, but less memory. I would not disagree there is more performance using the bulk bitset operations.

risdenk:
From David on Jira an idea to explore:
> Create a subclass of BitDocSet called MutableBitDocSet. Its intersect() and andNot() methods can return itself, and be implemented that way. In getProcessedFilter, before those two tight loops that call those two methods, upgrade/replace "answer" with MutableBitDocSet if end (the number of combinations to do) is greater than 1. If it's 1 or 0 then there's no point in the upgrade.

risdenk:
This is inefficient as David pointed out and caused timeouts on a bunch of queries during testing. I was hoping to at least capture some profile from it but didn't last long enough to do that. https://github.com/apache/solr/pull/1184 is next iteration using a `MutableBitSet` as David pointed out.

PR #1184, opened 2022-11-18
SOLR-16555: SolrIndexSearcher - FilterCache intersections/andNot should not clone bitsets repeatedly

risdenk:
https://issues.apache.org/jira/browse/SOLR-16555

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/BitDocSet.java
+++ b/solr/core/src/java/org/apache/solr/search/BitDocSet.java
@@ -40,7 +40,7 @@ public class BitDocSet extends DocSet {
 
   // TODO consider SparseFixedBitSet alternative
 
-  private final FixedBitSet bits;
+  protected final FixedBitSet bits;
   int size; // number of docs in the set (cached for perf)
 
   public BitDocSet() {
--- a/solr/core/src/java/org/apache/solr/search/MutableBitDocSet.java
+++ b/solr/core/src/java/org/apache/solr/search/MutableBitDocSet.java
@@ -1,0 +1,54 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.search;
+
+import org.apache.lucene.util.FixedBitSet;
+
+public class MutableBitDocSet extends BitDocSet {
+  public MutableBitDocSet(FixedBitSet bits) {
+    super(bits);
+  }
+
+  @Override
+  public DocSet andNot(DocSet other) {
+    if (other instanceof BitDocSet) {
+      bits.andNot(((BitDocSet) other).bits);
+    } else {
+      DocIterator iter = other.iterator();
+      while (iter.hasNext()) {
+        int doc = iter.nextDoc();
+        if (doc < bits.length()) {
+          bits.clear(doc);
+        }
+      }
+    }
+    return new MutableBitDocSet(bits);
+  }
+
+  @Override
+  public DocSet intersection(DocSet other) {
+    // intersection is overloaded in the smaller DocSets to be more
+    // efficient, so dispatch off of it instead.
+    if (!(other instanceof BitDocSet)) {
+      return other.intersection(this);
+    }
+
+    // Default... handle with bitsets.
+    bits.and(other.getFixedBitSet());
+    return new MutableBitDocSet(bits);
+  }
+}
--- a/solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
+++ b/solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
@@ -1211,17 +1211,22 @@ public class SolrIndexSearcher extends IndexSearcher implements Closeable, SolrI
     }
 
     // Are all of our normal cached filters negative?
-    if (end > 0 && answer == null) {
-      answer = getLiveDocSet();
-    }
+    if (end > 0) {
+      if (answer == null) {
+        answer = getLiveDocSet();
+      }
 
-    // do negative queries first to shrink set size
-    for (int i = 0; i < end; i++) {
-      if (neg[i]) answer = answer.andNot(sets[i]);
-    }
+      answer =
+          new MutableBitDocSet(FixedBitSet.ensureCapacity(answer.getFixedBitSetClone(), maxDoc()));
 
-    for (int i = 0; i < end; i++) {
-      if (!neg[i] && i != smallestIndex) answer = answer.intersection(sets[i]);
+      // do negative queries first to shrink set size
+      for (int i = 0; i < end; i++) {
+        if (neg[i]) answer = answer.andNot(sets[i]);
+      }
+
+      for (int i = 0; i < end; i++) {
+        if (!neg[i] && i != smallestIndex) answer = answer.intersection(sets[i]);
+      }
     }
 
     // ignore "answer" if it simply matches all docs


solr/core/src/java/org/apache/solr/search/MutableBitDocSet.java
```diff
@@ -0,0 +1,54 @@
...
+    } else {
+      DocIterator iter = other.iterator();
+      while (iter.hasNext()) {
+        int doc = iter.nextDoc();
+        if (doc < bits.length()) {
+          bits.clear(doc);
+        }
+      }
+    }
+    return new MutableBitDocSet(bits);
```
risdenk, line 39: Need comment - can't return `this` since `BitSetDoc` stores size cached. Easier to just create new lightweight object.

risdenk:
@dsmiley here is the `MutableBitDocSet` version. I'll consolidate down to one PR once I do some testing and get some perf numbers. I just pushed it up separate to get some eyes on it.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1211,17 +1211,22 @@ public ProcessedFilter getProcessedFilter(DocSet setFilter, List<Query> queries)
     }
 
     // Are all of our normal cached filters negative?
-    if (end > 0 && answer == null) {
-      answer = getLiveDocSet();
-    }
+    if (end > 0) {
```
dsmiley, line 1214: As I said, I think we shouldn't do this if equals to 1 since there will be at most one copy any way (sometimes optimizations lead to none).  Furthermore, consider that "answer" may be a SortedIntDocSet and thus upgrading it to a BitSet may be wasteful.

dsmiley: Also, iff smallestIdx is >= 0 then "end" is effectively one smaller for the purposes of this check.

risdenk: > Also, iff smallestIdx is >= 0 then "end" is effectively one smaller for the purposes of this check.

not sure I understand this comment. I'll dig in further. The other changes I'll push in a few minutes.

solr/core/src/java/org/apache/solr/search/MutableBitDocSet.java
risdenk: addressed in bc97db7183cedabbb7615c081298f45f34750bb1

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
risdenk: Pushed some of these optimizations in 5e753aa44b848f4e45361864c84be23a0c774edd - I checked if answer is instance of BitDocSet since thats the only case we can optimize easily.

solr/core/src/java/org/apache/solr/search/MutableBitDocSet.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.lucene.util.FixedBitSet;
+
+/**
+ * A {@link BitDocSet} based implementation that mutates the underlying bits for andNot and
+ * intersection. This allows for computing the combinations of sets without duplicating the
+ * underlying array.
+ *
+ * @since solr 9.2
+ */
+public class MutableBitDocSet extends BitDocSet {
```
dsmiley, line 28: Let's not make this public.

solr/core/src/java/org/apache/solr/search/MutableBitDocSet.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+package org.apache.solr.search;
+
+import org.apache.lucene.util.FixedBitSet;
+
+/**
+ * A {@link BitDocSet} based implementation that mutates the underlying bits for andNot and
+ * intersection. This allows for computing the combinations of sets without duplicating the
+ * underlying array.
+ *
```
dsmiley, line 25: Lets warn that this ought not to be cached, because it might inadvertently be modified.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1211,17 +1211,28 @@ public ProcessedFilter getProcessedFilter(DocSet setFilter, List<Query> queries)
...
+      }
 
-    // do negative queries first to shrink set size
-    for (int i = 0; i < end; i++) {
-      if (neg[i]) answer = answer.andNot(sets[i]);
-    }
+      // This optimizes for the case where we have more than 2 filters and instead
+      // of copying the bitsets we make one mutable bitset. We only need to do this
+      // for BitDocSet since it clones the backing bitset for andNot and intersection.
+      if (end > 1 && answer instanceof BitDocSet) {
```
dsmiley, line 1222: at this spot, end is guaranteed to be greater than 1 based on you putting a big if-else around this section (line 1214); right?

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1211,17 +1211,28 @@ public ProcessedFilter getProcessedFilter(DocSet setFilter, List<Query> queries)
...
+        FixedBitSet fixedBitSet =
+            FixedBitSet.ensureCapacity(answer.getFixedBitSetClone(), maxDoc());
+        answer = new MutableBitDocSet(fixedBitSet);
```
dsmiley, line 1225: I suggest adding a small static utility method on MutableBitDocSet `fromBitDocSet` that takes only BitDocSet and maxDoc.

dsmiley (changes requested):
Looking good!
At the end of the loops, we should make sure we don't return a MutableBitDocSet to avoid modifying it if it gets cached later, even if that might be impossible today.  So you could add a static method on MutableBitDocSet `unwrapIfMutable` to make this a nice one-liner.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
risdenk: the line 1214 is checking if `end > 0` - which was just an expansion of the previous if statement that was there. Previously `end > 0` was checked before updating `answer` only and then the for loops would skip since i = 0 and end = 0 in that case. I wanted it to be clear - if `end <= 0` we don't need to do this whole block.

The end > 1 here is to make sure we have 2 bitsets to work with?

Unless I'm misunderstanding your command.

solr/core/src/java/org/apache/solr/search/MutableBitDocSet.java
risdenk: Gah yea good call - I updated the javadoc on the methods but will add that to the class too.

risdenk:
I ran this through some benchmarks - same one as the screenshots on SOLR-16555 and got the following:

Screenshot 2022-11-18 at 14 44 07
Screenshot 2022-11-18 at 14 44 24

I found that `[FixedBitSet.ensureCapacity](https://github.com/apache/solr/pull/1184/commits/f1fe35ca91a9a219dcadc29186a963e8f665e5b5)` isn't even needed so can skip that too. I pushed that commit [f1fe35c](https://github.com/apache/solr/pull/1184/commits/f1fe35ca91a9a219dcadc29186a963e8f665e5b5) a few minutes ago.

I'll update based on David's comments in a few minutes as well.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
dsmiley: RE smallestIdx that I was saying, the logic here is confusing.  I think a refactor is in order to make this clearer.  I can do a push to your branch doing that, and thus adopting a small improvement here.  Without _something_, end could easily be equal to 1 and smallestIdx == 0 which is a scenario where we should do nothing because none of the pos/neg loops will do any work.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
risdenk: Addressed in bdf47db

solr/core/src/java/org/apache/solr/search/MutableBitDocSet.java
risdenk: Addressed in bdf47db

risdenk:
So I reran this update especially after f1fe35c but included the review changes as well.

Memory allocation is no longer dominated by getProcessedFilter. I'd share screenshots, but custom scorers now dominate the flamegraph after cleaning this up.

Some performance numbers from a cluster test:
* median latency baseline from prod - something like 23ms. iteration prior to f1fe35c -> 18ms. second iteration after f1fe35c and cleanup -> 11ms
* p75 in the 16ms range vs ~60ms range from prod
* p99 in the 100ms range vs ~325ms range from prod
* prod was hitting ~75% cpu. this is 30% CPU
* GC counts were in the 250 range for prod at ~2.25s. during this test maxing around 75 range and around 500ms max
* total estimated memory allocation in prod for 5 min period - 1.5TB. total memory allocated in test during 5 min period = 356GB
* ~50% reduction in latency in ms across the board for percentiles under 99.99%.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1190,38 +1187,53 @@ public ProcessedFilter getProcessedFilter(DocSet setFilter, List<Query> queries)
...
       Query posQuery = QueryUtils.getAbs(q);
-      sets[end] = getPositiveDocSet(posQuery);
+      DocSet docSet = getPositiveDocSet(posQuery);
       // Negative query if absolute value different from original
       if (Objects.equals(q, posQuery)) {
-        neg[end] = false;
-        // keep track of the smallest positive set.
+        // keep track of the smallest positive set; use "answer" for this.
         // This optimization is only worth it if size() is cached, which it would
         // be if we don't do any set operations.
```
risdenk, line 1195: I think this comment can move down to just above line 1200

dsmiley (approved):
Hopefully someone else takes a look as well.  Like @magibney or @madrob maybe.

risdenk:
Thanks @dsmiley appreciate the reviews. I added a bunch of others to hopefully make sure we get another set of eyes on this. I'll test this change again today with your commit as well.

solr/core/src/java/org/apache/solr/search/MutableBitDocSet.java
```diff
@@ -0,0 +1,103 @@
...
+   * bits so do not cache the returned bitset.
+   *
+   * @return a DocSet representing this AND NOT other
+   */
+  @Override
+  public DocSet andNot(DocSet other) {
+    if (other instanceof BitDocSet) {
+      bits.andNot(((BitDocSet) other).bits);
+    } else {
+      DocIterator iter = other.iterator();
```
madrob, line 68: This is mostly copied from BitDocSet, right? It feels like there could be a static method `andNot(DocSet left, DocSet right)` and this class calls `andNot(this.bits, other)` while BitDocSet calls `andNot(this.bits.clone(), other)` and we have the logic in one place.

solr/core/src/java/org/apache/solr/search/MutableBitDocSet.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.lucene.util.FixedBitSet;
+
+/**
+ * A {@link BitDocSet} based implementation that mutates the underlying bits for andNot and
+ * intersection. This allows for computing the combinations of sets without duplicating the
+ * underlying array. This MutableBitDocSet should not be cached because it can be modified.
+ *
+ * @since solr 9.2
+ */
+class MutableBitDocSet extends BitDocSet {
```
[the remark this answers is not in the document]
dsmiley: As a TODO, okay.  But not essential.  That method isn't used where we are using MutableBitDocSet.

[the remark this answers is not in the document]
risdenk: I can go a step further here and explicitly throw UnsupportedOperationException for methods that we don't use and don't plan to support now.

solr/core/src/java/org/apache/solr/search/MutableBitDocSet.java
risdenk: Fixed in 26754f8e2f8db5ad6a5535eb10cb73d43d4739d3

solr/core/src/java/org/apache/solr/search/MutableBitDocSet.java
```diff
@@ -0,0 +1,103 @@
...
+ * limitations under the License.
+ */
+package org.apache.solr.search;
+
+import org.apache.lucene.util.FixedBitSet;
+
+/**
+ * A {@link BitDocSet} based implementation that mutates the underlying bits for andNot and
+ * intersection. This allows for computing the combinations of sets without duplicating the
+ * underlying array. This MutableBitDocSet should not be cached because it can be modified.
```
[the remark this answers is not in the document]
risdenk: Fixed in e4c92289514e91bfdefb727437f6e2a042e0577e

solr/core/src/java/org/apache/solr/search/MutableBitDocSet.java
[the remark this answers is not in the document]
risdenk: Fixed in 414c8749dcdf2530068014c4ee81a3a518de50bf

solr/core/src/java/org/apache/solr/search/BitDocSet.java
```diff
@@ -194,19 +194,30 @@ public void addAllTo(FixedBitSet target) {
...
+
+  /**
+   * Helper method for andNot that takes FixedBitSet and DocSet. This returns the resulting bits
+   * andNoted together.
+   *
+   * @param bits bits to operate on
+   * @param other The DocSet to compare to
+   * @return Resulting andNoted FixedBitSet
+   */
+  protected FixedBitSet andNot(FixedBitSet bits, DocSet other) {
```
magibney, line 208: nit: this could be a static method?

Also, a couple of minor issues with javadoc wording:
1. it could be clearer that this method modifies the docset provided as the first arg and returns that same object (making this a static method might help clarify that point).
2. "andNoted together" implies that "andNot" is a symmetric operation, which of course it's not. Maybe something like "modifies the first arg input BitDocSet to remove all bits contained in the second DocSet arg -- equivalent to calling `a.andNot(b)`, but modifies the state of the base docset instead of returning a new docSet"?

solr/core/src/java/org/apache/solr/search/MutableBitDocSet.java
```diff
@@ -0,0 +1,137 @@
...
+  /**
+   * Returns the documents in this set that are not in the other set. This mutates the underlying
+   * bits so do not cache the returned bitset.
+   *
+   * @return a DocSet representing this AND NOT other
+   */
+  @Override
+  public DocSet andNot(DocSet other) {
+    // We can't return just this since `size` is cached and
+    // we are changing the cardinality of the underlying bits.
```
magibney, line 119: True; but doesn't that potentially leave `this` in an inconsistent state -- with a cached `size` that doesn't match the size according to the underlying bits? Shouldn't we clear/reset size to `-1`? And, if we do that, could we then just return `this`?

solr/core/src/java/org/apache/solr/search/BitDocSet.java
risdenk: addressed in a1af3895af00b37b99d738556ce14cd75441f502

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1190,38 +1192,52 @@ public ProcessedFilter getProcessedFilter(DocSet setFilter, List<Query> queries)
...
+        if (!neg[i]) answer = answer.intersection(sets[i]);
+      }
 
-    for (int i = 0; i < end; i++) {
-      if (!neg[i] && i != smallestIndex) answer = answer.intersection(sets[i]);
+      // Make sure to keep answer as an immutable DocSet if we made it mutable
+      answer = MutableBitDocSet.unwrapIfMutable(answer);
     }
 
     // ignore "answer" if it simply matches all docs
```
magibney, line 1243: Out of scope of this PR, but while looking at this code: will `answer.size()` always be called downstream of this? If so, then it's fine to force size calculation here; if _not_, I think there should be better ways to check this condition than naively calling `answer.size()` after all intersections are done.

magibney (approved):
A couple more minor questions/suggestions around avoiding unnecessary `size` calculation, but perhaps not practically that significant. Overall this looks great.

solr/core/src/java/org/apache/solr/search/BitDocSet.java
```diff
@@ -194,19 +194,31 @@ public void addAllTo(FixedBitSet target) {
...
-      newbits.andNot(((BitDocSet) other).bits);
+      bits.andNot(((BitDocSet) other).bits);
     } else {
       DocIterator iter = other.iterator();
       while (iter.hasNext()) {
         int doc = iter.nextDoc();
-        if (doc < newbits.length()) {
-          newbits.clear(doc);
+        if (doc < bits.length()) {
+          bits.clear(doc);
```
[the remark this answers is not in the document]
risdenk: eh not sure it would be worth it.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
risdenk: so in theory you are probably correct. in practice don't plan on tackling this here :)

risdenk:
Thanks @madrob and @magibney for the reviews!

dsmiley:
Awesome perf analysis by the way!

PR #1185, opened 2022-11-22
SOLR-16553 and SOLR-16559

risdenk:
Contains two commits that are related to fixing the benchmark stuff.
* https://issues.apache.org/jira/browse/SOLR-16553
* https://issues.apache.org/jira/browse/SOLR-16559

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/NodeConfig.java
+++ b/solr/core/src/java/org/apache/solr/core/NodeConfig.java
@@ -253,13 +253,13 @@ public class NodeConfig {
   /**
    * Obtain the path of solr's binary installation directory, e.g. <code>/opt/solr</code>
    *
-   * @return path to install dir
-   * @throws SolrException if property 'solr.install.dir' has not been initialized
+   * @return path to install dir or null if solr.install.dir not set.
    */
   public Path getSolrInstallDir() {
     String prop = System.getProperty(SolrDispatchFilter.SOLR_INSTALL_DIR_ATTRIBUTE);
     if (prop == null || prop.isBlank()) {
-      throw new SolrException(ErrorCode.SERVER_ERROR, "solr.install.dir property not initialized");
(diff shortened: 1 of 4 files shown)


solr/core/src/java/org/apache/solr/core/NodeConfig.java
```diff
@@ -476,18 +479,22 @@ private void addFoldersToSharedLib(Set<String> libDirs) {
 
   // Adds modules to shared classpath
   private void initModules() {
+    Path solrInstallDir = getSolrInstallDir();
+    if (solrInstallDir == null) {
+      return;
```
janhoy, line 484: I'm not a fan of returning null to handle exceptional cases. Why not catch the exception instead? 

And certainly not a fan of silenlty returning like this. At the very least, it should do a check to see if any modules are requested, and throw exception if they are but we cannot load them due to, in this case, a configuration error.

risdenk: I don't think `getSolrInstallDir()`  should return an exception if the property isn't set. Returning `null` from that method is 100% valid since it is saying "there is no install dir" - `initModules` and `setupSharedLib` should handle the case of no install dir. I agree that there should be a message returned in that case if there is no install dir.

risdenk: @janhoy I pushed a change that gives some more detail in the case of modules and shared lib. Let me know if that looks better for you.

janhoy: This was better, although we'd probably be better off for all tests to also set `solr.install.dir` and treat lack thereof as a configuration error. But now we at least report why things are failing if it is not set. I'm ok with this too.

janhoy (approved):
Approving - have only reviewed the solr.install.dir part

PR #1197, opened 2022-11-28
Refactor: remove unused "filter" in QueryCommand & getProcessedFilter

dsmiley:
1. QueryCommand: remove unused "filter"
2. SolrIndexSearcher.getProcessedFilter: remove unused setFilter param

Background:
While looking at SolrIndexSearcher.getProcessedFilter (on multiple occassions), it seemed to me the setFilter param was needless because a caller could simply provide the DocSet as a Query via DocSet.makeQuery (new in Solr 9).  Then I realized that nobody actually passes setFilter any way so I simply removed it, leading me to remove QueryCommand.filter in-kind.  Maybe it was used a long time ago; this is old code.

Not sure this needs a JIRA; it's a refactoring.  I'm thinking of only doing this in Solr 10 for back-compat sake.

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/component/ExpandComponent.java
+++ b/solr/core/src/java/org/apache/solr/handler/component/ExpandComponent.java
@@ -432,7 +432,7 @@ public class ExpandComponent extends SearchComponent implements PluginInfoInitia
       newFilters.add(groupQuery);
     }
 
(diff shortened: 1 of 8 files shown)


mkhludnev (approved):

risdenk (approved):

dsmiley:
I think this PR is ready except fixing conflicts & adding a CHANGES.txt.

PR #1118, opened 2022-10-24
[SOLR-16489] CaffeineCache puts thread into infinite loop

tboeghk:
https://issues.apache.org/jira/browse/SOLR-16489

# Description

Under heavy machine load, accessing any Solr CaffeineCache can lead to threads spinning in an endless loop. In our setup we had machines spinning at 70% cpu load without receiving any request for hours.

# Solution

The problem is caused by delegating the `SolrCache#put` method to `Cache#asMap#put` Under heavy machine load with concurrent read and write access to the same key, the `ConcurrentHashMap#put` method does not terminate an endless loop.

# Tests

The behavior happens under heavy load when the cache is under congestion and read / write access is happening in different threads. We had this small fix live for a couple of weeks now without any stuck threads.

# Checklist

Please review the following and check all that apply:

- [x] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [x] I have created a Jira issue and added the issue ID to my pull request title.
- [x] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [x] I have developed this patch against the `main` branch.
- [x] I have run `./gradlew check`.
- [ ] I have added tests for my changes.
- [ ] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

PR #1198, opened 2022-11-29
SOLR-16562: Upgrade to Caffeine 3.1.2

risdenk:
https://issues.apache.org/jira/browse/SOLR-16562

This update of Caffeine includes infinite loop checking.

This also upgrades the following:
    * SOLR-16579 - Jackson 2.14.1
    * SOLR-16578 - errorprone 2.16
    * and other minor dependencies based on transitive dependencies.


solr/core/src/java/org/apache/solr/core/backup/BackupManager.java
```diff
@@ -252,7 +252,7 @@ public void uploadConfigDir(
       String sourceConfigName, String targetConfigName, ConfigSetService configSetService)
       throws IOException {
     URI source = repository.resolveDirectory(getZkStateDir(), CONFIG_STATE_DIR, sourceConfigName);
-    Preconditions.checkState(repository.exists(source), "Path {} does not exist", source);
+    Preconditions.checkState(repository.exists(source), "Path %s does not exist", source);
```
risdenk, line 255: This was found on the errorprone upgrade.

risdenk:
Relates to https://github.com/apache/solr/pull/1118 - the underlying issue w/ infinite loop is mutable Lucene/Solr query classes.

epugh (approved):
amazing how many files an upgrade touches, however it all LGTM

fsparv:
Can we fix the typo comment at the top to say woodstox-core 6.4 not 2.4? Also it looks like this happens to address CVE-2022-40152

risdenk:
Thanks for catching that @fsparv - updated the commit message. Bummer not all dependencies start with `2.x` :P

magibney (approved):

solr/CHANGES.txt
```diff
@@ -140,6 +140,8 @@ Other Changes
 
 * SOLR-15861: ConcurrentUpdateSolrClient should work with ManagedExecutorService (Sammy Chu, Kevin Risden)
 
+* SOLR-16562: Upgrade to Caffeine 3.1.2 (Kevin Risden)
```
HoustonPutman, line 143: Probably want to mention the other dependency upgrades as well.

dsmiley: definitely!  Dependency changes need to be very visible in a JIRA issue & CHANGES.txt

dsmiley:
Why is this upgrading more than one thing?  Pure convenience?  Caffeine has no dependencies.

risdenk:
> Why is this upgrading more than one thing? Pure convenience? Caffeine has no dependencies.

https://mvnrepository.com/artifact/com.github.ben-manes.caffeine/caffeine/3.1.2

Caffeine definitely has dependencies. Upgrading Caffeine triggered these dependency updates.

risdenk:
yea I'm not even sure I trust the mvnrepository.com output - https://github.com/ben-manes/caffeine/blob/master/gradle/dependencies.gradle shows lots of dependencies at least managed. https://github.com/ben-manes/caffeine/blob/master/caffeine/build.gradle is the limited list just for `caffeine` library. I wouldn't have guessed these dependencies updates would happen without the consistent versions highlighting it.

FWIW I'm not opposed to splitting the `versions.props` specific dependency changes (jackson, woodstox, errorprone) into other commits.

dsmiley:
I have no problem doing a bunch of updates at once so long as it's clear in JIRA & CHANGES.txt that we did so

ben-manes:
Caffeine has no required dependencies. It does have optional annotations that can be excluded (checker, errorprone) that are compile scoped. The managed dependencies are internal for testing, benchmarks, analysis and should not be in the pom.

The actual bug in solr hasn’t been fixed, so while this is good to fail fast, I am s little concerned that the root problem won’t be addressed.

ben-manes:
I believe that the managed dependencies in the pom are Gradle constraints added to protect the build itself from exploits. It’s noise that doesn’t impact users, but I’ll look into removing it from the pom as an implementation detail to avoid CI attacks like the codecov breach.

ben-manes:
Solr worked around the same issue wrt managed dependencies being unintentionally exported. While this fixes the pom, the gradle module metadata still includes these and, as Gradle prefers that over the pom file, it takes precedence for consumers who use a Gradle build. To resolve that, I [switched](https://github.com/ben-manes/caffeine/commit/7321a56a0fe92308c5b20c2a789ab77907ca480e) to not applying constraint to the published configuration, which removed the section from the metadata files.

https://github.com/apache/solr/blob/1a940ade468074fb6bfed8acf699d2f2c570edf5/gradle/maven/defaults-maven.gradle#L152-L161

solr/core/src/java/org/apache/solr/core/backup/BackupManager.java
risdenk: Separated out to a different commit - SOLR-16578 - errorprone 2.16

solr/CHANGES.txt
risdenk: Made this 3 commits since the woodstox stuff was done via contributor PR. Reused the PR branch and filed the jiras to make it all clear what got upgraded in versions.props.

* SOLR-16579 - Jackson 2.14.1
* SOLR-16578 - errorprone 2.16

risdenk:
Thanks @ben-manes - I decided to separate these upgrades for Jackson and errorprone into separate commits to make it easier to see in solr/CHANGES.txt and not conflate the caffeine upgrade.

PR #1200, opened 2022-11-30
Invert error-prone configuration to be allow-list vs deny-list

risdenk:
This ports https://github.com/apache/lucene/pull/11970 to Solr. Copied commit message from Lucene below:

This does not change the semantics or performance of our setup.

Instead, it explicitly enables checks that we want vs disabling checks that we don't want.

Also reordered checks to match the error-prone website list of checks for easier maintenance.

It is now clear that many useless checks are enabled, we can disable some of them and try to get the performance reasonable.


risdenk:
I'm planning to follow up with disabling useless checks - like https://github.com/apache/lucene/pull/11971 is doing for Lucene.

gradle/validation/error-prone.gradle
```diff
@@ -62,53 +62,421 @@ allprojects { prj ->
...
+            '-Xep:BadAnnotationImplementation:ERROR',
+            '-Xep:BadShiftAmount:ERROR',
+            '-Xep:BanJNDI:ERROR',
+            '-Xep:BoxedPrimitiveEquality:ERROR',
+            '-Xep:BundleDeserializationCast:ERROR',
+            '-Xep:ChainingConstructorIgnoresParameter:ERROR',
+            '-Xep:CheckNotNullMultipleTimes:ERROR',
+            '-Xep:CheckReturnValue:ERROR',
+            '-Xep:CollectionToArraySafeParameter:ERROR',
             '-Xep:ComparableType:OFF', // SolrTestCaseJ4.Doc and Fld are messy
```
madrob, line 93: Can you comment out all the off checks so they are visually distinct?

rmuir: i think i ended up commenting them out in the followup PR. It does make it easier for my editor to see the disabled ones.

I'm not concerned about disabling-all-checks not actually disabling all checks. I think it only enables a few critical ones, such as unicode directional checker. These checks override `disableable = false` in their annotation in the source code, e.g. https://github.com/google/error-prone/blob/4c7dff57339bba6b4cf41fc776a51de6c6bd5e00/core/src/main/java/com/google/errorprone/bugpatterns/UnicodeDirectionalityCharacters.java#L34

And if you explicitly disable them, error-prone fails: 
```
Execution failed for task ':lucene:core:compileJava'.
> UnicodeDirectionalityCharacters may not be disabled
```

So I think it is best to make life easier on ourselves by using the comments? In this example, yeah we check for these unicode characters elsewhere in our build, so it is redundant, but it is not possible to disable with error-prone so best to just have it configured as `ERROR` and move on.

risdenk: addressed in 9245fec36031360bb46b5c43a2d48412ad2de096

PR #1201, opened 2022-11-30
Disable useless errorprone checks

risdenk:
This ports https://github.com/apache/lucene/pull/11971 to Solr and builds on https://github.com/apache/solr/pull/1200

### DIFF
--- a/gradle/validation/error-prone.gradle
+++ b/gradle/validation/error-prone.gradle
@@ -67,11 +67,13 @@ allprojects { prj ->
 
             // List of enabled/disabled checks
             // Please keep this synced with https://errorprone.info/bugpatterns when upgrading!
+            // Do *NOT* enable checks based on their name or description. Read the source code and make sure they are useful!
+            // Most error-prone checks are not useful for non-google software.
 
             // On by Default : ERROR
 
             '-Xep:AlwaysThrows:ERROR',
-            '-Xep:AndroidInjectionBeforeSuper:ERROR',
+            // '-Xep:AndroidInjectionBeforeSuper:OFF', // we don't use android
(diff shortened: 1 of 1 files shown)


gradle/validation/error-prone.gradle
```diff
@@ -67,43 +67,45 @@ allprojects { prj ->
 
             // List of enabled/disabled checks
             // Please keep this synced with https://errorprone.info/bugpatterns when upgrading!
+            // Do *NOT* enable checks based on their name or description. Read the source code and make sure they are useful!
+            // Most error-prone checks are not useful for non-google software.
 
             // On by Default : ERROR
```
dsmiley, line 73: Why have "On" by default if "most error-prone checks are not useful"?
Based on what I see below, if commenting makes it go away, then we are not "on" by default, thus perhaps this comment is wrong.

rmuir: the comment refers to the "On by default : ERROR" section of this page: https://errorprone.info/bugpatterns

the idea is to organize our (grouping, order) checks consistent with that page and to update it when updating the error-prone.

dsmiley: aaaah.  I thinking updating these comments like this would be more clear:
// Section "On by Default : ERROR"

and maybe a separate comment that we've disabled all checks by default, overriding the error-prone default setting.  At least I'm assuming we've done this.

rmuir: i think it already says all this. click "view file" because github isn't showing you enough context?

dsmiley: Indeed; thanks :-)

PR #1202, opened 2022-11-30
Lower gradle heap: 3GB is unnecessary (#11936)

risdenk:
Ports https://github.com/apache/lucene/pull/11936 over to Solr. I've been testing this the past week or so and haven't seen any issues.

You need to clear `gradle.properties` in the Solr directory and regenerate with `./gradlew localSettings` (or let the build regenerate it for you next time.

### DIFF
--- a/gradle/generation/local-settings.gradle
+++ b/gradle/generation/local-settings.gradle
@@ -65,10 +65,9 @@ configure(rootProject) {
 systemProp.file.encoding=UTF-8
 
 # Set up gradle JVM defaults.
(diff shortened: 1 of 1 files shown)


risdenk:
Anecdotally - this has improved using Gradle in Intellij for builds - the machine doesn't seem to freak out as much with smaller Gradle heap.

risdenk:
Not that I've seen - at least not drastically so. I haven't measured it to be honest. I can say that `./gradlew check -Pvalidation.errorprone=true` on my machine stays at around 20min plus or minus a few minutes.

PR #1204, opened 2022-12-01
SOLR-15955: Fix dependency in hadoop-auth

HoustonPutman:
This dependency is not compileTimeOnly. I think this was only done because of the weird old way of handling jetty dependencies.

After https://issues.apache.org/jira/browse/SOLR-16158 this should no longer be an issue.

### DIFF
--- a/solr/modules/hadoop-auth/build.gradle
+++ b/solr/modules/hadoop-auth/build.gradle
@@ -36,7 +36,7 @@ dependencies {
 
   implementation 'org.slf4j:slf4j-api'
 
-  compileOnlyApi 'org.eclipse.jetty.toolchain:jetty-servlet-api'
+  api 'org.eclipse.jetty.toolchain:jetty-servlet-api'
 
   implementation 'com.fasterxml.jackson.core:jackson-core'
   implementation 'com.google.guava:guava'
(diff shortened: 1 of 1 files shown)


solr/modules/hadoop-auth/build.gradle
```diff
@@ -103,7 +103,7 @@ dependencies {
     exclude group: "org.apache.yetus", module: "audience-annotations"
   })
   // required for instantiating a Zookeeper server in tests or embedded
-  runtimeOnly ('org.xerial.snappy:snappy-java')
+  testRuntimeOnly ('org.xerial.snappy:snappy-java')
```
HoustonPutman, line 106: this already exists above as `runtimeOnly`, but maybe we need it for tests as well? I don't know.

risdenk: I'd be surprised if the `runtimeOnly` version is needed?

at least in the `hdfs` module - there is only the `testRuntimeOnly`

risdenk: However in solr/core/build.gradle - `runtimeOnly ('org.xerial.snappy:snappy-java')` exists... so I'm conflicted.

Requires more digging but not going to stop this from getting merged.

PR #1149, opened 2022-10-30
SOLR-16498: Tests to retrieve url from SolrClient

joshgog:
https://issues.apache.org/jira/browse/SOLR-16498

# Description

Some tests are either getting the URL  from legacy/deprecated SolrClient classes or Apache HttpClient instance out of the client. Tests need to  get the URL from the SolrClient in a more general way..

# Solution

This is still a work in progress. Created a utility method that interrogates the passed SolrClient and returns the base url.

# Tests

Tests are not needed since this PR refactors existing tests.

# Checklist

Please review the following and check all that apply:

- [x] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [x] I have created a Jira issue and added the issue ID to my pull request title.
- [x] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [x] I have developed this patch against the `main` branch.
- [ ] I have run `./gradlew check`.
- [ ] I have added tests for my changes.
- [ ] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

PR #1206, opened 2022-12-01
SOLR-16564: Migrate to using DEFAULT_TEST_COLLECTION_NAME constant through out tests

epugh:
https://issues.apache.org/jira/browse/SOLR-16564

# Description
Following up from #1149 where we started replacing `"collection1"` strings with the `DEFAULT_TEST_COLLECTION_NAME` constant, this continues that process.

# Solution

Use constants where it makes sense.

If a test was interaction with a *collection1* and a *collection2*, then I left it, as I found that more readable.

# Tests

# Checklist


solr/core/src/test/org/apache/solr/cloud/LeaderVoteWaitTimeoutTest.java
```diff
@@ -179,74 +179,78 @@ public void basicTest() throws Exception {
 
   @Test
   public void testMostInSyncReplicasCanWinElection() throws Exception {
-    final String collectionName = "collection1";
-    CollectionAdminRequest.createCollection(collectionName, 1, 3)
+    // final String collectionName = "collection1";
```
risdenk, line 182: I think this could be 

```
final String collectionName = DEFAULT_TEST_COLLECTION_NAME;
```

which would make the diff a lot smaller?

solr/core/src/test/org/apache/solr/cloud/OverseerTest.java
```diff
@@ -127,7 +127,6 @@ public class OverseerTest extends SolrTestCaseJ4 {
   private final List<UpdateShardHandler> updateShardHandlers =
       Collections.synchronizedList(new ArrayList<>());
   private final List<CloudSolrClient> solrClients = Collections.synchronizedList(new ArrayList<>());
-  private static final String COLLECTION = SolrTestCaseJ4.DEFAULT_TEST_COLLECTION_NAME;
```
risdenk, line 130: Not sure this change is needed? It looks like this is doing the "right thing" already?

solr/core/src/test/org/apache/solr/cloud/TestBaseStatsCacheCloud.java
```diff
@@ -111,23 +111,23 @@ public void testBasicStats() throws Exception {
                 "q", "foo_t:\"bar baz\"", "fl", "*,score", "rows", "" + NUM_DOCS, "debug", "true"));
     QueryResponse controlRsp =
         control.query(
-            "collection1",
+                DEFAULT_TEST_COLLECTION_NAME,
```
risdenk, line 114: probably need `./gradlew tidy` due to indentation btw

solr/core/src/test/org/apache/solr/cloud/TestCloudRecovery.java
```diff
@@ -49,7 +49,6 @@
 
 public class TestCloudRecovery extends SolrCloudTestCase {
 
-  private static final String COLLECTION = "collection1";
```
risdenk, line 52: Could change to the following to reduce size of diff?:

```
private static final String COLLECTION = DEFAULT_TEST_COLLECTION_NAME;
```

solr/core/src/test/org/apache/solr/handler/TestReplicationHandler.java
```diff
@@ -87,18 +85,18 @@
  *
  * @since 1.4
  */
-@LuceneTestCase.Nightly
+// @LuceneTestCase.Nightly
```
risdenk, line 88: These changes shouldn't be here?

epugh: correct..   I need to figure out how to run a nightly unit test from inside of intellij..  i keep commenting this out to run the tests.....

solr/core/src/test/org/apache/solr/cloud/TestCloudRecovery.java
epugh: i started acoutally out with that...    And my thought was that the value of this entire PR is directly based on how much we get every test in every file to using the same name...    I really didn't want to go through by hand and fix all of these... ;-).   My thought is that by using the label DEFAULT_TEST_COLLECTION_NAME everywhere, it signals to you that you don't need to think about what the `COLLECTION` or `collection` variable means..   and that if you see either of those variables, then it presumably matters for some reason to that test....

solr/core/src/test/org/apache/solr/handler/TestReplicationHandler.java
epugh: also, on my slow laptop it takes 110 seconds roughly to run this test..   what is the criteria for being a nightly?

epugh:
This is starting to appear to be a sisphean task...  106 test files..  admittedly, not everyone needs the constant....

solr/core/src/test/org/apache/solr/cloud/LeaderVoteWaitTimeoutTest.java
epugh: I think the value of this PR is to use the same pattern everywhere (even though maybe the diff is bigger).    I want to only have a variable called collectionName if that is somehow important to the function of the test..  when I see the variable I immediately think, oh, it might change, and I need to think about that as I read the test.    Where as seeing the constant everywhere means I know that the name of the collection isn't important to the function of the test...   So thats why I made these changes, and yeah, it did make the diff bigger...   Does that argument sway your opinion?   I could split up this patch into multiple patches too if that was better...   My end goal on this PR is to use the same constant the same way everywhere that it makes sense......

solr/core/src/test/org/apache/solr/cloud/OverseerTest.java
epugh: Same argument as above...   If we learn, reading the tests, taht DEFAULT_TEST_COLLECTION_NAME means that the test using the default collection name, and that you don't need to think about it, then let's use that name everywhere...

solr/core/src/test/org/apache/solr/cloud/TestBaseStatsCacheCloud.java
epugh: Thanks!   Doing it now.

epugh:
Okay, I think I found a few more places, and at this point think I am done..  I don't know that this is the most impactful of PR's, but I'd love to merge it and call it done and close the JIRA if I can get a LGTM...

dsmiley:
I think I question the value in this.  "collection1" is friggin pervasive.  We all know what it is.  To the vast majority of tests, the test doesn't actually care what the collection it's going to create is named.  It's whatever.

dsmiley:
l'll add a disclaimer that, I've observed to some developers, use of constants is like some sort of religion.  I am most definitely *not* religious about any coding practice.  I think sometimes constants are a needless indirection to discern a value (not very relevant to this PR) and sometimes a bit more verbose as well.  Yet another import statement.

epugh:
Just not feeling it....

PR #1208, opened 2022-12-02
SOLR-16571: Add java system property to disable config watch

nginthfs:
# Description

Solr creates a ZooKeeper watch per-core. On nodes with a lot of cores, this results in a lot of ZK watches, and thus increased resource consumption. ZK config watches aren't strictly necessary, so we should be able to disable them if we want to. 

# Solution

To remedy this, I've added a Java system property to disable the ZK config watch per core. The default behavior is still to create a ZK config watch per core. 

# Tests

I've added a test confirming that the ZK watch is not created when the system property is set to true. 

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/ZkController.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ZkController.java
@@ -2736,6 +2736,10 @@ public class ZkController implements Closeable {
     return confDirListeners;
   }
 
+  public boolean hasConfDirectoryListeners(final String confDir) {
+    return confDirectoryListeners.containsKey(confDir) && !confDirectoryListeners.isEmpty();
+  }
+
   private final Map<String, Set<Runnable>> confDirectoryListeners = new HashMap<>();
 
   private class WatcherImpl implements Watcher {
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -188,6 +188,7 @@ import org.slf4j.LoggerFactory;
 public class SolrCore implements SolrInfoBean, Closeable {
 
   public static final String version = "1.0";
+  public static final String DISABLE_ZK_CONFIG_WATCH = "disable.zk.config.watch";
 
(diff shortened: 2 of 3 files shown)


solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -188,6 +188,7 @@
 public class SolrCore implements SolrInfoBean, Closeable {
 
   public static final String version = "1.0";
+  public static final String DISABLE_ZK_CONFIG_WATCH = "disable.zk.config.watch";
```
dsmiley, line 191: I would prefer this be named something like `configSet.core.watch.enable`

nginthfs: Would `configSet.core.watch.disable` be okay, or should I change the logic to check for the Boolean and default to false?

dsmiley: Yeah, you're right.  The boolean should represent the change from its default state.

dsmiley:
Let's see where my comment goes on the parent JIRA.  Note that this test fails no matter your change to no-op it :-).  It's because you didn't actually create any collections.  As we take this further, I found a good home for this test -- org.apache.solr.handler.TestConfigReload.  Yes it has the wrong base class; maybe you were inspired from this one originally.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -2736,6 +2736,12 @@ private Set<Runnable> getConfDirListeners(final String confDir) {
     return confDirListeners;
   }
 
+  public boolean hasConfDirectoryListeners(final String confDir) {
```
noblepaul, line 2739: so, this method was added just for the JUnit? do we not have any other way to know if there is a watch from Zookeeper client?

noblepaul:
`If you want to increase the scope to mark these things deprecated, that's be nice!`

I'll do that .

`so just move the test to the existing TestConfigReload `

Yes, please

PR #1215, opened 2022-12-06
DocRouter: strengthen abstraction

dsmiley:
DocRouter is an abstraction with 3 implementations; it isn't pluggable.  Years ago at work, I hacked a custom subclass of CompositeIdRouter.  In the process I found a number of spots, especially with splits, that were making assumptions of how CompositeIdRouter parsed IDs.  So here I'm strengthening the separation of concerns so that the split code can do its job, delegating to the DocRouter (specifically a CompositeIdRouter) on how to parse doc IDs.  It also makes CompositeIdRouter more extendable, but I am not truly making the DocRouter choice pluggable because perhaps it's too advanced and would take more work today to do that cleanly & tested.

This is an internal refactoring; not too big, so I was thinking no JIRA.

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/api/collections/MigrateCmd.java
+++ b/solr/core/src/java/org/apache/solr/cloud/api/collections/MigrateCmd.java
@@ -53,7 +53,6 @@ import org.apache.solr.common.params.ModifiableSolrParams;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.common.util.Utils;
 import org.apache.solr.handler.component.ShardHandler;
-import org.apache.solr.update.SolrIndexSplitter;
 import org.apache.solr.util.TimeOut;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -253,7 +252,7 @@ public class MigrateCmd implements CollApiCmds.CollectionApiCommand {
             SHARD_ID_PROP,
             sourceSlice.getName(),
             "routeKey",
-            SolrIndexSplitter.getRouteKey(splitKey) + "!",
+            sourceRouter.getRouteKeyNoSuffix(splitKey) + "!",
             "range",
             splitRange.toString(),
             "targetCollection",
@@ -283,7 +282,7 @@ public class MigrateCmd implements CollApiCmds.CollectionApiCommand {
       sourceSlice = sourceCollection.getSlice(sourceSlice.getName());
       Map<String, RoutingRule> rules = sourceSlice.getRoutingRules();
       if (rules != null) {
-        RoutingRule rule = rules.get(SolrIndexSplitter.getRouteKey(splitKey) + "!");
+        RoutingRule rule = rules.get(sourceRouter.getRouteKeyNoSuffix(splitKey) + "!");
         if (rule != null && rule.getRouteRanges().contains(splitRange)) {
           added = true;
           break;
--- a/solr/core/src/java/org/apache/solr/handler/admin/SplitOp.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/SplitOp.java
@@ -263,8 +263,9 @@ class SplitOp implements CoreAdminHandler.CoreAdminOp {
         DocCollection collection = clusterState.getCollection(collectionName);
         String sliceName = parentCore.getCoreDescriptor().getCloudDescriptor().getShardId();
         Slice slice = collection.getSlice(sliceName);
-        DocRouter router =
-            collection.getRouter() != null ? collection.getRouter() : DocRouter.DEFAULT;
+        CompositeIdRouter router =
+            (CompositeIdRouter)
+                (collection.getRouter() != null ? collection.getRouter() : DocRouter.DEFAULT);
         DocRouter.Range currentRange = slice.getRange();
 
         Object routerObj =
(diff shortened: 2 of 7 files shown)


noblepaul:
why can't we make it properly pluggable?

```
"router" : {"class": "fully.qualified.clasName"}
```

dsmiley:
We could; I just chose not to go that far in this PR.  If I did something trivial, I'd be inclined to then document it, test it.  If you think I could skip some of that (rationale: this is "expert" option only) then I'd be willing to do that in this PR.

noblepaul:
IMHO  this PR is mixing both the new capability and impl. If we could just have a pluggable option, then only experts would use it and that should be fine

dsmiley:
I can sympathize somewhat but the change/PR as it is strengthens a weak abstraction, even if it's not pluggable.  _Not all abstractions need to be pluggable_.  I could remove some comments added in the PR that may be suggestive of the possibility of a subclass that actually isn't present.

solr/core/src/java/org/apache/solr/handler/admin/SplitOp.java
```diff
@@ -263,8 +263,9 @@ private void handleGetRanges(CoreAdminHandler.CallInfo it, String coreName) thro
         DocCollection collection = clusterState.getCollection(collectionName);
         String sliceName = parentCore.getCoreDescriptor().getCloudDescriptor().getShardId();
         Slice slice = collection.getSlice(sliceName);
-        DocRouter router =
-            collection.getRouter() != null ? collection.getRouter() : DocRouter.DEFAULT;
+        CompositeIdRouter router =
```
noblepaul, line 266: is it just a formatting change? why is this required ?

dsmiley: Changed declared type to be specifically CompositeIdRouter because in fact splits depend on CompositeIdRouter in particular.  I added a method or two there as well that SplitOp calls.

solr/core/src/java/org/apache/solr/cloud/api/collections/MigrateCmd.java
```diff
@@ -253,7 +252,7 @@ private void migrateKey(
             SHARD_ID_PROP,
             sourceSlice.getName(),
             "routeKey",
-            SolrIndexSplitter.getRouteKey(splitKey) + "!",
+            sourceRouter.getRouteKeyNoSuffix(splitKey) + "!",
```
bruno-roustant, line 255: Above, in the existing code, there is a cast to CompositeIdRouter. It is not clear to me if it is known that the router must be of type CompositeIdRouter. Should we add a check like checkRouterSupportsSplitKey()?

solr/core/src/java/org/apache/solr/handler/admin/SplitOp.java
bruno-roustant: As I understand, the router was 'expected' to be a CompositeIdRouter, even if not clear here, because of the code below manipulating the terms and expecting to find a CompositeIdRouter.SEPARATOR.
It becomes clearer.
Should we add a check like checkRouterSupportsSplitKey() with a clearer exception?

solr/core/src/java/org/apache/solr/update/SolrIndexSplitter.java
```diff
@@ -765,18 +766,11 @@ static FixedBitSet[] split(
...
-    int idx = idString.indexOf(CompositeIdRouter.SEPARATOR);
-    if (idx <= 0) return null;
-    String part1 = idString.substring(0, idx);
-    int commaIdx = part1.indexOf(CompositeIdRouter.bitsSeparator);
-    if (commaIdx > 0 && commaIdx + 1 < part1.length()) {
-      char ch = part1.charAt(commaIdx + 1);
-      if (ch >= '0' && ch <= '9') {
-        part1 = part1.substring(0, commaIdx);
-      }
+  private static void checkRouterSupportsSplitKey(HashBasedRouter hashRouter, String splitKey) {
```
bruno-roustant, line 769: The expectation is much clearer with this method.
I'm not familiar with the other DocRouter. Does this mean that split is not supported at all with other DocRouter types?

bruno-roustant (approved):
I find it much clear to have all code related to CompositeIdRouter inside CompositeIdRouter itself (like key parsing). And we probably need even more checkRouterSupportsSplitKey().

bruno-roustant:
@noblepaul I don't fully understand your point RE _"Introducing a new type of Router needs more reviews"_.
To me this PR does not add a new type of DocRouter, as it is centered around CompositeIdRouter. It makes it clearer that only CompositeIdRouter is supported for split operations in SplitOp, SolrIndexSplitter, MigrateCmd. I see no new code, only some refactoring to move key parsing logic inside CompositeIdRouter. Do I miss something?

solr/core/src/java/org/apache/solr/update/SolrIndexSplitter.java
noblepaul: the other one is PLAIN router. I don't think Split is possible for that

noblepaul:
```To me this PR does not add a new type of DocRouter, as it is centered around CompositeIdRouter```

I was confused by the original description.  thought this was trying to introduce a new `DocRouter` by enhancing `CompositeIdRouter`

This PR is about  moving all the logic of routing /splitting into  `CompositeIdRouter`

solr/core/src/java/org/apache/solr/handler/admin/SplitOp.java
dsmiley: Correcting myself.  This method, `handleGetRanges`, is only called for splitByPrefix (as its javadocs say), which is what depends on CompositeIdRouter.

I suppose if there was some weird/unexpected code path (maybe in the future), a ClassCastException wouldn't be particularly unfriendly?

dsmiley:
I suppose this is ready to merge?  I'll do so in a couple days unless I hear there's more to do.

PR #1218, opened 2022-12-07
SOLR-16573:SolrClientTestRule for EmbeddedSolrServer

joshgog:
https://issues.apache.org/jira/browse/SOLR-16573

# Description

This PR introduces a SolrClientTestRule for EmbeddedSolrServer

# Solution

Introduced SolrClientTestRule abstract class that extends JUnit's ExternalResource class, which is a TestRule that allows for specifying methods to be run before and after a test. EmbeddedSolrServerTestRule is a concrete implementation of the abstract class and is declared as a ClassRule field in EmbeddedSolrServerTestBase. All subclasses of EmbeddedSolrServerTestBase use the rule to access SolrClient

# Tests

# Checklist


solr/test-framework/src/java/org/apache/solr/util/SolrClientTestRule.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+package org.apache.solr.util;
+
+import java.lang.invoke.MethodHandles;
+import org.apache.solr.client.solrj.SolrClient;
+import org.junit.rules.ExternalResource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SolrClientTestRule extends ExternalResource {
```
epugh, line 25: For classes that are going to have a long life, like this one in the `test-framework` starting to introduce some JavaDocs to explain what this class does would be great!

solr/core/src/test/org/apache/solr/update/processor/AbstractAtomicUpdatesMultivalueTestBase.java
```diff
@@ -31,41 +33,46 @@
...
-  private static void assertQR(
-      final String fieldName, final String queryValue, final int numFound) {
-    assertQ(
-        req("q", fieldName + ":" + queryValue, "indent", "true"),
-        "//result[@numFound = '" + numFound + "']");
+  private void assertQR(final String fieldName, final String queryValue, final int numFound)
+      throws SolrServerException, IOException {
+
+    SolrQuery query = new SolrQuery();
+    query.setParam("q", fieldName + ":" + queryValue);
```
dsmiley, line 73: note that SolrQuery has a constructor that would simplify this

solr/solrj/src/test/org/apache/solr/client/solrj/GetByIdTest.java
```diff
@@ -30,15 +35,19 @@ public class GetByIdTest extends EmbeddedSolrServerTestBase {
...
 
   @Before
   @Override
   public void setUp() throws Exception {
     super.setUp();
-    getSolrClient().deleteByQuery("*:*");
-    getSolrClient()
+    solrClientTestRule.getSolrClient().deleteByQuery("*:*");
+    solrClientTestRule
+        .getSolrClient()
```
dsmiley, line 50: Here and in many places in this file, you no longer call getSolrClient(), instead, more verbosely, you go to the solrClientTestRule first.  Is getSolrClient is still defined in EmbeddedSolrServerTestBase which is fine to continue to use?

solr/core/src/test/org/apache/solr/update/RootFieldTest.java
```diff
@@ -124,4 +132,8 @@ public void testUpdateWithChildDocs() throws Exception {
     }
     client.commit();
   }
+
+  public EmbeddedSolrServer getSolrClient() {
+    return solrClientTestRule.getSolrClient();
+  }
```
dsmiley, line 138: Isn't this already defined on EmbeddedSolrServerTestRule?  (if so, then we needn't specify this)

solr/solrj/src/test/org/apache/solr/client/solrj/embedded/LargeVolumeJettyTest.java
```diff
@@ -16,12 +16,17 @@
...
+import java.nio.file.Paths;
+import org.apache.solr.SolrJettyTestBase;
 import org.apache.solr.client.solrj.LargeVolumeTestBase;
 import org.junit.BeforeClass;
 
 public class LargeVolumeJettyTest extends LargeVolumeTestBase {
   @BeforeClass
   public static void beforeTest() throws Exception {
-    initCore();
+    solrClientTestRule
```
dsmiley, line 27: Let's add a TODO here.  AFAICT, long ago, @gerlowskija removed the Jetty aspect of this test, surely a mistake.

solr/solrj/src/test/org/apache/solr/client/solrj/request/SolrPingTest.java
```diff
@@ -32,17 +34,19 @@ public class SolrPingTest extends EmbeddedSolrServerTestBase {
   @BeforeClass
   public static void beforeClass() throws Exception {
     File testHome = createTempDir().toFile();
-    FileUtils.copyDirectory(getFile("solrj/solr"), testHome);
-    initCore("solrconfig.xml", "schema.xml", testHome.getAbsolutePath(), "collection1");
+    FileUtils.copyDirectory(SolrTestCaseJ4.getFile("solrj/solr"), testHome);
+
+    SolrTestCaseJ4.newRandomConfig();
+    // TODO NO COMMIT
+    solrClientTestRule.build().setSolrHome(testHome.toPath()).setSchemaFile("schema.xml").init();
```
dsmiley, line 41: We specify this as exactly `nocommit` (no space).

Any way, I think you are now in a position to fix it by simply removing the needless schema reference.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,273 @@
...
+
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  private EmbeddedSolrServer client = null;
+
+  public EmbeddedSolrServerTestRule() {}
+
+  public class Builder {
+    private Path solrHome;
+    private String schemaFile;
```
dsmiley, line 55: should have a default as well

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,273 @@
...
+    String schemaFile = b.getSchemaFile();
+    String configFile = b.getConfigFile();
+    String collectionName = b.getCollectionName();
+
+    if (b.getRequestWriterSupplier() != null) {
+      SolrConfig solrConfig;
+
+      try {
+        solrConfig = new SolrConfig(solrHome.resolve(collectionName), configFile);
+      } catch (Exception e) {
```
dsmiley, line 122: if it's already a RuntimeException, don't wrap it

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,273 @@
...
+      }
+
+      NodeConfig nodeConfig = buildTestNodeConfig(solrHome);
+
+      TestCoresLocator testCoreLocator =
+          new TestCoresLocator(
+              collectionName,
+              LuceneTestCase.createTempDir("data-dir").toFile().getAbsolutePath(),
+              solrConfig.getResourceName(),
+              IndexSchemaFactory.buildIndexSchema("schema.xml", solrConfig).getResourceName());
```
dsmiley, line 133: should use builder's config for the schema

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,273 @@
...
+    }
+  }
+
+  private void init(Builder b) {
+    Path solrHome = b.getSolrHome();
+    String schemaFile = b.getSchemaFile();
+    String configFile = b.getConfigFile();
+    String collectionName = b.getCollectionName();
+
+    if (b.getRequestWriterSupplier() != null) {
```
dsmiley, line 117: instead of if-else 3 branches, can't we always build basically one way?  Using the logic in this first branch.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,273 @@
...
+      client.close();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+    client.getCoreContainer().shutdown();
+  }
+
+  @Override
+  public EmbeddedSolrServer getSolrClient() {
+    return client;
```
dsmiley, line 188: I would assert that the client isn't null.  I doubt any test checks for null-ness; they assume it's set.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+  @Override
+  public void clearIndex() throws SolrServerException, IOException {
+    client.deleteByQuery("*:*");
```
dsmiley, line 193: if you can combine a delete & commit at the same time; it'd be perfect.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,273 @@
...
+              System.getProperty("collection", "collection1"),
+              CoreDescriptor.CORE_SHARD,
+              System.getProperty("shard", "shard1")));
+    }
+  }
+
+  // From TestHarness
+  private NodeConfig buildTestNodeConfig(Path solrHome) {
+    CloudConfig cloudConfig =
+        (null == System.getProperty("zkHost"))
```
dsmiley, line 239: Since EmbeddedSolrServer is not SolrCloud, you can assume "null" here.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,273 @@
...
+    attributes.put("name", "default");
+    attributes.put("class", SolrJmxReporter.class.getName());
+    PluginInfo defaultPlugin = new PluginInfo("reporter", attributes);
+    MetricsConfig metricsConfig =
+        new MetricsConfig.MetricsConfigBuilder()
+            .setMetricReporterPlugins(new PluginInfo[] {defaultPlugin})
+            .build();
+
+    return new NodeConfig.NodeConfigBuilder("testNode", solrHome)
+        .setUseSchemaCache(Boolean.getBoolean("shareSchema"))
```
dsmiley, line 267: TODO remove

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
dsmiley: also, this is so generic that it cold be the base impl in the superclass

dsmiley (commented):
So overall, I think the biggest impacting change to make to simplify the PR is to ensure that EmbeddedSolrServerTestBase continues to define getSolrClient so that you can then revert many of the changes in the tests that could simply continue to call it.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,273 @@
...
+import org.apache.solr.core.NodeConfig;
+import org.apache.solr.core.PluginInfo;
+import org.apache.solr.core.SolrConfig;
+import org.apache.solr.metrics.reporters.SolrJmxReporter;
+import org.apache.solr.schema.IndexSchemaFactory;
+import org.apache.solr.update.UpdateShardHandlerConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class EmbeddedSolrServerTestRule extends SolrClientTestRule {
```
epugh, line 45: I'd love to see some Javadocs that explain when you would want to use this TestRule.     I find it often confusing to figure out which type of unit test base calss to use...   Maybe text about that needs to also go in our dev-docs?

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,273 @@
...
+
+  private EmbeddedSolrServer client = null;
+
+  public EmbeddedSolrServerTestRule() {}
+
+  public class Builder {
+    private Path solrHome;
+    private String schemaFile;
+    private String configFile = "solrconfig.xml";
+    private String collectionName = "collection1";
```
epugh, line 57: In another PR, we are trying to use the constant for places where we say `collection1`, something like `DEFAULT_TEST_COLLECTION_NAME`....    I like it from the perspective that it cues me that the actual name of the collection doesn't matter to the behaviour of the test.

solr/test-framework/src/java/org/apache/solr/util/SolrClientTestRule.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.solr.client.solrj.SolrServerException;
+import org.junit.rules.ExternalResource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * SolrClientTestRule is an abstract class that extends JUnit's {@link ExternalResource} class,
+ * which is a TestRule that allows for specifying methods to be run before and after a test.
+ * SolrClientTestRule provides access to a {@link SolrClient} instance. SolrClientTestRule should be
+ * used in tests that need to interact with a Solr server and provides a convenient way to manage
```
epugh, line 32: Do we need more detail on Solr server?   I.e, does this support solr cloud mode?   Does this support http apis?

dsmiley: I don't know the "Solr server" reference you speak about?  Any way, this PR is an incremental step.  Obviously there is no Jetty/HTTP or SolrCloud stuff in this PR.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
dsmiley: Agreed; I held back on that feedback that came to my mind as well.  I think we want to get the API right before documenting said API.

solr/solrj/src/test/org/apache/solr/client/solrj/GetByIdTest.java
joshgog: Should I define getSolrClient() in EmbeddedSolrServerTestBase to return the client created in the test rule or I should revert to how it was defined initially?

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -53,7 +55,7 @@ public class Builder {
     private Path solrHome;
     private String schemaFile = "schema.xml";
     private String configFile = "solrconfig.xml";
-    private String collectionName = "collection1";
+    private String collectionName = DEFAULT_TEST_COLLECTION_NAME;
```
epugh, line 58: thanks!

solr/solrj/src/test/org/apache/solr/client/solrj/GetByIdTest.java
dsmiley: > Should I define getSolrClient() in EmbeddedSolrServerTestBase to return the client created in the test rule

Yes.

solr/test-framework/src/java/org/apache/solr/util/SolrClientTestRule.java
```diff
@@ -0,0 +1,56 @@
...
+ * the lifecycle of the SolrClient instance.
```
dsmiley, line 36: A good first start.  I recommend that you review the [Javadoc style guide](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#styleguide), especially around the importance of the first sentence and brevity generally.  I think the first sentence should be the one you specified 2nd, so I suggest the following (along with some other changes).
```
Provides access to a {@link SolrClient} instance and a running Solr in tests.
Implementations could run Solr in different ways (e.g. strictly embedded, adding HTTP/Jetty, adding SolrCloud, or an external process).
It's a JUnit {@link ExternalResource} (a {@code TestRule}), and thus closes the client and Solr itself when the test completes.
This test utility is encouraged to be used by external projects that wish to test communicating with Solr, especially for plugin providers.
```

solr/test-framework/src/java/org/apache/solr/util/SolrClientTestRule.java
```diff
@@ -0,0 +1,56 @@
...
+public abstract class SolrClientTestRule extends ExternalResource {
+
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  protected SolrClientTestRule() {}
+
+  @Override
+  protected void before() throws Throwable {
+    super.before();
+  }
```
dsmiley, line 47: can be removed

solr/test-framework/src/java/org/apache/solr/util/SolrClientTestRule.java
```diff
@@ -0,0 +1,56 @@
...
+
+  public abstract SolrClient getSolrClient();
+
+  public void clearIndex() throws SolrServerException, IOException {
```
dsmiley, line 51: needs docs

solr/test-framework/src/java/org/apache/solr/util/SolrClientTestRule.java
```diff
@@ -0,0 +1,56 @@
...
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  protected SolrClientTestRule() {}
+
+  @Override
+  protected void before() throws Throwable {
+    super.before();
+  }
+
+  public abstract SolrClient getSolrClient();
```
dsmiley, line 49: needs docs

solr/test-framework/src/java/org/apache/solr/util/SolrClientTestRule.java
```diff
@@ -0,0 +1,56 @@
...
+
+  public void clearIndex() throws SolrServerException, IOException {
+    new UpdateRequest().deleteByQuery("*:*").commit(solrClientTestRule.getSolrClient(), null);
+  }
+
+  public abstract Path getSolrHome();
```
dsmiley, line 55: needs docs.
BTW I recall there is some doubt as to wether we actually need this.  I suspect we don't?  If we can eliminate it -- even better.

solr/test-framework/src/java/org/apache/solr/util/SolrClientTestRule.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * SolrClientTestRule is an abstract class that extends JUnit's {@link ExternalResource} class,
+ * which is a TestRule that allows for specifying methods to be run before and after a test.
+ * SolrClientTestRule provides access to a {@link SolrClient} instance. SolrClientTestRule should be
+ * used in tests that need to interact with a Solr server and provides a convenient way to manage
+ * the lifecycle of the SolrClient instance.
+ */
+public abstract class SolrClientTestRule extends ExternalResource {
+
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
dsmiley, line 40: can be removed

solr/test-framework/src/java/org/apache/solr/util/SolrClientTestRule.java
```diff
@@ -0,0 +1,56 @@
...
+
+  @Override
+  protected void before() throws Throwable {
+    super.before();
+  }
+
+  public abstract SolrClient getSolrClient();
+
+  public void clearIndex() throws SolrServerException, IOException {
+    new UpdateRequest().deleteByQuery("*:*").commit(solrClientTestRule.getSolrClient(), null);
```
dsmiley, line 52: woah, this is suspicious -- the `solrClientTestRule.` part.  What's going on here?

solr/test-framework/src/java/org/apache/solr/util/SolrClientTestRule.java
```diff
@@ -0,0 +1,56 @@
...
+ * which is a TestRule that allows for specifying methods to be run before and after a test.
+ * SolrClientTestRule provides access to a {@link SolrClient} instance. SolrClientTestRule should be
+ * used in tests that need to interact with a Solr server and provides a convenient way to manage
+ * the lifecycle of the SolrClient instance.
+ */
+public abstract class SolrClientTestRule extends ExternalResource {
+
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  protected SolrClientTestRule() {}
```
dsmiley, line 42: can be removed

dsmiley (changes requested):
Just to be clear on the scope of this JIRA, it is only for this new EmbeddedSolrServerTestRule; no others (not Jetty, SolrCloud).  And it's not scoped to upgrade all tests using TestHarness; no, it's only tests that are already using EmbeddedSolrServer.

solr/solrj/src/test/org/apache/solr/client/solrj/embedded/LargeVolumeJettyTest.java
gerlowskija: 🤔 I don't remember doing that in particular, but I guess that'd be expected if it was an unintentional mistake 🤷 

You don't have a commit hash handy do you @dsmiley ?  (No worries if not; I can dig it up with git log/blame, was just hoping to save myself the trouble...)

solr/solrj/src/test/org/apache/solr/client/solrj/response/TestSpellCheckResponse.java
```diff
@@ -37,14 +40,18 @@ public class TestSpellCheckResponse extends EmbeddedSolrServerTestBase {
...
-    initCore();
+    solrClientTestRule
+        .build()
+        .setSolrHome(Paths.get(SolrJettyTestBase.legacyExampleCollection1SolrHome()))
+        .init();
   }
 
   static String field = "name";
 
+  SolrClient client = getSolrClient();
```
gerlowskija, line 51: [0] I don't think it's _wrong_ per se, but I'd rather see this in a method (maybe at the end of `beforeClass` or in a separate `@Before` annotated method?

Not a strong opinion though; if there's a reason you've done it this way then ignore me.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,187 @@
...
+    private String configFile;
+    private String schemaFile;
+    private RequestWriterSupplier requestWriterSupplier = RequestWriterSupplier.JavaBin;
+
+    public Builder setSolrHome(Path solrHome) {
+      this.solrHome = solrHome;
+      return this;
+    }
+
+    public Builder useTempDataDir() {
```
gerlowskija, line 57: [Q] Maybe trimming down the functionality exposed is beyond scope of this issue/PR, since at this point we need to  support the tests that exist currently.  But I am curious: what's the value of exposing this option to users?

Do tests actually care about where data lives, or is this just something that has historically been exposed to support tests that need multiple EmbeddedSolrServers?  If the latter, could we just always use a temp-dir and eliminate this method entirely?

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,187 @@
...
+      this.dataDir = LuceneTestCase.createTempDir("data-dir");
+      return this;
+    }
+
+    public Builder setSchemaFile(String schemaFile) {
```
gerlowskija, line 62: [0] Eric Pugh has been working recently on a Builder pattern for creating Solr clients that has methods with names like: `withFoo`, `withBar`, etc.

I love that you're consistent across this Builder here with using `set`.  But is there any value in being consistent with the SolrClient builders and using `with` instead of `set`?  (Maybe there's not - solrj and solr-test-framework are different animals.  Just bringing it up for conversation in case it hasn't been raised yet.)

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,187 @@
...
+import org.apache.solr.core.CoreDescriptor;
+import org.apache.solr.core.NodeConfig;
+import org.apache.solr.update.UpdateShardHandlerConfig;
+
+/** TODO NOCOMMIT document */
+public class EmbeddedSolrServerTestRule extends SolrClientTestRule {
+
+  private EmbeddedSolrServer client = null;
+
+  public Builder build() {
```
gerlowskija, line 40: [0] Definitely getting into "personal-preference territory" here, but I don't love the `build()/init()` method names that are used to create the builder and finish its initialization.

In most of the builders I've worked with (e.g. [ToStringBuilder](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/builder/ToStringBuilder.html#build--), [HttpClientBuilder](https://hc.apache.org/httpcomponents-client-4.5.x/current/httpclient/apidocs/org/apache/http/impl/client/HttpClientBuilder.html#build()), and of course [HttpSolrClient.Builder](https://solr.apache.org/docs/7_4_0/solr-solrj/org/apache/solr/client/solrj/impl/HttpSolrClient.Builder.html#build--)), `build()` is the final method in the call chain that takes all of the provided settings and uses them to actually create an instance of the object in question.

Using `build()` to create the builder itself breaks that common convention and (I think) will be confusing for readers/users.  Is there a good reason we're not just sticking to the common convention here?

gerlowskija (commented):
Leaving a few comments; though nothing serious.

Not involved enough in the tickets to 'Approve', but worth saying that I like almost all of what I see here.  Thanks for all the hard work Josh+David!

solr/solrj/src/test/org/apache/solr/client/solrj/embedded/LargeVolumeJettyTest.java
dsmiley: 319cb005d314e6f15dd73339e99657d8e39218a4

solr/solrj/src/test/org/apache/solr/client/solrj/response/TestSpellCheckResponse.java
dsmiley: Agreed on putting this at the end of beforeClass

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
dsmiley: It's rather common for tests to create Solr pointed at SOLR_HOME -- an existing directory that should be considered read-only (instead of some build/ or temp dir).  Tests will complain (via the SecurityManager) if this mistake is overlooked.  I could easily imagine plugin writers making use of this as well.  If the solr home provided is a temp dir populated from something, then there's no need for this.  It could be worth exploring creating one combo method like withMutableSolrHome vs withTemplateSolrHome (thus read-only) but I'm kind of "meh" on this.  Regardless, javadoc is necessary to help users understand when to use these methods.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
dsmiley: Should use "with" in general -- basically whatever reads well.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
dsmiley: Yeah I hear you on breaking precedent so maybe not a good idea.  How about initializeWithSolrHome(...).withBlahBlah().init();  We are "initializing" not "building"?  This idea standardizes on the Solr home coming first which is mandatory.

solr/solrj/src/test/org/apache/solr/client/solrj/request/TestCoreAdmin.java
```diff
@@ -46,20 +51,73 @@
...
+    System.setProperty("dataDir1", dataDir.getAbsolutePath());
+    System.setProperty("dataDir2", dataDir2.getAbsolutePath());
+    System.setProperty("tempDir", tempDir.getAbsolutePath());
+    System.setProperty("tests.shardhandler.randomSeed", Long.toString(random().nextLong()));
+    SolrTestCaseJ4.newRandomConfig();
+    solrClientTestRule.build().withSolrHome(SOLR_HOME).init();
+
+    solrClientTestRule
+        .newCollection(DEFAULT_TEST_COLLECTION_NAME)
+        .withConfigSet(CONFIG_HOME.resolve("../configsets").normalize().toString())
```
dsmiley, line 110: a configSet reference is a simple name, not a path (although it's on disk in a like-named directory).  To know which to use, look in `solr/solrj/src/test-files/solrj/solr/shared/core0/core.properties` which points to one named "shared".

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,225 @@
...
+    public NewCollectionBuilder withSchemaFile(String schemaFile) {
+      this.schemaFile = schemaFile;
+      return this;
+    }
+
+    public String getName() {
+      return name;
+    }
+
+    public String getConfigSet() {
```
dsmiley, line 136: Nobody is using this yet, thus this is probably why the test failed.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,225 @@
...
+    public void create() throws SolrServerException, IOException {
+      EmbeddedSolrServerTestRule.this.create(this);
+    }
+  }
+
+  private void create(NewCollectionBuilder b) throws SolrServerException, IOException {
+
+    client = new EmbeddedSolrServer(container, b.getName());
+
+    CoreAdminRequest.createCore(
```
dsmiley, line 157: I recommend inlining this so that you can set the configSet.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,225 @@
...
+        b.getName(),
+        container.getCoreRootDirectory().resolve(b.getName()).toString(),
```
dsmiley, line 159: can be simply `b.getName()` as it's relative to the core root dir.

solr/solrj/src/test/org/apache/solr/client/solrj/response/TermsResponseTest.java
```diff
@@ -31,16 +35,19 @@ public class TermsResponseTest extends EmbeddedSolrServerTestBase {
 
   @BeforeClass
   public static void beforeClass() throws Exception {
-    initCore();
+    solrClientTestRule.startSolr(LuceneTestCase.createTempDir("solrhome"));
+
+    solrClientTestRule
+        .newCollection(DEFAULT_TEST_COLLECTION_NAME)
+        .withConfigSet(ExternalPaths.TECHPRODUCTS_CONFIGSET)
+        .create();
```
dsmiley, line 43: Exemplifies we're not after shaving lines of code.  Starting Solr and creating a collection with a known configSet is explicit and should be -- isn't verbose either.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
dsmiley: @joshgog still waiting for docs in general.  Just give it a shot; I can clean up after.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,199 @@
...
+  public NodeConfig.NodeConfigBuilder newNodeConfigBuilder(Path solrHome) {
+    // TODO nocommit dedupe this with TestHarness
+    var updateShardHandlerConfig =
+        new UpdateShardHandlerConfig(
+            HttpClientUtil.DEFAULT_MAXCONNECTIONS,
+            HttpClientUtil.DEFAULT_MAXCONNECTIONSPERHOST,
+            30000,
+            30000,
+            UpdateShardHandlerConfig.DEFAULT_METRICNAMESTRATEGY,
+            UpdateShardHandlerConfig.DEFAULT_MAXRECOVERYTHREADS);
```
dsmiley, line 85: Let's create a constant declared at package scope so that TestHarness can refer to it, and we'll be done.  Actually, defining it on UpdateShardHandlerConfig would be nicer (debatable).

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,199 @@
...
+
+    return new NodeConfig.NodeConfigBuilder("testNode", solrHome)
+        .setUpdateShardHandlerConfig(updateShardHandlerConfig)
+        .setCoreRootDirectory(LuceneTestCase.createTempDir("cores").toString());
+  }
+
+  public NewCollectionBuilder newCollection(String name) {
```
dsmiley, line 92: Define on SolrClientTestRule as it's going to be common to all rules.  Obviously, the builder would tag along with it.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,199 @@
...
+    public NewCollectionBuilder withConfigSet(String configSet) {
+      // Chop off "/conf" if found -- configSet can be a path.
+      // This is a hack so that we can continue to use ExternalPaths.DEFAULT_CONFIGSET etc. as-is.
+      // TODO FileSystemConfigSetService.locateInstanceDir should have this logic.
+      // Without this, managed resources might be written to
+      // conf/conf/_schema_analysis_stopwords_english.json because SolrResourceLoader points to the
+      // wrong dir.
+      if (configSet != null && configSet.endsWith("/conf")) {
+        configSet = configSet.substring(0, configSet.length() - "/conf".length());
+      }
```
dsmiley, line 115: I'll tend to this improvement in another issue/PR.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+  }
+
+  @Override
+  public EmbeddedSolrServer getSolrClient() {
+    if (client == null) {
+      return adminClient;
+    } else {
+      return client;
+    }
```
dsmiley, line 188: Let's not assume the last call to create a collection is the default client (the `client` field).  Instead, let's establish that `getSolrClient` will always return a client to "collection1".  It should work even if there is such a core loaded by Solr in startSolr (i.e. no call to explicitly create a collection).  Don't return adminClient here either.

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  @Override
+  public EmbeddedSolrServer getSolrClient(String name) {
+    if (client == null) {
+      return new EmbeddedSolrServer(adminClient.getCoreContainer(), name);
+    } else {
+      return new EmbeddedSolrServer(client.getCoreContainer(), name);
+    }
```
dsmiley, line 197: always use adminClient; no point in flip-flopping.

solr/solrj/src/test/org/apache/solr/client/solrj/response/TermsResponseTest.java
```diff
@@ -31,16 +35,19 @@ public class TermsResponseTest extends EmbeddedSolrServerTestBase {
 
   @BeforeClass
   public static void beforeClass() throws Exception {
-    initCore();
+    solrClientTestRule.startSolr(LuceneTestCase.createTempDir("solrhome"));
```
epugh, line 38: Does this need the "solrhome" parameter?   When I see it, it makes me think it's important somehow..   I saw we also have the use of `.createTempDir()` with no name, is that better in the sense of conveying that it isn't important to have a named directory?

epugh:
I think I am slowly starting to grok what this is all about...   Let me see if I can state the value....   "Using a TestRule simplifies the setup of a test by abstracting away a lot of the Solr environment setup into the specific TestRule you choose.   This means choosing the right TestRule to use is an important decision when writing a test, to make sure you have the right abstraction of Solr you need"...???

solr/solrj/src/test/org/apache/solr/client/solrj/response/TermsResponseTest.java
dsmiley: > Does this need the "solrhome" parameter? When I see it, it makes me think it's important somehow

I suppose you are wondering, could we have a default of some temp directory so that the test doesn't even need to specify it?  We could do that.  I leaned to specifying it because this is such a foundational thing to Solr.

>  I saw we also have the use of .createTempDir() with no name

It's cosmetic so that if we debug issues, it's more evident what purpose this temp dir is for.

dsmiley:
> I think I am slowly starting to grok what this is all about... Let me see if I can state the value.... "Using a TestRule simplifies the setup of a test by abstracting away a lot of the Solr environment setup into the specific TestRule you choose. This means choosing the right TestRule to use is an important decision when writing a test, to make sure you have the right abstraction of Solr you need"...???

Yes!  One could argue we have similar things already -- EmbeddedSolrServer, JettySolrRunner, MiniSolrCloudCluster, and corresponding base classes.  But the look & feel is very different.  So the aim standardizes the approach between them so that using any of these requires little new knowledge.  This isn't evident in this PR as it's just one rule but there are more to create.

solr/core/src/test/org/apache/solr/update/RootFieldTest.java
```diff
@@ -54,7 +54,7 @@ public static void beforeTest() throws Exception {
     solrClientTestRule.startSolr(Path.of(SolrTestCaseJ4.TEST_HOME()));
 
     solrClientTestRule
-        .newCollection(DEFAULT_TEST_COLLECTION_NAME)
```
epugh: are we going away from using `DEFAULT_TEST_COLLECTION_NAME` everywhere?   I'm okay with that, just want to do it everywhere then ;-)

joshgog: That was a misunderstanding. We aren't going away from using going away from using `DEFAULT_TEST_COLLECTION_NAME`

solr/test-framework/src/java/org/apache/solr/util/EmbeddedSolrServerTestRule.java
dsmiley: @joshgog still pending use in TestHarness

joshgog: @dsmiley I already used it in TestHarness line 231  `.setUpdateShardHandlerConfig(UpdateShardHandlerConfig.TEST_DEFAULT)`

solr/solrj/src/test/org/apache/solr/client/solrj/request/json/DirectJsonQueryRequestFacetingEmbeddedTest.java
```diff
@@ -38,7 +39,6 @@
 @SuppressSSL
 public class DirectJsonQueryRequestFacetingEmbeddedTest extends EmbeddedSolrServerTestBase {
 
-  private static final String COLLECTION_NAME = "techproducts";
```
dsmiley, line 44: Why not leave this but merely change the value to collection1?  This way you avoid changing many lines below.

joshgog: Sure that's logical

dsmiley:
This is ready!  I'm looking to commit this Tuesday if I don't hear feedback to the contrary.

Suggested CHANGES.txt:
---------
SOLR-16573: Introduce EmbeddedSolrServerTestRule, a JUnit 4 TestRule for Solr testing.  (Joshua Ouma, David Smiley)

Suggested commit summary:
---------
New SolrClientTestRule abstraction with one implementation: EmbeddedSolrServerTestRule

EmbeddedSolrServerTestBase and AbstractEmbeddedSolrServerTestCase were modified to use this abstraction, and yet were also marked deprecated to encourage direct use of the rule.  A solr.xml file is optional.  Some of the affected tests will run more efficiently because (A) they no longer do file I/O to create a temporary solr-home because the rule is able to use existing solr homes and configSets in-place and because (B) initCore() can double-parse a solrconfig.xml and schema.xml.

SolrTestCaseJ4 / SolrTestCase:  Moved ObjectReleaseTracker checking to use a Rule so that it's compatible with Rules that use ObjectReleaseTracker.

solr/core/src/java/org/apache/solr/update/UpdateShardHandlerConfig.java
```diff
@@ -32,6 +32,15 @@ public class UpdateShardHandlerConfig {
           DEFAULT_METRICNAMESTRATEGY,
           DEFAULT_MAXRECOVERYTHREADS);
 
+  public static final UpdateShardHandlerConfig TEST_DEFAULT =
```
epugh, line 35: Shouldn't a TEST_DEFAULT be declared in some sort of testing class or in the Test Framework jar ???

solr/solrj/src/test/org/apache/solr/client/solrj/GetByIdTest.java
```diff
@@ -30,7 +35,9 @@ public class GetByIdTest extends EmbeddedSolrServerTestBase {
 
   @BeforeClass
   public static void beforeClass() throws Exception {
-    initCore();
+    solrClientTestRule.startSolr(LuceneTestCase.createTempDir("solrhome"));
```
epugh, line 38: Why do some use `solrClientTestRule.startSolr(Paths.get(SolrTestCaseJ4.TEST_HOME()));` and some need their own temp dir like in this?   Maybe this is really asking about some documentation somewhere on how to write good tests that use this Rule?

solr/solrj/src/test/org/apache/solr/client/solrj/LargeVolumeTestBase.java
```diff
@@ -41,7 +41,7 @@ public abstract class LargeVolumeTestBase extends EmbeddedSolrServerTestBase {
 
   @Test
   public void testMultiThreaded() throws Exception {
-    SolrClient client = this.getSolrClient();
+    SolrClient client = getSolrClient();
```
epugh, line 44: Should this be `solrClientTestRule.getSolrClient()` ?   When do we want to use the explicit `solrClientTestRule` and when do we have a helper like `getSolrClient()`.   In some ways, I like the explicit verbosity of the `solrClientTestRule.getSolrClient()` telling me where the client came from...

solr/solrj/src/test/org/apache/solr/client/solrj/embedded/AbstractEmbeddedSolrServerTestCase.java
```diff
@@ -104,4 +110,8 @@ protected SolrClient getSolrCore1() {
   protected SolrClient getSolrCore(String name) {
     return new EmbeddedSolrServer(cores, name);
   }
+
+  protected SolrClient getSolrAdmin() {
```
epugh, line 114: Cool...    So in many of our tests we need an admin client that doesn't have a core/collection name in the path, and then a collection specific one, and this solves that first need?

solr/solrj/src/test/org/apache/solr/client/solrj/embedded/LargeVolumeJettyTest.java
```diff
@@ -16,12 +16,17 @@
...
+import org.apache.lucene.tests.util.LuceneTestCase;
 import org.apache.solr.client.solrj.LargeVolumeTestBase;
+import org.apache.solr.util.ExternalPaths;
 import org.junit.BeforeClass;
 
 public class LargeVolumeJettyTest extends LargeVolumeTestBase {
   @BeforeClass
   public static void beforeTest() throws Exception {
-    initCore();
+    // TODO
```
epugh, line 27: left over or real?

solr/solrj/src/test/org/apache/solr/client/solrj/request/SolrPingTest.java
```diff
@@ -31,18 +30,18 @@ public class SolrPingTest extends EmbeddedSolrServerTestBase {
 
   @BeforeClass
   public static void beforeClass() throws Exception {
-    File testHome = createTempDir().toFile();
-    FileUtils.copyDirectory(getFile("solrj/solr"), testHome);
-    initCore("solrconfig.xml", "schema.xml", testHome.getAbsolutePath(), "collection1");
+    solrClientTestRule.startSolr(SolrTestCaseJ4.getFile("solrj/solr").toPath());
+
+    SolrTestCaseJ4.newRandomConfig();
+    solrClientTestRule.newCollection().withConfigSet("../collection1").create();
```
epugh, line 36: do we need relative pathing like this?   I don't think that I see this normally...    Here is an example of the more common pattern: `solrClientTestRule.newCollection().withConfigSet(ExternalPaths.TECHPRODUCTS_CONFIGSET).create();`

solr/solrj/src/test/org/apache/solr/client/solrj/request/SolrPingTest.java
```diff
@@ -31,18 +30,18 @@ public class SolrPingTest extends EmbeddedSolrServerTestBase {
...
   }
 
   @Before
   @Override
   public void setUp() throws Exception {
     super.setUp();
-    clearIndex();
-    assertU(commit());
-    assertU(optimize());
+    solrClientTestRule.clearIndex();
```
epugh, line 43: maybe a nit pick, but do we need to do this kind of clean up in our before's?   Shouldn't the overall framework take care of having a clean index?   I would hope that every test would start with an empty index right?

solr/solrj/src/test/org/apache/solr/client/solrj/response/TestSpellCheckResponse.java
```diff
@@ -34,17 +38,26 @@
  * @since solr 1.3
  */
 public class TestSpellCheckResponse extends EmbeddedSolrServerTestBase {
+  private static SolrClient client;
```
epugh, line 41: Do we need a static client?

epugh: Maybe this is partly to avoid changing so many lines of the test?

epugh:
This is really exciting..   I love seeing this effort to simplify how our tests are written.   I have found them at times to be quite difficult to understand, so this is a great step.

solr/solrj/src/test/org/apache/solr/client/solrj/LargeVolumeTestBase.java
joshgog: `getSolrClient()` in `EmbeddedSolrServerTestBase` returns `solrClientTestRule.getSolrClient()`

epugh: I get that..   What I'm wondering is that it appears that sometimes we use the helper pattern of a `getSolrClient()`, but in other tests we use the full explicit `solrClientTestRule.getSolrClient()`....    Feels like we should try and have one pattern?   Maybe just a bikeshed..

solr/solrj/src/test/org/apache/solr/client/solrj/embedded/AbstractEmbeddedSolrServerTestCase.java
joshgog: True. The client used by these tests do not need a core

solr/solrj/src/test/org/apache/solr/client/solrj/embedded/LargeVolumeJettyTest.java
dsmiley: @joshgog in the new Jetty based TestRule you are working on, lets address this issue above (a test that should be using Jetty but isn't any longer) on this test by using the new rule!  Basically, change LargeVolumeTestBase to subclass SolrTestCase instead of EmbeddedSolrServerTestBase, and instead declare whichever rule we want based on the test.

CC @epugh this is the TODO that wasn't clear

solr/core/src/java/org/apache/solr/update/UpdateShardHandlerConfig.java
dsmiley: That's debatable... because there is no obvious place where it would live exactly.  I suppose TestHarness *was* okay maybe but now that TestHarness has a competitor, ehh....

As it's just a "config" thing and not some heavy object, its home seems arbitrary to me -- not a problem that it's created in production.  I don't have a strong opinion though.

solr/solrj/src/test/org/apache/solr/client/solrj/GetByIdTest.java
dsmiley: Usually it's a reflection of whoever wrote the test and was maybe arbitrary.  I'm not sure we can strongly advocate one or the other.  I like the empty home dir insofar as the node configuration (usually read from solr.xml) is non-existent so  then the test infrastructure is in a position to tweak things for tests; perhaps being more minimalistic.  Perhaps a no-arg startSolr() would make this easy-path even easier and thus used more, which is a good thing.

solr/solrj/src/test/org/apache/solr/client/solrj/LargeVolumeTestBase.java
dsmiley: EmbeddedSolrServerTestBase is henceforth deprecated; it needn't exist; it barely does anything.  If someone wishes to remove it, then they can make that change then.  But those tests use this test base today so... why change them today.

solr/solrj/src/test/org/apache/solr/client/solrj/embedded/LargeVolumeJettyTest.java
dsmiley: See previous thread I at-mentioned you on.

solr/solrj/src/test/org/apache/solr/client/solrj/request/SolrPingTest.java
dsmiley: I concur that "I don't think that I see this normally".  But hey why not?  Your example looks nicer but it only applies to a well-known config.  Here in the SolrJ module, there are configSets in its own module that some tests use.  Would you propose creating a SolrJExternalPaths.COLLECTION1?  _Meh_.

solr/solrj/src/test/org/apache/solr/client/solrj/request/SolrPingTest.java
dsmiley: TestRules (a JUnit thing) can either be declared statically and thus become a so-called `@ClassRule` (they are annotated this way; see EmbeddedSolrServerTestBase in this PR), or non-statically declared rule is annotated `@Rule`.  ClassRule operate for the whole class; don't participate in life-cycle between individual tests in the class.  Rule is for each test.

So with that background, if a test needs a separate Solr instance per test, it could do so by defining this rule using `@Rule` and it would not need to bother clearing the index.

If a test does not need to restart Solr between tests (typical; why bother), then if a test also creates data, it'll need to arrange for that data to be cleaned up.  This is not a new concern; it exists today.  It's sometimes annoying to call clearIndex.  I could imagine some new mechanism where we create a new separate `@Rule` _from_ the `@ClassRule` that represents the data so that it cleans up so that it has the life-cycle hooks to call clearIndex.  If it works, it'll be one line of code in a test.  It'd be a pretty unusual / weird thing though.  :shrug:  could warrant an exploration.

solr/solrj/src/test/org/apache/solr/client/solrj/response/TestSpellCheckResponse.java
dsmiley: Indeed; some subclasses assumed this and we are restraining ourselves a bit from doing more work than needed.

solr/solrj/src/test/org/apache/solr/client/solrj/GetByIdTest.java
epugh: Maybe it's just my love for the Rails world's view of "convention over configuration"...    I'd love to see us strongly advocate for one approach whenever there are two approaches available because of the wide variety of folks working on Solr.  I suspect most folks are copy/pasting to decide how to do things, and aren't thinking carefully thorugh one option or the other.   If you provide two options that appear similarily good then you will live with two ways of doing it henceforth!   I'd love to see that `startSolr()` ;-)

solr/solrj/src/test/org/apache/solr/client/solrj/LargeVolumeTestBase.java
epugh: Right...   I think you suggested that...   So let's make sure to create a JIRA post merging this to get EmbeddedSolrServerTestBase removed ;-)   I'd be happy to sign up for that, removing deprecations is fun!

solr/solrj/src/test/org/apache/solr/client/solrj/request/SolrPingTest.java
epugh: Fair...   I have a old PR open where I tried to deal with that issue of the solrj module...

solr/solrj/src/test/org/apache/solr/client/solrj/request/SolrPingTest.java
epugh: Thanks for clarifying!

solr/solrj/src/test/org/apache/solr/client/solrj/GetByIdTest.java
```diff
@@ -35,7 +34,7 @@ public class GetByIdTest extends EmbeddedSolrServerTestBase {
 
   @BeforeClass
   public static void beforeClass() throws Exception {
-    solrClientTestRule.startSolr(LuceneTestCase.createTempDir("solrhome"));
+    solrClientTestRule.startSolr();
```
epugh, line 37: AWESOME!

dsmiley (commented):
@joshgog please see the failing "checks" above in this GitHub UI.  Ignore the Crave one; it's probably because this PR is out of date.  I recommend running `gw precommit` locally before committing next time.

PR #1220, opened 2022-12-07
SOLR-10463: setRetryExpiryTime should be deprecated in favor of Solr Client Builder methods

epugh:
https://issues.apache.org/jira/browse/SOLR-10463

# Description
Similar to a number of others, this moves the property to the Builder.

HOWEVER, one thing is that it appears in my searching the source that we do NOT use this property!   I'd love to have someone confirm this..  Is there a chance this should actually be deprecated and removed????

# Solution

did the migration, however maybe we rip it all out?

# Tests

# Checklist


dsmiley (changes requested):
You added to the Builder but never consumed it!  Do this in CloudSolrClient which is the base of the new & old impls.

You said we do not *use* this property but I believe you mean we do not *set* this property.  It is used/retrieved in CloudSolrClient.

epugh:
> You added to the Builder but never consumed it! Do this in CloudSolrClient which is the base of the new & old impls.
> 
> You said we do not _use_ this property but I believe you mean we do not _set_ this property. It is used/retrieved in CloudSolrClient.

I was about to post a screenshot showing that this value was never used, and finally found it in a method `shouldRetry`.    I duplicated the setting so if you use the builder you get it, if you don't you still get it.   Maybe when we remove the deprecated methods, these defaults will just livei n the builder?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudHttp2SolrClient.java
```diff
@@ -245,6 +249,16 @@ public Builder withParallelUpdates(boolean parallelUpdates) {
       return this;
     }
 
+    /**
+     * This is the time to wait to refetch the state after getting the same state version from ZK
+     *
+     * <p>secs
+     */
+    public Builder setRetryExpiryTime(int secs) {
```
dsmiley, line 257: No "with" terminology?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudHttp2SolrClient.java
```diff
@@ -245,6 +249,16 @@ public Builder withParallelUpdates(boolean parallelUpdates) {
       return this;
     }
 
+    /**
+     * This is the time to wait to refetch the state after getting the same state version from ZK
+     *
+     * <p>secs
```
dsmiley, line 255: Delete, I think

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudHttp2SolrClient.java
dsmiley: Can you incorporate TimeUnit as a parameter so that it's clearer what the unit is at the call site?
IntelliJ tip: do this by selecting "TimeUnit.SECONDS" on the line below and doing cmd-opt-P (introduce parameter)

solr/solrj/src/java/org/apache/solr/client/solrj/impl/SolrClientBuilder.java
```diff
@@ -34,6 +35,8 @@
   protected boolean useMultiPartPost;
   protected Integer connectionTimeoutMillis = 15000;
   protected Integer socketTimeoutMillis = 120000;
+  protected long retryExpiryTime =
```
dsmiley, line 38: SolrClientBuilder is for all the Apache HttpClient based SolrClients, but retryExpiryTime is specifically for ZK.  It's wrong to put here.

dsmiley (changes requested):

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudHttp2SolrClient.java
epugh: are you suggesting the method look like "public Builder setRetryExpiryTime(int secs, TimeUnit seconds)"

epugh: we do have the "int secs" or "int connectionTimeoutMillis" as parameters on various methods.   I wonder if we should have another JIRA to go through and make all timeunits on builders be the same????

solr/solrj/src/java/org/apache/solr/client/solrj/impl/SolrClientBuilder.java
epugh: Good catch...

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudHttp2SolrClient.java
epugh: @dsmiley I added https://issues.apache.org/jira/browse/SOLR-16590 to track standardizing across all builders the naming format of the methods.

epugh:
@dsmiley  @risdenk I think I am ready for another review after addressing feedback.  Can I get a LGTM?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudHttp2SolrClient.java
dsmiley: TimeUnit class was introduced in part to add clarity to call-sites of a method so the unit is clear.  `blah.setTime(TimeUnit.SECOND, 1)` is fine as well as `blah.setTime(TimeUnit.MINUTE,2)` -- the caller picks the unit convenient to them.  With that design, the method is designed unit-free -- definitely NOT with variables named "second" as you proposed since the unit could be anything.  Internally (implementation of the setter), we need to pick a unit to standardize to on some internal field to store the result, and name the field to be clear as to what the internal unit chosen is. (e.g. retryExpirySecs).  Again, that's *internal*, the caller choses a unit convenient to them.

epugh: I have added https://issues.apache.org/jira/browse/SOLR-16595 to do this over all builders...

PR #1221, opened 2022-12-07
SOLR-16577: Always log core load issues

heythm:
https://issues.apache.org/jira/browse/SOLR-16577


dsmiley:
FYI everyone, I already code reviewed this.

BTW a nice improvement here is that we remove needless complexity.  These Futures were pointless.  Perhaps there was once a point to them but I see no point now.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -973,29 +969,31 @@ public void load() {
...
+                  if (asyncSolrCoreLoad) {
+                    solrCores.markCoreAsNotLoading(cd);
+                  }
+                }
+                try {
+                  zkSys.registerInZk(core, true, false);
+                } catch (RuntimeException e) {
+                  log.error("Error registering SolrCore", e);
+                }
+                return core;
```
risdenk, line 995: Do we need to return anything? Since this isn't stored in a future?

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -973,29 +969,31 @@ public void load() {
...
+                } finally {
+                  solrCores.removeFromPendingOps(cd.getName());
+                  if (asyncSolrCoreLoad) {
+                    solrCores.markCoreAsNotLoading(cd);
+                  }
+                }
+                try {
+                  zkSys.registerInZk(core, true, false);
+                } catch (RuntimeException e) {
+                  log.error("Error registering SolrCore", e);
```
risdenk, line 993: Shouldn't this throw or at least do something besides just log?

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -973,29 +969,31 @@ public void load() {
...
+                SolrCore core;
+                try {
+                  if (zkSys.getZkController() != null) {
+                    zkSys.getZkController().throwErrorIfReplicaReplaced(cd);
+                  }
+                  solrCores.waitAddPendingCoreOps(cd.getName());
+                  core = createFromDescriptor(cd, false, false);
+                } catch (Exception e) {
+                  log.error("SolrCore failed to load on startup", e);
+                  return null;
```
risdenk, line 983: What does returning null do here?

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1004,24 +1002,9 @@ public void load() {
...
         coreContainerWorkExecutor.submit(
-            () -> {
-              try {
-                for (Future<SolrCore> future : futures) {
-                  try {
-                    future.get();
-                  } catch (InterruptedException e) {
-                    Thread.currentThread().interrupt();
-                  } catch (ExecutionException e) {
-                    log.error("Error waiting for SolrCore to be loaded on startup", e);
```
risdenk, line 1018: At least the old code was just logging on failure. I guess this was ok to not throw anything?

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -973,29 +969,31 @@ public void load() {
...
-                      }
-                    }
-                    try {
-                      zkSys.registerInZk(core, true, false);
-                    } catch (RuntimeException e) {
-                      SolrException.log(log, "Error registering SolrCore", e);
-                    }
-                    return core;
-                  }));
+          coreLoadExecutor.submit(
```
risdenk, line 972: Do we only need to do this in an executor if `asyncSolrCoreLoad`? Or if this is async should it use `coreContainerWorkExecutor`?

I'm a little surprised by looking at this diff there are two executors - `coreLoadExecutor` and `coreContainerWorkExecutor`?

risdenk (changes requested):

dsmiley (commented):
Just one little change needed.  I know you're not feeling well today @heythm; no hurry on this.

dsmiley:
We should also probably set & clear the MDC so that if we log an issue, we have more context (e.g. *which* core).

dsmiley:
While working on another issue & troubleshooting, I would have benefited from this improvement.  Some core initialization errors were getting swallowed.

risdenk:
@heythm hope all is well - curious of the comments will be addressed? I think this is a beneficial change

solr/CHANGES.txt
```diff
@@ -160,7 +160,7 @@ Other Changes
 
 SOLR-16575: splitshard should honour createNodeSet (noble)
 
-* SOLR-16577: Always log core load issues (Haythem Khiri)
+* SOLR-16577: Encore core load failures are always logged. (Haythem Khiri, David Smiley)
```
dsmiley, line 163:
```suggestion
* SOLR-16577: Ensure core load failures are always logged. (Haythem Khiri, David Smiley)
```

dsmiley:
@heythm , for a very old PR, first get it up to date with main (merge main into this) *then* update CHANGES.txt.  Doing CHANGES.txt too early is no fun merging.  Also remember there is a bunch of feedback here that is still unanswered.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
heythm: yes, hence no need to throw an exception at a lower level; in task body

dsmiley:
@risdenk I think your concerns have been addressed?

risdenk (approved):

mkgudla:
After upgrading to 9.5, we ran into issues while loading the Solr Cores.

We got interrupted exceptions and the cores didn't get loaded completely.

Later we saw that in this fix, the Future has been removed and the shutdownAndAwaitTermination is called in 60 seconds -- whereas the collections are taking more time to be loaded

PR #1222, opened 2022-12-07
SOLR-8975: Ensure poll time default of 250 is set on builder

risdenk:
https://issues.apache.org/jira/browse/SOLR-8975

Followup to fix ConnectionReuseTest.testConnectionReuse


epugh:
this makes sense...   I see how we had it on the main `ConcurrentUpdateSolrClient` originally.

Thanks for debugging this!

PR #1230, opened 2022-12-10
SOLR-16099: Upgrade to Jetty 10.0.13

risdenk:
https://issues.apache.org/jira/browse/SOLR-16099


epugh (approved):
LGTM..  I find these hard to review as in "it's a wall of straight up changes", but I guess the real value is having another set of eyes to see if a new dependency has snuck in!

Glad to get this in, and then we can start digging into the feasiblity of moving more to http2...

janhoy (approved):
LGTM. No-brainer I suppose, being a patch release.

New deps cannot “sneak” in unless a human adds License/Notice files manually. But if a new jetty includes a new jetty jar due to a split or something, then it will be covered by existing license file, and if the dep is optional we’d likely want to vet whether we need it or not.

Btw, I’be been testing Renovatebot on our repo for automated dependency PRs. Looks promising (after implementing a parser for consistent-versions plug-in first that is). Going to propose it on the dev list soon. Would you be supportive?

epugh:
> LGTM. No-brainer I suppose, being a patch release.
> 
> New deps cannot “sneak” in unless a human adds License/Notice files manually. But if a new jetty includes a new jetty jar due to a split or something, then it will be covered by existing license file, and if the dep is optional we’d likely want to vet whether we need it or not.
> 
> Btw, I’be been testing Renovatebot on our repo for automated dependency PRs. Looks promising (after implementing a parser for consistent-versions plug-in first that is). Going to propose it on the dev list soon. Would you be supportive?

anything that helps with reducing housekeeping and keeps us closer to the latest and greatest sounds good to me.

I suspect that making it easier to stay up to date would reduce the frequency of CVE reports we get as well ;-).  

I know that I have found in stewarding Quepid that falling behind makes the eventual upgrade that much harder...

PR #1232, opened 2022-12-10
SOLR-13243: Correct initial capacity of ops

heythm:
https://issues.apache.org/jira/browse/SOLR-13243

Quick improvement for initial ops capacity


madrob (commented):
Why don't we do List.of and not worry about the sizing?

heythm:
I think it's more readable this way, but it's still a preference.

stillalex:
I think @madrob's proposal fits better, a List.of() looks cleaner to me too.

risdenk (approved):

risdenk:
Thanks @heythm

PR #1233, opened 2022-12-10
SOLR-788: POC transfer MLT queries back and forth.

mkhludnev:
The way how distributed MLT works now lacks of perfection, you know. 
Here's not a real proposal, it just demonstrates current state. 

1. MLT respond with Query instance. However, it might be list of terms, or so. I keep it as-is
2. These queries should be responded to query coordinator. In HEAD it's just `BoolQ.toString()` that's a pity. There are two options instead: Lucene query can be structured as named list, or as a blob. I started from the former approach, but then decided to dump it to blob with Jackson+Smile. So, in the patch MLT query successfully arrives to coordinator. 
3. Then, we need to spread it across shards. And now it's prototyped in a little bit weird way. Since, it's deadly hard to send shard request with payload in body, I have to encode it to base64 and hook it up by dedicated QParser. Yeah... 

So, where to tweak it further?


alessandrobenedetti:
Hi Mikhail, thanks for looking into this!
I am currently in Japan with quite a tight schedule, but during the
Christmas holidays I'll be more than happy to review it!

PR #1238, opened 2022-12-13
Update caches-warming.adoc with a note about maxRamMB

EreMaijala:
Add a note about documentCache not supporting maxRamMB setting properly.

Since Lucene's RamUsageEstimator doesn't understand the Document objects, it will default to 256 bytes. Depending on the real document size this can be wildly inaccurate and cause the cache to use several magnitudes more memory than expected.


EreMaijala:
For reference, I had documentCache with `maxRamMB="128"` consume 4.9 GB of memory. A quick web search indicated that others have stumbled on this issue before as well, so a nice little note in the documentation might help someone avoid the pitfall.

janhoy (approved):

epugh:
Thank you for the doc improvement, sorry it took so long to merge!

PR #1239, opened 2022-12-13
SOLR-15787: Fix FileSystemConfigSetService test failure on Windows

NazerkeBS:
https://issues.apache.org/jira/browse/SOLR-15787


solr/core/src/java/org/apache/solr/core/FileSystemConfigSetService.java
```diff
@@ -238,9 +238,11 @@ public List<String> getAllConfigFiles(String configName) throws IOException {
           @Override
           public FileVisitResult visitFile(Path file, BasicFileAttributes attrs)
               throws IOException {
-            // don't include hidden (.) files
-            if (!Files.isHidden(file)) {
-              filePaths.add(configDir.relativize(file).toString());
+            String filePath = configDir.relativize(file).toString();
+            filePath = filePath.replace('\\', '/'); // normalize slashes
```
dsmiley, line 242: I think that backslash is `path.getFileSystem().getSeparator()`.  In theory, we could be running on some new system that uses some other symbol and this method would return it; it'd be found in the path.toString as the separator.

solr/core/src/java/org/apache/solr/core/FileSystemConfigSetService.java
```diff
@@ -238,9 +238,11 @@ public List<String> getAllConfigFiles(String configName) throws IOException {
...
+            // don't include .metadata.json hidden file
+            if (!filePath.startsWith(METADATA_FILE)) {
```
dsmiley, line 244: Why startsWith instead of equals?

dsmiley (changes requested):

solr/core/src/java/org/apache/solr/core/FileSystemConfigSetService.java
NazerkeBS: ohh it should be definitely `equals`

risdenk (changes requested):

solr/core/src/java/org/apache/solr/core/FileSystemConfigSetService.java
```diff
@@ -238,9 +238,14 @@ public List<String> getAllConfigFiles(String configName) throws IOException {
...
-            if (!Files.isHidden(file)) {
-              filePaths.add(configDir.relativize(file).toString());
+            Path filePath = configDir.relativize(file);
+            String filePathStr = filePath.toString();
+            filePathStr =
+                filePathStr.replace(
+                    filePath.getFileSystem().getSeparator(), "/"); // normalize slashes
+            // don't include .metadata.json hidden file
+            if (!filePathStr.equals(METADATA_FILE)) {
+              filePaths.add(filePathStr);
```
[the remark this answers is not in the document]
NazerkeBS: There might be other subdirs  which would have a separator, not only METADATA_FILE.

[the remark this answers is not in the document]
risdenk: Not sure I understand. In `visitFile` you end up with a file path. Why do we need to do anything about the slashes if we stick with Path? 

Why are we storing strings for file paths? We should be storing `Path` only and comparing paths. This avoids the whole string manipulation of file paths.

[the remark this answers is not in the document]
dsmiley: I hear you on embracing Path.  In this particular case, the ConfigSetService API (that is being implemented here) is to return a List<String> not a List<Path>.  The ConfigSetService cannot return Paths because the data isn't necessarily on the file system.  And I've never seen a Path/FileSystem implementation for ZK; it's hypothetical.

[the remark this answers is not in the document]
risdenk: Hmmm I see what you mean about `List<String>` I guess I'm confused maybe by the variable name and what is going on here. I haven't looked at the original PR or the whole change just related to this error.

I'll try to lay out my thoughts and maybe there is some misunderstanding I have.

* normalizing paths to `/` - if we do this - then on Windows this is no longer a filesystem path right? We are forcing these to be what? just strings to compare?
* In my head, we don't need to normalize paths - since `FileSystemConfigSetService` is run per node. Even if you had a mixed Windows/Linux Solr setup - each node (either Windows or Linux) would only care about file paths in its format (`/` or `\`). The strings being compared should just be comparable regardless of this "normalizing" - if its a path on windows - who cares if its `\` separating. 
* We are looking at filePathStr for `METADATA_FILE` - which should be just a `Path` comparison - regardless of what ends up in `filePaths` right?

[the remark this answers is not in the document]
risdenk: @NazerkeBS / @dsmiley any update here? Windows tests are still failing.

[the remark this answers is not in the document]
risdenk: I'm putting together a PR to show what I mean about using Path instead of string manipulation. I'll have it up shortly.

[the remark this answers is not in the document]
risdenk: Opened https://github.com/apache/solr/pull/1278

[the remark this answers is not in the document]
risdenk: Well https://github.com/apache/solr/pull/1278 fails since the test is testing both ZK and Filesystem. ZK always returns `/` and Filesystem doesn't. I'm not sure the actual string values matter but the test is trying to assert that ZK and Filesystem should return the same values. Not sure which is correct or not.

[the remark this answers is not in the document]
dsmiley: This ought to be documented in the method contract for getAllConfigFiles() better, although at least the existing docs give examples using `/` as a separator.  I see two users of this method:
* `org.apache.solr.core.backup.BackupManager#downloadConfigToRepo` which takes these and calls `org.apache.solr.core.backup.repository.BackupRepository#resolve` which isn't documented clearly on separators either.  Its file system based implementation assumes it's slash-compatible, which means it'll break if getAllConfigFiles() were to normalize them.
* `org.apache.solr.handler.configsets.UploadConfigSetAPI#uploadConfigSet` which will do two things:
  * compare the result with a zip entry file name.  I dug a little and[ Zip files should normalize to '/' separator](https://stackoverflow.com/a/44387973).
  * it'll call `org.apache.solr.core.ConfigSetService#deleteFilesFromConfig` which doesn't document this separator matter either.  

So it's a bit of a mess.  I think getAllConfigFiles() should normalize to '/' due to the generic nature of the abstraction (isn't specific to a local file system) and ensure that `org.apache.solr.core.backup.repository.BackupRepository#resolve` and  `org.apache.solr.core.ConfigSetService#deleteFilesFromConfig` translate a `/` in its input a file system native separator if there's a difference.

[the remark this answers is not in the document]
risdenk: I think I addressed these in 086ff2cf02e88757c18cf517491f4a740706e3c2 - added comments to the javadocs and cleaned up the places in question to either normalize or not to either '/' or the OS file path separator.

risdenk:
Tested the latest changes on Mac and Windows and both pass (`.\gradlew test --tests TestConfigSetService`)

solr/core/src/java/org/apache/solr/core/FileSystemConfigSetService.java
```diff
@@ -259,6 +260,14 @@ public FileVisitResult postVisitDirectory(Path dir, IOException ioException) {
     return filePaths;
   }
 
+  private String normalizePathToForwardSlash(String path) {
+    return path.replace(File.separatorChar, '/');
+  }
+
+  private String normalizePathToOSSeparator(String path) {
```
dsmiley, line 267: [Google Java Style guide](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case) would say "Os" end not "OS" here.

solr/core/src/java/org/apache/solr/core/FileSystemConfigSetService.java
```diff
@@ -259,6 +260,14 @@ public FileVisitResult postVisitDirectory(Path dir, IOException ioException) {
...
+    return path.replace('/', File.separatorChar);
```
dsmiley, line 268: Instead of referencing the old `java.io.File`, please use `configSetBase.getFileSystem().getSeparator()`

solr/core/src/java/org/apache/solr/core/FileSystemConfigSetService.java
```diff
@@ -250,7 +251,7 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs)
           public FileVisitResult postVisitDirectory(Path dir, IOException ioException) {
             String relativePath = configDir.relativize(dir).toString();
             if (!relativePath.isEmpty()) {
-              filePaths.add(relativePath + "/");
+              filePaths.add(normalizePathToForwardSlash(relativePath + File.separator));
```
dsmiley, line 254: shouldn't this always end in '/' ?   If we agree on this, then there's a gap in the tests.

solr/core/src/java/org/apache/solr/core/FileSystemConfigSetService.java
```diff
@@ -259,6 +260,14 @@ public FileVisitResult postVisitDirectory(Path dir, IOException ioException) {
     return filePaths;
   }
 
+  private String normalizePathToForwardSlash(String path) {
+    return path.replace(File.separatorChar, '/');
```
dsmiley, line 264: A short-circuit to do nothing if these are the same; would be nice but don't have to.

solr/core/src/test/org/apache/solr/core/TestConfigSetService.java
```diff
@@ -77,24 +72,31 @@ public void testConfigSetServiceOperations() throws IOException {
...
+    String file1 = "file1";
+    Files.write(configDir.resolve(file1), testdata);
+    String file2 = "file2";
+    Files.createFile(configDir.resolve(file2));
+    String subDirPath = "subdir";
+    String subdir = subDirPath + "/";
+    Files.createDirectory(configDir.resolve(subDirPath));
+    String file3 = subdir + "file3";
+    Files.createFile(configDir.resolve(subDirPath).resolve("file3"));
+    String file4 = subdir + "file4";
```
dsmiley, line 87: Ooops.
When I said to keep these changes from your trial PR, I wasn't paying attention at all.  I glanced I guess and thought you were converting from java.io.File to the Path API -- quite wrong and embarrassing on my part.  I think the test was easier to read before, honestly, because the assertions had the strings plainly without having to decipher what each variable added means.  **But it's fine; professionals can disagree.**

solr/core/src/test/org/apache/solr/core/TestConfigSetService.java
```diff
@@ -108,16 +110,14 @@ public void testConfigSetServiceOperations() throws IOException {
     assertTrue(configSetService.getConfigMetadata(configName).containsKey("foo"));
 
     List<String> configFiles = configSetService.getAllConfigFiles(configName);
-    assertEquals(
-        configFiles.toString(),
-        "[file1, file2, solrconfig.xml, subdir/, subdir/file3, subdir/file4]");
+    assertEquals(configFiles, List.of(file1, file2, solrConfigXml, subdir, file3, file4));
```
dsmiley, line 113: Perfect example of how it was so readable before and now... it's for example not evident that the output has slashes and what type.  Test code is not like main/prod code; different stylistic choices emphasizing easy readability above more idealistic decomposition.

dsmiley (changes requested):
Hey I'm glad you stepped up to take over -- I was about to do so tonight after you got the ball rolling yesterday.

solr/core/src/java/org/apache/solr/core/FileSystemConfigSetService.java
risdenk: yea I can fix this.

solr/core/src/test/org/apache/solr/core/TestConfigSetService.java
risdenk: yea I'll clean this up. I'm not really a fan of how it turned out.

solr/core/src/test/org/apache/solr/core/TestConfigSetService.java
risdenk: Yea agreed - I'll leave the list,of but will revert back to actual strings.

risdenk: Fixed in c445c08e2f31b18a8b047051a9cea5ca3f582bf1

solr/core/src/java/org/apache/solr/core/FileSystemConfigSetService.java
risdenk: So this was normalizing to `/` just semi convolutedly. Fixed in 836f33ad38a2bcca839e37d63ff35a78acbee404 and added a comment there to clarify.

solr/core/src/java/org/apache/solr/core/FileSystemConfigSetService.java
risdenk: Fixed in 836f33ad38a2bcca839e37d63ff35a78acbee404

solr/core/src/java/org/apache/solr/core/FileSystemConfigSetService.java
risdenk: Fixed in 836f33ad38a2bcca839e37d63ff35a78acbee404

mkhludnev (approved):
test passed on my windows laptop

dsmiley (approved):
+1 Thanks @risdenk for finishing this and @NazerkeBS for starting this!

No CHANGES.txt necessary because this is really just an update to an existing entry there which was never released.

PR #1240, opened 2022-12-14
SOLR-16584 BATS test fails stopping two instances

risdenk:
https://issues.apache.org/jira/browse/SOLR-16584

### DIFF
--- a/solr/packaging/test/test_start_solr.bats
+++ b/solr/packaging/test/test_start_solr.bats
@@ -31,7 +31,8 @@ teardown() {
 @test "SOLR11740 check f" {
   solr start
   solr start -p 7574
-  sleep 5
+  solr assert --started http://localhost:8983/solr --timeout 5000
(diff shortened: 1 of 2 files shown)


solr/packaging/test/test_start_solr.bats
```diff
@@ -31,6 +31,7 @@ teardown() {
 @test "SOLR11740 check f" {
   solr start
   solr start -p 7574
+  sleep 5
```
janhoy, line 34: You can assert solr is running, which will return status 0 immediately if solr is running, else wait.
```suggestion
  bin/solr assert --started http://localhost:8983/solr --timeout 5000
  bin/solr assert --started http://localhost:7574/solr --timeout 5000
```

risdenk:
@janhoy are you happy with this? I think its an improvement. I'm thinking get it merged and then see how things look.

janhoy (approved):

PR #1241, opened 2022-12-15
SOLR-16589: Large fields with large=true can be truncated when using unicode values

risdenk:
https://issues.apache.org/jira/browse/SOLR-16589


mkhludnev (approved):

dsmiley (commented):
Nice!

magibney:
Esp. given that a 9.1.1 release looks [inevitable](https://github.com/apache/solr/pull/1236), would you consider backporting this to `branch_9_1`?

risdenk:
> Esp. given that a 9.1.1 release looks https://github.com/apache/solr/pull/1236, would you consider backporting this to branch_9_1?

yup done and done :)

PR #1242, opened 2022-12-15
SOLR-16580: Avoid making copies of DocCollection for PRS updates

noblepaul:


noblepaul:
@justinrsweeney @patsonluk this is still WIP , reviews are welcome

solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java
```diff
@@ -488,4 +468,35 @@ public interface CollectionStateProps {
...
+      this.prs = prs;
+    }
+
+    @Override
+    public PerReplicaStates get() {
+      return prs;
+    }
+  }
+
+  public static final ThreadLocal<DocCollection.PrsSupplier> REPLICASTATES_PROVIDER =
```
hiteshk25, line 488: Quick question: is there any alternative to avoid threadlocal. Give me some more context on this

noblepaul: Yes, there is. The  only reason why I used the ThreadLocal Option was to minimize the changes. However, I think I should explore that route

solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java
```diff
@@ -139,30 +138,10 @@ public static String getCollectionPathRoot(String coll) {
    * only a replica is updated
    */
   public DocCollection copyWith(PerReplicaStates newPerReplicaStates) {
```
patsonluk, line 140: I assume we will need to modify `getSlices()` so it would return the correct "view" of slices from replica states too? Anyway, this is probably still WIP ! 😊

solr/solrj/src/java/org/apache/solr/common/cloud/ClusterState.java
```diff
@@ -261,9 +259,10 @@ private static DocCollection collectionFromObjects(
       if (log.isDebugEnabled()) {
         log.debug("a collection {} has per-replica state", name);
       }
-      // this collection has replica states stored outside
-      ReplicaStatesProvider rsp = REPLICASTATES_PROVIDER.get();
-      if (rsp instanceof StatesProvider) ((StatesProvider) rsp).isPerReplicaState = true;
+    } else {
+      // prior to this call, PRS provider is set. We should unset it before
+      // deserializing the replicas and slices
+      DocCollection.clearReplicaStateProvider();
```
patsonluk, line 265: To my understand this is required as otherwise the Provider might interfere and overrides the input values here?

I agree that using ThreadLocals could avoid modification of method signatures as you pointed out, but I also share similar concern as @hiteshk25 that it's a bit hard to track code flow with ThreadLocal as it requires "internal knowledge" of the code in order to know where things get added/modified.

This invocation of `clearReplicateStateProvider` could be one of the places that could be hard for dev that are not familiar with the ThreadLocal to understand.

I do understand the goal of this PR is NOT the removal of threadlocal usage 😊 , it would be nice though to consider other designs as a replacement of Threadlocal. That probably would include bigger changes (subclassing ClusterState that includes PRS, or adding overloading method etc). For the moment, more comments to explain the rational would be very helpful (which this comment also does a pretty good job, but perhaps it could also mention how the threadlocal PRS provider could override the values if not cleared ?)

patsonluk:
> @justinrsweeney @patsonluk this is still WIP , reviews are welcome

Thanks for the work @noblepaul ! Just want to confirm that there are mainly 2 goals here:
1. Avoid making copies of DocCollections when `copyWith` is called (related to PRS updates)
2. Avoid fetching PRS states until the state is actually queried (the Lazy PRS provider part)

Do we know the general overhead of the current designs and whether they are causing issues? I agree that both changes will for sure reduce resource usage!

solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java
noblepaul: Wy would slices change when PRS change?

solr/solrj/src/java/org/apache/solr/common/cloud/ClusterState.java
noblepaul: I'm aware of this problem . The ideal solution would be to pass on the `PrsSupplier` with the constructor. I'm trying to do that

solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java
patsonluk: Based on the old code, the modified replicas would be read from the newPerReplicaStates provided to construct a list as `modifiedShards` (which each Slice  value instead contains a map of replicas, which info relies on the input `newPerReplicaStates`).

Since we are not constructing a new DocCollection here and instead returning the same `this` instance, how can we ensure the such DocCollection instance getSlice calls return the updated replica info from `newPerReplicaStates`?

noblepaul: when per-replica states change the slices remain the exactly the same

noblepaul:
>Avoid making copies of DocCollections when copyWith is called (related to PRS updates)

Yes

> Avoid fetching PRS states until the state is actually queried (the Lazy PRS provider part)

This behavior is not changed. But what is changed is all states are **always** queried just in time. Prior to this , the states were copied into the object when it was constructed

solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java
patsonluk: By following the old code logic, it appears to be necessary to update the "replicas" stored within "Slice"?

For example if someone call docCollection.getSlice("shard1").getReplica("core_node_2"), the `Replica` stored in the `Slice` might not reflect the latest one from `newPerReplicaStates`  since they are all direct field and map lookup afaik :)

noblepaul:
@hiteshk25 @patsonluk I have gotten rid of the `ThreadLocal` completely

patsonluk:
> @hiteshk25 @patsonluk I have gotten rid of the `ThreadLocal` completely

Nice! I see that DocCollection has a new ctor which makes things much easier to follow! I think having a different ctor is probably a right step, though it might be hard to convince the Solr community to adapt that as they don't see PRS as something to be enabled by default so far? If it stays that way, we might want to introduce a subclass that extends `DocCollection` which has PRS support? (to avoid modification to the existing "default" `DocCollection`) But that's a much more refactoring...

I have also added https://github.com/cowpaths/fullstory-solr/commit/8a83fdfc20d4ae7dc832217ed8533aca3c0bb4ee to demonstrate the issue raised in https://github.com/apache/solr/pull/1242#discussion_r1054654494 . If we run that test on main branch it passes but on the current branch the replica within slice could miss the update.

solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java
noblepaul: The `DocCollection` , all the Slices & Replicas share the same `PrsSupplier`. When the PRS data changes, it's immediately visible to all the objects at once

patsonluk: The test here https://github.com/cowpaths/fullstory-solr/commit/8a83fdfc20d4ae7dc832217ed8533aca3c0bb4ee fails with the current branch as Slice and Replica are direct access to read from `Map` (as in https://github.com/cowpaths/fullstory-solr/blob/ef15b8615d01b9a4c00731c489c57dfe36621201/solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java#L233 and  https://github.com/cowpaths/fullstory-solr/blob/ef15b8615d01b9a4c00731c489c57dfe36621201/solr/solrj/src/java/org/apache/solr/common/cloud/Slice.java#L304). Unless the `Map` is a specialized version that knows to read from `PrsSupplier`?

noblepaul: The test was wrong. I have fixed it there.

patsonluk: Thanks i see what you mean now! 😅  I didn't notice the how the information was shared/mutated by the same `prsSupplier`.

I guess what confused me was that I have always assumed Slice/Replica to be immutable so it was unexpected on how the flow now works.

I do understand we are trying to avoid making copies here by reusing the existing Slice/Shard instances, and it might then become inevitable to mutate them. 

Wondering whether the existing design (immutable/making copies of the replica/shard on update) has caused any performance concern? Being immutable does have certain advantages - such as thread safety and ease to read the flow - considered that i just missed the shared prsSupplier details, this might not be obvious of how changes are shared, as such linkage is built at the ctor of `DocCollection`, which is outside of the scope of Slice/Replica

noblepaul: The objective is to closely map the new design with the way objects are created/modified/used

Before PRS, the `state.json` contained everything. So, once the `state.json` is loaded `DocCollection`, `Slice`, `Replica` Objects  remain immutable. After PRS, that assumption is no longer true. The optimisation that we did in ZK should reflect in the in memory structure as well.

> ... which is outside of the scope of Slice/Replica

Yes, the PRS data is itself outside the scope of `Slice`/`Replica`

solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java
```diff
@@ -465,7 +467,11 @@ public boolean isPerReplicaState() {
   }
 
   public PerReplicaStates getPerReplicaStates() {
-    return perReplicaStates;
+    return prsSupplier != null ? prsSupplier.get() : null;
```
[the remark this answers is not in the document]
noblepaul: > What is the purpose of fetching prs state lazily. 

PRS changes all the time. it has to be always fetched just in time

> would be good to make doccollection object immutable
 
We cannot make `DocCollection` immutable. it defeats the purpose. PRS states change too frequently. If we make make copies of these very large objects every-time a PRS entry is modified, we end up creating several times more objects than in non-PRS collections

solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java
```diff
@@ -488,4 +494,27 @@ public interface CollectionStateProps {
...
+      this.supplier = supplier;
+    }
+
+    public PrsSupplier(PerReplicaStates prs) {
+      this.prs = prs;
+    }
+
+    @Override
+    public PerReplicaStates get() {
+      if (prs == null) {
```
hiteshk25, line 514: what if multiple threads call this same time? do we see this can be fetched twice/more ?

solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java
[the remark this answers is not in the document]
hiteshk25: In general, who creates DocCollection Object ? it is inside some lock ?

solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java
noblepaul: yes. but it is OK as it doesn't cost much

solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java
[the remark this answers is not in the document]
noblepaul: why does it need a lock?

solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java
hiteshk25: Not sure it justify lazy loading. Better to initialize this in DocCollection'ctor, keep it simple. 
It seems like you want some delay in fetching prs state, if so we should consider that problem separately.

noblepaul: this object is constructed before `DocCollection` is constructed

hiteshk25: Sure. I think we want to avoid `LazyPrsSupplier` interface. So that we can construct immutable `DocCollection` object. I'm happy to file separate jira for it.

PR #1243, opened 2022-12-16
SOLR-16591 Make SolrCores pluggable.

dsmiley:
Remove transient core processing from SolrCores.  Move to a new subclass of SolrCores: TransientSolrCores. Remove TransientSolrCoreCacheFactory; not a useful intermediary.

Ideally move TransientSolrCoreCacheDefault into TransientSolrCores but not yet.

Improved/standardized visibility of members of SolrCores. Deprecated SolrCores.getCores().
SolrCores lists & lock should not have been static. SolrCores.getUnloadedCoreDescriptor isn't needed.

https://issues.apache.org/jira/browse/SOLR-16591

------
WIP... TestCoreDiscovery & TestLazyCores don't work.  Seems like a cleanup issue


solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -114,11 +94,8 @@ protected void close() {
       synchronized (modifyLock) {
         // make a copy of the cores then clear the map so the core isn't handed out to a request
         // again
-        coreList.addAll(cores.values());
+        coreList.addAll(getCores());
```
stillalex, line 97: if this is already holding 'modifyLock' does it still need to call getCores which also attempts to grab the lock?

solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -193,9 +165,9 @@ List<SolrCore> getCores() {
    * @return An unsorted list. This list is a new copy, it can be modified by the caller (e.g. it
    *     can be sorted).
    */
-  List<String> getLoadedCoreNames() {
+  public List<String> getLoadedCoreNames() {
     synchronized (modifyLock) {
-      return distinctSetsUnion(cores.keySet(), getTransientCacheHandler().getLoadedCoreNames());
+      return new ArrayList<>(cores.keySet());
```
stillalex, line 170: would you consider returning an immutable list here? (as in List.copyOf())

solr/core/src/test-files/solr/solr.xml
```diff
@@ -35,11 +35,6 @@
     <int name="connTimeout">${connTimeout:15000}</int>
   </shardHandlerFactory>
 
-  <transientCoreCacheFactory name="transientCoreCacheFactory" class="TransientSolrCoreCacheFactoryDefault">
```
noblepaul, line 38: TBH , I do not understand this abstraction.

Why is `transientCoreCacheFactory` an abstration?

We may have a `CoreFactory` or something like that as an abstraction and `TransientSolrCoreCacheFactory` could be one impl

dsmiley: TransientCoreCache was pluggable.  I don't think this is the right pluggable thing, so I removed its plug-ability.  Instead, I think "SolrCores" is better suited to be pluggable.  In so doing, the default/base SolrCores doesn't have to know about transient cores, and is thus simpler to understand for nearly everyone.  Businesses like my employer (in the definite minority) can deal with it in a different way than previously.

You propose a CoreFactory?  Today, cores are created in CoreContainer (as you know, I'm sure).  I'm trying to narrowly limit the scope of this PR to not be noticed by code that doesn't care about transient cores.

solr/core/src/java/org/apache/solr/core/SolrCores.java
dsmiley: There's no harm in synchronizing twice -- it's not worth trying to avoid it (e.g. perf).

solr/core/src/java/org/apache/solr/core/SolrCores.java
dsmiley: These methods are documented to return mutable copies, not immutable ones.  If I recall, maybe there were some callers that sorted the result?  I'd prefer immutable... not sure if we should change the contract right now (in 9.x); maybe.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1772,11 +1786,16 @@ private void resetIndexDirectory(CoreDescriptor dcore, ConfigSet coreConfig) {
...
+   * close.
+   *
+   * <p>NOTE: rather dangerous API because each core is not reserved (could in theory be closed).
+   * Prefer {@link #getLoadedCoreNames()} and then call {@link #getCore(String)} then close it.
    *
    * @return An unsorted list. This list is a new copy, it can be modified by the caller (e.g. it
-   *     can be sorted).
+   *     can be sorted). Don't need to close them.
    */
+  @Deprecated
```
dsmiley, line 1798: Deprecated because it's trappy -- cores are not incremented; perhaps they might be closed in the middle of a caller's processing.  It's not hard for callers to return a list of core names and fetch each calling close instead.  I didn't change most callers yet though.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -2284,17 +2303,6 @@ public boolean isLoaded(String name) {
...
   }
 
-  /**
-   * Gets a solr core descriptor for a core that is not loaded. Note that if the caller calls this
-   * on a loaded core, the unloaded descriptor will be returned.
-   *
-   * @param cname - name of the unloaded core descriptor to load. NOTE:
-   * @return a coreDescriptor. May return null
-   */
-  public CoreDescriptor getUnloadedCoreDescriptor(String cname) {
```
dsmiley, line 2294: wasn't useful

solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -36,60 +36,46 @@
...
-
   // This map will hold objects that are being currently operated on. The core (value) may be null
   // in the case of initial load. The rule is, never to any operation on a core that is currently
   // being operated upon.
-  private static final Set<String> pendingCoreOps = new HashSet<>();
+  private final Set<String> pendingCoreOps = new HashSet<>();
 
   // Due to the fact that closes happen potentially whenever anything is _added_ to the transient
   // core list, we need to essentially queue them up to be handled via pendingCoreOps.
-  private static final List<SolrCore> pendingCloses = new ArrayList<>();
```
dsmiley, line 62: a couple fields in here like modifyLock and this were static which made no sense so I removed that.

solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -99,37 +85,34 @@ protected void close() {
...
     // It might be possible for one of the cores to move from one list to another while we're
     // closing them. So loop through the lists until they're all empty. In particular, the core
     // could have moved from the transient list to the pendingCloses list.
-    do {
-      coreList.clear();
+    while (true) {
+
       synchronized (modifyLock) {
-        // make a copy of the cores then clear the map so the core isn't handed out to a request
-        // again
```
dsmiley, line 116: This comment is wrong, I think.  Clearing this.cores does not clear the descriptors; a request coming in will lazy load it.  This is finicky stuff so I opted to simply remove this comment instead of try to fix a possible existing bug that I've never seen.

solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -99,37 +85,34 @@ protected void close() {
...
+          if (core != null) {
+            coreList.add(core);
+          }
         }
 
         coreList.addAll(pendingCloses);
         pendingCloses.clear();
       }
 
+      if (coreList.isEmpty()) {
```
dsmiley, line 106: I changed the loop to exit at this point, which is more efficient and I think equally clear.

solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -99,37 +85,34 @@ protected void close() {
...
+        break;
+      }
+
       ExecutorService coreCloseExecutor =
           ExecutorUtil.newMDCAwareFixedThreadPool(
               Integer.MAX_VALUE, new SolrNamedThreadFactory("coreCloseExecutor"));
       try {
         for (SolrCore core : coreList) {
-          coreCloseExecutor.submit(
+          coreCloseExecutor.execute(
```
dsmiley, line 115: submit() returns a Future and we clearly weren't using it.
Inexplicably, this improvement appears to cause SolrMetricsIntegrationTest to fail because it double-closed cores.  I looked and made that test not do that.  I'm not sure why it worked before.  Maybe some fluke/accidental circumstance.

solr/core/src/java/org/apache/solr/core/SolrXmlConfig.java
```diff
@@ -379,8 +378,13 @@ private static NodeConfig fillSolrSection(NodeConfig.NodeConfigBuilder builder,
               case "replayUpdatesThreads":
                 builder.setReplayUpdatesThreads(it.intVal(-1));
                 break;
-              case "transientCacheSize":
-                builder.setTransientCacheSize(it.intVal(-1));
+              case "transientCacheSize": // TODO declare deprecated
```
dsmiley, line 381: TODO is to remember to state this in the release notes

solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -99,37 +85,34 @@ protected void close() {
...
-        // make a copy of the cores then clear the map so the core isn't handed out to a request
-        // again
-        coreList.addAll(cores.values());
-        cores.clear();
-        if (transientSolrCoreCacheFactory != null) {
-          coreList.addAll(getTransientCacheHandler().prepareForShutdown());
+        // remove all loaded cores; add to our working list.
+        for (String name : getLoadedCoreNames()) {
+          final var core = remove(name);
+          if (core != null) {
```
stillalex, line 97: not sure I follow the null check here. both 'getLoadedCoreNames' and 'remove' happen under lock, what is the scenario where this would be null? also, please reconsider the use of dedicated methods for basically 'keySet' and 'remove' with extra locking, it is adding a bunch of complexity to this code that is not needed.

solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -99,37 +85,34 @@ protected void close() {
     waitForLoadingCoresToFinish(30 * 1000);
     Collection<SolrCore> coreList = new ArrayList<>();
```
stillalex, line 86: what about moving the coreList definition inside the while block?

solr/core/src/java/org/apache/solr/core/SolrCores.java
```diff
@@ -99,37 +85,34 @@ protected void close() {
...
       ExecutorService coreCloseExecutor =
           ExecutorUtil.newMDCAwareFixedThreadPool(
               Integer.MAX_VALUE, new SolrNamedThreadFactory("coreCloseExecutor"));
       try {
         for (SolrCore core : coreList) {
-          coreCloseExecutor.submit(
+          coreCloseExecutor.execute(
               () -> {
                 MDCLoggingContext.setCore(core);
                 try {
```
stillalex, line 118: what do you think about adding the core name to the error log?

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -423,6 +424,20 @@ public CoreContainer(NodeConfig config, CoresLocator locator, boolean asyncSolrC
     this.allowListUrlChecker = AllowListUrlChecker.create(config);
   }
 
+  private SolrCores newSolrCores(NodeConfig config) {
```
stillalex, line 427: does this need the 'config' param? it could use the already existing 'this.cfg' (similar to the loader)?

solr/core/src/java/org/apache/solr/core/SolrCores.java
dsmiley: I'd be happy to do that; I debated that but your suggestion now motivates me :-)

solr/core/src/java/org/apache/solr/core/SolrCores.java
dsmiley: I suppose core can't be null; I just added the check to be sure.  I'll change it to an assertion.  
What complexity is in SolrCores.close() that isn't needed?  I suspect maybe you mean the total work that is done (i.e. adding code that these methods call), but I don't think that should _count_ in terms of complexity of this method -- being able to understand what it does.

If your point wasn't actually complexity but was about performance, I think it's a trivial matter.  Consider a SolrCores is closed once :-) and I disagree it'd be noticeably slower in any benchmark.

solr/core/src/java/org/apache/solr/core/SolrCores.java
dsmiley: Can you comment on the error log statement you refer to?  Note that the use of MDC will have this contextual information in the log automatically.

stillalex: you are correct,  please ignore the above

solr/core/src/java/org/apache/solr/core/SolrCores.java
stillalex: I meant cognitive complexity. I needed to bounce around a bit to verify what those methods do only to find that they are calling the Map methods (and so realized that the null check was not needed). what is the need for extra methods when they don't do anything else, makes the code more difficult to understand.

dsmiley: Firstly, I don't think SolrCore.close calling getLoadedCoreNames adds much cognitive complexity since its name is clear and it does what you'd think it does.  I think you are asking, why does `getLoadedCoreNames` exist at all if it's so trivial.  It existed before but the reason is to allow a subclass that could implement the concept of transient cores, which is the point of this PR.

stillalex: I agree with your reasoning, thanks for the clarifications!

solr/core/src/test-files/solr/solr.xml
noblepaul: honestly , a `CoreFactory` is a better abstraction. I know it's more work. But, it is better

dsmiley: I don't think it's an either-or, I think of this as more evolutionary.  Feel free to propose CoreFactory separately.  You haven't outlined your whole idea but I presume SolrCores would *become* CoreFactory.  Progress.

dsmiley: ?

This PR removes one "Factory" -- TransientCoreCacheFactory (the thing you just referred to).  I don't think it's needed at all.  But it doesn't remove TransientCoreCache, perhaps what you meant?  Regardless, I think of this thing as more of an implementation detail inside how the new TransientSolrCores works.  If you haven't reviewed this PR, please do before asking more questions.

I suspect your proposal is for there to be some abstraction that is responsible for creating the cores _and_ managing/holding them generally.  I think it makes sense for "SolrCores" to be that abstraction.  Maybe you didn't mean one abstraction for both.  I'm not sure it matters; maybe you should propose something somewhere.

noblepaul: I'm seeing a new pluggable component as follows
```
 <solrCores class="org.apache.solr.core.TransientSolrCores">
    <int name="transientCacheSize">4</int>
  </solrCores>
```

The name ` <solrCores>` actually doesn't make any sense. It would need a more descriptive name say `<coreFactory>` , `<coreManager>`  or something like that

dsmiley: `<coreManager>` works for me because its a reasonable name today and in the future if this class takes on core creation responsibility.

> The ticket doesn't even say that you are introducing a new type of component , what the configuration looks like or what the responsibilities are for that component type

From the JIRA issue: "This proposal would make SolrCores itself pluggable".  So there is no new type of component, and no change of responsibilities.  Yeah I could have shared what the plug-ability looks like.  I'm thinking this is so expert that we shouldn't really say anything about it in the ref guide.

noblepaul: As we are making `SolrCores` pluggable, we need to clearly define the following in JIRA itself:

- the responsibilities of the plugin
- the configuration of the plugin

dsmiley: There is no change of responsibilities here.  Agreed on the configuration; will do.

noblepaul: > There is no change of responsibilities here.

I didn't say there is a change in responsibilities. If it is not a pluggable component, the responsibilities do not matter and we can change them as often as we wish. Making it a pluggable component means we have to have a clear definition of the contract

dsmiley: I'm confused how I should satisfy the "clear definition of the contract", particularly for pre-existing things that are considered "expert" to plug in / not documented in the ref guide.

solr/core/src/java/org/apache/solr/core/SolrCores.java
dsmiley: The null check is actually needed, as I found TestCoreDiscovery.testTooManyTransientCores will fail.  It's vaguely explained in a comment here.  In more detail, transient core cache eviction can happen on any core manipulation, including the removal call.  So we might get core list A, B, C.  Then we remove A, triggering a cache maintenance resulting in B getting removed/evicted.  Then we try to remove B and it's already removed (and was put in pendingCloses).

solr/core/src/test-files/solr/solr.xml
noblepaul: `SolrCores` is a concrete class . I'm against making a concrete class pluggable. Lets start with an interface extracted from `SolrCores`

dsmiley: I could imagine your asks of me snow-balling to fully documenting every method.  I'd like to remind the context as to why this JIRA issue exists and its scope.  Today there is too much complexity around this code relating to transient cores -- a very niche feature that that is basically always enabled even though almost nobody configures cores to be transient.  I'm trying to ensure that the typical code-path doesn't involve any transient core stuff, and also not needlessly change APIs that don't need to change.  In other words, do a fairly narrow refactoring that is a positive step forward.  I don't think this should _require_ documenting existing things lacking documentation.  Not even splitting an interface honestly but I'll do it.

noblepaul: >  I don't think this should require documenting existing things lacking documentation.

I'm not asking for a ref guide section. The JIRA can have a basic definition of what this plugin means.

solr/core/src/java/org/apache/solr/core/TransientSolrCores.java
```diff
@@ -0,0 +1,197 @@
...
+      }
+    } else {
+      return super.putCore(cd, core);
+    }
+  }
+
+  @Override
+  public List<String> getLoadedCoreNames() {
+    synchronized (modifyLock) {
+      return distinctSetsUnion(
```
bruno-roustant, line 80: We can get rid of distinctSetsUnion and manipulate directly the List returned by the super method because it is a mutable copy. Something like:
```
if (getNumLoadedPermanentCores() == 0) {
  return new ArrayList<>(getTransientCacheHandler().getLoadedCoreNames());
}
List<String> coreNames = super.getLoadedCoreNames();
coreNames.addAll(getTransientCacheHandler().getLoadedCoreNames());
return coreNames;
```

solr/core/src/java/org/apache/solr/core/TransientSolrCores.java
```diff
@@ -0,0 +1,197 @@
...
+          super.getLoadedCoreNames(), getTransientCacheHandler().getLoadedCoreNames());
+    }
+  }
+
+  @Override
+  public List<String> getAllCoreNames() {
+    synchronized (modifyLock) {
+      return distinctSetsUnion(
```
bruno-roustant, line 88: Same as line 80.

solr/core/src/java/org/apache/solr/core/TransientSolrCores.java
```diff
@@ -0,0 +1,197 @@
...
+        return coreDescriptor;
+      }
+      return getTransientCacheHandler().getTransientDescriptor(coreName);
+    }
+  }
+
+  @Override
+  public List<CoreDescriptor> getCoreDescriptors() {
+    synchronized (modifyLock) {
+      List<CoreDescriptor> coreDescriptors = new ArrayList<>(getNumAllCores());
```
bruno-roustant, line 184: Same as line 80.

solr/core/src/java/org/apache/solr/core/TransientSolrCores.java
```diff
@@ -0,0 +1,197 @@
...
+      coreDescriptors.addAll(super.getCoreDescriptors());
+      coreDescriptors.addAll(getTransientCacheHandler().getTransientDescriptors());
+      return coreDescriptors;
+    }
+  }
+
+  /**
+   * @return the cache holding the transient cores; never null.
+   */
+  private TransientSolrCoreCache getTransientCacheHandler() {
```
bruno-roustant, line 194: Maybe inline?

bruno-roustant (approved):
I agree to not make SolrCores pluggable and to keep separating transient core cache from it.
I have some minor comments, but overall LGTM. Thanks David!

solr/core/src/java/org/apache/solr/core/TransientSolrCores.java
dsmiley: I pushed a change that did away with the distinctSetsUnion but did _not_ keep that optimization you have there on several lines.  I opted for simplicity over adding several lines and calling a particular metric name that isn't always obvious.  If you think the optimization is important, I can add it back in the way you did.

dsmiley:
I marked the pertinent classes deprecated, added a note about this in the upgrade notes, and added this CHANGES.txt entry:
> SOLR-16591: The "Transient Cores" feature is now deprecated.  In solr.xml "\<transientCoreCacheFactory\>"
  no longer works; it wasn't documented either.  Some internals were changed as well to simplify
  the standard case of having no transient cores.  (David Smiley)

Obviously the title & description of the JIRA issue (and this PR) should change accordingly.

dsmiley:
I plan to commit in a couple days if I don't hear to the contrary.

stillalex:
for what it's worth, I spent some time reviewing the latest version of the patch and it looks good to me.

PR #1245, opened 2022-12-19
SOLR-16567: KnnQueryParser support for both pre-filters and post-filter

alessandrobenedetti:
https://issues.apache.org/jira/browse/SOLR-16567

# Description

Frange and in general post filters were abandoned for KNN search in favour of pre-filters.
This PR brings back supports for post-filters when they have a cost>0 (in line with the rest of Solr)

# Tests

tests have been added

# Checklist

- [ X] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [ X] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
+++ b/solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
@@ -89,7 +89,7 @@ public class KnnQParser extends QParser {
       String[] filterQueries = req.getParams().getParams(CommonParams.FQ);
       if (filterQueries != null && filterQueries.length != 0) {
         try {
-          List<Query> filters = QueryUtils.parseFilterQueries(req, true);
+          List<Query> filters = QueryUtils.parseFilterQueries(req);
           SolrIndexSearcher.ProcessedFilter processedFilter =
               req.getSearcher().getProcessedFilter(filters);
           return processedFilter.filter;
--- a/solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
+++ b/solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
@@ -82,10 +82,10 @@ public class KnnQParser extends QParser {
     float[] parsedVectorToSearch = parseVector(vectorToSearch, denseVectorType.getDimension());
 
     return denseVectorType.getKnnVectorQuery(
-        schemaField.getName(), parsedVectorToSearch, topK, buildPreFilter());
+        schemaField.getName(), parsedVectorToSearch, topK, getFilterQuery());
   }
 
-  private Query buildPreFilter() throws SolrException {
+  private Query getFilterQuery() throws SolrException {
     boolean isSubQuery = recurseCount != 0;
     if (!isFilter() && !isSubQuery) {
       String[] filterQueries = req.getParams().getParams(CommonParams.FQ);
@@ -93,7 +93,7 @@ public class KnnQParser extends QParser {
         List<Query> preFilters;
 
         try {
-          preFilters = acceptPreFiltersOnly(QueryUtils.parseFilterQueries(req, true));
+          preFilters = excludePostFilters(QueryUtils.parseFilterQueries(req, true));
         } catch (SyntaxError e) {
           throw new SolrException(SolrException.ErrorCode.SERVER_ERROR, e);
         }
@@ -122,7 +122,7 @@ public class KnnQParser extends QParser {
    * @param filters filter queries in the knn query request
    * @return the list of pre-filters
    */
-  private List<Query> acceptPreFiltersOnly(List<Query> filters) {
+  private List<Query> excludePostFilters(List<Query> filters) {
     return filters.stream()
         .filter(
             q -> {
--- a/solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
+++ b/solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
@@ -53,7 +53,7 @@ public class KnnQParser extends QParser {
   }
 
   @Override
-  public Query parse() {
+  public Query parse() throws SyntaxError {
     String denseVectorField = localParams.get(QueryParsing.F);
     String vectorToSearch = localParams.get(QueryParsing.V);
     int topK = localParams.getInt(TOP_K, DEFAULT_TOP_K);
@@ -83,7 +83,7 @@ public class KnnQParser extends QParser {
         schemaField.getName(), parsedVectorToSearch, topK, getFilterQuery());
   }
 
-  private Query getFilterQuery() throws SolrException {
+  private Query getFilterQuery() throws SolrException, SyntaxError {
     boolean isSubQuery = recurseCount != 0;
     if (!isFilter() && !isSubQuery) {
       String[] filterQueries = req.getParams().getParams(CommonParams.FQ);
@@ -93,7 +93,7 @@ public class KnnQParser extends QParser {
           SolrIndexSearcher.ProcessedFilter processedFilter =
               req.getSearcher().getProcessedFilter(filters);
           return processedFilter.filter;
-        } catch (SyntaxError | IOException e) {
+        } catch (IOException e) {
           throw new SolrException(SolrException.ErrorCode.SERVER_ERROR, e);
         }
       }
--- a/solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
+++ b/solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
@@ -16,6 +16,8 @@
  */
 package org.apache.solr.search.neural;
 
+import java.util.List;
+import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.lucene.search.BooleanClause;
 import org.apache.lucene.search.BooleanQuery;
@@ -32,9 +34,6 @@ import org.apache.solr.search.QParser;
 import org.apache.solr.search.QueryParsing;
 import org.apache.solr.search.QueryUtils;
 import org.apache.solr.search.SyntaxError;
-
-import java.util.List;
-import java.util.stream.Collectors;
 
 public class KnnQParser extends QParser {
 
@@ -98,7 +97,7 @@ public class KnnQParser extends QParser {
         } catch (SyntaxError e) {
           throw new SolrException(SolrException.ErrorCode.SERVER_ERROR, e);
         }
-        
+
         if (preFilters.size() == 0) {
           return null;
         } else if (preFilters.size() == 1) {
@@ -116,14 +115,17 @@ public class KnnQParser extends QParser {
   }
 
   private List<Query> acceptPreFiltersOnly(List<Query> filters) {
-    return filters.stream().filter(q -> {
-      if (q instanceof ExtendedQuery) {
-        ExtendedQuery eq = (ExtendedQuery) q;
-        return eq.getCost() == 0;
-      } else {
-        return true;
-      }
-    }).collect(Collectors.toList());
+    return filters.stream()
+        .filter(
+            q -> {
+              if (q instanceof ExtendedQuery) {
+                ExtendedQuery eq = (ExtendedQuery) q;
+                return eq.getCost() == 0;
+              } else {
+                return true;
+              }
+            })
+        .collect(Collectors.toList());
   }
 
   /**
@@ -133,6 +135,7 @@ public class KnnQParser extends QParser {
    * @return a float array
    */
   private static float[] parseVector(String value, int dimension) {
+
     if (!value.startsWith("[") || !value.endsWith("]")) {
       throw new SolrException(
           SolrException.ErrorCode.BAD_REQUEST,
--- a/solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
+++ b/solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
@@ -16,7 +16,6 @@
  */
 package org.apache.solr.search.neural;
 
-import java.util.List;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.lucene.search.BooleanClause;
 import org.apache.lucene.search.BooleanQuery;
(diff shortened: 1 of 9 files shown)


solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
```diff
@@ -80,44 +82,60 @@ public Query parse() {
...
-        }
-
-        return builder.build();
       }
     }
-
     return null;
   }
 
+  private List<Query> acceptPreFiltersOnly(List<Query> filters) {
```
dsmiley, line 117: This method looks suspicious.  The cost is merely a hint; it isn't explicitly an indicator about being a post-filter or not.  Maybe I'm missing the point?

dsmiley: Look at the javadocs for PostFilter.  Basically `implements PostFilter` and `score >= 100`.

alessandrobenedetti: Thanks for the useful insight David! Is it any better now?

solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
```diff
@@ -80,44 +82,68 @@ public Query parse() {
     float[] parsedVectorToSearch = parseVector(vectorToSearch, denseVectorType.getDimension());
 
     return denseVectorType.getKnnVectorQuery(
-        schemaField.getName(), parsedVectorToSearch, topK, getFilterQuery());
+        schemaField.getName(), parsedVectorToSearch, topK, buildPreFilter());
   }
 
-  private Query getFilterQuery() throws SolrException {
-    if (!isFilter()) {
+  private Query buildPreFilter() throws SolrException {
```
dsmiley, line 88: IMO its name was just fine.  "pre-filter" is not something I've ever heard someone say in Solr.

solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
dsmiley: Can we name this method `excludePostFilters`?  I've never heard of "pre-filter" before.  Any part of a query that isn't a PostFilter is not necessarily executed in a defined order (is not "pre"); it's a combined query tree with interesting leap-frogging behavior based on various costs.

The logic appears correct but I'd like to recommend a simplification to make it even easier to read.  I suggest adding a static utility method on PostFilter called `boolean runAsPostFilter(Query q)`.  You could then simply make this a method reference.  Perhaps use this elsewhere (e.g. in SolrIndexSearcher.getProcessedFilter).

dsmiley (changes requested):
I'm looking closer at KnnQParser; I've never reviewed it before.  I see that it consumes the top-level request's `fq` params and it supplies them into the `new KnnVectorQuery(...)`.  I think this is a performance issue in that it _ignores_ Solr's caching / interpretation of 'fq' into the well-known filter cache.  I believe it ought to be using Solr's `SolrIndexSearcher.getProcessedFilter` method to get a combined Query of the parts, which is "cache-aware".  Other features of Solr are similar (notably QueryElevationComponent & grouping) so you can get inspiration / understanding by looking at some callers of that method.  Ultimately this should mean less code because getProcessedFilter is going to do most of the work, including understanding PostFilter.

A ramification of the lack of this use of getProcessedFilter is that typical filter queries are going to be processed twice -- once at the top level (perhaps retrieved from a cache), and again by this KNN stuff (not using a cache).

alessandrobenedetti:
> I'm looking closer at KnnQParser; ...

David, your help has been invaluable!
You are absolutely right and this is an oversight of mine when I did the original review of the pre-filtering work(this naming is quite common in the neural search community).

I think that I have found the perfect place for this fix and it literally would require few lines of code, rather than the complicate methods that are in place now:

org.apache.solr.search.QueryUtils#combineQueryAndFilter
`
...
} else if(scoreQuery instanceof KnnVectorQuery){
        (KnnVectorQuery)scoreQuery.setFilter(filterQuery);
      }
      else {
        return new BooleanQuery.Builder()
            .add(scoreQuery, Occur.MUST)
            .add(filterQuery, Occur.FILTER)
            .build();
      }
`
Basically when we combine the query with the filter, we manage the thing differently for KNN queries, and the filter(excluding post filters) is set in the Lucene KnnVectorQuery.

So far so good, elegant and minimal code change, filters are processed ones , everyone is happy...
BUT
currently KnnVectorQuery in Lucene has no getters and setters and has all variable as final!!
This is extremely annoying but we are where we are (to be honest, for a library class I would have gone with private variables with getters/setters since the beginning).

Also, now Lucene is a separate project, so I basically should do the change in Lucene, then wait for a Lucene release, include it in Solr ect ect 

So, long story short, my suggestion:
1) we proceed with the current hack, renaming where necessary to make it nicer, but no massive change
2) contribute the change Lucene side, probably removing final, and adding getters/setters or if the community disagree, just adding getters so that It's possible to create a new class from the input one
3) once Lucene releases and we have the code in Solr, do the nice and clean implementation

let me know what do you think!

alessandrobenedetti:
Ok, I updated the PR, this is what I have done:
1) workaround to fix the bug and renames suggested by @dsmiley 
2) opened a pull request in Lucene to implement getters in the KnnVectorQuery:  https://github.com/apache/lucene/pull/12029/files

Unless any additional good ideas, I would go with this now.
Then as soon as the Lucene side is sorted out and in Solr, I would implement the optimal approach, removing all the redundant code and just managing the filters in the org.apache.solr.search.QueryUtils#combineQueryAndFilter ( it will be so easy and clean, it's a shame I can't do it immediately)

dsmiley:
Thanks for your complements.

Can't you simply use SolrIndexSearcher#getProcessedFilter now?

As to your proposal, I am confused as to exactly where you propose inserting the logic you provided a snippet of.  If you propose SolrIndexSearcher somewhere then I don't like it because it's clearly special casing a specific query which is a design problem.  Are you trying to basically *move* certain FQs out of their top level position and into/embedded in a particular parsed query?

alessandrobenedetti:
Hi @dsmiley ,
once we have the Lucene changes, the idea is to change this method:
org.apache.solr.search.QueryUtils#combineQueryAndFilter
with an additional if clause:
`else {
        return new BooleanQuery.Builder()
            .add(scoreQuery, Occur.MUST)
            .add(filterQuery, Occur.FILTER)
            .build();
      }`
      
      will become:
      `else if{scoreQuery instanceof KnnVectorQuery}{
      return new KnnVectorQuery(scoreQuery,filterQuery);} 
      else{
        return new BooleanQuery.Builder()
            .add(scoreQuery, Occur.MUST)
            .add(filterQuery, Occur.FILTER)
            .build();
      }`

Just to give an idea, the final code will look different as we will have to create a new instance of KnnVectorQuery using the getters of the old one.
With this change, we will be able to simplify the KnnQueryParser removing all the stuff for pre-filters and post-filters.
GetProcessedFilters will be called just once as usual and we'll get the benefit of caching and post-0filter separation automatically.
SolrIndexSearcher won't be touched at all.

alessandrobenedetti:
And yes, in the workaround I can use the getProcessedFilters and I will, but once we have the Lucene side it will go away.

dsmiley:
A special case nearly anywhere (except directly in KNN oriented code of course) is a design/maintenance issue.  Some special cases like MatchAllDocs are understandable but a check for KNN in QueryUtils... eh... :-/    Maybe you could show in a new PR what this would look like so I could see.  Perhaps when I understand better what you are trying to accomplish, I'll see a better solution.

> (me:) Are you trying to basically move certain FQs out of their top level position and into/embedded in a particular parsed query?

Could you respond to that please?

alessandrobenedetti:
Ok, I'll produce another branch with the example code assuming Lucene changes are there.

So what I am trying to accomplish:
1) Knn Query has a Query filter as a constructor parameter (and instance variable). This filter is meant to be a pre-filter(in Approximate Nearest Neighbour search means a filter that is executed before the Top K nearest neighbors are returned).
It is used internally in the approximate nearest neighbor search Lucene code to only accept certain neighbors from the graph(along with the alive bitSet)
2) In Apache Solr we need to make sure that all filter queries except explicit post-filters are processed and set in the Knn Query.
3) But parsing happens before the Searcher will process the filters. So if are able to modify the Lucene KnnQuery in org.apache.solr.search.QueryUtils#combineQueryAndFilter (or create a new one), we are done.

Right now we process the filters at parsing time (ad do it again in the Searcher).
Potentially we could process and remove the filters from the request in the query parser, but it seems nasty to me, hence my idea of modifying the combineQueryAndFilter.
Because that method has the responsibility of building a new Query, combining the main Query and all filters(except post-filters).
So it seems the perfect place for implementing the custom logic for the KnnQuery, that behaves differently when you combine it with filters.

Hope this helps with context @dsmiley !

alessandrobenedetti:
here a rough pull request just to give you the idea @dsmiley :
https://github.com/apache/solr/pull/1246/files

alessandrobenedetti:
Refactored a bit, removed custom code and used getProcessedFilter.
It looks much better now, but still, the getProcessedFilter is done twice.
It's probably an acceptable workaround at the moment?
I created the tentative branch to show what it would be like when we have access to the internal of the KNN Query (which I still believe to be ideal).
Open to better ideas and thanks again for the review @dsmiley !

epugh:
Just wanted to comment what a great discussion this has been...  I learned some new stuff...

solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
```diff
@@ -84,30 +84,20 @@ public Query parse() {
...
-        List<Query> filters;
-
         try {
-          filters = QueryUtils.parseFilterQueries(req, true);
-        } catch (SyntaxError e) {
+          List<Query> filters = QueryUtils.parseFilterQueries(req, true);
+          SolrIndexSearcher.ProcessedFilter processedFilter =
+              req.getSearcher().getProcessedFilter(filters);
+          return processedFilter.filter;
+        } catch (SyntaxError | IOException e) {
```
dsmiley, line 96: Don't catch SyntaxError; propagate it. getFilterQuery will indirectly parse and this is in the context of query parsing; SyntaxError is a declared exception of QueryParser.parse.

solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
```diff
@@ -84,30 +84,20 @@ public Query parse() {
   }
 
   private Query getFilterQuery() throws SolrException {
-    if (!isFilter()) {
+    boolean isSubQuery = recurseCount != 0;
+    if (!isFilter() && !isSubQuery) {
```
dsmiley, line 88: Why is isFilter being checked?  Why is isSubQuery being checked?  In my experience, use of these is extremely rare.  isFilter... that would only be pertinent if the KNN query was specified in a filter query (`fq`), which is kind of surprising because AFAIK, KNN is for relevancy (scoring).  Does this query need to operate differently or to make more optimal decisions if it's in a filter query (to only filter and not rank)?

dsmiley (commented):
Overall I think this is superior.
Yes, Solr will wind up calling getProcessedFilter twice but typically, filter queries are cached, and so looking up cached queries is nearly free.  It's certainly better than the logic that was happening here before -- to actually evaluate those filter queries _every time_.

solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
alessandrobenedetti: Done, didn't notice in my original review of the code for 9.1 !
Thanks for pointing this out!

solr/core/src/java/org/apache/solr/search/neural/KnnQParser.java
```diff
@@ -84,30 +84,20 @@ public Query parse() {
...
       String[] filterQueries = req.getParams().getParams(CommonParams.FQ);
       if (filterQueries != null && filterQueries.length != 0) {
-        List<Query> filters;
-
         try {
-          filters = QueryUtils.parseFilterQueries(req, true);
-        } catch (SyntaxError e) {
+          List<Query> filters = QueryUtils.parseFilterQueries(req, true);
```
dsmiley, line 92: Boolean parameters are kind of sadly confusing in methods when it's not obvious what it is in code review tools (because the method name isn't a hint either).  Here, it is "fixNegativeQueries".  It should be rare to need to do that.  It's redundant/needless here because getProcessedFilter deals with the matter.

alessandrobenedetti: Ok,  now I remember this bit, I was initially (and I am still) not a fan of that boolean param (as I tend to forget what it means and the name also doesn't help).
We discussed for a better name(we couldn't find) and the reason to put it that was was because that piece of code was shared across various Solr parts and other reviewers thought it was better to isolate it and move it to the queryUtils.

Not sure of course, if it's still the best solution to be honest

dsmiley: I think this boolean could be removed (logic is to not call makeQueryable inside).  The only spot to be improved is then RealTimeGetComponent.  Simply update the spot that loops over the filters (RealTimeGetComponent.process line ~ 328 to ensure the query is "queryable".

alessandrobenedetti: done!

dsmiley (approved):
Just needs a CHANGES.txt

dsmiley:
I read your response to my question:
> (me:) Are you trying to basically move certain FQs out of their top level position and into/embedded in a particular parsed query?

You didn't give a simple yes-no answer but my interpretation is definitely a "yes".  Another possible approach is to make embedding specific filter queries very explicit -- put them in local-params of the KNN query parser *instead of* standard 'fq' at the top of the request.  This makes it clear there's no double-interpretation problem and it provides a great deal of control to the Solr user/dev.  But of course it's less friendly / more complex.  I don't know enough about KNN to know if certain expensive filter queries (that are not post-filters) might be better off happening ~after the KNN instead of "pre".  This is not supported by KNN today but could be.

gabrielmagno:
Just to bring a "data-science KNN user" view on this.

I think there is actually _two_ related but different ways one could use the Dense Vectors and KNN features. The first one is the most obvious and straightforward: you use it as KNN per si, i.e. you bring the top K documents most similar to the target. 

The second way to use the KNN is to use the _similarity score_ that is calculated with the vectors. In this case we use the vectors and their similarity for _ranking_ rather than for _retrieval_. And there are some use cases that we could even combine multiple similarity scores to create an aggregated score, or even combine the similarity score with the actual "lexical score" (BM25). For this second use case we often use a very high K to guarantee that we calculate the similarity score for _all_ the relevant documents.

So, considering this second use case of using the KNN for calculating the similarity score, having the ability to filter the documents based on the value of the similarity is very useful. And as previously mentioned, sometimes this is not even directly a single similarity score, it could be a combination of multiple scores, and we could apply the Post Filter on top of the aggregated score.

PR #1246, opened 2022-12-21
Jira/solr 16567 tentative

alessandrobenedetti:
just for @dsmiley to have the idea of what I meant


dsmiley:
I suppose this should be closed; you were merely showing me something?  I suppose this approach of showing somebody something can also be done on one's personal fork as well; probably better.

PR #1252, opened 2022-12-27
SOLR-16598: Upgrade Protobuf to 3.21.12

dsmiley:
from 3.21.8
Fixes some CVEs.

https://issues.apache.org/jira/browse/SOLR-16598

This is the latest patch version, thus compatible.  I spot-checked some [version notes](https://groups.google.com/g/protobuf/c/ONHxwxTfDxE) from the project for compatibility.

Note this fixes a possible issue of mixed versions.  We depended on protobuf-java-3.21.8 and protobuf-java-util-3.21.4.  I believe this was introduced by the Caffeine upgrade.

I didn't add a CHANGES.txt because I intend to add it to branch_9_1.  I believe post-release, the CHANGES.txt is synchronized out from this.  Any way, it'll be in "Other Changes" and look just like the issue title here, plus my name.


mkhludnev (approved):

risdenk:
So I'm not a fan of putting non Solr direct dependencies in versions.props - basically they will never get removed. We should be upgrading the libraries that use this transitive dependency - which will have the same outcome of using the upgraded protobuf.

PS `com.google.protobuf:protobuf-java*=3.21.12` would probably be better than what was added - this ensures that ALL `protobuf-java` dependencies use the same version instead of just the `protobuf-java-util` dependency.

Was out for the holidays so catching up on stuff.

dsmiley:
> So I'm not a fan of putting non Solr direct dependencies in versions.props - basically they will never get removed. We should be upgrading the libraries that use this transitive dependency - which will have the same outcome of using the upgraded protobuf.

Sure; I'm being pragmatic (trying to get something done *now*).  We can remove this eventually.

> com.google.protobuf:protobuf-java*=3.21.12

Cool!  BTW protobuf-java-util *and* the non-util do get upgraded to 3.21.12 because -util depends on the non-util of the same version :-). Any way, what you suggested is superior.

PR #1253, opened 2022-12-28
SOLR-10452: setQueryParams should be deprecated in favor of SolrClientBuilder methods

epugh:
https://issues.apache.org/jira/browse/SOLR-10452

# Description

setQueryParams should be deprecated in favor of SolrClientBuilder methods.

# Solution

I moved the the setQueryParams over for the Http2SolrClient, it already had been done for HttpSolrClient.   I noticed we have a `addQueryParams` method, which I marked deprecated and we shouldn't use, as that goes against the idea of a Solr Client being immutable.   

One area I tried to fix and gave up on was the `DelegationTokenHttpSolrClient`, I couldn't quite figure out what to do there, would love a suggestion or a fix ;-)

# Tests

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/update/UpdateShardHandler.java
+++ b/solr/core/src/java/org/apache/solr/update/UpdateShardHandler.java
@@ -148,7 +148,7 @@ public class UpdateShardHandler implements SolrInfoBean {
           .idleTimeout(cfg.getDistributedSocketTimeout())
           .maxConnectionsPerHost(cfg.getMaxUpdateConnectionsPerHost());
     }
-    updateOnlyClientBuilder.withQueryParams(queryParams);
+    updateOnlyClientBuilder.withTheseParamNamesInTheUrl(queryParams);
     updateOnlyClient = updateOnlyClientBuilder.build();
     updateOnlyClient.addListenerFactory(updateHttpListenerFactory);
 
--- a/solr/core/src/java/org/apache/solr/update/UpdateShardHandler.java
+++ b/solr/core/src/java/org/apache/solr/update/UpdateShardHandler.java
@@ -138,6 +138,9 @@ public class UpdateShardHandler implements SolrInfoBean {
         HttpClientUtil.createClient(
             clientParams, defaultConnectionManager, false, httpRequestExecutor);
 
+    Set<String> queryParams = new HashSet<>(2);
+    queryParams.add(DistributedUpdateProcessor.DISTRIB_FROM);
+    queryParams.add(DistributingUpdateProcessorFactory.DISTRIB_UPDATE_PARAM);
     Http2SolrClient.Builder updateOnlyClientBuilder = new Http2SolrClient.Builder();
     if (cfg != null) {
       updateOnlyClientBuilder
@@ -145,12 +148,9 @@ public class UpdateShardHandler implements SolrInfoBean {
           .idleTimeout(cfg.getDistributedSocketTimeout())
           .maxConnectionsPerHost(cfg.getMaxUpdateConnectionsPerHost());
     }
+    updateOnlyClientBuilder.withQueryParams(queryParams);
     updateOnlyClient = updateOnlyClientBuilder.build();
     updateOnlyClient.addListenerFactory(updateHttpListenerFactory);
-    Set<String> queryParams = new HashSet<>(2);
-    queryParams.add(DistributedUpdateProcessor.DISTRIB_FROM);
-    queryParams.add(DistributingUpdateProcessorFactory.DISTRIB_UPDATE_PARAM);
-    updateOnlyClient.setQueryParams(queryParams);
 
     ThreadFactory recoveryThreadFactory = new SolrNamedThreadFactory("recoveryExecutor");
     if (cfg != null && cfg.getMaxRecoveryThreads() > 0) {
--- a/solr/core/src/java/org/apache/solr/update/UpdateShardHandler.java
+++ b/solr/core/src/java/org/apache/solr/update/UpdateShardHandler.java
@@ -137,7 +137,7 @@ public class UpdateShardHandler implements SolrInfoBean {
         HttpClientUtil.createClient(
             clientParams, defaultConnectionManager, false, httpRequestExecutor);
 
-    Set<String> queryParams =
+    Set<String> urlParamNames =
         Set.of(
             DistributedUpdateProcessor.DISTRIB_FROM,
             DistributingUpdateProcessorFactory.DISTRIB_UPDATE_PARAM);
@@ -148,7 +148,7 @@ public class UpdateShardHandler implements SolrInfoBean {
           .idleTimeout(cfg.getDistributedSocketTimeout())
           .maxConnectionsPerHost(cfg.getMaxUpdateConnectionsPerHost());
     }
-    updateOnlyClientBuilder.withTheseParamNamesInTheUrl(queryParams);
+    updateOnlyClientBuilder.withTheseParamNamesInTheUrl(urlParamNames);
     updateOnlyClient = updateOnlyClientBuilder.build();
     updateOnlyClient.addListenerFactory(updateHttpListenerFactory);
 
--- a/solr/core/src/java/org/apache/solr/update/UpdateShardHandler.java
+++ b/solr/core/src/java/org/apache/solr/update/UpdateShardHandler.java
@@ -20,7 +20,6 @@ import static org.apache.solr.util.stats.InstrumentedHttpRequestExecutor.KNOWN_M
 
 import com.google.common.annotations.VisibleForTesting;
 import java.lang.invoke.MethodHandles;
-import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
@@ -138,9 +137,10 @@ public class UpdateShardHandler implements SolrInfoBean {
         HttpClientUtil.createClient(
             clientParams, defaultConnectionManager, false, httpRequestExecutor);
 
-    Set<String> queryParams = new HashSet<>(2);
-    queryParams.add(DistributedUpdateProcessor.DISTRIB_FROM);
-    queryParams.add(DistributingUpdateProcessorFactory.DISTRIB_UPDATE_PARAM);
+    Set<String> queryParams =
+        Set.of(
+            DistributedUpdateProcessor.DISTRIB_FROM,
+            DistributingUpdateProcessorFactory.DISTRIB_UPDATE_PARAM);
     Http2SolrClient.Builder updateOnlyClientBuilder = new Http2SolrClient.Builder();
     if (cfg != null) {
       updateOnlyClientBuilder
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
@@ -128,7 +128,7 @@ public class Http2SolrClient extends SolrClient {
   private static final List<String> errPath = Arrays.asList("metadata", "error-class");
 
   private HttpClient httpClient;
-  protected volatile Set<String> urlParamNames = Set.of();
+  private volatile Set<String> urlParamNames = Set.of();
   private int idleTimeout;
   private int requestTimeout;
 
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
@@ -128,7 +128,7 @@ public class Http2SolrClient extends SolrClient {
   private static final List<String> errPath = Arrays.asList("metadata", "error-class");
 
   private HttpClient httpClient;
-  protected Set<String> queryParams = Set.of();
+  protected Set<String> urlParamNames = Set.of();
   private int idleTimeout;
   private int requestTimeout;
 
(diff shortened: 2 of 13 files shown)


epugh:
Hey @stillalex would love a review.

epugh:
tests all pass!

solr/core/src/java/org/apache/solr/update/UpdateShardHandler.java
```diff
@@ -138,19 +138,19 @@ public UpdateShardHandler(UpdateShardHandlerConfig cfg) {
         HttpClientUtil.createClient(
             clientParams, defaultConnectionManager, false, httpRequestExecutor);
 
+    Set<String> queryParams = new HashSet<>(2);
+    queryParams.add(DistributedUpdateProcessor.DISTRIB_FROM);
+    queryParams.add(DistributingUpdateProcessorFactory.DISTRIB_UPDATE_PARAM);
```
stillalex, line 143: Could this block be queryParams = Set.of(...);

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -991,6 +992,7 @@ public static class Builder {
     private ExecutorService executor;
     protected RequestWriter requestWriter;
     protected ResponseParser responseParser;
+    private volatile Set<String> queryParams = Collections.emptySet();
```
stillalex, line 995: I don't think this needs to be 'volatile' anymore

stillalex:
I find LBHttp2SolrClient's use of queryParams a bit confusing (it seems to be on LBSolrClient level), apologies if this is just my incorrect understanding. Why does the LBSolrClient need to maintain a separate set of query params? LBHttp2SolrClient already receives a Http2SolrClient which ideally would have been init with correct query params. any (now deprecated) method to modify query params would just delegate to Http2SolrClient instance. I would cautiously suggest to remove the extra set of query params, which would also simplify a bit the code.

stillalex:
proposal for DelegationTokenHttpSolrClient: 
* move constructor 'setQueryParams(' call to Builder itself. the call comes from the builder already, so no added complexity https://github.com/apache/solr/blob/0f912f6b84c6b0a42079df2479ac6bbecc3c2c67/solr/solrj/src/java/org/apache/solr/client/solrj/impl/DelegationTokenHttpSolrClient.java#L40
* mark setQueryParams method deprecated
I think this should be it. in the future we can remove the DelegationTokenHttpSolrClient#setQueryParams method and the query params will already have the expected param because the builder injected it at build time.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java
```diff
@@ -661,12 +661,13 @@ private void verifyServletState(Http2SolrClient client, SolrRequest<?> request)
   public void testQueryString() throws Exception {
 
     final String clientUrl = jetty.getBaseUrl().toString() + "/debug/foo";
-    try (Http2SolrClient client = getHttp2SolrClient(clientUrl)) {
+    UpdateRequest req = new UpdateRequest();
+    setReqParamsOf(req, "serverOnly", "notServer");
```
dsmiley, line 665: it doesn't matter but its clearer to place this after client creation instead of before.  This is consistent with tests below.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java
```diff
@@ -325,9 +338,24 @@ public LBHttp2SolrClient.Builder setAliveCheckInterval(int aliveCheckInterval) {
...
 
+    /**
+     * Expert Method
+     *
+     * @param queryParams set of param keys that are only sent via the query string. Note that the
+     *     param will be sent as a query string if the key is part of this Set or the SolrRequest's
+     *     query params.
+     * @see org.apache.solr.client.solrj.SolrRequest#getQueryParams
+     */
+    public LBHttp2SolrClient.Builder withQueryParams(Set<String> queryParams) {
```
dsmiley, line 349: I didn't know about this thing before.  The javadocs are nice but I nonetheless find this confusing; I previously might have thought this was something like default SolrParams, but no.  The odd thing is that this doesn't hold the actual parameters, instead it's _names_ of parameters that only result in some change if the name-value is specified on the SolrRequest.    If this were `withUrlQueryParams` (my preference) or `withQueryStringParams`, it'd be clearer.   WDYT?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
dsmiley: definitely not volatile in a Builder!
Eric, I think you were merely copying the field definition in the client.  It's only volatile there because of the setters we are removing, in case two threads work with the client at the same time.

solr/core/src/java/org/apache/solr/update/UpdateShardHandler.java
epugh: it can be!   I think there are a lot of places we could do that...

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
epugh: we should make this set.of everywehre ;-)  And prevent the Collections.emptySet via forbidden apis...

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java
epugh: Yeah, I don't like the name either, and the confusion is real, witness the test that had `key=value` being used as query param!  instead of just `key`.    Any thoguhts on a name that really focuses on the fact that it's the key?   `withUrlQueryKeys` ???    or `withQueryKeys` ?

epugh:
> LBHttp2SolrClient

Would you be interested in pushing up a commit to make this change to remove the method?  I'd lvoe to see it!

epugh:
> * setQueryParams

Likewise @stillalex I think I get what you are suggesting, you want to push up a change for this?

epugh:
Okay, I think I've dealt with the small changes.   @stillalex I'd love you to push up changes for the two suggestions you had, they sound great.   @dsmiley maybe you can weigh in on the name of the field, and I'll make that change, and then this should be ready to merge!   Thanks for the feedback.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java
dsmiley: in the spirit of the Builder reading like English more naturally, how about `withTheseParamNamesInTheUrl`.  Yes it's long but who cares -- this is some expert method.

epugh: ROTFL...   But I can totally get behind that!

dsmiley: as you are showing willingness to change the name of fields in this PR (i.e. http2SolrClient), I would also really appreciate any field/parameter named "queryParams" to be something like "urlParamNames" so we are less likely to be confused at a glance with SolrParams.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -128,7 +128,7 @@ public class Http2SolrClient extends SolrClient {
   private static final List<String> errPath = Arrays.asList("metadata", "error-class");
 
   private HttpClient httpClient;
-  private volatile Set<String> queryParams = Collections.emptySet();
+  protected Set<String> queryParams = Set.of();
```
stillalex, line 131: I think this should remain 'volatile', see @dsmiley's comments on the builder's use of volatile.

epugh: Thanks for the sharp eyes.. yeah, I got moving too fast.

stillalex:
> > * setQueryParams
> 
> Likewise @stillalex I think I get what you are suggesting, you want to push up a change for this?

sure. how should we do this? push a proposed change to your branch directly for review?

epugh:
> > > * setQueryParams
> > 
> > 
> > Likewise @stillalex I think I get what you are suggesting, you want to push up a change for this?
> 
> sure. how should we do this? push a proposed change to your branch directly for review?

Yeah, I think you can checkout my PR, or maybe my `epugh/solr` repo?  Honestly, my Git Fu isn't great!   I am always very  happy to have other folks directly contribute to my branches ;-)   Casue generally I am bumbling along!

epugh:
> > > > * setQueryParams
> > > 
> > > 
> > > Likewise @stillalex I think I get what you are suggesting, you want to push up a change for this?
> > 
> > 
> > sure. how should we do this? push a proposed change to your branch directly for review?
> 
> Yeah, I think you can checkout my PR, or maybe my `epugh/solr` repo? Honestly, my Git Fu isn't great! I am always very happy to have other folks directly contribute to my branches ;-) Casue generally I am bumbling along!

With small concrete commits, I can always revert something ;-)

stillalex:
ok. I was not able to push to your git, not sure why, I don't think I have the rights. I pushed to my repo based on your branch. could you take a look and let me know if it looks ok? https://github.com/stillalex/solr/commit/0490ab420ff1d90536f3572fc7b17b194f0724c1

epugh:
odd, cause I think you have permissiosn to apache/solr repo?   Can you check out the PR from apache/solr instead of my branch from epugh/solr, and then push up changes????

stillalex:
> odd, cause I think you have permissiosn to apache/solr repo? Can you check out the PR from apache/solr instead of my branch from epugh/solr, and then push up changes????

I don't, I'm not a Solr committer :)

stillalex:
suggested fix for the DelegationTokenHttpSolrClient https://github.com/stillalex/solr/commit/5086e0fe099d5c2fa5942c6b3f18b7d803370136 . 

The constructor code looks a bit iffy to me. does it really overwrite any existing param from the builder to set this single param? it can't be correct.
```
setQueryParams(new TreeSet<>(Arrays.asList(DELEGATION_TOKEN_PARAM)));
```

epugh:
@stillalex I learned somethign new!  added you as another repo, and was able to cherry-pick your commit!

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -128,7 +128,7 @@ public class Http2SolrClient extends SolrClient {
   private static final List<String> errPath = Arrays.asList("metadata", "error-class");
 
   private HttpClient httpClient;
-  private volatile Set<String> queryParams = Collections.emptySet();
+  protected volatile Set<String> urlParamNames = Set.of();
```
stillalex, line 131: I don't think this needs to be protected anymore (following latest changes)

stillalex:
I've seen you only partially took changes from https://github.com/stillalex/solr/commit/0490ab420ff1d90536f3572fc7b17b194f0724c1 was there anything failing? or was the change too drastic?

epugh:
> I've seen you only partially took changes from [stillalex@0490ab4](https://github.com/stillalex/solr/commit/0490ab420ff1d90536f3572fc7b17b194f0724c1) was there anything failing? or was the change too drastic?

Huh...   I don't know why my cherry pick of that commit didn't bring it all?   I am going to manually review and copy over now ;-(

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
stillalex: @epugh bumping this comment up for visibility

epugh: thank you!

epugh:
Okay, are we done?

stillalex:
yep, looks good to me!

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java
```diff
@@ -82,25 +83,25 @@
  * @since solr 8.0
  */
 public class LBHttp2SolrClient extends LBSolrClient {
-  private final Http2SolrClient httpClient;
+  private final Http2SolrClient http2SolrClient;
```
dsmiley, line 86: I suggest `solrClient` instead -- it's shorter and it need not specify `http2` in it.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClient.java
```diff
@@ -1041,6 +1052,15 @@ public HttpSolrClient build() {
...
+        urlParamNames =
+            urlParamNames == null
+                ? Set.of(DelegationTokenHttpSolrClient.DELEGATION_TOKEN_PARAM)
+                : urlParamNames;
+        if (!urlParamNames.contains(DelegationTokenHttpSolrClient.DELEGATION_TOKEN_PARAM)) {
+          urlParamNames = new HashSet<>(urlParamNames);
+          urlParamNames.add(DelegationTokenHttpSolrClient.DELEGATION_TOKEN_PARAM);
+          urlParamNames = Collections.unmodifiableSet(urlParamNames);
+        }
         return new DelegationTokenHttpSolrClient(this);
```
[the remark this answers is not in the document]
stillalex: I don't disagree, and I like the Set methods more. but if you look at the impl, the UnmodifiableCollection is only a thin wrapper over the original collection, so I'm not sure this type of optimizations is really that useful in the end.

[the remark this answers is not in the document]
dsmiley: My intention is only to make it obvious reading the code that urlParamNames will be immutable in the end.  That's all.

PR #1259, opened 2022-12-31
SOLR-16602: tweaks to package manager docs to make following along easier

epugh:
https://issues.apache.org/jira/browse/SOLR-16602

# Description

While trying to understand how PackageManager api works, I went through this page.

# Solution

Fix some of the examples and make it easier to use the ref guide "copy" command.

# Tests
manually built

### DIFF
--- a/solr/solr-ref-guide/modules/configuration-guide/pages/package-manager-internals.adoc
+++ b/solr/solr-ref-guide/modules/configuration-guide/pages/package-manager-internals.adoc
@@ -46,7 +46,7 @@ Start all your nodes with the system property `-Denable.packages=true` to use th
 [source,bash]
 ----
 
-$ bin/solr -c -Denable.packages=true
+bin/solr -c -Denable.packages=true
 ----
 
 === Upload Your Keys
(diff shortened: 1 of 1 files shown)


epugh:
I removed lots of `$` from the bash commands, which solves the copy function, otherwise every copy and paste into a terminal would start with a `$` that you have to remove!  I wish Antora would just automtically prefix any bash commands in the UI with a $, but not have it in the copy version.  Any ideas on how to make that happen @ctargett ??

risdenk:
> I wish Antora would just automtically prefix any bash commands in the UI with a 

this isn't necessary. `$` is just a terminal prompt display thing. Its not necessary and my terminal doesn't just have `$` in the prompt. I'm all for removing the `$` to make copy/paste work well.

risdenk (approved):

epugh:
> > I wish Antora would just automtically prefix any bash commands in the UI with a
> 
> this isn't necessary. `$` is just a terminal prompt display thing. Its not necessary and my terminal doesn't just have `$` in the prompt. I'm all for removing the `$` to make copy/paste work well.

Cool, I created https://issues.apache.org/jira/browse/SOLR-16610 to track this.

PR #1260, opened 2023-01-01
SOLR-16682: transfer mlt query for component right

mkhludnev:
https://issues.apache.org/jira/browse/SOLR-788
I think this is a somewhat reliable solution for the problem with distributed mlt component. Reminder, now it naively relies on parsing `booleanQuery.toString()` that's unacceptable.

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/MoreLikeThisHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/MoreLikeThisHandler.java
@@ -459,8 +459,7 @@ public class MoreLikeThisHandler extends RequestHandlerBase {
     /**
      * Yields terms with boosts from the boosted MLT query.
      *
-     * @param maxTerms how many terms to return, a negative value means all terms are
-     *     returned
+     * @param maxTerms how many terms to return, a negative value means all terms are returned
      */
     public List<InterestingTerm> getInterestingTerms(BooleanQuery boostedMLTQuery, int maxTerms) {
       assert boostedMLTQuery != null : "strictly expecting it's set";
--- a/solr/core/src/java/org/apache/solr/handler/MoreLikeThisHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/MoreLikeThisHandler.java
@@ -52,7 +52,6 @@ import org.apache.solr.handler.component.ResponseBuilder;
 import org.apache.solr.request.SimpleFacets;
 import org.apache.solr.request.SolrQueryRequest;
 import org.apache.solr.response.SolrQueryResponse;
-import org.apache.solr.schema.IndexSchema;
 import org.apache.solr.schema.SchemaField;
 import org.apache.solr.search.DocIterator;
 import org.apache.solr.search.DocList;
@@ -395,15 +394,11 @@ public class MoreLikeThisHandler extends RequestHandlerBase {
         int id, int start, int rows, List<Query> filters, List<InterestingTerm> terms, int flags)
         throws IOException {
       Document doc = reader.document(id);
-      rawMLTQuery = mlt.like(id);
-      boostedMLTQuery = getBoostedQuery(rawMLTQuery);
-      if (terms != null) {
-        fillInterestingTermsFromMLTQuery(boostedMLTQuery, terms);
-      }
+      final Query boostedQuery = getBoostedMLTQuery(id, terms);
 
       // exclude current document from results
       BooleanQuery.Builder realMLTQuery = new BooleanQuery.Builder();
-      realMLTQuery.add(boostedMLTQuery, BooleanClause.Occur.MUST);
+      realMLTQuery.add(boostedQuery, BooleanClause.Occur.MUST);
       realMLTQuery.add(
           new TermQuery(
               new Term(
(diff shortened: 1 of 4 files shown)


solr/core/src/java/org/apache/solr/handler/MoreLikeThisHandler.java
```diff
@@ -422,6 +417,24 @@ public DocListAndSet getMoreLikeThis(
...
     }
 
+    public List<InterestingTerm> getInterestingTerms(int docid) throws IOException {
+      final ArrayList<InterestingTerm> interestingTerms = new ArrayList<>();
+      getBoostedMLTQuery(docid, interestingTerms);
+      return interestingTerms;
+    }
+    /**
+     * Sets {@link #boostedMLTQuery} and return, also put interesting terms into terms list if
+     * provided non-null. Sorry. It lacks of perfection and overall sense.
```
dsmiley, line 427: LOL.... so it can help make it less ugly to name the parameter outputTerms or something like that.  I've also seen a strategy where the return value is a Pair or something like that.  There is such a class in SolrJ.

solr/core/src/java/org/apache/solr/handler/component/MoreLikeThisComponent.java
```diff
@@ -127,6 +128,53 @@ public void process(ResponseBuilder rb) throws IOException {
     }
   }
 
+  private static NamedList<String> putMLTIntoParamList(
```
dsmiley, line 131: We strive to match the Google Java Style guide, and that means using "Mlt" and not "MLT" inside names.  But maybe MLT is so pervasive capitalized this way in the code already (before striving for this) that you are trying to be consistent?  Any way; do as you wish.

dsmiley (commented):
I just did a *super* quick review looking at some things that stood out to me.  Really Alessandro should review this.

solr/core/src/java/org/apache/solr/handler/MoreLikeThisHandler.java
mkhludnev: Got it. This collection also let to bypass terms extraction. I'm not sure it allows to save a lot of computation, I prefer to keep it lazy. I tries to toss code a little to improve. The root cause of this ugliness is lack of proper public method in Lucene. It's worth to decouple terms extraction, field assignment and query factoring. It will let to boost clauses without boolean query ripping.

solr/core/src/java/org/apache/solr/handler/component/MoreLikeThisComponent.java
```diff
@@ -127,6 +128,53 @@ public void process(ResponseBuilder rb) throws IOException {
...
+      StringBuilder reuseStr,
+      CharsRefBuilder reuseChar) {
+    reuseStr.setLength(0);
+    if (boost != 1f) {
+      reuseStr.append("{!boost b=");
+      reuseStr.append(boost);
+      reuseStr.append("}");
+    }
+    final String field = term1.field();
+    final CharsRef val =
```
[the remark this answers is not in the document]
mkhludnev: Thanks for suggestion, but I don't think it often matters much for field name.

solr/core/src/java/org/apache/solr/handler/component/MoreLikeThisComponent.java
mkhludnev: We have a hundred of occurrences of `MLT`, it might be great idea, but out of scope.

mkhludnev: ok. I put it as `mltViaQueryParams`

alessandrobenedetti (approved):
I have done a deep review and this pull request seems reasonable to me.
It's true that it mixes a nice refactor of the interesting terms extraction and communication with the shards (effectively solving two problems at once).
But I believe it's fine.
I left some minor comments that we should probably address, but not blockers

solr/core/src/test/org/apache/solr/handler/component/DistributedMLTComponentTest.java
```diff
@@ -389,5 +389,29 @@ public void test() throws Exception {
...
+        "mlt.fl",
+        "lowerfilt1,lowerfilt",
+        "mlt.qf",
+        "lowerfilt1^1.2 lowerfilt^3.4",
+        "qt",
+        requestHandlerName,
+        "shards.qt",
+        requestHandlerName,
+        "mlt.count",
+        "20");
```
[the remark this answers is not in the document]
mkhludnev: It boils down to org.apache.solr.BaseDistributedSearchTestCase#query(boolean, org.apache.solr.common.params.SolrParams) which compares distributed and standalone responses. // ..so, if both of them returns nothing, it's perfectly fine ;)

mkhludnev:
Thanks a lot @alessandrobenedetti

alessandrobenedetti:
You are welcome! I saw you changed the anonymous 'id' to 'docNum'.
It's probably fine, but it didn't look intuitive to me as a name so I checked in the Solr code base, and it seems 'docId' to be more present and widespread used?
You know it's a super minor thing, I have a soft preference for using 'docId' but up to you!
And thanks for the nice work and refactoring, the MoreLikeThis needed some love, I really appreciate it!

mkhludnev:
I think more often about Lucene `docNums`. Ok turned to id. Do you think it's worth to bother with open some MLT methods in Lucene to allow to grab interesting terms without ripping boolean query? Or it's worth to think about generic query serialization framework?

alessandrobenedetti:
Back in the days, in 2017 when I was not a committer yet, I did a massive refactor of the Lucene More Like This code (https://www.slideshare.net/SeaseLtd/how-the-lucene-more-like-this-works, https://issues.apache.org/jira/browse/LUCENE-8326) and also presented it at various conf.
But I didn't get much traction and the work basically got wasted (a lot of tests and code cleaned up)
Long time has passed so it's very likely I would do it differently nowadays but I don't have the time (or fundings?) to do it again right now.
So in general I am in total favor of refactoring the Lucene side of things, happy to review it, I can't guarantee when, but I'll do my best!

mkhludnev:
I'll push it tomorrow. Please chime it.

PR #1267, opened 2023-01-03
SOLR-16608: Ability to compress the collection state

justinrsweeney:
https://issues.apache.org/jira/browse/SOLR-16608

# Description

This PR provides the ability to configured a minimum size of state.json above which it will be compressed when written to Zookeeper. Solr will be able to handle compressing/decompressing in ZLib format in all areas where Solr reads state.json from Zookeeper.

# Solution

This uses ZLib compression to optionally compress state.json, the core of this logic is contained within a new class called CompressionUtil that handles both compression and decompression as well as some logic to check if bytes are compressed in a very efficient way. This can handle reading/writing both compressed and uncompressed data so it changing the configuration for compression is backward compatible and will not break reading state.json.

# Tests

There are tests added to check various layers to ensure compression/decompression is working, including unit tests for CompressionUtil and tests for ZkStateReader and ZkStateWriter.

# Checklist


noblepaul:
LGTM

noblepaul (approved):

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/SolrZkClient.java
```diff
@@ -433,6 +435,14 @@ public byte[] getData(
     } else {
       result = keeper.getData(path, wrapWatcher(watcher), stat);
     }
+    if (CompressionUtil.isCompressedBytes(result)) {
+      log.debug("Zookeeper data at path {} is compressed", path);
+      try {
+        result = CompressionUtil.decompressBytes(result);
+      } catch (DataFormatException e) {
+        throw new RuntimeException(e);
```
HoustonPutman, line 443: Please don't introduce random RuntimeExceptions that have no additional context. You can probably use a SolrException and explain the context where the decompression was happening.

solr/solrj/src/test/org/apache/solr/common/cloud/SolrZkClientCompressedDataTest.java
```diff
@@ -0,0 +1,86 @@
...
+      byte[] compressedData = CompressionUtil.compressBytes(arr);
+      ZkACLProvider aclProvider = new DefaultZkACLProvider();
+      String path = ZkStateReader.COLLECTIONS_ZKNODE + "/c1/state.json";
+      zkClient
+          .getZooKeeper()
+          .create(path, compressedData, aclProvider.getACLsToAdd(path), CreateMode.PERSISTENT);
+
+      byte[] data =
+          zkClient.getData(ZkStateReader.COLLECTIONS_ZKNODE + "/c1/state.json", null, null, true);
+      Map<?, ?> map = (Map<?, ?>) Utils.fromJSON(data);
```
HoustonPutman, line 79: Don't we want to also verify that the bytes are the same before compression and after decompression?

solr/core/src/test/org/apache/solr/cloud/overseer/ZkStateWriterTest.java
```diff
@@ -407,4 +409,89 @@ public void testExternalModification() throws Exception {
...
+            new ZkWriteCommand(
+                "c2", new DocCollection("c2", slices, new HashMap<>(), DocRouter.DEFAULT, 0));
+
+        writer.enqueueUpdate(reader.getClusterState(), Collections.singletonList(c1), null);
+        writer.writePendingUpdates();
+
+        byte[] data =
+            zkClient
+                .getZooKeeper()
+                .getData(ZkStateReader.COLLECTIONS_ZKNODE + "/c2/state.json", null, null);
```
HoustonPutman, line 486: Do we want to ensure that the size is within the acceptable range here?

HoustonPutman (changes requested):
This would be a pretty scary feature to use for me (since there are times where modifying cluster state manually can be unfortunately necessary).

But overall I have a few comments.

Is there a full e2e test here? I see all tests generally use a ZkStateWriter. But it would be good to see a test just start a cloud with compression enabled, create a collection and make sure it is compressed by fetching the data from ZK.

Also for upgrade notes, this feature should not be enabled until all Solr Nodes are running a version of Solr that supports it.

justinrsweeney:
> This would be a pretty scary feature to use for me (since there are times where modifying cluster state manually can be unfortunately necessary).
> 
> But overall I have a few comments.
> 
> Is there a full e2e test here? I see all tests generally use a ZkStateWriter. But it would be good to see a test just start a cloud with compression enabled, create a collection and make sure it is compressed by fetching the data from ZK.
> 
> Also for upgrade notes, this feature should not be enabled until all Solr Nodes are running a version of Solr that supports it.

Thanks for taking a look @HoustonPutman, one of the reasons we chose ZLib compression is because tools like [zk-shell](https://github.com/rgs1/zk_shell) can handle it natively when doing a get on a compressed state.json. Also since this is done in ZKStateReader it is handled across Solr including Admin pages and Solr CLI, which hopefully helps alleviate some of the concern. There is certainly some tradeoff here in terms of manual modification of state, but we've found value in this for very large collections to keep files in ZK small.

+1 on upgrade notes and I'll follow up on additional tests and the other comments you provided.

noblepaul:
> This would be a pretty scary feature to use for me..

It's an opt-in feature. So, a user would not get an unpleasant surprise

solr/core/src/java/org/apache/solr/cloud/Overseer.java
```diff
@@ -199,9 +199,15 @@ private class ClusterStateUpdater implements Runnable, Closeable {
 
     private SolrMetricsContext clusterStateUpdaterMetricContext;
 
+    private final int minimumStateSizeForCompression;
```
dsmiley, line 202: Consider "Bytes" instead of "Size" as it captures the units, and implies a size.

solr/solrj/src/java/org/apache/solr/common/util/CompressionUtil.java
```diff
@@ -0,0 +1,118 @@
...
+package org.apache.solr.common.util;
+
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+import java.util.zip.DataFormatException;
+import java.util.zip.Deflater;
+import java.util.zip.Inflater;
+
+/**
+ * Utility class for dealing with LZ4 compressed data, such as state.json in Zookeeper if configured
```
dsmiley, line 27: I actually wish it was LZ4 but you're using ZLIB.  I think the methods on this class should indicate the compression used.  I imagine we might want to actually use LZ4 and then we'd have a naming issue.

solr/solrj/src/java/org/apache/solr/common/util/CompressionUtil.java
```diff
@@ -0,0 +1,118 @@
...
+        buf = Arrays.copyOf(buf, actualDecompressedSize);
+      }
+      return buf;
+    } finally {
+      inflater.end();
+    }
+  }
+
+  /**
+   * Compresses bytes into zlib compressed bytes using java native compression
```
dsmiley, line 86: There is no such thing as "java native compression".

```suggestion
   * Compresses bytes into zlib compressed bytes using ZLIB (AKA Deflate).
```

FYI https://stackoverflow.com/a/20765054

dsmiley (commented):
I'm super-pleased to see this contribution to Solr!  Funny enough, it was on my plans to do similarly.

HoustonPutman:
> @HoustonPutman what if you could edit zk through the solr admin? Then presumably the api that pushed that data up and down could deal with the compression? I've done the same ZK editing in the past ;-)

Yeah then that would be fine 🙂

solr/solrj/src/java/org/apache/solr/common/util/CompressionUtil.java
justinrsweeney: We actually originally started down the path of LZ4, but a number of different tools/languages we use have built in support for ZLIB compression vs having to use a third party library or build additional support for LZ4. I can do some refactoring here to make it open to additional compression methods. Out of curiosity, what would push you towards LZ4 instead?

dsmiley: I recall the compress/decompress speed was better even if it didn't shrink the size as much as ZLIB, which seemed to be a better trade-off for this use-case.  But I totally understand wanting to prefer convenience of ease of access to ZLIB in tooling.

solr/core/src/java/org/apache/solr/cloud/Overseer.java
noblepaul: `size` is better compared to bytes

`minimumBytesForCompression` vs `minimumStateSizeForCompression`

however, we could use `minStateSizeForCompression` instead of `minimumStateSizeForCompression`

noblepaul:
> > @HoustonPutman what if you could edit zk through the solr admin? Then presumably the api that pushed that data up and down could deal with the compression? I've done the same ZK editing in the past ;-)
> 
> Yeah then that would be fine 🙂

compression will be used when the `state.json` is too big. Loading such a big JSON file in admin UI is not a good idea. most likely the workflow is as follows

1. download the json using ZkCLI
2. edit the file using some tool
3. upload it back using ZkCLI

solr/core/src/java/org/apache/solr/cloud/Overseer.java
dsmiley: You didn't supply a reason why "size" is better than "bytes".  I gave a reason why "bytes" is better.

Other idea is "ByteLen".

Agreed on "min" vs "minimum".

Note that if we ultimately land on a unit ambiguous name (as is right now), the very least we could do is add a line of docs to specify the unit.

dsmiley (commented):
Other than size units ambiguity, +1 for me!

justinrsweeney:
> Other than size units ambiguity, +1 for me!

I made the update here to use minStateByteLenForCompression in all places.

justinrsweeney:
> > > @HoustonPutman what if you could edit zk through the solr admin? Then presumably the api that pushed that data up and down could deal with the compression? I've done the same ZK editing in the past ;-)
> > 
> > 
> > Yeah then that would be fine 🙂
> 
> compression will be used when the `state.json` is too big. Loading such a big JSON file in admin UI is not a good idea. most likely the workflow is as follows
> 
> 1. download the json using ZkCLI
> 2. edit the file using some tool
> 3. upload it back using ZkCLI

Working on getting this working as the last piece here.

solr/core/src/java/org/apache/solr/cloud/Overseer.java
noblepaul: I thnk solr uses `size` more often . Either is fine

noblepaul (approved):

dsmiley (approved):

justinrsweeney:
The change to the ZkCLI has some complexity that requires its own discussion, enough so that I believe it warrants its own PR once this one is merged. Since this change does not introduce compression unless intentionally configured, I am proposing that we merge this change and I will follow up with an additional PR for the ZkCLI changes to handle compressed state through that CLI tool.

@HoustonPutman Does that sound reasonable to you?

HoustonPutman:
I think it would be nice to have all together, but if you want to split it up thats fine with me. Be sure not to close the JIRA, and if the second PR doesnt make it by 9.2 then the limitations need to be spelled out in the ref guide.

PR #1271, opened 2023-01-04
SOLR-16612: fix-hadoop-openj9

risdenk:
https://issues.apache.org/jira/browse/SOLR-16612

I'm looking into the Hadoop integration test failures on openj9 that are related to https://lists.apache.org/thread/b3053cvc2jqw768jrrw8npxkow4k70r6

I know Hadoop had some IBM java checks that look to be not necessary anymore after SOME change in IBM. I made an initial check change here, but not sure its correct.

@uschindler is there a good way to determine old IBM java vs new IBM Java based on openjdk?

### DIFF
--- a/solr/modules/hdfs/src/test/org/apache/hadoop/util/PlatformName.java
+++ b/solr/modules/hdfs/src/test/org/apache/hadoop/util/PlatformName.java
@@ -18,6 +18,7 @@ package org.apache.hadoop.util;
 
 import org.apache.hadoop.classification.InterfaceAudience;
 import org.apache.hadoop.classification.InterfaceStability;
+import org.apache.solr.client.solrj.util.Constants;
 
 /**
  * A helper class for getting build-info of the java-vm.
@@ -45,9 +46,7 @@ public class PlatformName {
    * A public static variable to indicate the current java vendor is
    * IBM java or not.
    */
-  public static final boolean IBM_JAVA = JAVA_VENDOR_NAME.contains("IBM")
-      // added the following check for Solr since newer OpenJ9 builds don't have IBM specific classes.
-      && System.getProperty("java.specification.vendor").contains("IBM");
+  public static final boolean IBM_JAVA = Constants.IS_IBM_JAVA;
 
   public static void main(String[] args) {
     System.out.println(PLATFORM_NAME);
--- a/solr/modules/hdfs/src/test/org/apache/hadoop/util/PlatformName.java
+++ b/solr/modules/hdfs/src/test/org/apache/hadoop/util/PlatformName.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.util;
+
+import org.apache.hadoop.classification.InterfaceAudience;
+import org.apache.hadoop.classification.InterfaceStability;
+
+/**
+ * A helper class for getting build-info of the java-vm.
+ *
+ */
+@InterfaceAudience.LimitedPrivate({"HBase"})
+@InterfaceStability.Unstable
+public class PlatformName {
+  /**
+   * The complete platform 'name' to identify the platform as
+   * per the java-vm.
+   */
+  public static final String PLATFORM_NAME =
+      (System.getProperty("os.name").startsWith("Windows")
+          ? System.getenv("os") : System.getProperty("os.name"))
+          + "-" + System.getProperty("os.arch")
+          + "-" + System.getProperty("sun.arch.data.model");
+
+  /**
+   * The java vendor name used in this platform.
+   */
+  public static final String JAVA_VENDOR_NAME = System.getProperty("java.vendor");
+
+  /**
+   * A public static variable to indicate the current java vendor is
+   * IBM java or not.
+   */
+  public static final boolean IBM_JAVA = JAVA_VENDOR_NAME.contains("IBM")
+      // added the following check for Solr since newer OpenJ9 builds don't have IBM specific classes.
+      && System.getProperty("java.specification.vendor").contains("IBM");
+
+  public static void main(String[] args) {
+    System.out.println(PLATFORM_NAME);
+  }
+}
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/util/Constants.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/util/Constants.java
@@ -17,8 +17,14 @@
 
 package org.apache.solr.client.solrj.util;
 
+import java.lang.invoke.MethodHandles;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 // Clone of org.apache.lucene.util.Constants, so SolrJ can use it
 public class Constants {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
   public static final String JVM_SPEC_VERSION = System.getProperty("java.specification.version");
 
   public static final boolean JRE_IS_MINIMUM_JAVA9 = true;
@@ -30,9 +36,14 @@ public class Constants {
 
   private static boolean isIBMJava() {
     try {
-      Class.forName("com.ibm.security.auth.module.Krb5LoginModule", false, null);
+      Class.forName(
+          "com.ibm.security.auth.module.Krb5LoginModule", false, Constants.class.getClassLoader());
       return true;
-    } catch (ClassNotFoundException e) {
+    } catch (SecurityException se) {
+      log.warn(
+          "Unable to determine if IBM Java due to SecurityException. Assuming not IBM Java.", se);
+      return false;
+    } catch (ClassNotFoundException ignore) {
       return false;
     }
   }
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/util/Constants.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/util/Constants.java
@@ -25,4 +25,15 @@ public class Constants {
   public static final boolean JRE_IS_MINIMUM_JAVA11 = true;
   // Future, enable if needed...
   // public static final boolean JRE_IS_MINIMUM_JAVA17 = Runtime.version().feature() >= 17;
+
+  public static final boolean IS_IBM_JAVA = isIBMJava();
+
+  private static boolean isIBMJava() {
+    try {
+      Class.forName("com.ibm.security.auth.module.Krb5LoginModule", false, null);
+      return true;
+    } catch (ClassNotFoundException e) {
+      return false;
+    }
+  }
 }
--- a/solr/modules/hadoop-auth/src/test/org/apache/solr/security/hadoop/KerberosTestServices.java
+++ b/solr/modules/hadoop-auth/src/test/org/apache/solr/security/hadoop/KerberosTestServices.java
@@ -190,6 +190,9 @@ public class KerberosTestServices {
 
   public static final String krb5LoginModuleName =
       System.getProperty("java.vendor").contains("IBM")
+              // added the following check for Solr since newer OpenJ9 builds don't have IBM
+              // specific classes.
+              && System.getProperty("java.specification.vendor").contains("IBM")
           ? "com.ibm.security.auth.module.Krb5LoginModule"
           : "com.sun.security.auth.module.Krb5LoginModule";
 
--- a/solr/modules/hadoop-auth/src/test/org/apache/solr/security/hadoop/KerberosTestServices.java
+++ b/solr/modules/hadoop-auth/src/test/org/apache/solr/security/hadoop/KerberosTestServices.java
@@ -30,6 +30,7 @@ import javax.security.auth.login.Configuration;
 import org.apache.commons.io.file.PathUtils;
 import org.apache.hadoop.minikdc.MiniKdc;
 import org.apache.solr.client.solrj.impl.Krb5HttpClientBuilder;
+import org.apache.solr.client.solrj.util.Constants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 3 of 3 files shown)


solr/modules/hdfs/src/test/org/apache/hadoop/util/PlatformName.java
```diff
@@ -0,0 +1,55 @@
...
+   */
+  public static final String JAVA_VENDOR_NAME = System.getProperty("java.vendor");
+
+  /**
+   * A public static variable to indicate the current java vendor is
+   * IBM java or not.
+   */
+  public static final boolean IBM_JAVA = JAVA_VENDOR_NAME.contains("IBM")
+      // added the following check for Solr since newer OpenJ9 builds don't have IBM specific classes.
+      && System.getProperty("java.specification.vendor").contains("IBM");
```
risdenk, line 50: This is a Hadoop class that is copied to override this specific IBM_JAVA check and added this additional check.

uschindler:
Maybe just check if one or the other class exists in runtime?

uschindler (approved):
I think you can hotfix it like this, but actually the bug is somewhere in Hadoop.

Actually what happened: in earlier versions, the java specification was a IBM one (using Apache Harmony). Nowadays, Semeru uses OpenJDK and they just added a few own APIs and replaced  Hotspot by OpenJ9. Because of this the check for the specification version is the only correct one. Depending on java.vendor makes no sense as this does not say anything about the API specification.

uschindler:
But on the long term, both Hadoop and also Solr should just check, which of both classes are available. Relying on system properties should be avoided if possible as this is just text. What happens if they remove IBM and write OpenJ9 (I think they plan to do this at some point).

uschindler:
Do you have an issue already opened at Hadoop?

uschindler:
In general for Java 17+ there is no longer a IBM "variant" available. For Java 11 I am not 100% sure, but I think they already switched to OpenJDK code base after Java 8. So maybe the check could go away completely?

uschindler:
As said before, for reads, the native varhandles can be used. There is also no separate "longBitsToRawDouble". What was the reason to add the latest commit?

solr/modules/hdfs/src/test/org/apache/hadoop/util/PlatformName.java
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * The java vendor name used in this platform.
+   */
+  public static final String JAVA_VENDOR_NAME = System.getProperty("java.vendor");
+
+  /**
+   * A public static variable to indicate the current java vendor is
+   * IBM java or not.
+   */
+  public static final boolean IBM_JAVA = Constants.IS_IBM_JAVA;
```
risdenk, line 49: Copies the Hadoop code and then replaces this with the Solr logic for IS_IBM_JAVA to consolidate this to one spot.

solr/solrj/src/java/org/apache/solr/client/solrj/util/Constants.java
```diff
@@ -25,4 +25,15 @@ public class Constants {
...
   // Future, enable if needed...
   // public static final boolean JRE_IS_MINIMUM_JAVA17 = Runtime.version().feature() >= 17;
+
+  public static final boolean IS_IBM_JAVA = isIBMJava();
+
+  private static boolean isIBMJava() {
+    try {
+      Class.forName("com.ibm.security.auth.module.Krb5LoginModule", false, null);
+      return true;
+    } catch (ClassNotFoundException e) {
```
uschindler, line 35: this works, but you really should also catch SecurityException, as this could happen with SecurityManager as enabled in SOlr by default.

Also to prevent a security Exception the last parameter should be `Constants.class.getClassLoader()` (this is what the default forName is doing). Sending null will use system classloader and that one is restricted by security manager.

uschindler (changes requested):

solr/solrj/src/java/org/apache/solr/client/solrj/util/Constants.java
uschindler: See also Javadoc: 

> If the parameter loader is null, the class is loaded through the bootstrap class loader.
> 
> [SecurityException](https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/lang/SecurityException.html) - if a security manager is present, and the loader is null, and the caller's class loader is not null, and the caller does not have the [RuntimePermission](https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/lang/RuntimePermission.html)("getClassLoader")

So to not get into security problems use the suggestion here:

> For example, in an instance method the expression:
> Class.forName("Foo")
> is equivalent to:
> Class.forName("Foo", true, this.getClass().getClassLoader())

so just replace by "false".

risdenk: yea `this.getClass().getClassLoader()` won't work with the static variable. `Constants.class.getClassLoader()` does.

risdenk: Fixed in e258862

uschindler: Thanks looks fine to me. You code including logging looks like I produced yesterday in Lucene two times - especially in static initializers you must be very careful with SecurityExceptions, they are unchecked and can appear at many places. The code then make class completely unuseable, which is horrible for some central, often used class like Constants - or the underlying directory implementation of Lucene: https://github.com/apache/lucene/blob/7f483bd6180c5c88176cc7c212405680d67ae249/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java#L376-L390

uschindler (approved):
Looks fine, let's give it a try. I have no Semeru JDK at hand to test it, but we will see what Jenkins says.

The Hadoop upstream changes are similar to ours (they also test for class).

risdenk:
I ran this locally for a bunch of tests with openj9 on Mac - The Hadoop tests at least pass now. I haven't gotten to these yet but they are a separate root cause from what I can tell:

```
RUNTIME_JAVA_HOME=/Library/Java/JavaVirtualMachines/ibm-semeru-open-17.jdk/Contents/Home ./gradlew check
...
  - org.apache.solr.security.TestPKIAuthenticationPlugin.testParseCipher (:solr:core)
    Test output: /Users/risdenk/repos/apache/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.security.TestPKIAuthenticationPlugin.txt
    Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.security.TestPKIAuthenticationPlugin.testParseCipher" -Ptests.jvms=5 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=B660BEBE6AAC0ADE -Ptests.file.encoding=US-ASCII

  - org.apache.solr.security.TestPKIAuthenticationPlugin.testProtocolMismatch (:solr:core)
    Test output: /Users/risdenk/repos/apache/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.security.TestPKIAuthenticationPlugin.txt
    Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.security.TestPKIAuthenticationPlugin.testProtocolMismatch" -Ptests.jvms=5 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=B660BEBE6AAC0ADE -Ptests.file.encoding=US-ASCII

  - org.apache.solr.security.TestPKIAuthenticationPlugin.testParseCipherNoTimestamp (:solr:core)
    Test output: /Users/risdenk/repos/apache/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.security.TestPKIAuthenticationPlugin.txt
    Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.security.TestPKIAuthenticationPlugin.testParseCipherNoTimestamp" -Ptests.jvms=5 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=B660BEBE6AAC0ADE -Ptests.file.encoding=US-ASCII

  - org.apache.solr.security.TestPKIAuthenticationPlugin.testParseCipherInvalidTimestampTooBig (:solr:core)
    Test output: /Users/risdenk/repos/apache/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.security.TestPKIAuthenticationPlugin.txt
    Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.security.TestPKIAuthenticationPlugin.testParseCipherInvalidTimestampTooBig" -Ptests.jvms=5 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=B660BEBE6AAC0ADE -Ptests.file.encoding=US-ASCII

  - org.apache.solr.security.TestPKIAuthenticationPlugin.testSuperUser (:solr:core)
    Test output: /Users/risdenk/repos/apache/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.security.TestPKIAuthenticationPlugin.txt
    Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.security.TestPKIAuthenticationPlugin.testSuperUser" -Ptests.jvms=5 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=B660BEBE6AAC0ADE -Ptests.file.encoding=US-ASCII

  - org.apache.solr.security.TestPKIAuthenticationPlugin.testParseCipherNoSpace (:solr:core)
    Test output: /Users/risdenk/repos/apache/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.security.TestPKIAuthenticationPlugin.txt
    Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.security.TestPKIAuthenticationPlugin.testParseCipherNoSpace" -Ptests.jvms=5 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=B660BEBE6AAC0ADE -Ptests.file.encoding=US-ASCII

  - org.apache.solr.security.TestPKIAuthenticationPlugin.testParseCipherInvalidTimestampTooSmall (:solr:core)
    Test output: /Users/risdenk/repos/apache/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.security.TestPKIAuthenticationPlugin.txt
    Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.security.TestPKIAuthenticationPlugin.testParseCipherInvalidTimestampTooSmall" -Ptests.jvms=5 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=B660BEBE6AAC0ADE -Ptests.file.encoding=US-ASCII

  - org.apache.solr.security.TestPKIAuthenticationPlugin.testParseCipherInvalidKey (:solr:core)
    Test output: /Users/risdenk/repos/apache/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.security.TestPKIAuthenticationPlugin.txt
    Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.security.TestPKIAuthenticationPlugin.testParseCipherInvalidKey" -Ptests.jvms=5 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=B660BEBE6AAC0ADE -Ptests.file.encoding=US-ASCII

  - org.apache.solr.security.TestPKIAuthenticationPlugin.testBasicRequest (:solr:core)
    Test output: /Users/risdenk/repos/apache/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.security.TestPKIAuthenticationPlugin.txt
    Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.security.TestPKIAuthenticationPlugin.testBasicRequest" -Ptests.jvms=5 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=B660BEBE6AAC0ADE -Ptests.file.encoding=US-ASCII

  - org.apache.solr.cloud.TestRSAKeyPair.testGenKeyPair (:solr:core)
    Test output: /Users/risdenk/repos/apache/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.cloud.TestRSAKeyPair.txt
    Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.cloud.TestRSAKeyPair.testGenKeyPair" -Ptests.jvms=5 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=B660BEBE6AAC0ADE -Ptests.file.encoding=US-ASCII

  - org.apache.solr.cloud.TestRSAKeyPair.testReadKeysFromDisk (:solr:core)
    Test output: /Users/risdenk/repos/apache/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.cloud.TestRSAKeyPair.txt
    Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.cloud.TestRSAKeyPair.testReadKeysFromDisk" -Ptests.jvms=5 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=B660BEBE6AAC0ADE -Ptests.file.encoding=US-ASCII
...
```

uschindler:
Those other tests complain about the private key not working, right? I thought this was also related to hadoop's kerberos library?

uschindler:
> Those other tests complain about the private key not working, right? I thought this was also related to hadoop's kerberos library?

Sorry, thats home-made:

```
org.apache.solr.common.SolrException: javax.crypto.BadPaddingException: RSA private key operation failed
	at __randomizedtesting.SeedInfo.seed([8471E4118F153A0C:11C67A010B918243]:0)
	at app//org.apache.solr.util.CryptoKeys$RSAKeyPair.encrypt(CryptoKeys.java:308)
	at app//org.apache.solr.security.PKIAuthenticationPlugin.generateToken(PKIAuthenticationPlugin.java:467)
	at app//org.apache.solr.security.PKIAuthenticationPlugin.setHeader(PKIAuthenticationPlugin.java:489)
	at app//org.apache.solr.security.TestPKIAuthenticationPlugin.testSuperUser(TestPKIAuthenticationPlugin.java:167)
```

PR #1273, opened 2023-01-04
SOLR-16610: Support Copy n Paste of Command Line commands in Ref Guide

epugh:
https://issues.apache.org/jira/browse/SOLR-16610

# Description

This is an example of the types of changes we would need to make to be more in line with Antora's handling of command lines and being able to cut n paste them.

# Solution

Conversion

# Tests

manual


epugh:
Ironically, from starting a journey to remove the `$` from all commands, now we would go the opposite direction and introduce them everywhere!  Right now it's kind of a mish mash.

risdenk:
https://docs.antora.org/antora-ui-default/copy-to-clipboard/#console-blocks

epugh:
I read the docs, but didn't find this section...   ;-)   Stumble across an example and then figured it out by just trialing...  Great to see the explanation of how it works!

risdenk:
I think we should use `console` then to make the code blocks explicit. Right now the indents make it hard to tell when editing the adoc if its actually a code block or not.

epugh:
I get that...   Does it deal properly with the multiple line thing?
```
$ bin/java
SOMEOUTPUT
```
and the copy only grabs `bin/java`?

epugh:
@risdenk so the `console` seemed to work out really nicely....     Thoughts?   Love a +1 before I start plowing through the Ref Guide ;-)

epugh:
The multiline `curl` command even worked!!!

risdenk (approved):

epugh:
umm..   I was a bit bummed I missed the cut off...   not sure I'm quite up for trying out merging to staging, so was just oign t to suffer in silence till 9.3 came out ;-)   I've never touched the staging site.

HoustonPutman:
Luckily since its just a doc change, whatever is on branch_9_2 is the official 9.2 docs, so backporting after the release is no issue at all.

This is the nightly build that you can try tomorrow after the jenkins job runs: https://nightlies.apache.org/solr/draft-guides/solr-reference-guide-nightly/solr/9_3/

HoustonPutman:
The nightly build works for me! Feel free to backport to branch_9_2!

And it would be great to get this spread across the rest of the guide as well! (not volunteering you though 😉)

PR #1168, opened 2022-11-09
SOLR-16532 New OTEL module with OTLP trace exporter

janhoy:
https://issues.apache.org/jira/browse/SOLR-16532

This is a new solr module which adds an OpenTelemetry tracer to Solr.
It is configured through env.variables. See ref guide for docs.

TODO:

- [x] Add JUnit tests for the config logic
- [x] Add license and notice for new libs
- [x] Add integration test for the tracing (may be hard)

PR #1275, opened 2023-01-05
SOLR-16532 Further improvements to opentelemetry module

janhoy:
https://issues.apache.org/jira/browse/SOLR-16532

I created a new PR where we can gather all followup cleanups that may arrive after the initial merge in #1168


janhoy:
@dsmiley Regarding the hope of excluding some dependencies, we received a reply in https://github.com/open-telemetry/opentelemetry-java/discussions/5079#discussioncomment-4595701, so it seems hard to get rid of the -log and -metric deps. 

This jack claims that we should be able to exclude 'okhttp' and 'kotlin', but when I tried IIRC it was a hard dependency of gRPC lib. If you want to give it a shot that would be great. It would make those test-dependencies only starting from 10.0.

janhoy:
Anything more we should fix in this same PR? I kind of gave up on slimming down the jars for now, but feel free to give it a shot if you like.

PR #1278, opened 2023-01-05
SOLR-15787: Fix Windows

risdenk:
https://issues.apache.org/jira/browse/SOLR-15787


dsmiley:
Nonetheless I liked some of the minor code improvements in this PR so I think we should bring them in.  Feel free to push them to Nazerke's PR.

PR #1279, opened 2023-01-06
SOLR-16613: CryptoKeys should handle RSA padding for OpenJ9

risdenk:
https://issues.apache.org/jira/browse/SOLR-16613

Tested with:

```
# With Temurin 17
./gradlew test --tests TestRSAKeyPair
./gradlew test --tests TestPKIAuthenticationPlugin

# With OpenJ9 17
RUNTIME_JAVA_HOME=/Library/Java/JavaVirtualMachines/ibm-semeru-open-17.jdk/Contents/Home ./gradlew test --tests TestRSAKeyPair
RUNTIME_JAVA_HOME=/Library/Java/JavaVirtualMachines/ibm-semeru-open-17.jdk/Contents/Home ./gradlew test --tests TestPKIAuthenticationPlugin
```

### DIFF
--- a/solr/core/src/java/org/apache/solr/util/CryptoKeys.java
+++ b/solr/core/src/java/org/apache/solr/util/CryptoKeys.java
@@ -39,7 +39,6 @@ import java.security.interfaces.RSAPrivateKey;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.X509EncodedKeySpec;
-import java.util.Arrays;
 import java.util.Base64;
 import java.util.Collection;
 import java.util.HashMap;
@@ -313,11 +312,12 @@ public final class CryptoKeys {
       return keySizeInBytes;
     }
 
-    public byte[] encrypt(byte[] buffer) {
+    public byte[] encrypt(ByteBuffer buffer) {
       // This is necessary to pad the plaintext to match the exact size of the keysize in openj9.
       // OpenJDK seems to do this padding internally, but OpenJ9 does not pad the byte input to
       // the key size in bytes without padding. This only works with "RSA/ECB/nopadding".
-      byte[] paddedPlaintext = Arrays.copyOf(buffer, getKeySizeInBytes());
+      byte[] paddedPlaintext = new byte[getKeySizeInBytes()];
+      buffer.get(paddedPlaintext, buffer.arrayOffset() + buffer.position(), buffer.limit());
 
       try {
         // This is better than nothing, but still not very secure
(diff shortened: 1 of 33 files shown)


solr/core/src/test/org/apache/solr/cloud/TestRSAKeyPair.java
```diff
@@ -36,26 +38,29 @@ public void testGenKeyPair() throws Exception {
...
     URL publicKey = getClass().getClassLoader().getResource("cryptokeys/pub_key512.der");
-
+    assertNotNull(privateKey);
+    assertNotNull(publicKey);
     testRoundTrip(new CryptoKeys.RSAKeyPair(privateKey, publicKey));
   }
 
   private void testRoundTrip(CryptoKeys.RSAKeyPair kp) throws Exception {
-    final byte[] plaintext = new byte[random().nextInt(64)];
-    random().nextBytes(plaintext);
```
risdenk, line 45: random().nextBytes() would end up with null bytes and that would cause issues. I switched this to generate a regular string with no null bytes.

solr/core/src/java/org/apache/solr/util/CryptoKeys.java
```diff
@@ -295,15 +304,28 @@ public PublicKey getPublicKey() {
       return publicKey;
     }
 
-    public byte[] encrypt(ByteBuffer buffer) {
```
risdenk, line 298: We only call `ByteBuffer.wrap` in all invocations of this so isn't helpful to have this as a `ByteBuffer`. The decryption also just uses a plain `byte[]` so this makes it consistent.

madrob (commented):
Should we switch the algorithm from no padding to something with padding?

risdenk:
> Should we switch the algorithm from no padding to something with padding?

probably at some point. I don't really want to tackle that with this. This at least cleans up the build failures.

solr/core/src/java/org/apache/solr/util/CryptoKeys.java
magibney: This is a public method on a public utility class though; I wonder if the intention was to have this used flexibly from elsewhere? This could in principle break plugin code ...

This change seems good still, but worth explicitly calling out this possible concern.

risdenk: Yea I'm just going to back out this change.

risdenk: Reverted this in 741e131ebbc43b7e49fc668e0c76cb6163946b3b

PR #1282, opened 2023-01-10
SOLR-16457: Use pwd -P

risdenk:
https://issues.apache.org/jira/browse/SOLR-16457

### DIFF
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -87,7 +87,7 @@ done
 CDPATH=''  # Prevent "file or directory not found" for 'cdpath' users
 SOLR_TIP=$(dirname "$SOLR_SCRIPT")/..
 # shellcheck disable=SC2164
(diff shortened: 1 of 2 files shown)


HoustonPutman (approved):
Nice! Glad we can get rid of that envVar

elyograg:
Looks good to me.  I'm not familiar with the docker stuff that's gets one line removed or what that will do.

risdenk:
> Looks good to me. I'm not familiar with the docker stuff that's gets one line removed or what that will do.

I added the docker env var line when I implemented SOLR-16457 which works around the symlink stuff specifically only for Docker. Since `pwd -P` fixes this issue in bin/solr no need to fix it in Docker anymore.

elyograg (approved):
@risdenk helped me find the approve button.

PR #1284, opened 2023-01-10
SOLR-16616: JWTAuthPlugin: Read trusted X509 certificates from multi files

janhoy:
https://issues.apache.org/jira/browse/SOLR-16616

A concrete use case for this is in a Kubernetes environment where a configMap exists with two `.cer` files, one for the SSL certificate of KeyCloak, and another file with SSL cert for the private CA. Currently we have been copying and concatenating the string content of those files into the `trustedCerts` configuration key. But that is cumbersome when you have several k8s clusters with different certificates for each.

With this feature, we'll supply a both files as this
```json
"trustedCertsFile": ["/path/to/hostCertFile.pem", "/path/to/caCertFile.pem"]
```

I considered adding a new parameter `trustedCertsFiles`, but went for accepting list on the existing one.

### DIFF
--- a/solr/modules/jwt-auth/src/java/org/apache/solr/security/jwt/JWTAuthPlugin.java
+++ b/solr/modules/jwt-auth/src/java/org/apache/solr/security/jwt/JWTAuthPlugin.java
@@ -17,7 +17,6 @@
 package org.apache.solr.security.jwt;
 
 import java.io.IOException;
-import java.io.InputStream;
 import java.lang.invoke.MethodHandles;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
@@ -78,7 +77,7 @@ import org.jose4j.lang.JoseException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/** Authenticaion plugin that finds logged in user by validating the signature of a JWT token */
+/** Authentication plugin that finds logged in user by validating the signature of a JWT token */
 public class JWTAuthPlugin extends AuthenticationPlugin
     implements SpecProvider, ConfigEditablePlugin {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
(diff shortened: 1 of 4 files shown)


solr/modules/jwt-auth/src/java/org/apache/solr/security/jwt/JWTAuthPlugin.java
```diff
@@ -297,6 +283,54 @@ public void init(Map<String, Object> pluginConfig) {
...
+    Collection<X509Certificate> certs = new HashSet<>();
+    if (trustedCertsFileObj instanceof String) {
+      String trustedCertsFile = (String) trustedCertsFileObj;
+      log.info("Reading trustedCerts from file {}", trustedCertsFile);
+      try {
+        certs.addAll(parseCertsFromFile(trustedCertsFile));
+      } catch (IOException e) {
+        throw new SolrException(
+            SolrException.ErrorCode.SERVER_ERROR, "Failed to read file " + trustedCertsFile, e);
+      }
```
risdenk, line 301: Couldn't you just make this string into a List of one and use the same logic as the list? That would simplify and remove the duplications.

solr/modules/jwt-auth/src/java/org/apache/solr/security/jwt/JWTAuthPlugin.java
```diff
@@ -297,6 +283,54 @@ public void init(Map<String, Object> pluginConfig) {
     lastInitTime = Instant.now();
   }
 
+  /**
+   * Given a configuration object of a file name or list of file names, read X509 certificates from
+   * each file
+   */
+  @SuppressWarnings("unchecked")
```
risdenk, line 290: What is unchecked here?

risdenk (changes requested):

solr/modules/jwt-auth/src/java/org/apache/solr/security/jwt/JWTAuthPlugin.java
janhoy: Unchecked cast from `Object` to `List<String>`

solr/modules/jwt-auth/src/java/org/apache/solr/security/jwt/JWTAuthPlugin.java
janhoy: I had the same thought on the train home from work today :) Will investigate.

solr/modules/jwt-auth/src/java/org/apache/solr/security/jwt/JWTAuthPlugin.java
```diff
@@ -297,6 +283,54 @@ public void init(Map<String, Object> pluginConfig) {
...
+    return certs;
+  }
+
+  /**
+   * Given a filename string, validate the file and then read any X509 certificates from it
+   *
+   * @return list of certificates found in file
+   */
+  Collection<? extends X509Certificate> parseCertsFromFile(String certFileName) throws IOException {
+    Path certFilePath = Paths.get(certFileName);
```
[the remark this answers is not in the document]
janhoy: @sonatype-lift ignore
We check in the line below that this path is safe.

risdenk (approved):

janhoy:
Thanks for the review. I'll merge this on Friday if no other comments.

PR #1285, opened 2023-01-10
SOLR-16617: DocSetBuilder should use DocSetUtil#smallSetSize

risdenk:
https://issues.apache.org/jira/browse/SOLR-16617

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/DocSetBuilder.java
+++ b/solr/core/src/java/org/apache/solr/search/DocSetBuilder.java
@@ -42,11 +42,7 @@ public final class DocSetBuilder {
 
   public DocSetBuilder(int maxDoc, long costEst) {
     this.maxDoc = maxDoc;
-    // For ridiculously small sets, we'll just use a sorted int[]
-    // maxDoc >>> 7 is a good value if you want to save memory, lower values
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/search/DocSetBuilder.java
```diff
@@ -42,11 +42,7 @@ public final class DocSetBuilder {
 
   public DocSetBuilder(int maxDoc, long costEst) {
     this.maxDoc = maxDoc;
-    // For ridiculously small sets, we'll just use a sorted int[]
-    // maxDoc >>> 7 is a good value if you want to save memory, lower values
-    // such as maxDoc >>> 11 should provide faster building but at the expense
-    // of using a full bitset even for quite sparse data
-    this.threshold = (maxDoc >>> 7) + 4; // the +4 is for better testing on small indexes
```
risdenk, line 49: There isn't an indication if the above is better than `(maxDoc >> 6) + 5` - they are both the same idea. There are fewer callers of `DocSetBuilder` than the `DocSetUtil#smallSetSize` so stuck w/ the `smallSetSize` formula but could be convinced otherwise.

dsmiley (approved):
LGTM

PR #1287, opened 2023-01-12
SOLR-16619: Fix `solr` script for IBM i

ThePrez:
https://issues.apache.org/jira/browse/SOLR-16619

# Description

The `solr` script runs an invocation of `ps` that is incompatible with the `ps` that ships on IBM i

# Solution

When running on IBM i, a simplified invocation is needed, namely `ps -f -p "$SOLR_PID"`

# Tests

Manual tests on IBM i have verified that the change causes Solr to behave correctly when using OpenJDK.

"System Java," also known as "JV1," will still fail, but there are likely several contributing factors beyond the startup script issue identified here. 

# Checklist

- ~~[ ] I have run `./gradlew check`. (N/A)~~
- ~~[ ] I have added tests for my changes. (N/A)~~
- ~~[ ] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide) (N/A)~~


solr/bin/solr
```diff
@@ -716,7 +716,11 @@ function solr_pid_by_port() {
 # extract the value of the -Djetty.port parameter from a running Solr process
 function jetty_port() {
   SOLR_PID="$1"
-  SOLR_PROC=$(ps -o command='' -p "$SOLR_PID" | grep start\.jar | grep jetty\.port)
+  if [ "OS400" == "$THIS_OS" ]; then
+    SOLR_PROC=`ps -f -p "$SOLR_PID" | grep start\.jar | grep jetty\.port`
```
janhoy, line 720: Please stick to the `$(foo)` syntax

ThePrez: Sorry for the delay. Fixed at latest commit/force-push. Thanks!

epugh:
What do we think?   Comittable?   (Someday we may need to think about how to test across many types of JVM's ;-( )

ThePrez:
@colvinco was doing some more in-depth investigation, as documented in Jira. 
It may be a good idea to revisit the overall usage of `ps` as was started, but I think it'd be safe to merge this in the meantime (deferring to maintainers for decision though).

colvinco:
Sorry I've not had a lot of time for Solr recently. Seems okay for now, cleaning up all the different uses of `ps` can still happen later

epugh:
@colvinco if you think this "LGTM" then I'm happy to shepherd committing this ;-).   I've been periodically reviewing this JIRA and dreading fixing it ;-).   Shell scripting magic isn't an area I'm super comofrotable with.   Is there any enhancement to our BATS test that would validate/verify this fix?

colvinco:
Thanks @epugh. It looks like test_status.bats should fail if something goes wrong here, but as ever the pain is coming from having different OSes and versions of `ps`.

I can't actually see any mention of someone having checked with a macos, so that's the risky bit. I don't have a mac to test with myself. It would be great if PR checks with changes to the bash scripts ran on macos and linux, but in the meantime it would best if someone could verify `-f` works as expected for mac users.

epugh:
I am on MacOS, so I can test that...  Assuming it passes, then I'll merge.  Thanks!

epugh:
I ran the bats test and the `bin/solr start -e example` with four nodes, and it all seemed to work.

epugh:
@ThePrez  How do you want to be credited in `CHANGES.txt`????

ThePrez:
> @ThePrez  How do you want to be credited in `CHANGES.txt`????

These changes are so small attribution isn't necessary, but if you choose, you can use my real name: Jesse Gorzinski. 
Thanks!

PR #1125, opened 2022-10-25
Fix useGpg signing without password

HoustonPutman:
Also make the release wizard work. It is currently broken with respect to GPG signing.

PR #1291, opened 2023-01-14
buildAndPushRelease should optionally pause before assembleRelease

magibney:
apache/solr#1125 works around gpg pinentry via a `setup_gpg` step that caches the user pin in `gpg-agent`.

For configurations that require explicit user PIN entry via non-tty-based pinentry program, as discussed in [help/publishing.txt](https://github.com/apache/solr/blob/releases/solr/9.1.0/help/publishing.txt#L125-L128) -- i.e., where gpg-agent is not configured to cache user PIN -- the approach followed in apache/solr#1125 will not work.

In such cases, we should provide the option to pause for user confirmation before proceeding to the `assembleRelease`/signing phase. Without such a pause, it's likely that an RM will step away while tests are running, and once tests complete but before the RM returns, the pinentry prompt will timeout, causing the entire invocation of `buildAndPushRelease.py` to fail unnecessarily (ask me how I know 😁).

This PR, if merged, should ideally be merged after https://github.com/apache/solr/pull/1290

### DIFF
--- a/dev-tools/scripts/buildAndPushRelease.py
+++ b/dev-tools/scripts/buildAndPushRelease.py
@@ -91,7 +91,7 @@ def getGitRev():
   return os.popen('git rev-parse HEAD').read().strip()
 
 
-def prepare(root, version, mf_username, gpg_key_id, gpg_password, gpg_home=None, sign_gradle=False):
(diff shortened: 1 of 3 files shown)


risdenk (approved):

janhoy (approved):
LGTM, but have not tested or tried to find bugs :)

PR #1293, opened 2023-01-16
SOLR-16593 : Redundant fields are present in the replica object of state.json

noblepaul:


solr/core/src/test/org/apache/solr/cloud/api/collections/SimpleCollectionCreateDeleteTest.java
```diff
@@ -104,6 +109,29 @@ public void testCreateAndDeleteThenCreateAgain() throws Exception {
     }
   }
 
+  @Test
+  @SuppressWarnings("unchecked")
```
hiteshk25, line 113: Would be good to add the test where it verifies all the expected properties. If test sees new property then it fails. So that person can justify any change in future.

noblepaul: Yeah, I think it is a good idea

PR #1295, opened 2023-01-17
SOLR-16621 Use $scope.isPermitted("my-permissiong") in security dash

janhoy:
https://issues.apache.org/jira/browse/SOLR-16621

Re-using this global function, we can simplify the js code of the dashboard. Note that the backend `/admin/info/system` already returns an array of user's permissions, not roles.

During testing I found that the app.js `isPermitted()` function did not support the `all` permission, so that's why I added that as well.

The `roleMatch()` function that I patched in #1294 is still in use one other place in `security.js`, in the `onPermFilterOptionChanged()` function. Not sure if we can get rid of that too here?

### DIFF
--- a/solr/webapp/web/js/angular/controllers/security.js
+++ b/solr/webapp/web/js/angular/controllers/security.js
@@ -320,10 +320,10 @@ solrAdminApp.controller('SecurityController', function ($scope, $timeout, $cooki
         $scope.roleNames = [];
       }
 
-      $scope.permissions = data.authorization["permissions"];
+      $scope.permissionsConfig = data.authorization["permissions"];
       $scope.permissionsTable = [];
-      for (p in $scope.permissions) {
-        $scope.permissionsTable.push(permRow($scope.permissions[p], parseInt(p)+1));
+      for (p in $scope.permissionsConfig) {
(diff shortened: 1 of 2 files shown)


janhoy:
I also found that `security.js` redefined the global `$scope.permissions` also defined in `app.js` which holds a static map of all permission names. I renamed the variable to `$scope.permissionsConfig` to avoid this, although I cannot really find any direct use in html angular directives, so I think it caused no bugs. wdyt?

risdenk (approved):

janhoy:
Would be great to get a quick review by @thelabdude on the security panel related changes.

@epugh Should we now use `var` and `const` everywhere instead of `let`?

epugh:
@janhoy we proobably should up our Javascript game, and do it in conjunction with a lint checker to ensure if we migrate to new ways, we don't slip back...   I've used `jshint` on Quepid and it works well...    I took a quick stab at getting it into gradle, but ended up stumbling on Gradle just never makes sense to me!

epugh (approved):
wasn't sure about the name `permissionsConfig`, but came around to it after reading.  LGTM.

PR #1297, opened 2023-01-17
SOLR-14336 Warn users running old version of Solr

janhoy:
https://issues.apache.org/jira/browse/SOLR-14336

This impl will simply parse release date from solr-impl string, count number of days old, and warn in Admin UI dashboard if older than a year. No warning in logs.

### DIFF
--- a/solr/webapp/web/js/angular/controllers/index.js
+++ b/solr/webapp/web/js/angular/controllers/index.js
@@ -20,6 +20,8 @@ solrAdminApp.controller('IndexController', function($scope, System, Cores, Const
   $scope.reload = function() {
     System.get(function(data) {
       $scope.system = data;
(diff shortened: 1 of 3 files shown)


risdenk (approved):

solr/webapp/web/js/angular/controllers/index.js
```diff
@@ -20,6 +20,8 @@ solrAdminApp.controller('IndexController', function($scope, System, Cores, Const
   $scope.reload = function() {
     System.get(function(data) {
       $scope.system = data;
+      const releaseDate = parse_release_date($scope.system.lucene['solr-impl-version'])
+      $scope.releaseDaysOld = (new Date() - releaseDate)/60/60/1000/24;
```
dsmiley, line 24: the ordering is nonintuitive should be /1000/60/60/24

PR #1299, opened 2023-01-18
SOLR-16625: Upgrade OWASP Dependency Check to 8.0.1

risdenk:
https://issues.apache.org/jira/browse/SOLR-16625

### DIFF
--- a/gradle/validation/owasp-dependency-check.gradle
+++ b/gradle/validation/owasp-dependency-check.gradle
@@ -26,9 +26,12 @@ configure(rootProject) {
   dependencyCheck {
     failBuildOnCVSS = propertyOrDefault("validation.owasp.threshold", 7) as Integer
     formats = ['ALL']
-    skipProjects = [':solr:solr-ref-guide']
+    skipProjects = [':solr:solr-ref-guide', ':solr-missing-doclet']
     skipConfigurations = ['unifiedClasspath', 'permitUnusedDeclared']
     suppressionFile = file("${resources}/exclusions.xml")
+    analyzers {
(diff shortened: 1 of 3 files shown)


gradle/validation/owasp-dependency-check.gradle
```diff
@@ -26,9 +26,12 @@ configure(rootProject) {
   dependencyCheck {
     failBuildOnCVSS = propertyOrDefault("validation.owasp.threshold", 7) as Integer
     formats = ['ALL']
-    skipProjects = [':solr:solr-ref-guide']
+    skipProjects = [':solr:solr-ref-guide', ':solr-missing-doclet']
```
risdenk, line 29: `solr-missing-doclet` is a build only dependency and keeps getting flagged as a Solr 1.0 version with lots of CVEs - this doesn't need to be checked.

gradle/validation/owasp-dependency-check.gradle
```diff
@@ -26,9 +26,12 @@ configure(rootProject) {
...
     skipConfigurations = ['unifiedClasspath', 'permitUnusedDeclared']
     suppressionFile = file("${resources}/exclusions.xml")
+    analyzers {
+      assemblyEnabled = false
+    }
```
risdenk, line 34: I have some .Net projects locally and for some reason dependency check is picking this up and trying to scan the project. Since we don't use .Net - we can disable the .Net Assembly scanner. See details here: https://jeremylong.github.io/DependencyCheck/dependency-check-gradle/configuration.html

PR #1301, opened 2023-01-18
SOLR-16627: Upgrade google-cloud-bom to 0.184.0, re2j to 1.6, and grpc to 1.51.0

risdenk:
https://issues.apache.org/jira/browse/SOLR-16627


solr/modules/gcs-repository/src/test/org/apache/solr/gcs/ConcurrentDelegatingStorage.java
```diff
@@ -56,6 +56,7 @@
  * synchronization across all instance methods. This allows google-cloud-nio's in-memory fake to be
  * used from multiple threads simultaneously as our tests require.
  */
+@SuppressWarnings("try")
```
risdenk, line 59: This is necessary since apparently Storage added a default close method that `throws Exception` which includes `InterruptedException` and thats expected from the Javadoc. This suppresses that warning that fails the build. The warning is coming from the compiler itself saying that close should never throw `InterruptedException` but apparently the `Storage` interface thinks otherwise.

PR #1303, opened 2023-01-18
Add temporary pipeline to test Solr using machines provided by Crave.io

uvatbc:

# Description

The Solr project needs machines that are much more powerful (eg. 100+ cores) to run all the tests.     
Crave.io offers a simple and cost effective way to provision such machines

# Solution

1. Create a separate yaml so that its possible to test the new functionality without impacting current workflows
1. Add a stage: "Download crave binary"
1. Prefix `crave run` behind the two gradlew commands

This should utlilize the "Zeroconf" Crave set up.

# Tests

Have not tested because it is a pipeline change that requires pipeline access.

# Checklist


.github/workflows/solrj-test-crave.yml
```diff
@@ -0,0 +1,41 @@
...
+  pull_request:
+    branches:
+      - 'main'
+    paths:
+      - '.github/workflows/solrj-test.yml'
+      - 'solr/solrj/**'
+
+jobs:
+  test:
+    name: Run SolrJ Tests
```
dsmiley, line 13:
```suggestion
    name: Run Solr Tests using Crave.io resources
```

.github/workflows/solrj-test-crave.yml
```diff
@@ -0,0 +1,41 @@
+name: SolrJ Tests
```
dsmiley, line 1:
```suggestion
name: Solr Tests
```

.github/workflows/solrj-test-crave.yml
uvatbc: Fixed!

.github/workflows/solrj-test-crave.yml
uvatbc: Fixed.

.github/workflows/solrj-test-crave.yml
```diff
@@ -0,0 +1,41 @@
...
+
+on:
+  pull_request:
+    branches:
+      - 'main'
+    paths:
+      - '.github/workflows/solrj-test.yml'
+      - 'solr/solrj/**'
```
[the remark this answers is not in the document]
uvatbc: Agreed. Fixed.

.github/workflows/solrj-test-crave.yml
```diff
@@ -0,0 +1,41 @@
...
+        path: |
+          ~/.gradle/caches
+        key: ${{ runner.os }}-gradle-solrj-${{ hashFiles('versions.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-gradle-solrj-
+          ${{ runner.os }}-gradle-
+    - name: Get the Crave binary
+      run: curl -s https://raw.githubusercontent.com/accupara/crave/master/get_crave.sh | bash -s --
+    - name: Initialize gradle settings
+      run: ./crave run -- ./gradlew localSettings
```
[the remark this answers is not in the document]
uvatbc: Re first comment: Yes, the output from `localSettings` will be retained between subsequent `crave run`s.     
Re second comment: I've changed the worker count to match the core count on the remote build/test node so that it matches whatever the ephemeral node is configured to be.

In other words, when we use a 16 core machine, we'll use 16 workers. When we configure crave to use `n2d-standard-224` instances, `nproc` will return 224 and automatically tell gradlew to use the maximum number of cores.

We can consider a linear scale factor for the number of cores in a later change as we understand or tweak how well gradle performs on multiple cores.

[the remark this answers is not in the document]
elyograg: Does that return the number of REAL cpu cores, or the same count you would see in something like /proc/cpuinfo?    I would wager that most modern CPUs support hyperthreading, which makes most programs think my 24 core server has 48 cores.  I don't want to disable HT, because even though it's not a true doubling of the core count, it does help some workloads run faster.  But I wouldn't want the solr build to decide it can handle 48 workers.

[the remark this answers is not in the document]
uvatbc: @elyograg `nproc` seems to return "the number of cores available to your process". Through empirical testing on my X1 carbon, this seems to mean "hyperthreaded cores".

Testing on the `n1-standard-96` set up that we have for Solr testing got me:
```
19:23:12 uv@uv-x1:~/src/github.com/java/solr/solr$ crave ssh -- nproc
96
```
So it looks like it is hyperthreaded cores for sure.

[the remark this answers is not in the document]
dsmiley: Half the number of available processors reported by the JVM; the JVM counts hyper threads individually.  Since 2x threads per core is common, our build divided this by two to guess at the number of physical cores.  Then this is capped at 12.  Logic is in `./gradle/generation/local-settings.gradle`.  There are lots of concurrent threads in the JVM so I this this approach of using physical cores in spite of hyper threading makes sense.  On a machine with 96 threads on 48 cores (which I used last weekend playing around on Crave.io), 48 gradle workers worked well (and so did 24).

.github/workflows/tests-via-crave.yml
```diff
@@ -0,0 +1,38 @@
...
+        key: ${{ runner.os }}-gradle-solrj-${{ hashFiles('versions.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-gradle-solrj-
+          ${{ runner.os }}-gradle-
+    - name: Get the Crave binary
+      run: curl -s https://raw.githubusercontent.com/accupara/crave/master/get_crave.sh | bash -s --
+    - name: Initialize gradle settings
+      run: ./crave run -- ./gradlew localSettings
+    - name: Test the Solr Package
+      run: ./crave run -- './gradlew --max-workers=`nproc` test'
```
dsmiley, line 38:
```suggestion
      run: ./crave run -- './gradlew --max-workers=`nproc` -Ptests.jvms=48 test'
```

I did some exploration and found that it's not enough to set the workers in this way as the "tests.jvms" will default to the value in gradle.properties which is again capped at 12.  When I used 48 workers & test JVMs on this machine, the tests finished in 5:45 which is awesome!  If you wish we could use nproc again thus 96 for both.  I did one such build and it pegged the cpu but then stalled out, probably due to misbehavior in a test or something.  I ultimately killed it.  Could just be bad luck.

dsmiley:
I imagine Crave might work on a "self hosted runner" _eventually_ instead of indirectly using GitHub runner:
https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners

dsmiley:
Looking at the GH Action results, the crave executable failed because of no crave conf:
> Error: Configuration file not found. Zero conf not found: No matching projects found for URL "https://github.com/apache/solr"

.github/workflows/tests-via-crave.yml
```diff
@@ -0,0 +1,38 @@
...
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
+    - uses: actions/cache@v2
+      with:
+        path: |
+          ~/.gradle/caches
+        key: ${{ runner.os }}-gradle-solrj-${{ hashFiles('versions.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-gradle-solrj-
+          ${{ runner.os }}-gradle-
```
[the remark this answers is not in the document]
uvatbc: Removed.

.github/workflows/tests-via-crave.yml
uvatbc: Updated as requested.

uvatbc:
Added the link "https://github.com/apache/solr" to the ZeroConf configuration as another alias for the Solr git repository.
The Github action should now work: The client binary should be able to recognize this as a ZeroConf request.

@Self hosted runner: Makes sense. Next step.

.github/workflows/tests-via-crave.yml
```diff
@@ -12,8 +12,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    # Setup
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
```
epugh, line 15: gets rid of the annoying "node needs updating" message too!

uvatbc:
Ok so that worked... Thanks @dsmiley !
Looks like really good CPU utilization of the 96 cores.

image

dsmiley:
If I'm not mistaken, there are 96 hyper threads on 48 *cores*?

The first two minutes in Crave was Crave downloading Solr's dependencies :-(. Hopefully won't repeat for each build.
Also the "fetch all branches" option I used was a blunt approach to satisfy what Crave needs.  There may be a better way.  A minute was lost in checking everything out -- dunno if it'll be repeated.  I'll re-run this to see.

The net of these delays and actual test execution was 11 minutes (as we can see here in the "all checks have passed" summary.  I'm happy with that :-)

uvatbc:
There's ways to eliminate both those "startup" times. Just to note them down:
Downloading Solr deps: That might be something to be put into the Docker container assigned to the Solr project. Can you point me to some logs to understand whats being downloaded? I'll try to put all of it into the container so that its not being pulled repeatedly.

Fetch all the branches: We may have a trick for that that involves eliminating the clone altogether, but let me try it out myself before I confirm it works.

dsmiley:
I think I'm done tweaking this thing.  Still 11 minutes, although the milestones of what happens changed a bit.  The checkout-everything delay is still there, as are Crave side delays of downloading everything.  Nevertheless; this is awesome -- worth merging!  I want to bring attention to this issue on the Solr dev list.

Crafting a custom Docker build image with the dependencies isn't worth the maintenance.  Build infrastructure (be it GitHub Actions or Crave) needs a feature of caching certain directories to avoid repeated work.  I saw this in GHA.  A fast network-local caching proxy (either HTTP generic or Maven repo specific) could ameliorate that and be complementary.

PR #1304, opened 2023-01-19
SOLR-16618 Admin UI Analysis page should include dynamic fields

stillalex:
https://issues.apache.org/jira/browse/SOLR-16618

# Description

Adding dynamic fields to analysis page, heavily inspired by the schema page.
As mentioned on Jira, the boolean type analyzer does not parse (code breaks) so I added a null check and a "?" as a fallback value.

# Solution

# Tests

# Checklist


stillalex:
@epugh please take a look

solr/webapp/web/js/angular/controllers/analysis.js
```diff
@@ -137,16 +168,24 @@ solrAdminApp.controller('AnalysisController',
         }
 
         if (fieldOrType == "fieldname") {
-          $location.search("analysis.fieldname", name);
-          $location.search("analysis.fieldtype", null);
+            $location.search("analysis.fieldname", name);
```
epugh, line 171: I don't love the whitespace change, however I see we have a werid mix of two and four space indentations on if/else throughout this method!   We need some code styling tools for jS ;-)

solr/webapp/web/js/angular/controllers/analysis.js
```diff
@@ -21,33 +21,64 @@ solrAdminApp.controller('AnalysisController',
...
-          for (var type in data.schema.types) {
-            $scope.fieldsAndTypes.push({
-              group: "Types",
-              value: "fieldtype=" + type,
-              label: type});
-          }
+          $scope.fieldsAndTypes = getFieldsAndTypes(data.schema);
           $scope.core = $routeParams.core;
+          $scope.parseQueryString(data.schema);
+          // @todo - set defaultSearchField either to context["analysis.fieldname"] or context["analysis.fieldtype"]
```
epugh, line 27: is this still a valid todo?

epugh: I don't see defaultSearchField used in the JS or HTML?

solr/webapp/web/js/angular/controllers/analysis.js
```diff
@@ -21,33 +21,64 @@ solrAdminApp.controller('AnalysisController',
...
+      $scope.verbose = true;
 
-        $scope.parseQueryString();
-        // @todo - set defaultSearchField either to context["analysis.fieldname"] or context["analysis.fieldtype"]
+      var getFieldsAndTypes = function(schema) {
+        var aggregatedFields = schema.fields;
+        for (var field in schema.fields) {
+          var copy_dests = schema.fields[field].copyDests;
+          for (var i in copy_dests) {
+            var copy_dest = copy_dests[i];
```
epugh, line 37: camelCase instead of snake_case

solr/webapp/web/js/angular/controllers/analysis.js
```diff
@@ -21,33 +21,64 @@ solrAdminApp.controller('AnalysisController',
...
+        var fieldsAndTypes = [];
+        var fields = Object.keys(aggregatedFields).sort();
+        for (var i in fields) {
+          fieldsAndTypes.push({
+            group: "Fields",
+            value: "fieldname=" + fields[i],
+            label: fields[i]
+          });
+        }
+        var dynamic_fields = Object.keys(schema.dynamicFields).sort();
```
epugh, line 53: camelCase

solr/webapp/web/js/angular/controllers/analysis.js
stillalex: I am not exactly sure, left the todo there just in case

solr/webapp/web/js/angular/controllers/analysis.js
stillalex: yep, fixed the formatting. tried as much as possible to keep the original style

solr/webapp/web/js/angular/controllers/analysis.js
epugh: Cool, what prompted this?     I guess I am wondering if we need this?   I am playing with the code, and it *appears* to all work ;-)

epugh:
Screenshot at Jan 23 10-34-38

I see the `?` from the bool...   I was trying to figure out if there is something we could put there, like "Bool" or something?   But regardless, I like how this works...

epugh:
image

Ahh...   Now I see the `@todo` reason....    I wonder if the right fix is to pick the drop down, but not trigger the Luke.schema call?   Does it make sense to submit the call when we have no Field Value (Index) or Field Value (Query) defined?    

Maybe we don't need the default value if we just show the page, but don't submit it to the backend, which then avoids the error message.

stillalex:
> I see the ? from the bool... I was trying to figure out if there is something we could put there, like "Bool" or something? But regardless, I like how this works...

I would keep it generic, but if "?" is too cryptic, we could put the original value in there like "org.apache.solr.schema.BoolField$1$1", but that might not be much better.

epugh:
> 

I started thinking about another param...  but that feels a bit like the `goto` statement of web dev.   How hard would it be to change the query to submit when you click the button?   Are there other places we link to the schema analysis page and have it auto run, so that requiring a button click would make the user experience worse?

epugh:
> > I see the ? from the bool... I was trying to figure out if there is something we could put there, like "Bool" or something? But regardless, I like how this works...
> 
> I would keep it generic, but if "?" is too cryptic, we could put the original value in there like "org.apache.solr.schema.BoolField$1$1", but that might not be much better.

Let's stick with `?`, cause the mouse over gives the long version.  And I suspect anything more then two or three characters would mess up the layout!

stillalex:
> How hard would it be to change the query to submit when you click the button? Are there other places we link to the schema analysis page and have it auto run, so that requiring a button click would make the user experience worse?

I fear this is going off scope for this PR, this behavior change would deserve its own PR and discussion (and probably someone that actually knows their way around js frontends).

epugh:
> 

Sure!  I can buy that.   So...   leave the todo?   It otherwise looks good to merge.

stillalex:
> Sure! I can buy that. So... leave the todo? It otherwise looks good to merge.

yep, leave the todo. there is only that string matching issue in the combo which I don't know how to approach. open to suggestions or we could leave it for another time.

stillalex:
@epugh gentle ping. are we good to merge, or is there something that needs more work?

epugh:
@magibney and I chatted, and we're going to leave it open till end of the week in case anyone else has concerns about having a potentially longer list of dynamic fields in the drop down ;-).  Sorry, I meant to update here.   If no one has strong concerns, then we'll merge.

stillalex:
makes sense, thanks for the update @epugh!

stillalex:
thank you @epugh!

PR #1307, opened 2023-01-23
Small code improvements in facet component

renatoh:
backporting "Debugging FacetComponent I noticed two little things which can be improved"


mkhludnev:
` Files changed 5,000+`
Not good. Ok maybe I put it wrong, but it's cherry-picking as described in dev docs.

PR #1312, opened 2023-01-24
SOLR-16635: Resolve crypto keys file path in MiniSolrCloudCluster

tflobbe:
By resolving the full path of the pre-generated crypto keys in `MiniSolrCloudCluster` we make sure it also works for third party projects testing using Solr test framework

### DIFF
--- a/solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
+++ b/solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
@@ -23,6 +23,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.lang.invoke.MethodHandles;
+import java.net.URL;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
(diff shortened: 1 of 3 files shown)


solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
```diff
@@ -89,6 +90,10 @@
 public class MiniSolrCloudCluster {
 
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  private static final URL PRE_GENERATED_PRIVATE_K_URL =
```
anshumg, line 93: nit-pick, but can we use `KEY` instead of `K` i.e. `PRE_GENERATED_PRIVATE_KEY_URL`

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
```diff
@@ -89,6 +90,10 @@
...
+      MiniSolrCloudCluster.class.getClassLoader().getResource("cryptokeys/priv_key512_pkcs8.pem");
+  private static final URL PRE_GENERATED_PUBLIC_K_URL =
```
anshumg, line 95: Same here

risdenk (approved):

PR #1318, opened 2023-01-30
SOLR-16636: ZkStateReader.waitForState should not register a watch if state already matches predicate

noblepaul:


solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -1682,6 +1687,37 @@ public DocCollection getCollectionLive(String coll) {
...
+    }
+    if (stat != null) {
+      if (!c.isModified(stat.getVersion(), stat.getCversion())) {
+        // we have the latest collection state
+        return c;
+      }
+      if (c.isPerReplicaState() && c.getChildNodesVersion() < stat.getCversion()) {
+        // only PRS is modified. just fetch it and return the new collection
+        return c.copyWith(PerReplicaStatesFetcher.fetch(collectionPath, zkClient, null));
+      }
```
tflobbe, line 1716: This wouldn't add the updated DocCollection to the cache, right? So any future calls to this method will continue on the same path (exists + fetch), until some other code path updates the cache, is that right?

noblepaul: actually, there is another PR that is waiting to be merged #1242 . that ensures that the this update is in-place. 

I'm waiting for perf tests to come back to merge that

hiteshk25:
I wonder this optimization should go into [get()](https://github.com/apache/solr/blob/e7e487e1694231b6e586276742e598de06acd9bf/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java#L809) method?

noblepaul:
> I wonder this optimization should go into [get()](https://github.com/apache/solr/blob/e7e487e1694231b6e586276742e598de06acd9bf/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java#L809) method?

The semantics is different there

PR #1321, opened 2023-01-31
SOLR-13396 - Disable deletion of unknown cores by default

colvinco:
https://issues.apache.org/jira/browse/SOLR-13396

# Description

This is a quick fix for SOLR-13396 to prevent the automatic deletion of cores that aren't referenced in the ZK Clusterstate.

# Solution

I've added a system property (`solr.deleteUnknownCores`) and associated ENV flag (`SOLR_DELETE_UNKNOWN_CORES`) to determine whether unreferenced cores should be deleted. This is a simple fix that should work for most people who have had an issue with SOLR-13396 (simply disabling the automatic deletion). A "better" fix may exist in the form of adding additional safety checks to try and detect misconfiguration, but this still feels like a safe option to have in addition to that in any case.

The default behavior is changed so cores will not be deleted automatically OOTB.

# Tests

I've found the test that was added for SOLR-12066 (https://github.com/apache/solr/blob/main/solr/core/src/test/org/apache/solr/cloud/DeleteInactiveReplicaTest.java) and updated it to set the new deleteUnknownCores property, otherwise it would fail. It might still be good to create a separate regression test for what happens when the property isn't set.

# Checklist


janhoy:
My point is the property name could be delete… with a default of true.

Users would need to set it to false to override. In 10.0 the default could change, unless we found a better / smarter solution to the entire problem.

Or we could default it to false if we are all ok with a change in behavior. The conservative approach is of course the safest.

janhoy (changes requested):
CHANGES.txt entry and short Reference Guide docs?

colvinco:
Where do you think this would fit best into the Reference Guide? I've not seen an obvious section to add it to, it doesn't feel big enough to get its own page, and I don't think there's a page that just lists out supported system properties / solr.in.sh options.

It could maybe slot in somewhere around here https://solr.apache.org/guide/solr/9_2/deployment-guide/taking-solr-to-production.html#zookeeper-chroot), or in https://solr.apache.org/guide/solr/9_2/deployment-guide/solrcloud-shards-indexing.html / https://solr.apache.org/guide/solr/9_2/deployment-guide/solrcloud-recoveries-and-write-tolerance.html

Is there somewhere better you have in mind?

janhoy:
Yea, good question.

If we decide to change the default to not delete (which I perhaps think we should), you'd for sure also add an entry to Major Changes in 9.3.

The feature is also self documenting in solr.in.sh of course, but a short mention in taking-solr-to-production would not hurt.

colvinco:
Sorry it's taken a long time, I've finally got back to this. See what you think?
Thanks

HoustonPutman (approved):
This looks good to me, but I would probably include Jan in the Changelog entry.

janhoy (approved):

colvinco:
Based on what we've discussed previously and what @elyograg said on Slack, it seems like disabling it by default is reasonable.

PR #1322, opened 2023-02-01
SOLR-16641 - Generate gradle.properties from template

colvinco:
This is a duplicate of #1320 atm, I just want to test the crave.io check


gradlew
```diff
@@ -163,10 +163,18 @@ fi
...
 CLASSPATH=$GRADLE_WRAPPER_JAR
 
-# Don't fork a daemon mode on initial run that generates local defaults.
-GRADLE_DAEMON_CTRL=
+# Generate gradle.properties if they don't exist
 if [ ! -e "$APP_HOME/gradle.properties" ]; then
-    GRADLE_DAEMON_CTRL=--no-daemon
+    echo "Populating gradle.properties from gradle.properties.template."
+    # Do the copy first. If the next steps fail, the gradle.properties will still be usable
+    cp gradle.properties.template gradle.properties
```
janhoy, line 170: Should this be absolute path?
```suggestion
    cp gradle.properties.template "$APP_HOME/gradle.properties"
```

gradlew
```diff
@@ -163,10 +163,18 @@ fi
...
+    NUM_CORES=$(grep ^cpu\\scores /proc/cpuinfo | uniq |  awk '{print $4}')
```
janhoy, line 171: `/proc/cpuinfo` won't work on macOS. Either detect OS and handle it specifically, or fall back to some decent default?

janhoy (commented):
I too like the approach of generating it in gradlew. But probably our customization should be clearly marked with some `# BEGIN/END SOLR CUSTOMIZATION` marker? Or we just do a one-liner `source solr_preinit.sh` which is easier to patch into new versions of gradlew?

janhoy:
Sorry for adding the review here instead of #1320, it was a mistake. I can add it there if you want, or I'll wait a few days to see how that PR evolves and review again.

colvinco:
Sorry, I'll close this one but I'll remember the comments on it

PR #304, opened 2021-09-24
SOLR-15455: joint solr/lucene development experiments

dweiss:

PR #1323, opened 2023-02-01
Improvements to joint Lucene/Solr dev workflow

magibney:
This is a followup to SOLR-15455 / #304

I'm happy to open a Jira issue for this if that would be helpful, or push to a branch that allows collaborative work.

In the course of exercising the joint Lucene/Solr local development workflow, a few opportunities for improvement presented themselves. I've separated these into logical commits. Some of these commits are more important than others; I think they're all good, but 87d041e9719b14d7b6a184d976fb88168fe4b8f2 and 64846a40af386b745e172b95577c260fbfcad8d7 in particular I'd say are just "nice to have" -- happy to revert them from this PR if anyone has a problem with them.

### DIFF
--- a/dev-tools/solr-missing-doclet/build.gradle
+++ b/dev-tools/solr-missing-doclet/build.gradle
@@ -20,7 +20,7 @@ plugins {
 }
 
 version = "1.0.0-SNAPSHOT"
-group = "org.apache.lucene.tools"
+group = "org.apache.solr.tools"
 description = 'Doclet-based javadoc validation'
 
 sourceCompatibility = JavaVersion.VERSION_11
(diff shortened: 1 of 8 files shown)


risdenk (approved):

magibney:
@dweiss I'd like to make sure you have a chance to look at this, if you're able. Particularly:
1. 3f6fb21ca7a36241520572ae9a906c829ecb10ef: I recall it was a bit tricky to read project properties at configuration time -- the solution that this replaces would read project props specified on the command-line, but wasn't able to read out of the `gradle.properties` file. The solution proposed here should be a bit cleaner, and is capable of doing both.
2. 600732f558e3052f5569ab3d189365f9b1edb5d3: I _think_ this is safe to do? If I'm right about it being arbitrary, this seems like the right way to go, and avoids conflicts.

dweiss:
Both changes look fine for me, @magibney

magibney:
Thanks, Kevin and Dawid! I plan to merge this today.

PR #1328, opened 2023-02-03
SOLR-16645: Extend start script to support ss and netstat as alternatives to lsof

colvinco:
https://issues.apache.org/jira/browse/SOLR-16645

# Description

The current start command requires `lsof` to be available. If you don't have `lsof` it does a hardcoded 10 second wait and then assumes the process has started. This updates the script to support netstat and ss as alternatives.

I'm also not sure if it's worth putting it this in, as it adds extra paths through the script so there's more potential for things to go wrong, as different versions of the tools may take different arguments, and it's generally not hard to install `lsof`.
However we have been using this logic in our product for about 4 years, so I thought I would offer it up and see if it's of interest.

# Solution

Use `which` to check if `lsof`, `netstat`, `ss` is available and pick the first one, then execute the appropriate command.

# Tests

Manually tested

# Checklist

### DIFF
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -2176,7 +2176,7 @@ function check_port_in_use() {
 
   case $tool in
     lsof)
-      lsof -iTCP:${port} 2>/dev/null -sTCP:LISTEN 2>/dev/null | grep -ow "java\s*\?[0-9]\+" | grep -ow "[0-9]\+"
+      lsof -PniTCP:${port} -sTCP:LISTEN 2>/dev/null | grep -ow "java\s*\?[0-9]\+" | grep -ow "[0-9]\+"
       ;;
     ss)
       ss -ntlp "( sport = :${port} )" 2>/dev/null | grep -ow '"java",pid=[0-9]\+' | grep -ow "[0-9]\+"
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -177,7 +177,7 @@ else
   fi
   JAVA_VENDOR="Oracle"
    # OpenJ9 was previously known as IBM J9, this will match both
-  if [ "$(echo "$JAVA_VER" | grep -i -E "OpenJ9|IBM J9")" != "" ]; then
+  if [ "$(echo "$JAVA_VER" | grep -i "OpenJ9\|IBM J9")" != "" ]; then
       JAVA_VENDOR="OpenJ9"
   fi
 fi
@@ -2176,13 +2176,13 @@ function check_port_in_use() {
 
   case $tool in
     lsof)
-      lsof -iTCP:${port} 2>/dev/null | grep "LISTEN" | grep -Eow "java\s*?[0-9]+" | grep -Eow [0-9]+
+      lsof -iTCP:${port} 2>/dev/null -sTCP:LISTEN 2>/dev/null | grep -ow "java\s*\?[0-9]\+" | grep -ow "[0-9]\+"
       ;;
     ss)
-      ss -ntlp "( sport = :${port} )" 2>/dev/null | grep -Eow "\"java\",(pid=)?[0-9]+?" | grep -Eow "[0-9]+"
+      ss -ntlp "( sport = :${port} )" 2>/dev/null | grep '"java\"' | grep -o "pid=[0-9]\+" | grep -ow "[0-9]\+"
       ;;
     netstat)
-      netstat -nlp 2>/dev/null | grep ":${port} " | grep -Eow " [0-9]+/java" | grep -Eow "[0-9]+"
+      netstat -nlp 2>/dev/null | grep ":${port} " | grep -ow " [0-9]\+/java" | grep -ow "[0-9]\+"
       ;;
   esac
 }
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -2160,9 +2160,8 @@ function mk_writable_dir() {
 
 # Check whether at least one of lsof/netstat/ss are available
 function get_port_tool() {
-  for tool in 'lsof' 'netstat' 'ss'; do
-      which $tool &> nul
-      if [[ $? -eq 0 ]]; then
+  for tool in 'lsof' 'ss' 'netstat'; do
+      if [[ -x "$(command -v $tool)" ]]; then
         echo $tool
         return
       fi
@@ -2172,9 +2171,10 @@ function get_port_tool() {
 # Will get the PID of the java process running on the specified port, if one is running
 # Otherwise empty result. Use get_port_tool to check whether lsof/ss/netstat is installed first
 function check_port_in_use() {
-  local port=$1
+  local tool=$1
+  local port=$2
 
-  case $(get_port_tool) in
+  case $tool in
     lsof)
       lsof -iTCP:${port} 2>/dev/null | grep "LISTEN" | grep -Eow "java\s*?[0-9]+" | grep -Eow [0-9]+
       ;;
@@ -2340,14 +2340,14 @@ function start_solr() {
 	echo "RNG might not work properly. To check for the amount of available entropy, use 'cat /proc/sys/kernel/random/entropy_avail'."
 	echo ""
     fi
-    # no lsof on cygwin though
-    if [ -n "$(get_port_tool)" ]; then
+    local port_tool=$(get_port_tool)
+    if [ -n "$port_tool" ]; then
       echo -n "Waiting up to $SOLR_START_WAIT seconds to see Solr running on port $SOLR_PORT"
       # Launch in a subshell to show the spinner
       (loops=0
       while true
       do
-        running=$(check_port_in_use $SOLR_PORT)
+        running=$(check_port_in_use $port_tool $SOLR_PORT)
         if [ -z "${running:-}" ]; then
           slept=$((loops * 2))
           if [ $slept -lt $SOLR_START_WAIT ]; then
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -2176,7 +2176,7 @@ function check_port_in_use() {
 
   case $(get_port_tool) in
     lsof)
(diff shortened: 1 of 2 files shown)


solr/bin/solr
```diff
@@ -2158,6 +2158,35 @@ function mk_writable_dir() {
...
+}
+
+# Will get the PID of the java process running on the specified port, if one is running
+# Otherwise empty result. Use get_port_tool to check whether lsof/ss/netstat is installed first
+function check_port_in_use() {
+  local port=$1
+
+  case $(get_port_tool) in
+    lsof)
+      lsof -iTCP:${port} 2>/dev/null | grep "LISTEN" | grep -Eow "java\s*?[0-9]+" | grep -Eow [0-9]+
```
colvinco, line 2179:
```suggestion
      lsof -t -PniTCP:${port} -sTCP:LISTEN || :
```

elyograg:
ss is probably more likely to have the same options on multiple platforms than netstat.

solr/bin/solr
```diff
@@ -2158,6 +2158,35 @@ function mk_writable_dir() {
   fi
 }
 
+# Check whether at least one of lsof/netstat/ss are available
+function get_port_tool() {
+  for tool in 'lsof' 'netstat' 'ss'; do
```
colvinco, line 2163:
```suggestion
  for tool in 'lsof' 'ss' 'netstat'; do
```

solr/bin/solr
```diff
@@ -2158,6 +2158,35 @@ function mk_writable_dir() {
...
+      which $tool &> nul
```
magibney, line 2164: Looks like it might be preferable to use `command` in place of `which`? -- (there's [precedent](https://github.com/apache/solr/blob/e35347e7cc12d91af91cb38f9cef43ccc7a7b40e/solr/bin/solr#L125-L126) in this script already)

solr/bin/solr
```diff
@@ -2312,13 +2341,13 @@ function start_solr() {
 	echo ""
     fi
     # no lsof on cygwin though
-    if lsof -v 2>&1 | grep -q revision; then
+    if [ -n "$(get_port_tool)" ]; then
```
magibney, line 2344: rather than seeking the port tool a second time (calling `get_port_tool()` from within the `check_port_in_use()` function), would it be just as clean/easy to assign the output of `get_port_tool` to a variable, check that variable for `[ -n "$PORT_TOOL" ]`, and pass `$PORT_TOOL` as an arg to `check_port_in_use()`?

Fine as-is, but seems there'd be very little downside to this change?

solr/bin/solr
```diff
@@ -2158,6 +2158,35 @@ function mk_writable_dir() {
...
+  case $(get_port_tool) in
+    lsof)
+      lsof -iTCP:${port} 2>/dev/null | grep "LISTEN" | grep -Eow "java\s*?[0-9]+" | grep -Eow [0-9]+
+      ;;
+    ss)
+      ss -ntpa 2>/dev/null | grep ":${port} " | grep 'LISTEN ' | grep -Eow "\"java\",(pid=)?[0-9]+?" | grep -Eow "[0-9]+"
+      ;;
+    netstat)
+      netstat -nlp 2>/dev/null | grep ":${port} " | grep -Eow " [0-9]+/java" | grep -Eow "[0-9]+"
+      ;;
```
magibney, line 2186: Analogous to the change above (restoring the more targeted use of `lsof`, avoiding the need for `grep`), I wonder if there are more targeted (but still portable) versions of this command for `ss` and `netstat`?

solr/bin/solr
```diff
@@ -2312,13 +2341,13 @@ function start_solr() {
...
     # no lsof on cygwin though
-    if lsof -v 2>&1 | grep -q revision; then
+    if [ -n "$(get_port_tool)" ]; then
       echo -n "Waiting up to $SOLR_START_WAIT seconds to see Solr running on port $SOLR_PORT"
       # Launch in a subshell to show the spinner
       (loops=0
       while true
       do
-        running=$(lsof -t -PniTCP:$SOLR_PORT -sTCP:LISTEN || :)
+        running=$(check_port_in_use $SOLR_PORT)
```
magibney, line 2350: The ` || : ` in the initial command is curious. I infer that this protects against the possibility that a failed `lsof` command might still write to stdout, which would get captured into the `running` variable. Whatever the reason, might we want to either add ` || : ` to the `ss` and `netstat` variants of the port command, or move the ` || : ` back to the outer level here (ensuring that failed invocations of `lsof`/`ss`/`netstat` propagate failure exit codes via `check_port_in_use()`?)

magibney (commented):
Thanks for this! questions/comments on code

solr/bin/solr
colvinco: My thinking here was that it's better to verify that the process that's listening is a java process, rather than some other process that is running the port you expect Solr on.

However, that's partly because my original use of these scripts was in a more general context. In this context, I guess we already have confidence that the port wasn't already in use?

solr/bin/solr
colvinco: I must admint can't see that it would have had any effect here. But I may be coming it at from the wrong angle. I see it was originally added by @madrob https://github.com/apache/solr/commit/3ed17c2737db038c2908a1475099647fb4ef5b38  

If I create a script:
```
echo hello
exit 1
```
And call it with ` || :` the stdout still gets captured. The exit code will be 0 rather than 1, but I don't think that's adding anything here?

magibney: ah yes, you're right! Maybe it just ensures that the script will proceed, regardless of how the script/environment does error handling (`set -e` etc.)?

solr/bin/solr
colvinco: Right yep. 

I see `ss` supports filtering to listening sockets and filtering by ports - `ss -l -at '( sport = :8983 )'`, but I don't know whether those options have been standard for a long time or not, but I also can't see anything suggesting that they're new.
The `netstat` `-l` is already filtering to listen, but it doesn't seem like there's an explicit port filter for it.

I'll change the `ss` command to use those options

magibney: Great! thanks for looking into it. This seems reasonable, and maybe if anyone else wants to weigh in wrt how standard these flags are, that could be helpful.

Assuming continuing to use grep, we really don't _need_ "extended" grep here (`grep -E` -- nor [here](https://github.com/apache/solr/blob/e35347e7cc12d91af91cb38f9cef43ccc7a7b40e/solr/bin/solr#L180)). Removing the `-E` flag would just mean a little bit of extra escaping of special characters in the pattern -- not sure it's worth it, totally fine as-is, just thought I'd mention.

colvinco: Thought I would try it anyway. 
Also hadn't realized that I had lost the `-sTCP:LISTEN` filter from lsof, so that's back in :)

solr/bin/solr
madrob: Yes, if lost doesn't exist we don't want the script to error out.

solr/bin/solr
```diff
@@ -2158,6 +2158,35 @@ function mk_writable_dir() {
...
+
+# Will get the PID of the java process running on the specified port, if one is running
+# Otherwise empty result. Use get_port_tool to check whether lsof/ss/netstat is installed first
+function check_port_in_use() {
+  local tool=$1
+  local port=$2
+
+  case $tool in
+    lsof)
+      lsof -iTCP:${port} 2>/dev/null -sTCP:LISTEN 2>/dev/null | grep -ow "java\s*\?[0-9]\+" | grep -ow "[0-9]\+"
```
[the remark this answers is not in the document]
colvinco: Oh yep, sorry copying too many things around now.

solr/bin/solr
colvinco: > if lost doesn't exist

Sorry do you mean if `lsof` doesn't exist?
With these changes, `command -v` will already have been used to check if lsof / ss / netstat is available, and this won't be called if they aren't, it will already have branched to https://github.com/apache/solr/blob/dcb7d2406b574e24fa8636cf717a8af2039c79b4/solr/bin/solr#L2369

solr/bin/solr
colvinco: Ah that's a pain. Yeah, on linux it gives the PID/program name column
```
# netstat -nlp 2>/dev/null
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp        0      0 127.0.0.1:8983          0.0.0.0:*               LISTEN      22862/java
tcp        0      0 0.0.0.0:9983            0.0.0.0:*               LISTEN      3380/java
tcp        0      0 127.0.0.1:7983          0.0.0.0:*               LISTEN      22862/java
```

https://apple.stackexchange.com/a/327967 suggests `-v` is the nearest alternative.

epugh:
What's the status on this?  I stumbled across it as part of looking for tickets related to the Solr CLI...  SOLR-16757.   Is this ready to be merged?

colvinco:
Hello, sorry I've been away for a while. The outstanding issue is to work out what to do with netstat as the standard arguments on mac differ to linux. https://github.com/apache/solr/pull/1328#discussion_r1096105999

So I guess the options are to either a) not include netstat, b) find common arguments that work on both OSes, or c) check the OS (or the arguments supported by the available netstat) and switch between them. 

Option b sounds like it would be best if it's possible, but I don't have a mac to test with.
As it's only working out whether a port is in use, returning the pid/process name isn't that important, so maybe just `netstat -nl 2>/dev/null | grep "LISTEN" | grep ":${port} "` would work well enough cross platform?

epugh:
@colvinco if we want to push this one along, I think your proposed solution made sense...    I'm going to close the associated JIRA, however happy to reopen this and work with you to get it to done...

colvinco:
@epugh no worries, will see what https://issues.apache.org/jira/browse/SOLR-17450 brings. To be fair, with containers it's not hard to ensure that lsof is available anyway so the need for alternatives diminished.

PR #1330, opened 2023-02-03
SOLR-15955: Jetty 10.x upgrade notes

risdenk:
https://issues.apache.org/jira/browse/SOLR-15955

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -51,6 +51,8 @@ New Features
 
 * SOLR-16596: Learning To Rank - Added support for null feature values in multiple additive trees models (Anna Ruggero via Alessandro Benedetti)
 
+* SOLR-15955: Upgrade to Jetty 10.x (Mark Miller, Kevin Risden)
+
 Improvements
 ---------------------
 
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -51,8 +51,6 @@ New Features
 
 * SOLR-16596: Learning To Rank - Added support for null feature values in multiple additive trees models (Anna Ruggero via Alessandro Benedetti)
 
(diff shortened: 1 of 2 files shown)


solr/CHANGES.txt
```diff
@@ -51,6 +51,8 @@ New Features
 
 * SOLR-16596: Learning To Rank - Added support for null feature values in multiple additive trees models (Anna Ruggero via Alessandro Benedetti)
 
+* SOLR-15955: Upgrade to Jetty 10.x (Mark Miller, Kevin Risden)
+
```
risdenk, line 55: Not sure that under `"New Feature"` makes sense, but figured it was good to call out up top.

tflobbe (approved):

solr/CHANGES.txt
dsmiley: Definitely not "New Feature" for bumping up a dependency unless somehow the user gets some new ability here.

risdenk: Thats fair - the only existing place that made rough sense was "Other Changes" or "Build" based on the existing categories being:

* New Features
* Improvements
* Optimizations
* Bug Fixes
* Build
* Other Changes

dsmiley: Typically goes in Other Changes.  Could be justified in some other categories if we expect an appreciable difference.
Obviously this is a big dependency upgrade so it should be called out when we do the release notes, maybe mentioning some benefits if you know of any in particular that matter.

epugh (approved):
Leaving aside the "New Feature" discussion, the rest makes sense.

solr/CHANGES.txt
risdenk: moved to other changes

PR #1334, opened 2023-02-04
SOLR-16590: Standardize Builder method names in solr clients

epugh:
https://issues.apache.org/jira/browse/SOLR-16590

# Description

We want to use `with` instead of `set` or even just a bare property name in our Builders.   

# Solution

Move to using nicer builder names, but try not to be pedantic on the pattern.   

I thought there were many many more places in our builders that needed updating, but it doesn't seem to be a widespread problem.

# Tests

Update tests and rerun.


epugh:
Fingers crossed that Crave runs on this and saves my poor laptop (and me) an hour of waiting for tests to pass before proceeding ;-)

epugh:
Okay, I think this is ready to run.  The Crave failure is an intermittant one ;-(

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -1096,38 +1096,89 @@ public Builder withTheseParamNamesInTheUrl(Set<String> urlParamNames) {
...
+    /**
+     * @deprecated Please use {@link #setUseHttp1_1(boolean)}
+     */
+    @Deprecated(since = "9.2")
     public Builder useHttp1_1(boolean useHttp1_1) {
+      setUseHttp1_1(useHttp1_1);
+      return this;
+    }
+
+    public Builder setUseHttp1_1(boolean useHttp1_1) {
```
risdenk, line 1140: How about `withHttp1_1`? not sure `setUseHttp1_1` is an improvement over `useHttp1_1`

risdenk (changes requested):

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
epugh: Or maybe `setHttp1_1`?    or `enableHttp_1_1()` and another `enableHttp_2()` that would flip taht flag?    This one felt a bit odd ot me too!

epugh: okay, went with `withHttp1_1`...    Looking forward to getting the merged so I can move on to sorting out the time parameters ;-)

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -1096,38 +1096,89 @@ public Builder withTheseParamNamesInTheUrl(Set<String> urlParamNames) {
...
+    public Builder withIdleTimeout(int idleConnectionTimeout) {
       this.idleTimeout = idleConnectionTimeout;
       return this;
     }
 
+    /**
+     * @deprecated Please use {@link #withHttp1_1(boolean)}
+     */
+    @Deprecated(since = "9.2")
     public Builder useHttp1_1(boolean useHttp1_1) {
```
dsmiley, line 1135: I actually think "useHttp1_1" sounds nicer than "with" in this case but if you think is an improvement then fine.

dsmiley (approved):

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
epugh: humm...    I like useHttp1_1 better than setHttp1_1..   and I don't love `withHttp1_1`...       I think I'll change back!

PR #1341, opened 2023-02-08
SOLR-16623:SolrClientTestRule for JettySolrRunner

joshgog:
https://issues.apache.org/jira/browse/SOLR-16623

# Description

This PR introduces a SolrClientTestRule for JettySolrRunner

# Solution

Introduced SolrClientTestRule abstract class that extends JUnit's ExternalResource class, which is a TestRule that allows for specifying methods to be run before and after a test. SolrJettyTestRule is a concrete implementation of the abstract class and is declared as a ClassRule field in SolrJettyTestBase.

# Tests

# Checklist

### DIFF
--- a/solr/core/src/test/org/apache/solr/schema/TestBinaryField.java
+++ b/solr/core/src/test/org/apache/solr/schema/TestBinaryField.java
@@ -17,13 +17,8 @@
 package org.apache.solr.schema;
 
 import java.io.File;
-import java.io.OutputStreamWriter;
-import java.io.Writer;
 import java.nio.ByteBuffer;
-import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
 import java.util.List;
-import java.util.Properties;
 import org.apache.commons.io.FileUtils;
 import org.apache.solr.SolrJettyTestBase;
 import org.apache.solr.SolrTestCaseJ4;
@@ -65,14 +60,14 @@ public class TestBinaryField extends SolrJettyTestBase {
         new File(src_dir, "solrconfig.snippet.randomindexconfig.xml"),
         new File(confDir, "solrconfig.snippet.randomindexconfig.xml"));
 
-    try (Writer w =
-        new OutputStreamWriter(
-            Files.newOutputStream(collDir.toPath().resolve("core.properties")),
-            StandardCharsets.UTF_8)) {
-      Properties coreProps = new Properties();
-      coreProps.put("name", "collection1");
-      coreProps.store(w, "");
-    }
+    //    try (Writer w =
+    //        new OutputStreamWriter(
+    //            Files.newOutputStream(collDir.toPath().resolve("core.properties")),
+    //            StandardCharsets.UTF_8)) {
+    //      Properties coreProps = new Properties();
+    //      coreProps.put("name", "collection1");
+    //      coreProps.store(w, "");
+    //    }
 
     createAndStartJetty(homeDir.getAbsolutePath());
   }
--- a/solr/test-framework/src/java/org/apache/solr/SolrJettyTestBase.java
+++ b/solr/test-framework/src/java/org/apache/solr/SolrJettyTestBase.java
@@ -18,11 +18,8 @@ package org.apache.solr;
 
 import java.io.File;
 import java.io.IOException;
-import java.io.OutputStreamWriter;
 import java.io.UncheckedIOException;
-import java.io.Writer;
 import java.lang.invoke.MethodHandles;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Properties;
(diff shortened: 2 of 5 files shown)


solr/core/src/test/org/apache/solr/schema/TestBinaryField.java
```diff
@@ -65,14 +60,14 @@ public static void beforeTest() throws Exception {
...
 
-    try (Writer w =
-        new OutputStreamWriter(
-            Files.newOutputStream(collDir.toPath().resolve("core.properties")),
-            StandardCharsets.UTF_8)) {
-      Properties coreProps = new Properties();
-      coreProps.put("name", "collection1");
-      coreProps.store(w, "");
-    }
+    //    try (Writer w =
```
epugh, line 63: this is nice!

solr/test-framework/src/java/org/apache/solr/SolrJettyTestBase.java
```diff
@@ -106,39 +108,16 @@ public static JettySolrRunner createAndStartJetty(String solrHome) throws Except
...
-
-    Properties nodeProps = new Properties(nodeProperties);
-    nodeProps.setProperty("coreRootDirectory", coresDir.toString());
-    nodeProps.setProperty("configSetBaseDir", solrHome);
-
-    jetty = new JettySolrRunner(solrHome, nodeProps, jettyConfig);
-    jetty.start();
-    port = jetty.getLocalPort();
-    log.info("Jetty Assigned Port#{}", port);
+    solrClientTestRule.startSolr(Path.of(solrHome));
```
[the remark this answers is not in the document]
dsmiley: This is a persistent false-positive that I'm pursuing some resolution with Lift/Infer/RacerD.  Just ignore it.

solr/test-framework/src/java/org/apache/solr/SolrJettyTestBase.java
```diff
@@ -74,16 +73,23 @@ public static JettySolrRunner createAndStartJetty(
             .withServlets(extraServlets)
             .withSSLConfig(sslConfig.buildServerSSLConfig())
             .build();
-
+    ;
     Properties nodeProps = new Properties();
-    if (configFile != null) nodeProps.setProperty("solrconfig", configFile);
```
dsmiley, line 79: You removed setting "solrconfig" but kept "schema" on the line below.  Doesn't make sense to me; either both or neither.  I suspect you can remove both since these are supplied when we create the collection explicitly below.

solr/test-framework/src/java/org/apache/solr/SolrJettyTestBase.java
```diff
@@ -93,7 +99,11 @@ public static JettySolrRunner createAndStartJetty(
 
   public static JettySolrRunner createAndStartJetty(String solrHome, JettyConfig jettyConfig)
       throws Exception {
-    return createAndStartJetty(solrHome, new Properties(), jettyConfig);
+
```
dsmiley, line 102: however the former code led to code that did a bunch more stuff than I see here, like using "collection1" as the configSet, for example.

solr/test-framework/src/java/org/apache/solr/SolrJettyTestBase.java
```diff
@@ -106,39 +116,16 @@ public static JettySolrRunner createAndStartJetty(String solrHome) throws Except
   public static JettySolrRunner createAndStartJetty(
       String solrHome, Properties nodeProperties, JettyConfig jettyConfig) throws Exception {
 
-    initCore(null, null, solrHome);
-
-    Path coresDir = createTempDir().resolve("cores");
-
-    Properties props = new Properties();
-    props.setProperty("name", DEFAULT_TEST_CORENAME);
-    props.setProperty("configSet", "collection1");
```
dsmiley, line 115: this particular choice wasn't retained in your changes -- the use of "collection1" as the configSet when creating a collection.

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
```diff
@@ -0,0 +1,103 @@
...
+  public JettySolrRunner getJetty() {
+    return jetty;
+  }
+
+  @Override
+  public SolrClient getSolrClient(String name) {
+    try {
+      throw new Exception("Unimplemented");
+    } catch (Exception e) {
+      throw new RuntimeException(e);
```
dsmiley, line 95: LOL!
FYI go look at UnsupportedOperationException

dsmiley (commented):
I looked at the failing tests.
* TestZkConfigSetService fails despite it not subclassing the tests you modified.  This is because it calls `SolrJettyTestBase.legacyExampleCollection1SolrHome()` which you modified to not put a core.properties there, which makes sense for your work here but broke assumptions in that test.  I used a debugger to figure this out and I was very familiar with that test.  I think this utility method should be protected so as to prevent tests outside this class hierarchy from being tempted to call it.  Admittedly a number of places call it though, and it would increase the scope to do something about that.  Looking closer at this method, and the fact that "legacy" is in its name, I wonder if we could do away with it entirely -- come up with alternatives, using our TestRule or something else depending on the expectations of each test calling it.  Basically, use an empty Solr home and reference org.apache.solr.util.ExternalPaths#TECHPRODUCTS_CONFIGSET as the configSet when creating a collection.
* TestSQLHandlerNonCloud fails; seems like a simple test.  solr.test.sys.prop2 isn't set -- this is a disease in our test configs.  Simple short term fix -- define it.  You did this in one spot in SolrJettyTestBase but not in another spot where a collection is created.
* SchemaTest (a RestTestBase subclass): Looking closer, this test failed because "dataDir1" wasn't defined -- it was referenced in a core descriptor (aka a core.properties file).  It turns out, this was in solrj/solr/shared/.... etc. which this test shouldn't even be using in the first place.  This suggests a blank (tempDir) coreRootDirectory was not used.  It also shows that SchemaTest.init() is copying way too much stuff.  A better fix would be to use a tempDir Solr home with nothing there and then reference the configSet we want when creating new collections.

dsmiley:
Replaced: https://github.com/apache/solr/pull/1374

PR #1342, opened 2023-02-08
SOLR-16650, SOLR-16532: OTEL tracer additional tags

janhoy:
https://issues.apache.org/jira/browse/SOLR-16650
https://issues.apache.org/jira/browse/SOLR-16532

Testing solr 9.2.0-SNAPSHOT at a client and missing hostname/port from traces. So this PR will add those by default, following the [otel semantic conventions](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/), both local logical host name as seen on the request, and peer host/port. Not yet tested in the wild.

### DIFF
--- a/solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
+++ b/solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
@@ -311,10 +311,7 @@ public abstract class ServletUtils {
             .withTag(Tags.SPAN_KIND, Tags.SPAN_KIND_SERVER)
             .withTag(Tags.HTTP_METHOD, request.getMethod())
             .withTag(Tags.HTTP_URL, request.getRequestURL().toString())
-            .withTag("net.host.name", request.getServerName())
-            .withTag("net.host.port", request.getServerPort())
-            .withTag("net.peer.name", request.getRemoteHost())
-            .withTag("net.peer.port", request.getRemotePort());
+            .withTag(Tags.PEER_HOSTNAME, request.getServerName());
     if (request.getQueryString() != null) {
       spanBuilder.withTag("http.params", request.getQueryString());
     }
--- a/solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
+++ b/solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
@@ -314,7 +314,9 @@ public abstract class ServletUtils {
     if (request.getQueryString() != null) {
       spanBuilder.withTag("http.params", request.getQueryString());
     }
-    spanBuilder.withTag(Tags.DB_TYPE, "solr");
+    spanBuilder.withTag(Tags.DB_TYPE, "solr")
+        .withTag("net.host.name", request.getServerName())
+        .withTag("net.host.port", request.getServerPort());
     return spanBuilder.start();
   }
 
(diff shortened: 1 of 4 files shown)


janhoy:
Not adding CHANGES line since none of this is yet released.

solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
```diff
@@ -310,7 +310,11 @@ protected static Span buildSpan(Tracer tracer, HttpServletRequest request) {
             .asChildOf(tracer.extract(Format.Builtin.HTTP_HEADERS, new HttpServletCarrier(request)))
             .withTag(Tags.SPAN_KIND, Tags.SPAN_KIND_SERVER)
             .withTag(Tags.HTTP_METHOD, request.getMethod())
-            .withTag(Tags.HTTP_URL, request.getRequestURL().toString());
+            .withTag(Tags.HTTP_URL, request.getRequestURL().toString())
+            .withTag("net.host.name", request.getServerName())
+            .withTag("net.host.port", request.getServerPort())
+            .withTag("net.peer.name", request.getRemoteHost())
+            .withTag("net.peer.port", request.getRemotePort());
```
risdenk, line 317: Is https://www.javadoc.io/static/io.opentracing/opentracing-api/0.31.0/io/opentracing/tag/Tags.html#PEER_HOSTNAME the right constant to use?

risdenk: Hmmm missing the `net.` it looks like... https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/tag/Tags.java#L88

dsmiley:
> Not adding CHANGES line since none of this is yet released.

But this change isn't restricted to OTEL; any tracer plugin would get these new tags.  Any way, it's fine with me.

solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
janhoy: I'm no expert here, but the opentelemetry semantic conventions seem to have evolved and differ from what OpenTracing used. I don't know the history here, but OT is now part of OTEL, so I interpreted that `peer.hostname` is the old OT tag name while `net.peer.name` is the new OTEL name. Thus there was no constant available. 

The `opentelemetry-semconv` dependency which we currently do not pull in, contains such constants in [SemanticAttributes.java](https://github.com/open-telemetry/opentelemetry-java/blob/main/semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/SemanticAttributes.java). A bit overkill to pull that in for four string constants though.

The question then is - since this is shared code between Jaeger and OTEL, and they both expect different tag names (the others, SPAN_KIND, HTTP_METHOD, HTTP_URL are the same), if we should tag both variants, or find some way to choose tags based on active module. Trick is that in `ServletUtils` class we only have a tracer instance and don't know the active module. Only way I can think of is to do `tracer instanceof io.opentelemetry.opentracingshim.TracerShim`.

janhoy:
> > Not adding CHANGES line since none of this is yet released.
> 
> But this change isn't restricted to OTEL; any tracer plugin would get these new tags. Any way, it's fine with me.

Yes, since this is on the OT Tracer level, that is true, so perhaps it deserves a line.

solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
janhoy: Hmm, I wonder if I'll skip these four tags in this PR, and just do the refGuide change to keep it contained.

What I initially planned was to add the solr host name in the trace, not the http request host name (which could be an alias, and is also part of the URL). It could be possible to add the `host.name` tag in a static way as part of the Otel configurator and avoid touching this common part for now.

Once we replace OT instrumentation with OTEL instrumentation, we can revisit what tags to set by default.

janhoy:
Going to merge this on Friday if no more comments from @risdenk , @dsmiley or @gus-asf

PR #1344, opened 2023-02-09
Tracking Field usage with DirectoryReader

risdenk:


solr/core/src/java/org/apache/solr/core/TrackingDirectoryReader.java
```diff
@@ -0,0 +1,158 @@
...
+    return wrap(in);
+  }
+
+  @Override
+  public CacheHelper getReaderCacheHelper() {
+    return in.getReaderCacheHelper();
+  }
+
+  private static void trackUsage(String field, String type) {
+    String key = field + '_' + type;
```
mkhludnev, line 65: won't it blow up heap? Couldn't be better to use a kind of tuple as a key, or just String.intern() ?

risdenk: Not sure its going to matter? very short lived object creation that will be immediately gc'd for the key? Not sure how a tuple would be better?

risdenk: I pushed e3e2364 which uses `AbstractMap.SimpleImmutableEntry` but the heap usage here shouldn't be in `trackUsage`

its going to be in the fieldUsage map if there are tons and tons of fields. each field would be a long entry plus the field name.

solr/core/src/java/org/apache/solr/core/TrackingDirectoryReader.java
```diff
@@ -0,0 +1,160 @@
...
+import org.apache.lucene.index.SortedSetDocValues;
+import org.apache.lucene.index.Terms;
+import org.apache.lucene.index.VectorValues;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TrackingDirectoryReader extends FilterDirectoryReader {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  private static final Map<Map.Entry<String, String>, LongAdder> fieldUsage =
```
mkhludnev, line 44: Pardon. Why it's static? Won't entries clash between TDReader instances?

risdenk: yea fair I have only been testing this on a Solr cluster that has one core per Solr node so hasn't been an issue. Will address.

risdenk: I think this is fixed in 91db3e9 - I need to add some tests. Thanks for taking a look so far!

PR #1348, opened 2023-02-10
Adding support for compression within the ZkCLI tool

justinrsweeney:
https://issues.apache.org/jira/browse/SOLR-16608

# Description

This provides support for compressed state.json for the ZkCLI tool. This will allow for using the ZkCLI tool to get and put state.json files and ensure they are correctly compressed when put and uncompressed when retrieved. This will allow Solr administrators to retrieve and modify state.json while maintaining compression.

# Solution

This uses the local solr home by default to get the configured settings from solr.xml for compressed state. If this tool is run without providing solr home, then you will not be able to rely on compression being configured.

# Tests

Added new tests to the ZkCLITest to cover cases with compression as well.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/ZkCLI.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ZkCLI.java
@@ -25,7 +25,9 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintStream;
+import java.lang.reflect.InvocationTargetException;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
 import java.util.Properties;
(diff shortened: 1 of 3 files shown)


epugh:
This looks like it does the job.  However, I was sort of thinking that compression would be more of a wrapper around the underlying data structure?   Kind of like a `ZipInputStream` might wrap a `InputStream`?    It seems like there is a lot of logic in the CLI around all the compression, and that it might be ripe for bugs...    

Maybe thinking it would look more like `ZKReader reader = getZKReader()` and you woudn't see all the compression logic?

Someday I want to break up ZKCli.java and make it easier to reason about.

epugh:
Oh, and thanks for giving the CLI some love!

justinrsweeney:
> This looks like it does the job. However, I was sort of thinking that compression would be more of a wrapper around the underlying data structure? Kind of like a `ZipInputStream` might wrap a `InputStream`? It seems like there is a lot of logic in the CLI around all the compression, and that it might be ripe for bugs...
> 
> Maybe thinking it would look more like `ZKReader reader = getZKReader()` and you woudn't see all the compression logic?
> 
> Someday I want to break up ZKCli.java and make it easier to reason about.

I don't think that'll be doable here. In this case, reading is actually handled fine. Putting files into ZK is where there is a bit of complexity. I did refactor the check if data should be compressed into its own method now to make it a little cleaner.

I do agree that a larger refactor of this code would be beneficial in the future.

PR #1350, opened 2023-02-10
Shard split when NRT and pull replica & prs enabled leads to inactive replica

hiteshk25:


solr/core/src/test/org/apache/solr/cloud/SplitShardWithNodeRoleTest.java
```diff
@@ -0,0 +1,91 @@
...
+    JettySolrRunner overseer2 = null;
+    System.setProperty(NodeRoles.NODE_ROLES_PROP, "data:off,overseer:preferred");
+    try {
+      overseer1 = cluster.startJettySolrRunner();
+      overseer2 = cluster.startJettySolrRunner();
+    } finally {
+      System.clearProperty(NodeRoles.NODE_ROLES_PROP);
+    }
+
+    Thread.sleep(10000);
```
noblepaul, line 36: We can't have a test waiting for 10 seconds

solr/core/src/java/org/apache/solr/cloud/overseer/ReplicaMutator.java
```diff
@@ -438,9 +438,6 @@ private ZkWriteCommand updateState(
 
     DocCollection newCollection = CollectionMutator.updateSlice(collectionName, collection, slice);
     log.debug("Collection is now: {}", newCollection);
-    if (collection.isPerReplicaState() && oldReplica != null) {
```
noblepaul, line 441: So, essentially, we always end up modifying state.json. 

Which property is being updated here for the replica?

solr/core/src/test/org/apache/solr/cloud/SplitShardWithNodeRoleTest.java
hiteshk25: Right, My attempt to listen zk node `overseer_elect` was too complex as child `leader` was deleted on each update. 

if there is any pattern let me know?

solr/core/src/java/org/apache/solr/cloud/overseer/ReplicaMutator.java
hiteshk25: The fix here is more about fetching(or triggering) the prs state in overseer node. This definitely needs more thoughts(when state.json is updated vs when prs updated vs how solrcore handles prs collection) to sync the prs state.

noblepaul:
There are still occasional test failures. I'm investigating

```./gradlew :solr:core:test --tests "org.apache.solr.cloud.SplitShardWithNodeRoleTest.testSolrClusterWithNodeRoleWithHA" -Ptests.jvms=5 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=606BE1C267080F99 -Ptests.file.encoding=US-ASCII```

chatman:
FYI, @hiteshk25 @noblepaul 

https://issues.apache.org/jira/browse/SOLR-16653?focusedCommentId=17693396&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17693396

PR #1353, opened 2023-02-12
SOLR-16563 DefaultCollectionSolrClient

dsmiley:
A draft of an idea expressed in
https://issues.apache.org/jira/browse/SOLR-16563 and earlier in https://issues.apache.org/jira/browse/SOLR-12502

### DIFF
--- a/solr/core/src/test/org/apache/solr/cloud/AliasIntegrationTest.java
+++ b/solr/core/src/test/org/apache/solr/cloud/AliasIntegrationTest.java
@@ -33,6 +33,7 @@ import org.apache.http.entity.StringEntity;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.util.EntityUtils;
 import org.apache.lucene.util.IOUtils;
+import org.apache.solr.client.solrj.DefaultCollectionSolrClient;
 import org.apache.solr.client.solrj.SolrClient;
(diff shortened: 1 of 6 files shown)


dsmiley:
The initial impl was more of a "default" and hence the name DefaultCollectionSolrClient.  But I felt it shouldn't be merely a default, it should be more of a chroot idea; an intentional limitation / scope.  Just `CollectionSolrClient` would be better.

epugh:
@dsmiley I tried using the `DefaultCollectionSolrClient` with out much success.  Everywhere wanted a CloudSolrClient...   Or, I didn't have setters in the testing infrastructure to take in a `DefaultCollectionSolrClient`....    So not sure here.

I am now wondering if having `setDefaultCollection` is that bad?   Maybe it's okay to have it there.  Or, rip the band-aid off and force the use of collectionName everywhere?   At least, in our test infrastrutture?

epugh:
I don't think right now I am going to push forward with this approach....   While it may have value in general, it didn't make it easier for me to migrate the tests away from changing the collection that the client was pointed at.  Instead, I am working on updating the tests to not require/need the use of a `solrClient.setDefaultCollection()` method...

dsmiley:
I'm seeing that this very DefaultCollectionSolrClient has value for users as an independent utility someone can opt-in to use.  I'm inclined to reposition it as that.

PR #1356, opened 2023-02-13
SOLR-16638 Fix Http2SolrClient's exception message when serverBaseUrl is null

stillalex:
https://issues.apache.org/jira/browse/SOLR-16638

# Description

Fixing Http2SolrClient's exception message when serverBaseUrl is null

# Solution

Instead of using the nullable serverBaseUrl, I am using the request data, this should reflect the actual request that caused the exception in the message.

Before change:
```
org.apache.solr.client.solrj.impl.BaseHttpSolrClient$RemoteSolrException: Error from server at null: Expected mime type in [application/octet-stream, application/vnd.apache.solr.javabin] but got text/html.
```

After change
```
org.apache.solr.client.solrj.impl.BaseHttpSolrClient$RemoteSolrException: Error from server at http://127.0.0.1:57676/solr/debug/foo/admin/ping?wt=javabin&version=2: Expected mime type in [application/octet-stream, application/vnd.apache.solr.javabin] but got text/html.
```

This helps when the baseUrl is null by using the real url involved in the request.

_Open Question_: the more intrusive part of the patch is that it changes the exception message in all cases, even when the base url is not null, so all the exceptions will get the _complete url with params_ instead of _base url_ only. if this is something that does not work, I am happy to poke at this some more.

# Tests

# Checklist

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
@@ -140,7 +140,7 @@ public class Http2SolrClient extends SolrClient {
   private List<HttpListenerFactory> listenerFactory = new ArrayList<>();
   private AsyncTracker asyncTracker = new AsyncTracker();
   /** The URL of the Solr server. */
-  private String serverBaseUrl;
+  private final String serverBaseUrl;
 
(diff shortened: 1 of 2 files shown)


risdenk (approved):

stillalex:
thanks @risdenk!

PR #1359, opened 2023-02-15
SOLR-16658 List of permissions returned to Admin UI is not complete

janhoy:
https://issues.apache.org/jira/browse/SOLR-16658

It's a trappy design to allow "null" keys in maps like this. And to define a "null" role in a permission in Admin UI security editor you need to actively de-select all roles, which took me a while to figure. Not trying to fix this here, but at least trying to resolve correct set of permissions for a given user, taking into account that permissions may have `*` or `null` roles.

I discovered this bug when working with a quite open-ended security.json for JWT, where some endpoints don't require authentication and all others do require auth, but we don't care what role the user has, e.g.

```json
"permissions":[
  {
    "name": "info",
    "role": null,
    "collection": null,
    "path": "/admin/info/system"
  },
  {
    "name": "health",
    "role": null
  },
  {
    "name": "all",
    "role": "*"
  }
]
```

With a config like the one above, a user logging in to Admin UI would not be able to create a collection or view schema editor or security editor, even if they basically has all the permissions in the world. After this PR, any user with role `foo` would get a minimum of `["info", "health", "all"]` in return. Of course in this example, with the presence of the `all` permission, any authenticated user could return just `["all"]` since the other permissions are implied, but that is an optimization that is unrelated.

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/admin/SystemInfoHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/SystemInfoHandler.java
@@ -34,6 +34,8 @@ import java.util.Date;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.apache.lucene.util.Version;
 import org.apache.solr.api.AnnotatedApi;
 import org.apache.solr.api.Api;
@@ -339,7 +341,10 @@ public class SystemInfoHandler extends RequestHandlerBase {
         RuleBasedAuthorizationPluginBase rbap = (RuleBasedAuthorizationPluginBase) auth;
         Set<String> roles = rbap.getUserRoles(req.getUserPrincipal());
         info.add("roles", roles);
-        info.add("permissions", rbap.getPermissionNamesForRoles(roles));
+        info.add(
+            "permissions",
+            rbap.getPermissionNamesForRoles(
+                Stream.concat(roles.stream(), Stream.of("*", null)).collect(Collectors.toSet())));
       }
     }
 
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -204,6 +204,8 @@ Bug Fixes
 
 * SOLR-16639: Fix jq parse error of solr-exporter metrics node_thread_pool_completed_total (Naoto Minami)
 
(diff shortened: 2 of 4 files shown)


solr/core/src/java/org/apache/solr/handler/admin/SystemInfoHandler.java
```diff
@@ -339,7 +341,10 @@ public SimpleOrderedMap<Object> getSecurityInfo(SolrQueryRequest req) {
         RuleBasedAuthorizationPluginBase rbap = (RuleBasedAuthorizationPluginBase) auth;
         Set<String> roles = rbap.getUserRoles(req.getUserPrincipal());
         info.add("roles", roles);
-        info.add("permissions", rbap.getPermissionNamesForRoles(roles));
+        info.add(
+            "permissions",
+            rbap.getPermissionNamesForRoles(
+                Stream.concat(roles.stream(), Stream.of("*", null)).collect(Collectors.toSet())));
```
janhoy, line 347: I considered letting `getPermissionNamesForRoles` always consider `*` and `null` roles, since this is the only place it is used, but the name of the method suggests that you want permissions for exactly those roles, so I expanded on the caller side instead...

risdenk (approved):

epugh (approved):
LGTM

solr/core/src/java/org/apache/solr/handler/admin/SystemInfoHandler.java
tflobbe: I'm hitting a NPE with Solr 9.2 in this line. I believe the issue is that `roles` can actually be null, if the user has no roles defined. Unfortunately, this is the case for the `$` user, that we use in `PKIAuthenticationPlugin`, this this causes the `/solr/admin/info/system` API not to work when using `nodes` (which the UI uses to populate the Nodes tab).

janhoy: https://issues.apache.org/jira/browse/SOLR-16729

tflobbe: Ugh, sorry, I created SOLR-16730

janhoy:
See https://issues.apache.org/jira/browse/SOLR-16730

PR #1360, opened 2023-02-15
SOLR 16642 : upgrade Solr to use Lucene 9.5.0

vinayakphegde:
https://issues.apache.org/jira/browse/SOLR-16642
draft PR


solr/core/src/java/org/apache/solr/handler/component/TermVectorComponent.java
```diff
@@ -277,15 +277,15 @@ public void process(ResponseBuilder rb) throws IOException {
       if (null != fields) {
         for (Map.Entry<String, FieldOptions> entry : fieldOptions.entrySet()) {
           final String field = entry.getKey();
-          final Terms vector = reader.getTermVector(docId, field);
+          final Terms vector = reader.termVectors().get(docId).terms(field);
```
risdenk, line 280: I think `reader.termVectors()` can be pulled out of the while loop? near line 253

solr/core/src/java/org/apache/solr/handler/component/TermVectorComponent.java
```diff
@@ -277,15 +277,15 @@ public void process(ResponseBuilder rb) throws IOException {
...
           if (vector != null) {
             TermsEnum termsEnum = vector.iterator();
             mapOneVector(docNL, entry.getValue(), reader, docId, termsEnum, field);
           }
         }
       } else {
         // extract all fields
-        final Fields vectors = reader.getTermVectors(docId);
+        final Fields vectors = reader.termVectors().get(docId);
```
risdenk, line 288: I think reader.termVectors() can be pulled out of the while loop? near line 253

solr/core/src/java/org/apache/solr/handler/component/ResponseLogComponent.java
```diff
@@ -92,7 +92,8 @@ protected void processIds(
     Set<String> fields = Collections.singleton(schema.getUniqueKeyField().getName());
     for (DocIterator iter = dl.iterator(); iter.hasNext(); ) {
 
-      sb.append(schema.printableUniqueKey(searcher.doc(iter.nextDoc(), fields))).append(',');
+      sb.append(schema.printableUniqueKey(searcher.storedFields().document(iter.nextDoc(), fields)))
```
risdenk, line 95: I think `searcher.storedFields()` can be pulled out of the for loop?

solr/core/src/java/org/apache/solr/handler/component/ResponseLogComponent.java
```diff
@@ -106,7 +107,7 @@ protected void processScores(
     StringBuilder sb = new StringBuilder();
     Set<String> fields = Collections.singleton(schema.getUniqueKeyField().getName());
     for (DocIterator iter = dl.iterator(); iter.hasNext(); ) {
-      sb.append(schema.printableUniqueKey(searcher.doc(iter.nextDoc(), fields)))
+      sb.append(schema.printableUniqueKey(searcher.storedFields().document(iter.nextDoc(), fields)))
```
risdenk, line 110: I think `searcher.storedFields()` can be pulled out of the for loop? probably reused with the one from line 95?

solr/core/src/java/org/apache/solr/handler/component/MoreLikeThisComponent.java
```diff
@@ -390,7 +390,7 @@ NamedList<DocList> getMoreLikeThese(
 
       DocListAndSet similarDocuments =
           mltHelper.getMoreLikeThis(id, 0, rows, null, interestingTerms, flags);
-      String name = schema.printableUniqueKey(searcher.doc(id));
+      String name = schema.printableUniqueKey(searcher.storedFields().document(id));
```
risdenk, line 393: `searcher.storedFields()` can be pulled out of the while loop?

solr/core/src/java/org/apache/solr/handler/admin/IndexSizeEstimator.java
```diff
@@ -388,7 +388,7 @@ private void estimateTermVectors(Map<String, Object> result) throws IOException
         if (liveDocs != null && !liveDocs.get(docId)) {
           continue;
         }
-        Fields termVectors = leafReader.getTermVectors(docId);
```
risdenk, line 391: `leafReader.termVectors()` should be pulled out of the docId for loop

solr/core/src/java/org/apache/solr/handler/admin/IndexSizeEstimator.java
```diff
@@ -612,7 +612,7 @@ private void estimateStoredFields(Map<String, Object> result) throws IOException
           if (liveDocs != null && !liveDocs.get(docId)) {
             continue;
           }
-          leafReader.document(docId, visitor);
+          leafReader.storedFields().document(docId, visitor);
```
risdenk, line 615: `leafReader.storedFields()` should be pulled out of the docId for loop

solr/core/src/java/org/apache/solr/handler/MoreLikeThisHandler.java
```diff
@@ -472,7 +472,7 @@ public NamedList<BooleanQuery> getMoreLikeTheseQuery(DocList docs) throws IOExce
       DocIterator iterator = docs.iterator();
       while (iterator.hasNext()) {
         int id = iterator.nextDoc();
-        String uniqueId = schema.printableUniqueKey(reader.document(id));
+        String uniqueId = schema.printableUniqueKey(reader.storedFields().document(id));
```
risdenk, line 475: `reader.storedFields()` should be pulled out of the while loop.

solr/core/src/java/org/apache/solr/core/QuerySenderListener.java
```diff
@@ -86,7 +86,7 @@ public void close() {}
             if (o instanceof DocList) {
               DocList docs = (DocList) o;
               for (DocIterator iter = docs.iterator(); iter.hasNext(); ) {
-                newSearcher.doc(iter.nextDoc());
+                newSearcher.storedFields().document(iter.nextDoc());
```
risdenk, line 89: `newSearcher.storedFields()` should be pulled out of the for loop.

solr/core/src/java/org/apache/solr/update/processor/ClassificationUpdateProcessor.java
```diff
@@ -60,7 +60,8 @@ public ClassificationUpdateProcessor(
       ClassificationUpdateProcessorParams classificationParams,
       UpdateRequestProcessor next,
       IndexReader indexReader,
-      IndexSchema schema) {
+      IndexSchema schema)
+      throws IOException {
```
risdenk, line 64: I see a bunch of these new `IOException` throws are they required?

solr/core/src/java/org/apache/solr/util/SolrPluginUtils.java
```diff
@@ -252,7 +252,7 @@ public static void optimizePreFetchDocs(
       // get documents
       DocIterator iter = docs.iterator();
       for (int i = 0; i < docs.size(); i++) {
-        searcher.doc(iter.nextDoc(), fieldFilter);
+        searcher.storedFields().document(iter.nextDoc(), fieldFilter);
```
risdenk, line 255: `searcher.storedFields()` should be pulled out of the for loop.

solr/core/src/java/org/apache/solr/util/SolrPluginUtils.java
```diff
@@ -409,7 +409,7 @@ public static NamedList<Explanation> getExplanations(
     for (int i = 0; i < docs.size(); i++) {
       int id = iterator.nextDoc();
 
-      Document doc = searcher.doc(id);
+      Document doc = searcher.storedFields().document(id);
```
risdenk, line 412: `searcher.storedFields()` should be pulled out of the for loop.

solr/core/src/test/org/apache/solr/handler/tagger/TaggerTestCase.java
```diff
@@ -144,7 +144,7 @@ protected TestTag[] pullTagsFromResponse(SolrQueryRequest req, SolrQueryResponse
     DocIterator iter = docList.iterator();
     while (iter.hasNext()) {
       int docId = iter.next();
-      Document doc = searcher.doc(docId);
+      Document doc = searcher.storedFields().document(docId);
```
risdenk, line 147: `searcher.storedFields()` should be pulled out of the for loop.

solr/core/src/test/org/apache/solr/search/function/TestOrdValues.java
```diff
@@ -96,7 +96,7 @@ private void doTestRank(String field, boolean inOrder) throws Exception {
             : "IC"; // smaller than all ids of docs in this test ("ID0001", etc.)
 
     for (int i = 0; i < h.length; i++) {
-      String resID = s.doc(h[i].doc).get(ID_FIELD);
+      String resID = s.storedFields().document(h[i].doc).get(ID_FIELD);
```
risdenk, line 99: `s.storedFields()` should be pulled out of the for loop.

solr/core/src/test/org/apache/solr/search/function/TestOrdValues.java
```diff
@@ -139,7 +139,7 @@ private void doTestExactScore(String field, boolean inOrder) throws Exception {
     ScoreDoc sd[] = td.scoreDocs;
     for (int i = 0; i < sd.length; i++) {
       float score = sd[i].score;
-      String id = s.getIndexReader().document(sd[i].doc).get(ID_FIELD);
+      String id = s.getIndexReader().storedFields().document(sd[i].doc).get(ID_FIELD);
```
risdenk, line 142: `s.getIndexReader().storedFields()` should be pulled out of the for loop.

solr/core/src/test/org/apache/solr/uninverting/TestFieldCacheVsDocValues.java
```diff
@@ -175,7 +175,7 @@ public void testHugeBinaryValues() throws Exception {
 
           BinaryDocValues s = FieldCache.DEFAULT.getTerms(ar, "field");
           for (int docID = 0; docID < docBytes.size(); docID++) {
-            Document doc = ar.document(docID);
+            Document doc = ar.storedFields().document(docID);
```
risdenk, line 178: `ar.storedFields()` should be pulled out of the for loop

risdenk (changes requested):
Most of my comments are about pulling the same call outside the for loop - as laid out here https://github.com/apache/lucene/blob/main/lucene/MIGRATE.md#removed-deprecated-indexsearcherdoc-indexreaderdocument-indexreadergettermvectors-github11998

magibney (changes requested):
So, I'm actually super-excited about the new storedFields() etc. APIs, but I suspect we may not want to dive right into them in this PR.

It's a big change, and I think we'd be better off changing as little as possible for the upgrade. (I agree with @risdenk's comments about pulling the new API abstractions out of loops, but I think that's best approached separately). Sticking with the existing (Lucene deprecated) API for stored fields and fixing any latent illegal usage would be my first choice. 

Basically I'm suggesting if possible to revert f1bbaa6f1a9d5d50a4ea2afd544dbd7e1df38485 and take a different approach to fixing the new assertion errors that crop up.

Curious what others think about this suggestion?

vinayakphegde:
@magibney you are right. I will revert that change.
at this moment I am not sure what causes these test failures
`Thread[searcherExecutor-320-thread-1,5,TGRP-TestJoin] and consumed in Thread[TEST-TestJoin.testRandomJoin-seed#[CA39E6C650AD9BC9],5,TGRP-TestJoin]`
but I have seen a lot of test cases failing with this error

magibney:
Sounds good! Definitely happy to take a look at the test failures, that's part of why I was thinking that a draft PR would be helpful (thanks for opening it btw!).

vinayakphegde:
Thanks @magibney

vinayakphegde:
that's nice!
let me also look into it

vinayakphegde:
> Sharing early here (not immediately sure about the _right_ way to fix this); it looks like there's a problem with `Terms` caching in `SlowCompositeReaderWrapper`; tests pass (aside from the test that explicitly tests for `Terms` caching in `SlowCompositeReaderWrapper` 🙂) with:
> 
> ```diff
> diff --git a/solr/core/src/java/org/apache/solr/index/SlowCompositeReaderWrapper.java b/solr/core/src/java/org/apache/solr/index/SlowCompositeReaderWrapper.java
> index f638fbd664f..5765b1e6d71 100644
> --- a/solr/core/src/java/org/apache/solr/index/SlowCompositeReaderWrapper.java
> +++ b/solr/core/src/java/org/apache/solr/index/SlowCompositeReaderWrapper.java
> @@ -140,6 +140,9 @@ public final class SlowCompositeReaderWrapper extends LeafReader {
>    public Terms terms(String field) throws IOException {
>      ensureOpen();
>      try {
> +      if (true) {
> +        return MultiTerms.getTerms(in, field);
> +      }
>        return cachedTerms.computeIfAbsent(
>            field,
>            f -> {
> ```
> 
> I also happened to hit an unrelated reproducible failure:
> 
> ```
> $ ./gradlew :solr:core:test --tests "org.apache.solr.search.TestComplexPhraseQParserPlugin.testPhraseHighlighter" -Ptests.jvms=5 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=56C83FBC07AAE2C3 -Ptests.file.encoding=ISO-8859-1
> ```

yeah, this change solved the problem, But I just want to know
what was the problem here and how this change(not involving the map) solved the problem?

magibney:
The map is explicitly designed to lazily cache these `Terms` instances (accumulating them over time for a given `SolrIndexSearcher`). The returned `Terms` instances are _not_ safe for concurrent use -- considering which, I'm surprised we didn't see actual errors resulting from this usage!

I suspect the correct solution will be to just skip caching the `Terms` instances altogether. We could also upgrade from Map to use a proper cache (with evictions, etc.) or some other caching strategy that effectively incorporates "creation thread" into the cache key. Not sure whether that would be worth it.

It's also possible that the _way_ in which `Terms` instances returned from `SlowCompositeReaderWrapper` are used in practice actually _is_ thread-safe? In that case there might be other options, such as pulling the required thread-safe derivative of the `Terms` instance, and wrapping it in a synthetic `Terms` instance that mostly throws `UnsupportedOperationException`, and caching/returning that synthetic instance?

Curious what others think here ... perhaps @dsmiley?

vinayakphegde:
even with the new change also we are returning the same `Term` instance, right?
How does that make the `Term` instance thread-safe?

vinayakphegde:
Okay, got it.
yeah, if you find any solution, ping me. I will try that out and test it in my system.

solr/core/src/java/org/apache/solr/update/processor/UpdateRequestProcessorChain.java
```diff
@@ -16,6 +16,7 @@
  */
 package org.apache.solr.update.processor;
 
+import java.io.IOException;
```
risdenk, line 19: I don't think `IOException` is needed in this file.

solr/core/src/java/org/apache/solr/update/processor/UpdateRequestProcessorFactory.java
```diff
@@ -16,6 +16,7 @@
  */
 package org.apache.solr.update.processor;
 
+import java.io.IOException;
```
risdenk, line 19: I don't think `IOException` is needed in this file.

solr/core/src/java/org/apache/solr/update/UpdateLog.java
```diff
@@ -1940,7 +1940,12 @@ public void doReplay(TransactionLog translog) {
         ThreadLocal<UpdateRequestProcessor> procThreadLocal =
             ThreadLocal.withInitial(
                 () -> {
-                  var proc = processorChain.createProcessor(req, rsp);
+                  UpdateRequestProcessor proc;
+                  try {
+                    proc = processorChain.createProcessor(req, rsp);
+                  } catch (IOException e) {
+                    throw new RuntimeException(e);
+                  }
```
risdenk, line 1948: After removing IOException from UpdateRequest files - this change isn't needed.

risdenk (changes requested):
I checked a few of the `IOExceptions` that were added and seems like a few are extraneous. I didn't check every case.

solr/core/src/java/org/apache/solr/update/UpdateLog.java
vinayakphegde: Hi @risdenk 
if we see this lucene commit https://github.com/apache/lucene/pull/11998/files , the `KNearestNeighborDocumentClassifier` constructor now throws IOException 
we are using that in `ClassificationUpdateProcessor`'s constructor 
so that's why I added `throws IOExceptio`n in the `ClassificationUpdateProcessor` constructor
and we are calling `ClassificationUpdateProcessor` from ClassificationUpdateProcessorFactory#getInstance() 
so I had to throw IOException from ClassificationUpdateProcessorFactory#getInstance() 
since it is inherited from UpdateRequestProcessorFactory, I had to add it there as well.
and so on for other related files like this.

risdenk: Yea lets not propagate `IOEXception` everywhere if we don't need to.

solr/core/src/java/org/apache/solr/index/SlowCompositeReaderWrapper.java
```diff
@@ -359,14 +373,26 @@ public PointValues getPointValues(String field) {
...
+  @Override
+  public TopDocs searchNearestVectors(
+      String field, float[] target, int k, Bits acceptDocs, int visitedLimit) {
+    return null;
   }
 
   @Override
   public TopDocs searchNearestVectors(
-      String field, float[] target, int k, Bits acceptDocs, int visitedLimit) throws IOException {
+      String field, byte[] target, int k, Bits acceptDocs, int visitedLimit) {
```
magibney, line 395: I think removing the `throws IOException` from the `float[]` variant of `searchNearestVectors()` is out-of-scope for this PR (the method it inherits throws IOException). Among other things, it makes the diff less clear wrt what's been added:

To clarify, this upgrade adds a `byte[]` variant of `searchNearestVectors()` and leaves the existing `float[]` variant as-is. True, because this class is final, there's probably no negative consequence to removing the `throws IOException` (from either variant, `byte[]` or `float[]`).

But my inclination would be to try to keep changes as tightly scoped as possible, and probably even follow existing precedent and add `throws IOException` to the new new `byte[]` variant.

Curious what others think? I definitely appreciate the clarity in the diff that results from restoring the `throws IOException` here.

solr/core/src/java/org/apache/solr/index/SlowCompositeReaderWrapper.java
```diff
@@ -328,6 +331,17 @@ public Fields getTermVectors(int docID) throws IOException {
     return in.getTermVectors(docID);
   }
 
+  @Override
+  public TermVectors termVectors() throws IOException {
+    return in.termVectors();
```
magibney, line 336: I know this follows precedent from the existing `termVectors()` method above, but I wonder why these don't call `ensureOpen()` the wall all the other "content-access" methods seem to do?

solr/core/src/java/org/apache/solr/update/processor/ClassificationUpdateProcessor.java
```diff
@@ -77,17 +77,21 @@ public ClassificationUpdateProcessor(
...
+                  classificationParams.getTrainingFilterQuery(),
+                  classificationParams.getK(),
+                  classificationParams.getMinDf(),
+                  classificationParams.getMinTf(),
+                  trainingClassField,
+                  field2analyzer,
+                  inputFieldNamesWithBoost);
+        } catch (IOException e) {
+          throw new RuntimeException(e);
+        }
```
magibney, line 94: Prefer to wrap this in SolrException with more information about context (as opposed to a generic RuntimeException).

magibney (changes requested):
I suspect it may be possible to remove more IOExceptions? Also, I have yet to review `OpenNLPExtractNamedEntitiesUpdateProcessorFactory`.

A handful of comments inline, but mostly looking pretty good so far I think!

solr/core/src/java/org/apache/solr/index/SlowCompositeReaderWrapper.java
vinayakphegde: Okay, Will add back the `throws IOException` to the `float[]` variant as well as the `byte[]` variant of `searchNearestVectors()`

solr/core/src/java/org/apache/solr/index/SlowCompositeReaderWrapper.java
vinayakphegde: Yeah, all the other functions seem to have `ensureOpen()` apart from `getTermVectors(int docID)` and `termVectors()` (of course I added this one).
Should I add for both functions or should I add only for `termVectors()` and will take care of `getTermVectors(int docID)` in some other PR?

solr/core/src/java/org/apache/solr/update/processor/ClassificationUpdateProcessor.java
vinayakphegde: okay.

solr/core/src/java/org/apache/solr/index/SlowCompositeReaderWrapper.java
magibney: `ensureOpen()` was removed as part of [this commit](https://github.com/apache/solr/commit/797bb38c50fed9e533ca20de920c63c6c35256a9#diff-a3934253a307f6bc36771c8d549e5b5bb7bad12fd5dbe4de06e7b5ef176e2083).

Looks like an oversight perhaps. I'd be inclined to put it back (for consistency with the other methods). @markrmiller any recollection why this was removed? Any objection to restoring it?

vinayakphegde: Hey @magibney, should we wait for his reply or should we add it back?

magibney: I'd recommend to go ahead and add it back. Thanks for following up!

solr/core/src/java/org/apache/solr/search/SolrQueryTimeoutImpl.java
```diff
@@ -45,7 +45,6 @@ public static Long getTimeoutAtNs() {
     return timeoutAt.get();
   }
 
-  @Override
```
dsmiley, line 48: This innocuous looking removal was very curious to me and I went digging into Lucene's changes that led to this.  Based on what I learned, I predict a performance degradation for any query that isn't using timeouts.  The Lucene issue is https://github.com/apache/lucene/issues/11914 and it means that if Solr does nothing about this, then ExitableDirectoryReader.wrap (used in SolrIndexSearcher) will wrap all underlying low level index components (Terms, ...) because Lucene's ExitableDirectoryReader no longer guards against this with a "isTimeoutEnabled" check as it no longer exists.  Solr could fork the old ExitableDirectoryReader, which would be sad but pretty easy.  Alternatively, maybe SolrIndexSearcher could only wrap for the execution of a query -- not sure as I haven't explored this.

vinayakphegde: Thanks, @dsmiley helping us here.
didn't understand fully, but will try again :)

uschindler (changes requested):

solr/core/src/java/org/apache/solr/core/MMapDirectoryFactory.java
```diff
@@ -63,7 +63,7 @@ protected Directory create(String path, LockFactory lockFactory, DirContext dirC
       throws IOException {
     MMapDirectory mapDirectory = new MMapDirectory(Path.of(path), lockFactory, maxChunk);
     try {
-      mapDirectory.setUseUnmap(unmapHack);
+      System.setProperty(MMapDirectory.ENABLE_UNMAP_HACK_SYSPROP, String.valueOf(unmapHack));
```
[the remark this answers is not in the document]
vinayakphegde: Okay, will remove the `unmapHack` references.

vinayakphegde:
tried to address all the issues/comments except "performance degradation for any query that isn't using timeouts" mentioned by @dsmiley 
now, only `org.apache.solr.search.TestComplexPhraseQParserPlugin > testPhraseHighlighter` test is failing
```
org.apache.solr.search.TestComplexPhraseQParserPlugin > testPhraseHighlighter FAILED
    java.lang.AssertionError: REQUEST FAILED: xpath=//lst[@name='1']/arr[@name='name']/str[.='<em>john</em> <em>smith</em> <em>smith</em> <em>john</em>']
        xml response was: <?xml version="1.0" encoding="UTF-8"?>
    <response>
    <lst name="responseHeader"><int name="status">0</int><int name="QTime">0</int></lst><result name="response" numFound="2" start="0" numFoundExact="true"><doc><str name="id">1</str></doc><doc><str name="id"
>2</str></doc></result><lst name="highlighting"><lst name="1"/><lst name="2"/></lst>
    </response>

        request was:q=name:"(john+johathon)+smith"&defType=complexphrase&qt=&hl=true&fl=id&hl.requireFieldMatch=false&hl.usePhraseHighlighter=false&hl.fragsize=0&start=0&hl.fl=name&rows=200
        at __randomizedtesting.SeedInfo.seed([3D5488D220BF1F0:B454EA75135D9B1]:0)
        at org.junit.Assert.fail(Assert.java:89)
        at org.apache.solr.SolrTestCaseJ4.assertQ(SolrTestCaseJ4.java:1025)
        at org.apache.solr.search.TestComplexPhraseQParserPlugin.testPhraseHighlighter(TestComplexPhraseQParserPlugin.java:216)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at com.carrotsearch.randomizedtesting.RandomizedRunner.invoke(RandomizedRunner.java:1758)
        at com.carrotsearch.randomizedtesting.RandomizedRunner$8.evaluate(RandomizedRunner.java:946)
        at com.carrotsearch.randomizedtesting.RandomizedRunner$9.evaluate(RandomizedRunner.java:982)
        at com.carrotsearch.randomizedtesting.RandomizedRunner$10.evaluate(RandomizedRunner.java:996)
```
looking into it now.

solr/core/src/java/org/apache/solr/core/MMapDirectoryFactory.java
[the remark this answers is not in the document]
uschindler: I'd still add a warning message.

The default was always "true". If somebody changed it from the default, I'd log a warning with  with the URL to the Lucene documentation as stated before.

> If we want to be nicer to users - then logging a warning is a valid option.

This would bring both together.

dsmiley:
TestComplexPhraseQParserPlugin:  I anticipated this failure.  https://github.com/apache/lucene/pull/11807
This weekend I'll see if I could fix it.  I'm going to look into overriding getHighlightQuery in ComplexPhraseQParserPlugin to rewrite the query.

solr/core/src/java/org/apache/solr/core/MMapDirectoryFactory.java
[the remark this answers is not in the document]
vinayakphegde: you mean something like this?
```
@Override 
  protected Directory create(String path, LockFactory lockFactory, DirContext dirContext)
      throws IOException {
    MMapDirectory mapDirectory = new MMapDirectory(Path.of(path), lockFactory, maxChunk);
    log.warn("to disable the unmap hack globally, see this Lucene documentation" +
            " https://lucene.apache.org/core/9_5_0/core/org/apache/lucene/store/MMapDirectory.html#ENABLE_UNMAP_HACK_SYSPROP");
    mapDirectory.setPreload(preload);
    return mapDirectory;
  }
```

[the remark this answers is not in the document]
uschindler: No not in create().
On init() just check for existence of directory property "unmap", if it exists show exactly this warning. Maybe add something "Per directory unmapping is no longer supported with Lucene 9.5. ...."
If user does not specify anything in the XML file, no warning is printed.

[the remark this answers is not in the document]
uschindler: Like `if (params.get("unmap") != null)` (this is a good check for existence in SolrParams.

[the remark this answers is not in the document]
uschindler: Do it like that:

```java
public class MMapDirectoryFactory extends StandardDirectoryFactory {
  boolean preload;
  private long maxChunk;

  @Override
  public void init(NamedList<?> args) {
    super.init(args);
    SolrParams params = args.toSolrParams();
    maxChunk = params.getLong("maxChunkSize", MMapDirectory.DEFAULT_MAX_CHUNK_SIZE);
    if (maxChunk <= 0) {
      throw new IllegalArgumentException("maxChunk must be greater than 0");
    }
    if (params.get("unmap") != null) {
      log.warn("It is no longer possible to configure unmapping of index files on DirectoryFactory level in solrconfig.xml.");
      log.warn("To disable unmapping, pass -Dorg.apache.lucene.store.MMapDirectory.enableUnmapHack=false on Solr's command line.");
    }
    preload = params.getBool("preload", false); // default turn-off
  }

  @Override
  protected Directory create(String path, LockFactory lockFactory, DirContext dirContext)
      throws IOException {
    MMapDirectory mapDirectory = new MMapDirectory(Path.of(path), lockFactory, maxChunk);
    mapDirectory.setPreload(preload);
    return mapDirectory;
  }
}
```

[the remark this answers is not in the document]
vinayakphegde: Sure, will do that. Thanks

dsmiley:
I'll be sad to see the Terms caching go in SlowCompositeReaderWrapper but maybe it has to be this way.  Although I added the test of this, removed here, the underlying caching pre-dated me (hiding in MultiFields) in Lucene and I don't recall when it was added.  I [asked in Lucene](https://github.com/apache/lucene/pull/11998#issuecomment-1454862916) about the change which brought about the new assertion which prompted reverting this caching.  I don't like the idea of adding more thread-locals as a work-around.  Instead, maybe Solr can avoid some uses of SlowCompositeReaderWrapper where terms is used and retain reasonable performance?  I'm thinking of JoinUtils which could change its algorithm to loop over the leaves instead of assuming one leaf.  For another issue; it's too complicated / exploratory to be in this PR.  There are uses elsewhere too like in Collapsing when top_fc is used.  BTW SlowCompositeReaderWrapper is named this on purpose so that people don't use it ;-)

Ideally we'd have some benchmark somewhere that uses join queries.  I know @gerlowskija and @joel-bernstein have worked on such as I've code-reviewed related things from them.  They are probably proprietary / not readily available but doesn't hurt to ping them :-).  After this change merges, we could broadcast to the users list that there is a change that maybe has a perf regression, inviting users to test it for themselves soon.  https://hub.docker.com/r/apache/solr-nightly/tags

rmuir:
Creating this MultiTerms doesn't really do anything. It doesn't do any IO or anything like that. It essentially makes a list of reader.terms() for each segment. It doesn't sort this list or anything.

I wouldn't suggest doing anything crazy to cache this thing, as it doesn't make sense to cache: there's no "hot" methods on Terms (nextDoc or nextTerm or anything) that you'd call a ton of times, and hopefully use of slow wrapper is minimized anyway.

dsmiley:
Made two commits.
I looked at the ComplexPhraseQParser changes and I don't think we should bend over backwards to support hl.usePhraseHighlighter=false, which is what the failing test failed to do.  Trying to support it could make performance worse or extra code/complexity.  It's less accurate and designed to be faster in some circumstances with older highlighters but I'd be pleased if this flag simply didn't exist at all.

dsmiley:
Thanks Rob for chiming in.

Made two commits.
I looked at the ComplexPhraseQParser test and I don't think we should bend over backwards to support `hl.usePhraseHighlighter=false`, which is what the failing test failed to do. Trying to support it could make performance worse or extra code/complexity. It's less accurate and designed to be faster in some circumstances with older highlighters but I'd be content if this flag simply didn't exist at all.  Also I noticed a commented out test that should work nowadays and it does (thanks to UnifiedHighlighter probably), so I enabled it.

vinayakphegde:
> Thanks Rob for chiming in.
> 
> Made two commits. I looked at the ComplexPhraseQParser test and I don't think we should bend over backwards to support `hl.usePhraseHighlighter=false`, which is what the failing test failed to do. Trying to support it could make performance worse or extra code/complexity. It's less accurate and designed to be faster in some circumstances with older highlighters but I'd be content if this flag simply didn't exist at all. Also I noticed a commented out test that should work nowadays and it does (thanks to UnifiedHighlighter probably), so I enabled it.

Thanks @dsmiley for helping

solr/core/src/java/org/apache/solr/search/SolrQueryTimeoutImpl.java
vinayakphegde: hi @dsmiley, can we address this issue in separate Jira or should we include it in this only?

dsmiley: Okay, let's.  It may be a release blocker; hopefully there are benchmarks that can show yes/no.  FYI @chatman

vinayakphegde: Hi @dsmiley, any update on this one?

vinayakphegde:
can I make it a regular PR now?

dsmiley:
No need to wait for me: https://github.com/apache/solr/pull/1450

dsmiley:
Lets get this updated and merged; ehh?

vinayakphegde:
> Lets get this updated and merged; ehh?

Sure, need anything from my side? I mean, at this moment?
I think I addressed all the previous comments!!

dsmiley (commented):
You forgot [Uwe's comment](https://issues.apache.org/jira/browse/SOLR-16642?focusedCommentId=17683095&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17683095) out reverting the `--enable-preview`.  I'll do this now.

I'll also throw UOE as I suggested in this review.

Can you comment on why org.apache.solr.update.processor.UpdateRequestProcessor#extractTypedNamedEntities needed to be updated?  Did the API change and you just updated it to comply with the change?  It's not apparently purely looking at the PR.

I'll add a CHANGES.txt as well and probably merge Monday.

uschindler (approved):

solr/core/src/java/org/apache/solr/index/SlowCompositeReaderWrapper.java
```diff
@@ -370,6 +380,12 @@ public TopDocs searchNearestVectors(
     return null;
   }
 
+  @Override
+  public TopDocs searchNearestVectors(
+      String field, byte[] target, int k, Bits acceptDocs, int visitedLimit) throws IOException {
+    return null;
```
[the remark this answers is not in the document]
vinayakphegde: there is an overloaded function with `float[] target`, which was returning null, So I just copied that. But UnsupportedOperationException makes more sense. Should I change both function?

vinayakphegde:
> You forgot [Uwe's comment](https://issues.apache.org/jira/browse/SOLR-16642?focusedCommentId=17683095&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17683095) out reverting the `--enable-preview`. I'll do this now.
> 
> I'll also throw UOE as I suggested in this review.

thanks @dsmiley for working on @uschindler comment
 
> Can you comment on why org.apache.solr.update.processor.UpdateRequestProcessor#extractTypedNamedEntities needed to be updated? Did the API change and you just updated it to comply with the change? It's not apparently purely looking at the PR.
> 
> I'll add a CHANGES.txt as well and probably merge Monday.

that is no longer needed, we are handling the exception in between. will change it.

risdenk (approved):

PR #1362, opened 2023-02-16
SOLR-15959 Deprecate loading of solr.xml from ZK

janhoy:
https://issues.apache.org/jira/browse/SOLR-15959

Deprecation warning in log, and WARNING in refguide.

Also fixed some other refguide mentions of the need for copying solr.xml to SOLR_HOME, which is no longer needed.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -282,6 +282,8 @@ Other Changes
 
 * SOLR-15955: Upgrade to Jetty 10.x (Mark Miller, Kevin Risden)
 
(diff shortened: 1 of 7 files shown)


epugh (approved):
LGTM.  I checked how I use solr.xml in Chorus, yep, we package it as part of the Dockerfile build, not using it from ZK.

risdenk (approved):

janhoy:
I intend to merge this on Wednesday next week (Feb 22nd) unless there are new concerns.

PR #1363, opened 2023-02-16
SOLR-16664: TestCoordinatorRole fails docs is null

risdenk:
https://issues.apache.org/jira/browse/SOLR-16664

I started to look at `TestCoordinatorRole` and besides just adding the assertion to make sure docs is not null there are other confusing things in this test.

### DIFF
--- a/solr/core/src/test/org/apache/solr/search/TestCoordinatorRole.java
+++ b/solr/core/src/test/org/apache/solr/search/TestCoordinatorRole.java
@@ -405,10 +405,8 @@ public class TestCoordinatorRole extends SolrCloudTestCase {
       } catch (SolrException ex) {
         // we know we're doing tricky things that might cause transient errors
         // TODO: all these query requests go to the QA node -- should QA propagate internal request
-        // errors
-        //  to the external client (and the external client retry?) or should QA attempt to failover
-        // transparently
-        //  in the event of an error?
+        // errors to the external client (and the external client retry?) or should QA attempt to
+        // failover transparently in the event of an error?
         if (i < 5) {
           log.info("swallowing transient error", ex);
         } else {
--- a/solr/core/src/test/org/apache/solr/search/TestCoordinatorRole.java
+++ b/solr/core/src/test/org/apache/solr/search/TestCoordinatorRole.java
@@ -176,8 +176,7 @@ public class TestCoordinatorRole extends SolrCloudTestCase {
             executor.submit(
                 () -> {
                   // we manipulate the jetty instances in a separate thread to more closely mimic
-                  // the behavior we'd
-                  // see irl.
+                  // the behavior we'd see irl.
                   try {
                     Thread.sleep(establishBaselineMs);
                     log.info("stopping NRT jetty ...");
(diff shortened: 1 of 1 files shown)


solr/core/src/test/org/apache/solr/search/TestCoordinatorRole.java
```diff
@@ -389,6 +390,7 @@ private String getHostCoreName(
     SolrQuery q = new SolrQuery("*:*");
     q.add("fl", "id,desc_s,_core_:[core]").add(OMIT_HEADER, TRUE);
     p.accept(q);
+    // TODO why is this building a string instead of using the proper SolrClient???
```
noblepaul, line 393: SolrClient will not send the request to COORDINATOR , so we can't  test the behaviour

solr/core/src/test/org/apache/solr/search/TestCoordinatorRole.java
```diff
@@ -296,6 +294,8 @@ public void testNRTRestart() throws Exception {
         assertEquals(nrtCore, hostCore);
         // allow any exceptions to propagate
         jettyManipulationFuture.get();
+
+        // TODO what is expected here? why are we just bailing? Can't the following code just be removed...
```
noblepaul, line 298: this got accidentally added while debugging. this line must go

solr/core/src/test/org/apache/solr/search/TestCoordinatorRole.java
```diff
@@ -338,6 +338,7 @@ public void testNRTRestart() throws Exception {
         start = new Date().getTime();
         try {
           do {
+            // TODO - What is expected here? equals not doing anything? hostCore=?
```
noblepaul, line 341: it should be this
```
      if( pullCore.equals(
                    getHostCoreName(
                        COLL,
                        qaJettyBase,
                        client,
                        p -> {
                          p.set(CommonParams.Q, "id:345");
                          p.add("shards.preference", "replica.type:NRT");
                        }))){
              done.set(true);
            }
```

solr/core/src/test/org/apache/solr/search/TestCoordinatorRole.java
risdenk: I don't understand what you mean. I pushed up ff1f547eada49b494f5943ff2bb0e95ef3f4a671 which does exactly what I said using SolrClient.

solr/core/src/test/org/apache/solr/search/TestCoordinatorRole.java
risdenk: Fixed in e409355e572f08491b199841f773166e0b4c7b3c I think.

solr/core/src/test/org/apache/solr/search/TestCoordinatorRole.java
risdenk: Removed in 869ee24dcfb1fbed253291d2f49945d186eaccda

solr/core/src/test/org/apache/solr/search/TestCoordinatorRole.java
noblepaul: if you use `SolrClient` how do you ensure that the request is actually going to coordinator node?

risdenk: Unless I misunderstand how this works `try (SolrClient solrClient = new Http2SolrClient.Builder(qaNode).build()) {` - this builds a SolrClient pointing to the qanode.

noblepaul: >this builds a SolrClient pointing to the qanode.

that is not what it does. `Http2SolrClient.Builder(qaNode).build()` builds a client with the `qaNode` as the "seed node". Thereafter, it queries the `qaNode` to get the collection topology and finds out where the request should be sent. Basically, it behaves the similar to the Zk based provider but using the REST APIs

HoustonPutman: Are you sure you don't mean `CloudHttp2SolrClient`? I don't think that the `Http2SolrClient` should know anything about the cloud topology.

noblepaul: Oh, my bad. You are right . I confused it with `CloudHttp2SolrClient`

noblepaul (approved):

PR #1366, opened 2023-02-17
SOLR-16668: Default to Java SSL for Http2SolrClient

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-16668

Also added a bit of cleanup in the code.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -221,6 +221,8 @@ Bug Fixes
 
 * SOLR-16670: Fix directory/file check in S3Repository (Houston Putman, Hakan Özler)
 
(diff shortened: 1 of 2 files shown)


risdenk (approved):

janhoy (approved):
This looks good. I must have overlooked the need for system-default truststore when doing that change. Will the warning logs re-appear with this change?

HoustonPutman:
I think that the warnings will be removed when https://github.com/apache/solr/pull/1367 is merged as well, since it changes the default of the `trustAll` parameter. So we can check after merging this PR and before merging https://github.com/apache/solr/pull/1367

PR #1367, opened 2023-02-17
SOLR-16669: Fix default checkPeerName in Http2SolrClient

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-16669

The primary issue is the `checkPeerNameStr == null` check. If the sysProp is not provided we want to default to `true`, not `false`.

Not back-compat, but actually follows the code and ref-guide documentation.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -223,6 +223,8 @@ Bug Fixes
 
 * SOLR-16668: Use default to Java SSL for Http2SolrClient when none is provided (Houston Putman)
 
+* SOLR-16669: Http2SolrClient now defaults checkPeerName to True, as the documentation specified (Houston Putman)
+
 Build
 ---------------------
 * Upgrade forbiddenapis to 3.4 (Uwe Schindler)
--- a/solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-9.adoc
+++ b/solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-9.adoc
@@ -76,6 +76,10 @@ This is an improvement to the binary release artifact, but Jetty does not allow 
 The `server/contexts/solr-jetty-context.xml` now explicitly removes these restrictions, allowing Solr to share these "server" jars which now live in `server/lib/ext`.
 * The "Transient Cores" feature is now deprecated.
 
+=== SSL Configuration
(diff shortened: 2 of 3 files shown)


risdenk (approved):

HoustonPutman:
Ok @janhoy , checking the logs message.

Main with #1366 included:

```
2023-02-22 09:28:49.232 WARN  (main) [] o.e.j.u.s.S.config Trusting all certificates configured for Client@1d901f20[provider=null,keyStore=null,trustStore=null]
2023-02-22 09:28:49.233 WARN  (main) [] o.e.j.u.s.S.config No Client EndPointIdentificationAlgorithm configured for Client@1d901f20[provider=null,keyStore=null,trustStore=null]
2023-02-22 09:28:49.339 WARN  (main) [] o.e.j.u.s.S.config Trusting all certificates configured for Client@760487aa[provider=null,keyStore=null,trustStore=null]
2023-02-22 09:28:49.339 WARN  (main) [] o.e.j.u.s.S.config No Client EndPointIdentificationAlgorithm configured for Client@760487aa[provider=null,keyStore=null,trustStore=null]
```

Then with this change:

```
2023-02-22 09:31:12.602 WARN  (main) [] o.e.j.u.s.S.config No Client EndPointIdentificationAlgorithm configured for Client@2c9a6717[provider=null,keyStore=null,trustStore=null]
2023-02-22 09:31:12.690 WARN  (main) [] o.e.j.u.s.S.config No Client EndPointIdentificationAlgorithm configured for Client@760487aa[provider=null,keyStore=null,trustStore=null]
```

That is due to this line:
```
sslContextFactory.setEndpointIdentificationAlgorithm(
        System.getProperty("solr.jetty.ssl.verifyClientHostName"));
```

It seems like this stems from https://issues.apache.org/jira/browse/SOLR-14163, so we have the perfect people to discuss this @janhoy & @risdenk ! I'll leave it to y'all if we want to use "HTTPS" as the default. That will make the last 2 warnings go away. We can also deal with this in a different PR/issue if y'all want to, it's pretty unrelated.

janhoy (approved):
Thanks for checking Houston. Agree to defer https / logs to another issue

risdenk:
Oooo thanks for digging @HoustonPutman agree that 

```
sslContextFactory.setEndpointIdentificationAlgorithm(
        System.getProperty("solr.jetty.ssl.verifyClientHostName"));
```

looks to be the issue. We should default to `HTTPS` if TLS is not enabled. It looks like we disable client hostname verification by default and the setting `solr.jetty.ssl.verifyClientHostName` only applies if TLS is enabled.

PR #1372, opened 2023-02-20
SOLR-16672: Remove use of setDefaultCollection in SolrCLI

epugh:
https://issues.apache.org/jira/browse/SOLR-16672

found unused (I believe!) parameter...


dsmiley (approved):
Dunno if the option is documented... please look but +1 in general.

PR #1374, opened 2023-02-21
SOLR-16623:SolrClientTestRule for JettySolrRunnerV2

joshgog:
https://issues.apache.org/jira/browse/SOLR-16623

# Description

This PR introduces a SolrClientTestRule for JettySolrRunner

# Solution

Introduced SolrClientTestRule abstract class that extends JUnit's ExternalResource class, which is a TestRule that allows for specifying methods to be run before and after a test. SolrJettyTestRule is a concrete implementation of the abstract class and is declared as a ClassRule field in SolrJettyTestBase.

# Tests

# Checklist


solr/core/src/test/org/apache/solr/TestSolrCoreProperties.java
```diff
@@ -79,11 +79,10 @@ public static void beforeTest() throws Exception {
     if (System.getProperty("solr.data.dir") == null) {
       nodeProperties.setProperty("solr.data.dir", createTempDir().toFile().getCanonicalPath());
     }
-    jetty =
-        new JettySolrRunner(homeDir.getAbsolutePath(), nodeProperties, buildJettyConfig("/solr"));
 
-    jetty.start();
-    port = jetty.getLocalPort();
+    new JettySolrRunner(homeDir.getAbsolutePath(), nodeProperties, buildJettyConfig("/solr"));
```
epugh, line 83: does this guy need to be set to something?

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
```diff
@@ -0,0 +1,122 @@
...
+import java.nio.file.Path;
+import java.util.Properties;
+import org.apache.solr.SolrTestCaseJ4;
+import org.apache.solr.client.solrj.SolrClient;
+import org.apache.solr.embedded.JettyConfig;
+import org.apache.solr.embedded.JettySolrRunner;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SolrJettyTestRule extends SolrClientTestRule {
```
epugh, line 35: Can we get some javadocs on how to use this?   I get that it's a SolrJettyTestRule, but can you provide some docson when I would use it?  What are good use cases for using this rule....????

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  public JettySolrRunner getJetty() {
+    if (jetty == null) throw new IllegalStateException("Jetty has not started");
+    return jetty;
+  }
+
+  @Override
+  public SolrClient getSolrClient(String name) {
+    throw new UnsupportedOperationException(); // TODO Client needed
```
epugh, line 115: is this still a todo?

solr/core/src/test/org/apache/solr/TestTolerantSearch.java
```diff
@@ -104,10 +104,7 @@ public static void destroyThings() throws Exception {
       collection2.close();
       collection2 = null;
     }
-    if (null != jetty) {
-      jetty.stop();
-      jetty = null;
-    }
+    solrClientTestRule.reset();
```
epugh, line 107: i mentally thought that by using a test rule, some of this clean up stuff would just be handled by us, because it's a rule?

epugh:
Really excited to see progress on anything that makes testing simpler!   I hope my comments are being asked prematurely, if you aren't quite ready for review.    I am somewhat trying to figure out how this makes our lives simpler in testing...   (Partly me coming up the curve on TestRule's!)...    For example, I expected to see many of the places where we shutdown Jetty to go completely away, as the TestRule would do it?   I see a lot of `solrClientTestRule.reset();` however....

solr/core/src/test/org/apache/solr/TestSolrCoreProperties.java
joshgog: Right! He was supposed to do this `solrClientTestRule.startSolr(Path.of(homeDir.getAbsolutePath()), nodeProperties, buildJettyConfig("/solr"));`

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
joshgog: Its now implemented.` return getHttpSolrClient(jetty.getBaseUrl().toString() + "/" + DEFAULT_TEST_CORENAME);`

joshgog:
The reset() method is not desired to be in the rule that's why its annotated with  ` @Deprecated // Prefer not to have this.`

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
epugh: awesome!

solr/test-framework/src/java/org/apache/solr/SolrJettyTestBase.java
```diff
@@ -125,7 +125,7 @@ public static void afterSolrJettyTestBase() throws Exception {
 
   public synchronized SolrClient getSolrClient() {
     if (client == null) {
-      client = createNewSolrClient();
+      client = solrClientTestRule.getSolrClient();
```
epugh, line 128: I LOVE this!    So many "createXYZClient" littering the test code!

solr/test-framework/src/java/org/apache/solr/SolrJettyTestBase.java
```diff
@@ -136,7 +136,7 @@ public synchronized SolrClient getSolrClient() {
    * options.
    */
   public SolrClient createNewSolrClient() {
```
epugh, line 138: can we just delete this method?  Do we still need createNewSolrClient?

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
```diff
@@ -52,13 +53,18 @@ protected void after() {
...
+        } catch (IOException ex) {
+          throw new RuntimeException(ex);
+        }
       }
       jetty = null;
     }
   }
 
-  public void reset() throws Exception {
+  @Deprecated // Prefer not to have this.
```
epugh, line 66: How would we remove these?    Is that part of the scope of this PR or it's own PR?

epugh:
Looking forward to this!

epugh:
Thanks for sharing the summary from the review...   I think you answered some questions I had!  Really looking forward to seeing this land ;-).

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
joshgog: It not part of the scope of this PR.

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
```diff
@@ -0,0 +1,130 @@
...
+import org.apache.solr.SolrTestCaseJ4;
+import org.apache.solr.client.solrj.SolrClient;
+import org.apache.solr.embedded.JettyConfig;
+import org.apache.solr.embedded.JettySolrRunner;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SolrJettyTestRule extends SolrClientTestRule {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  private SolrClient client = null;
```
dsmiley, line 36: There will no longer be one client but a map of clients keyed by collection name.

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
dsmiley: It communicates we have a method that we don't really love; debatable.  If we had the time to tend to each and every subclass of SolrJettyTestBase, I think we wouldn't have needed this.  A comment would be helpful Joshua.

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
```diff
@@ -0,0 +1,130 @@
...
+  public void startSolr(Path solrHome, Properties nodeProperties, JettyConfig jettyConfig) {
+
+    jetty = new JettySolrRunner(solrHome.toString(), nodeProperties, jettyConfig);
+    try {
+      jetty.start();
+    } catch (RuntimeException e) {
+      throw e;
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
```
dsmiley, line 104: Finally; you figured it out :-)

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
```diff
@@ -0,0 +1,130 @@
...
+    startSolr(
+        solrHome,
+        new Properties(),
+        JettyConfig.builder()
+            .withSSLConfig(SolrTestCaseJ4.sslConfig.buildServerSSLConfig())
+            .build());
+  }
+
+  public void startSolr(Path solrHome, Properties nodeProperties, JettyConfig jettyConfig) {
+
```
dsmiley, line 96: Let's check "jetty" isn't already set?  Would be nasty if this were to happen accidentally.

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
```diff
@@ -0,0 +1,130 @@
...
+    if (jetty == null) throw new IllegalStateException("Jetty has not started");
+    return jetty;
+  }
+
+  @Override
+  public SolrClient getSolrClient(String name) {
+    if (clients.containsKey(name)) {
+      return clients.get(name);
+    } else {
+      clients.put(
```
dsmiley, line 120: see computeIfAbsent or similar

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
```diff
@@ -0,0 +1,130 @@
...
+          name, getHttpSolrClient(jetty.getBaseUrl().toString() + "/" + DEFAULT_TEST_CORENAME));
+      return getHttpSolrClient(jetty.getBaseUrl().toString() + "/" + DEFAULT_TEST_CORENAME);
```
dsmiley, line 122: Why do you create two clients in this method?  Surely not correct.

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
```diff
@@ -0,0 +1,130 @@
...
+import org.slf4j.LoggerFactory;
+
+public class SolrJettyTestRule extends SolrClientTestRule {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  private SolrClient client = null;
+  private JettySolrRunner jetty;
+  private SolrClient adminClient = null;
+
+  private ConcurrentHashMap<String, SolrClient> clients =
+      new ConcurrentHashMap<>(); // TODO close them
```
dsmiley, line 41: That's a critical TODO by the way, and an easy one to get done.

solr/test-framework/src/java/org/apache/solr/util/SolrClientTestRule.java
```diff
@@ -135,14 +135,16 @@ protected void create(NewCollectionBuilder b) throws SolrServerException, IOExce
...
+   */
   public abstract SolrClient getAdminClient();
 
-  /** Provides a SolrClient instance for collection1 */
+  /** Provides a SolrClient instance for collection1. The caller doesn't need to close it */
   public SolrClient getSolrClient() {
     return getSolrClient("collection1");
   }
-
+  /** Provides a SolrClient instance for provided name. The caller doesn't need to close it */
```
dsmiley, line 147: "For provided name"?  And what is this "name"?

solr/core/src/test/org/apache/solr/core/TestConfigSetImmutable.java
```diff
@@ -65,11 +66,8 @@ public void before() throws Exception {
 
   @After
   public void after() throws Exception {
-    if (jetty != null) {
-      jetty.stop();
-      jetty = null;
-    }
-    client = null;
+    solrClientTestRule.reset();
+    SolrJettyTestRule.client = null;
```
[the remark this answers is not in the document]
dsmiley: Looking at the state of SolrJettyTestRule right now, it's already *not* static and is public -- which is good.  Your PR doesn't compile.  I suggest never pushing code that doesn't even compile.

solr/test-framework/src/java/org/apache/solr/SolrJettyTestBase.java
```diff
@@ -138,17 +142,15 @@ public static void afterSolrJettyTestBase() throws Exception {
...
-   * options.
-   */
-  public SolrClient createNewSolrClient() {
-    return solrClientTestRule.getSolrClient();
+  public synchronized SolrClient getSolrClient() {
+    if (getClient() == null) {
+      return solrClientTestRule.getSolrClient();
+    }
+    return getClient();
   }
```
dsmiley, line 150: When the method fundamentally changes away from using fields on this class to calling the rule, it no longer needs to be synchronized because the rule's getSolrClient is thread-safe.

also, don't see why getClient exists nor why it would return null.

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
```diff
@@ -31,18 +31,13 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+// TODO write javadocs
```
dsmiley, line 34: you will find a ton of TODOs across Solr but no "nocommit".  "nocommit" is checked for in the build and the PR is unmergeable until these are addressed.

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
```diff
@@ -31,18 +31,13 @@
...
 public class SolrJettyTestRule extends SolrClientTestRule {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
-  private SolrClient client = null;
   private JettySolrRunner jetty;
+  private SolrClient client = null;
```
dsmiley, line 38: Why is "client" its own field yet we've got this "clients" map?

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
```diff
@@ -75,9 +70,22 @@ protected void after() {
       }
       client = null;
     }
+
+    for (SolrClient solrClient : clients.values()) {
```
dsmiley, line 74: The sequence of actions in a lifecycle method (and this is one) can sometimes matter.  Do things in reverse order in a closing/after type of method.  Thus clients should be closed and *then* close Jetty.  It's the reverse order from starting.

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
dsmiley: I don't think you understood my meaning; admittedly what I said was ambiguous.  Let me put it like this, if someone calls startSolr yet startSolr was already called, I believe this is quite erroneous!  To *not* throw an exception and pretend everything is fine could easily lead to strange behavior to the caller (the test).  Like maybe the startup parameters changed.  But fundamentally, the test is simply using this wrong.

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
```diff
@@ -94,33 +102,37 @@ public void startSolr(Path solrHome) {
...
-    log.info("Jetty Assigned Port#{}", port);
-    adminClient = getHttpSolrClient(jetty.getBaseUrl().toString());
   }
 
   public JettySolrRunner getJetty() {
     if (jetty == null) throw new IllegalStateException("Jetty has not started");
     return jetty;
   }
 
+  public SolrClient getClient() {
```
dsmiley, line 126: What's this for?  Seems redundant with getSolrClient; no?  If you had documented it, maybe you would have self-discovered the duplicity

dsmiley (commented):
MUCH closer, finally

epugh:
This is going to be great!

solr/test-framework/src/java/org/apache/solr/SolrJettyTestBase.java
joshgog: Deleting it makes tests to fail with the error `java.lang.IllegalStateException: Connection pool shut down` I'm trying to crack the cause

solr/solrj/src/test/org/apache/solr/client/solrj/response/TestSuggesterResponse.java
```diff
@@ -119,8 +119,8 @@ public void testEmptySuggesterResponse() throws Exception {
   }
 
   private void addSampleDocs() throws SolrServerException, IOException {
-    getClient().deleteByQuery("*:*");
-    getClient().commit(true, true);
+    solrClientTestRule.getSolrClient().deleteByQuery("*:*");
+    solrClientTestRule.getSolrClient().commit(true, true);
```
dsmiley, line 123: Nope; we agreed we'd leave much test code as it was, simply calling `getSolrClient` defined by SolrJettyTestBase which will indeed get the client from the rule.

dsmiley:
I don't recommend adding `// nocommit: Fails` type comments.  The CI results tells you what fails, which can change as you improve things.  In fact you just made a critical improvement and it seems there are only a few failures left.  How's it going with them?

dsmiley:
I was going to mention that your "admin client" thing could be removed in lieu of a simple getSolrClient("") (empty string) with special handling?  I think this would simplify the code / lifecycle.  WDYT?

joshgog:
> I was going to mention that your "admin client" thing could be removed in lieu of a simple getSolrClient("") (empty string) with special handling? I think this would simplify the code / lifecycle. WDYT?

Sure will work on that

joshgog:
> I don't recommend adding `// nocommit: Fails` type comments. The CI results tells you what fails, which can change as you improve things. In fact you just made a critical improvement and it seems there are only a few failures left. How's it going with them?

Still working on them

epugh:
Some nice cleans up @dsmiley in the current round!  The tests read much simpler...

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
dsmiley: I'd say that in general you should test with the most basic Solr possible based on what you are testing.  If you can use EmbeddedSolrServerTestRule then do that.  But if you need to test something relating to HTTP or Jetty then you're going to have to use this.   Likewise, this isn't good enough for SolrCloud.

As a subclass of SolrClientTestRule, the hope is that it be familiar enough so that how to use it is self-explanatory provided you're familiar with the other SolrClientTestRule's (there is just one other right now).

If this is embraced everywhere that it should, fewer tests need to even be aware of JettySolrRunner.  And maybe SolrJettyTestBase won't exist.

I'm maybe not directly answering your question but I'm hoping it's more self evident as it's used more and more.  Initially, mostly it's just SolrJettyTestBase.  Maybe more of SJTB should be folded into this rule.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientProxyTest.java
dsmiley, line 1: @stillalex please take a look at my changes here and let me know if you think they are fine, or if you have any comments in general

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
```diff
@@ -0,0 +1,129 @@
...
+  public void startSolr(Path solrHome) {
+    startSolr(
+        solrHome,
+        new Properties(),
+        JettyConfig.builder()
+            .withSSLConfig(SolrTestCaseJ4.sslConfig.buildServerSSLConfig())
+            .build());
+  }
+
+  /** Enables proxy feature to disable connections. Must be called prior to starting. */
```
dsmiley, line 82: I copied the first sentence from a comment in JettySolrRunner written by Mark Miller but honestly I don't understand it.  @stillalex maybe you could recommend better info as you just recently used it?

dsmiley (approved):
Summary
----
Introduce SolrJettyTestRule, a new subclass of SolrClientTestRule for tests requiring Jetty or HTTP.  
Perhaps most tests using JettySolrRunner could convert to this eventually.

Only converted SolrJettyTestBase, (thus the many tests using it) which should be considered a deprecated test utility now.  Don’t need a base test class just to use Jetty; you simply need to use SolrJettyTestRule.  Changed many of these subclasses in trivial ways like to reduce Jetty references.

Moved SolrJettyTestBase.legacyExampleCollection1SolrHome to SolrTestCaseJ4 because wasn’t Jetty related; was used by non-Jetty tests.

----

I don't love that the clients it returns are Apache HttpClient and not Jetty but that would be a follow-on to migrate some of those tests.

Other than the possibility of further review, I think it's ready to merge, finally.  And of course a CHANGES.txt entry...

SOLR-16623: new SolrJettyTestRule for tests needing HTTP or Jetty.  (David Smiley, Joshua Ouma)

solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientProxyTest.java
stillalex: looks good! I like this approach and being the only proxy test as far as I can see, I have to say thank you for adding support for it :)

solr/test-framework/src/java/org/apache/solr/util/SolrJettyTestRule.java
stillalex: I think the following does a slightly better job
> Enables proxy feature to allow for failure injection testing at the inter-node communication level

please rewrite as needed :)

PR #1375, opened 2023-02-22
Enable renovatebot for the project

janhoy:
This PR adds automatic dependency housekeeping to the project through [Renovatebot](https://github.com/renovatebot/renovate), which is an alternative to dependabot. See [slack discussion here](https://the-asf.slack.com/archives/CEKUCUNE9/p1674220374400369).

So once this is merged, a self-hosted bot called [SolrBot](https://github.com/solrbot), will kick off every Sunday and file PRs for dependencies that need upgrades. Here is an [example of such a PR](https://github.com/cominvent/solr-playground/pull/83) from my sandbox test project. A nice feature is that the PR will also include changes from `gradlew updateLicenses` and `--write-locks`, so for a majority of upgrades, there's nothing left to do for the committer.

We can't use Dependabot since it does not support our gradle "consistent-versions" plugin. We can't use the public Renovatebot github app since it cannot be configured to run custom post-commands like `gradlew updateLicenses`. So we follow GitHub's advice of creating a custom github account for the bot, and that user will run renovatebot in a GitHub Action, and fork the solr repo and file PRs against the Solr repo. This has the nice property that the bot won't need commit-rights to the solr-repo, which would be disallowed by ASF policy.

~I plan to share the github credentials for the [SolrBot](https://github.com/solrbot) account with the PMC if I can find a secure way of doing it. I remember there is a private svn repo per PMC, but could not find it.~

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,5 +1,6 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "description": "Runs Renovate with solrbot, see dev-docs/dependency-upgrades.adoc for more",
   "enabled": true,
   "dependencyDashboard": false,
   "enabledManagers": ["gradle"],
@@ -11,15 +12,28 @@
   },
   "packageRules": [
     {
+      "description": "Fix for non-semantic versions for older artifacts",
       "matchDatasources": ["maven"],
       "matchPackageNames": ["commons-collections:commons-collections", "commons-io:commons-io", "commons-lang:commons-lang"],
       "versioning": "regex:^(?<major>\\d{1,4})\\.(?<minor>\\d+)(\\.(?<patch>\\d+))?$"
     },
     {
+      "description": "Group these together in same PR",
+      "matchPackagePatterns": ["org.apache.httpcomponents"],
+      "groupName": "httpcomponents"
+    },
+    {
+      "description": "Test-dependencies are checked less often than the shipped deps",
+      "matchDepTypes": ["test"],
+      "extends": ["schedule:monthly"]
+    },
+    {
+      "description": "Noisy, frequently updated dependencies checked less often",
       "matchPackagePrefixes": ["software.amazon.awssdk"],
       "extends": ["schedule:monthly"]
     },
     {
+      "description": "Workaround for https://github.com/renovatebot/renovate/issues/19226",
       "matchPackageNames": ["solr:modules", "HH:mm"],
       "enabled": false
     }
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,5 +1,6 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "enabled": true,
   "dependencyDashboard": false,
   "enabledManagers": ["gradle"],
   "includePaths": ["versions.props", "build.gradle"],
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -20,7 +20,7 @@
     {
       "description": "Group these together in same PR",
       "matchPackagePatterns": ["org.apache.httpcomponents"],
-      "groupName": "httpcomponents"
+      "groupName": "org.apache.httpcomponents"
     },
     {
       "description": "Test-dependencies are checked less often than the shipped deps",
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -24,8 +24,10 @@
     }
   ],
   "schedule": [
-    "every sunday"
+    "* 8-20 * * *"
   ],
   "labels": ["dependencies"],
-  "prConcurrentLimit": 5
+  "prConcurrentLimit": 5,
+  "prHourlyLimit": 5,
+  "stabilityDays": 5
 }
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,30 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "dependencyDashboard": false,
+  "enabledManagers": ["gradle"],
+  "ignorePaths": ["gradle/**", "solr/**", "buildSrc/**", "dev-docs/solr-missing-doclet/**"],
+  "postUpgradeTasks": {
+    "commands": ["./gradlew updateLicenses"],
+    "fileFilters": ["solr/licenses/*.sha1"],
+    "executionMode": "branch"
+  },
+  "packageRules": [
+    {
+      "matchDatasources": ["maven"],
+      "matchPackageNames": ["commons-collections:commons-collections", "commons-io:commons-io", "commons-lang:commons-lang"],
+      "versioning": "regex:^(?<major>\\d{1,4})\\.(?<minor>\\d+)(\\.(?<patch>\\d+))?$"
+    },
+    {
+      "matchPackagePrefixes": ["software.amazon.awssdk"],
+      "extends": ["schedule:monthly"]
+    },
+    {
+      "matchPackageNames": ["solr:modules", "HH:mm"],
+      "enabled": false
+    }
+  ],
+  "schedule": [
+    "every sunday"
+  ],
+  "prConcurrentLimit": 5
+}
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -4,7 +4,7 @@
   "enabled": true,
   "dependencyDashboard": false,
   "enabledManagers": ["gradle"],
-  "includePaths": ["versions.props", "build.gradle"],
+  "includePaths": ["versions.*", "build.gradle"],
   "postUpgradeTasks": {
     "commands": ["./gradlew updateLicenses"],
     "fileFilters": ["solr/licenses/*.sha1"],
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -2,7 +2,7 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "dependencyDashboard": false,
   "enabledManagers": ["gradle"],
-  "ignorePaths": ["gradle/**", "solr/**", "buildSrc/**", "dev-docs/solr-missing-doclet/**"],
+  "includePaths": ["versions.props", "build.gradle"],
   "postUpgradeTasks": {
     "commands": ["./gradlew updateLicenses"],
     "fileFilters": ["solr/licenses/*.sha1"],
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -41,7 +41,6 @@
   "schedule": [
     "* 8-20 * * *"
   ],
-  "labels": ["dependencies"],
   "prConcurrentLimit": 5,
   "prHourlyLimit": 5,
   "stabilityDays": 5
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -26,5 +26,6 @@
   "schedule": [
     "every sunday"
   ],
+  "labels": ["dependencies"],
   "prConcurrentLimit": 5
 }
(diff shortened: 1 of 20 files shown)


.github/renovate.json
```diff
@@ -0,0 +1,30 @@
...
+      "matchPackagePrefixes": ["software.amazon.awssdk"],
+      "extends": ["schedule:monthly"]
+    },
+    {
+      "matchPackageNames": ["solr:modules", "HH:mm"],
+      "enabled": false
+    }
+  ],
+  "schedule": [
+    "every sunday"
```
janhoy, line 27: The schedule for when to run can be adjusted. I chose every sunday as a start, to avoid getting a constant flow of PRs throughout the week. We can adjust it as we gain experience. We may also decide to flip this to "never" to mute new PRs until closer to a release.

.github/renovate.json
```diff
@@ -0,0 +1,30 @@
...
+  ],
+  "prConcurrentLimit": 5
```
janhoy, line 29: This setting will keep a maximum of 5 PRs open at a time. It is just a temporary setting for the first run. I don't think we need a limit here. Nice to be able to filter by a label or author and see all outstanding dep upgrades.

.github/renovate.json
```diff
@@ -0,0 +1,30 @@
...
+  "postUpgradeTasks": {
+    "commands": ["./gradlew updateLicenses"],
+    "fileFilters": ["solr/licenses/*.sha1"],
+    "executionMode": "branch"
+  },
+  "packageRules": [
+    {
+      "matchDatasources": ["maven"],
+      "matchPackageNames": ["commons-collections:commons-collections", "commons-io:commons-io", "commons-lang:commons-lang"],
+      "versioning": "regex:^(?<major>\\d{1,4})\\.(?<minor>\\d+)(\\.(?<patch>\\d+))?$"
```
janhoy, line 15: Workaround for some maven artifacts that used `yyyymmdd.xxx` format for some early versions, causing a suggestion to "upgrade" to that version.

.github/renovate.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "dependencyDashboard": false,
+  "enabledManagers": ["gradle"],
+  "ignorePaths": ["gradle/**", "solr/**", "buildSrc/**", "dev-docs/solr-missing-doclet/**"],
```
janhoy, line 5: We only care about versions recorded in `versions.props`, not inside every `*.gradle` file.

janhoy (commented):
Here are some comments on the suggested config. The reference guide for how to configure the bot is here https://docs.renovatebot.com

janhoy:
As of now, the git author for the PRs will be `SolrBot <solrbot@cominvent.com>`. I used an email address on my own downain, the same I used to create the github account. The git commit message could of course be anything, so perhaps better with something like `SolrBot <solrbot@solr.apache.org>`?

Regarding the email of the github account for the bot, I'd like to switch that to e.g. private@solr.apache.org. I turned off all unnecessary notification mails, but if there was a failed Github workflow run, there would be an email, and for password recovery etc. Could request a new (private) mail list, but seems overkill?

janhoy:
> I remember there is a private svn repo per PMC, but could not find it.

I found https://svn.apache.org/repos/private/pmc/lucene/ for Lucene, but there is no equivalent for Solr at https://svn.apache.org/repos/private/pmc/solr/ - I can ask Infra to create one, with ACL for PMC only, a place we can store secrets.

risdenk (approved):

.github/renovate.json
dsmiley: Why ignore solr-missing-doclet specifically; I think this "ignorePaths" is too specific in this case.  In other words, why not ignore all of dev-docs?  Trying to reduce maintenance to this config line & to be simpler.

Also; no idea if it's possible to configure a GHA to be include-list instead of exclude-list.  Include-list is obviously much more sensible for this PR!

janhoy: Originally I allowed renovate to suggest upgrades to everything, including `dev-tools/scripts/requirements.txt`, webapp JS libs, Dockerfile (image versions), antora etc. Then I restricted it to only "gradle" (maven pkgs).

I think this will work to only consider the two files where we define versions:
```json
"includePaths": ["versions.props", "build.gradle"]
```

chatman:
> I changed teh commit email to SolrBot <solrbot@solr.apache.org>, see [renovate config](https://github.com/solrbot/renovate-github-action/blob/main/solr/renovate-config.js).

Who controls that email address?

> Regarding the email of the github account for the bot, I'd like to switch that to e.g. [private@solr.apache.org](mailto:private@solr.apache.org).

Oh god! Of all the bad ideas in the world, why am I seeing several of the worst ones here in this single issue?

.github/renovate.json
```diff
@@ -0,0 +1,33 @@
+{
```
chatman, line 1: Does this file need to be in solr repository for the entire integration to work? Can't it be added to the folder after a clone by the GHA by the https://github.com/solrbot/renovate-github-action project?

dsmiley: If it can live elsewhere easily then cool.  But I think it's inclusion isn't a blocker either.  I believe Ishan's concern is an appearance of "official"-ness?  Easily addressed with comments... but not sure in JSON.  Hmm.  A renovate-README.txt in this dir could have some commentary.

janhoy:
> IMO, this separate GH account (solrbot) is outside the purview of Apache Solr project, and is a _3rd party_ project. We shouldn't need the private@ or PMC level involvement for this project. If you disagree, let us discuss in the private@ list about it.

After conclusion from private@ list discussion, we don't need this bot / github action to be under official control of the Solr PMC. So I'll re-brand this project as being a 3rd party community project, both in this PR and over in the solrbot account.

That means I'm reverting back to `solrbot@cominvent.com` as the commit email address.

.github/renovate.json
janhoy: I'll add a `dev-docs/dependency-upgrades.adoc` which both explains manual dep upgrades as well as documents what `.github/renovate.json` is for.

While tecnically we could perhaps embed all the config from `renovate.json` into the runner project, I like the fact that `renovate.json` gives the project a place to disable/enable the bot, tune schedules and handle future custom rules for dependencies that may require it.

janhoy: Please review the new dev-doc. It has some overlap with `help/dependencies.txt` but from different angles.

chatman: > While tecnically we could perhaps embed all the config from renovate.json into the runner project

+1 to renovate.json staying in the runner project. Dependabot didn't need any custom config files in the main Solr project, AFAIK. I don't see a need for us to maintain configurations for a third party integration in our main repository.

dev-docs/dependency-upgrades.adoc
```diff
@@ -0,0 +1,64 @@
...
+1. Identify the available versions from e.g. https://search.maven.org[Maven Central]
+2. Update the version in `versions.props` file
+3. Run `./gradlew --write-locks` to re-generate `versions.lock`. Note that this may cause a cascading effect where
+   the locked version of other dependencies also change.
+4. Run `./gradlew updateLicenses` to re-generate SHA1 checksums of the new jar files.
+5. Once in a while, a new version of a dependency will transitively bring in brand-new dependencies.
+   You'll need to decide whether to keep or exclude them. See `help/dependencies.txt` for details.
+
+== Renovate bot Pull Requests
+A member of the Solr community operates a Github bot running https://github.com/renovatebot/renovate[Renovate], which
```
chatman, line 40: Instead of all this text on renovatebot in our docs, we should put them all in the external repository and maybe add a one liner here referring to it.

epugh: At the risk of being flamed, I don't know why you are so against the word "renovatebot" in our repo?   If I see PR's showing up in Solr, I would love to see some documentation about it...   I think more documentation is a good thing.

dev-docs/dependency-upgrades.adoc
```diff
@@ -0,0 +1,64 @@
...
+files Pull Requests to Solr with dependency upgrade proposals. The PRs are labeled `dependencies` and do include
+changes resulting from `--write-locks` and `updateLicenses`.
+
+Community members and committers can then review, and if manual changes are needed, help bring the PR to completion.
+For many dependencies, a changelog is included in the PR text, which may help guide the upgrade decision.
+
+The scans are run on a schedule. New PRs are filed every Sunday, and only dependency versions that are at least
```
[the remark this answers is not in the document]
epugh: I suspect you have more energy for this discussion than I do...   I have some points of why I disagree, but I don't have the emotional energy for the argument.

dev-docs/dependency-upgrades.adoc
chatman: I didn't say I'm against the "word renovatebot" in our repo. If you read carefully, I said that so much verbose explanation can reside in the external project and a concise one liner describing renovatebot with a link to the external project should be better.

dev-docs/dependency-upgrades.adoc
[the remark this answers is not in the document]
chatman: > I suspect you have more energy for this discussion than I do ...  I don't have the emotional energy for the argument.

I don't see why you have to be snarky about this. Jan requested more eyes and feedback on the document, and I provided mine. I have no vested interest in arguing with anyone. It is up to Jan to decide how to handle my feedback.

dev-docs/dependency-upgrades.adoc
chatman: >  If I see PR's showing up in Solr, I would love to see some documentation about it

If the bot can add a link to the documentation page (in its own repo) along with each PR, then this concern can be addressed.

.github/renovate.json
janhoy: Thanks for the input. However, I disagree, and choose to keep it. It's analoguys to our `.lift.toml` and similar files used to enable and configure various tools.

dev-docs/dependency-upgrades.adoc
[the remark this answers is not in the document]
janhoy: Let's fail on more rather than less documentation. This is a dev docs.

dev-docs/dependency-upgrades.adoc
janhoy: Let's gain some experience and then revisit docs.

janhoy: @chatman I removed the "Maintaining the bot" paragraph from the dev-doc. Agree it is enough to mention it over at the bot README.

janhoy:
Turns out the bot account needs "triage" rights on the repo in order to set the "dependencies" label on PRs. ASF.yaml [provides a way to grant that](https://cwiki.apache.org/confluence/display/INFRA/Git+-+.asf.yaml+features#Git.asf.yamlfeatures-AssigningexternalcollaboratorswiththetriageroleonGitHub).

However, since we currently don't use labels at all for our PRs, I'll instead skip the `labels` config. We can always filter by Author. We can introduce labels later.

janhoy:
Thanks for weighing in, now let's see the first PRs flowing in :)

.github/renovate.json
```diff
@@ -0,0 +1,47 @@
...
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "description": "Runs Renovate with solrbot, see dev-docs/dependency-upgrades.adoc for more",
+  "enabled": true,
+  "dependencyDashboard": false,
+  "enabledManagers": ["gradle"],
+  "includePaths": ["versions.*", "build.gradle"],
+  "postUpgradeTasks": {
```
HoustonPutman, line 8: This is great, and a step that's missing in the dependabot feature for the solr-operator.

PR #1376, opened 2023-02-22
SOLR-16679: Fix solr.jetty.ssl.verifyClientHostName logging

risdenk:
https://issues.apache.org/jira/browse/SOLR-16679

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -223,6 +223,8 @@ Bug Fixes
 
 * SOLR-16668: Use default to Java SSL for Http2SolrClient when none is provided (Houston Putman)
 
+* SOLR-16679: Fix solr.jetty.ssl.verifyClientHostName logging (Kevin Risden)
+
 Build
 ---------------------
 * Upgrade forbiddenapis to 3.4 (Uwe Schindler)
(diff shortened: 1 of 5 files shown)


HoustonPutman (approved):
This looks good to me, but could you provide more information as to why we want a default value of `null` when SSL is enabled? Why is `HTTPS` a bad default in all cases?

risdenk:
> This looks good to me, but could you provide more information as to why we want a default value of null when SSL is enabled? Why is HTTPS a bad default in all cases?

So ideally `HTTPS` would be set by default in all cases except when we want it disabled. The disabled case is setting it to `null`.

The way that bin/solr is setup today, `solr.jetty.ssl.verifyClientHostName` is defaulted to `HTTPS` when TLS is enabled. So in a round about way the default is `HTTPS`. In order to change this, we would need to update `bin/solr`.

I'll take a crack at doing that since it would make the logic much cleaner I think.

risdenk:
@HoustonPutman can you take another look? I updated the logic to be clearer I think. It also makes `SOLR_SSL_CLIENT_HOSTNAME_VERIFICATION` match the existing documentation in `solr/bin/solr.in.sh` and `solr/bin/solr.cmd`

PR #1378, opened 2023-02-23
SOLR-16585: examples test for match all paging

mkhludnev:
https://issues.apache.org/jira/browse/SOLR-16585

### DIFF
--- a/solr/solrj/src/test/org/apache/solr/client/solrj/SolrExampleTests.java
+++ b/solr/solrj/src/test/org/apache/solr/client/solrj/SolrExampleTests.java
@@ -32,6 +32,7 @@ import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
@@ -569,6 +570,57 @@ public abstract class SolrExampleTests extends SolrExampleTestsBase {
     assertEquals(2, out.getNumFound());
     assertEquals(0, out.get(0).size());
     assertEquals(0, out.get(1).size());
+  }
+
+  @Test
+  public void testMatchAllPaging() throws Exception {
+    SolrClient client = getSolrClient();
+
+    // Empty the database...
+    client.deleteByQuery("*:*"); // delete everything!
+    if (random().nextBoolean()) {
+      client.commit();
+    }
+    // Add eleven docs
+    List<SolrInputDocument> docs = new ArrayList<>();
+    for (int i = 0; i < 11; i++) {
+      SolrInputDocument doc = new SolrInputDocument();
+      doc.addField("id", "id" + i);
+      doc.addField("name", "doc" + i);
+      doc.addField("price", "" + i);
+      docs.add(doc);
+      if (rarely()) {
+        client.commit();
+      }
+    }
+    client.add(docs);
+    if (random().nextBoolean()) {
+      client.commit();
+    } else {
+      client.optimize();
+    }
+    final List<String> sorts = Arrays.asList("_docid_", "id", "name", "price", null);
+    Collections.shuffle(sorts, random());
+    final List<Integer> starts = Arrays.asList(0, 1, 2, 10, 11, 12);
+    Collections.shuffle(starts, random());
+    for (String sort : sorts.subList(0, 1 + random().nextInt(sorts.size() - 1))) {
+      final SolrQuery query = new SolrQuery("*:*");
+      if (sort != null) {
+        query.setSort(sort, random().nextBoolean() ? SolrQuery.ORDER.asc : SolrQuery.ORDER.desc);
+      }
+      for (int start : starts.subList(0, 1 + random().nextInt(starts.size() - 1))) {
+        if (start > 0 || random().nextBoolean()) {
+          query.setStart(start);
+        }
+        SolrDocumentList results = client.query(query).getResults();
+        assertEquals(11, results.getNumFound());
+        assertEquals("page from " + start, Math.max(Math.min(10, 11 - start), 0), results.size());
+        for (SolrDocument doc : results) {
+          assertTrue(doc.containsKey("id") && doc.containsKey("name") && doc.containsKey("price"));
+          break;
+        }
+      }
+    }
   }
 
   private String randomTestString(int maxLength) {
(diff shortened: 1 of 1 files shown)


solr/solrj/src/test/org/apache/solr/client/solrj/SolrExampleTests.java
```diff
@@ -571,6 +572,57 @@ public void testGetEmptyResults() throws Exception {
...
+    // Add eleven docs
+    List<SolrInputDocument> docs = new ArrayList<>();
+    for (int i = 0; i < 11; i++) {
+      SolrInputDocument doc = new SolrInputDocument();
+      doc.addField("id", "id" + i);
+      doc.addField("name", "doc" + i);
+      doc.addField("price", "" + i);
+      docs.add(doc);
+      if (rarely()) {
+        client.commit();
```
cpoerschke, line 593: Is `client.add(docs);` need here and optionally `docs.clear()` after?

```suggestion
        client.add(docs);
        client.commit();
        // docs.clear();
```

solr/solrj/src/test/org/apache/solr/client/solrj/SolrExampleTests.java
```diff
@@ -571,6 +572,57 @@ public void testGetEmptyResults() throws Exception {
...
+        final SolrQuery query = new SolrQuery("*:*");
+        if (sort != null) {
+          query.setSort(sort, random().nextBoolean() ? SolrQuery.ORDER.asc : SolrQuery.ORDER.desc);
+        }
+        if (start > 0 || random().nextBoolean()) {
+          query.setStart(start);
+        }
+        SolrDocumentList results = client.query(query).getResults();
+        assertEquals(11, results.getNumFound());
+        assertEquals("page from " + start, Math.max(Math.min(10, 11 - start), 0), results.size());
```
cpoerschke, line 617: Hmm, not sure about the `Math.max(Math.min(10, 11 - start), 0)` expression e.g. for `start=0` it would evaluate to `10` but paging from `0` would return `11` results, no?

mkhludnev: but, default `rows=0`, it will cut 11 results to 10-items page.

solr/solrj/src/test/org/apache/solr/client/solrj/SolrExampleTests.java
mkhludnev: Oh..gosh. Thanks for spotting it @cpoerschke .

solr/solrj/src/test/org/apache/solr/client/solrj/SolrExampleTests.java
cpoerschke: ah, hadn't spotted about implicit `rows=` usage there, thanks!

how about explaining the connection like this?
```suggestion
        assertEquals("page from " + start, Math.max(Math.min(CommonParams.ROWS_DEFAULT, 11 - start), 0), results.size());
```

solr/solrj/src/test/org/apache/solr/client/solrj/SolrExampleTests.java
```diff
@@ -571,6 +572,67 @@ public void testGetEmptyResults() throws Exception {
...
+        }
+        SolrDocumentList results = client.query(query).getResults();
+        assertEquals(docsTotal, results.getNumFound());
+        assertEquals(
+            "page from " + start,
+            Math.max(Math.min(CommonParams.ROWS_DEFAULT, docsTotal - start), 0),
+            results.size());
+        for (SolrDocument doc : results) {
+          assertTrue(doc.containsKey("id") && doc.containsKey("name") && doc.containsKey("price"));
+          break;
```
cpoerschke, line 630: Curious why the `break;` is here? And to consider splitting the `&&` clauses to make any failures easier to diagnose.

```suggestion
          assertTrue(doc.containsKey("id"));
          assertTrue(doc.containsKey("name"));
          assertTrue(doc.containsKey("price"));
```

cpoerschke (approved):

solr/solrj/src/test/org/apache/solr/client/solrj/SolrExampleTests.java
mkhludnev: Pardon. With `for(..) {.. break;}` I tried to say if not empty checking the first one, do nothing otherwise. I saves a few micros (I guess) in comparison to looping all five docs. Let it be.

solr/solrj/src/test/org/apache/solr/client/solrj/SolrExampleTests.java
```diff
@@ -571,6 +572,68 @@ public void testGetEmptyResults() throws Exception {
...
+        final SolrQuery query = new SolrQuery("*:*");
+        if (sort != null) {
+          query.setSort(sort, random().nextBoolean() ? SolrQuery.ORDER.asc : SolrQuery.ORDER.desc);
+        }
+        if (start > 0 || random().nextBoolean()) {
+          query.setStart(start);
+        }
+        if (usually()) {
+          query.setRows(CommonParams.ROWS_DEFAULT);
+        }
```
magibney, line 621: This is ok, but just to point out: issuing a request with rows=ROWS_DEFAULT (either explicit or unset/default) will I think usually cause the docList to be cached in its entirety for the associated sort; so this test will I think largely end up testing queryResultCache?

magibney (approved):
A couple of minor comments, but this looks good, thanks for adding!

Just to check my understanding: it looks like the main difference between this and the test added in bfccca2837e is that this one varies the `sort` param, and is much more likely to hit the queryResultCache. To be sure, those are both good things to test -- but I wonder whether  it would then also be worth varying the main query. It's good to have tests that cover `*:*`, but the test added in bfccca2837e was more specific/nuanced in that it validated specific metrics/caching behavior; the utility of this test is more general. I suspect there may be no existing tests that explicitly target pagination (if there were, that's likely where I would have added tests for `*:*`, and if there were, such tests probably would already have covered `*:*` already, as one of the most obvious query cases!).

It would hit different code paths, and wouldn't require any changes to the logic of expected results, to add a couple of queries that are _not_ `*:*`, but that return the same results -- e.g., `[* TO *]`, `{!prefix f=name v=doc}`?

solr/solrj/src/test/org/apache/solr/client/solrj/SolrExampleTests.java
mkhludnev: Pardon, Michael. I can't get this point, ~~beside of thinking twice~~. Here I just emphasize that following asserts relies on default rows. I don't think it somehow impacts queryResultCache usage, really.

magibney: I see. I get that this block doesn't change the behavior -- so perhaps misleading that I commented on it. My point was that unless `rows` param is set artificially small (and it's always effectively `10` here), the results for all offsets should be within the queryResultCache window, so the first request for a given `sort` will construct a docList, and all subsequent requests for that `sort` will I think be entirely served from the queryResultCache.

mkhludnev:
Thanks to everyone for reviewing. Here's the last confirmation: Is it really worth to push in?

cpoerschke (approved):

PR #1407, opened 2023-02-27
SOLR-16427: Evaluate and fix errorprone rules - part 5

risdenk:
https://issues.apache.org/jira/browse/SOLR-16427

builds on https://github.com/apache/solr/pull/1128

Updates comments on a few rules since we do use them. Enables the following rules
* `MutablePublicArray` - most of these were in SolrCLI and switched to list vs array
* `NonCanonicalType` - fixed a few type references

The changes are in separate commits to help make reviews easier.

### DIFF
--- a/solr/solrj/src/java/org/noggit/JSONUtil.java
+++ b/solr/solrj/src/java/org/noggit/JSONUtil.java
@@ -17,11 +17,12 @@
 package org.noggit;
 
 public class JSONUtil {
-  public static final char[] TRUE_CHARS = new char[] {'t', 'r', 'u', 'e'};
-  public static final char[] FALSE_CHARS = new char[] {'f', 'a', 'l', 's', 'e'};
-  public static final char[] NULL_CHARS = new char[] {'n', 'u', 'l', 'l'};
-  public static final char[] HEX_CHARS =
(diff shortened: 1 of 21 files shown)


epugh (approved):
LGTM

solr/solrj/src/java/org/noggit/JSONUtil.java
```diff
@@ -17,11 +17,19 @@
...
+  @SuppressWarnings("MutablePublicArray")
   public static final char[] NULL_CHARS = new char[] {'n', 'u', 'l', 'l'};
+
+  @SuppressWarnings("MutablePublicArray")
   public static final char[] HEX_CHARS =
       new char[] {'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'};
+
   public static final char VALUE_SEPARATOR = ',';
   public static final char NAME_SEPARATOR = ':';
   public static final char OBJECT_START = '{';
```
dsmiley, line 35: I think we can change the access modifiers to be package-protected.  These are internals that should not be public.  Perhaps we will need the SuppressWarnings annotation but lets do this little thing now too please?

risdenk: addressed and removed the suppresswarnings since it wasn't needed anymore.

PR #1410, opened 2023-03-01
SOLR-16676 Http2SolrClient loss of MDC context

stillalex:
https://issues.apache.org/jira/browse/SOLR-16676

# Description

Adding a helper class to attach MDC context to the request and to clean it up once it finishes. there is a new instance of the helper created on each request to avoid sharing of context data.

I don't know how to add tests to this, it's a very involved piece of machinery. if anyone has any ideas I am open.

# Solution

# Tests

# Checklist


dsmiley:
For testing, see the LogListener utility and see how it's used.  Perhaps you should modify TestRequestId to add an async test of the same using Http2... maybe that makes sense?  Granted that test isn't at an Http level so instead maybe a new TestHttpRequestId subclassing SolrJettyTestBase.

stillalex:
@dsmiley good news and bad news :) I finally managed to add tests (and cover failure scenarios too) thanks for the pointers, I was tempted to not go the extra mile.
now the not so great news. I believe there might be a deadlock on the Http2SolrClient when configured with an ArrayBlockingQueue. I was not able to find tests for this config, but it's definitely possible to have this enabled via the HttpShardHandlerFactory configs. please take a look at https://github.com/apache/solr/pull/1410/commits/a9f9af819ed094a504c60b17da8607e66afc6a67 and possibly tell me what I did wrong.

stillalex:
trying the `ShardHandlerFactory#sizeOfQueue` setting on a Solr 9 definitely shows some weirdness, I'm unable to create a collection:
```
2023-03-02 15:15:40.028 INFO  (qtp2048013503-26) [] o.a.s.s.HttpSolrCall [admin] webapp=null path=/admin/collections params={replicationFactor=1&collection.configName=_default&name=test&router.name=compositeId&action=CREATE&numShards=5&wt=json&_=1677798725094} status=500 QTime=180057
2023-03-02 15:15:40.029 ERROR (qtp2048013503-26) [] o.a.s.s.HttpSolrCall org.apache.solr.common.SolrException: create the collection time out:180s => org.apache.solr.common.SolrException: create the collection time out:180s
	at org.apache.solr.handler.admin.CollectionsHandler.submitCollectionApiCommand(CollectionsHandler.java:433)
org.apache.solr.common.SolrException: create the collection time out:180s
	at org.apache.solr.handler.admin.CollectionsHandler.submitCollectionApiCommand(CollectionsHandler.java:433) ~[?:?]
	at org.apache.solr.handler.admin.CollectionsHandler.invokeAction(CollectionsHandler.java:345) ~[?:?]
	at org.apache.solr.handler.admin.CollectionsHandler.handleRequestBody(CollectionsHandler.java:297) ~[?:?]
	at org.apache.solr.handler.RequestHandlerBase.handleRequest(RequestHandlerBase.java:224) ~[?:?]
	at org.apache.solr.servlet.HttpSolrCall.handleAdmin(HttpSolrCall.java:941) ~[?:?]
	at org.apache.solr.servlet.HttpSolrCall.handleAdminRequest(HttpSolrCall.java:893) ~[?:?]
	at org.apache.solr.servlet.HttpSolrCall.call(HttpSolrCall.java:584) ~[?:?]
	at org.apache.solr.servlet.SolrDispatchFilter.dispatch(SolrDispatchFilter.java:250) ~[?:?]
```

stillalex:
deadlock avoided by setting corePoolSzize to 3. thanks @risdenk! waiting for test results but PR should be open for review again.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java
```diff
@@ -469,6 +474,43 @@ public void testGetById() throws Exception {
     }
   }
 
+  @Test
+  public void testGetByIdArrayBlockingQueue() throws Exception {
+    DebugServlet.clear();
+
+    // client setup needs to be same as HttpShardHandlerFactory
+
```
dsmiley, line 482: ah..... okay, so you are seeing an existing problem.  I would have recommended modifying an existing distributed-search test.  Your new test is fine though.

solr/core/src/test/org/apache/solr/handler/TestHttpRequestId.java
```diff
@@ -0,0 +1,143 @@
...
+import org.apache.solr.common.util.ExecutorUtil;
+import org.apache.solr.common.util.NamedList;
+import org.apache.solr.common.util.SolrNamedThreadFactory;
+import org.eclipse.jetty.client.api.Request;
+import org.eclipse.jetty.client.api.Result;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.slf4j.MDC;
+
+public class TestHttpRequestId extends SolrJettyTestBase {
```
dsmiley, line 38: Your test approach is a bit invasive... uses the internals of Http2 client.  It's okay but I was hoping you would have more of an integration test that like TestRequestId, which is why I referred you to that test.  It does not depend on internal details of SolrClients etc.  I love integration tests as it's more realistic, doesn't assume internals, and also makes it easier to change internals at some point.  Since I now understand you observed specifically an issue with distributed search because of HttpShardHandler's use of that, maybe you could even use a simple query that uses that under the hood?

stillalex: I don't think your example does covers this scenario out of the box. this bug is about MDC being present across the threads running the request. there are no logs at this level, unless I add some specifically for this test, something trace level, but that feels awkward.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java
stillalex: it's not strictly needed anymore, was only trying to reproduce the deadlock. I can remove it if it doesn't fit with the rest of the tests

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -1273,4 +1287,33 @@ static SslContextFactory.Client getDefaultSslContextFactory() {
 
     return sslContextFactory;
   }
+
+  private static class MDCCopyHelper extends RequestResponseListener {
```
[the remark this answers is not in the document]
stillalex: it felt like a good representation of the lifecycle I needed to hook into, so I used it. in the end it turned out it's not stritly needed but I liked the idea of reusing an existing model instead of inventing a new set of names.

solr/core/src/test/org/apache/solr/handler/TestHttpRequestId.java
dsmiley: I'm confused.  You are basically saying that the improvements here are un-observable?  I wonder what motivated this dubious improvement then -- something custom you have?
FWIW I don't think it'd be awkward if you added a trace log.  And FYI it's easy to configure a log level for just one test via the `LogLevel` annotation -- a wonderful thing.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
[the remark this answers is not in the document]
dsmiley: Okay then a comment to this affect would be helpful.

solr/core/src/test/org/apache/solr/handler/TestHttpRequestId.java
stillalex: > You are basically saying that the improvements here are un-observable? I wonder what motivated this dubious improvement then -- something custom you have?

there might be a small misunderstanding here. It's not am improvement I am chasing here, it's a bug, a regression in Solr 9 compared to Solr 8. and I don't have anything custom, this is inside the http2 client as it is used by the HttpShardHandler.
I am trying to setup LogLevel + LogListener in a SolrCloudTestCase, see if I can get something better looking.

stillalex: in the end it was SolrJettyTestBase not the SolrCloudTestCase as I mentioned before. please take a look at this version of the test: it uses logs to register context and verifies existence of MDC context.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -447,21 +450,42 @@ public Cancellable asyncRequest(
...
+                MDCCopyHelper mdcCopyHelper = new MDCCopyHelper();
                 executor.execute(
                     () -> {
                       InputStream is = listener.getInputStream();
                       assert ObjectReleaseTracker.track(is);
                       try {
                         NamedList<Object> body =
                             processErrorsAndResponse(solrRequest, parser, response, is);
+                        mdcCopyHelper.onBegin(null);
+                        if (log.isDebugEnabled()) {
```
dsmiley, line 466: Can you please remove all the needless "log.isDebugEnabled" checks?  It's needlessly verbose.  I suspect you are doing this because you are seeing it in many places but we only do this when there are interpolated arguments `{}` with non-simply references to objects.

dsmiley (commented):
Looking great now!

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
stillalex: I was actually following the 'local style' of the class
```
      if (log.isDebugEnabled()) {
        log.debug("Create Http2SolrClient with HTTP/1.1 transport");
      }
```
but sure, I will remove shortly

stillalex:
I think all feedback was covered

solr/CHANGES.txt
```diff
@@ -236,6 +236,8 @@ Bug Fixes
 
 * SOLR-16656: rid parameter missing from query logs (Alex Deparvu)
 
+* SOLR-16676: Http2SolrClient loss of MDC context (Alex Deparvu)
```
dsmiley, line 239: I like succinctness but here it's unclear.

```suggestion
* SOLR-16676: Logs: Http2SolrClient.async() lacked MDC (Alex Deparvu)
```

If this helps distributed search (which uses this I think) then mention that.  But I haven't noticed such a problem so maybe it doesn't?

stillalex:
thank you @dsmiley, really appreciate your help on getting this merged.

PR #1412, opened 2023-03-01
SOLR-16372: Migrate collection listing, deletion to JAX-RS

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-16372

# Description

Solr is in the process of migrating its v2 APIs over to a JAX-RS framework.  This gives us a more feature-rich framework for expressing APIs, makes it easier for the APIs themselves to be more REST-ful and standardized, and also makes it easier for Solr to integrate with tooling such as OpenAPI, which can be used to autogenerate client bindings for a variety of different languages.

But many non-JAX-RS APIs remain.

# Solution

This PR introduces a baby-step in the right direction by migrating two additional APIs to JAX-RS: the "list collections" and "delete collection" APIs.

# Tests

See `DeleteCollectionAPITest`.  Existing v2 tests continue to pass.

# Checklist


gerlowskija:
Alright, this should be ready to go.  I'll plan on merging midweek and then backporting later on pending any feedback from folks.

solr/core/src/java/org/apache/solr/jersey/SolrJacksonMapper.java
```diff
@@ -18,15 +18,44 @@
...
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import java.io.IOException;
 import javax.ws.rs.ext.ContextResolver;
 import javax.ws.rs.ext.Provider;
+import org.apache.solr.common.util.NamedList;
 
 /** Customizes the ObjectMapper settings used for serialization/deserialization in Jersey */
+@SuppressWarnings("rawtypes")
```
dsmiley, line 32: Did you try to fix this or too tricky?  These are annoying

solr/core/src/java/org/apache/solr/jersey/SolrJacksonMapper.java
```diff
@@ -18,15 +18,44 @@
...
+    }
+
+    public NamedListSerializer(Class<NamedList> nlClazz) {
+      super(nlClazz);
+    }
+
+    @Override
+    public void serialize(NamedList value, JsonGenerator gen, SerializerProvider provider)
+        throws IOException {
+      gen.writeObject(value.asShallowMap());
```
dsmiley, line 58: I suspect if benchmarks are done, this is going to be revisited to be more efficient.  That NamedList.shallowMap looks trivial but the implementations of several of its methods call NamedList.asMap which is not.  Fine for now I guess.
CC @noblepaul

solr/core/src/java/org/apache/solr/jersey/SubResponseAccumulatingJerseyResponse.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * <p>Many Solr APIs, particularly those historically reliant on overseer processing, return a
+ * response to the user that is composed in large part of the responses from all sub-requests made
+ * during the APIs execution. (e.g. the collection-deletion response itself contains the responses
+ * from the 'UNLOAD' call send to each core.) This class encapsulates those responses as possible.
+ */
+public class SubResponseAccumulatingJerseyResponse extends SolrJerseyResponse {
+
+  @JsonProperty("requestid")
+  public String requestId;
```
dsmiley, line 33: Isn't the requestId here the same as "rid" which is only used in logging?  But maybe not because it's a response JSON property?

solr/core/src/test/org/apache/solr/handler/admin/api/DeleteCollectionAPITest.java
```diff
@@ -0,0 +1,62 @@
...
+
+/** Unit tests for {@link DeleteCollectionAPI} */
+public class DeleteCollectionAPITest extends SolrTestCaseJ4 {
+
+  @Test
+  public void testConstructsValidOverseerMessage() {
+    // Only required properties provided
+    {
+      final ZkNodeProps message =
+          DeleteCollectionAPI.createRemoteMessage("someCollName", null, null);
```
dsmiley, line 40: Sigh... one day if we ever want to remove the Overseer fully, I see here an ever increasing number of tests that assume its existence.  Yeah we've spoken about this, just reading this test with a heavy sigh.

solr/core/src/java/org/apache/solr/jersey/SolrJacksonMapper.java
gerlowskija: Yeah, I looked into it a bit but couldn't quite figure out the right syntax unfortunately ☹️

solr/core/src/java/org/apache/solr/jersey/SubResponseAccumulatingJerseyResponse.java
gerlowskija: I _think_ that Solr spits out 'requestid' here because that's the name of the parameter that [v1 REQUESTSTATUS](https://solr.apache.org/guide/6_6/collections-api.html#CollectionsAPI-requeststatus) has taken, historically.

Definitely something worth cleaning up at some point.  (Personally my preference would be for something more explicit, like `asynchronous-request-id`, but admittedly that's probably too verbose for most people's tastes.)  I'd be onboard with any option that helped requestid be less overloaded.

That said - definitely beyond the scope of this ticket to make any field naming changes, unfortunately.  Changing the response format would impact v1 APIs as well, have backcompat implications, etc.

epugh: For these types of feedback, what's the best way foreward?   Should we create a JIRA for each one of these and make it be related to the current migration related JIRA, (in this case SOLR-16372)?   If so, do you want to do them as the person moving the current issue forward, or me to do it as "the kibitzing person coming up with ideas" person?  I don't want to lose these ideas improvements ;-)

PR #1413, opened 2023-03-01
SOLR-16683 Fix frequent failures of OtelTracerConfiguratorTest

janhoy:
https://issues.apache.org/jira/browse/SOLR-16683

The `@ThreadLeakLingering`  should be in millis, not seconds, so 10 means 10ms, but we should wait for 10s.

### DIFF
--- a/solr/modules/opentelemetry/src/test/org/apache/solr/opentelemetry/OtelTracerConfiguratorTest.java
+++ b/solr/modules/opentelemetry/src/test/org/apache/solr/opentelemetry/OtelTracerConfiguratorTest.java
@@ -27,7 +27,7 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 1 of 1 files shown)


risdenk (approved):

risdenk:
I haven't tried these changes, but its sad that the worker thread is just unmanaged and can't be shutdown.

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java#L85

dsmiley (approved):

janhoy:
> I haven't tried these changes, but its sad that the worker thread is just unmanaged and can't be shutdown.

Yes, they should allow passing in an executor or otherwise control it. But Hopefully the timeouts set in the test will now cause the worker to shutdown within 10s.

PR #1393, opened 2023-02-26
Update org.apache.httpcomponents

solrbot:
This PR contains the following updates:

| Package | Type | Update | Change |
|---|---|---|---|
| [org.apache.httpcomponents:httpmime](http://hc.apache.org/httpcomponents-client-ga) | dependencies | patch | `4.5.13` -> `4.5.14` |
| [org.apache.httpcomponents:httpcore](http://hc.apache.org/httpcomponents-core-ga) | dependencies | patch | `4.4.15` -> `4.4.16` |
| [org.apache.httpcomponents:httpclient](http://hc.apache.org/httpcomponents-client-ga) | dependencies | patch | `4.5.13` -> `4.5.14` |

---

### Configuration

📅 **Schedule**: Branch creation - "* * * * 0" (UTC), Automerge - At any time (no schedule defined).

🚦 **Automerge**: Disabled by config. Please merge this manually once you are satisfied.

♻ **Rebasing**: Whenever PR becomes conflicted, or you tick the rebase/retry checkbox.

👻 **Immortal**: This PR will be recreated if closed unmerged. Get [config help](https://togithub.com/renovatebot/renovate/discussions) if that's undesired.

---

 - [ ] If you want to rebase/retry this PR, check this box

---

This PR has been generated by [Renovate Bot](https://togithub.com/solrbot/renovate-github-action)

PR #1429, opened 2023-03-02
Renovate tweaks

risdenk:
* Group together Apache Calcite and Apache Calcite Avatica
* Use `matchPackagePrefixes` instead of regex
* Add google cloud to monthly checks like AWS SDK


janhoy (approved):
Looks ok. The only disadvantage of grouping this way is that the PR title and commit message will not mention the version, since there may be many different pkgs with different vers grouped together, see https://github.com/apache/solr/pull/1393

The committer doing the merge can consider adding some version info into the commit message - for more friendly changelog msgs if we automate this during release.

PR #1434, opened 2023-03-06
SOLR-16653 :  sub-shard remains inactive after split for NRT+PULL collection

noblepaul:


noblepaul:
The `force_set_state` flag was set to `false` after the replica come up and this was leading to unnecessary updates to `state.json`

PR #123, opened 2021-05-12
SOLR-12697 Add pure DocValues support to FieldValueFeature

tomglk:
# Description

This PR refers to the issue [SOLR-12697](https://issues.apache.org/jira/browse/SOLR-12697).
The problem is, that the current FieldValueFeature only works for stored fields.
This is not optimal, because using DocValues is faster for this use case. Also it increases the index size if you have to store fields only to use them for ltr.

# Solution

**Note:** This PR is based on the work of Stanislav Livotov and Christine Poerschke that can be seen in the jira ticket.

It uses the latest patch (17th May 2019) from the jira ticket as base. I combined that with suggestions from Mrs. Poerschke and my own approach to the problem.

This PR adds the DocValuesFieldValueFeatureScorer as a new Scorer used by the FieldValueFeatureWeight.
The new scorer is used whenever a field has docValues and is not stored. Therefore it does not affect the current functionality but only is applied for fields that could not be used before.
The new scorer checks the type of docValues a field has and handles NUMERIC and SORTED types. For NUMERIC fields, it simply uses the value, the SORTED type gets parsed as number or boolean-flag.

# Tests

New fields that have docValues=true were added to the schema.xml in order to test in TestLTROnSolrCloud that the feature-requests also return values for these fields.
The TestLTRReRankingPipeline was changed from a SolrTestCase  to a SolrTestCaseJ4 in order to improve readability.

I ran all tests in the package `org.apache.solr.ltr`.

# Please note

I am aware that the structure of the FieldValueFeature is now quite hard to read and the new Scorer is a bit hidden. I decided to add another nested class to the FieldValueFeatureWeight to avoid having to duplicate a lot of code just to change the inner functionality.

~~Unit tests for the handleBytesRef are still missing. I plan to add them, but wanted to create the PR already so that the general approach can be reviewed and discussed.~~
Edit: Tests for the parsing of different Types of sorted docValues (T and F flag, numbers, no number) were added in e07c432.

# Checklist

Please review the following and check all that apply:

- [X] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [X] I have created a Jira issue and added the issue ID to my pull request title. (**Issue was already present**)
- [X] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
[...]

PR #1436, opened 2023-03-07
Add solrbot dependency upgrades to CHANGES for 9.2

janhoy:
Here I move all dep upgrade changes lines to a new section, and add all the PRs filed by solrbot. These do not have JIRAs but are identified by `PR#123`.

There were multiple upgrades of same dependency in 9.2, in that case I merged them into one line and comma-separated the Jira/PR refs. If anyone feel that they want to add themselves for credit in addition to `solrbot` for any of these, feel free to do so in this PR.

I modified `changes2html.pl` to recognize this pattern and add link in HTML version of changes, see http://www.cominvent.com/pub/releasenotes.html#v9.2.0.dependency_upgrades for preview

I also modify `addVersion.py` so that the `Dependency Upgrades` section is auto generated when a new version is added.


risdenk (approved):

HoustonPutman (approved):

janhoy:
Lucene uses `GITHUB#123` for their CHANGES github links. But I like `PR#123` better as it is shorter and should be well understood.

In the `changes2html` script I added support for both `GITHUB#123`, `GH#123` and `PR#123`. I also considered a plain `#123` or `(#123)`, which is a pretty standard way of linking to github issue or PR. I really like the plain `#123` best, as it should be well understood by developers too. However, as long as we have both JIRA and GH it may be ambigous to read "number 123" and assume it is Github.

Shout out if you have strong opinions either way.

@HoustonPutman may will cause some merge conflicts for other in-flight PRs, and will also need syncing with branch_9x, which happens to be a release manager task that needs doing anyway. So if you want to take over this PR and finalize it as part of your RM task, that is OK with me.

janhoy:
Do you want me to complete this @HoustonPutman or will you merge it as part of your other rm work with CHANGES ?

HoustonPutman:
Sorry I missed this somehow, I will merge it before starting the RC!

PR #1437, opened 2023-03-07
SOLR-16487: Improve Pull Replica Interval

justinrsweeney:
https://issues.apache.org/jira/browse/SOLR-16487

# Description

The polling for replication is fairly simply, but can lead to polling too often. As an example if you had the following config for commits:
```
<autoCommit>
    <maxTime>15000</maxTime>
    <openSearcher>false</openSearcher>
</autoCommit>

<autoSoftCommit>
    <maxTime>60000</maxTime>
</autoSoftCommit>
```
The current logic would setup polling to be half of the autoCommit time, so poll every 7.5 seconds. However since a new searcher isn't opened, there will only be changes reflected every 60 seconds on the leader. We can make this logic a bit smarter knowing that the replication handler won't reflect changes until a new searcher is opened.

# Solution

Modified the code setting the polling interval for replication to only use the autoCommit time if openSearcher is true, otherwise it will use the soft commit time.

# Tests

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
@@ -80,7 +80,8 @@ public class ReplicateFromLeader {
       if (System.getProperty("jetty.testMode") != null) {
         pollIntervalStr = "00:00:01";
       }
-      if (uinfo.autoCommmitMaxTime != -1) {
+
+      if (uinfo.openSearcher && uinfo.autoCommmitMaxTime != -1) {
         pollIntervalStr = toPollIntervalStr(uinfo.autoCommmitMaxTime / 2);
       } else if (uinfo.autoSoftCommmitMaxTime != -1) {
         pollIntervalStr = toPollIntervalStr(uinfo.autoSoftCommmitMaxTime / 2);
--- a/solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
@@ -155,6 +155,13 @@ public class ReplicateFromLeader {
     }
   }
 
+  /**
+   * Determine the poll interval for replicas based on the auto soft/hard commit schedule
+   *
+   * @param uinfo the update handler info containing soft/hard commit configuration
+   * @return a poll interval string representing a cadence of polling frequency in the form of
+   *     hh:mm:ss
+   */
   public static String determinePollInterval(SolrConfig.UpdateHandlerInfo uinfo) {
     int hardCommitMaxTime = uinfo.autoCommmitMaxTime;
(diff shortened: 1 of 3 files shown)


solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
```diff
@@ -155,6 +155,38 @@ public static String getCommitVersion(SolrCore solrCore) {
...
+        // softCommit is configured.
+        // Usually if softCommit is configured, `hardCommitNewSearcher==false`, in which case you
+        // want
+        // to calculate poll interval wrt the max of hardCommitTime (when segments are available to
+        // replicate) and softCommitTime (when changes are visible).
+        // But in the unusual case that hardCommit _does_ open a new searcher and
+        // `hardCommitMaxTime < softCommitMaxTime`, then fallback to `else` clause, setting poll
+        // interval
+        // wrt `hardCommitMaxTime` alone.
+        pollIntervalStr = toPollIntervalStr((Math.max(hardCommitMaxTime, softCommitMaxTime)) / 2);
```
dsmiley, line 178: nitpick: extra parentheses in here that are not adding clarity at all

solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
```diff
@@ -155,6 +155,38 @@ public static String getCommitVersion(SolrCore solrCore) {
...
   }
 
+  public static String determinePollInterval(SolrConfig.UpdateHandlerInfo uinfo) {
+    int hardCommitMaxTime = uinfo.autoCommmitMaxTime;
+    int softCommitMaxTime = uinfo.autoSoftCommmitMaxTime;
+    boolean hardCommitNewSearcher = uinfo.openSearcher;
+    String pollIntervalStr = null;
+    if (hardCommitMaxTime != -1) {
+      // configured hardCommit places a ceiling on the interval at which new segments will be
+      // available
```
dsmiley, line 165: Code comments need to be re-flowed after "tidy" did its work.
When writing comments that flow to this many lines (kind of a lot) consider /* */ style, which I *think* Tidy reflows for you.

solr/core/src/test/org/apache/solr/cloud/ReplicateFromLeaderTest.java
```diff
@@ -0,0 +1,69 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.solr.cloud;
+
+import org.apache.solr.core.SolrConfig;
+import org.junit.Test;
+
+public class ReplicateFromLeaderTest extends SolrCloudTestCase {
```
dsmiley, line 23: Why SolrCloudTestCase?

justinrsweeney: No good reason, thanks for pointing that out.

PR #1438, opened 2023-03-08
SOLR-16690: enqueueUpdate should execute callback even if cmds is empty or noop

patsonluk:
https://issues.apache.org/jira/browse/SOLR-16690

# Description

`ZkStateWriter#enqueueUpdate(ClusterState prevState, List<ZkWriteCommand> cmds, ZkWriteCallback callback)` accepts a callback but such callback would not be called if either the cmds is empty or a no-op.

However, this could cause issues in some scenarios, for example in [Overseer$ClusterStateUpdater#run](https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/cloud/Overseer.java#L373), it calls

```
clusterState = processQueueItem(message, clusterState, zkStateWriter, true,
() ->
{ stateUpdateQueue.remove(processedNodes); processedNodes.clear(); }
);
```

Which it relies on the callback to remove items from the work queue, however if the message does not create an empty `List<ZkWriteCommand>` - for example "DOWNNODE" for PRS enabled collection, then the callback will be ignored hence the `stateUpdateQueue.remove(processedNodes);` would not be run even though it has finished processing the message. 

# Solution

Execute the callback if cmds is empty or noop

# Checklist


solr/core/src/java/org/apache/solr/cloud/overseer/ZkStateWriter.java
```diff
@@ -139,8 +139,12 @@ public ClusterState enqueueUpdate(
       throw new IllegalStateException(
           "ZkStateWriter has seen a tragic error, this instance can no longer be used");
     }
-    if (cmds.isEmpty()) return prevState;
-    if (isNoOps(cmds)) return prevState;
+    if (cmds.isEmpty() || isNoOps(cmds)) { // empty and no-ops are valid, should call callback
+      if (callback != null) {
```
hiteshk25, line 143: Thanks @patsonluk. is it possible to add unit test for this method `enqueueUpdate` ?

patsonluk: Added `ZkStateWriterTest#testEnqueueCallback`

patsonluk:
Closing this as the proposed change can introduce incorrect behavior. For example, if there's already pending changes in the `ZkStateWriter`, and a no op changes invocation to `enqueueUpdate` can invoke the callback while the actual pending changes are not flushed.

The existing behavior (shortcut no-op/empty ops and skip callback) is not ideal but is technically correct, as the `ZkWriteCallback#onWrite` method has description of `Called by ZkStateWriter if state is flushed to ZK`, since the empty/no-op operation does not trigger any state flush, it should not invoke the callback neither.

This callback might not work in certain edge case, for example if there are a mix of both PRS and non PRS enabled collection, a PRS related operation might invoke `ZkWriteCallback#onWrite` while the operations for non PRS collections are still pending. However, this is probably a rather unlikely case and is beyond the scope of this PR

PR #1440, opened 2023-03-08
SOLR-16691 : Use Jackson for JSON serialization

noblepaul:


magibney:
I could be mistaken, but iiuc most normal writes (what we've been benchmarking in the related dev@solr discussion) run through [here](https://github.com/apache/solr/blob/c99af207c761ec34812ef1cc3054eb2804b7448b/solr/core/src/java/org/apache/solr/response/QueryResponseWriterUtil.java#L51-L73).

I wonder in that case how much of the speedup we've observed from this change might simply be related to bypassing the shim `OutputStream` that only overrides `write(int byte)`. Put another way: what would be the impact of simply overriding `write(byte[] buf, int offset, int length)` on that shim `OutputStream`?

`writeStrRaw` is basically equivalent to both the `gen.writeRawValue(String)` and `gen.writeRaw(String)` methods. The distinction between `writeRawValue()` and `writeRaw()` is I think a bit of an impedance mismatch between Solr's `TextWriter` API and the Jackson `JsonGenerator` API -- the clue is in the no-op methods for writing separators in `WriterImpl` in this PR, and I think the consequence is that because we're no-opping those, we should _always_ use `gen.writeRawValue()` from the context of Solr. Related: I also note that `gen.writeRawValue(String)` says nothing about wrapping Strings in quotes -- in fact the `gen.writeRawValue()` methods don't know the type context at all, so I think that may requires a change in the else clause of `writeStr()`, to:
```java
    public void writeStr(String name, String val, boolean needsEscaping) throws IOException {
      if (needsEscaping) {
        gen.writeString(val);
      } else {
        gen.writeRawValue("\"");
        gen.writeRaw(val);
         gen.writeRaw("\"");
     }
    }
```

Fikavec:
> I don't have any means to compare the performance. 

You can run the wget command and see the transfer time, its speed and the amount of data sent.
`wget --report-speed=bits --server-response -O /dev/null --header='Accept-Encoding: ' --post-data 'indent=false&q=*%3A*&rows=1000000&wt=RESPONSE_WRITER' http://SOLR_IP:8983/solr/TEST_COLLECTION/select
`

Replace SOLR_IP, TEST_COLLECTION with your values, and run the command with various RESPONSE_WRITER (json or javabin or smile or jacksonjson). The collection documents should be large enough and the number of rows should be such that at least several gigabytes are transferred from the database. This will allow you to see the difference in speed. I tested on different collections and data, it always turned out faster from 1.7x to 7x. The disk and the network must be fast (for example, NVME disk or ram disk), otherwise the difference will not be noticeable due to long data reading from the disk or network limitations. To avoid a bottleneck in the network, the command can be executed on the Solr server by replacing the SOLR_IP with 127.0.0.1.

Fikavec (commented):
[JacksonJsonWriter.zip](https://github.com/apache/solr/files/10946072/JacksonJsonWriter.zip)
Moved JsonFactory to JacksonJsonWriter constructor and remove this.out = out - I tested it on my Solr, everything works (I commented out writeStrRaw - it was not compiled with it). Are there any plans to make an extension, a patch for Solr version 8.11.2.

Fikavec (commented):
If we create WriterImpl as AutoCloseable why we [here ](https://github.com/apache/solr/pull/1440/commits/8a5a5c687595505d804252d1291ef381ce1aff8f#diff-d2bc03d28f4721504e5b0956887dd9223765b8dee702e3e9766a56ef010afc9fL75) close gen? 
Solr has indent=true|false option in url - [gen.setPrettyPrinter](https://github.com/apache/solr/pull/1440/commits/8a5a5c687595505d804252d1291ef381ce1aff8f#diff-d2bc03d28f4721504e5b0956887dd9223765b8dee702e3e9766a56ef010afc9fR66) was always setted independently of it?

noblepaul:
@Fikavec , I'm not sure if there is a release planned for `8.11.2` . if yes, we will try to incorporate this. However, We can create a patch and make it available for anyone who wishes to use it

solr/core/src/java/org/apache/solr/response/JacksonJsonWriter.java
```diff
@@ -29,6 +29,16 @@
 import org.apache.solr.request.SolrQueryRequest;
 
 public class JacksonJsonWriter extends BinaryResponseWriter {
+
+  protected final JsonFactory jsonfactory;
```
Fikavec, line 33: `protected static final JsonFactory jsonfactory = new JsonFactory();`
    I'm sorry, maybe it would be better to create a factory right here than in the constructor (just like you did with "PrettyPrinter pretty = ")? Then the "public JacksonJsonWriter(){super()..." constructor itself can be deleted below, since it will no longer be needed.

solr/test-framework/src/java/org/apache/solr/SolrTestCase.java
```diff
@@ -183,4 +186,11 @@ public void checkSyspropForceBeforeAssumptionFailure() {
     final String PROP = "tests.force.assumption.failure.before";
     assumeFalse(PROP + " == true", systemPropertyAsBoolean(PROP, false));
   }
+
+  public static void jsonAssert(String expected, String actual) {
```
dsmiley, line 190: Without looking at the code, the method's name alone gives no hint as to what is being asserted.  Perhaps renaming to assertJsonEquals instead?  And javadocs of course.

noblepaul:
`TestRandomFlRTGCloud` is failing in this PR. @magibney  Can you give me any hint as to why it is failing ?

magibney:
I can take a closer look tomorrow, but (esp. based on the fact that you tagged me) I suspect this will be around handling of the "json" raw doc transformer ... just have to figure out how to translate the special handling of raw json values to the new context.

magibney:
@noblepaul take a look at https://github.com/apache/solr/pull/513 -- in particular, the RawShimTextResponseWriter which together with TextResponseWriter (from which it is invoked) actually handles flipping raw transformer fields over such that `TextWriter.writeStrRaw()` will be called. I think you need an analogous multiplexing (flipping between raw and non-raw) somewhere in JsonResponseWriter.

Again it's possible that any concrete performance improvement might be largely a consequence of avoiding [this](https://github.com/apache/solr/blob/c99af207c761ec34812ef1cc3054eb2804b7448b/solr/core/src/java/org/apache/solr/response/QueryResponseWriterUtil.java#L51-L73), which IMO looks really really wrong, and would be the first thing I'd change, even if pursuing using Jackson for JSON serialization anyway.

noblepaul:
looks like the  problem was [here](https://github.com/apache/solr/blob/1b06505c0f956ba6d112acec461c5946446dbdad/solr/core/src/java/org/apache/solr/response/transform/GeoTransformerFactory.java#L137)

@magibney

solr/test-framework/src/java/org/apache/solr/SolrTestCase.java
```diff
@@ -183,4 +186,11 @@ public void checkSyspropForceBeforeAssumptionFailure() {
     final String PROP = "tests.force.assumption.failure.before";
     assumeFalse(PROP + " == true", systemPropertyAsBoolean(PROP, false));
   }
+
+  public static void assertJSONEquals(String expected, String actual) {
```
dsmiley, line 190: assertJSONEquals should be capitalized as assertJsonEquals but not a big deal

PR #1442, opened 2023-03-08
Improving efficiency of replication process

justinrsweeney:
https://issues.apache.org/jira/browse/SOLR-16689

# Description

This PR resolves two inefficiencies in the replication process:
1. Avoiding repeated attempts to replicate an empty index
2. Avoiding an unnecessary commit call to leader when replicating to a PULL replica

# Solution

Fixed the check of an empty index to use commit timestamp equal to 0 as opposed to generation which starts at 1.
Added a check when a replica is recovering to only commit the leader if the recovering replica is a leader type. PULL replicas which cannot become leader, do not need to force a commit on the leader in order to recover.

# Tests

# Checklist


solr/core/src/java/org/apache/solr/cloud/RecoveryStrategy.java
```diff
@@ -223,8 +223,10 @@ private final void replicate(String nodeName, SolrCore core, ZkNodeProps leaderp
 
     log.info("Attempting to replicate from [{}].", leaderUrl);
 
-    // send commit
-    commitOnLeader(leaderUrl);
+    // send commit if replica could be a leader
+    if (replicaType == null || Replica.Type.isLeaderType(replicaType)) {
```
risdenk, line 227: minor: `Replica.Type.isLeaderType` could also check for null and make this logic clearer

PR #1443, opened 2023-03-08
SOLR-16676: Test improvements

risdenk:
https://issues.apache.org/jira/browse/SOLR-16676

### DIFF
--- a/solr/core/src/test/org/apache/solr/handler/TestHttpRequestId.java
+++ b/solr/core/src/test/org/apache/solr/handler/TestHttpRequestId.java
@@ -36,7 +36,6 @@ import org.junit.Test;
 import org.slf4j.MDC;
 
 @LogLevel("org.apache.solr.client.solrj.impl.Http2SolrClient=DEBUG")
-@SuppressForbidden(reason = "We need to use log4J2 classes directly to test MDC impacts")
 public class TestHttpRequestId extends SolrJettyTestBase {
 
   @BeforeClass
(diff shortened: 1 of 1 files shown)


solr/core/src/test/org/apache/solr/handler/TestHttpRequestId.java
```diff
@@ -126,7 +116,7 @@ public void onFailure(Throwable throwable) {
       assertEquals(3, reqLog.getQueue().size());
       while (!reqLog.getQueue().isEmpty()) {
         var reqEvent = reqLog.getQueue().poll();
-        assertTrue(reqEvent.getContextData().containsKey(key));
+        assertNotNull(reqEvent);
```
stillalex, line 119: don't think this change is correct. we are checking if the MDC actually contains the unique key. the log will always be there, so the `assertNotNull` will always hold

solr/core/src/test/org/apache/solr/handler/TestHttpRequestId.java
```diff
@@ -45,79 +44,70 @@ public static void beforeTest() throws Exception {
...
+    final String key = "mdcContextTestKey" + System.nanoTime();
+    final String value = "TestHttpRequestId" + System.nanoTime();
 
     AsyncListener<NamedList<Object>> listener =
         new AsyncListener<>() {
-
           @Override
           public void onSuccess(NamedList<Object> t) {
-            assertTrue(value, value.equals(MDC.get(key)));
+            assertEquals(value, MDC.get(key));
```
stillalex, line 83: this is a good catch, much better to have the values included in the assertion message.

solr/core/src/test/org/apache/solr/handler/TestHttpRequestId.java
risdenk: the check is that reqEvent could be null from the poll(). I'm going to push a quick addendum quick to remove the repeated `reqLog.getQueue` which then doesn't need the null check.

risdenk:
> sorry, there are a lot of cosmetic changes here. while I don't disagree with these, they are 'nice to have' rather than adding some meaningful change to understand why this might be failing. if you don't mind, could you focus on failure-fixing patches first to get the builds passing reliably, and then we can make the code clean(er) and nice(er).

I'm with you I just put this together to try to see if there were missing assertions and general cleanup. I agree it would be better to fix the underlying issue first.

solr/core/src/test/org/apache/solr/handler/TestHttpRequestId.java
risdenk: Fixed in 117960e0d521c18ce48a7f2b520a65a0eab4870b

risdenk:
combined into https://github.com/apache/solr/pull/1444

PR #1447, opened 2023-03-10
SOLR-16694: Shortcut Http2SolrClient mimeType check for default response parse

tflobbe:


solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -136,6 +136,11 @@ public class Http2SolrClient extends SolrClient {
   private long requestTimeoutMillis;
 
   protected ResponseParser parser = new BinaryResponseParser();
+  private final Set<String> defaultMimeTypes =
```
stillalex, line 139: wouldn't it be better if this happens after the parser is initialized from the builder? https://github.com/apache/solr/blob/237ff32720ac02c9d950e1aa863d0c957a0c6702/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java#L191 could cover the case there a different parser is passed in, in case that is useful.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -909,6 +889,44 @@ private NamedList<Object> processErrorsAndResponse(
...
+   */
+  private void checkContentType(
+      ResponseParser processor, InputStream is, String mimeType, String encoding, int httpStatus) {
+    if (processor == this.parser && defaultMimeTypes.contains(mimeType)) {
+      // Shortcut the default scenario
+      return;
+    }
+    final Collection<String> processorSupportedContentTypes = processor.getContentTypes();
+    if (processorSupportedContentTypes != null && !processorSupportedContentTypes.isEmpty()) {
+      final Collection<String> processorMimeTypes =
```
stillalex, line 904: I know this code was moved 'as is' but will leave the comment here in case there is interest in the proposal: what about using a lookup on the stream that can shortcut instead of building the entire set of mimeTypes. something along the lines of 
```
boolean contains =  processorSupportedContentTypes.stream()
    .map(ct -> ContentType.parse(ct).getMimeType().trim())
    .anyMatch(supported -> mimeType.equalsIgnoreCase(supported));
```
this would stop as soon as the mimeType is found. granted there are probably not many entries in that set anyway so maybe not worth the trouble

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
tflobbe: Ah, definitely, you are right.

stillalex:
(non binding) +1 the parser part looks good to me. I don't have background on the `MiniClusterState` change to `.useHttp1_1(true)`.

stillalex:
ah interesting, I missed that other one. yeah I would maybe move the MiniClusterState and benchmark to its own PR maybe, so the discuss can focus on that aspect only.

tflobbe:
> I would maybe move the MiniClusterState and benchmark to its own PR maybe, so the discuss can focus on that aspect only.

PR Here: https://github.com/apache/solr/pull/1464. PTAL

PR #1450, opened 2023-03-11
SOLR-16693: timeAllowed: use more efficient TimeLimitingBulkScorer

dsmiley:
Instead of ExitableDirectoryReader

https://issues.apache.org/jira/browse/SOLR-16693

TODO... maybe allow use of ExitableDirectoryReader via sys prop?  BTW there are code paths using the reader that wouldn't use the search method yet may have a timeAllowed param.  I think I've seen this in an erroneous (bug or accident; haven't root caused yet) scenario at work but could be valid.

Wish we had benchmarks on this.  Any way, it's interesting that all tests pass so easily :-)


mkhludnev (approved):

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -713,6 +713,33 @@ public QueryResult search(QueryResult qr, QueryCommand cmd) throws IOException {
     return qr;
   }
 
+  @Override
+  protected void search(List<LeafReaderContext> leaves, Weight weight, Collector collector)
+      throws IOException {
+    final var queryTimeout = SolrQueryTimeoutImpl.getInstance();
+    if (queryTimeout.isTimeoutEnabled() == false) {
+      // no timeout.  Pass through to subclass
```
cpoerschke, line 721:
```suggestion
      // no timeout.  Pass through to super class
```

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -713,6 +713,33 @@ public QueryResult search(QueryResult qr, QueryCommand cmd) throws IOException {
...
+      // So we need to make a new IndexSearcher instead of using "this".
+      new IndexSearcher(reader) { // cheap, actually!
+        void searchWithTimeout() throws IOException {
+          setTimeout(queryTimeout.makeLocalImpl());
+          super.search(leaves, weight, collector); // FYI protected access
+          if (timedOut()) {
+            // We might not even be using ExitableDirectoryReader yet various places in Solr check
+            //  for this exception.  Sometimes others but this seems most suitable.
+            throw new ExitableDirectoryReader.ExitingReaderException(
+                "Timed out.  Not actually via ExitableDirectoryReader");
```
cpoerschke, line 736: Maybe an alternative might be to have a `class SolrNotActuallyExitableDirectoryReader extends ExitableDirectoryReader.ExitingReaderException` approach which would be thrown here and then over time the various places could be switched over to check for `SolrNotActuallyExitableDirectoryReader` instead or to have the checks phased out.

cpoerschke (approved):

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
dsmiley: I assume you're not serious about the name SolrNotActuallyExitableDirectoryReader.  Two name proposals:
* TimeAllowedExceededException -- generic.  But despite it being generic, the reality is that there are other exceptions that are effectively similar that would still exist, so maybe give up on trying to be generic?
* TimeAllowedExceededFromScorerException -- more specific as to what makes this timeout what it is.

cpoerschke: > I assume you're not serious about the name SolrNotActuallyExitableDirectoryReader. Two name proposals: ...

Yeah, that was not meant to be a serious name suggestion but only a way to convey the `class ... extends ExitableDirectoryReader.ExitingReaderException` idea.

`TimeAllowedExceededException` vs. `TimeAllowedExceededFromScorerException` -- the latter to me seems closer to the current `ExitableDirectoryReader.ExitingReaderException` which is also pretty specific.

dsmiley: Good; I chose the latter; PR updated.
I think it'd be worthwhile for Lucene to have a base class of these exceptions that is public, therefore Lucene apps can catch one general exception no matter which approach is being taken.

dsmiley:
I plan to merge end of Monday or so.

PR #1453, opened 2023-03-13
SOLR-16391: Migrate collection creating APIs to JAX-RS

gerlowskija:

# Description

Solr is in the process of migrating its v2 APIs over to a JAX-RS framework. This gives us a more feature-rich framework for expressing APIs, makes it easier for the APIs themselves to be more REST-ful and standardized, and also makes it easier for Solr to integrate with tooling such as OpenAPI, which can be used to autogenerate client bindings for a variety of different languages.

But many non-JAX-RS APIs remain.

# Solution

This commit migrates a handful of APIs that either directly or indirectly create collections over to the JAX-RS framework.  These APIs are somewhat unrelated, but they have a large number of overlapping parameters relating to collection-creation so it made sense to migrate them together.

This PR also makes a few cosmetic changes to these v2 APIs to bring them closer into line with the more REST-ful design we're targeting for v2.  The APIs modified by this commit, and their new forms, may be found below:

* "Create Collection" POST /api/collections {...}
* "Create Alias" POST /api/aliases {...}

In order to keep the v2 API in line with any potential v1 changes, this commit moves the actual API logic for collection-creation over to the v2 definition.

# Tests 

Still needs unit tests and there are a number of existing tests failing.

# Checklist


gerlowskija:
As this PR stands today, the new API and refactoring appear to work correctly for v1 requests, but actually fails for v2 requests.

The root of the problem is that multiple APIs now claim `POST /api/collections` as their endpoint.  (See the v2 backup, restore, create alias, set-alias-property, and delete alias APIs).

This wasn't a problem while the APIs all shared the legacy framework, as the legacy framework is designed to inspect a "command" indicator at the top level of the request body (e.g. `{"delete-alias": {...}}).  But moving "create" over to JAX-RS while the other APIs remain on legacy proved to be a problem: V2HttpCall sends incoming "create" requests to the legacy framework, which no longer handles the "create" command, so an error is thrown.

This isn't a problem long-term, but we won't be able to proceed here until the other `POST /collections` APIs are moved over to JAX-RS.

gerlowskija:
Towards the end of last week, I was able to add the "restore collection" API to this PR, which is the last of the "collection-creating" APIs I'd hoped to do in this PR.

In hindsight though, it's grown a little larger than I was expecting and I'm a little leery of merging as is: both for the onerous-ness of review, and because the size would make any test-noise harder to pinpoint.

To this end, I filed a PR this morning that gives the create-collection, create-alias, and restore-collection v2 APIs their own (unique) REST-ful paths.  Splitting them up to be on different paths removes the blocker I mentioned [in my previous comment](https://github.com/apache/solr/pull/1453#issuecomment-1466671301), and will allow me to split this PR into 3 separate PRs (one for each API).  I'll aim to merge those individually, and close this PR without merging.

The first of these single-API PRs, for create-collection, can be found [here](https://github.com/apache/solr/pull/1453#issuecomment-1466671301).  Hoping to have single-API PRs for Create-Alias and Restore shortly as well!

PR #1459, opened 2023-03-14
SOLR-16393 Alias properties management

stillalex:
https://issues.apache.org/jira/browse/SOLR-16393

# Description

Alias properties management

* Update properties on alias

```
curl -X PUT http://localhost:8983/api/aliases/t0/properties -H 'Content-Type: application/json' -d '
{ 
  "properties": {"foo":"bar"} 
}'
```

* Get all properties on alias

```
curl -X GET http://localhost:8983/api/aliases/t0/properties
{
  "responseHeader": {
    "status": 0,
    "QTime": 1
  },
  "properties": {
    "foo": "bar"
  }
}
```

* Get specific property value

```
curl -X GET http://localhost:8983/api/aliases/t0/properties/foo
{
  "responseHeader": {
    "status": 0,
    "QTime": 2
  },
  "property": "foo",
  "value": "bar"
}
```

* Add property

```
curl -X PUT http://localhost:8983/api/aliases/t0/properties -H 'Content-Type: application/json' -d '{ "properties": {"abc":"fgh"} }'
```

* Update property by PUT

```
curl -X PUT http://localhost:8983/api/aliases/t0/properties -H 'Content-Type: application/json' -d '{ "properties": {"abc":"rty"} }'
```

* Remove property by setting to empty

```
curl -X PUT http://localhost:8983/api/aliases/t0/properties -H 'Content-Type: application/json' -d '{ "properties": {"abc":""} }'
```

* Update property via PUT to property api

```
curl -X PUT http://localhost:8983/api/aliases/t0/properties/abc -H 'Content-Type: application/json' -d '{ "value": "update0" }'
```

* Delete property via DELETE to property api

```
curl -X DELETE http://localhost:8983/api/aliases/t0/properties/abc -H 'Content-Type: application/json'
```

# Solution

# Tests

# Checklist

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -45,6 +45,12 @@ Improvements
 * SOLR-16393: The path of the v2 "delete alias" API has been tweaked slightly to be more intuitive, and is now available at
   `DELETE /api/aliases/aliasName`. (Jason Gerlowski)
 
+* SOLR-16393: Solr's v2 "aliasprop" CRUD APIs have been tweaked slightly to be more intuitive.  Alias property modification
+  is now available at `PUT /api/aliases/aliasName/properties` (for bulk modification) and `PUT /api/aliases/aliasName/properties/propName`
+  (for single property updates).  Additionally new APIs have been added for listing properties (`GET /api/aliases/aliasName/properties`),
+  fetching single property values (`GET /api/aliases/aliasName/properties/propName`), and property deletion
+  (`DELETE /api/aliases/aliasName/properties/propName`). (Alex Deparvu via Jason Gerlowski)
+
 Bug Fixes
 ---------------------
 


solr/core/src/java/org/apache/solr/handler/admin/api/AliasPropertyAPI.java
```diff
@@ -0,0 +1,262 @@
...
+  @Operation(
+      summary = "Update properties for a collection alias.",
+      tags = {"aliases"})
+  public SolrJerseyResponse updateAliasProperties(
+      @Parameter(description = "Alias Name") @PathParam("aliasName") String aliasName,
+      @RequestBody(description = "Properties that need to be updated", required = true)
+          UpdateAliasPropertiesRequestBody requestBody)
+      throws Exception {
+
+    if (requestBody == null) {
```
stillalex, line 133: @gerlowskija leaving a comment for you here. I am fuzzy on how enforcing the not-null param works here. I realized the annotations with 'required=true' are just for docs, the actual null check is not performed by the framework. checked some of the other apis and I could not find details. please let me know your thoughts.

solr/core/src/test/org/apache/solr/cloud/AliasIntegrationTest.java
```diff
@@ -291,20 +316,25 @@ public void testModifyPropertiesCAR() throws Exception {
...
-    checkFooAndBarMeta(aliasName, zkStateReader);
+    checkFooAndBarMeta(aliasName, zkStateReader, "baz", "bam");
 
     // now verify we can delete
     setAliasProperty = CollectionAdminRequest.setAliasProperty(aliasName);
     setAliasProperty.addProperty("foo", "");
     setAliasProperty.process(cluster.getSolrClient());
+    checkFooAndBarMeta(aliasName, zkStateReader, null, "bam");
+
+    // TODO is this a bug? should "bar" be null after the following
```
stillalex, line 327: @gerlowskija this might be a bug? I added some assertions to this older test and results are a bit surprising. my understanding is that setting value to null should remove it, but it doesn't. maybe I am reading it wrong.

solr/core/src/java/org/apache/solr/handler/admin/api/AliasPropertyAPI.java
gerlowskija: Good question!

You're right that any constraints added to the `@RequestBody` annotation are only considered in the docs and generated OpenAPI spec.

Jersey does offer a way to validate requests based on constraint annotations, but we're not using it currently.  They call it "bean validation".  Basically you add an extra gradle dep on `org.glassfish.jersey.ext:jersey-bean-validation`, enable validation in your Jersey apps with a specific property (`jersey.config.beanValidation.enableOutputValidationErrorEntity.server`), and add any constraint annotations (like `@NotNull`) that you need.  You can even define your own annotations which is kindof cool.

I think that we'll eventually want to utilize this everywhere we can, but right now it's hard to take advantage of it due to how the v1 code explicitly calls the Jersey classes and methods internally.  The v1 code isn't using the Jersey framework, so it can't take advantage of any bean validation, so any input checking still needs to happen explicitly as you've done here.

At least, unless there's some other option I'm missing?  Maybe we should enable bean-validation for the benefit of those v2 APIs that don't have v1 counterparts (like the list-aliases you've added in this PR)?

> checked some of the other apis and I could not find details

This is probably just looseness on my part.  I'm probably not being as rigorous in input-validation as I should be 😬

solr/core/src/test/org/apache/solr/cloud/AliasIntegrationTest.java
gerlowskija: I'm not sure whether it's a bug or not, but I agree that it's very surprising.  It seems like empty-string is the only way to delete alias properties.  Might be worth a thread on dev@ to see whether anyone knows if this is intentional or not.

solr/solr-ref-guide/modules/deployment-guide/pages/alias-management.adoc
```diff
@@ -657,7 +657,7 @@ If a key is set with a value that is empty it will be removed.
 
 [source,bash]
 ----
-http://localhost:8983/admin/collections?action=ALIASPROP&name=techproducts_alias&property.foo=bar
+curl -X POST 'http://localhost:8983/admin/collections?action=ALIASPROP&name=techproducts_alias&property.foo=bar'
```
gerlowskija, line 660: [-1] Would you be willing to add some coverage here for the new v2 APIs that you added (alias-prop listing, single-alias-prop fetching, etc.)

Would really help people use the alias prop functionality - I think right now the ref-guide suggests that people look in ZK if they need to know alias prop values!

gerlowskija (changes requested):
Looks good generally Alex - thanks again for the PR.  Still missing a few docs, if you're willing to add them, and then I think this will be good to go.

solr/core/src/java/org/apache/solr/handler/admin/api/AliasPropertyAPI.java
stillalex: I can't tell which way you are leaning here. are you suggesting I go for the new dependency and enable it for this api, or are you ok with leaving this as is and having a different PR for enabling annotation driven input validation for all existing apis that might need it? for me it works either way.

solr/core/src/test/org/apache/solr/cloud/AliasIntegrationTest.java
stillalex: email sent to dev list

solr/core/src/java/org/apache/solr/handler/admin/api/AliasPropertyAPI.java
```diff
@@ -0,0 +1,262 @@
...
+      @Parameter(description = "Property Name") @PathParam("propName") String propName)
+      throws Exception {
+    recordCollectionForLogAndTracing(null, solrQueryRequest);
+
+    final GetAliasPropertyResponse response =
+        instantiateJerseyResponse(GetAliasPropertyResponse.class);
+    final Aliases aliases = readAliasesFromZk();
+    if (aliases != null) {
+      response.property = propName;
+      response.value = aliases.getCollectionAliasProperties(aliasName).get(propName);
```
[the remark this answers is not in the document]
stillalex: sounds good. will add 404 if property is missing

solr/solr-ref-guide/modules/deployment-guide/pages/alias-management.adoc
stillalex: not sure I follow, could you clarify a bit?

I have alias prop listing https://github.com/apache/solr/pull/1459/files#diff-2cc6e0ae57c6bf17247dd5a4386f98b3123544cb816a12fdb864068705970f00R670 and property level get https://github.com/apache/solr/pull/1459/files#diff-2cc6e0ae57c6bf17247dd5a4386f98b3123544cb816a12fdb864068705970f00R685

stillalex:
to answer the questions you posted:


> Does this PR change any of the API responses (for those APIs which preexisted this PR? It doesn't look like it (which is good), but just wanted to double check.

hmm. I did not touch the v1 apis, but I did remove the v2 collections support. let me know if this was not the right decision

> I'm a little ambivalent about the response for GET /api/aliases/aName/properties/foo (which looks to be an API that's brand new to this PR?). Do we need to include "property": "foo" - presumably the user already knows which replica property they requested? I don't feel strongly in either direction here, just wondering aloud...

I think I agree to remove it. you are right, doesn't feel like it's adding too much value

> Does the v2 code need to support "deleting" a property via empty-string? It legitimately may because of the way that the v1 code is now structured to call into v2. But it'd be a nice thing to avoid bringing into v2, since v2 has an explicit DELETE API for that purpose now.

I'm using much of the preexisting code, it might not be possible to change the apis to this level. plus if you are using the bulk property update, there would be no way to remove more properties in one go.

solr/solr-ref-guide/modules/deployment-guide/pages/alias-management.adoc
stillalex: will add a property listing section, let's see how it looks

stillalex:
@gerlowskija PR updated based on feedback. please take another look. thanks!

stillalex:
fyi there were some merge errors on crave side, so I merged main branch again in the hope this gets unstuck

solr/core/src/java/org/apache/solr/handler/admin/api/AliasPropertyAPI.java
gerlowskija: Sorry for being unclear.  I was attempting to convey that there _are_ mechanisms in Jersey to enforce this, but that we shouldn't use those mechanisms (yet).

What you have here is perfect for now 👍

gerlowskija (approved):
LGTM - thanks for another PR Alex!

PR #1460, opened 2023-03-14
SOLR-16701: Race condition on PRS enabled collection deletion

patsonluk:
https://issues.apache.org/jira/browse/SOLR-16701

# Description

This fixes a race condition on PRS enabled collection deletion, which triggers the exception:
```
org.apache.solr.common.SolrException: Error fetching per-replica states
    at __randomizedtesting.SeedInfo.seed([C2BFFBF8FE49C1E1:F1C8D9E308D2745]:0)
    at app//org.apache.solr.common.cloud.PerReplicaStatesFetcher.fetch(PerReplicaStatesFetcher.java:49)
    at app//org.apache.solr.common.cloud.PerReplicaStatesFetcher$LazyPrsSupplier.lambda$new$0(PerReplicaStatesFetcher.java:62)
    at app//org.apache.solr.common.cloud.DocCollection$PrsSupplier.get(DocCollection.java:515)
    at app//org.apache.solr.common.cloud.Replica.isLeader(Replica.java:314)
    at app//org.apache.solr.common.cloud.Slice.findLeader(Slice.java:242)
    at app//org.apache.solr.common.cloud.Slice.setPrsSupplier(Slice.java:56)
    at app//org.apache.solr.common.cloud.DocCollection.<init>(DocCollection.java:123)
    at app//org.apache.solr.common.cloud.ClusterState.collectionFromObjects(ClusterState.java:305)
    at app//org.apache.solr.common.cloud.ClusterState.createFromCollectionMap(ClusterState.java:254)
    at app//org.apache.solr.client.solrj.impl.ZkClientClusterStateProvider.createFromJsonSupportingLegacyConfigName(ZkClientClusterStateProvider.java:117)
    at app//org.apache.solr.common.cloud.ZkStateReader.fetchCollectionState(ZkStateReader.java:1695)
```

This could be triggered by:
1. `fetchCollectionState` is called, and the state.json is fetched
2. But before the `fetchCollectionState` fetches the PRS entries, the collection state.json/PRS are deleted by someone else
3. `fetchCollectionState` would throw below exception when it reaches the PRS fetching logic as the Zk node state.json is no longer around

# Solution

Create a specific exception `PrsZkNodeNotFoundException` (that extends `SolrException`) when the PRS entries cannot be fetched. Then in `ZkStateReader#fetchCollectionState`, we have added a new catch clause for this exception, we will use the similar check as the handling for `NoNodeException`, but in this case if exists is null, then it's the expected race condition (prs entries deleted after state.json is fetched), and  `fetchCollectionState` should just return null (collection gone), otherwise it will rethrow the `PrsZkNodeNotFoundException` case (unexpected, PRS entries gone but state.json is still around)

# Tests

Added `ZkStateReaderTest#testDeletePrsCollection` which reproduce such race condition, and verify that:
1. The `ZkStateReader#fetchCollectionState` should not throw exception, instead, it should eventually return `null` which indicates the collection is deleted
2. The `PrsZkNodeNotFoundException` was indeed triggered

Please take note that the test case was built on the `Breakpoint` introduced by another PR https://github.com/apache/solr/pull/1457

# Checklist

### DIFF
--- a/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
+++ b/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
@@ -1628,7 +1628,8 @@ public class ZkStateReader implements SolrCloseable {
         }
         return null;
       } catch (PerReplicaStatesOps.PrsZkNodeNotFoundException e) {
-        assert CommonTestInjection.injectBreakpoint(ZkStateReader.class.getName() + "/exercised", e);
+        assert CommonTestInjection.injectBreakpoint(
+            ZkStateReader.class.getName() + "/exercised", e);
         // could be a race condition that state.json and PRS entries are deleted between the
         // state.json fetch and PRS entry fetch
         Stat exists = zkClient.exists(collectionPath, watcher, true);
--- a/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
+++ b/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
@@ -1629,15 +1629,19 @@ public class ZkStateReader implements SolrCloseable {
         return null;
       } catch (PerReplicaStatesFetcher.PrsZkNodeNotFoundException e) {
         CommonTestInjection.injectBreakpoint(ZkStateReader.class.getName() + "/exercised", e);
-        //could be a race condition that state.json and PRS entries are deleted between the state.json fetch and PRS entry fetch
+        // could be a race condition that state.json and PRS entries are deleted between the
+        // state.json fetch and PRS entry fetch
         Stat exists = zkClient.exists(collectionPath, watcher, true);
         if (exists == null) {
           if (log.isInfoEnabled()) {
-            log.info("PRS entry for collection " + coll + " not found in ZK. It was probably deleted between state.json read and PRS entry read.");
+            log.info(
+                "PRS entry for collection "
+                    + coll
+                    + " not found in ZK. It was probably deleted between state.json read and PRS entry read.");
           }
           return null;
         } else {
-          throw e; //unexpected, PRS node not found but the collection state.json still exists
+          throw e; // unexpected, PRS node not found but the collection state.json still exists
         }
       }
     }
--- a/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
+++ b/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
@@ -1616,7 +1616,8 @@ public class ZkStateReader implements SolrCloseable {
 
         ClusterState.CollectionRef collectionRef = state.getCollectionStates().get(coll);
         return collectionRef == null ? null : collectionRef.get();
-      } catch (KeeperException.NoNodeException e) {
+      } catch (KeeperException.NoNodeException | PerReplicaStatesFetcher.PrsZkNodeNotFoundException e) {
+        assert CommonTestInjection.injectBreakpoint(ZkStateReader.class.getName() + "/exercised", e);
         if (watcher != null) {
           // Leave an exists watch in place in case a state.json is created later.
           Stat exists = zkClient.exists(collectionPath, watcher, true);
--- a/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
+++ b/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
@@ -1616,10 +1616,7 @@ public class ZkStateReader implements SolrCloseable {
 
         ClusterState.CollectionRef collectionRef = state.getCollectionStates().get(coll);
         return collectionRef == null ? null : collectionRef.get();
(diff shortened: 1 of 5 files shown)


solr/solrj/src/java/org/apache/solr/common/util/CommonTestInjection.java
```diff
@@ -73,4 +76,68 @@ public static boolean injectDelay() {
...
+   * <p>This should always be a part of an assert statement (ie assert injectBreakpoint(key)) such
+   * that it will be skipped for normal code execution
+   *
+   * @see CommonTestInjection#setBreakpoint(String, Breakpoint)
+   * @param key could simply be the fully qualified class name or more granular like class name +
+   *     other id (such as method name). This should only be set by corresponding unit test cases
+   *     with CommonTestInjection#setBreakpoint
+   * @param args optional arguments list to be passed to the Breakpoint
+   */
+  public static boolean injectBreakpoint(String key, Object... args) {
```
magibney, line 126: The way this is used currently, we're passing a single arg (the exception). For a utility class like this I guess I'd slightly prefer to have the signature be a _single_ arg, esp. since the way this will _ever_ be used, afaict, would involve close coordination between caller and breakpoint-setter. Basically: if an array needs to be passed, the calling code can just construct an array. I don't think varargs gets us anything much here. That said, I don't feel too strongly about it either way.

magibney (changes requested):

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -1627,6 +1627,20 @@ private DocCollection fetchCollectionState(String coll, Watcher watcher)
           }
         }
         return null;
+      } catch (PerReplicaStatesOps.PrsZkNodeNotFoundException e) {
+        CommonTestInjection.injectBreakpoint(ZkStateReader.class.getName() + "/exercised", e);
```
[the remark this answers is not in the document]
patsonluk: good catch thanks!

solr/solrj/src/java/org/apache/solr/common/util/CommonTestInjection.java
patsonluk: I do personally like the flexibilty of the varargs, especially for most cases when there's no need for any args, then the invocation can just be `injectBreakpoint("mykey")` instead of `injectBreakpoint("mykey", null)`

magibney: That makes sense; I'm fine with that.

magibney (approved):

solr/solrj/src/java/org/apache/solr/common/util/CommonTestInjection.java
```diff
@@ -129,15 +111,61 @@ public static boolean injectBreakpoint(String key, Object... args) {
       log.info("Breakpoint with key {} is triggered", key);
       breakpoint.executeAndResume(args);
       log.info("Breakpoint with key {} was executed and normal code execution resumes", key);
+    } else {
+      log.info(
+          "Breakpoint with key {} is triggered but there's no implementation set. Skipping...",
+          key);
```
magibney, line 117: I don't think we should log here (and fwiw there is no corresponding logging for absent `injectDelay()`. The issue is that if you were to enable assertions on a running system (not in a test context), you might get a ton of spam from this. If you want to leave it, maybe set log level from `log.info()` to `log.debug()`?

solr/solrj/src/java/org/apache/solr/common/util/CommonTestInjection.java
```diff
@@ -129,15 +111,61 @@ public static boolean injectBreakpoint(String key, Object... args) {
...
+     *
+     * @see CommonTestInjection#injectBreakpoint(String, Object...)
+     * @param key could simply be the fully qualified class name or more granular like class name +
+     *     other id (such as method name). This should batch the key used in injectBreakpoint
+     * @param implementation The Breakpoint implementation
+     */
+    public void setImplementation(String key, Breakpoint implementation) {
+      if (breakpoints.containsKey(key)) {
+        throw new IllegalArgumentException(
+            "Cannot refine Breakpoint implementation with key " + key);
```
magibney, line 159: nit: "modify" or "replace" rather than "refine"? We'd be replacing, not refining.

solr/solrj/src/java/org/apache/solr/common/util/CommonTestInjection.java
patsonluk: yes let's change that to debug, i was a bit concerned if someone accidentally used different key names meant for the same Breakpoint for `injectBreakpoint` and `setImplementation`, so with this log message they can see whether the breakpoint is triggered.

Though I guess this could be really noisy as you pointed out, and I assume the dev can search whether there's "Breakpoint with key ... is triggered" message instead and absence of that means either the execution path does not reach the breakpoint or the breakpoint has no implementation set

solr/solrj/src/java/org/apache/solr/common/util/CommonTestInjection.java
patsonluk: ahh was a typo, supposed to be "redefine"

PR #1464, opened 2023-03-16
SOLR-16704: Add simple search benchmark to benchmark module

tflobbe:
The test is very simple, and it's goal (so far at least) is to test the client library more than the server


stillalex:
looks good! if it helps in any way, I tried some of the other benchmarks with 16 threads and I was able to repro the jetty ioexception on my local.

risdenk (approved):

PR #1469, opened 2023-03-17
SOLR-16706: Reduce the ZLibCompressor compression buffer size.

bruno-roustant:
Currently ZLibCompressor.compress() works with a compression buffer initialized with a capacity equal to the source data (to compress) length.
Since we expect a good compression ratio, specially for state.json, we propose a small adjustment to have the initial capacity of the buffer configurable. That way, when the caller knows it is compressing a state.json data, it can set the initial buffer capacity to sourceData.length / 10.

This PR adds ZLibCompressor.compressBytes(byte[] data, int initialBufferCapacity)

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/overseer/ZkStateWriter.java
+++ b/solr/core/src/java/org/apache/solr/cloud/overseer/ZkStateWriter.java
@@ -286,7 +286,8 @@ public class ZkStateWriter {
           } else {
             byte[] data = Utils.toJSON(singletonMap(c.getName(), c));
             if (minStateByteLenForCompression > -1 && data.length > minStateByteLenForCompression) {
-              data = compressor.compressBytes(data);
+              // When compressing state.json, we expect at least a 10:1 compression ratio.
+              data = compressor.compressBytes(data, data.length / 10);
             }
             if (reader.getZkClient().exists(path, true)) {
               if (log.isDebugEnabled()) {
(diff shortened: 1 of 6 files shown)


solr/core/src/java/org/apache/solr/cloud/overseer/ZkStateWriter.java
```diff
@@ -286,7 +286,8 @@ public ClusterState writePendingUpdates(
           } else {
             byte[] data = Utils.toJSON(singletonMap(c.getName(), c));
             if (minStateByteLenForCompression > -1 && data.length > minStateByteLenForCompression) {
-              data = compressor.compressBytes(data);
+              // When compressing state.json, we expect at least a 10:1 compression ratio.
+              data = compressor.compressBytes(data, data.length / 10);
```
justinrsweeney, line 290: Is this expectation based on a test? 

The intention here was that by doing it as the same length and uncompressed data, we consume a bit more memory initially, but save on having to potentially resize the array. The ZLib docs seem to indicate an expected compression ratio of between 2:1 and 5:1. Would we be better off erring on the side of less compression and save resizing?

bruno-roustant: The json format is quite verbose, and state.json even more.
I tested with some state.json, including the one in ZkStateWriterTest.testSingleExternalCollectionCompressedState() which triggers the compression at 500KB. I got a compression ratio of 20:1. That's why I chose 10:1 here to have a high probability of not resizing the array, and saving significant memory at the same time.

justinrsweeney: Nice, that sounds good. I can see how state.json would get a particularly high compression ratio. glad it is backed up by the test. Thanks for the improvement!

justinrsweeney (approved):

PR #1477, opened 2023-03-20
SOLR-16712: Simplify DocCollection ctor for PRS enabled collection

patsonluk:
https://issues.apache.org/jira/browse/SOLR-16712

# Description

The current implementation of PRS requires an extra param to the DocCollection, the `PrsSupplier`, when `get` is called, would fetch the PRS states from ZK. The implementation of such supplier `PrsSupplier/LazyPrsSupplier` would only fetch the state on first call. 

While this flow does work properly, this flow might introduce some unnecessary complexity:
1. PRS entry fetching from ZK is done either during or after the `DocCollection` construction, this could be a bit inconsistent with existing non PRS `DocCollection` design which `DocCollection` is simply a immutable container that does not fetch data after its instantiation
2. The lazy fetching could introduce some uncertainties as to when exactly the fetching happens (and if any Zookeeper IO exceptions arises) . Currently it's fetching via this call chain `DocCollection ctor -> Slice#setPrsSupplier -> Slice#findLeader() -> Replica#isLeader() -> PrsSuppler#get`
 
My guess was that the lazy loading was introduced in https://issues.apache.org/jira/browse/SOLR-16580 as to avoid fetching the PRS states multiple times in the ctor of `DocCollection`, however, if we only fetch the `PerReplicaStates` once calling the `DocCollection` ctor, and pass the `PerReplicaStates` object (as a wrapped reference object) to the `DocCollection` instead, it can probably achieve similar result but with reduced uncertainty after `DocCollection` construction.

There's another branch which experimented with making DocCollection, Slice and Replica immutable as well for PRS enabled collection [https://github.com/cowpaths/fullstory-solr/pull/84] but is beyond the discussion of this

# Solution
Centralize the Prs state fetching to a new builder method `buildDocCollection` in `DocCollection` and accept the identical signature as the `DocCollection` before this change (which accepts `PrsSupplier`). Such that all existing methods that invoke the old signature can call such builder method instead. 

This new method will always fetch the PRS states once before constructing the `DocCollection` object. Once the states are fetched, the rest of the instantiation of `DocCollection`, `Slice` and `Replica` will no longer trigger any fetch as we no longer pass down the supplier.

The existing `PrsSupplier` serves 2 purposes:
1. Wrap the actual suppler provided in the ctor argument, and fetch the PerReplicaStates from the wrapped supplier when needed
2. Cache and store the state of `PerReplicaStates` as a field, act as a container of `PerReplicaStates` value. Expose such field so that `DocCollection#setReplicaStates` can mutate it. Such mutation will then be visible to all other classes (Slice/Replica) that holds the same reference of such supplier.

I'm proposing to modify the `PrsSupplier` such that:
1. It becomes pure Supplier that provides `PerReplicaStates`, no wrapping necessary. 
2. No longer keeps the state of `PerReplicaStates`. Classes such as `DocCollection/Slice/Replica` will keep a reference of `AtomicReference<PerReplicaStates>` instead of a shared reference of `PrsSupplier`. This might improve code readability as a shared reference is made obvious here.

Take note that `DocCollection#copyWith(PerReplicaStates)` has also been renamed to `DocCollection#setReplicaStates(PerReplicaStates)`, this is to better describes that such method simply set the `PerReplicaStates` but does not "copy" (ie creates new `DocCollection` instance).

# Tests

No new test cases are introduced, all the existing unit test cases should provide sufficient coverage

# Checklist


noblepaul (changes requested):
Unnecessary Object creation

patsonluk:
> Unnecessary Object creation

@noblepaul AFAIK, this PR does not introduce any extra object creation? Would you mind to elaborate a bit more about your concern? 😊

noblepaul:
@patsonluk 
yes, the latest one doesn't create new Objects, I shall review it

patsonluk:
ping @noblepaul do u have a chance to take a look? Please lemme know if u have any questions/concerns. thanks again!

noblepaul:
> ping @noblepaul do u have a chance to take a look? Please lemme know if u have any questions/concerns. thanks again!
 
@patsonluk 
I had posted my comments above

patsonluk:
@noblepaul would u mind to link it again please? i tried the only link i found https://github.com/apache/solr/pull/1477/files/6f682a2febf9e5e4a03b161f61615115d3aea60d and it did not show any comments, perhaps they were not saved?

noblepaul:
@patsonluk 
as it is mutating an existing Object, there is a small period during which the state of a given replica is unpredictable . some replicas of the collection would fetch state from old PRS and some will point to new PRS

The question is, what did we achieve with this change?

patsonluk:
> @patsonluk as it is mutating an existing Object, there is a small period during which the state of a given replica is unpredictable . some replicas of the collection would fetch state from old PRS and some will point to new PRS
> 
> The question is, what did we achieve with this change?

Thank you for the comments/questions @noblepaul!

Yes you are correct that mutating the prs in  `DocCollection#copyWith(PerReplicaStates) with `this.prsSupplier.prs = newPerReplicaStates` would set the actual prs instance in one place! TBH, i did find such design a bit hard to trace when the PRS state mutates in `Replica` as such state is kept outside of `Replica` and modified in `DocCollection`. 

I guess that is a reasonable trade-off since `DocCollection/Slice/Replica` are mutable for PRS enabled collection and if we want to keep using the same instances for PRS updates, this is likely the best we could do, so let's just keep your existing implementation! 😊 

The goal of this PR is make the loading/fetching of PRS states a bit more deterministic - ie, always fetch before the construction of  `DocCollection` instead of relying on implementation. Currently it's `DocCollection ctor -> Slice#setPrsSupplier -> Slice#findLeader() -> Replica#isLeader() -> PrsSuppler#get`

I will see if we can achieve that with the Supplier still in place for `DocCollection/Slice/Replica`.

Will ping you again once the change is made. Thank you again!

patsonluk:
@noblepaul I have pushed the updates, it's using `AtomicReference<PerReplicaStates>` very similar to the `PrsSupplier`, but it does not fetch. The fetching is still done by the `PrsSupplier` but is only called before the `DocCollection` is constructed.

Let me know your thoughts thanks!

noblepaul:
This change looks good now. There are no outstanding concerns now.

noblepaul (approved):

PR #1479, opened 2023-03-22
SOLR-16465: Replace Login w/ React App

MarcusSorealheis:
[SOLR-16465](https://issues.apache.org/jira/browse/SOLR-16465)

# Description

This PR will replace the login screen functionality with new, modern technologies.

# Solution

TBD

# Tests

TBD 

# Checklist


MarcusSorealheis:
As is the nature of these things, I need to blow away a lot of the work to take a different approach. As I mentioned before, it had been a while since I dusted off my JS chops. Looks like Create React Aoo has died after many years being the scaffolding leader. I'm going to take a new approach and I'm sure developers will be enthusiastic to work with the new framework and approach I will push up in a bit.

MarcusSorealheis:
> Please reach out if you need insight into current Admin UI. As you may know, Solr backend is 100% stateless, so there is no "login" as such. The current Admin UI will do the auth, caching Authorization header in browser storage, and pass in on all subsequent reqeusts to Solr backend.

Will do for sure probably beginning this week as I think I've done sufficient research to have an opinion on the best approach. Haven't pushed it up yet but will very soon. I am aware of the backend being stateless.

> Question: Is the goal still to host the entire Admin UI, including old non-ported angular-pages, from the new "modern" Admin App? Or do you plan to have two independent Admin apps and send users between them?

I think the best approach is to host both the non-ported and ported UI's at the same time for a while. Given the scope and complexity, it will take some time for us to port over. I think the best path is to chip away at it here over the next 9 months until it is entirely ported over. Over time, I expecting porting to accelerate dramatically, and that estimate could be way overblown. There's also the fact that part of the decision I've chosen the framework that I have chosen is that it is extremely modern, seeing a lot of investment, and makes many strong choices about design. 

I'm detailing the approach in a document for the next community meeting where people can review and comment, in addition to pushing up the example here shortly. The first example I started with was fine, but the framework and approach that was very popular for 6+ years is now obsolete. This approach is newer and should be good for another 6-10 years, which will be vital for preventing the situation that hit us more recently.

epugh:
We've got an active effort going on with the new Kotlin based "Solr UI", so going to close this.

PR #1481, opened 2023-03-22
SOLR-16707: Avoid "Recursive update" ISE in non-async cache `computeIfAbsent()`

magibney:
an alternative to #1480 
See: SOLR-16707

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/CaffeineCache.java
+++ b/solr/core/src/java/org/apache/solr/search/CaffeineCache.java
@@ -264,6 +264,11 @@ public class CaffeineCache<K, V> extends SolrCacheBase
     if (computed == null) {
       return null;
     }
+    // using get-then-put, by the time the below call completes, we will have
+    // double-incremented `lookups` as internally tracked by CaffeineCache
+    // `stats.requestCount()` (considering the initial call to `getIfPresent()`);
+    // so decrement `lookups` here to compensate.
+    lookups.decrement();
     return cache.get(
         key,
         k -> {
@@ -271,10 +276,6 @@ public class CaffeineCache<K, V> extends SolrCacheBase
           // singleton return values and consistent updates to `inserts` and `ramBytes`.
           recordRamBytes(key, null, computed);
           inserts.increment();
-          // using get-then-put, we will have double-incremented `lookups` as internally
-          // tracked by CaffeineCache `stats.requestCount()` (here and in the initial
-          // call to `getIfPresent()`); so decrement `lookups` here to compensate.
-          lookups.decrement();
           return computed;
         });
   }
--- a/solr/core/src/java/org/apache/solr/search/CaffeineCache.java
+++ b/solr/core/src/java/org/apache/solr/search/CaffeineCache.java
@@ -25,6 +25,7 @@ import com.github.benmanes.caffeine.cache.RemovalListener;
 import com.github.benmanes.caffeine.cache.stats.CacheStats;
 import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.lang.invoke.MethodHandles;
 import java.time.Duration;
 import java.util.Collections;
@@ -97,6 +98,13 @@ public class CaffeineCache<K, V> extends SolrCacheBase
   private int maxIdleTimeSec;
   private boolean cleanupThread;
   private boolean async;
+  private final ThreadLocal<boolean[]> detectRecursion =
+      new ThreadLocal<>() {
+        @Override
+        protected boolean[] initialValue() {
+          return new boolean[] {false};
+        }
+      };
 
   private MetricsMap cacheMap;
   private SolrMetricsContext solrMetricsContext;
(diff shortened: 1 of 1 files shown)


risdenk (approved):

magibney:
I'm thinking this warrants a "bugfix" CHANGES.txt entry?

solr/core/src/java/org/apache/solr/search/CaffeineCache.java
```diff
@@ -250,26 +249,35 @@ public V computeIfAbsent(K key, IOFunction<? super K, ? extends V> mappingFuncti
...
+    `computeIfAbsent()`-type behavior. A number of Solr queries (and consequently,
+    `mappingFunction`s passed into this method) would modify the cache recursively.
+    This is supported in `async` mode, but when `async==false`, it may yield
+    "IllegalStateException: Recursive update" (see SOLR-16707).
+     */
+    V cached = cache.getIfPresent(key);
+    if (cached != null) {
+      return cached;
+    }
+    final V computed = mappingFunction.apply(key);
```
dsmiley, line 263: Computing the value outside the cache means a potentially expensive query that is requested concurrently will be computed redundantly.  I think that's the value that was added previously, now defeated.  Maybe we can get this behavior back by always doing async mode in 10?
CC @madrob who I think added this

magibney: Oh 100%, no doubt. In fact I was pretty clear about being in favor of removing the possibility of non-async in [comments on the related issue](https://issues.apache.org/jira/browse/SOLR-16707?focusedCommentId=17703668#comment-17703668).

But until we can deprecate/remove that option, we have two options to continue to support `async=false`, both not great:
1. document that users need to avoid join queries, avoid certain configurations of range queries and boolean queries, and ?? maybe other things?, or
2. revert the behavior of `async=false` to what it would have been pre-SOLR-15555, and have everything work, albeit with possible double-computation.

This PR goes the second route. Which, faced with the tradeoff that route implies for `async=false`, makes one wonder why anyone would ever configure `async=false`? But then we're [back to the question](https://issues.apache.org/jira/browse/SOLR-16707?focusedCommentId=17702769#comment-17702769) of whether there's any benefit to keeping `async=false` as an option at all.

In the meantime, this PR will make things work for users who (for whatever reason) enable this option, and it will also stop the unhelpful test failures.

magibney:
This fell through the cracks; circling back now, I plan to merge this within the next couple of days, pending any objections/feedback. cc @dsmiley , @mkhludnev.

dsmiley:
Looking back at this, the problem is that the cache implementation can't recursively compute new cache entries.  Given this situation, couldn't we use a ThreadLocal boolean to indicate if we're recursively being invoked and make different decisions based on this?  If recursive invocation, need to compute the query prior to the computeIfAbsent call.  The vast majority of invocations will not be recursive though.

dsmiley:
You could even modify the logic so we optimistically assume non-recursion (the fast path).  if the lambda in compute-if-absent sees we aren't so lucky, it could throw an exception to bail out of the computeIfAbsent call.  Catch it outside, and do the fallback logic.

magibney:
This sounds good; I had been thinking to not optimize the `async=false` case, and to instead move toward deprecating it. But I like this; it's straightforward, and will also provide a template for a simple way for plugin SolrCache impl's to handle this situation without requiring them to implement full-blown async support.

magibney:
@dsmiley thanks again for your suggestion; it's implemented in 3fb684f0291b508e7cbca3ad964dbe779211ff8d if you care to take a look!

solr/core/src/java/org/apache/solr/search/CaffeineCache.java
```diff
@@ -98,6 +98,13 @@ public class CaffeineCache<K, V> extends SolrCacheBase
   private int maxIdleTimeSec;
   private boolean cleanupThread;
   private boolean async;
+  private final ThreadLocal<boolean[]> detectRecursion =
+      new ThreadLocal<>() {
+        @Override
+        protected boolean[] initialValue() {
```
dsmiley, line 104: Why a boolean array; why not a Boolean?  We can easily change the ThreadLocal value; it's not immutable.
Also, it's less code to use ThreadLocal.[withInitial](https://www.tabnine.com/code/java/methods/java.lang.ThreadLocal/withInitial)

solr/core/src/java/org/apache/solr/search/CaffeineCache.java
```diff
@@ -269,7 +289,38 @@ public V computeIfAbsent(K key, IOFunction<? super K, ? extends V> mappingFuncti
           });
     } catch (UncheckedIOException e) {
       throw e.getCause();
+    } catch (RecursionException ignored) {
+      // recursive invocation detected; fallback to get-then-put
+    }
+
```
dsmiley, line 295: At this line, a comment saying "Recursive update detected; fall back on get-then-put" would clarify the overall flow of this complicated method.

dsmiley:
@magibney you seemed so close to this nice change being merged

magibney:
Ah, thanks for not letting this slip through the cracks; I'll follow up on this shortly

PR #1483, opened 2023-03-23
SOLR-16714 ReleaseWizard to add solrbot dependency upgrades to CHANGES

janhoy:
https://issues.apache.org/jira/browse/SOLR-16714

Here is a python script that runs `git log` to find all `solrbot` commits since last release.

NOTE: The wizard change is not yet tested, only the python script.

To test the python script yourself, check out this branch and run e.g. `dev-tools/scripts/addDepsToChanges.py --version 10.0.0`

### DIFF
--- a/dev-tools/scripts/addDepsToChanges.py
+++ b/dev-tools/scripts/addDepsToChanges.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


dev-tools/scripts/addDepsToChanges.py
```diff
@@ -0,0 +1,139 @@
...
+
+line_re = re.compile(r"(.*?) \(#(\d+)\)$")
+
+
+def get_prev_release_tag(ver):
+    """
+    Based on a given version, compute the git tag for the "previous" version to calculate changes since.
+    For a major version, we want all solrbot commits since last release, i.e. X-1.Y.Z
+    For a minor version, we want all solrbot commits since X.Y-1.0
+    For a patch version, we want all solrbot commits since X.Y.Z-1
```
janhoy, line 38: Alternatively, for a major release, we could add all commits since `X-1.Y.0`, even if that would repeat some of the upgrades made in patch releases?

dev-tools/scripts/addDepsToChanges.py
```diff
@@ -0,0 +1,139 @@
...
+                continue
+            if not appended:
+                if header_line_re.search(line):
+                    buffer.append(line)
+                    appended = True
+                    for change_line in changes_lines:
+                        buffer.append(change_line)
+                        buffer.append("\n")
+                    continue
+            buffer.append(line)
```
janhoy, line 105: This logic fins a `Dependency Upgrades` heading in the correct version section of CHANGES, and adds all entries directly below that heading. 

TODO:
* It won't delete the `(No changes)` line that would be there by default
* If the release version does not contain a `Dependency Upgrades` heading, the script fails. This is not likely to happen, as the `addVersion.py` script now adds this header by default.

PR #1487, opened 2023-03-23
SOLR-15493: Throw an error message if the feature store doesn't exist

ilariapet:
https://issues.apache.org/jira/browse/SOLR-15493
https://issues.apache.org/jira/browse/SOLR-15494

# Description

When a wrong/non-existent feature store name is passed to the query for feature extraction, an empty list is returned;
it also happens when inspecting the contents of a non-existent feature store. It would be helpful if a message was returned indicating that the feature store I am calling is not present.
Also, this non-existent (and empty) feature store is inappropriately added to the feature store list and only disappears after the collection is reloaded.
This contribution wants to add the ability to throw an exception when a non-existent feature store name is passed and thus wants to prevent it from being inappropriately added to the feature store list.

# Solution

A check for a non-existent feature store was added in the doGet method of the _ManagedFeatureStore.java_ class, throwing a Solr exception.

Also, a getter method of the "stores" variable was added in the _ManagedFeatureStore.java_ class and used to throw the exception during the feature extraction (logging phase) if the passed feature store is missing from the store list.
In this phase, there is the need to differentiate between the two cases where no feature store name is passed (i.e., it is null):
If a reranking query (qr) is not passed, it means the model name is not passed either, so the only feature store that could be used is the default feature store (which should not be empty, otherwise an error is thrown).
On the other hand, if a reranking query is passed, it means that a model name has been specified, so a temporarily empty feature store is created to later add features in the implTransform() method, using the methods of the FeatureLogger class.

# Tests

To verify that an Exception is thrown when a non-existent feature store is passed to the query (with fl parameter) 3 tests have been added in _solr/modules/ltr/src/test/org/apache/solr/ltr/feature/TestFeatureLogging.java_:

- _featureExtraction_NonExistentFeatureStore_shouldThrowException_
- _featureExtraction_NonExistentFeatureStoreReRanking_shouldThrowException_
- _featureExtraction_NoFeatureAndModelStore_shouldThrowException_

To verify that an Exception is thrown when trying to get a missing feature store and that this feature store is not inappropriately added to the feature store list, some tests in solr/modules/ltr/src/test/org/apache/solr/ltr/store/rest/TestModelManagerPersistence.java have been modified.
# Checklist

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -98,6 +98,8 @@ Improvements
 
 * SOLR-16711: Extract SolrCLI tool implementations into their own classes and reorganize CLI related classes into new org.apache.solr.cli package. (Eric Pugh)
 
+* SOLR-15493: Throw an error message if the feature store doesn't exist.  (Ilaria Petreti via Alessandro Benedetti)
+
 Optimizations
 ---------------------
 
--- a/solr/modules/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
+++ b/solr/modules/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
@@ -230,7 +230,8 @@ public class LTRFeatureLoggerTransformerFactory extends TransformerFactory {
       Map<String, String[]> transformerExternalFeatureInfo =
           LTRQParserPlugin.extractEFIParams(localparams);
 
-      final LoggingModel loggingModel = createLoggingModel(transformerFeatureStore);
+      final LoggingModel loggingModel =
+          createLoggingModel(transformerFeatureStore, docsWereNotReranked);
       setupRerankingQueriesForLogging(
           transformerFeatureStore, transformerExternalFeatureInfo, loggingModel);
       setupRerankingWeightsForLogging(context);
@@ -241,10 +242,31 @@ public class LTRFeatureLoggerTransformerFactory extends TransformerFactory {
      *
      * @param transformerFeatureStore the explicit transformer feature store
      */
-    private LoggingModel createLoggingModel(String transformerFeatureStore) {
-      final ManagedFeatureStore fr = ManagedFeatureStore.getManagedFeatureStore(req.getCore());
+    private LoggingModel createLoggingModel(
+        String transformerFeatureStore, Boolean docsWereNotReranked) {
+      final ManagedFeatureStore featureStores =
+          ManagedFeatureStore.getManagedFeatureStore(req.getCore());
 
-      final FeatureStore store = fr.getFeatureStore(transformerFeatureStore);
+      // check for non-existent feature store name
+      if (transformerFeatureStore != null) {
+        if (!featureStores.getStores().containsKey(transformerFeatureStore)) {
+          throw new SolrException(
+              SolrException.ErrorCode.BAD_REQUEST,
+              "Missing feature store: " + transformerFeatureStore);
+        }
+      }
+
+      final FeatureStore store = featureStores.getFeatureStore(transformerFeatureStore);
+
+      // check for empty feature store only if there is no reranking query, otherwise the model
+      // store will be used later for feature extraction
+      if (docsWereNotReranked) {
+        if (store.getFeatures().isEmpty()) {
+          throw new SolrException(
+              SolrException.ErrorCode.BAD_REQUEST,
+              "Empty feature store. If no ranking query (rq) is passed and DEFAULT feature store is empty, an existing feature store name must be passed for feature extraction.");
+        }
+      }
 
       // if transformerFeatureStore was null before this gets actual name
       transformerFeatureStore = store.getName();
--- a/solr/modules/ltr/src/test/org/apache/solr/ltr/feature/TestFeatureLogging.java
+++ b/solr/modules/ltr/src/test/org/apache/solr/ltr/feature/TestFeatureLogging.java
@@ -153,6 +153,74 @@ public class TestFeatureLogging extends TestRerankBase {
   }
 
   @Test
+  public void featureExtraction_nonExistentFeatureStore_shouldThrowException() throws Exception {
+    loadFeature("store9f1", ValueFeature.class.getName(), "store9", "{\"value\":3.0}");
+    loadModel(
+        "store9m1",
+        LinearModel.class.getName(),
+        new String[] {"store9f1"},
+        "store9",
+        "{\"weights\":{\"store9f1\":1.0}}");
+
+    final SolrQuery query = new SolrQuery();
+    query.setQuery("id:7");
+    query.add("rows", "1");
+
+    // Non-existent store specified, should throw exception
+    query.add("fl", "fv:[fv store=nonExistentFeatureStore]");
+    assertJQ(
+        "/query" + query.toQueryString(),
+        "/error/msg=='Missing feature store: nonExistentFeatureStore'");
+    assertJQ("/query" + query.toQueryString(), "/error/code==400");
+  }
+
+  @Test
+  public void featureExtraction_nonExistentFeatureStoreReRanking_shouldThrowException()
+      throws Exception {
+    loadFeature("store9f1", ValueFeature.class.getName(), "store9", "{\"value\":3.0}");
+    loadModel(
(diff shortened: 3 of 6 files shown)


solr/modules/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
```diff
@@ -241,10 +242,31 @@ public void setContext(ResultContext context) {
...
-    private LoggingModel createLoggingModel(String transformerFeatureStore) {
-      final ManagedFeatureStore fr = ManagedFeatureStore.getManagedFeatureStore(req.getCore());
+    private LoggingModel createLoggingModel(
+        String transformerFeatureStore, Boolean docsWereNotReranked) {
+      final ManagedFeatureStore featureStore =
+          ManagedFeatureStore.getManagedFeatureStore(req.getCore());
+
+      // check for non-existent feature store name
+      if (transformerFeatureStore != null) {
+        if (!featureStore.getStores().containsKey(transformerFeatureStore)) {
```
alessandrobenedetti, line 252: do we need this? isn't this already in the  'final FeatureStore store = featureStore.getFeatureStore(transformerFeatureStore);' ?

Do we even need 'featureStore.getStores()' ?

solr/modules/ltr/src/test/org/apache/solr/ltr/feature/TestExternalFeatures.java
```diff
@@ -200,6 +200,16 @@ public void featureExtraction_valueFeatureRequired_shouldThrowException() throws
         "/error/msg=='Exception from createWeight for ValueFeature [name=popularity, params={value=${myPop}, required=true}] ValueFeatureWeight requires efi parameter that was not passed in request.'");
   }
 
+  @Test
+  public void featureExtraction_wrongFeatureStore_shouldThrowException() throws Exception {
+    final SolrQuery query = new SolrQuery();
+    query.setQuery("*:*");
+    query.add("rows", "1");
+    query.add("fl", "fvalias:[fv store=featureStoreWrong]");
```
alessandrobenedetti, line 208: should this be 'nonExistent' rather than 'wrong' ?

solr/modules/ltr/src/test/org/apache/solr/ltr/feature/TestFeatureLogging.java
```diff
@@ -125,6 +126,11 @@ public void testDefaultStoreFeatureExtraction() throws Exception {
             + FeatureLoggerTestUtils.toFeatureVector("defaultf1", "1.0")
             + "'}");
 
+    // Wrong store specified, should throw exception
+    query.remove("fl");
+    query.add("fl", "fv:[fv store=storeWrong]");
```
alessandrobenedetti, line 131: should it be 'nonExistent' ?

alessandrobenedetti (commented):
Some minor to review

solr/modules/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
ilariapet: When a nonexistent feature store is passed with _"final FeatureStore store = featureStore.getFeatureStore(transformerFeatureStore);"_ no error is received and the store is improperly added (empty) to the list of feature stores. In fact, the getFeatureStore() method is also used for the 'put' phase.

So a check was added to get the error and avoid creating it.

alessandrobenedetti: mmmm what is the difference with:
`public synchronized FeatureStore getFeatureStore(String name) {
     if (name == null) {
       name = FeatureStore.DEFAULT_FEATURE_STORE_NAME;
 @@ -160,11 +164,12 @@ public void doGet(BaseSolrResource endpoint, String childId) {
     if (childId == null) {
       response.add(FEATURE_STORE_JSON_FIELD, stores.keySet());
     } else {
       // check for non-existent feature store name
       if (!stores.containsKey(childId)) {
         throw new SolrException(
             SolrException.ErrorCode.BAD_REQUEST, "Missing feature store: " + childId);
       }` ?

alessandrobenedetti: the formatting is terrible but you should get the idea

ilariapet: If I have understood correctly your doubt... 
I added two checks because _doGet_ method is used when you want to get a feature store, for example:
`GET http://localhost:8983/solr/books/schema/feature-store/nonExistentFeatureStore`

While _createLoggingModel()_ method is used when you want to extract features, for example:
`GET http://localhost:8983/solr/books/query?q=test&fl=id,[features store=nonExistentFeatureStore]
`

Here I also handled the cases when the store name is not passed (i.e., is null [features]) and when the rerank query is present.

solr/CHANGES.txt
```diff
@@ -85,7 +85,9 @@ Improvements
   `GET /api/node/logging/messages`.  And the threshold for the log "listener" can be modified with
   `PUT /api/node/logging/messages/threshold`.  (Jason Gerlowski, Calvince Otieno)
 
-* SOLR-16504: Convert CLI tools to use Jetty HTTP 2 client.  (Bence Szabo via Eric Pugh)  
+* SOLR-16504: Convert CLI tools to use Jetty HTTP 2 client.  (Bence Szabo via Eric Pugh)
```
alessandrobenedetti, line 88: line 88 should result unchanged

solr/modules/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
```diff
@@ -241,10 +242,31 @@ public void setContext(ResultContext context) {
...
+          ManagedFeatureStore.getManagedFeatureStore(req.getCore());
+
+      // check for non-existent feature store name
+      if (transformerFeatureStore != null) {
+        if (!featureStores.getStores().containsKey(transformerFeatureStore)) {
+          throw new SolrException(
+              SolrException.ErrorCode.BAD_REQUEST,
+              "Missing feature store: " + transformerFeatureStore);
+        }
+      }
```
alessandrobenedetti, line 257: let me rephrase it:
what happens if you entirely remove this block
`// check for non-existent feature store name
       if (transformerFeatureStore != null) {
         if (!featureStores.getStores().containsKey(transformerFeatureStore)) {
           throw new SolrException(
               SolrException.ErrorCode.BAD_REQUEST,
               "Missing feature store: " + transformerFeatureStore);
         }
       }`
       
       Shouldn't 'final FeatureStore store = featureStores.getFeatureStore(transformerFeatureStore);' raise the same exception?

solr/modules/ltr/src/test/org/apache/solr/ltr/feature/TestFeatureLogging.java
```diff
@@ -152,6 +152,74 @@ public void testDefaultStoreFeatureExtraction() throws Exception {
             + "'}");
   }
 
+  @Test
+  public void featureExtraction_NonExistentFeatureStore_shouldThrowException() throws Exception {
```
alessandrobenedetti, line 156: after the _, lowercase 'nonExistent...'

solr/modules/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
ilariapet: I added the exception in the _doGet_ method, so _getFeatureStore()_ remains unchanged and doesn't raise any exception as it is also used to put new feature stores.

solr/CHANGES.txt
ilariapet: oh ok... I have also seen that this file was recently modified at the same point and our fork doesn't contain the change

PR #1490, opened 2023-03-23
Remove duplicate code analytics tests

janhoy:
Just some hygiene - I know the analytics module is going away, but this was a low hanging fruit :)


HoustonPutman (approved):
tests pass, so this should be good!

risdenk (approved):

PR #1491, opened 2023-03-24
SOLR-16716: Replace commons-io usages with pure Java

risdenk:
https://issues.apache.org/jira/browse/SOLR-16716


solr/core/src/java/org/apache/solr/cloud/api/collections/TimeRoutedAlias.java
epugh, line 1: I like getting rid of the werid MoreObjects class that I've never seen before, boy I wish Java had nicer string building ;-)

epugh (approved):
LGTM....    Love the simpliication towards classes that I'm more familiar with ;-)

solr/core/src/test/org/apache/solr/schema/ChangedSchemaMergeTest.java
```diff
@@ -83,15 +84,15 @@ private void addDoc(SolrCore core, String... fieldValues) throws IOException {
...
   private CoreContainer init() throws Exception {
-    File changed = new File(solrHomeDirectory, "changed");
-    copyMinConf(changed, "name=changed");
+    Path changed = solrHomeDirectory.toPath().resolve("changed");
+    copyMinConf(changed.toFile(), "name=changed");
     // Overlay with my local schema
-    schemaFile = new File(new File(changed, "conf"), "schema.xml");
-    FileUtils.writeStringToFile(schemaFile, withWhich, StandardCharsets.UTF_8);
+    schemaFile = changed.resolve("conf").resolve("schema.xml").toFile();
+    Files.writeString(schemaFile.toPath(), withWhich, StandardCharsets.UTF_8);
```
dsmiley, line 91: The fields `solrHomeDirectory` and `schemaFile` should both be changed to be of type Path.  I took a look and we're always using them where we actually want a Path.

dsmiley (commented):
Maybe block org.apache.commons.io.FileUtils altogether.  It's java.io.File oriented; we want Path nowadays.

risdenk:
> Maybe block org.apache.commons.io.FileUtils altogether. It's java.io.File oriented; we want Path nowadays.

yea agreed I just didn't want to tackle that all at once. FileUtils moves to PathUtils for almost all remaining cases, but requires quite a few changes to go from File -> Path. https://issues.apache.org/jira/browse/SOLR-15919 started w/ Mike converting a few things. I'll tackle switching File -> Path separately.

solr/core/src/test/org/apache/solr/schema/ChangedSchemaMergeTest.java
risdenk: gah good call. I figured schemaFile and solrHomeDirectory were part of the test framework and not private variables.

risdenk: Fixed in 2d79b286b652e6e5da265a9e3767252fa7063532

PR #1495, opened 2023-03-27
SOLR-16720: Defer PKI header creation to send-time

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-16720

# Description

Prior to this commit, the PKIAuthPlugin decorates outgoing requests at 'queue' time instead of later when they're actually sent.  This can cause scenarios where, under heavy load, queue time and sending time are far enough apart that much or all of the TTL window for the outgoing request is expired by the time its even sent.

# Solution

This commit switches header-decoration to occur using a different lifecycle hook offered by Jetty, 'onBegin', so that it's done as close as possible to the actual time the request is sent.

This won't eliminate the possibility of errant TTL expiry, but should reduce it substantially under heavy load.

# Tests

Testing this is difficult, as it requires a heavy load profile to reproduce.  Solr (rightfully) doesn't offer any way to patch or customize PKIAuthPlugin, which makes forcing a delay difficult.  We may have to rely on manual testing and the existing test suite here, unless a novel idea arises?

    
# Checklist

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -101,6 +101,9 @@ Bug Fixes
 * SOLR-16730: Fix NPE in SystemInfoHandler for inter-node requests that would cause the Nodes page not to load.
   SystemInfoHandler no longer populates the username, roles and permissions in inter-node requests. (Tomás Fernández Löbbe)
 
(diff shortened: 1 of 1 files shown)


HoustonPutman (approved):
Looks good to me! (with a changelog entry)

PR #1501, opened 2023-03-28
SOLR-16713: Replace Guava usages with pure Java (part 2)

risdenk:
https://issues.apache.org/jira/browse/SOLR-16713

Cleans up the following by adding forbiddenapis rules:
* com.google.common.base.Throwables
* com.google.common.collect.ImmutableMap
* com.google.common.collect.**
* com.google.common.io.** (and consequently org.apache.commons.io.IOUtils#toString(java.io.Reader) since its the same usage)

### DIFF
--- a/solr/core/src/java/org/apache/solr/cluster/placement/plugins/MinimizeCoresPlacementFactory.java
+++ b/solr/core/src/java/org/apache/solr/cluster/placement/plugins/MinimizeCoresPlacementFactory.java
@@ -17,8 +17,6 @@
 
 package org.apache.solr.cluster.placement.plugins;
 
-import com.google.common.collect.Ordering;
-import com.google.common.collect.TreeMultimap;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
@@ -28,6 +26,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeMap;
 import org.apache.solr.cluster.Node;
 import org.apache.solr.cluster.Replica;
 import org.apache.solr.cluster.SolrCollection;
@@ -100,14 +99,15 @@ public class MinimizeCoresPlacementFactory
         }
 
         // Get number of cores on each Node
-        TreeMultimap<Integer, Node> nodesByCores =
-            TreeMultimap.create(Comparator.naturalOrder(), Ordering.arbitrary());
+        Map<Integer, Set<Node>> nodesByCores = new TreeMap<>(Comparator.naturalOrder());
 
         Set<Node> nodes = request.getTargetNodes();
 
         // Get the number of cores on each node and sort the nodes by increasing number of cores
         for (Node node : nodes) {
-          nodesByCores.put(coresPerNodeTotal.get(node.getName()), node);
+          nodesByCores
+              .computeIfAbsent(coresPerNodeTotal.get(node.getName()), k -> new HashSet<>())
+              .add(node);
         }
 
         Set<ReplicaPlacement> replicaPlacements =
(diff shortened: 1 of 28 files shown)


solr/core/src/java/org/apache/solr/logging/jul/JulWatcher.java
```diff
@@ -157,7 +158,9 @@ public SolrDocument toSolrDocument(LogRecord event) {
     doc.setField("message", event.getMessage().toString());
     Throwable t = event.getThrown();
     if (t != null) {
-      doc.setField("trace", Throwables.getStackTraceAsString(t));
+      StringWriter trace = new StringWriter();
+      t.printStackTrace(new PrintWriter(trace));
```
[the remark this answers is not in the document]
risdenk: @sonatype-lift ignore

solr/core/src/java/org/apache/solr/logging/log4j2/Log4j2Watcher.java
```diff
@@ -272,7 +273,11 @@ public SolrDocument toSolrDocument(LogEvent event) {
     Message message = event.getMessage();
     doc.setField("message", message.getFormattedMessage());
     Throwable t = message.getThrowable();
-    if (t != null) doc.setField("trace", Throwables.getStackTraceAsString(t));
+    if (t != null) {
+      StringWriter trace = new StringWriter();
+      t.printStackTrace(new PrintWriter(trace));
```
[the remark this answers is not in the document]
risdenk: @sonatype-lift ignore

HoustonPutman:
Why do we want to get rid of `IOUtils.toString()`

risdenk:
> Why do we want to get rid of `IOUtils.toString()`

This is a specific version of IOUtils.toString() specifically one that deals with Java Reader. Basically limits usage of commons-io. Guava and commons-io both had methods converting a Java Reader to a String so added a simple helper to StrUtils to do that conversion.

risdenk:
> Ahh ok I get that, but why do all of the readers need to be wrapped by BufferedReader first?

yea so they don't that is left over from a previous iteration where I was using BufferedReader.lines() I can fix that.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/MinimizeCoresPlacementFactory.java
```diff
@@ -123,20 +123,23 @@ public List<PlacementPlan> computePlacements(
...
           // nodeEntriesToAssign list so the right nodes get replicas.
-          ArrayList<Map.Entry<Integer, Node>> nodeEntriesToAssign =
+          List<Map.Entry<Integer, Node>> nodeEntriesToAssign =
               new ArrayList<>(totalReplicasPerShard);
-          Iterator<Map.Entry<Integer, Node>> treeIterator = nodesByCores.entries().iterator();
+          Iterator<Map.Entry<Integer, Node>> treeIterator =
+              nodesByCores.entrySet().stream()
+                  .flatMap(e -> e.getValue().stream().map(n -> Map.entry(e.getKey(), n)))
+                  .iterator();
           for (int i = 0; i < totalReplicasPerShard; i++) {
```
dsmiley, line 132: Since you are adding the use of streams above, you might as well remove the need for an iterator and for loop as well.  use `Stream.limit(totalReplicasPerShard).collect(Collectors.toList())` and then the whole concoction becomes rather satisfying IMO.

solr/solrj/src/java/org/apache/solr/common/util/StrUtils.java
```diff
@@ -363,4 +366,16 @@ public static String formatString(String pattern, Object... args) {
   public static boolean isNullOrEmpty(String string) {
     return string == null || string.isEmpty();
   }
+
+  public static String stringFromReader(Reader inReader) throws IOException {
+    try (Reader reader = new BufferedReader(inReader)) {
+      char[] arr = new char[8 * 1024];
+      StringBuilder buffer = new StringBuilder();
```
dsmiley, line 373: triple buffering?!

dsmiley: instead I recommend inReader.transferTo(StringWriter)

risdenk: thanks looks a lot nicer - fixed in b8e3fd33d327d0fd0c57ea226a4ea90764c3298f

solr/core/src/java/org/apache/solr/cluster/placement/plugins/MinimizeCoresPlacementFactory.java
risdenk: Yes much cleaner - fixed in 983a8306b78c1fa0237eb5661c2a8147a58a6a5d

dsmiley (approved):

PR #1509, opened 2023-03-30
WIP SOLR-16441: Upgrade Jetty to 11.x

risdenk:
https://issues.apache.org/jira/browse/SOLR-16441

This is just an example of what this might look like.


risdenk:
This is NOT ready to merge yet, but tagging some folks to at least look and see if this is something that we want to pursue. I think eventually we will need to go in this direction, but I don't think it needs to happen in Solr 9.x

thomaswoeckinger:
Will this be part of Solr 10?

risdenk:
@thomaswoeckinger no idea. this is updated to main now though. Still has the hadoop-auth and s3-repository modules disabled.

jorgerod:
Hi @risdenk 

Spring boot 3.0 and 3.1 use jetty 11x but Spring boot 3.2 will use jetty 12x.

Maybe it would be more interesting to upgrade directly to jetty 12x, what do you think? Or release two versions, one with 11x and one with 12x?

uschindler:
I would recommend to use Jetty 12. It allows to use both servlet APIs so we can stay with javax.servlet. You can run them in parallel, but actually a single context must use one version. SO means all of Solr needs to decide which version. 

I migarted one project here and it was great. There were some bugs in Jetty 12.0.0, but since 12.0.1 all worked fine.

epugh:
Let's rip that bandaid off and get to 12!   Jakarta EE!

risdenk:
I filed https://issues.apache.org/jira/browse/SOLR-17069 to look at Jetty 12.x

Marty1703:
I think this one can be closed, since version 12 seems to be released with https://github.com/apache/solr/pull/2876

PR #1511, opened 2023-03-31
SOLR-16730: Exclude username, roles and permissions for inter-node requests to SystemInfoHandler

tflobbe:
This PR addresses `SOLR-16730` by completely skipping including the username, roles and permissions when the request is an inter-node request to `SystemInfoHandler`. 
Having this part alone:
```
if (roles == null) {
     info.add("permissions", Set.of());
}
```
should be enough to handle the NPE described in the Jira issue, but I think it may be better to just skip this section for internal requests, since the data displayed is unrelated to the user that issues the request.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -96,6 +96,9 @@ Bug Fixes
 
 * SOLR-16734: SOLR_DATA_HOME is only honored in verbode mode (Kevin Risden)
 
+* SOLR-16730: Fix NPE in SystemInfoHandler for inter-node requests that would cause the Nodes page not to load.
+  SystemInfoHandler no longer populates the username, roles and permissions in inter-node requests. (Tomás Fernández Löbbe)
+
 Dependency Upgrades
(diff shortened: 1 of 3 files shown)


janhoy:
According to RBAP api, the `[getUserRoles()](https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/security/RuleBasedAuthorizationPluginBase.java#L371)` call should not return null but empty set. So I think that can be fixed too, by making [this](https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/security/RuleBasedAuthorizationPlugin.java#L65) null-safe.

After coding Kotlin for a few years I really miss its not-null-by-default!

Basically, this looks like a decent solution.

janhoy (approved):

tflobbe:
> Add a CHANGES entry

Yes, I typically avoid adding an entry until I'm ready to merge, to prevent conflicts :). Added now.

PR #1517, opened 2023-04-01
Update dependencies com.adobe.testing:s3mock-junit4 to v2.11.0, org.springframework.boot:spring-boot to 2.7.10, org.springframework:spring to 5.3.26, and software.amazon.awssdk to 2.19.1

solrbot:
This PR contains the following updates:

| Package | Type | Update | Change |
|---|---|---|---|
| [com.adobe.testing:s3mock-junit4](https://togithub.com/adobe/S3Mock) | test | minor | `2.1.34` -> `2.11.0` |

---

### Release Notes

<details>
<summary>adobe/S3Mock</summary>

### [`v2.11.0`](https://togithub.com/adobe/S3Mock/blob/HEAD/CHANGELOG.md#&#8203;2110)

[Compare Source](https://togithub.com/adobe/S3Mock/compare/2.10.2...2.11.0)

2.x is JDK8 LTS bytecode compatible, with Docker and JUnit / direct Java integration.

-   Features and fixes
    -   Support for GetBucketLocation API (fixes [#&#8203;985](https://togithub.com/adobe/S3Mock/issues/985))
-   Version updates
    -   Bump aws-java-sdk-s3 from 1.12.346 to 1.12.369
    -   Bump aws-v2.version from 2.18.21 to 2.19.1
    -   Bump docker-maven-plugin from 0.40.2 to 0.40.3
    -   Bump maven-dependency-plugin from 3.3.0 to 3.4.0
    -   Bump mockito-kotlin from 4.0.0 to 4.1.0
    -   Bump checkstyle from 10.4 to 10.5.0
    -   Bump kotlin.version from 1.7.21 to 1.7.22
    -   Bump alpine from 3.16.3 to 3.17.0 in /docker
    -   Bump spring-boot.version from 2.7.5 to 2.7.6

### [`v2.10.2`](https://togithub.com/adobe/S3Mock/blob/HEAD/CHANGELOG.md#&#8203;2102)

[Compare Source](https://togithub.com/adobe/S3Mock/compare/2.10.1...2.10.2)

2.x is JDK8 LTS bytecode compatible, with Docker and JUnit / direct Java integration.

-   Features and fixes
    -   Let S3Mock return correct errors for invalid bucket names (fixes [#&#8203;935](https://togithub.com/adobe/S3Mock/issues/935))
        -   Previous implementation returned a Spring generated error which does not disclose what's actually wrong
[...]

PR #1535, opened 2023-04-03
Update dependencies com.adobe.testing:s3mock-junit4 to v2.11.0, org.springframework.boot:spring-boot to 2.7.10, org.springframework:spring to 5.3.26, and software.amazon.awssdk to 2.19.1

risdenk:
This replaces https://github.com/apache/solr/pull/1517


HoustonPutman:
😮 yeah for this and Jetty 11, it would be great to move away from the Spring requirement of the adobe s3-mock.

risdenk:
@HoustonPutman can you take a look at 468310c4a856b2264e322f4b859d9599c372c406 - it does fix the tests, but I just feel weird adding/removing `/` from directories. I think there were other changes around this in the past?

solr/modules/s3-repository/src/test/org/apache/solr/s3/S3PathsTest.java
```diff
@@ -48,7 +48,7 @@ public void testFiles() throws S3Exception {
   public void testDirectory() throws S3Exception {
 
     client.createDirectory("/simple-directory");
-    assertTrue(client.pathExists("/simple-directory"));
+    assertFalse("Dir path needs a trailing slash", client.pathExists("/simple-directory"));
```
HoustonPutman, line 51: Hmmm so it appears adobe mock no-longer supports this, but I'm pretty sure that S3 supports it... Kind of annoying.

Not sure what this does with regards to test coverage... As long as the integration test works, I imagine we would be fine here....

I'll try to run the nightlies on the module and see how it goes.

risdenk: so I haven't looked in detail, but the s3 client did upgrade so maybe this is a client side AWS change?

HoustonPutman: So the integration tests work when I changed one line (which I have pushed), so this _should_ be good, though we have seen other strange errors that the integration tests didn't address. But overall I think we should be good here. directories are generally always treated as directories in the actual backup code...

HoustonPutman: The nightlies pass across all implementations but hdfs. I don't think the hdfs failures were related to the change (the tests wouldn't even start)

risdenk: > I don't think the hdfs failures were related to the change (the tests wouldn't even start)

ut oh... I'll take a closer look at this since I just upgrade Hadoop to 3.3.5... I haven't seen any Jenkins failures related to it though.

risdenk: hmmm I don't see the same issue w/ hdfs tests running `./gradlew test -Ptests.filter="@Nightly and not(@awaitsfix)"`

risdenk: Also https://ci-builds.apache.org/job/Solr/job/Solr-NightlyTests-main/748/ passed w/o issue for HDFS so not sure what you see :/

PR #1540, opened 2023-04-03
Remove unused joda-time version

risdenk:
This changes the joda-time version back: `2.12.5` -> `2.8.1`. I think this is ok since Solr doesn't use any joda-time stuff and this looks like a leftover from Solr 8 where all dependencies had to be defined. 

```
./gradlew why --hash d61aef93
To honour the JVM settings for this build a single-use Daemon process will be forked. See https://docs.gradle.org/7.6/userguide/gradle_daemon.html#sec:disabling_the_daemon.
Daemon will be stopped at the end of the build

> Task :why
joda-time:joda-time:2.8.1
        com.amazonaws:aws-java-sdk-core -> 2.8.1
        edu.ucar:cdm -> 2.2
        edu.ucar:udunits -> 2.2
```


janhoy (approved):
+1 to remove (and thus let AWS dependency dictate version), unless we fear regressing some joda-time vulnerability?

PR #1542, opened 2023-04-04
SOLR-16728: Fix SSL & HTTP2 classloading

HoustonPutman:
Need to add integration tests with SSL.

Also do we want to remove this security?

### DIFF
--- a/solr/server/contexts/solr-jetty-context.xml
+++ b/solr/server/contexts/solr-jetty-context.xml
@@ -29,10 +29,7 @@
       <New id="removeServerClasses" class="org.eclipse.jetty.webapp.ClassMatcher">
         <Arg>
           <Array type="java.lang.String">
-            <Item>-org.eclipse.jetty.io.</Item>
-            <Item>-org.eclipse.jetty.http.</Item>
-            <Item>-org.eclipse.jetty.http2.</Item>
-            <Item>-org.eclipse.jetty.util.</Item>
+            <Item>-org.eclipse.jetty.</Item>
           </Array>
         </Arg>
       </New>


risdenk (approved):

HoustonPutman:
I added an SSL integration test, which hangs before this change and passes afterwards.

You can reproduce via:

```
./gradlew integrationTest --tests "test_ssl.bats"
```

solr/packaging/test/test_ssl.bats
```diff
@@ -0,0 +1,61 @@
...
+  export SOLR_SSL_TRUST_STORE_PASSWORD=secret
+  export SOLR_SSL_NEED_CLIENT_AUTH=false
+  export SOLR_SSL_WANT_CLIENT_AUTH=false
+  export SOLR_SSL_CHECK_PEER_NAME=true
+  export SOLR_HOST=localhost
+
+  solr start -c
+  solr assert --started https://localhost:8983/solr --timeout 5000
+
+  run curl --cacert "$ssl_dir/solr-ssl.pem" 'https://localhost:8983/solr/admin/collections?action=CREATE&collection.configName=_default&name=test&numShards=2&replicationFactor=1&router.name=compositeId&wt=json'
```
risdenk, line 56: nitL I think this can be done with `curl --cert-type P12 --cert ...` that should remove the need for converting with openssl?

HoustonPutman: So I tried this out a few ways, and since this is a self-signed cert, it doesn't work. 

We have to use `--cacert` for self-signed certs, and that doesn't except a p12 type.

HoustonPutman:
I'm going to backport this to 9.2. This is a breaking change, so it will likely require a 9.2.1 release (even though users can fix it by updating the jetty context.)

How quickly do y'all think we should get 9.2.1 out there? I'm happy to do the release again since I make the bug.

janhoy:
Makes sense to start planning for a 9.2.1, and backport all relevant bug fixes, from this list https://issues.apache.org/jira/issues/?jql=project%20%3D%20SOLR%20AND%20issuetype%20%3D%20Bug%20AND%20affectedVersion%20%3D%209.2

HoustonPutman:
I'll send an email to @dev in a bit then.

PR #1543, opened 2023-04-04
Gradle 8.0.2

risdenk:


magibney:
Not specifically related to this PR, but I've been wondering whether there'd be any objection (unless there's a specific reason to upgrade gradle versions) to effectively treat gradle like a transitive dependency of Lucene. For the purpose of joint Solr/Lucene development this definitely makes things easier.

Thoughts?

risdenk:
@magibney not sure I understand what you mean? Are you saying that Lucene Gradle upgrades should always come before Solr? That Solr and Lucene should always match Gradle versions?

There is a lot of sharing going back and forth between Lucene and Solr related to the build. I don't fully support only Lucene getting upgrades first to the build. The two projects are separate.

magibney:
Purely as a convenience, for supporting dev workflows involving [Lucene dependency substitution](https://github.com/apache/solr/blob/32dd924b14ee7160cb3a47b75bbe9f51ee36a483/help/dependencies.txt#L111-L142).

The workflow requires the gradle versions be the same; when they are, joint development is _really_ straightforward. I'm suggesting that, from my own perspective, if Solr were to upgrade gradle in sync with Lucene, it would make certain things really easy (selfish, definitely!).

I don't intend to make an issue of it -- but there are lots of things that are treated as transitive dependencies, so to some extent it's arbitrary when we upgrade. What I've done to date is backport gradle upgrades to Lucene if/when needed, which is pretty straightforward -- worst case I can continue doing that.

cc @dweiss, @uschindler in case they have any thoughts?

risdenk:
Superseded by https://github.com/apache/solr/pull/1997

PR #1546, opened 2023-04-05
SOLR-16240: Fix kerberosPlugin module classloading

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-16240

The easy way out vs waiting on a hadoop fix and a hadoop release.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -107,6 +107,9 @@ Bug Fixes
 * SOLR-16728: Fix Classloading Exception for inter-node requests when using SSL and HTTP2.
   All Jetty classes are able to be shared between the Jetty server and webApp now. (Houston Putman)
 
+* SOLR-16240: Fix KerberosPlugin module classloading when using the hadoop-auth module.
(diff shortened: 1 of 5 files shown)


janhoy (approved):
Surprisingly simple workaround. Need to add exception to ForbiddenAPIs, and perhaps a code comment with JIRA link as a reminder to remove the workaround some day.

HoustonPutman:
Yeah will get that stuff cleaned up. Not sure we will want to remove this workaround, it does let us use non-request centric plugins without worrying if those plugins use the thread classloader or not. (This is similar workaround we do for the sql module, which has libraries that will never accept a cloassloader probably...)

I don't want to move merge this without an integration test, but I'm not sure how to get that working for kerberos... Might play around with that later if no one else has ideas.

HoustonPutman:
@janhoy added an integration test, which seems to work as expected, and fails when I remove the line that fixes this PR. Let me know if it looks good to you.

risdenk (approved):

PR #1549, opened 2023-04-07
SOLR-16741: CLUSTERSTATUS API returns wrong value for state ,leader for PRS collections

noblepaul:


solr/solrj/src/java/org/apache/solr/common/cloud/Replica.java
```diff
@@ -415,6 +415,13 @@ public void write(JSONWriter jsonWriter) {
     Map<String, Object> map = new LinkedHashMap<>();
     // this serializes also our declared properties
     _allPropsWriter().toMap(map);
+    boolean isLeader = isLeader();
```
hiteshk25, line 418: Does it make sense to include this in `_allPropsWriter` method? as this method included in other method `writeMap` as well.

noblepaul: done @hiteshk25

PR #1559, opened 2023-04-12
Gradle 8.1

risdenk:

This builds on https://github.com/apache/solr/pull/1543

This currently fails since palintir consistent versions tries to copy the classpath. 

```
* What went wrong:
A problem occurred configuring root project 'solr-root'.
> Cannot change the allowed usage of configuration ':solr:benchmark:compileClasspathCopy', as it was locked upon creation to the role: 'Deprecated Resolvable'.
  This role permits the following usage:
        Consumable - this configuration can be selected by another project as a dependency (but this behavior is marked deprecated)
        Resolvable - this configuration can be resolved by this project to a set of files
        Declarable Against - this configuration can have dependencies added to it (but this behavior is marked deprecated)
  Ideally, each configuration should be used for a single purpose.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.

* Get more help at https://help.gradle.org

BUILD FAILED in 18s
2 actionable tasks: 2 up-to-date
```

https://github.com/palantir/gradle-consistent-versions/issues


risdenk:
Superseded by https://github.com/apache/solr/pull/1997

PR #1562, opened 2023-04-13
SOLR-16747: Fix source-release gradle build

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-16747

Basically if the root directory is not a git repository, then don't include ref-guide tasks in the build by default.

The ref guide tasks require that the root directory is a git repository, otherwise they will not work. So we can only remove them. No way to "fix" these tasks.


risdenk (approved):

dsmiley (approved):

andywebb1975:
Thanks for fixing this Houston!

PR #1563, opened 2023-04-14
SOLR-16748: Official Dockerfile cleanup Args & GPG/SHA

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-16748

This PR finally takes a look at what we actually need to do in the official Dockerfile and how we should allow customization (A TODO that hadn't been addressed since the migration).

- [x] Check SHA even if GPG isn't checked
- [x] Remove all download servers except for the closer script (`https://www.apache.org/dyn/closer.lua`)
- [x] Only option for customizing the download location is via the `SOLR_DOWNLOAD_SERVER` arg. This must accept requests to `<version>/solr-<version>(-slim).tgz(.asc)`.
- [x] Only check GPG if the url contains `apache.org`, otherwise there is very little point in verifying the signature. (especially since mirrors are no longer a thing in apache).
- [x] Cleanup GPG logic, so that it only executes when GPG checking is enabled.
- [x] Make sure this works with the `gradle testBuildDockerfileOfficial`
- [x] Make sure that the release wizard instructions will work here.
- [x] Upgrade notes to describe the changes to accepted docker build args
- [x] Changelog entry

Big wins: The file is less complex, and there is only one place that the download url can come from, and the tests (and RC process) will ensure the GPG check succeeds.


solr/docker/build.gradle
```diff
@@ -453,10 +453,10 @@ if (''.equals(releaseGpgFingerprint)) {
         exec {
           commandLine 'docker', 'run',
               '--cidfile', mockServerIdFile,
-              '--rm',
```
risdenk, line 456: why is `--rm` removed? This would leave containers around afterwards

solr/docker/templates/Dockerfile.official.header.template
```diff
@@ -21,72 +21,62 @@
...
 # Override the default solr download location with a prefered mirror, e.g.:
 #   docker build -t mine --build-arg SOLR_DOWNLOAD_SERVER=https://downloads.apache.org/solr/solr .
-ARG SOLR_DOWNLOAD_SERVER
-
-# These should never be overridden except for the purposes of testing the Dockerfile before release
-ARG SOLR_CLOSER_URL="http://www.apache.org/dyn/closer.lua?action=download&filename=/solr/solr/$SOLR_VERSION/solr-$SOLR_VERSION$SOLR_DIST.tgz"
-ARG SOLR_DIST_URL="https://www.apache.org/dist/solr/solr/$SOLR_VERSION/solr-$SOLR_VERSION$SOLR_DIST.tgz"
-ARG SOLR_ARCHIVE_URL="https://archive.apache.org/dist/solr/solr/$SOLR_VERSION/solr-$SOLR_VERSION$SOLR_DIST.tgz"
+# This server must support downloading at: ${SOLR_DOWNLOAD_SERVER}/${SOLR_VERSION}/solr-${SOLR_VERSION}(-slim).tgz(.asc)
+ARG SOLR_DOWNLOAD_SERVER="https://www.apache.org/dyn/closer.lua?action=download&filename=/solr/solr"
```
risdenk, line 33: I've not checked but just to confirm this needs to support older Solr versions that only live on archive.apache.org right? for say building Solr 9.0 docker image? or when 9.3 is removed from https://www.apache.org/dist/solr/solr/

solr/docker/build.gradle
HoustonPutman: Good call, this was for debugging

solr/docker/templates/Dockerfile.official.header.template
HoustonPutman: It does, and it works for `.asc` files as well:

https://www.apache.org/dyn/closer.lua?action=download&filename=/solr/solr/9.2.0/solr-9.2.0-src.tgz
https://www.apache.org/dyn/closer.lua?action=download&filename=/solr/solr/9.2.0/solr-9.2.0-src.tgz.asc
https://www.apache.org/dyn/closer.lua?action=download&filename=/solr/solr/9.0.0/solr-9.0.0-src.tgz
https://www.apache.org/dyn/closer.lua?action=download&filename=/solr/solr/9.0.0/solr-9.0.0-src.tgz.asc

(linking the sources because they are smaller)

HoustonPutman:
Eventually this will allow non-committers to run the official dockerfile tests, and not require that a GPG signature is generated... But we can save that for another ticket.

HoustonPutman:
Ok I'm pretty sure the release will work, though some things are pretty hard to check...

janhoy (approved):
I read through the changes, built the official docker file, and also ran the testDockerFileOfficial target.

Did not test the release wizard or the smoke tester - guess the next release will battle-test those.

PR #1568, opened 2023-04-15
SOLR-16711: Extract SolrCLI tool implementations into their own package and classes (Take 2)

epugh:
https://issues.apache.org/jira/browse/SOLR-16711

# Description

Second take on this.

# Solution

# Tests

# Checklist


epugh:
Question?  Should we move `org.apache.solr.util.cli` up to `org.apache.solr.cli`?    And related...  Should we move `SolrCLI.java` into the `cli` package?

HoustonPutman:
> Question? Should we move `org.apache.solr.util.cli` up to `org.apache.solr.cli`? And related... Should we move `SolrCLI.java` into the `cli` package?

Yes to both IMO

solr/packaging/test/test_export.bats
```diff
@@ -29,7 +29,7 @@ teardown() {
 }
 
 @test "Check export command" {
-  run solr start -c -Dsolr.modules=sql
```
risdenk, line 32: Ummm why is this removed? This is checking a query so need sql module right?

solr/core/src/java/org/apache/solr/packagemanager/PackageManager.java
```diff
@@ -356,7 +357,7 @@ private boolean deployPackage(
     List<String> previouslyDeployedOnCollections = deployResult.second();
 
     // Verify
-    boolean verifySuccess = true;
+    boolean verifySuccess;
     // Verify that package was successfully deployed
     verifySuccess =
```
risdenk, line 362: Combine these lines?

risdenk (changes requested):

solr/packaging/test/test_export.bats
epugh: I don't think it uses the sql module..  it's just a straight up solr query, not a sql one....

gerlowskija (commented):
I haven't had time to review all of the changes (5k LOC 🤯 ), but things LGTM from the brief skimming I was able to do.

I'd feel more confident in large refactors like this if the bats coverage was beefier to catch regressions in syntax/output, but that's probably an unrealistic expectation at this point and this is definitely a step in the right direction IMO.

epugh:
> I haven't had time to review all of the changes (5k LOC 🤯 ), but things LGTM from the brief skimming I was able to do.
> 
> I'd feel more confident in large refactors like this if the bats coverage was beefier to catch regressions in syntax/output, but that's probably an unrealistic expectation at this point and this is definitely a step in the right direction IMO.

Yeah, I added two more bats tests, and actually the PDF that @bszabo97 uploaded on his work was super useful to me in thinking about these tests.     

I hope to add more bats tests as I look at each Tool in seperate PR's...

epugh:
> Huge improvement to split this up. LGTM although have not studied every change.
> 
> Are there places where you did more than simply breaking classes out, where you'd like us to pay special attention?

I tried to restrain myself from "fixing" too many things.  The edits were mostly just responding to some intellij warnings..   I think as long as the changes to the bash/cmd scripts look good?   The bats tests run as do the unit tests, so I think it's actually not too much risk..

janhoy:
Yea, and the script changes is basically java package name change. Looks good.

PR #1570, opened 2023-04-16
Factor out ModifiableSolrParams.setShardAttributesToParams method

renatoh:
small refactoring to get rid of duplicated code


epugh:
@renatoh  can you look at the precommit failure?  I think a `tidy` is needed ;-)

epugh:
Good questions @renatoh..  I am not possiitve, however if you run `gradlew tidy` then it applies our spotless code formatter....

renatoh:
@epugh  so what tool is used within gradlew tidy to format the code? I am wondering  it is possible to load these code style rules into a Intellij Code Style schema
I ran tidy and pushed the changes to my feature branch. is it possible to add the commit to the current PR or do I need to open a new PR?

epugh:
I *think* you pushed your tidy fixes up to the PR right?  (Which is great!)..   I clicked the "run precommit" github action...

renatoh:
In the log I see:
error: patch failed: gradlew.bat:80
error: gradlew.bat: patch does not apply
Not sure what this means, but given the checks were successful last week I assume it is because my branch got out of sync with main-branch thus I merged the main-branch into my branch. How do I run the checks again?

epugh:
sometimes "crave" get's in a odd state..   well, I should say "frequently" with how it merges stuf....    I see the precommit passed.    How would you like to credited in the CHANGES.txt file?

epugh:
I actually want to look a bit more at the method name...   And if it should actually be on the SearchComponent or elsewhere...  it isn't (I don't think!) specific to TaskManagement???

renatoh:
so in which class should the method be in?

cpoerschke:
From a very quick look only, maybe `ModifiableSolrParams.java` could be a home for it?

epugh:
i like @cpoerschke  suggestion, let's do that @renatoh and then we can merge!

renatoh:
@cpoerschke @epugh 
As suggested I have move the method to ModifiableSolrParams.java
I had to change the method-signature slightly in order not to depend on org.apache.solr.handler.component.ShardRequest

epugh (approved):
LGTM

solr/solrj/src/java/org/apache/solr/common/params/ModifiableSolrParams.java
```diff
@@ -175,6 +177,18 @@ public boolean remove(String name, String value) {
   // ----------------------------------------------------------------
   // ----------------------------------------------------------------
 
+  public static void setShardAttributesToParams(
+      int purpose, String shard, ModifiableSolrParams params) {
```
cpoerschke, line 181: subjective: method could be non-static maybe and/or to add test coverage e.g. in `ModifiableSolrParamsTest.java`
```suggestion
  public void setShardAttributes(
      int purpose, String shard) {
```

renatoh: true, I did not realize an instance of ModifiableSolrParams is present at the place where the method is called.
I have added a unit test validating the code, please have a look.
Also I have made two changes to the method, please have a look at me know if this makes sense to you.
Thanks!

renatoh:
@epugh I have done 2 or 3 very small commits before, e.g. fixing a typo or doing a minor change to 1 or 2 lines of code.
Yes, that is OK, I would like to be referenced with my full name, which is Renato Haeberli.
Any idea why the job is failing again? I do not see any meaning full error message in the log

epugh:
I don't know why, but yeah, Crave gets confused at times...

renatoh:
@risdenk Could you please have a look at it any time soon, thanks

epugh:
Going to merge today once checks run!

dsmiley:
See https://github.com/apache/solr/pull/2486 where I move it.  Would love a +1

PR #1576, opened 2023-04-19
SOLR-16755: bin/solr's '-noprompt' option no longer works for examples

janhoy:
https://issues.apache.org/jira/browse/SOLR-16755

The new test fails with `# ERROR: No line found` without this patch.

### DIFF
--- a/solr/packaging/test/test_example_noprompt.bats
+++ b/solr/packaging/test/test_example_noprompt.bats
@@ -29,8 +29,6 @@ teardown() {
 }
 
 @test "SOLR16755 test works with noprompt" {
-  echo starting solr
   solr start -e cloud -noprompt
-  echo waiting for solr to start
   solr assert --started http://localhost:8983/solr --timeout 10000
 }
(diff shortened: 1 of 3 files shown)


HoustonPutman:
@janhoy used bash arrays instead of the string.

We should also do this with `SOLR_SSL_OPTS` as it's the only other env to use `+="...."`, but we can tackle that in a separate issue that isn't backported to 9.2.1

janhoy:
Thanks Houston. I removed the now obsolete shellcheck disable comment. Will commit and backport shortly

solr/packaging/test/test_example_noprompt.bats
```diff
@@ -0,0 +1,36 @@
...
+
+teardown() {
+  # save a snapshot of SOLR_HOME for failed tests
+  save_home_on_failure
+
+  solr stop -all >/dev/null 2>&1
+}
+
+@test "SOLR16755 test works with noprompt" {
+  echo starting solr
```
epugh, line 32: We don't normally echo things out in our bats tests....

epugh (changes requested):
LGTM once two echos stripped out ;-)

PR #1577, opened 2023-04-19
SOLR-16719: Let AffinityPlacementFactory have a spread_domain label

tflobbe:
This PR adds the option to require anti-affinity within an AZ (essentially giving another level of anti-affinity). 
The usecase is, to support something like rack diversity between an AZ, or host diversity, when multiple Solr instances are running on the same physical host.

I plan to add a CHANGES entry before merging

### DIFF
--- a/solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
+++ b/solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
@@ -604,17 +604,11 @@ public class AffinityPlacementFactory implements PlacementPluginFactory<Affinity
           // available
           HashMap<String, List<Node>> antiAffinityNameToListOfNodesMap = new HashMap<>();
           for (Node node : availableNodesForPlacement) {
-            antiAffinityNameToListOfNodesMap.compute(
+            antiAffinityNameToListOfNodesMap.computeIfAbsent(
                 attributeValues
                     .getSystemProperty(node, AffinityPlacementConfig.ANTI_AFFINITY_SYSPROP)
                     .get(),
-                (k, v) -> {
-                  if (v == null) {
-                    v = new ArrayList<>();
-                  }
-                  v.add(node);
-                  return v;
-                });
+                ArrayList<>::new).add(node);
           }
           sortedAntiAffinityGroups =
               new TreeSet<>(new AntiAffinityComparator(currentAntiAffinityUsage));
--- a/solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
+++ b/solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
@@ -39,7 +39,6 @@ import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
-import org.apache.commons.lang3.mutable.MutableInt;
 import org.apache.solr.cluster.Cluster;
 import org.apache.solr.cluster.Node;
 import org.apache.solr.cluster.Replica;
@@ -542,7 +541,7 @@ public class AffinityPlacementFactory implements PlacementPluginFactory<Affinity
       private final List<Node> availableNodesForPlacement;
       private final AttributeValues attributeValues;
       private TreeSet<AffinityGroupWithNodes> sortedAntiAffinityGroups;
-      private final Map<String, MutableInt> currentAntiAffinityUsage;
+      private final Map<String, Integer> currentAntiAffinityUsage;
       private int numNodesForPlacement;
 
       AzWithNodes(
@@ -552,7 +551,7 @@ public class AffinityPlacementFactory implements PlacementPluginFactory<Affinity
           Comparator<Node> nodeComparator,
           Random random,
           AttributeValues attributeValues,
-          Map<String, MutableInt> currentAntiAffinityUsage) {
+          Map<String, Integer> currentAntiAffinityUsage) {
         this.azName = azName;
         this.availableNodesForPlacement = availableNodesForPlacement;
         this.useAntiAffinity = useAntiAffinity;
@@ -576,7 +575,7 @@ public class AffinityPlacementFactory implements PlacementPluginFactory<Affinity
 
       private void sort() {
         assert !listIsSorted && sortedAntiAffinityGroups == null
-                : "We shouldn't be sorting this list again";
+            : "We shouldn't be sorting this list again";
 
         // Make sure we do not tend to use always the same nodes (within an AZ) if all
         // conditions are identical (well, this likely is not the case since after having added
@@ -592,7 +591,8 @@ public class AffinityPlacementFactory implements PlacementPluginFactory<Affinity
           // When we use anti-affinity, we don't just sort the list of nodes, instead we generate a
           // TreeSet of AffinityGroupWithNodes,
           // sorted by the number of times the affinity label has been used. Each
-          // AffinityGroupWithNodes internally contains the list of nodes that use a particular affinity
+          // AffinityGroupWithNodes internally contains the list of nodes that use a particular
+          // affinity
           // label, and it's sorted internally by the comparator passed to this
           // class (which is the same that's used when not using anti-affinity).
           // Whenever a node from a particular AffinityGroupWithNodes is selected as the best
@@ -616,7 +616,8 @@ public class AffinityPlacementFactory implements PlacementPluginFactory<Affinity
                   return v;
                 });
           }
-          sortedAntiAffinityGroups = new TreeSet<>(new AntiAffinityComparator(currentAntiAffinityUsage));
+          sortedAntiAffinityGroups =
+              new TreeSet<>(new AntiAffinityComparator(currentAntiAffinityUsage));
 
           int i = 0;
           for (Map.Entry<String, List<Node>> entry : antiAffinityNameToListOfNodesMap.entrySet()) {
@@ -653,11 +654,9 @@ public class AffinityPlacementFactory implements PlacementPluginFactory<Affinity
               group.affinityGroupName,
               (k, v) -> {
                 if (v == null) {
-                  v = new MutableInt(1);
-                } else {
-                  v.increment();
+                  return 1;
                 }
-                return v;
+                return v + 1;
               });
           if (!group.sortedNodesForPlacement.isEmpty()) {
             sortedAntiAffinityGroups.add(group);
@@ -898,7 +897,7 @@ public class AffinityPlacementFactory implements PlacementPluginFactory<Affinity
 
       // This Map will include the affinity labels for the nodes that are currently hosting replicas
       // of this shard. It will be modified with new placement decisions.
-      Map<String, MutableInt> affinityLabelsInUse = new HashMap<>();
+      Map<String, Integer> affinityLabelsInUse = new HashMap<>();
       Shard shard = solrCollection.getShard(shardName);
       if (shard != null) {
         // shard is non null if we're adding replicas to an already existing collection.
@@ -922,10 +921,9 @@ public class AffinityPlacementFactory implements PlacementPluginFactory<Affinity
                       .get(),
                   (k, v) -> {
                     if (v == null) {
-                      return new MutableInt(1);
+                      return 1;
                     }
-                    v.increment();
-                    return v;
+                    return v + 1;
                   });
             }
           }
@@ -1205,10 +1203,9 @@ public class AffinityPlacementFactory implements PlacementPluginFactory<Affinity
     }
 
     static class AntiAffinityComparator implements Comparator<AffinityGroupWithNodes> {
-      private static final MutableInt ZERO = new MutableInt(0);
-      private final Map<String, MutableInt> antiAffinityUsage;
+      private final Map<String, Integer> antiAffinityUsage;
 
-      AntiAffinityComparator(Map<String, MutableInt> antiAffinityUsage) {
+      AntiAffinityComparator(Map<String, Integer> antiAffinityUsage) {
         this.antiAffinityUsage = antiAffinityUsage;
       }
 
@@ -1220,8 +1217,8 @@ public class AffinityPlacementFactory implements PlacementPluginFactory<Affinity
         // 2. On equal number of usages, by the internal comparator (which uses core count and disk
         // space) on the best node for each group (which, since the list is sorted, it's always the
         // one in the position 0)
-        MutableInt usagesLabel1 = antiAffinityUsage.getOrDefault(group1.affinityGroupName, ZERO);
-        MutableInt usagesLabel2 = antiAffinityUsage.getOrDefault(group2.affinityGroupName, ZERO);
+        Integer usagesLabel1 = antiAffinityUsage.getOrDefault(group1.affinityGroupName, 0);
+        Integer usagesLabel2 = antiAffinityUsage.getOrDefault(group2.affinityGroupName, 0);
         if (usagesLabel1.equals(usagesLabel2)) {
           return group1.compareTo(group2);
         }
--- a/solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
+++ b/solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
@@ -904,17 +904,13 @@ public class AffinityPlacementFactory implements PlacementPluginFactory<Affinity
               azToNumReplicas.put(az, azToNumReplicas.get(az) + 1);
             }
             if (doUseAntiAffinity) {
-              affinityLabelsInUse.compute(
+              affinityLabelsInUse.merge(
                   attrValues
                       .getSystemProperty(
                           replica.getNode(), AffinityPlacementConfig.ANTI_AFFINITY_SYSPROP)
                       .get(),
-                  (k, v) -> {
-                    if (v == null) {
-                      return 1;
-                    }
-                    return v + 1;
-                  });
+                  1,
+                  Integer::sum);
             }
           }
         }
--- a/solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
+++ b/solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
@@ -542,7 +542,7 @@ public class AffinityPlacementFactory implements PlacementPluginFactory<Affinity
       private final List<Node> availableNodesForPlacement;
       private final AttributeValues attributeValues;
       private TreeSet<AffinityGroupWithNodes> sortedAntiAffinityGroups;
-      private final Map<String, MutableInt> antiAffinityUsage;
+      private final Map<String, MutableInt> currentAntiAffinityUsage;
       private int numNodesForPlacement;
 
       AzWithNodes(
@@ -552,16 +552,14 @@ public class AffinityPlacementFactory implements PlacementPluginFactory<Affinity
           Comparator<Node> nodeComparator,
           Random random,
           AttributeValues attributeValues,
-          Map<String, MutableInt> antiAffinityUsage) {
+          Map<String, MutableInt> currentAntiAffinityUsage) {
         this.azName = azName;
-        // Once the list is sorted to an order we're happy with, this flag is set to true to avoid
-        // sorting multiple times unnecessarily.
         this.availableNodesForPlacement = availableNodesForPlacement;
         this.useAntiAffinity = useAntiAffinity;
         this.nodeComparator = nodeComparator;
         this.random = random;
         this.attributeValues = attributeValues;
-        this.antiAffinityUsage = antiAffinityUsage;
+        this.currentAntiAffinityUsage = currentAntiAffinityUsage;
         this.numNodesForPlacement = availableNodesForPlacement.size();
       }
 
@@ -577,6 +575,9 @@ public class AffinityPlacementFactory implements PlacementPluginFactory<Affinity
       }
 
       private void sort() {
+        assert !listIsSorted && sortedAntiAffinityGroups == null
+                : "We shouldn't be sorting this list again";
+
         // Make sure we do not tend to use always the same nodes (within an AZ) if all
         // conditions are identical (well, this likely is not the case since after having added
         // a replica to a node its number of cores increases for the next placement decision,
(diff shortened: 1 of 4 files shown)


tflobbe:
This is the output of the `testScalability` on my machine running with nightly property:

```
Patch: 
1225 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.SolrTestCaseJ4 ###Starting testScalability
1229 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ==== numNodes ====
1559 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 1000 nodes, 100 shards, 10000 total replicas, anti-affinity=false, elapsed time 316 ms.
2119 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 2000 nodes, 100 shards, 10000 total replicas, anti-affinity=false, elapsed time 550 ms.
3515 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 10000 total replicas, anti-affinity=false, elapsed time 1385 ms.
6411 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 10000 nodes, 100 shards, 10000 total replicas, anti-affinity=false, elapsed time 2875 ms.
12579 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 20000 nodes, 100 shards, 10000 total replicas, anti-affinity=false, elapsed time 6133 ms.
12584 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ==== numShards ====
13987 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 10000 total replicas, anti-affinity=false, elapsed time 1393 ms.
16791 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 200 shards, 20000 total replicas, anti-affinity=false, elapsed time 2794 ms.
23470 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 500 shards, 50000 total replicas, anti-affinity=false, elapsed time 6666 ms.
36979 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 1000 shards, 100000 total replicas, anti-affinity=false, elapsed time 13488 ms.
63988 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 2000 shards, 200000 total replicas, anti-affinity=false, elapsed time 26970 ms.
64063 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ==== numReplicas ====
64624 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 10000 total replicas, anti-affinity=false, elapsed time 556 ms.
65268 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 20000 total replicas, anti-affinity=false, elapsed time 636 ms.
66351 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 50000 total replicas, anti-affinity=false, elapsed time 1073 ms.
68766 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 100000 total replicas, anti-affinity=false, elapsed time 2397 ms.
76679 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 200000 total replicas, anti-affinity=false, elapsed time 7873 ms.
76753 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ==== antiAffinity ====
78120 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 10000 total replicas, anti-affinity=false, elapsed time 1363 ms.
79883 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 10000 total replicas, anti-affinity=true, elapsed time 1756 ms.
79888 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[29B84ABBD0C09F4A]) [] o.a.s.SolrTestCaseJ4 ###Ending testScalability


Main:
1160 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.SolrTestCaseJ4 ###Starting testScalability
1164 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ==== numNodes ====
1498 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 1000 nodes, 100 shards, 10000 total replicas, elapsed time 323 ms.
2081 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 2000 nodes, 100 shards, 10000 total replicas, elapsed time 574 ms.
3543 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 10000 total replicas, elapsed time 1451 ms.
6570 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 10000 nodes, 100 shards, 10000 total replicas, elapsed time 3008 ms.
13138 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 20000 nodes, 100 shards, 10000 total replicas, elapsed time 6539 ms.
13144 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ==== numShards ====
14612 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 10000 total replicas, elapsed time 1462 ms.
17527 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 200 shards, 20000 total replicas, elapsed time 2907 ms.
24854 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 500 shards, 50000 total replicas, elapsed time 7311 ms.
39991 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 1000 shards, 100000 total replicas, elapsed time 15117 ms.
68719 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 2000 shards, 200000 total replicas, elapsed time 28684 ms.
68797 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ==== numReplicas ====
69462 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 10000 total replicas, elapsed time 661 ms.
70184 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 20000 total replicas, elapsed time 715 ms.
71320 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 50000 total replicas, elapsed time 1127 ms.
73796 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 100000 total replicas, elapsed time 2459 ms.
82653 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.c.p.p.AffinityPlacementFactoryTest ComputePlacement: 5000 nodes, 100 shards, 200000 total replicas, elapsed time 8823 ms.
82725 INFO  (TEST-AffinityPlacementFactoryTest.testScalability-seed#[CE15B43BFC926FB0]) [] o.a.s.SolrTestCaseJ4 ###Ending testScalability
```

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -204,6 +206,8 @@ static class AffinityPlacementPlugin implements PlacementPlugin {
     private final Random replicaPlacementRandom =
```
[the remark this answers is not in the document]
risdenk: @sonatype-lift ignore

risdenk (approved):

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -694,6 +913,19 @@ private void makePlacementDecisions(
...
+                  attrValues
+                      .getSystemProperty(
+                          replica.getNode(), AffinityPlacementConfig.ANTI_AFFINITY_SYSPROP)
+                      .get(),
+                  (k, v) -> {
+                    if (v == null) {
+                      return 1;
+                    }
+                    return v + 1;
+                  });
```
HoustonPutman, line 927:
```suggestion
              affinityLabelsInUse.merge(
                  attrValues
                      .getSystemProperty(
                          replica.getNode(), AffinityPlacementConfig.ANTI_AFFINITY_SYSPROP)
                      .get(),
                  1,
                  Integer::sum);
```

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -703,15 +935,26 @@ private void makePlacementDecisions(
...
+        for (Node node : candidateNodes) {
+          String nodeAz = getNodeAZ(node, attrValues);
+          List<Node> nodesForAz = nodesPerAz.computeIfAbsent(nodeAz, k -> new ArrayList<>());
+          nodesForAz.add(node);
+        }
       }
 
+      Comparator<Node> interGroupNodeComparator =
+          new CoresAndDiskComparator(attrValues, coresOnNodes, prioritizedFreeDiskGB);
+      ;
```
HoustonPutman, line 956:
```suggestion
```

HoustonPutman (commented):
### Spread vs Anti Affinity

So overall I think this is a great feature, I do think a different name would be more appropriate though.

From what I can glean this is more making sure that the "spread" of replicas across affinity groups is consistent. Anti-affinity is more making sure that multiple of the same "thing" can't be in the same affinity group (this is what is described in the JIRA). I believe "anti-affinity" is more like the old (and removed) `maxShardsPerNode` param.

I like the functionality, but I think a better name would be `affinity_spread` or something with `spread` in it. Also I completely support adding both of these features, there should be both a `spread` and an `anti_affinity`.

### Spread deviation logic

Though if we are definining this as "spread" there are two ways we could go:
- Any new replica should increase the spread as best as possible (That is the current logic of the PR)
- Any new replica must be spread across all affinityGroups

Basically we are only looking at eligible nodes when computing spread, where the other option is to look at all affinity groups. Say affinity groups are:

```json
{
  a: ["node1", "node2"],
  b: ["node3"],
  c: ["node4", "node5", "node6"]
}
```
and there is a replica on node1 and node4. However node3 is not eligible for some reason. We should have a parameter to say how much the spread can deviate. If that is `0` however, an error should be thrown and the replica placement should fail because the desired spread cannot be achieved with the eligible nodes. If the deviation can be `1`, then one of `node2`, `node5` or `node6` would be acceptable.

I think with the current way the affinityPlacementFactory works, your approach is the right way. I just wanted to provide the options and explicitly say what we are choosing versus the other ways we could go.

### Sidenote

I'm kind of going off of the very powerful [Kubernetes Pod Node Assignment logic](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity) for this. Its more powerful and you can set various "sysProps" to be affinity and anti-affinity and spread, giving each rule a weight and whether it is a "must" or "should'. It would be great if we could eventually get there, but it would certainly be a much bigger rewrite than this.

tflobbe:
> I do think a different name would be more appropriate though.

I see. I wasn't aware of the "spread" feature in Kubernetes, but yes, it does look closer to what I have here. I was actually going with this naming based on Kubernetes' use of affinity/anti-affinity and making this a "soft" anti-affinity, and I was initially thinking about giving up on the rule if it couldn't be satisfied, but ended up implementing a logic similar to the "spread" in such case.

> Also I completely support adding both of these features, there should be both a spread and an anti_affinity.

If I understand correctly, you are suggesting:
* Let users configure the "spread", which is exactly what the PR is doing for now (new features such as the `maxSkew` could be added in the future)
* Make anti-affinity always "required" when configured: Attempt to spread the replicas, but fail if more than one replica is placed on the same affinity group.

It probably makes sense to keep both features using the same system property for defining the affinity groups (or "domains"?), right? I see no reason to have both features enabled but using different labels.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -503,15 +534,199 @@ private String getNodeAZ(Node n, final AttributeValues attrValues) {
...
+      }
+
+      void ensureSorted() {
+        if (!hasBeenSorted()) {
+          sort();
+        }
+      }
+
+      private void sort() {
+        assert !listIsSorted && sortedAntiAffinityGroups == null
```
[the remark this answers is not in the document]
tflobbe: Given these are internal private classes, I'd rather make sure they are being used as designed. Returning here would feel like hiding misuse IMO

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -503,15 +534,199 @@ private String getNodeAZ(Node n, final AttributeValues attrValues) {
...
+                new AffinityGroupWithNodes(entry.getKey(), entry.getValue(), i++, nodeComparator));
+          }
+        } else {
+          availableNodesForPlacement.sort(nodeComparator);
+          listIsSorted = true;
+        }
+      }
+
+      Node getBestNode() {
+        assert hasBeenSorted();
```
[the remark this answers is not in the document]
tflobbe: There isn't, for now. The reason I decided to have the assert is that the API of this class relies on the class being correctly used, in the correct order, etc. I want to protect against future changes that may start using this class incorrectly

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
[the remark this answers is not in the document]
HoustonPutman: Understood. As long as it doesn't turn into a runtime error for users, it's fine. I'm just not very familiar with asserts outside of tests.

HoustonPutman:
> Let users configure the "spread", which is exactly what the PR is doing for now (new features such as the maxSkew could be added in the future)

Yep 👍 

> Make anti-affinity always "required" when configured: Attempt to spread the replicas, but fail if more than one replica is placed on the same affinity group.

Yes, but have the requirement be configurable. So anti-affinity could have a parameter "maxAffinity" (awful name, we would need a better one), which is a rule that "the maximum replicas in this affinityGroup is `<maxAffinity>`". 

> It probably makes sense to keep both features using the same system property for defining the affinity groups (or "domains"?), right? I see no reason to have both features enabled but using different labels.

If we have both of these options available (anti-affinity and spread), I can imagine that they might be used with the different affinityGroup "keys". Like you would have spread setup for "availabilityZone" and anti-affinity setup for "rack".  In that case, you want to spread the nodes across availabilityZones as best as possible, and you want to avoid putting two replicas on the same rack.

Thinking about it more given my answer above, I guess this does make more sense as a single option, with separate toggles like "maxSkew" and "maxAffinity" (still and awful name). And maybe we eventually allow for two of these to be configured.

So in my example you would have the `spread` configured for both `availabilityZone` and `rack`. `availabilityZone` could have a `maxAffinity` of `-1` or `0`, meaning there is no issue with there being tons of replicas in that `availabilityZone`. But `rack` would have a `maxAffinity` of `2`, saying you don't want more than 2 replicas in a single `rack`. 

I know that having multiple `affinityKeys` specified is not yet supported, just kind of thinking out loud. So in the end, I do think it makes sense to keep both features using the same system property, just in the future allow for defining multiple of them.

tflobbe:
I see. Do keep in mind that "availabilityZone" is already doing the first level of spreading, and this configuration is only within one AZ (for racks, in your example). I guess someone may want AZ spreading + rack spreading + host anti-affinity or something like that, but maybe that's not a very common case.

It gets more complicated with some of the other variables already implemented, like replica types, or the "withCollection" setting

HoustonPutman:
> Do keep in mind that "availabilityZone" is already doing the first level of spreading, and this configuration is only within one AZ (for racks, in your example).

I did not realize that, well in that case, I'm not sure we will need another layer. So I'm good to go at this point. Sorry for throwing out the erroneus things.

I think the only thing left here is to rename it from "anti-affinity" to "spread" or something similar.
And additions to the ref-guide page for this.

Both "maxAffinity" and "spreadSkew" can be added in future tickets.

> It gets more complicated with some of the other variables already implemented, like replica types, or the "withCollection" setting

Absolutely. I'm assuming this already plays well with those?

tflobbe:
I've updated the code with a possible rename. the new validation of "maxReplicasPerShardInDomain" is not being used yet. 
Still, there is a need to improve javadocs and add something to the reference guide (couldn't find existing docs for this plugin)

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
```diff
@@ -503,15 +533,203 @@ private String getNodeAZ(Node n, final AttributeValues attrValues) {
...
+                    k -> new ArrayList<>())
+                .add(node);
+          }
+          sortedSpreadDomains =
+              new TreeSet<>(new SpreadDomainComparator(currentSpreadDomainUsageUsage));
+
+          int i = 0;
+          for (Map.Entry<String, List<Node>> entry : spreadDomainToListOfNodesMap.entrySet()) {
+            if (maxReplicasCountInDomain != Integer.MAX_VALUE
+                && this.currentSpreadDomainUsageUsage.getOrDefault(entry.getKey(), 0)
```
HoustonPutman, line 624: I don't think we need the maxReplicasCountInDomain check, since the second clause will return the same either way. Right?
```suggestion
            if (this.currentSpreadDomainUsageUsage.getOrDefault(entry.getKey(), 0)
```

solr/solr-ref-guide/modules/configuration-guide/pages/replica-placement-plugins.adoc
```diff
@@ -184,6 +192,24 @@ This property defines an additional constraint that collections (keys) must be l
...
+Spread replicas for a given shard/replica type across spread domains, that are defined by nodes using the system property `spread_domain`
+
+`maxReplicasPerShardInDomain`::
++
+[%autowidth,frame=none]
+|===
+|Optional |Default: `-1`
+|===
++
+Set a maximum number of replicas (for a given shard and replica type) that can be placed in a particular spread domain. Fail the request if this can't be satisfied. This can be used only in combination of `spreadAcrossDomains`.
```
HoustonPutman, line 211:
```suggestion
Set a maximum number of replicas (for a given shard and replica type) that can be placed in a particular spread domain. Fail the request if this can't be satisfied. This can be used only in combination with `spreadAcrossDomains`.
```

HoustonPutman (approved):
Two small changes, but overall this looks great!

I think the affinityPlugin is pretty much exactly what anyone could need at this point...

solr/core/src/java/org/apache/solr/cluster/placement/plugins/AffinityPlacementFactory.java
tflobbe: Yes, it should be the same. I just included the first part to avoid doing the `map.get` N times if the feature is not enabled (which I believe would be a very common case). Probably not a big deal, but since this class is already putting so much attention to performance and scalability I thought this wouldn't hurt

HoustonPutman: Fair enough!

PR #1578, opened 2023-04-20
SOLR-16756: SchemaCodecFactory, add test coverage for getKnnVectorsFormatForField

potenzal:
https://issues.apache.org/jira/browse/SOLR-16756

# Description

The getKnnVectorsFormatForField() override in SchemaCodecFactory doesn't seem to be covered by any of the existing tests.

Adding basic test points could be useful to detect future regressions, similar to the recent https://issues.apache.org/jira/browse/SOLR-16588

# Solution

Added testKnnVectorsFormat() and testUnsupportedKnnVectorsFormat() to the existing TestCodecSupport

# Tests

The tests provide a basic coverage of both branches. I believe that testKnnVectorsFormat would have detected SOLR-16588

# Checklist


potenzal:
@alessandrobenedetti since you were involved in the fix to SOLR-16588, I was wondering what you thoughts are on adding this simple regression test, any flags/suggestions to improve it?

serkankirbas (approved):
@potenzal Thanks for the suggestion. Adding automated tests and increasing test coverage is a good way to catch issues proactively before the release. 

Do you have a coverage report showing which lines of getKnnVectorsFormatForField are now covered by the tests added with this PR?  
https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java#L121

I suspect line 138 isn't covered. if that's the case, would it be possible to cover that line too?  https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java#L138

solr/core/src/test/org/apache/solr/core/TestCodecSupport.java
```diff
@@ -265,4 +266,27 @@ public void testCompressionModeDefault() throws IOException {
...
   }
+
+  public void testKnnVectorsFormat() {
+    Codec codec = h.getCore().getCodec();
+    Map<String, SchemaField> fields = h.getCore().getLatestSchema().getFields();
+    SchemaField schemaField = fields.get("vector");
+
+    PerFieldKnnVectorsFormat format = (PerFieldKnnVectorsFormat) codec.knnVectorsFormat();
+    assertEquals(
+        "Lucene95HnswVectorsFormat",
```
cpoerschke, line 277: minor: perhaps something like this could result in a compile-time rather than test-run breakage when `Lucene95HnswVectorsFormat` becomes `Lucene98HnswVectorsFormat` or so in a future Lucene version upgrade?
```suggestion
        Lucene95HnswVectorsFormat.class.getName(),
```

potenzal: Makes sense, thanks!

I have just replaced class.getName() with class.getSimpleName() to get rid of the full package info.

PR #1593, opened 2023-04-27
SOLR-16773: UI: Cloud>Nodes screen fix display of cores with non-standard shard names

janhoy:
https://issues.apache.org/jira/browse/SOLR-16773
Skjermbilde 2023-04-27 kl  13 45 23

### DIFF
--- a/solr/webapp/web/js/angular/controllers/cloud.js
+++ b/solr/webapp/web/js/angular/controllers/cloud.js
@@ -107,11 +107,6 @@ function isNumeric(n) {
   return !isNaN(parseFloat(n)) && isFinite(n);
 }
 
-// Makes a shorter label if possible, by shortening 'shard1' to 's1' and 'replica_n2' to 'r2'
-function makeCoreLabel(collection, shard, replica) {
-  return collection + "_" + (shard + "_").replace(/shard(\d+)_/, 's\$1') + replica.replace(/replica_?[ntp]?(\d+)/, 'r\$1');
-}
-
 var nodesSubController = function($scope, Collections, System, Metrics) {
   $scope.pageSize = 10;
   $scope.showNodes = true;
@@ -203,7 +198,9 @@ var nodesSubController = function($scope, Collections, System, Metrics) {
             core.collection = collection.name;
             core.shard = shard.name;
             core.shard_state = shard.state;
-            core.label = makeCoreLabel(core['collection'], core['shard'], core['replica']);
+            core.label = core['collection'] + "_"
+              + (core['shard'] + "_").replace(/shard(\d+)_/, 's\$1')
+              + core['replica'].replace(/replica_?[ntp]?(\d+)/, 'r\$1');
 
             var node_name = core['node_name'];
             var node = getOrCreateObj(node_name, nodes);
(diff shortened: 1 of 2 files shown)


solr/webapp/web/js/angular/controllers/cloud.js
```diff
@@ -198,10 +194,13 @@ var nodesSubController = function($scope, Collections, System, Metrics) {
...
             var core = replicas[replicaName];
             core.name = replicaName;
-            core.label = coreNameToLabel(core['core']);
+            core.replica = core['core'].replace(/.*_(replica_.*)$/, '\$1');
             core.collection = collection.name;
             core.shard = shard.name;
             core.shard_state = shard.state;
+            core.label = core['collection'] + "_"
+              + (core['shard'] + "_").replace(/shard(\d+)_/, 's\$1')
+              + core['replica'].replace(/replica_?[ntp]?(\d+)/, 'r\$1');
```
MarcusSorealheis, line 203: I think could be made a bit clearer as to what is going here in the GitHub description, e.g. super imposing a box on what has changed  or a before/after, which is somewhat standard for UI. Alternatively, you could show how some of the names were treated in the past and how they will be treated going forward. Otherwise, mostly looks good.

However, I know this is in maintenance mode.

MarcusSorealheis: My only question is, what is the ntp here for? I'm assuming NRT, Pull, and TLog replicas?

MarcusSorealheis (approved):
good simplification. Of course, I generally think we should move away from code like this as it is difficult to maintain, here is a bug fix.

solr/webapp/web/js/angular/controllers/cloud.js
MarcusSorealheis: Are there any reasons why we would not do any processing whatsoever of the shard names and simply show them as they were named? I assume there are some other screens or other confusion that could arise, but I am trying to doubly check that I understand the approach here fully.

janhoy: The label is for display purpose only. There is no before/after. This code produces exactly same display label as previously, but now also works for the custom shard names in implicit mode.

janhoy: Yes, `ntp` is replica type.

MarcusSorealheis: Does the test failure have anything to do with the change? It doesn't seem like it.

MarcusSorealheis: so, when implicit mode what would happen before this change?

After, it will show up. It would be helpful to know if (1) there was a particular error, (2) didn't show up at all, (3) was a broken page, etc.

epugh (approved):
LGTM.    We all know this area could be much better in the future ;-)

solr/webapp/web/js/angular/controllers/cloud.js
janhoy: > so, when implicit mode what would happen before this change?

The label would be ok, but we’d use wrong key for fetching metrics and get empty values in UI.

PR #1594, opened 2023-04-27
SOLR-16392: Convert v2 shard, replica deletion to JAX-RS 

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-16392

# Description

Solr is slowly migrating its v2 API definitions over to the JAX-RS framework.  Some APIs have already been converted, but many remain: including the APIs for replica and shard deletion.

# Solution

This PR migrates the v2 delete-replica and delete-shard APIs over to JAX-RS.  No user-visible changes were made to the v2 delete-shard endpoint.  Delete-replica however has been split into two separate endpoints (one to delete a single named replica, and one to delete a specified number of (unnamed) replicas.  The v2 APIs now appear as below:

- "Delete shard" - `DELETE /api/collections/collName/shards/shardName` (no changes)
- "Delete named replica" - `DELETE /api/collections/collName/shards/shardName/replicas/replicaName`
- "Delete multiple replicas" - `DELETE /api/collections/collName/shards/shardName/replicas?count=3`

# Tests

See `DeleteReplicaAPITest` and `DeleteShardAPITest`

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/api/JerseyResource.java
+++ b/solr/core/src/java/org/apache/solr/api/JerseyResource.java
@@ -22,6 +22,7 @@ import static org.apache.solr.jersey.RequestContextKeys.SOLR_JERSEY_RESPONSE;
 import java.util.function.Supplier;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.core.Context;
+import org.apache.solr.common.SolrException;
 import org.apache.solr.jersey.CatchAllExceptionMapper;
 import org.apache.solr.jersey.SolrJerseyResponse;
 import org.apache.solr.servlet.HttpSolrCall;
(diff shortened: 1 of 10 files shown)


HoustonPutman:
Maybe we don't count it as a delete anymore, when the action is more of a scale. So `PUT /api/collections/Cn/replicas/scale?count=123`. This would cover scaling up or scaling down replicas. If we support scaling shards, which I don't think we do, we could do the same thing under `PUT /api/collections/Cn/shards/scale`. (I don't think another API uses PUT with that shards path)

gerlowskija:
> Maybe we don't count it as a delete anymore, when the action is more of a scale. So `PUT /api/collections/Cn/replicas/scale?count=123`

Thanks for the suggestion!

I have some REST-fulness quibbles: having a "replicas" path that doesn't live under `/shards/Sn` might be confusing to users who are familiar with the `/collections/Cn/shards/Sn/replicas/Rn` pattern we've established with other APIs.  But it's definitely the best option at this point IMO!

Thinking aloud a bit, maybe we keep the idea of this being a "scaling" API, but drop the "replicas" path segment.  e.g `PUT /api/collections/Cn/scale`

Longer term, maybe this behavior could even be folded into MODIFYCOLLECTION: when someone does a "modify" (`PUT /api/collections/Cn {...}`) which happens to change `nrtReplicas`, etc, Solr would do the scaling for them.  Though of course that'd be a larger change and not something to tackle here...

PR #1595, opened 2023-04-28
SOLR-16774: Docker version requirement documented

janhoy:
https://issues.apache.org/jira/browse/SOLR-16774


janhoy:
I'm thinking about adding a TIP box in the same paragraph, to give more options:
Skjermbilde 2023-04-28 kl  23 18 48
But if major Linux distros (Ubuntu, RHEL, Debian, Suse) already default to 20.10.10 or newer then it may not be a real need?

janhoy:
Or perhaps we only need to mention the cli param?
Skjermbilde 2023-04-30 kl  02 24 18

epugh:
> Or perhaps we only need to mention the cli param? Skjermbilde 2023-04-30 kl 02 24 18

I like this one.   I wish I had known about this when running some other Java based images ;-)

PR #1597, opened 2023-04-29
SOLR-16777: Schema Designer blindly trusts potentially malicious configset

chatman:
Details in JIRA.


chatman:
This is merged now.

PR #1603, opened 2023-04-30
SOLR-16776: Disable remote streaming via system property

chatman:
Details in JIRA.

### DIFF
--- a/solr/core/src/java/org/apache/solr/servlet/SolrRequestParsers.java
+++ b/solr/core/src/java/org/apache/solr/servlet/SolrRequestParsers.java
@@ -86,7 +86,11 @@ public class SolrRequestParsers {
 
   private final HashMap<String, SolrRequestParser> parsers = new HashMap<>();
   private final boolean enableRemoteStreams;
+  public final boolean enableRemoteStreamsAtNodeLevel =
+      Boolean.parseBoolean(System.getProperty("solr.enableRemoteStreaming", "false"));
   private final boolean enableStreamBody;
+  public final boolean enableStreamBodyAtNodeLevel =
+      Boolean.parseBoolean(System.getProperty("solr.enableStreamBody", "true"));
   private StandardRequestParser standard;
   private boolean handleSelect = true;
   private boolean addHttpRequestToContext;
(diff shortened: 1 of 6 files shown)


solr/core/src/test/org/apache/solr/request/TestRemoteStreaming.java
```diff
@@ -48,14 +48,17 @@ public class TestRemoteStreaming extends SolrJettyTestBase {
 
   @BeforeClass
   public static void beforeTest() throws Exception {
+    System.setProperty("solr.enableRemoteStreaming", "true");
```
janhoy, line 51: Should add a new test case here to verify streaming fails when sysprop is false.

solr/core/src/java/org/apache/solr/servlet/SolrRequestParsers.java
```diff
@@ -86,7 +86,11 @@ public class SolrRequestParsers {
 
   private final HashMap<String, SolrRequestParser> parsers = new HashMap<>();
   private final boolean enableRemoteStreams;
+  public final boolean enableRemoteStreamsAtNodeLevel =
+      Boolean.parseBoolean(System.getProperty("solr.enableRemoteStreaming", "false"));
   private final boolean enableStreamBody;
+  public final boolean enableStreamBodyAtNodeLevel =
+      Boolean.parseBoolean(System.getProperty("solr.enableStreamBody", "true"));
```
janhoy, line 93: Do we need two different sysprops here? This `solr.enableStreamBody` is not documented in refGuide. Can't the same `solr.enableRemoteStreaming` be used for all remote streaming?

janhoy (changes requested):
Since this is a back-compat break in default values, you should also call it out in major-changes-in page for 9.3, that remote streaming now defaults to false.

I have some concern about the two layers here (node vs core), and they being named similarly. It is confusing that you can set `-Dsolr.enableRemoteStreaming=true` but still cannot do `&stream.url=foo` before you also flip the core-level setting. So I'd prefer what David proposed, to rip out the core-level switch and let the node-level be the only one.

chatman:
I don't have time to pursue this issue any further. Thanks for the feedback, @janhoy .

janhoy:
Superceded by #1615

PR #372, opened 2021-10-26
SOLR-11623 Every request handler in Solr implement PermissionNameProvider

janhoy:
https://issues.apache.org/jira/browse/SOLR-11623

This is a back-compat breaking change that likely will require lots of 3rd party request handlers out there to explicitly declare their permission requirement in order to be used with Solr 9. But I think we cannot let this be optional anymore.

I added one permission `HEALTH_PERM` for the `/node/health` and `/<coll>/ping` endpoints. That way you can easily open those to the world or connect them to a "health" role that can be given to e.g. solr-operator user. I thought of using the `METRICS_READ_PERM` but that reveals more data.

I struggle a bit with what perm to assign to our `/admin/info/system` handler, since it outputs some metrics, some config, all commandline args, some (insensitive) security info etc. I chose `CONFIG_READ_PERM` but this also means that all Admin UI users will need this permission to do almost anything, since Dashboard calls this. The same struggle goes for the `<coll>/analysis/field`, `<coll>/admin/thread` and several others. They reveal some config and sometimes some content. Please review whether things look sane.

The `ZookeeperInfoHandler` is special since it allows read of everything in Zookeeper. I tried to give it `CONFIG_READ_PERM`, unless you request `path=/security.json&detail=true`, then it will require `SECURITY_READ_PERM`. 

Do we have other endpoints in the system that reveal Zookeeper content, where security.json might leak?

PR #1615, opened 2023-05-02
SOLR-14853 enableRemoteStreaming and enableStreamBody are now sys props

dsmiley:
No longer configSet / per-core.

Existing configuration (e.g. in solrconfig.xml or via config edit API) is ignored with a deprecation warning.

Even though this is a backwards incompatible change, I think it's fine in a minor release because:
* is probably mostly being used in a local developer setting, not in production, thus backwards incompatibility is much less of a concern
* it's easy to adapt
* the user gets a warning if they don't
* existing configs or config-edit won't fail (but will no-op admittedly)
* improves security (defense-in-depth)

ref guide / changes.txt edits needed

https://issues.apache.org/jira/browse/SOLR-14853


chatman:
There are lots of XML files that need to be fixed.

ishan@x1extreme ~/code/chatman-solr (branch_9_2) $ grep enableRemoteStr -rn *|grep xml|cut -d ":" -f 1|sort|uniq
solr/core/src/test-files/solr/collection1/conf/solrconfig-analytics-query.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-collapseqparser.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-components-name.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-delaying-component.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-doctransformers.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-elevate.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-follower1.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-follower.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-hash.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-leader1-keepOneBackup.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-leader1.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-leader2.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-leader3.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-leader-throttled.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-leader.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-minhash.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-nocache.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig_perf.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-plugcollector.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-repeater.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-replication-legacy.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig-sql.xml
solr/core/src/test-files/solr/collection1/conf/solrconfig.xml
solr/modules/extraction/src/test-files/extraction/solr/collection1/conf/solrconfig.xml
solr/modules/langid/src/test-files/langid/solr/collection1/conf/solrconfig-languageidentifier.xml
solr/server/solr/configsets/_default/conf/solrconfig.xml
solr/server/solr/configsets/sample_techproducts_configs/conf/solrconfig.xml
solr/solrj/src/test-files/solrj/solr/collection1/conf/solrconfig-follower1.xml
solr/solrj/src/test-files/solrj/solr/collection1/conf/solrconfig-sql.xml
solr/solrj/src/test-files/solrj/solr/collection1/conf/solrconfig.xml
solr/solrj/src/test-files/solrj/solr/configsets/ml/conf/solrconfig.xml
solr/solrj/src/test-files/solrj/solr/configsets/shared/conf/solrconfig.xml
solr/solrj/src/test-files/solrj/solr/configsets/streaming/conf/solrconfig.xml
solr/solrj/src/test-files/solrj/solr/multicore/core0/conf/solrconfig.xml
solr/solrj/src/test-files/solrj/solr/multicore/core1/conf/solrconfig.xml
solr/solrj-streaming/src/test-files/solrj/solr/configsets/ml/conf/solrconfig.xml
solr/solrj-streaming/src/test-files/solrj/solr/configsets/streaming/conf/solrconfig.xml
solr/solr-ref-guide/modules/indexing-guide/pages/content-streams.adoc
solr/test-framework/src/test-files/solr/collection1/conf/solrconfig.xml

chatman:
> Existing configuration (e.g. in solrconfig.xml or via config edit API) is ignored with a deprecation warning.

I think they should be removed as well.

janhoy:
Did a quick browse, looks good. I agree that this change is acceptable in a minor release, for the reasons given.

@chatman, I think it makes sense to keep the core-level settings and APIs as no-ops in 9.x, declare them as deprecated in 9x and remove in main branch only.

For all the test-solrconfig.xml files I think we can do a global search/replace `<requestParsers.*?enableRemoteStreaming=".*?"` with `<requestParsers` and the same with stream body, as these are no-op anyway now.

janhoy:
We should also give these two new sysProps their own line and env.var in `solr.in.sh`? And unfortunately we'll also need some code in bin/solr[.cmd] for parsing it. I can contribute commits for this if you wish.

dsmiley:
Sure; will remove from the configs in Solr -- easy enough.

Why modify solr.in.sh at all?  I'd rather not; to be honest, I'd rather that file not exist at all (by default).

If/once Solr has a solr.env=prod|dev|test then we'd enable this setting in dev mode, and then feel no need later to give it such first class treatment to deserve it's very own env (which I assume is what motivated you to suggest an entry in solr.in.sh?).

chatman:
> If/once Solr has a solr.env=prod|dev|test then we'd enable this setting in dev mode 

@dsmiley , there's already "-Dsolr.environment=prod" in Solr.

dsmiley:
I propose in this issue not looking at solr.environment -- leave that for https://issues.apache.org/jira/browse/SOLR-15875

chatman:
> But do we want remote streaming to be enabled by default in dev/test?

Not in my opinion. Rather, I think we should remove remote streaming from Solr. If something shouldn't be used in production, it shouldn't exist in Solr.

dsmiley:
BTW I want to call attention to https://github.com/apache/solr/blob/6d5529a39fd995c207425da7f39e32721955b616/solr/core/src/java/org/apache/solr/handler/DumpRequestHandler.java#L129 
which feels weird to me.  I don't think I love varying the permission of a handler based on wether remote streaming is enabled or not.  Nor do I know why CONFIG_READ_PERM in particular was chosen.  I suppose this is here because it's super easy to have Solr fetch whatever URL for you on your behalf... but if you enabled remote streaming then, shrug, it all seems moot honestly.

chatman:
Good catch, David. I think the remote streaming "functionality" should be scrapped. DumpRequestHandler should either be scrapped or, at least, shouldn't be allowed to work with remote streaming, and the https://solr.apache.org/guide/8_11/content-streams.html page should be re-written to remove all references to such features.

janhoy:
> Nor do I know why CONFIG_READ_PERM in particular was chosen.

You and I discussed this in the original PR to secure all request handlers: https://github.com/apache/solr/pull/372/files/ae46e8d101026c3439fb23a07d1603c884048765#r737730716

janhoy:
> Why modify solr.in.sh at all? I'd rather not; to be honest, I'd rather that file not exist at all (by default).

I'll push a change that only adds parsing of `SOLR_ENABLE_REMOTE_STREAMING` and `SOLR_ENABLE_STREAM_BODY` to the scripts, but nothing in `solr.in.xx`. That way we don't advertise this feature, while still allowing Solr to pick it up from the environment, and we can document both sys.prop and env.var in refGuide.

janhoy:
I'm going to commit solrconfig.xml fixes (removing now noop params and xml comments about them)

janhoy:
Only RefGuide & Changes left to do here I believe?

dsmiley:
Yes.  Thanks for your help, BTW.  I'll be busy today but I offer this:

Proposed CHANGES.txt
* SOLR-14853: Security: Converted enableRemoteStreaming and enableStreamBody solrconfig options into system properties and env vars.  Attempts to set them the old way will no-op and log a warning.  (me, Jan, Ishan)

upgrade notes:
In order to improve security, use of stream.file and stream.url and stream.body is no longer enabled via configuration in solrconfig.xml, nor dynamic equivalents with the config API.  Instead, set an env var: ___ or ___ or system property equivalents.

solr/solr-ref-guide/modules/configuration-guide/pages/requestdispatcher.adoc
```diff
@@ -107,45 +81,11 @@ This `HttpServletRequest` is not used by any Solr component, but may be useful w
...
 ----
-<requestParsers enableRemoteStreaming="false"
-                enableStreamBody="false"
-                multipartUploadLimitInKB="2048"
+<requestParsers multipartUploadLimitInKB="2048"
                 formdataUploadLimitInKB="2048"
                 addHttpRequestToContext="false" />
 ----
 
-The below command is an example of how to enable RemoteStreaming and BodyStreaming through the xref:config-api.adoc#commands-for-common-properties[Config API]:
```
dsmiley, line 117: I prefer to not show any example on how to set any of these requestParsers details dynamically, thus I did not swap out the examples for the other still-valid configuration attributes.  IMO none of these should be in solrconfig.xml in the first place -- they are obscure things that I don't think someone would want to set on any one config (collection); instead I think all of them should be global.  Making stuff configurable has some maintenance cost and are arcane details to many people.  This is just my opinion.

dsmiley (commented):
This should do it.  I'll commit in a couple days if nobody says anything.

solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-9.adoc
```diff
@@ -74,9 +74,14 @@ Please refer to the https://solr.apache.org/downloads.html[Solr Downloads] site
 * Solr now provides an xref:deployment-guide:shard-management.adoc#installshard[INSTALLSHARD] API to allow users who have built (per-shard) indices offline to import them into SolrCloud shards.
 
 == Solr 9.3
-=== Improvements to Solr CLI
+=== Solr CLI
 * the bin/solr -i and bin/solr -info removed in favour of bin/solr status command.
 
+=== Security
+* Use of `stream.file` and `stream.url` and `stream.body` params are no longer enabled via configuration in solrconfig.xml, nor dynamic equivalents with the config API.
```
janhoy, line 81:
```suggestion
* Use of `stream.file`, `stream.url` and `stream.body` params are no longer enabled via configuration in solrconfig.xml, nor dynamic equivalents with the config API.
```

janhoy (approved):

PR #1620, opened 2023-05-02
SOLR-16470: Create V2 equivalent of V1 Replication: Get IndexVersion

mlbiscoc:

# Description

Create a V2 equivalent API for the replication "GET indexversion"

V1
`/solr/{coreName}/replication?command=indexversion`

V2
`/api/core/{coreName}/replication/indexversion`

# Solution

Created a new CoreReplicationAPI class using JAS-RX framework exposing the new V2 endpoint. The new V2 endpoint is injected with the `coreContainer` to retrieve the replicationHandler to retrieve the indexversion. Then updated the V1 endpoint to use SolrJerseyResponse. 

# Tests

Wrote new Unit test CoreReplicationAPITest that will mock a response from the replicationHandler for the new V2 api. 
# Checklist


solr/core/src/java/org/apache/solr/handler/replication/ReplicationAPIBase.java
```diff
@@ -0,0 +1,57 @@
...
+      throws IOException {
+
+    if (coreContainer.getCore(coreName) == null) {
+      throw new SolrException(
+          SolrException.ErrorCode.BAD_REQUEST, "Solr core " + coreName + " does not exist");
+    }
+
+    ReplicationHandler replicationHandler =
+        (ReplicationHandler)
+            coreContainer.getCore(coreName).getRequestHandler(ReplicationHandler.PATH);
```
mlbiscoc, line 53: @gerlowskija This is my first attempt and progress at creating a new V2 api for indexversion and would love your feedback. One of the issues I ran into was this replicationHandler being called in the V2 endpoint. 

It wasn't entirely obvious to me on how to move the replicationHandler logic into the v2 API as there were a few fields that were needed inside the replicationHandler in order to retrieve the indexversion because of race conditions. So I retrieved the handler from the core but I'm not sure if this is the best solution.

Was hoping I could get some guidance on this?

mlbiscoc:
Has anyone had a chance to review this for any fixes or improvements I could do to this PR?

gerlowskija:
Hi @mlbiscoc - thanks for the PR!  Sorry for the delay in reviewing, I didn't see it right away.  But will set it to review it now!

solr/core/src/java/org/apache/solr/handler/replication/ReplicationAPIBase.java
gerlowskija: Hmm...it'd really be ideal if we could move the logic into the v2 API class, rather than looking up the request-handler, but I agree it's not always easy or feasible to do.

I'll play with it a bit to see if I can come up with some alternative.  If I can't though, what you've done here looks 👍  to me.  (With the small quibble that we should be able to "inject" the SolrCore the same way we're injecting the SolrQueryRequest, etc. rather than going to the CoreContainer to look it up by name)

mlbiscoc: Hi Jason, thanks for the review! Glad most of it looks good.

As for your comment on injecting the SolrCore, I'm not sure how this would be done in this API. The user specifies the `coreName` for which the API then uses to retrieve the SolrCore from the coreContainer. 

Maybe I need to read more on Solr's Jersey Injections, but how would you inject the SolrCore in the constructor without knowing what `coreName` the user was asking for that is specified in the URI path?

solr/core/src/java/org/apache/solr/handler/replication/package-info.java
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/** V2 API implementations for "replication" APIs. */
+package org.apache.solr.handler.replication;
```
gerlowskija, line 19: [0] If you want to create a package specifically for the APIs we're pulling out of ReplicationHandler, that's fine by me.  But:

1.  Could you add 'api' to the package name somewhere?  (e.g. `o.a.s.handler.api.replication`). That's the convention I've tried to follow elsewhere.
2. Do you envision each ReplicationHandler API/command its own class, or do you imagine multiple of them going into CoreReplicationAPI?  If the former, maybe we should rename CoreReplicationAPI to be more specific (e.g. `GetIndexVersionAPI`?). If the latter, maybe we don't need a separate package for this code after all?

solr/core/src/test/org/apache/solr/handler/replication/CoreReplicationAPITest.java
```diff
@@ -0,0 +1,79 @@
...
+  public void testGetIndexVersion() throws Exception {
+    CoreReplicationAPI.IndexVersionResponse expected =
+        new CoreReplicationAPI.IndexVersionResponse(123L, 123L, "testGeneration");
+    when(mockReplicationHandler.getIndexVersionResponse()).thenReturn(expected);
+
+    CoreReplicationAPI.IndexVersionResponse response =
+        coreReplicationAPI.IndexVersionResponse(coreName);
+    assertEquals(response.indexVersion, expected.indexVersion);
+    assertEquals(response.generation, expected.generation);
+    assertEquals(response.status, expected.status);
```
gerlowskija, line 69: [0] Tiny nit, but I think the "expected" value is supposed to be the first argument to `assertEquals`.  (Otherwise the error message that gets produced on assertion-failure has the actual/expected values backwards.)

solr/core/src/java/org/apache/solr/handler/replication/CoreReplicationAPI.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import org.apache.solr.core.CoreContainer;
+import org.apache.solr.jersey.PermissionName;
+import org.apache.solr.jersey.SolrJerseyResponse;
+import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.response.SolrQueryResponse;
+
+@Path("/core/{coreName}/replication")
```
gerlowskija, line 36: [-1] The pattern we've established for collection/core APIs is `/cores/{coreName}/....`.  (Note that 'cores' in the first path segment is plural.)  Looks like we need an 's' at the end of the first path segment for us to stick with that pattern.

solr/solr-ref-guide/modules/deployment-guide/pages/user-managed-index-replication.adoc
```diff
@@ -378,9 +378,26 @@ http://_leader_host:port_/solr/_core_name_/replication?command=disablereplicatio
...
 
+----
+====
++
+====
+[.tab-label]*V2 API*
+
+[source,bash]
+----
+http://_host:port_/api/_core_name_/replication/indexversion
```
gerlowskija, line 396: [-1] L396 should be `http://_host:port_/api/cores/_core_name_/replication/indexversion` I think. (The "/cores" segment is missing)

gerlowskija (changes requested):
Hi @mlbiscoc - thanks again for the PR!

I left a few change-requests inline, but the overall approach looks great!

It's a shame that we're tied to some of those internal ReplicationHandler fields, but there's nothing to be done I don't think until we refactor things a bit in a follow-on PR that I'll take on.  I would've done just as you did, I think.

gerlowskija (changes requested):
Hi @mlbiscoc - thanks again for the PR!

I left a few change-requests inline, but the overall approach looks great!

It's a shame that we're tied to some of those internal ReplicationHandler fields, but there's nothing to be done I don't think until we refactor things a bit in a follow-on PR that I'll take on.  I would've done just as you did, I think.

solr/core/src/java/org/apache/solr/handler/ReplicationHandler.java
```diff
@@ -798,6 +777,38 @@ private void getFileList(SolrParams solrParams, SolrQueryResponse rsp) {
     }
   }
 
+  public CoreReplicationAPI.IndexVersionResponse getIndexVersionResponse() throws IOException {
```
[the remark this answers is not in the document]
mlbiscoc: Personally I wouldn't create public getters for those fields just for the V2 API. If at some point all the logic is moved then that should go in that PR.

But if you feel strongly about it, I can definitely move it just lmk.

solr/core/src/test/org/apache/solr/handler/replication/CoreReplicationAPITest.java
mlbiscoc: Swapped the arguments to correct assertion.

solr/core/src/java/org/apache/solr/handler/replication/CoreReplicationAPI.java
mlbiscoc: Thanks for letting me know. I changes the path so it is now `cores` instead.

solr/solr-ref-guide/modules/deployment-guide/pages/user-managed-index-replication.adoc
mlbiscoc: Ah nice catch on the missed typo by me. Fixed

solr/core/src/java/org/apache/solr/handler/replication/CoreReplicationAPI.java
```diff
@@ -0,0 +1,80 @@
...
+public class CoreReplicationAPI extends ReplicationAPIBase {
+
+  @Inject
+  public CoreReplicationAPI(
+      CoreContainer coreContainer, SolrQueryRequest req, SolrQueryResponse rsp) {
```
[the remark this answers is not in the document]
mlbiscoc: Changed to inject the `SolrCore` and register the api by overriding `registerV2`

mlbiscoc:
@gerlowskija I believe I have addressed all your comments above. Please feel free to review again if you think any other changes are necessary.

solr/core/src/java/org/apache/solr/handler/ReplicationHandler.java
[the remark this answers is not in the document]
gerlowskija: Definitely don't feel strongly about it; it's fine as-is 👍

gerlowskija:
Awesome, thanks for the quick turnaround @mlbiscoc !  Will review shortly

gerlowskija:
Thanks again @mlbiscoc - let me know if you decide to pick up any other v2 API tickets that need review!

PR #1621, opened 2023-05-03
SOLR-16784 Badapple CollectionReloadTest

janhoy:
https://issues.apache.org/jira/browse/SOLR-16784


risdenk (approved):

epugh:
I am going to merge as a badapple for now, and maybe someday if I learn more about how all this works, I can dig into the cause!   Thanks for doing this @janhoy

PR #1622, opened 2023-05-03
SOLR-16661 Remove deprecated Analytics Component in 10.0

janhoy:
https://issues.apache.org/jira/browse/SOLR-16661

Many files changed.
You may see separate commits when reviewing.


solr/solr-ref-guide/modules/deployment-guide/pages/metrics-reporting.adoc
```diff
@@ -156,6 +156,7 @@ If this section is not defined, the following default configuration is used whic
 </metrics>
 ----
 
+[#the-metrics-reporters-element]
```
janhoy, line 159: This and the similar edit in configuring-solr-xml.adoc were done due to IntelliJ throwing errors, that the asciidoc plugin could not find the anchor. So I added these explicit anchors to silence it. Antora build was apparently ok with it.

solr/core/src/java/org/apache/solr/handler/component/ResponseBuilder.java
```diff
@@ -60,7 +60,6 @@ public class ResponseBuilder {
   public boolean doExpand;
   public boolean doStats;
   public boolean doTerms;
-  public boolean doAnalytics;
```
cpoerschke, line 63: observation only and I am not aware of such a use case: if someone wanted to take the removed analytics module code and have it locally or as a package plugin or so then the removal of the `ResponseBuilder` members and setter/getters here would present a small challenge. aside from patching the equivalent of what is removed here into a local variant of the `ResponseBuilder`, it's possible to have a custom response builder and the notes on the https://issues.apache.org/jira/browse/SOLR-13511 illustrate how a custom response builder/custom search handler combination can be done and configured.

solr/solr-ref-guide/modules/query-guide/pages/analytics-expression-sources.adoc
```diff
@@ -1,95 +0,0 @@
-= Analytics Expression Sources
```
cpoerschke, line 1: Via `git grep` found "analytics" references in `dev-tools` e.g. https://github.com/apache/solr/blob/df84356a816442c6b2a7b578cb193f22b418cdae/dev-tools/scripts/refguide/new-guide.txt#L75 though not sure if the tooling would automatically handle that i.e. maybe no change needed there.

cpoerschke (commented):
328 / 328 files viewed

solr/core/src/java/org/apache/solr/handler/component/ResponseBuilder.java
janhoy: Thanks for commenting on this. I was surprised to find analytics specific vars in ResponseBuilder, and could not find any other use of these, so I removed them. If someone were to package the component as a package and maintain it on GitHub, it would be nice if this was more generic.

From SOLR-13511 I understand that you can provide your own custom ResponseBuilder if you own the SearchHandler. Fair enough, could be part of that plugin package.

Or would it be possible to have some kind of `Map<Sting,Object> responseContext` structure on ResponseBuilder where plugins could register arbitrary data? I'm on thin ice here...

solr/solr-ref-guide/modules/query-guide/pages/analytics-expression-sources.adoc
janhoy: These are data files for the 9.0 transition. The ouput of that script is served in `solr-site` repo as `.htaccess` rules, so this is not affected. Obviously we need to add new rewrite rules when 10.0 is released, for the analytics pages, so they redirect to latest 9.x and not 10.0...

cpoerschke (approved):

solr/solr-ref-guide/modules/query-guide/pages/analytics-expression-sources.adoc
HoustonPutman: I believe that antora should do a redirect to the last known version with that page, but I could be wrong...

solr/core/src/java/org/apache/solr/handler/component/ResponseBuilder.java
HoustonPutman: Yeah unfortunately this was necessary at the time, although it is very regrettable.


> From [SOLR-13511](https://issues.apache.org/jira/browse/SOLR-13511) I understand that you can provide your own custom ResponseBuilder if you own the SearchHandler. Fair enough, could be part of that plugin package.

Yeah, this seems like a pretty easy win. But a `Map<Sting,Object> responseContext` would also work pretty easily. Basically there just needs to be a way for SearchComponents to store customState in the response builder.

HoustonPutman (approved):

solr/solr-ref-guide/modules/query-guide/pages/analytics-expression-sources.adoc
janhoy: Ok, cool. We should be able to test that on the nightlies server once this is merged?

HoustonPutman: Kind of, for the nightlies, `latest` still refers to 9.2 (or the current released version), since `10.0.0` is an unreleased version. So `latest` will redirect to the 9.2 page even if the page exists in 10.0.0. We can test it locally by temporarily making `main` a released version and asking it to generate an htaccess file.

I can try that out in a bit

HoustonPutman: Ok this is apparently not a feature, so I've added an issue for antora so we can eventually support it.
https://gitlab.com/antora/antora/-/issues/1060

Hopefully this is implemented by the time Solr 10.0 is released.

janhoy: We can always continue with htaccess as earlier..

PR #1624, opened 2023-05-03
SOLR-16783: Ignore NoNodeException when deleting clusterstate.json on startup

tflobbe:
This PR addresses a race condition when deleting the clusterstate.json from multiple instances concurrently.


janhoy (approved):
LGTM

PR #1625, opened 2023-05-04
SOLR-16507: Refactor out ImplicitSnitch

noblepaul:
This is work in progress


solr/solrj-zookeeper/src/java/org/apache/solr/client/solrj/impl/MetricsFetcher.java
```diff
@@ -0,0 +1,202 @@
...
+import org.apache.solr.common.SolrException;
+import org.apache.solr.common.cloud.ZkStateReader;
+import org.apache.solr.common.params.CommonParams;
+import org.apache.solr.common.params.ModifiableSolrParams;
+import org.apache.solr.common.util.NamedList;
+import org.apache.solr.common.util.StrUtils;
+import org.apache.solr.common.util.Utils;
+import org.apache.zookeeper.KeeperException;
+
+// uses metrics API to get node information
```
dsmiley, line 22: Javadocs please.  If it was written for something in particular in mind, then it's very useful to say that, even if it may theoretically have future uses we don't know about.  It's not clear why would something use this instead of going directly to the Metrics API?  As a newcomer to this code, is it related to having short/simply well-known named "tags" (formerly maybe called snitch?) that loosely correlate to a more complex metrics expression, so that some other code can use these tags to make node placement decisions based on those same tags?

The lifecycle of an instance of this class is not obvious; please add a sentence on that.

solr/solrj-zookeeper/src/java/org/apache/solr/client/solrj/impl/MetricsFetcher.java
```diff
@@ -0,0 +1,202 @@
...
+          if (((List<?>) e.getValue()).contains(solrNode)) {
+            ctx.tags.put(key, e.getKey());
+            break;
+          }
+        }
+      }
+    }
+  }
+
+  public enum Variable {
```
dsmiley, line 177: At least a one line javadoc on what this is.  Seems similar to the other "tags" although the value varies?  Say that.

dsmiley (changes requested):
Looks like a vast improvement; thanks!

Is an objective here to remove the word "snitch"?  Unless you think it holds value, I'd be in favor of that.  I see it still used though.

noblepaul:
> Looks like a vast improvement; thanks!
> 
> Is an objective here to remove the word "snitch"? Unless you think it holds value, I'd be in favor of that. I see it still used though.

That too and get rid of a lot of classes because we just need a mechanism to fetch Per node details. Maybe we need to find a better term , but "snitch" must go

solr/solrj-zookeeper/src/java/org/apache/solr/client/solrj/impl/SolrClientNodeStateProvider.java
```diff
@@ -221,110 +217,18 @@ static void fetchReplicaMetrics(
...
-  }
-
   @Override
   public String toString() {
     return Utils.toJSONString(this);
   }
 
-  static class ClientSnitchCtx extends SnitchContext {
-    private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  static class RemoteCallCtx {
```
dsmiley, line 225: Can you please add at least one javadoc line explaining what this is for?  I don't get it.  It seems to be the "Command" software pattern?  Maybe call it "RemoteCallCmd" if so.

solr/solrj-zookeeper/src/java/org/apache/solr/client/solrj/impl/SolrClientNodeStateProvider.java
```diff
@@ -43,28 +43,24 @@
...
 /** The <em>real</em> {@link NodeStateProvider}, which communicates with Solr via SolrJ. */
 public class SolrClientNodeStateProvider implements NodeStateProvider, MapWriter {
-  public static final String METRICS_PREFIX = "metrics:";
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   private final CloudLegacySolrClient solrClient;
   protected final Map<String, Map<String, Map<String, List<Replica>>>>
       nodeVsCollectionVsShardVsReplicaInfo = new HashMap<>();
-  private Map<String, Object> snitchSession = new HashMap<>();
+  private Map<String, Object> session = new HashMap<>();
```
dsmiley, line 63: Can you please add a comment illustrating why the keys & values look like?

solr/solrj-zookeeper/src/java/org/apache/solr/client/solrj/impl/NodeValueFetcher.java
```diff
@@ -0,0 +1,203 @@
...
+import org.apache.solr.common.params.CommonParams;
+import org.apache.solr.common.params.ModifiableSolrParams;
+import org.apache.solr.common.util.NamedList;
+import org.apache.solr.common.util.StrUtils;
+import org.apache.solr.common.util.Utils;
+import org.apache.zookeeper.KeeperException;
+
+/**
+ * This class is responsible for fetching metrics and other attributes from a given node in Solr
+ * cluster. This is a helper class that is used by {@link SolrClientNodeStateProvider}
```
dsmiley, line 41: Still looking for a lifecycle like bit of documentation.   It has no state; right?  In that case, all methods should probably be static and we say this is a "utility" like class and document it's not instantiated.

solr/solrj-zookeeper/src/java/org/apache/solr/client/solrj/impl/NodeValueFetcher.java
```diff
@@ -0,0 +1,203 @@
...
+        Number num = (Number) val;
+        return num.doubleValue();
+
+      } else {
+        throw new IllegalArgumentException("Unknown type : " + val);
+      }
+    }
+  }
+
+  protected void getRemoteInfo(
```
dsmiley, line 103: is not subclassed; should be private

solr/solrj-zookeeper/src/java/org/apache/solr/client/solrj/impl/NodeValueFetcher.java
dsmiley: Can you please respond to my comment @noblepaul ?

Before merging PRs, please wait for either an explicit approval or announce a 2-day plan to merge ("lazy consensus").

noblepaul: > It has no state; right? In that case, all methods should probably be static 

Can it be made static? Yes, But we could potentially add some state later. That's why I left it as is. However, please feel free to make any change to this. That's why I made this branch is the public repo instead of my own

PR #1627, opened 2023-05-05
SOLR-16273: Prometheus Metric Exporter is very slow when collecting large amounts of sample data

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-16273

# Description

MetricFamilySamples from prometheus contains a List of samples and a "contains" check is called to avoid added duplicate metrics. With large lists, this reduces the prometheus exporters performance with large amounts of collections or lots of metrics scraping

# Solution

Add in a Hashset into MetricSamples.java as a cache and cache the whole metric as a string and check against the cache instead of checking against the List with contains.

# Tests

Update unit tests where cache is implemented and verify no duplicate samples were added to a metric family.

# Checklist


mlbiscoc:
Hi everyone, this should be ready for review now.

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/collector/MetricSamples.java
```diff
@@ -19,24 +19,42 @@
...
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 public class MetricSamples {
 
   private final Map<String, Collector.MetricFamilySamples> samplesByMetricName;
 
+  private Set<String> sampleMetricsCache;
```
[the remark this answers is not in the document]
mlbiscoc: Changed into final. Thanks!

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/collector/MetricSamples.java
```diff
@@ -47,7 +65,8 @@ public void addSampleIfMetricExists(
       return;
     }
 
-    if (!sampleFamily.samples.contains(sample)) {
+    if (!sampleMetricsCache.contains(sample.toString())) {
+      sampleMetricsCache.add(sample.toString());
```
cpoerschke, line 69: Thanks @mlbiscoc for explaining the issue in the PR description, I'm new to this part of the code base and the explanation really helped!

So here the `sampleFamily.samples.contains` is inefficient for a large list and the idea is that a `HashSet.contains` would be more efficient.

It seems that `sample.toString()` doesn't include the `exemplar` member -- https://github.com/prometheus/client_java/blob/parent-0.16.0/simpleclient/src/main/java/io/prometheus/client/Collector.java#L256-L292 -- i.e. comparison of `sample` vs. comparison via `sample.toString()` could theoretically differ? Hmm.

mlbiscoc: No problem! I am also still learning this code base as well but please ask anything if it needs clarification.

Exactly, this PR is to address the fact that we are searching in `O(n)` but we can change that to `O(1)`. Testing locally, I found significant performance increase.

As for the `.toString` not including `exemplar`, I felt it was a non-issue because when we scrape the metrics from Solr, the prometheus exporter creates the sample object, but with `timestamp and exemplar = null`. So there will never be a string with a `timestamp` or `exemplar`. You can see here -> https://github.com/apache/solr/blob/7a31a1e23f6034c14639d7a208c5aaec64608cfd/solr/prometheus-exporter/src/java/org/apache/solr/prometheus/scraper/SolrScraper.java#L205

mlbiscoc:
Hello, is there any other additional comments or suggestions to review for improving this PR?

mlbiscoc:
@dsmiley I noticed you've done some work with the prometheus exporter. Would you have any input for these changes?

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/collector/MetricSamples.java
```diff
@@ -19,24 +19,42 @@
...
+    }
   }
 
   public MetricSamples() {
     this(new HashMap<>());
   }
 
-  public void addSamplesIfNotPresent(String metricName, Collector.MetricFamilySamples samples) {
-    samplesByMetricName.putIfAbsent(metricName, samples);
+  private void addSamplesToCache(Collector.MetricFamilySamples metricFamilySamples) {
```
dsmiley, line 46: The diff was confusing to read here.  I think it would be clearer if you move this method to *below* the method that calls it; not before.

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/collector/MetricSamples.java
```diff
@@ -19,24 +19,42 @@
...
+    for (Collector.MetricFamilySamples.Sample sample : metricFamilySamples.samples) {
+      sampleMetricsCache.add(sample.toString());
+    }
+  }
+
+  public void addSamplesIfNotPresent(
+      String metricName, Collector.MetricFamilySamples metricFamilySamples) {
+    if (!samplesByMetricName.containsKey(metricName)) {
+      samplesByMetricName.put(metricName, metricFamilySamples);
```
dsmiley, line 55: There used to be a putIfAbsent call here, which is more elegant (and efficient) than a check and put.  You can still use that; just rely on the return value of putIfAbsent being null to check if you then should call addSamplesToCache.  If you think it's still unclear, you could add a comment.

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/collector/MetricSamples.java
dsmiley: Again, you could use a sampleMetricCache.add(sample.toString()) and look at the response boolean to know if it was added (wasn't already present).  Is more efficient and moreover avoids a second sample.toString() call easily.

Why is sampleMetricsCache using the toString of the Sample at all; why isn't it simply `Set<Sample>`?  Also; sampleMetricsCache isn't a cache; it's not the ideal name/word for it.  I would call it "seenSamples".

dsmiley (commented):
BTW I observe the only test dependent on this deduplication is the unit test you added.  Can you consider tweaking a higher level test?  Not critical.

mlbiscoc:
@dsmiley Thanks for reviewing with suggestions. I made the changes accordingly.
Only part I did not change was for adding a higher level test.

I also thought about doing this earlier in development but getting duplicate metrics is caused from more than 1 collection. The current configuration for CloudScraperTest is using 1 collection so I would need to refactor all the tests with another collection and anything using that configuration so I felt that a unit test to catch the duplication was suffice and easiest to test for without potentially breaking tests.

I know you said it is not critical but if you feel strong on adding a high level test to catch duplication, I can go forward with that doing that.

dsmiley (approved):
> getting duplicate metrics is caused from more than 1 collection

I had no idea this is why; I'm not familiar with this part of the exporter.  Just thinking about this on the surface (not diving through code details), that explanation is suspicious -- why would different collections have the same metrics?  If this deduplication mechanism was absent; would the exporter return the same dame data (but perhaps internally operate more efficiently?) or does this prevent more data from being returned?  Higher level tests would help answer that (and I'm not insisting you write one; it's okay).

Perhaps the "seenSamples" field should have a comment to reflect this?  Or not -- as you wish.

Please add a CHANGES.txt entry under the "Optimization" section.  A simple / brief sentence is fine; no details.

PR #1628, opened 2023-05-05
SOLR-16789: Admin UI Dashboard shows "unauthenticated" warning on authenticated servers

tflobbe:
I'm not sure this is the right fix, I don't trust my UI skills. I believe this was introduced in https://github.com/apache/solr/pull/1296

### DIFF
--- a/solr/webapp/web/js/angular/controllers/index.js
+++ b/solr/webapp/web/js/angular/controllers/index.js
@@ -29,7 +29,7 @@ solrAdminApp.controller('IndexController', function($scope, System, Cores, Const
         sessionStorage.setItem("auth.username", data.security.username);
       }
 
-      if ("authenticationPlugin" in data.security && data.security.authenticationPlugin != null) {
+      if (data.security.authenticationPlugin) {
         $scope.isSecurityEnabled = true
       }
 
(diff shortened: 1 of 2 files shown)


solr/webapp/web/js/angular/controllers/index.js
```diff
@@ -29,6 +29,10 @@ solrAdminApp.controller('IndexController', function($scope, System, Cores, Const
         sessionStorage.setItem("auth.username", data.security.username);
       }
 
+      if ("authenticationPlugin" in data.security && data.security.authenticationPlugin != null) {
```
epugh, line 32: I'd have to see if we ahve an established pattern for these if's!   `if data.security.authenticationPlugin ` migth work...

tflobbe: Yes, that worked for me

janhoy (approved):
Have not reproduced or tested the fix, but if it works, it's probably right :)

tflobbe:
Ran these steps:
`gw clean  -p solr/packaging dev && cd  solr/packaging/build/dev`
`bin/solr start -e cloud -noprompt`
Checked that the warning was present in the dashboard
`bin/solr auth enable -credentials foo:bar -z localhost:9983`
On `main`, the warning is still present. With the patch I get:
Screenshot 2023-05-05 at 2 38 27 PM

PR #1632, opened 2023-05-08
SOLR-14886 : suppress stack traces in query response

igiguere:
https://issues.apache.org/jira/browse/SOLR-14886

# Description

Stack traces are considered a security risk.  Please refer to OWASP for a full explanation:

https://owasp.org/Top10/A05_2021-Security_Misconfiguration/

# Solution

Add a property "hideStackTrace" to solr.xml.  In NodeConfig, the default value is "false", for back-compatibility.

Use the new property in ResponseUtils, to print out, or not, the stack trace.

Adapt code that calls ResponseUtils.

Also check if stack traces should be hidden everywhere the "trace" element is added to a response.  (I searched for the string"trace" - with quotes - throughout the Solr code.  Hope I got everything.)

# Tests

org.apache.solr.servlet.HideStackTraceTest : force an exception into the response, and assert that if hideStackTrace=true, the stack trace is not shown.

# Checklist


solr/core/src/java/org/apache/solr/servlet/ResponseUtils.java
```diff
@@ -70,10 +90,13 @@ public static int getErrorInfo(Throwable ex, NamedList<Object> info, Logger log)
...
     // For any regular code, don't include the stack trace
     if (code == 500 || code < 100) {
-      StringWriter sw = new StringWriter();
-      ex.printStackTrace(new PrintWriter(sw));
-      log.error("500 Exception", ex);
-      info.add("trace", sw.toString());
+      // hide all stack traces, as configured
+      if (!hideStackTrace) {
+        StringWriter sw = new StringWriter();
+        ex.printStackTrace(new PrintWriter(sw));
```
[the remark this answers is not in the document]
igiguere: This PR aims at mitigating "information exposure through an error message".  With this PR, the stack trace will only be output if Solr is configured to output stack traces.
The PR does not, however, attempt to fix pre-existing code, such as this call to "printStackTrace"

solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
```diff
@@ -971,7 +975,11 @@ protected void writeResponse(
 
       if (solrRsp.getException() != null) {
         NamedList<Object> info = new SimpleOrderedMap<>();
-        int code = ResponseUtils.getErrorInfo(solrRsp.getException(), info, log);
+        boolean hideStackTrace =
+            core != null
```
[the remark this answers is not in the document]
igiguere: HttpSolrCall.init() is called during initialization.  It will not race with HttpSolrCall.writeResponse(...).  There is no response to be written before or during initialization.

solr/core/src/java/org/apache/solr/servlet/ResponseUtils.java
```diff
@@ -70,10 +90,13 @@ public static int getErrorInfo(Throwable ex, NamedList<Object> info, Logger log)
...
+        log.error("500 Exception", ex);
+        info.add("trace", sw.toString());
```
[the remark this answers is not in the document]
igiguere: Good catch.  Thanks, I'll do that

solr/core/src/java/org/apache/solr/rest/BaseSolrResource.java
```diff
@@ -143,7 +143,11 @@ protected void handleException(Logger log) {
     Exception exception = getSolrResponse().getException();
     if (null != exception) {
       NamedList<Object> info = new SimpleOrderedMap<>();
-      this.statusCode = ResponseUtils.getErrorInfo(exception, info, log);
+      boolean hideStackTrace =
+          solrCore != null
+              ? solrCore.getCoreContainer().hideStackTrace()
+              : Boolean.parseBoolean(System.getProperty("solr.hideStackTrace"));
```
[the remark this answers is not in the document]
igiguere: addressed in commit cd8459605081f993ba6e16e5d0f631b3b4de60be

solr/core/src/java/org/apache/solr/servlet/ResponseUtils.java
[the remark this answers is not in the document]
igiguere: addressed in commit cd8459605081f993ba6e16e5d0f631b3b4de60be

solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
```diff
@@ -971,7 +971,7 @@ protected void writeResponse(
 
       if (solrRsp.getException() != null) {
         NamedList<Object> info = new SimpleOrderedMap<>();
-        int code = ResponseUtils.getErrorInfo(solrRsp.getException(), info, log);
+        int code = ResponseUtils.getErrorInfo(solrRsp.getException(), info, log, core);
```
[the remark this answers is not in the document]
igiguere: @sonatype-lift ignore

HttpSolrCall.init() is called during initialization. It will not race with HttpSolrCall.writeResponse(...). There is no response to be written before or during initialization.

solr/core/src/java/org/apache/solr/servlet/ResponseUtils.java
```diff
@@ -42,6 +43,26 @@ private ResponseUtils() {}
...
+   * added to the given NamedList.
+   *
+   * <p>Status codes less than 100 are adjusted to be 500.
+   *
+   * <p>Stack trace will not be output if hideTrace=true OR system property
+   * solr.hideStackTrace=true.
+   *
+   * @see #getTypedErrorInfo(Throwable, Logger)
+   */
+  public static int getErrorInfo(Throwable ex, NamedList<Object> info, Logger log, SolrCore core) {
```
stillalex, line 65: could you replace 'core' instance with just the hideStackTrace flag? there is no reason to pass in the entire core.

solr/core/src/java/org/apache/solr/servlet/ResponseUtils.java
```diff
@@ -96,6 +120,22 @@ public static int getErrorInfo(Throwable ex, NamedList<Object> info, Logger log)
...
+   * <p>Primarily used by v2 API code, which can handle such typed information.
+   *
+   * <p>Status codes less than 100 are adjusted to be 500.
+   *
+   * <p>Stack trace will not be output if hideTrace=true OR system property
+   * solr.hideStackTrace=true.
+   *
+   * @see #getErrorInfo(Throwable, NamedList, Logger)
+   */
+  public static ErrorInfo getTypedErrorInfo(Throwable ex, Logger log, SolrCore core) {
```
stillalex, line 138: same as above. pass the boolean only

solr/core/src/java/org/apache/solr/servlet/ResponseUtils.java
```diff
@@ -135,4 +177,10 @@ public static ErrorInfo getTypedErrorInfo(Throwable ex, Logger log) {
     errorInfo.code = code;
     return errorInfo;
   }
+
+  private static boolean hideStackTrace(SolrCore core) {
+    return core != null
+        ? core.getCoreContainer().hideStackTrace()
+        : Boolean.parseBoolean(System.getProperty("solr.hideStackTrace"));
```
stillalex, line 184: no need to check the system property value every time you need to log an error. you can move this to the init phase, in the NodeConfig maybe? and pass that value to the log method.

igiguere: @stillalex  : Thanks for your comments.  For some reason, the unit test in this PR is now failing, before I even try to modify anything.  Is anyone aware of changes in the code path of a response that would prevent going though these methods of ResponseUtils?  I'll keep trying, but, no luck after a couple of hours so far.

solr/core/src/java/org/apache/solr/servlet/ResponseUtils.java
igiguere: Done in latest commit.  Thanks.

solr/core/src/java/org/apache/solr/servlet/ResponseUtils.java
igiguere: Something similar was done in latest commit.  Thanks.

solr/core/src/java/org/apache/solr/servlet/ResponseUtils.java
igiguere: Done in latest commit. Thanks.

solr/core/src/java/org/apache/solr/servlet/ResponseUtils.java
```diff
@@ -28,6 +28,11 @@
 public class ResponseUtils {
   private ResponseUtils() {}
 
+  // System property to use if the Solr core does not exist or solr.hideStackTrace is not
+  // configured. (i.e.: a lot of unit test).
+  private static final boolean SYSTEM_HIDE_STACK_TRACES =
```
stillalex, line 33: this can be `Boolean.getBoolean("solr.hideStackTrace")`

solr/core/src/test/org/apache/solr/servlet/HideStackTraceTest.java
```diff
@@ -0,0 +1,151 @@
...
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.w3c.dom.Document;
+
+/** SOLR-14886 : Suppress stack trace in Query response */
+@SuppressSSL
+public class HideStackTraceTest extends SolrJettyTestBase {
```
stillalex, line 58: I don't think you need to write as much code to boot a Solr. take a look at https://github.com/apache/solr/blob/473cea28b55dcf03fb2338399af321f425c90002/solr/core/src/test/org/apache/solr/handler/TestRequestId.java#L36 for example (there are many examples around)

also `@SuppressSSL` is not needed here

solr/core/src/java/org/apache/solr/core/NodeConfig.java
```diff
@@ -567,6 +575,7 @@ public static class NodeConfigBuilder {
     private String defaultZkHost;
     private Set<Path> allowPaths = Collections.emptySet();
     private List<String> allowUrls = Collections.emptyList();
+    private boolean hideStackTrace = false;
```
stillalex, line 578: the default should be the system property value I think. this way it's consistent with the ResponseUtils implementation. (if you only set the system property, both code paths should behave the same)

solr/core/src/test/org/apache/solr/servlet/HideStackTraceTest.java
igiguere: I'm having too much trouble with that unit test (on Windows) if SSL is enabled.  And SSL is not needed to demonstrate the functionality.

solr/core/src/java/org/apache/solr/servlet/ResponseUtils.java
igiguere: done in latest commit

solr/core/src/test/org/apache/solr/servlet/HideStackTraceTest.java
igiguere: test initialization simplified in the latest commit.  But keeping @SuppresSSL: refer to my [previous comment|https://github.com/apache/solr/pull/1632#discussion_r1303381080].

solr/core/src/java/org/apache/solr/core/NodeConfig.java
igiguere: done in latest commit

solr/core/src/java/org/apache/solr/servlet/ResponseUtils.java
```diff
@@ -135,4 +182,8 @@ public static ErrorInfo getTypedErrorInfo(Throwable ex, Logger log) {
     errorInfo.code = code;
     return errorInfo;
   }
+
+  private static boolean hideStackTrace(final Boolean hideTrace) {
```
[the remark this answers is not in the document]
igiguere: It took a bit of refactoring elsewhere, but yep, done in latest commit.

solr/core/src/java/org/apache/solr/core/NodeConfig.java
stillalex: hmm, this doesn't look right :) it will now ignore the builder value. line 616 needs to be 

```
private boolean hideStackTrace = Boolean.getBoolean("solr.hideStackTrace");
```

and line 190 needs to be as before
```
this.hideStackTraces
```

igiguere: Wow!  Thanks.  I don't know what I was thinking ;)

dsmiley (commented):
While I can understand that Solr needs NodeConfig / solr.xml, this feels way more like a simple Java system property situation.  We certainly have many of the latter.

Maybe someday we would have a unification in a generic way so that we needn't go through all the NodeConfig plumbing just to have some global flag.

stillalex:
@igiguere could you rebase on top of main branch (as suggested on the dev list)?
also what name should I use in the CHANGES.txt entry for the credit?

igiguere:
@stillalex  :  I'm in CHANGES.txt a couple of times with my full name 
Isabelle Giguere

igiguere:
I rebased this branch on the fork's main (up to date with /solr main).  There were no changes to push, so I don't know if a build will be triggered.
Hopefully, it will work.

stillalex:
I'm not seeing a new build, I think you have to force push the rebase so the build can apply correctly

solr/core/src/test/org/apache/solr/servlet/HideStackTraceTest.java
```diff
@@ -0,0 +1,131 @@
...
+    afterSolrJettyTestBase();
+    if (solrHome != null) {
+      cleanUpJettyHome(solrHome.toFile());
+    }
+  }
+
+  @Test
+  public void testHideStackTrace() throws Exception {
+    final String url = jetty.getBaseUrl().toString() + "/collection1/withError?q=*:*&wt=xml";
+    final HttpGet get = new HttpGet(url);
```
dsmiley, line 72: Does this test need to work with HTTP at all?  I believe not; the stack trace should be in the response structure.  See EmbeddedSolrServerTestRule -- a new test mechanism for a new breed of tests that can work with a SolrClient.

dsmiley (commented):
See org.apache.solr.search.json.TestJsonRequestWithEdismaxDefType to see a new breed of tests.  Notice the SolrClientTestRule.  Notice a ConfigRequest is used to manipulate a configuration instead of creating yet another config file.  We have too many such config files; it's a maintenance problem.  Feedback on this is welcome!

stillalex:
@dsmiley I am attempting to rewrite the test using thew new `EmbeddedSolrServerTestRule` but I am falling short.
First it seems to shortcut exceptions in the response via `checkForExceptions(rsp);` method so there is no opportunity for the stacktrace to be populated in the response, what I see now is just the exception being thrown so cannot verify the `hideStackTrace` flag does anything.
Second it seems to hardcode the response to the binary response (see `BinaryResponseWriter.getParsedResponse`), so not sure if this might need to change to support other types, for this test it is needed to serialize the error to verify the trace is not there.
please correct if I am missing something obvious here.

dsmiley:
Thanks for trying Alex!
I suppose this test is more sensitive to the serialization, in which case you should probably use SolrJettyTestRule.

EmbeddedSolrServer:  Ideally, a test (or more likely the test infrastructure behind the scenes) could pick the response format if it wants, say, XML or JSON instead of the parsed NamedList, such as if assertJQ (JSON path custom thing) or assertQ (XPath).  The InputStreamResponseParser may be needed, which is a bit of a special thing (see Find-Usages to see for yourself).

stillalex:
@dsmiley I refactored the test as best I could. please take a look and let me know if more cleanup is needed.
I had to rebase+force push to get crave build running again

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -301,6 +301,8 @@ && getZkController().getOverseer() != null
 
   private final AllowListUrlChecker allowListUrlChecker;
 
+  private final boolean hideStackTrace;
```
dsmiley, line 304: I don't think a simple global boolean needs to be yet another field on CoreContainer.  Even the NodeConfig stuff is overkill for what could be a system property.

If we do keep NodeConfig, we could still do away with this field because the NodeConfig stays around as a field on CoreContainer as `cfg`.

solr/core/src/java/org/apache/solr/servlet/ResponseUtils.java
```diff
@@ -28,6 +28,10 @@
 public class ResponseUtils {
   private ResponseUtils() {}
 
+  // System property to use if the Solr core does not exist or solr.hideStackTrace is not
+  // configured. (i.e.: a lot of unit test).
+  private static final boolean SYSTEM_HIDE_STACK_TRACES = Boolean.getBoolean("solr.hideStackTrace");
```
dsmiley, line 33: This is all we need for internal config IMO.

solr/core/src/test/org/apache/solr/servlet/HideStackTraceTest.java
```diff
@@ -0,0 +1,175 @@
...
+ * limitations under the License.
+ */
+package org.apache.solr.servlet;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.util.EntityUtils;
```
dsmiley, line 24: As a project, we'd like to get away from needless use of Apache HttpClient.  We can use JDK HttpClient here.

dsmiley:
It's sad to see so much ceremony in NodeConfig :-( -- an issue for another PR.

stillalex:
having to rebase + force commit to get crave build unstuck is becoming a real pain

dsmiley (approved):

PR #1633, opened 2023-05-09
SOLR-16691 :Use Jackson for JSON serialization (2)

noblepaul:


dizzu:
@noblepaul I'm seeing some very weird formatting using the default interface query output (wt=json/default and indent=true) in SolrCloud mode. Either all the indented content is shifted by many white spaces (while keeping indents after this), either there is not indenting at all (keeping multiple lines, different from indent=false), either we get a new line between every line. Why was this made the instant default and not tested as an alternative in at least one release? 😕

PR #1637, opened 2023-05-10
SOLR-16766: Remove -usage as option, and ensure both -help and -h work

bszabo97:
https://issues.apache.org/jira/browse/SOLR-16766

# Description

Nowhere do we mention `-usage` as a parameter, and it's not one that I ever use. `-help` is common, as is `-h`, so make sure that works everywhere. There are places where `-h` doesn't work!

# Solution

Deleted appearances of `-usage` and added `-h` where it was not present. Also made some "beautification" here and there to make the help more consistent and complete.

# Tests

# Checklist


epugh (approved):
LGTM

epugh:
in the not too distant future, I think all that detailed "help" information is going to move to the Ref Guide, and we'll only have the help that the commons-cli (or the other tool that @elyograg found) being generated in bin/solr -h...    That way we can put all the tips and tricks and examples in one place, the Ref Guide!

bszabo97:
I was a bit uncertain to add all of the options to the ref guide as well (for example all the `-verbose` options for every command). But if I understand correctly you would suggest to add all of these options and descriptions (where it is more detailed) to the ref guide?

epugh:
> I was a bit uncertain to add all of the options to the ref guide as well (for example all the `-verbose` options for every command). But if I understand correctly you would suggest to add all of these options and descriptions (where it is more detailed) to the ref guide?

I think so....    We should point users to the ref guide for detailed help information when using our various bin/solr XXX tools...    Otehrwise, what we see is that we duplicate content and it gets out of alignment...

bszabo97:
So you suggest to add everything to the ref guide and change the help of the tools to a link which points the users to the corresponding ref guide paragraph instead of having the whole description duplicated?

PR #1647, opened 2023-05-16
Make small improvements to integration tests

HoustonPutman:
- Fix some of the test names
- Add timing information to test output
- Fix `test_bats.bats` so that it waits for the server to start before trying to stop it. This should probably be fixed somewhere else, but I tried a lot of things and none helped.
- `test_create.bats` is faster by removing an unnecessary `delete_all_collections`

### DIFF
--- a/solr/packaging/test/test_placement_plugin.bats
+++ b/solr/packaging/test/test_placement_plugin.bats
@@ -37,7 +37,7 @@ teardown() {
   assert_file_contains "${SOLR_LOGS_DIR}/solr.log" 'Default replica placement plugin set in solr\.placementplugin\.default to affinity'
 }
 
-@test "Affinity placement plugin using ENV" {
+@test "Random placement plugin using ENV" {
   export SOLR_PLACEMENTPLUGIN_DEFAULT=random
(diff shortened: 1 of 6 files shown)


risdenk (approved):

epugh (approved):
LGTM.    There are some places wehre we have a `run` but I don't think we actually look at the captured content...

solr/packaging/test/test_placement_plugin.bats
```diff
@@ -37,7 +37,7 @@ teardown() {
   assert_file_contains "${SOLR_LOGS_DIR}/solr.log" 'Default replica placement plugin set in solr\.placementplugin\.default to affinity'
 }
 
-@test "Affinity placement plugin using ENV" {
+@test "Random placement plugin using ENV" {
   export SOLR_PLACEMENTPLUGIN_DEFAULT=random
   run solr start -c
```
epugh, line 42: do we need the `run` ?

HoustonPutman: probably not...

PR #1648, opened 2023-05-17
SOLR-16804 Simplify testing modified dependencies.

gus-asf:


HoustonPutman (approved):
Makes sense to me. Also makes sense why the lucene-dev-path option worked, but the mavenLocal() option didn't.

gus-asf:
With one +1 via review and no other feedback after a few days, merging

PR #1655, opened 2023-05-22
SOLR-16812: Support CBOR format for update/query

noblepaul:


solr/solrj/src/java/org/apache/solr/client/solrj/request/GenericSolrRequest.java
```diff
@@ -42,6 +44,22 @@ public GenericSolrRequest setContentWriter(ContentWriter contentWriter) {
     return this;
   }
 
+  public GenericSolrRequest withContent(byte[] buf, String type) {
```
mkhludnev, line 47: Isn't it too harmful to be advertised as public for users?

mkhludnev (approved):

solr/solrj/src/java/org/apache/solr/client/solrj/request/GenericSolrRequest.java
noblepaul: I don't think so.
Users can already do so with their own `ContentWriter` impl. This is just a helper method

solr/core/src/test/org/apache/solr/util/TestCborDataFormat.java
```diff
@@ -0,0 +1,228 @@
...
+import org.apache.solr.cloud.SolrCloudTestCase;
+import org.apache.solr.common.params.MapSolrParams;
+import org.apache.solr.common.util.JavaBinCodec;
+import org.apache.solr.common.util.NamedList;
+import org.apache.solr.common.util.SuppressForbidden;
+import org.apache.solr.common.util.Utils;
+import org.apache.solr.handler.loader.CborStream;
+import org.apache.solr.response.XMLResponseWriter;
+
+public class TestCborDataFormat extends SolrCloudTestCase {
```
dsmiley, line 54: IMO our request/response formats, to include this new one, should be tested in a general way without a dedicated test, especially not a SolrCloud test.  For example, we randomly pick compatible substitutes for the codec.  Shouldn't we do the same for the underlying "wt" encoding too?

solr/core/src/java/org/apache/solr/handler/loader/CborStream.java
```diff
@@ -0,0 +1,150 @@
...
+import org.apache.solr.common.SolrInputDocument;
+import org.apache.solr.common.params.SolrParams;
+import org.apache.solr.common.params.UpdateParams;
+import org.apache.solr.common.util.ContentStream;
+import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.response.SolrQueryResponse;
+import org.apache.solr.update.AddUpdateCommand;
+import org.apache.solr.update.processor.UpdateRequestProcessor;
+
+public class CborStream {
```
dsmiley, line 37: javadoc on public classes

solr/core/src/java/org/apache/solr/response/CborResponseWriter.java
```diff
@@ -0,0 +1,189 @@
...
+import com.fasterxml.jackson.dataformat.cbor.CBORGenerator;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import org.apache.solr.client.solrj.impl.BinaryResponseParser;
+import org.apache.solr.common.util.NamedList;
+import org.apache.solr.request.SolrQueryRequest;
+
+public class CborResponseWriter extends BinaryResponseWriter {
```
dsmiley, line 29: javadoc on public classes

solr/core/src/java/org/apache/solr/handler/loader/CborStream.java
dsmiley: Since this appears to be an internal class; maybe comment that this is so, and mention who uses it.

solr/core/src/java/org/apache/solr/handler/UpdateRequestHandler.java
```diff
@@ -146,6 +147,7 @@ protected Map<String, ContentStreamLoader> createDefaultLoaders(NamedList<?> arg
     registry.put("application/json", new JsonLoader().init(p));
     registry.put("application/csv", new CSVLoader().init(p));
     registry.put("application/javabin", new JavabinLoader(instance).init(p));
+    registry.put("application/cbor", CborStream.createLoader(p));
```
dsmiley, line 150: Looking at nearby lines of code, this suggests that perhaps CborStream should be called "CborLoader"?

dsmiley (commented):
Cool to see this new format!

solr/core/src/test/org/apache/solr/util/TestCborDataFormat.java
```diff
@@ -0,0 +1,221 @@
...
+    try {
+      System.setProperty("managed.schema.mutable", "true");
+      CloudSolrClient client = cluster.getSolrClient();
+      CollectionAdminRequest.createCollection(testCollection, "conf", 1, 1).process(client);
+      modifySchema(testCollection, client);
+
+      byte[] b =
+          Files.readAllBytes(
+              new File(ExternalPaths.SOURCE_HOME, "example/films/films.json").toPath());
+      // every operation is performed twice. We should only take the second number
```
gerlowskija, line 73: [-0] Is this class a test meant to validate correctness, or a benchmark for measuring performance?  I think both are needed and valuable; it's great you thought of both.  But I think it's a mistake to combine the two in a single class.  Especially when we have a whole "benchmark" module built for exactly the sort of measurements you're taking.  It offers tools (e.g. JMH annotations, data generators) that make it easy to test performance much more robustly than is feasible to do here.  (See my comment below).

[-1] It's a good start, but I think there's a handful of problems with how this JUnit method gathers numbers around CBOR queries, that make it misleading for gathering even anecdotal performance data.

I'm sure you've done other benchmarking at this point to prove out the CBOR idea, but so far these are the only numbers the rest of the community has seen.  So we need to make sure they're rock solid!  In particular, here are a few of my concerns:

1. Even for tiny methods called in a tight loop, JVMs often take more than n=2 iterations to optimize bytecode.  The 'N' here is just too small to assume the JVM is fully "hot" on the ser-de code for each format.  Even ignoring JVM optimizations, n=2 is too small to rule out random noise (other processes competing for CPU, JVM GC, etc.) on the host machine.
2. The order in which you're testing formats likely skews results.  e.g. The two JSON queries run on an almost entirely cold JVM, whereas CBOR queries run when the JVM, OS caches, Solr caches, etc. are considerably warmer.
3. "films.json" is pretty small as a data set and homogenous in terms of a few field-types.  I worry that would make the gathered data unrepresentative and noisy.

I think all three of these would be easy to address if the code was moved to the benchmarking module.  It has data generators for building larger and more diverse datasets.  JMH has a lot of support built in that make it easy to configure "Warming" iterations that get ignored in the collected statistics. etc.

solr/core/src/test/org/apache/solr/util/TestCborDataFormat.java
```diff
@@ -0,0 +1,221 @@
...
+  private byte[] runQuery(String testCollection, CloudSolrClient client, String wt)
+      throws SolrServerException, IOException {
+    NamedList<Object> result;
+    QueryRequest request;
+    RTimer timer = new RTimer();
+    request = new QueryRequest(new SolrQuery("*:*").setRows(1111));
+    request.setResponseParser(new InputStreamResponseParser(wt));
+    result = client.request(request, testCollection);
+    byte[] b = copyStream((InputStream) result.get("stream"));
+    System.out.println(wt + "_time : " + timer.getTime());
```
gerlowskija, line 122: [Q] So, as I read this code - `runQuery` will capture CBOR serialization in its perf numbers but **not** deserialization since we're not deserializing the received data on the client side.

Am I reading that right?  And if so, is that intentional?

[0] Also, doesn't forbiddenApis complain about the `System.out.println` usage here?  I swear that was one of the things we disallowed, though maybe we're looser about that in our tests...

gerlowskija (commented):
The CBOR serialization code itself looks OK, but the test/benchmarking code could really use some improvement IMO.

I'm also a little concerned about what's **not** in this PR.  In particular: documentation anywhere about the new format.  And, less crucially, deprecation for javabin (or guidance around when users should use javabin vs cbor).

PR #1661, opened 2023-05-24
SOLR-16816: Update node metrics when making affinityPlacement selections

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-16816

This is not backcompat for people using custom PlacementPlugin implementations, but we could just use a NO-OP for `increase()` and `decrease()` if that is an issue.

Unfortunately I had to refactor the built in metrics, as you can't use a subclass as a static variable in the superclass.


HoustonPutman:
Tests added!

tflobbe (approved):
> This is not backcompat for people using custom PlacementPlugin implementations, but we could just use a NO-OP for increase() and decrease() if that is an issue.

I think this could work, maybe done only in 9.x?

PR #1664, opened 2023-05-24
SOLR-16799 upgrade Solr to use Lucene 9.6.0

stillalex:
https://issues.apache.org/jira/browse/SOLR-16799

# Description

Lucene 9.6.0 upgrade. All tests seem to be passing.

* DocValuesTermsQuery is gone. this triggered some changes that will need a close look in TermsQParserPlugin and GraphEdgeCollector.

* TopLevelDocValuesTermsQuery  uses getTermData() which is deprecated, likely this part will need another rewrite

* Field storedValue() method was introduced as an optimization to detect value type:

    - this broke the EnumFieldType - org.apache.solr.TestDistributedSearch.test was failing. added details about the fix on the Jira ticket - https://issues.apache.org/jira/browse/SOLR-16799?focusedCommentId=17726446&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17726446
    - added handling for this to the LargeLazyField, needs a close look

* updated renovate config to skip lucene updates. not sure how to test this.

# Solution

# Tests

# Checklist


solr/core/src/java/org/apache/solr/search/SolrDocumentFetcher.java
```diff
@@ -539,6 +541,21 @@ public Reader readerValue() {
...
+      } else {
+        return new StoredValue(stringValue());
+      }
+    }
+
+    @Override
+    public InvertableType invertableType() {
+      // TODO SOLR-16799 when to use InvertableType.TOKEN_STREAM and when to use null?
+      return null;
+    }
```
tflobbe, line 558: It shouldn't matter, right? this is only for stored field retrieval, not for indexing. This should never be called AFAIK

solr/core/src/java/org/apache/solr/search/SolrDocumentFetcher.java
```diff
@@ -539,6 +541,21 @@ public Reader readerValue() {
...
       return null;
     }
+
+    @Override
+    public StoredValue storedValue() {
+      if (fieldType().stored() == false) {
+        return null;
+      } else {
+        return new StoredValue(stringValue());
+      }
```
tflobbe, line 551: I guess `fieldType().stored()` will never be `false` here?

tflobbe (commented):
LGTM. A couple minor comments added

solr/core/src/java/org/apache/solr/search/SolrDocumentFetcher.java
stillalex: I think you are correct. will remove the if block

solr/core/src/java/org/apache/solr/search/SolrDocumentFetcher.java
stillalex: hmm. I agree, will remove the TODO

stillalex:
thank you for talking a look @tflobbe! I updated the PR.

tflobbe (approved):
LGTM

tflobbe:
@stillalex, would you add a CHANGES entry for 9.3?

stillalex:
sorry for the delay @tflobbe! pushed a changes.txt update. I rebased too, running the tests again on my local just for good measure.

PR #1669, opened 2023-05-28
SOLR-16826: Validate if authentication plugin class name are valid

senthh:
https://issues.apache.org/jira/browse/SOLR-16826

# Description

Validate if authentication plugin class name are valid authentication class names; and if it is not valid then Solr has to write a WARN log

As per Solr 9 documents following are the valid authentication plugin names

 

{{

"solr.BasicAuthPlugin", "solr.JWTAuthPlugin", "solr.HadoopAuthPlugin", 
"solr.ConfigurableInternodeAuthHadoopPlugin", "solr.CertAuthPlugin", "solr.MultiAuthPlugin", 
"solr.KerberosPlugin"

}}

# Solution

Validating authentication plugin class names, if it is not matching with class names mentioned in Documentation then write WARN logs

# Tests

# Checklist

- [*] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [*] I have created a Jira issue and added the issue ID to my pull request title.
- [*] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [*] I have developed this patch against the `main` branch.
- [*] I have run `./gradlew check`.
- [*] I have added tests for my changes.
- [*] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -517,6 +517,11 @@ public class CoreContainer {
     authenticationConfig = Utils.getDeepCopy(authenticationConfig, 4);
     int newVersion = readVersion(authenticationConfig);
     String pluginClassName = null;
+    ArrayList<String> authPluginList = new ArrayList<String>(Arrays.asList(
+            "solr.BasicAuthPlugin", "solr.JWTAuthPlugin", "solr.HadoopAuthPlugin",
+            "solr.ConfigurableInternodeAuthHadoopPlugin", "solr.CertAuthPlugin",
+            "solr.MultiAuthPlugin", "solr.KerberosPlugin"
+    ));
     if (authenticationConfig != null) {
       if (authenticationConfig.containsKey("class")) {
         pluginClassName = String.valueOf(authenticationConfig.get("class"));
@@ -547,7 +552,12 @@ public class CoreContainer {
 
     // Initialize the plugin
     if (pluginClassName != null) {
-      log.info("Initializing authentication plugin: {}", pluginClassName);
+      if (authPluginList.contains(pluginClassName)) {
+        log.info("Initializing authentication plugin: {}", pluginClassName);
+      } else {
+        log.warn("Wrong Authentication plugin {} configured." +
+                "Please refer Solr 9+ documentation for correct Authentication plugin", pluginClassName);
+      }
       authenticationPlugin =
           new SecurityPluginHolder<>(
               newVersion,
--- a/solr/packaging/test/test_modules.bats
+++ b/solr/packaging/test/test_modules.bats
@@ -60,6 +60,27 @@ teardown() {
   refute_output --partial "java.lang.ClassNotFoundException"
 }
 
+@test "Hadoop-Auth Module: Validate KerberosPlugin Classloading" {
+  # Write a security.json that uses the KerberosPlugin
+  local security_json="${BATS_TEST_TMPDIR}/kerberos-security.json"
+  echo '{"authentication": {"class": "org.apache.solr.security.KerberosPlugin"}}' > "${security_json}"
+
+  # Start Solr
+  export SOLR_MODULES=hadoop-auth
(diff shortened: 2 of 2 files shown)


senthh:
@epugh Eric Pugh, could you please review this PR

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -547,7 +552,12 @@ private synchronized void initializeAuthenticationPlugin(
 
     // Initialize the plugin
     if (pluginClassName != null) {
-      log.info("Initializing authentication plugin: {}", pluginClassName);
+      if (authPluginList.contains(pluginClassName)) {
+        log.info("Initializing authentication plugin: {}", pluginClassName);
+      } else {
+        log.warn("Wrong Authentication plugin {} configured." +
```
epugh, line 558: I wonder if this needs to be more of a "The plugin X doesn't match any of the default plugins shipped with Solr"...   Wouldn't there be a possibility of someone writing their own plugin, so it wouldn't be in the authPluginList?

senthh: Yes Eric, I am aiming this PR just to validate solr's inbuilt Auth plugins write WARN logs if it is not matching as per Solr 9+ documentation. It will help users to, easily, find what mistake they commited.

janhoy:
I'm not a  fan of this solution. Yet a place of strings to maintain. 

What is wrong with the ClassNotFound exception you (probably) already get? Does it not tell you clearly the problem? Then you fix your class name and it works. If we should start adding such lists to all kind of plugins it would be a maintenance nightmare.

HoustonPutman:
The `org.apache.solr.security.*Plugin` still works, but the issue is that the package name changed for the KerberosPlugin from `org.apache.solr.security` (8.x) to `org.apache.solr.security.hadoop` (9.x). This should be included in the upgrade notes, I don't know why it wasn't.

I think that's the path forward, just document the change (and we can backport the upgrade notes change, so that it's included in 9.0, 9.1 and 9.2).

senthh:
> The `org.apache.solr.security.*Plugin` still works, but the issue is that the package name changed for the KerberosPlugin from `org.apache.solr.security` (8.x) to `org.apache.solr.security.hadoop` (9.x). This should be included in the upgrade notes, I don't know why it wasn't.
> 
> I think that's the path forward, just document the change (and we can backport the upgrade notes change, so that it's included in 9.0, 9.1 and 9.2).

Oh okay, then both org.apache.solr.security.hadoop.*Plugin and solr.*Plugin works? If it is so I can change in doc

HoustonPutman:
Oh okay, then both org.apache.solr.security.hadoop.*Plugin and solr.*Plugin works? If it is so I can change in doc

> Oh okay, then both org.apache.solr.security.hadoop.*Plugin and solr.*Plugin works? If it is so I can change in doc

It should work yes. We should just add a line in `solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-9.adoc` for 9.0, that the hdfs file locations have changed. There's a line about the code moving to a module, but it doesn't mention the classpaths.

senthh:
> Oh okay, then both org.apache.solr.security.hadoop.*Plugin and solr.*Plugin works? If it is so I can change in doc
> 
> > Oh okay, then both org.apache.solr.security.hadoop.*Plugin and solr.*Plugin works? If it is so I can change in doc
> 
> It should work yes. We should just add a line in `solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-9.adoc` for 9.0, that the hdfs file locations have changed. There's a line about the code moving to a module, but it doesn't mention the classpaths.

This is better way to proceed

janhoy:
Closing this, please open a new PR with Upgrade Notes doc changes...

PR #1670, opened 2023-05-29
SOLR-16829: Delegate arg parsing and help usage to java.  (Round 1)

epugh:
https://issues.apache.org/jira/browse/SOLR-16829

# Description

Going thorugh and attempting to remove the help usage (which seems to require also removing the arg parsing!)

I have migrated on the *nix side these commands:

```
bin/solr (by itself)
create
create_core
create_collection
delete
healthcheck
status
```

@Willdotwhite has done the same on the Windows side.

# Solution

# Tests

# Checklist


epugh:
So...   I'm starting to think this is a Solr 10 only change...   There are so many small cleanups and other awkwardnesses in the CLI, that making it work in Solr 9 is going to cause a LOT of folks some pain.   Not to mention, when we start moving all the commands like `-solrUrl` to `--solrUrl` pattern (in another PR)...    Plus, I'm thinking that this is going to need to bake for a while......

epugh:
More ambitious is this PR that I haven't tried to change the `solr.cmd` yet, but it should mimic what was done in `solr`.   cc @willdotwhite

Willdotwhite:
> More ambitious is this PR that I haven't tried to change the `solr.cmd` yet, but it should mimic what was done in `solr`. cc @Willdotwhite

Do you want me to take a crack at the solr.cmd changes (and would that be a PR from my fork into your source branch?) @epugh or do you plan to make the changes on this PR soon (in which case I'll focus on getting my local environment setup ready for testing)? :D

epugh:
That would be perfect.  This PR is ready for merging as soon as we have the Windows side dealt with.   There will be at least a Round 2, and maybe a Round 3 when we tackle `package` and `zk` commands ;-).

Willdotwhite:
Awesome 👍🏻 I'll have a read through the changes to `bin/solr` tomorrow and have a crack at it on Sunday if that's ok time-wise? (Busy couple of days coming up, apologies if that's inconvenient)

epugh:
That is perfect !

mkhludnev:
Hi, @epugh .
I've caught that you need to test something on windows. Could let me know what you need to check exactly?, I can do it tomorrow.

mkhludnev:
ok I've got https://github.com/epugh/solr/pull/2/files, I'll check it out. Thanks.

epugh:
@mkhludnev I've chased down the latest merge conflicts ;-).   Let me know how it goes, and if you want to do a pairing session, I can jump on next week.

epugh:
@Willdotwhite @mkhludnev so what is the best way to bring in the changes that both of you have made?  https://github.com/apache/solr/compare/main...mkhludnev:solr:SOLR-16829 shows some great improvments from each of you?    Do you each propose a PR to the branch that I have in my github.com/epugh/solr?   Just not sure how to get those commits into the PR?    @Willdotwhite are the changes you've made ready to go once `status` is done?

epugh:
@Willdotwhite I had a thought which is that maybe I'll back out the `bin/solr status`command changes, since that was giving you some problems, and move that to the Round 2 of work, which then would make this committable?

epugh:
@Willdotwhite can we delete the method `status_usage` ?   Line 282 has: 

```
IF "%SCRIPT_CMD%"=="status" goto status_usage
```

epugh:
okay @Willdotwhite I had to restore a change to `StatusTool.java` to make it work on Unix..   Is it possible that we have the .cmd and the .sh somewhat working at cross purposes in how we decide to output the help?

Willdotwhite:
> okay @Willdotwhite I had to restore a change to `StatusTool.java` to make it work on Unix.. Is it possible that we have the .cmd and the .sh somewhat working at cross purposes in how we decide to output the help?

Sorry @epugh , I think I was just a bit too tired yesterday and flubbed my testing - I've just put the condition back to `if ((cli.getOptions().length == 0 && cli.getArgs().length == 0) || cli.hasOption("h") || cli.hasOption("help")) {` on my machine and the status checks work as before; not really sure what happened there 🤔 

I think both are inline (although I'll fully rebuild from this branch during my coffee break in ~2 hours and confirm).

By the way, a bit of copypasta has slipped through the net on the hashbang of `solr.sh` [just here](https://github.com/apache/solr/pull/1670/files#diff-49c29c8f653341f48008c38f5f2cf970fa430cdca29181c819d7bdcfcc722980R1)

epugh:
I *think* we are finally there...   Responding to the last bits of feedback.

solr/bin/solr
```diff
@@ -937,8 +814,8 @@ function stop_solr() {
 
 if [ $# -eq 1 ]; then
   case $1 in
-    -help|-h)
-        print_usage ""
+    -help|-h)    
+        run_tool ""
```
gerlowskija, line 818: [Q] Does nothing in `tidy`/`check` complain about trailing whitespace?  I don't care much, but I'm surprised given how we've leaned into code-formatting in recent years.

solr/core/src/java/org/apache/solr/cli/HealthcheckTool.java
```diff
@@ -60,11 +60,11 @@ public List<Option> getOptions() {
         SolrCLI.OPTION_SOLRURL,
         SolrCLI.OPTION_ZKHOST,
         Option.builder("c")
-            .argName("COLLECTION")
+            .longOpt("name")
+            .argName("NAME")
             .hasArg()
             .required(false)
```
gerlowskija, line 66: [-1] AFAICT from [this bin/solr code](https://github.com/apache/solr/pull/1670/files#diff-49c29c8f653341f48008c38f5f2cf970fa430cdca29181c819d7bdcfcc722980L1003), the collection parameter should be required for the healthcheck tool.

Prior to this commit, we enforced this requirement in the bash/Windows-scripts, but now that we don't parse args or enforce required params there any more - we should update the `.required(...)` call here to be correct.

(Assuming I'm reading the old `bin/solr` code correctly at least; always possible I'm wrong.)

solr/core/src/java/org/apache/solr/cli/HealthcheckTool.java
```diff
@@ -60,11 +60,11 @@ public List<Option> getOptions() {
         SolrCLI.OPTION_SOLRURL,
         SolrCLI.OPTION_ZKHOST,
         Option.builder("c")
-            .argName("COLLECTION")
+            .longOpt("name")
```
gerlowskija, line 63: [Q] I've picked HealthcheckTool's handling of its "collection" parameter as a concrete example, but the concern below applies in a number of places.

----

Historically, `bin/solr healthcheck` has accepted the collection name using either `-c` or `-collection`.  But with this PR, AFAICT it now accepts `-c` and `--collection` (note the double-dash in the longopt).

Personally, I like the new syntax better - double-dashes on longopts are pretty ubiquitous in other CLI tools, and Solr's usage of `-longoptName` always struck me as quirky.  That said though, it's a backwards incompatible change so we need to roll it out carefully.

If you **do** plan on committing some form of this to branch_9x, do you have plans for addressing backcompat?  (i.e. Would the branch_9x version accept both `--collection` and `-collection` as longopt forms?) 

If you **don't** plan on committing some form of this to branch_9x, things are better from a backcompat perspective but we'd still need to call out the syntax changes in CHANGES.txt or our Upgrade Notes, presumably?Do you have a general sense or (even better) a list of the syntax changes that'd need documented?

gerlowskija (commented):
Awesome to see so many lines in `bin/solr` and `bin/solr.cmd` get nuked, though I have some concerns around backcompat and the user experience.  See inline comments for more details...

solr/bin/solr
epugh: Not that I know of...   Maybe because this is shell?  I didn't run shellcheck after making changes...

solr/core/src/java/org/apache/solr/cli/HealthcheckTool.java
epugh: https://issues.apache.org/jira/browse/SOLR-16882 for fixing this...   I will investigate the .bats test...

solr/core/src/java/org/apache/solr/cli/HealthcheckTool.java
epugh: Thanks for jumping on this...   https://issues.apache.org/jira/browse/SOLR-16824 was opened to specifically track that conversion to the `--collection` approach.   I agree on the syntax changes needing tracking in Upgrade Notes.   I might start with the Upgrade Notes for 10.x, but if we back port, then bring them back to that version of Upgrade Notes.

epugh: also...   one more thing is that the long name is actually `--name`..   I was struggling with the CreateTool which used `--name` since it can do core or collection.   Can't use `--collection` for a Core!

epugh: LIkewise the same thinking with the DeleteTool

PR #1682, opened 2023-06-02
SOLR-16395 JAX-RS conversion for zkversion and field related endpoints.

bszabo97:
https://issues.apache.org/jira/browse/SOLR-16395

# Description

This PR contains the JAX-RS conversion of endpoints:
* /zkversion
* /fields
* /fields/{fieldname}
* /copyfields
* /dynamicfields
* /dynamicfields/{fieldname}
* /fieldtypes
* /fieldtypes/{fieldtypename}

I also moved SchemaNameAPI to GetSchemaAPI so that everything is in one place.

There should be no difference in usage so there is no need for documentation modifications, I believe.

# Solution

# Tests

I added unit tests similar to what can be seen in PR #1649 

# Checklist


gerlowskija:
> I also moved SchemaNameAPI to GetSchemaAPI so that everything is in one place

🎉 Perfect, thanks!  That'd been bothering me as well and was on my todo list.

Will review this over the next few days - thanks again for the PR @bszabo97 !

bszabo97:
Thank you @gerlowskija !

Also I had an issue I forgot to mention when I uploaded the PR. I tried to create unit test for endpoint `/zkversion` I was not able to mock `ManagedIndexSchema` class since it is a final class. I have tried several solutions found online but I was not happy with any of those. 
Do you have any suggestions how I could test the `ManagedIndexSchema` branch in the code?

solr/core/src/java/org/apache/solr/handler/admin/api/GetSchemaFieldAPI.java
```diff
@@ -0,0 +1,255 @@
...
+        SolrException.ErrorCode.NOT_FOUND, "No such path /" + realName + "/" + fieldTypeName);
+  }
+
+  public static class SchemaGetFieldTypeInfoResponse extends SolrJerseyResponse {
+    @JsonProperty("fieldType")
+    public SimpleOrderedMap<?> fieldTypeInfo;
+  }
+
+  private Object listAllFieldsOfType(String realName, SolrParams params) {
+    String camelCaseRealName = IndexSchema.nameMapping.get(realName);
```
gerlowskija, line 187: [0] Mostly thinking aloud here, but I wonder what parameters the schema code looks for in this `SolrParams` object.  Obviously there's 'meta' below, but there's probably others in `getNamedPropertyValues`.

The more explicit we can be about the accepted parameters, the easier this code will be to read.  And the nicer any sort of client code will be if we start doing automated client generation (see SOLR-16835).  If the number of parameters is relatively small it might be worth calling each param out explicitly in the API methods above, and then   packaging them up in a SolrParam to satisfy the IndexSchema call below.

solr/core/src/java/org/apache/solr/handler/admin/api/GetSchemaZkVersionAPI.java
```diff
@@ -0,0 +1,69 @@
+package org.apache.solr.handler.admin.api;
```
gerlowskija, line 1: [-1] Ditto, re: missing copyright header.

solr/core/src/java/org/apache/solr/jersey/InjectionFactories.java
```diff
@@ -104,6 +106,24 @@ public IndexSchema provide() {
     public void dispose(IndexSchema instance) {}
   }
 
+  public static class ReuseFromContextSolrParamsFactory implements Factory<SolrParams> {
```
gerlowskija, line 109: [+1] Technically, v2 code that needs the SolrParams could inject the SolrQueryRequest and call `getParams`.  So this injector isn't strictly required.

But I'm a big fan of only injecting what you actually need - it makes the code easier to understand, and really helps simplify unit tests.  So 👍

solr/core/src/test/org/apache/solr/handler/admin/api/GetSchemaAPITest.java
```diff
@@ -55,6 +59,38 @@ public void testReliesOnIndexSchemaWhenFetchingWholeSchema() {
...
+   * Test the v2 to v1 response mapping for /schema/name
+   *
+   * <p>{@link SchemaHandler} uses the v2 {@link GetSchemaAPI} (and its response class {@link
+   * GetSchemaAPI.SchemaNameResponse}) internally to serve the v1 version of this functionality. So
+   * it's important to make sure that our response stays compatible with SolrJ - both because that's
+   * important in its own right and because that ensures we haven't accidentally changed the v1
+   * response format.
+   */
+  @Test
+  public void testResponseCanBeParsedBySolrJ() {
```
gerlowskija, line 82: [+1] Great idea to test the SolrJ compatibility - I should've been doing this in a lot of other places but hadn't thought of it.

solr/core/src/test/org/apache/solr/handler/admin/api/GetSchemaFieldsAPITest.java
```diff
@@ -0,0 +1,133 @@
+package org.apache.solr.handler.admin.api;
```
gerlowskija, line 1: [-1] Ditto, re: missing copyright header

gerlowskija (changes requested):
Overall this looks great.  There are a few small things that will need changed I think (missing copyright headers, etc).  But otherwise we should be able to get this in soon; thanks for the PR @bszabo97 !

gerlowskija:
> I tried to create unit test for endpoint /zkversion I was not able to mock ManagedIndexSchema class since it is a final class. I have tried several solutions found online but I was not happy with any of those

Yeah, I've run into these sort of restrictions at different points as well.  Unfortunately there's no easy workaround that I know of.  I think more recent versions of Mockito have a way to mock 'final' classes.  I tried to upgrade the Mockito that Solr uses at some point but ran into some dep conflicts that were tricky to resolve.

If the test would add enough value to justify the effort, we could look at removing the 'final' modifier from ManagedIndexSchema.  I could be wrong, but I don't think anyone would object to that as long as there's no especially compelling reason for MIS to be "final".  I can take that on and get back to you shortly

solr/core/src/test/org/apache/solr/handler/admin/api/GetSchemaAPITest.java
bszabo97: It was your idea, I just copied this test from the already existing one :D I can add similar tests to more places if you think it would be useful.

solr/core/src/java/org/apache/solr/handler/admin/api/GetSchemaFieldAPI.java
```diff
@@ -0,0 +1,255 @@
...
+    final String realName = "fields";
+
+    response.fields = listAllFieldsOfType(realName, params);
+
+    return response;
+  }
+
+  public static class SchemaListFieldsResponse extends SolrJerseyResponse {
+    @JsonProperty("fields")
+    public Object fields;
```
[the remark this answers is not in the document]
bszabo97: To be honest I used the type `Object` since I have seen that the call `propertyValues.get()` returns an Object and did not really knew if there could be any possible scenarios where this is not a `List`... But the tests also suggest that this should always be an `ArrayList` so I will go ahead and change it everywhere

solr/core/src/java/org/apache/solr/handler/admin/api/GetSchemaFieldAPI.java
bszabo97: I was doing an experiment before this approach to try and pass all the necessary parameters, but I have found [here](https://github.com/apache/solr/blob/8089e6fc6a9d37299df9a63e5372ec5c235e4f9a/solr/core/src/java/org/apache/solr/schema/IndexSchema.java#L1643-L1647) that we would need at least 6 parameters. I considered it a bit too much to nicely handle so I went ahead and used the parameter injection instead. 
If you think 6 parameters is a viable option I can easily change it back to that approach. Let me know what you think!

solr/core/src/java/org/apache/solr/handler/admin/api/GetSchemaZkVersionAPI.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.solr.jersey.PermissionName;
+import org.apache.solr.jersey.SolrJerseyResponse;
+import org.apache.solr.schema.ManagedIndexSchema;
+import org.apache.solr.schema.ZkIndexSchemaReader;
+import org.apache.solr.security.PermissionNameProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Path("/{a:cores|collections}/{collectionName}/schema")
+public class GetSchemaZkVersionAPI extends JerseyResource {
```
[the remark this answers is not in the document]
bszabo97: I was doing the extension more because `GetSchemaFieldAPI` also uses the index schema and not because we inherit the path as well. I was thinking that if we need to add the Solr core here anyways it is nicer to only have that one attribute than both Solr core and Index schema.

solr/core/src/java/org/apache/solr/handler/admin/api/GetSchemaFieldAPI.java
gerlowskija: 6 is a lot, 🤔 

Ultimately, I think there will be value in being as explicit as possible here.  But most of those benefits rely on JIRA tickets that haven't been finished yet (e.g. using these API definitions to [auto-generate client code](https://github.com/apache/solr/pull/1681)).

So, I'm happy to go either way here.  It'll set us up really nicely if we're more explicit with the API params.  But if you don't want the method signature to be noisy in the meantime, that also makes sense to me.

solr/core/src/test/org/apache/solr/handler/admin/api/GetSchemaAPITest.java
gerlowskija: Yikes - it worries me that I didn't remember haha.

> I can add similar tests to more places if you think it would be useful.

Absolutely, if you're willing. I think this would be a great type of test to have for other APIs as well!

gerlowskija (approved):
LGTM - still need to test, but then I'll aim to commit!

solr/core/src/java/org/apache/solr/handler/admin/api/GetSchemaFieldAPI.java
gerlowskija: I ended up adding a TODO comment to the ctor where we inject the SolrParams as a reminder to circle back to this and be more explicit.

But I'm happy to merge as-is for now

solr/core/src/java/org/apache/solr/handler/admin/api/GetSchemaFieldAPI.java
```diff
@@ -0,0 +1,255 @@
...
+import org.apache.solr.common.util.NamedList;
+import org.apache.solr.common.util.SimpleOrderedMap;
+import org.apache.solr.core.PluginInfo;
+import org.apache.solr.jersey.PermissionName;
+import org.apache.solr.jersey.SolrJerseyResponse;
+import org.apache.solr.pkg.PackageListeningClassLoader;
+import org.apache.solr.schema.IndexSchema;
+import org.apache.solr.security.PermissionNameProvider;
+
+public class GetSchemaFieldAPI extends GetSchemaAPI {
```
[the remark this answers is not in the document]
stillalex: minor suggestion. could we have the 'strict' validation based off a system property? so we can have a smoke test (just a simple Solr startup) running nightly or something.

PR #1599, opened 2023-04-30
Update dependency com.fasterxml.jackson:jackson-bom to v2.15.2

solrbot:
This PR contains the following updates:

| Package | Type | Update | Change |
|---|---|---|---|
| [com.fasterxml.jackson:jackson-bom](https://togithub.com/FasterXML/jackson-bom) | dependencies | minor | `2.14.2` -> `2.15.2` |

---

### Configuration

📅 **Schedule**: Branch creation - "* * * * 0" (UTC), Automerge - At any time (no schedule defined).

🚦 **Automerge**: Disabled by config. Please merge this manually once you are satisfied.

♻ **Rebasing**: Whenever PR becomes conflicted, or you tick the rebase/retry checkbox.

🔕 **Ignore**: Close this PR and you won't be reminded about this update again.

---

 - [ ] If you want to rebase/retry this PR, check this box

---

This PR has been generated by [Renovate Bot](https://togithub.com/solrbot/renovate-github-action)

PR #1694, opened 2023-06-07
SOLR-16812: update to cbor 2.15.2

noblepaul:


chatman:
I think there's a solrbot version of this same PR. We should close one of them.
https://github.com/apache/solr/pull/1599

PR #1699, opened 2023-06-09
Remove deprecated BackupRepository delete with ignoreNoSuchFileException API

risdenk:


HoustonPutman (approved):
This looks good. Probably doesn't warrant a Ref Guide upgrade note for 10.0, since users will know when upgrading the libraries...

PR #1704, opened 2023-06-12
SOLR-16470 : Create V2 equivalent of V1 Replication: GET files

mlbiscoc:

# Description

Create a V2 equivalent API for the replication "GET filelist"

V1
`/solr/{coreName}/replication?command=filelist&generation={genNumber}`

V2
`/api/cores/{coreName}/replication/files?generation={genNumber}`

# Solution

Created a new endpoint in CoreReplicationAPI class using JAS-RX framework. I then moved the logic of the filelist from the ReplicationHandler class into the CoreReplicationAPI class and updated the V1 endpoint to use the CoreReplicationAPI class so that the ReplicationHandler can be deprecated in the future.

# Tests

Wrote new Unit test in CoreReplicationAPITest that will mock a response from getFileList into a NamedList object. 

# Checklist


mlbiscoc:
@gerlowskija Could you take a look at this PR when you get a chance?

solr/core/src/java/org/apache/solr/handler/admin/api/CoreReplicationAPI.java
```diff
@@ -46,12 +50,24 @@ public CoreReplicationAPI(SolrCore solrCore, SolrQueryRequest req, SolrQueryResp
...
   @Path("/indexversion")
-  @Produces({"application/json", "application/xml", BINARY_CONTENT_TYPE_V2})
+  @Produces({MediaType.APPLICATION_JSON, MediaType.APPLICATION_XML, BINARY_CONTENT_TYPE_V2})
   @PermissionName(CORE_READ_PERM)
   public IndexVersionResponse fetchIndexVersion() throws IOException {
     return doFetchIndexVersion();
   }
 
+  @GET
+  @Path("/files/generation/{gen}")
```
gerlowskija, line 60: [Q] For a lot of these v2 APIs, I've been working from a [spreadsheet](https://docs.google.com/spreadsheets/d/1HAoBBFPpSiT8mJmgNZKkZAPwfCfPvlc08m5jz3fQBpA) that describes what the new endpoints should be.

In the spreadsheet, the suggested v2 API was just `GET /cores/coreName/replication/files`.  (I think at the time I was assuming "generation" would be a required query parameter, rather than a path-param.  Though it's possible I didn't even know about it at that point.)

Not that we have to stick to that necessarily - if you think it makes sense to keep "generation" in the path, it's def an option.

Any thoughts?

solr/core/src/java/org/apache/solr/handler/admin/api/CoreReplicationAPI.java
```diff
@@ -46,12 +50,24 @@ public CoreReplicationAPI(SolrCore solrCore, SolrQueryRequest req, SolrQueryResp
...
+  @Produces({MediaType.APPLICATION_JSON, MediaType.APPLICATION_XML, BINARY_CONTENT_TYPE_V2})
+  @PermissionName(CORE_READ_PERM)
+  public NamedList<Object> fetchFiles(
```
gerlowskija, line 63: [0] The method name "fetchFiles" sounds a little bit like you're getting the file-content itself, rather than a list of file names/metadata.  Might be worth updating.

[-0] One of the values of the JAX-RS framework is that we can be explicit about the inputs and outputs of our APIs.  If possible, we should avoid returning Namedlist here and instead return some other strongly typed class that makes it clearer what the response should look like?

solr/core/src/java/org/apache/solr/handler/admin/api/ReplicationAPIBase.java
```diff
@@ -38,10 +68,140 @@ public ReplicationAPIBase(
...
+  }
 
-    return replicationHandler.getIndexVersionResponse();
+  protected NamedList<Object> getFileList(long generation, ReplicationHandler replicationHandler) {
+    final IndexDeletionPolicyWrapper delPol = solrCore.getDeletionPolicy();
+    final NamedList<Object> filesResponse = new NamedList<>();
+
+    IndexCommit commit = null;
+    try {
+      if (generation == -1) {
```
gerlowskija, line 88: [Q] Mostly thinking out loud here, but I wonder whether we should just fall back on "-1"
 as a default.  (Right now we require users to provide an explicit generation value).
 
 Probably not something to handle in this PR regardless, just wanted to jot down the thought somewhere...

solr/solr-ref-guide/modules/deployment-guide/pages/user-managed-index-replication.adoc
```diff
@@ -433,9 +433,27 @@ http://_follower_host:port_/solr/_core_name_/replication?command=details
...
+====
+[.tab-label]*V1 API*
+
 [source,bash]
+----
 http://_host:port_/solr/_core_name_/replication?command=filelist&generation=<_generation-number_>
+
+----
+====
++
```
gerlowskija, line 447: [+1] Thanks for thinking of the docs!

gerlowskija (commented):
Looks good overall - particularly the tests and ref-guide updates.

There were one or two small things that we might want to consider updating, but otherwise this looks pretty close!

solr/core/src/java/org/apache/solr/handler/admin/api/ReplicationAPIBase.java
mlbiscoc: This makes sense to me and like that idea. I could try changing it myself in this PR but if you think this requires more discussion or would like to do it yourself in another PR I'll leave as is.

solr/core/src/java/org/apache/solr/handler/admin/api/CoreReplicationAPI.java
mlbiscoc: Changed the method name and returned a POJO object of a strongly typed class. Also updated the tests to reflect that. Let me know what you think.

solr/core/src/java/org/apache/solr/handler/admin/api/CoreReplicationAPI.java
gerlowskija: This seems reasonable to me, though it might be undermined if we decide to set a default "generation" value as we discussed elsewhere on this PR.  Let's consolidate discussion of this [here](https://github.com/apache/solr/pull/1704#discussion_r1238667428)

solr/core/src/java/org/apache/solr/handler/ReplicationHandler.java
```diff
@@ -764,12 +765,8 @@ public FileInfo(long lasmodified, String name, long size, long checksum) {
       this.checksum = checksum;
     }
 
-    Map<String, Object> getAsMap() {
-      Map<String, Object> map = new HashMap<>();
-      map.put(NAME, name);
-      map.put(SIZE, size);
-      map.put(CHECKSUM, checksum);
-      return map;
+    CoreReplicationAPI.FileMetaData getAsMap() {
```
gerlowskija, line 768: [-1] We probably need to rename this method, since it's no longer returning a map 😛

solr/core/src/java/org/apache/solr/handler/admin/api/CoreReplicationAPI.java
```diff
@@ -88,4 +90,137 @@ public IndexVersionResponse(Long indexVersion, Long generation, String status) {
...
+    }
+
+    public void addToConfFiles(List<FileMetaData> confFilesMetaData) {
+      if (confFiles == null) {
+        confFiles = new ArrayList<>();
+      }
+      confFiles.addAll(confFilesMetaData);
+    }
+
+    public void setFileList(List<FileMetaData> fileList) {
```
gerlowskija, line 127: [Q] Do setters/getters here provide any value?  We kindof have to make the fields in this class 'public' bc of how we use Jackson, and it doesn't look like the getters/setters are doing anything that'd be different from users just accessing the fields directly.

solr/core/src/java/org/apache/solr/handler/admin/api/ReplicationAPIBase.java
```diff
@@ -73,22 +62,24 @@ protected CoreReplicationAPI.IndexVersionResponse doFetchIndexVersion() throws I
...
     ReplicationHandler replicationHandler =
         (ReplicationHandler) solrCore.getRequestHandler(ReplicationHandler.PATH);
     return getFileList(generation, replicationHandler);
   }
 
-  protected NamedList<Object> getFileList(long generation, ReplicationHandler replicationHandler) {
+  protected CoreReplicationAPI.FileListResponse getFileList(
+      long generation, ReplicationHandler replicationHandler) {
     final IndexDeletionPolicyWrapper delPol = solrCore.getDeletionPolicy();
-    final NamedList<Object> filesResponse = new NamedList<>();
```
gerlowskija: [+1] Woohoo - always nice to see NamedList usages go away!

gerlowskija (commented):
LGTM - I'm mostly just holding off on "official" Approval bc we have a few outstanding discussions around the "generation" param (path vs query param, default value, etc.)

solr/core/src/java/org/apache/solr/handler/admin/api/ReplicationAPIBase.java
mlbiscoc: Hmmm I agree it makes less sense to put generation in the path if we decide to support a default generation. There doesn't seem to be a clean way to generation in the path while also defaulting so if I had to choose, I think there is more value in generation being in the query parameter and being able to default generation than the other way around. 

I'll modify the API with generation in the query parameter.

solr/core/src/java/org/apache/solr/handler/admin/api/CoreReplicationAPI.java
mlbiscoc: I took another look and removed them. After some though, I don't think they do provide any value. These are supposed to be POJO objects and should be nothing but a response object without being overly complex.

solr/core/src/java/org/apache/solr/handler/admin/api/ReplicationAPIBase.java
mlbiscoc: API updated with generation now as a query parameter

gerlowskija:
Merged - will target backporting next week if test runs over the weekend don't surface any issues.  Thanks again @mlbiscoc!

HoustonPutman:
@gerlowskija, this broke precommit (for jenkins and github actions) https://github.com/apache/solr/actions/runs/5425690231/jobs/9866774872

mlbiscoc:
> @gerlowskija, this broke precommit (for jenkins and github actions) https://github.com/apache/solr/actions/runs/5425690231/jobs/9866774872

Sorry about that. ```./gradlew check``` passed running locally on my laptop. Did I miss a step and if so, what can I do to avoid this in the future?

HoustonPutman:
No worries, its annoying because that gradle validation only runs if it can tell its running in a Github Action or in Jenkins. So you can't tell locally.

If you look at the checks for the last commits in this PR, you can see that the precommit check failed. That's the only real way to catch it before committing.

mkhludnev:
> > @gerlowskija, this broke precommit (for jenkins and github actions) https://github.com/apache/solr/actions/runs/5425690231/jobs/9866774872
> 
> Sorry about that. `./gradlew check` passed running locally on my laptop. Did I miss a step and if so, what can I do to avoid this in the future?

if fails for me locally with `-Pvalidation.errorprone=true` 
fyi https://github.com/apache/solr/pull/1751

PR #1706, opened 2023-06-13
SOLR-16846: Use IndexOrDocValuesQuery for PointFields on Set queries

tflobbe:
Similar to Lucene's https://github.com/apache/lucene/pull/12087

### DIFF
--- a/solr/benchmark/src/java/org/apache/solr/bench/search/NumericSearch.java
+++ b/solr/benchmark/src/java/org/apache/solr/bench/search/NumericSearch.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
@@ -94,7 +95,7 @@ public class NumericSearch {
               .field("term_high_s", highCardinalityTerms);
       // .field("numbers_dt", setValues);
 
-      miniClusterState.index(COLLECTION, docs, numDocs);
+      miniClusterState.index(COLLECTION, docs, numDocs, false);
       basePath = miniClusterState.nodes.get(0);
       SolrQuery q = new SolrQuery("*:*");
       q.setParam("facet", "true");
@@ -183,7 +184,10 @@ public class NumericSearch {
       private final AtomicInteger idx;
 
       CircularIterator(Collection<T> collection) {
-        this.collection = collection.toArray();
+        this.collection = Objects.requireNonNull(collection).toArray();
+        if (this.collection.length == 0) {
+          throw new IllegalArgumentException("This iterator doesn't support empty collections");
+        }
         this.idx = new AtomicInteger();
       }
 
--- a/solr/benchmark/src/java/org/apache/solr/bench/search/NumericSearch.java
+++ b/solr/benchmark/src/java/org/apache/solr/bench/search/NumericSearch.java
@@ -1,0 +1,233 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.bench.search;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.apache.solr.bench.BaseBenchState;
+import org.apache.solr.bench.Docs;
+import org.apache.solr.bench.MiniClusterState;
+import org.apache.solr.bench.generators.IntegersDSL;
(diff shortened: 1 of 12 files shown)


solr/core/src/java/org/apache/solr/schema/IntPointField.java
```diff
@@ -119,7 +121,16 @@ public Query getSetQuery(QParser parser, SchemaField field, Collection<String> e
       values[i] = parseIntFromUser(field.getName(), val);
       i++;
     }
-    return IntPoint.newSetQuery(field.getName(), values);
+    Query pointsQuery = IntPoint.newSetQuery(field.getName(), values);
```
stillalex, line 124: with the caveat that most of this is above my head, I think this code could be written in a simplified manner as
```
if (field.hasDocValues()) {
  return IntField.newSetQuery(field.getName(), values);
} else {
  return IntPoint.newSetQuery(field.getName(), values);
}
```
most of the new code seems to be the same as the IntField.newSetQuery method, so I am thinking calling into it could be a good idea. haven't checked the other types yet.

stillalex:
discovered a hidden feature of the benchmark, turns out you can run the following and it will only perform the `intSet` and `intDvSet` benchmarks! 
```
./jmh.sh NumericSearch.int
```

just in case it helps, providing my numbers for reference:
* with patch applied but disabled (just commented out that code addition)

```
Benchmark                Mode  Cnt     Score    Error  Units
NumericSearch.intDvSet  thrpt    5  1981.450 ± 22.090  ops/s
NumericSearch.intSet    thrpt    5  1865.810 ± 24.893  ops/s
```

* with patch

```
Benchmark                Mode  Cnt     Score    Error  Units
NumericSearch.intDvSet  thrpt    5  1907.598 ± 35.759  ops/s
NumericSearch.intSet    thrpt    5  1880.214 ± 68.973  ops/s
```

so if I am reading this correctly, I am seeing an improvement from `22.090  ops/s` to `35.759  ops/s`.

tflobbe:
Thanks @stillalex. I've applied your suggestion. I've updated the benchmark to show the case where we should be getting the benefits of this change. I'm going to run it against `main` to check the results there, but this is what I get running this test with the patch:
```
Benchmark                   Mode     Cnt       Score       Error    Units
NumericSearch.intDvSet      thrpt    5         2129.873  ± 875.960  ops/s
NumericSearch.intSet        thrpt    5         1355.741  ± 366.968  ops/s
```

tflobbe:
Here is the result of running the latest test in `main` vs `branch`:

Main:
```
Benchmark                   Mode  Cnt     Score     Error  Units
NumericSearch.doubleDvSet  thrpt    5  1399.982 ± 113.036  ops/s
NumericSearch.doubleSet    thrpt    5  1416.364 ±  50.363  ops/s
NumericSearch.floatDvSet   thrpt    5  1391.298 ±  63.938  ops/s
NumericSearch.floatSet     thrpt    5  1431.984 ±  52.668  ops/s
NumericSearch.intDvSet     thrpt    5  1615.806 ±  28.131  ops/s
NumericSearch.intSet       thrpt    5  1564.974 ±  36.716  ops/s
NumericSearch.longDvSet    thrpt    5  1507.846 ±  42.412  ops/s
NumericSearch.longSet      thrpt    5  1559.993 ±  39.056  ops/s
```
Branch:
```
Benchmark                   Mode  Cnt     Score     Error  Units
NumericSearch.doubleDvSet  thrpt    5  2227.143 ± 182.832  ops/s
NumericSearch.doubleSet    thrpt    5  1611.681 ±  89.757  ops/s
NumericSearch.floatDvSet   thrpt    5  2408.225 ± 113.256  ops/s
NumericSearch.floatSet     thrpt    5  1505.764 ±  85.578  ops/s
NumericSearch.intDvSet     thrpt    5  2455.584 ± 142.097  ops/s
NumericSearch.intSet       thrpt    5  1472.866 ±  83.228  ops/s
NumericSearch.longDvSet    thrpt    5  2315.010 ± 510.850  ops/s
NumericSearch.longSet      thrpt    5  1492.064 ±  41.869  ops/s
```

solr/benchmark/src/java/org/apache/solr/bench/search/NumericSearch.java
```diff
@@ -0,0 +1,298 @@
...
+              new SolrQuery(
+                  "q",
+                  termQueryField + ":" + lowCardTerms.next(),
+                  "fq",
+                  "{!terms cache=false f='" + field + "'}" + queries.next()));
+      q.setBasePath(basePath);
+      return q;
+    }
+
+    private static class CircularIterator<T> implements Iterator<T> {
```
stillalex, line 180: minor suggestion. flavors of this exist already in commons-collections. you could consider replacing this with something like `org.apache.commons.collections4.iterators.LoopingIterator` or `org.apache.commons.collections4.iterators.LoopingListIterator`.

tflobbe: While the benchmark is currently using just one thread, we could increase it to test different things, and those two aren't thread safe. I'd rather keep the one we have here for now

solr/benchmark/src/resources/configs/cloud-minimal/conf/schema.xml
```diff
@@ -44,11 +44,16 @@
     <dynamicField name="*_b" type="boolean" indexed="true" stored="true"/>
     <dynamicField name="*_s" type="string" indexed="true" stored="false"/>
     <dynamicField name="*_t" type="text" indexed="true" stored="false"/>
-    <dynamicField name="*_i" type="int" indexed="false" stored="false"/>
```
HoustonPutman, line 47: Have you run the full test suite to make sure this is compatible with tests that use `cloud-minimal`?

HoustonPutman (approved):
Awesome to see usage of new Lucene improvements!

solr/benchmark/src/resources/configs/cloud-minimal/conf/schema.xml
tflobbe: Good call, I thought about it when I changed the schema but I forgot after. 
I've updated `JsonFaceting` test to use the dv + indexed field (used to be dv only). Should be no difference. `CloudIndexing` was using the non-indexed fields which is strange to me, maybe the schema was updated after the test or maybe the test was focusing just on the distribution? I left it using the indexed fields now, which is more realistic, if someone is trying to test something in particular with non-indexed fields they can update again.

PR #1724, opened 2023-06-26
Refactor: move PerReplicaStatesFetcher code into PerReplicaStatesOps 

dsmiley:
This is a trivial & mechanical refactoring.  The PerReplicaStatesFetcher class was something I created with @heythm when we created solrj-zookeeper; we didn't know where to put this functionality at the time.  I think PerReplicaStatesOps is a good home.


noblepaul (approved):
LGTM

PR #1726, opened 2023-06-27
SOLR-16840: Workaround HK2 DI bug

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-16840

# Description

Prior to this commit many v2 invocations would succeed but log scary looking stacktraces to Solr's console log.  This was because of a bug in how Jersey creates and cleans up "RequestScoped" objects.  (See [Jersey#3503](https://github.com/eclipse-ee4j/jersey/issues/3503) for more details.)

# Solution

This commit works around the bug by changing how several "factory" classes obtain a "ContainerRequestContext": switching away from DI and towards looking up the CRC more explicitly using a "ServiceLocator".

# Tests

Manual testing to verify that the warning stacktraces no longer appear in Solr's console logging.  Automated tests continue to pass.

# Checklist


epugh (approved):
LGTM.    Finally learned what the hk2 prefix is.  https://en.wikipedia.org/wiki/GlassFish_HK2

Just out of curiosity, are we behind on our version of Glassfish and HK2?  I.e, are we establishing a pattern of not upgrading and then having to work around things?   If/when the bug in HK2 gets fixed, how will we know to eliminate our workaround?  DOn't we sometimes annoate code with a bug url?

HoustonPutman:
So the Github Issue that Jason linked has been open for 6 years, and not closed. So definitely not an upgrade issue.

Yeah we can probably link the Issue url in a comment somewhere, so this is documented in the code. Good idea

gerlowskija:
> are we behind on our version of Glassfish and HK2?

No, we're not behind on hk2 - it's just a long standing bug I guess.  I've added a comment as you guy suggested to we can use more conventional injection when the bug does eventually get fixed.

PR #1729, opened 2023-06-27
BackupRepository permit

dsmiley:
A draft PR for a POC to show how we might limit the number of concurrent backups or restores happening per node.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/backup/repository/BackupRepository.java
+++ b/solr/core/src/java/org/apache/solr/core/backup/repository/BackupRepository.java
@@ -48,6 +48,17 @@ public interface BackupRepository extends NamedListInitializedPlugin, Closeable 
   }
 
   /**
+   * Acquires a permit, used with try-finally. Should be used to wrap an aggregate heavy operation
+   * (backup or restore an index; many files).
+   *
+   * @param operation for logging; possibly for decision-making too
+   * @throws InterruptedException if it takes a long time
+   */
+  default AutoCloseable permit(String operation) throws InterruptedException {
+    return new RepoPermit(operation);
+  }
+
+  /**
    * This method returns the location where the backup should be stored (or restored from).
    *
    * @param override The location parameter supplied by the user.
--- a/solr/core/src/java/org/apache/solr/core/backup/repository/RepoPermit.java
+++ b/solr/core/src/java/org/apache/solr/core/backup/repository/RepoPermit.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.solr.core.backup.repository;
+
+import java.lang.invoke.MethodHandles;
+import java.util.concurrent.Semaphore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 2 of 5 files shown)


dsmiley:
@psalagnac LMK what you think.

I briefly toyed with a FilterDirectory wrapper but then realized that the mechanism needn't implement anything; we just want a Semaphore.  Also I was unhappy with the DirectoryFactory relationship with Directory... I couldn't simply delegate close because the API is rather custom.  Could be improved.  If we wanted to rate limit, this could be explored more.

I briefly toyed with BackupRepository constructor & close having this mechanism.  But I don't see that we actually close BackupRespository ?!  Also there are simple use-cases like listing backups and deleting and other things that are not heavy and don't want to hold a permit.

A colleague named Gilles expressed concern that a "restore" should not wait in line like a backup; it's relatively rare and done in emergencies.  I think this is a good point, but I'd still want a similar mechanism described here to limit it somehow.  Maybe we have separate permit calls (backup vs restore).  And maybe a simple default implementation in which we track that restores are in progress.  If a backup gets a permit but sees restores in progress, it gives it up immediately, waits, and then tries again.

murblanc:
I think there are multiple options for throttling:
- Is the throttling decision made at the source (CollectionsHandler or Overseer) or on the node actually executing the action?
- Are the limits per command execution or overall for a node? (the second option seems better for protecting resources)

This proposal is deciding to throttle at the node and takes into account all (backup/restore) commands that need to run on that node.
IMO this is too late in the command execution lifecycle. It can lead to starting execution of a very large number of commands and blocking them all waiting for each other.
It might be more efficient to block earlier so that commands can complete faster and not compete for resources with newer commands starting execution in an already saturated environment.
It can also allow prioritizing one type of command while blocking others (for example if there are restores in the queue they are unblocked first and backups wait for the restores to complete - assuming for example that restores are more urgent than backups).

Throttling earlier could also be surfaced back to the client by rejecting new API calls if the system can tell it won't be able to handle them.

dsmiley:
Thanks for weighing in Ilan.

My proposal here was motivated by a goal of providing something very simple which I hope you can see is a desirable quality.  @psalagnac's POC in BackupCmd (i.e. "early" instead of "late") is very complex in tracking and doesn't protect against multiple concurrent backups being issued nor restore I/O.

>  It can lead to starting execution of a very large number of commands and blocking them all waiting for each other.

There is no free lunch if we limit resources -- commands will compete.  But I do get your point.  For a _very naive_ backup client that might choose to bombard Solr with backup requests, they would all be accepted and be using threads within Solr waiting for a very long time.  I suppose a way to wrestle with this decision is to consider what a backup invoking daemon  should do to do its job _well_ (not naively).  I imagine a cron/timer that loops over all collections to back them up.  One-at-a-time would take poor advantage of resources, but we could imagine a client that uses a Semaphore with permits equal to the number of nodes * a rough number simultaneous backups desired.  Backup of a collection would grab a number of permits equal to its shard count, limited by the maximum chosen for the Semaphore.  The POC here protects against what a backup client can't possibly protect against -- a collection with way more shards than there are nodes.  By the way -- that's really rare as far as Solr use-cases out there, so the POC is motivated by providing a simple solution to something rare.

It would be nice to handle the naive/abuse case better though.  It wouldn't be hard to fail the backup if the number of waiters on the Semaphore in this POC is very large (> 100?). Yes it's possible for a partial backup to fail but that's okay.  Simple protection vs absolute efficiency / guarantee.

I already provided an option for prioritizing restores over backups -- which doesn't seem to favor an "early" vs "late" design.

psalagnac:
Main issue with your approach is thread starvation in the collection API.

I think we have 50 threads dedicated to this in each node. Assuming we do a backup for a very large collection (with much more shards per node than this number of threads), all threads will be either doing a core backup or waiting for another to be done. My understanding is all other operations at the collection API level will be on hold.

I don't think throttling should be done at the cost of blocking most of the threads. An other operation which is not an IO consumer (like creating an empty collection) should ideally not be impacted by the large backup. That's why I think throttinlgw should happen at a higher level. 

When throttinlgw in `BackupCmd`, a single thread is waiting for all others of the backup.

psalagnac:
Following our offline discussion, adding details here on my point about the thread starvation.

When running a core admin request in async mode, they are all done by a thread pool defined in `CoreAdminHandler`.
```
    private ExecutorService parallelExecutor =
        ExecutorUtil.newMDCAwareFixedThreadPool(
            50, new SolrNamedThreadFactory("parallelCoreAdminAPIBaseExecutor"));
```

If we block many core snapshots (or any other core operation) with a semaphore, we'll have the next eventual operations that may not be processed immediately, even if they should (for example, a large backup may block a collection creation).

PR #1732, opened 2023-06-27
Remove DistributedQueueFactory

dsmiley:
It's not used since Solr 9, nor do I think we need such a thing.


solr/core/src/test/org/apache/solr/cloud/OverseerTest.java
```diff
@@ -1947,8 +1947,7 @@ private SolrCloudManager getCloudDataProvider(String zkAddress, SolrZkClient zkC
             .withConnectionTimeout(15000, TimeUnit.MILLISECONDS)
             .build();
     solrClients.add(client);
-    SolrClientCloudManager sccm =
-        new SolrClientCloudManager(new ZkDistributedQueueFactory(zkClient), client);
+    SolrClientCloudManager sccm = new SolrClientCloudManager(client);
```
cpoerschke, line 1950: minor: the `zkClient` argument for the method is unused now then

sigram (approved):
Looks good to me +1.

PR #1738, opened 2023-06-29
SOLR-15753: Ensure v2 API requests are logged

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-15753

# Description

v1 requests are logged out in two main places: HttpSolrCall and SolrCore.  This request logging is very useful for debugging and other purposes.  But no request logging exists for v2 API calls.

# Solution

This commit adds request logging for v2 requests, via a JAX-RS "post-request filter".  In an attempt to maintain compatibility with existing logging dashboards, v2 logging uses the same HttpSolrCall and SolrCore.Request loggers that are currently used to log v1 requests.

Some of the niceties built up around our v1 request logging are also supported for v2: including the `logParamsList` parameter used to filter which parameters get logged, and SLF4J "marker" support.

# Tests

Manual testing that request logging works as expected for v2 requests.  Added unit tests (mostly around how parts of requests are flattened into a string, not so much around the logging itself) in `PostRequestLoggingFilterTest`.

# Checklist


HoustonPutman (commented):
This is awesome!

I am so thankful you are a wizard with the Jersey stuff 😅 

This might not be possible, but given that so many APIs are moving from GET to POST/PUT, would it be possible to log the body of the request as well?

gerlowskija:
> I am so thankful you are a wizard with the Jersey stuff

Haha, thanks - but I swear I'm just one or two frantic google-searches ahead of everyone else haha!

> would it be possible to log the body of the request as well?

I definitely want to log out the body of POST and PUT requests, but I'm still figuring out how to make that work.

The main hurdle is that Jersey filters have access to the request body as an InputStream, but since the logging filter in this PR is a "response filter" that runs _after_ the API method, that InputStream has already been consumed and is mostly useless.  The workaround for this is to have something cache whatever POJO the request body is deserialized into, so that it's accessible by the logging filter, but I haven't found a non-hacky way to do that yet.

That's the main issue, though there are a few others (e.g. do we want to try to extend `logParamsList` in some way to cover potentially nested request bodies?). So unless you think we really should include it here, I was probably gonna work on request-body logging in a separate PR (but the same JIRA).

gerlowskija:
> [Houston] would it be possible to log the body of the request as well?

> [Jason] I was probably gonna work on request-body logging in a separate PR

I've created a PR [here](https://github.com/gerlowskija/solr/pull/3) that logs the request body as well.  It ended up not being too bad!

I think it's the best we're likely to come by, but there's still a bit of hackiness to the approach.  It also doesn't extend `logParamsList` to the request body itself, and I'm curious whether anyone feels particularly strongly about that one way or another.  Both of those concerns seem to merit a bit more time in review IMO, so I'm going to keep it in a separate PR.

HoustonPutman:
I think it's great, doesn't look particularly hacky to me. But I guess I don't know much here.

Anyways, I think this would be great to get in for 9.3, and something we can always improve on in later versions!

gerlowskija:
> doesn't look particularly hacky to me

I guess I'm thinking particularly of a few bits, like manually stripping newlines out of the JSON string that gets produced by `Utils.toJsonString(o)`.

But I agree that those warts probably aren't significant enough to merit missing 9.3.  I'll fold the changes in here and look to merge tomorrow or shortly after.

gerlowskija:
Alright - tests and check are passing, I've added tests in PostRequestLoggingFIlterTest, and a few doc tweaks clarifying `logParamsList` non-effect on request bodies and path parameters.  Plan on merging later today!

PR #1740, opened 2023-06-30
SOLR-16490: Create a v2 equivalent for /admin/cores BACKUPCORE

iamsanjay:
https://issues.apache.org/jira/browse/SOLR-16490

# Description

Create a v2 equivalent for /admin/cores BACKUPCORE

# Solution

Created a new class BackupAPI.java, a new Jax RS endpoints, that contains the BackUpCoreOp.java code. Now BackupCoreOp.java class is calling this new class to create SnapShooter or the Incremental Backup.

# Tests

Created one test class that testing both Incremental and SnapShooter backup commands on.
	1	The Valid response.
	2	Missing Parameters.

# Checklist


solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
```diff
@@ -0,0 +1,179 @@
...
+      SolrQueryResponse solrQueryResponse,
+      CoreAdminHandler.CoreAdminAsyncTracker coreAdminAsyncTracker) {
+    super(coreContainer, coreAdminAsyncTracker, solrQueryRequest, solrQueryResponse);
+  }
+
+  @POST
+  @Produces({"application/json", "application/xml", BINARY_CONTENT_TYPE_V2})
+  @PermissionName(COLL_EDIT_PERM)
+  public SolrJerseyResponse createBackup(
+      @Schema(description = "The name of the core.") @PathParam("core") String coreName,
```
gerlowskija, line 65: [-1] Shouldn't the PathParam annotation value be "coreName" instead of "core"?  I've been thinking the value of these annotations need to match whatever the placeholder is up in the `@Path` annotation (see L50).  If this works as-is, then I've been misunderstanding how these work for a loong time 😮

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
```diff
@@ -0,0 +1,179 @@
...
+import org.apache.solr.core.backup.repository.BackupRepository;
+import org.apache.solr.handler.IncrementalShardBackup;
+import org.apache.solr.handler.SnapShooter;
+import org.apache.solr.handler.admin.CoreAdminHandler;
+import org.apache.solr.jersey.PermissionName;
+import org.apache.solr.jersey.SolrJerseyResponse;
+import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.response.SolrQueryResponse;
+
+@Path(("/cores/{coreName}/backup/{name}"))
```
gerlowskija, line 50: [Q] Are the extra set of parenthesis around the path value serving a particular purpose that I'm missing?

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
```diff
@@ -0,0 +1,179 @@
...
+  @Produces({"application/json", "application/xml", BINARY_CONTENT_TYPE_V2})
+  @PermissionName(COLL_EDIT_PERM)
+  public SolrJerseyResponse createBackup(
+      @Schema(description = "The name of the core.") @PathParam("core") String coreName,
+      @Schema(description = "The backup will be created in a directory called snapshot.<name>")
+          @PathParam("name")
+          String name,
+      @Schema(description = "The POJO for representing additional backup params.") @RequestBody
+          BackupCoreRequestBody backupCoreRequestBody,
+      @Parameter(description = "The id to associate with the async task.") @QueryParam("async")
```
gerlowskija, line 71: [-1] I really like the idea of making "async" a query parameter instead of including it in the request body.

The request body of a POST or PUT is supposed to represent the resource being created or updated.  Since "async" is more about _how_ the request is processed, and less about the resource itself, I think it makes sense to keep out of the request body.

But I'm worried about inconsistency across the many APIs that accept the "async" param.  So far, all of these put "async" in the request body.  And, personally, I'd rather keep the Solr's APIs as a whole a bit more consistent with one another than make this specific API a little bit better.

That's all to say: can we keep "async" in the request body for now?

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
```diff
@@ -0,0 +1,179 @@
...
+        "backup",
+        () -> {
+          try (BackupRepository repository =
+                  coreContainer.newBackupRepository(backupCoreRequestBody.repository);
+              SolrCore core = coreContainer.getCore(coreName)) {
+            String location = repository.getBackupLocation(backupCoreRequestBody.location);
+            if (location == null) {
+              throw new SolrException(
+                  SolrException.ErrorCode.BAD_REQUEST,
+                  "'location' is not specified as a query"
```
gerlowskija, line 91: [-0] Should probably update "query parameter" in this message to just read "parameter" - since `location` now appears in the request body of v2 requests.

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
```diff
@@ -0,0 +1,179 @@
...
+                      + " parameter or as a default repository property");
+            }
+            URI locationUri = repository.createDirectoryURI(location);
+            repository.createDirectory(locationUri);
+
+            if (backupCoreRequestBody.incremental) {
+              if ("file".equals(locationUri.getScheme())) {
+                core.getCoreContainer().assertPathAllowed(Paths.get(locationUri));
+              }
+              if (backupCoreRequestBody.shardBackupId == null) {
```
gerlowskija, line 101: [0] You can replace the null-check-and-throw block here (and in a few other places) with a call to `JerseyResource.ensureRequiredParameterProvided` which should be inherited and available here.

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
```diff
@@ -0,0 +1,179 @@
...
+              SnapShooter snapShooter =
+                  new SnapShooter(
+                      repository, core, locationUri, name, backupCoreRequestBody.commitName);
+              // validateCreateSnapshot will create parent dirs instead of throw; that choice is
+              // dubious.
+              // But we want to throw. One reason is that this dir really should, in fact must,
+              // already
+              // exist here if triggered via a collection backup on a shared file system. Otherwise,
+              // perhaps the FS location isn't shared -- we want an error.
+              if (!snapShooter.getBackupRepository().exists(snapShooter.getLocation())) {
```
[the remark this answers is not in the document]
gerlowskija: Present in code only moved as a part of this PR.

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
```diff
@@ -0,0 +1,179 @@
...
+
+    @Schema(
+        description =
+            "The name of the commit which was used while taking a snapshot using the CREATESNAPSHOT command.")
+    @JsonProperty("commitName")
+    public String commitName;
+
+    @Schema(description = "To turn on incremental backup feature")
+    @JsonProperty("incremental")
+    public boolean incremental;
```
gerlowskija, line 177: [-1] We should have this be a `Boolean`, so that BackupCoreAPI has a way to tell whether users actually requested 'false' in their request or whether they just didn't provide a value.  This property is supposed to default to 'true', so it's pretty important to know whether users explicitly requested 'false' or not.

solr/core/src/test/org/apache/solr/handler/admin/api/BackupCoreAPITest.java
```diff
@@ -0,0 +1,178 @@
...
+
+    CoreAdminHandler.CoreAdminAsyncTracker coreAdminAsyncTracker =
+        new CoreAdminHandler.CoreAdminAsyncTracker();
+    backupCoreAPI =
+        new BackupCoreAPI(
+            coreContainer, solrQueryRequest, solrQueryResponse, coreAdminAsyncTracker);
+  }
+
+  @Test
+  public void testCreateBackupReturnsValidResponse() throws Exception {
```
gerlowskija, line 66: [0] Maybe add "NonIncremental" before "Backup" in the test name, to differentiate it a bit more from the Incremental version of this test below?

solr/core/src/test/org/apache/solr/handler/admin/api/BackupCoreAPITest.java
```diff
@@ -0,0 +1,178 @@
...
+            });
+    assertEquals(500, solrException.code());
+    assertTrue(
+        "Exception message differed from expected: " + solrException.getMessage(),
+        solrException.getMessage().contains("'location' is not specified as a query"));
+  }
+
+  @Test
+  public void testMissingCoreNameParameter() throws Exception {
+    final String backupName = "my-new-backup";
```
gerlowskija, line 100: [0] If you wanted you could create a static constant at the top of the test, and then use that in each test method.

solr/core/src/java/org/apache/solr/handler/IncrementalShardBackup.java
```diff
@@ -134,14 +135,13 @@ private IndexCommit getAndSaveLatestIndexCommit(IndexDeletionPolicyWrapper delPo
   }
 
   // note: remember to reserve the indexCommit first so it won't get deleted concurrently
-  protected NamedList<Object> backup(final IndexCommit indexCommit) throws Exception {
+  protected IncrementalBackupCoreResponse backup(final IndexCommit indexCommit) throws Exception {
```
gerlowskija, line 138: [+1] Always great to see NamedList usages go away!

solr/core/src/java/org/apache/solr/handler/SnapShooter.java
```diff
@@ -277,7 +281,8 @@ public void createSnapAsync(final int numberToKeep, Consumer<NamedList<?>> resul
    * @see IndexDeletionPolicyWrapper#saveCommitPoint
    * @see IndexDeletionPolicyWrapper#releaseCommitPoint
    */
-  protected NamedList<Object> createSnapshot(final IndexCommit indexCommit) throws Exception {
+  protected SnapShooterBackupCoreResponse createSnapshot(final IndexCommit indexCommit)
```
gerlowskija, line 284: [Q] Looking at createSnapAsync, it looks like this same `createSnapshot()` code is also called by Solr's "ReplicationHandler" (i.e. `/replication`) APIs.

Maybe the name of the response class should be made more generic so that it's not so "Backup Core API" specific?  Maybe something like `CoreSnapshotResponse` or `CoreSnapshotDetails`?  Idk, just thinking aloud...

solr/core/src/java/org/apache/solr/handler/admin/BackupCoreOp.java
```diff
@@ -17,84 +17,48 @@
...
+      backupCoreRequestBody.prevShardBackupId =
+          params.get(CoreAdminParams.PREV_SHARD_BACKUP_ID, null);
+      backupCoreRequestBody.incremental = true;
+    }
+    BackupCoreAPI backupCoreAPI =
+        new BackupCoreAPI(
+            it.handler.coreContainer, it.req, it.rsp, it.handler.coreAdminAsyncTracker);
+    try {
+      SolrJerseyResponse response =
+          backupCoreAPI.createBackup(cname, name, backupCoreRequestBody, null);
```
gerlowskija, line 57: [Q/-1] Is there a reason why the last parameter of this method is always 'null'?  I would've expected it to get  `params.get("async")` or something similar?

solr/core/src/test/org/apache/solr/handler/admin/api/BackupCoreAPITest.java
```diff
@@ -0,0 +1,178 @@
...
+import org.apache.solr.handler.SnapShooter;
+import org.apache.solr.handler.admin.CoreAdminHandler;
+import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.response.SolrQueryResponse;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class BackupCoreAPITest extends SolrTestCaseJ4 {
```
gerlowskija, line 39: [+1] Great tests!

gerlowskija (changes requested):
Thanks for the PR Sanjay!  Exciting to close these gaps in v2 coverage.  The bones of this all look pretty good, though there were a few things I commented on that I think will need changed.  Feel free to reply inline and we can go from there.

I know our ref-guide docs around the core-backup functionality is pretty patchy, we'll probably want to update those to include the new "v2" version of the API.

solr/core/src/java/org/apache/solr/handler/admin/BackupCoreOp.java
iamsanjay: I believe v1 calls come through `CoreAdminHandler.handleRequestBody()` and it looks for `async` parameter. If present, they create an CoreAdminOp object, such as  BackupCoreOp in our case and then run it asynchronously via calling `CoreAdminAsyncTracker.submitAsyncTask()`. As per my understanding, if we start submitting `async` in `BackupCoreOp` as well then same task will be submitted twice to `ExecutorService`. https://github.com/apache/solr/blob/583fdf0cc7438cd3a50e4d0aa865281ff209c5c6/solr/core/src/java/org/apache/solr/handler/admin/CoreAdminHandler.java#L196-L227
But if call come via v2 api, then we are passing the actual taskID, not null. 
https://github.com/apache/solr/blob/cbe661547d7f12f0c4d21e6fef4e03455b97575c/solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java#L78-L84
Same way we are handling in `CreateSnapshotOp.java`.
https://github.com/apache/solr/blob/583fdf0cc7438cd3a50e4d0aa865281ff209c5c6/solr/core/src/java/org/apache/solr/handler/admin/CreateSnapshotOp.java#L37-L38

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
iamsanjay: It should be the same name as the one in parenthesis. Will change it!

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
iamsanjay: Not sure. Going to remove the extra ones!

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
iamsanjay: There is one API (`CoreSnapshotAPI.java`) class in which it used as QueryParam. But most of the classes are using via RequestBody. So I am going to include in RequestBody as well.

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
iamsanjay: I was thinking of using the `JerseyResource.ensureRequiredParameterProvided` here as well. But looks like here we need more custom error message. I will just change the error message!

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
iamsanjay: I will change it to `Boolean`. Also,
[Q] Should I set default value to 'true' here? Because, and also you mentioned, in `BackupCoreOp.java, If `incremental` is missing they set it to 'true'.

iamsanjay:
> Thanks for the PR Sanjay! Exciting to close these gaps in v2 coverage. The bones of this all look pretty good, though there were a few things I commented on that I think will need changed. Feel free to reply inline and we can go from there.
> 
> I know our ref-guide docs around the core-backup functionality is pretty patchy, we'll probably want to update those to include the new "v2" version of the API.

I search through documentation we never really mentioned 'BACKUPCORE' operation. We only provided documentation where the backup is done via `ReplicationHandler` for single core mode.

solr/core/src/java/org/apache/solr/handler/SnapShooter.java
iamsanjay: `CoreSnapshotResponse` looks good because most of our JerseyResponse classes ends with 'Response' keyword. There is another class named `IncrementalBackupCoreResponse.java`. Should I change it to `IncrementalShardSnapshotResponse.java`.

solr/core/src/java/org/apache/solr/handler/admin/BackupCoreOp.java
gerlowskija: OK, makes sense; good catch.

solr/core/src/java/org/apache/solr/handler/SnapShooter.java
gerlowskija: Sure, renaming them both consistently makes sense to me?

gerlowskija (approved):
LGTM; thanks @iamsanjay !  Working through 'check' and tests now, but hoping to merge soon!

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
```diff
@@ -0,0 +1,171 @@
...
+                  "'location' parameter is not specified in the request body or as a default repository property");
+            }
+            URI locationUri = repository.createDirectoryURI(location);
+            repository.createDirectory(locationUri);
+
+            if (backupCoreRequestBody.incremental) {
+              if ("file".equals(locationUri.getScheme())) {
+                core.getCoreContainer().assertPathAllowed(Paths.get(locationUri));
+              }
+              ensureRequiredParameterProvided("shardBackupId", backupCoreRequestBody.shardBackupId);
```
gerlowskija, line 93: [-1] `shardBackupId` shouldn't be required idt, since we technically still support "non-incremental" backups which don't build off of a previous backup.

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
```diff
@@ -0,0 +1,171 @@
...
+import org.apache.solr.core.backup.repository.BackupRepository;
+import org.apache.solr.handler.IncrementalShardBackup;
+import org.apache.solr.handler.SnapShooter;
+import org.apache.solr.handler.admin.CoreAdminHandler;
+import org.apache.solr.jersey.PermissionName;
+import org.apache.solr.jersey.SolrJerseyResponse;
+import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.response.SolrQueryResponse;
+
+@Path("/cores/{coreName}/backup/{name}")
```
gerlowskija, line 48: [-1] I think I steered you towards this using this path value, so this is totally my fault, but I don't think we can put the backup name in the path.

I hadn't realized it until debugging some tests this afternoon, but the backup name only exists for non-incremental core backups.  Incremental core backups don't even have a concept of "backup names" (for reference see [this code in BackupCoreOp that parses the param](https://github.com/apache/solr/pull/1740/files#diff-0c6a9444b6016570c565957acf8a067084b4808a50cd96e89e3e514cdcbbb675R80))

Path parameters are almost always required values, so we'll have to move the param elsewhere (probably into the request body).

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
```diff
@@ -0,0 +1,171 @@
...
+      @Parameter(description = "The name of the core.") @PathParam("coreName") String coreName,
+      @Schema(description = "The backup will be created in a directory called snapshot.<name>")
+          @PathParam("name")
+          String name,
+      @Schema(description = "The POJO for representing additional backup params.") @RequestBody
+          BackupCoreRequestBody backupCoreRequestBody)
+      throws Exception {
+    ensureRequiredParameterProvided("coreName", coreName);
+    return handlePotentiallyAsynchronousTask(
+        null,
```
gerlowskija, line 72: [-1] This value cannot be 'null' - `handlePotentiallyAsynchronousTask` uses it to squash receive whatever response data is generated by the lambda below.  I think we're getting an NPE in some tests because of this.

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
```diff
@@ -0,0 +1,171 @@
...
+            String location = repository.getBackupLocation(backupCoreRequestBody.location);
+            if (location == null) {
+              throw new SolrException(
+                  SolrException.ErrorCode.BAD_REQUEST,
+                  "'location' parameter is not specified in the request body or as a default repository property");
+            }
+            URI locationUri = repository.createDirectoryURI(location);
+            repository.createDirectory(locationUri);
+
+            if (backupCoreRequestBody.incremental) {
```
gerlowskija, line 89: [-1] If a v2 request comes in that doesn't specify "incremental" in the request body, the `Boolean` value here might be null and this conditional will cause a NPE

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
gerlowskija: My mistake - I thought I was tracing a test failure to this, but was misreading the stack.  Ignore

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
gerlowskija: Gah - another misconception I had while debugging test failures.  Forget I said anything.

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
gerlowskija: I've updated the path to `/cores/<coreName>/backups`, and created a field in the request body to hold the "backup name" parameter.

solr/core/src/java/org/apache/solr/handler/admin/api/BackupCoreAPI.java
gerlowskija: Fixed - there's a nice little recipe to handle the extra 'null' case around `Boolean` types: `Boolean.TRUE.equals(backupCoreRequestBody.incremental)`

PR #1751, opened 2023-07-04
SOLR-16470: Inner class is non-static but does not reference enclosin…

mkhludnev:


mkhludnev:
@gerlowskija it looks like I don't need to put to branch_9x. Right?

risdenk (commented):
Fix looks good

PR #1758, opened 2023-07-06
SOLR-16869: Use Panama Vector apis for Java 20/21

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-16869

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -229,6 +229,9 @@ Optimizations
 * SOLR-16846: Use IndexOrDocValuesQuery for PointFields on Set queries. This improves the performance in some scenarios when
   the set query is combined with other selective queries. (Tomás Fernández Löbbe, Alex Deparvu)
 
+* SOLR-16869: Enable Incubating Java Panama Vector APIs when using Java 20 or 21.
+  When run using compatible hardware, Vector search will see considerable speed up. (Houston Putman, Uwe Schindler)
+
 Bug Fixes
 ---------------------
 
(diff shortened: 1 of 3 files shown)


stillalex:
+1 tested on my local with jdk20 and was the message, as expected:

```
Java 20 detected. Incubating Panama Vector APIs have been enabled
```

uschindler (approved):
Tested on Windows with Java 17, 20 and 21. All start correctly and print right messages.

Theres still the problem of Solr not using the Java Util Logging appender, so Lucene's logs get lost in console and are not correctly logged to its main log file.

I did not check Vector indexing (I have no vectors available). The techdocs example should maybe add some vectors.

HoustonPutman:
> Theres still the problem of Solr not using the Java Util Logging appender provided by Log4j, so Lucene's logs get lost in console and are not correctly logged to Solr's main log file.

I'll try to fix this in a separate Issue/PR. Thanks for raising it.

uschindler:
This one should be installed: https://logging.apache.org/log4j/2.x/log4j-jul.html

uschindler:
I think we need to add the JAR file next to the log4j JAR files in the server distribution and pass the correct command line arg. If that's done Lucene's logging messages should appear in Solr's log file. You will see this when Java 20 starts and it tells you that vector API was enabled or Memory Segments are used.

HoustonPutman:
Sounds good. Will have a PR soon. https://issues.apache.org/jira/browse/SOLR-16878

PR #1760, opened 2023-07-06
Fix TestSolrCLIRunExample Nightly Thread Leak

risdenk:
TestSolrCLIRunExample has been failing thread leak in nightly since the Http2SolrClient has not been closed in `RunExampleTool`

### DIFF
--- a/solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
@@ -328,75 +328,64 @@ public class RunExampleTool extends ToolBase {
     } else if ("films".equals(exampleName) && !alreadyExists) {
       try (SolrClient solrClient = new Http2SolrClient.Builder(solrUrl).build()) {
         echo("Adding dense vector field type to films schema \"_default\"");
-        try {
-          SolrCLI.postJsonToSolr(
-              solrClient,
-              "/" + collectionName + "/schema",
-              "{\n"
-                  + "        \"add-field-type\" : {\n"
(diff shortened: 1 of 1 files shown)


HoustonPutman (changes requested):
Looks good to me, but I don't think the inner try {} statements are needed, you can just catch the exception of the outer try and wrap in a SolrException, since that is done everytime.

solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
```diff
@@ -326,98 +326,98 @@ protected void runExample(CommandLine cli, String exampleName) throws Exception
             "exampledocs directory not found, skipping indexing step for the techproducts example");
       }
     } else if ("films".equals(exampleName) && !alreadyExists) {
-      SolrClient solrClient = new Http2SolrClient.Builder(solrUrl).build();
+      try (SolrClient solrClient = new Http2SolrClient.Builder(solrUrl).build()) {
+        echo("Adding dense vector field type to films schema \"_default\"");
+        try {
+          SolrCLI.postJsonToSolr(
```
epugh, line 332: Look at all the string handling for JSON!  I wondered about maybe using the FieldTypeDefinition.java, but after looking at how it's used, it seems to add another level of complexity/obfuscation over what is happening.   I hope our future V2 api's make doing thinks like adding a field type simpler and cleaner ;-).

epugh (approved):
LGTM.   THanks for fixing this!

risdenk:
Thanks @epugh for the review and @HoustonPutman makes sense to make and for the commit.

PR #1762, opened 2023-07-06
SOLR-16871: Race condition in `CoordinatorHttpSolrCall` synthetic collection/replica init

patsonluk:
https://issues.apache.org/jira/browse/SOLR-16871

# Description

From a unit test case [that issue concurrent select queries to coordinator nodes](https://github.com/cowpaths/fullstory-solr/blob/e4226eb8fa2afb01d7615f7faea01f71b144cd58/solr/core/src/test/org/apache/solr/search/TestCoordinatorRole.java#L486), it’s found that there could be 3 race condition issues:

1. If multiple concurrent requests find the synthetic collection is not yet created, they might all attempt to create the synthetic collection. This could trigger SolrException on `collection already exists`

2. Similarly, if multiple concurrent requests find there’s no replica of the synthetic collection for current node (multiple coordinator node scenario), then CoordinatorHttpSolrCall#addReplica could be invoked multiple times. This should not trigger any exception, but would create multiple replicas for the same node in the synthetic collection

3. The existing logic [here](https://github.com/cowpaths/fullstory-solr/blob/6c8531f08301a291478502c262499abed0d5075c/solr/core/src/java/org/apache/solr/servlet/CoordinatorHttpSolrCall.java#L102) assumes if syntheticColl.getReplicas(solrCall.cores.getZkController().getNodeName()) returns non empty result, then the following call in [here](https://github.com/cowpaths/fullstory-solr/blob/6c8531f08301a291478502c262499abed0d5075c/solr/core/src/java/org/apache/solr/servlet/CoordinatorHttpSolrCall.java#L112) should return a core. Unfortunately, the first call can return a non empty list but with a DOWN replica if another request is in the progress of creating such replica. In this case, the solrCall.getCoreByCollection(syntheticCollectionName, isPreferLeader) would call super.getCoreByCollection at [here](https://github.com/cowpaths/fullstory-solr/blob/6c8531f08301a291478502c262499abed0d5075c/solr/core/src/java/org/apache/solr/servlet/CoordinatorHttpSolrCall.java#L69) which would return a null (since super impl only returns active replica). So CoordinatorHttpSolrCall#getCoreByCollection would end up calling CoordinatorHttpSolrCall#getCore , introducing an infinite loop and cause stack overflow

# Solution

1. For collection creation exception, check again if the collection exists, if so, ignore the exception and proceed
2. For replica, if the replica for such node already found in the DocCollection, then ensure that it's active using `zkStateReader.waitForState`. This avoids the infinite loop caused by the presence of `down` replica.

Take note that this does NOT avoid the 2nd issue above, concurrent requests can still create multiple replica for the same node in the synthetic collection, though it's probably benign (and unlikely)

Remarks: First attempt was actually provide proper locking to avoid race condition. However, it's quite tricky to get it right - might need to force refresh the zkReader and do multiple extra reads. The extra cost and complexity probably does not justify the gain.

# Tests

Added `TestCooridnatorRole#testConcurrentAccess` to reproduce the issue

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/servlet/CoordinatorHttpSolrCall.java
+++ b/solr/core/src/java/org/apache/solr/servlet/CoordinatorHttpSolrCall.java
@@ -208,9 +208,12 @@ public class CoordinatorHttpSolrCall extends HttpSolrCall {
   private static void addReplica(String syntheticCollectionName, CoreContainer cores) {
     SolrQueryResponse rsp = new SolrQueryResponse();
     try {
+      String coreName = syntheticCollectionName + "_" + "r1";
       CollectionAdminRequest.AddReplica addReplicaRequest =
           CollectionAdminRequest.addReplicaToShard(syntheticCollectionName, "shard1")
-              .setCreateNodeSet(cores.getZkController().getNodeName());
+              // we are fixing the name, so that no two replicas are created in the same node
+              .setCoreName(coreName)
+              .setNode(cores.getZkController().getNodeName());
       addReplicaRequest.setWaitForFinalState(true);
       cores
           .getCollectionsHandler()
@@ -221,6 +224,15 @@ public class CoordinatorHttpSolrCall extends HttpSolrCall {
             "Could not auto-create collection: " + Utils.toJSONString(rsp.getValues()));
       }
     } catch (SolrException e) {
+      if (e.getMessage().contains("replica with the same core name already exists")) {
+        // another request has already created a replica for this synthetic collection
+        if (log.isInfoEnabled()) {
+          log.info(
+              "A replica is already created in this node for synthetic collection: {}",
+              syntheticCollectionName);
+        }
+        return;
+      }
       throw e;
 
     } catch (Exception e) {
--- a/solr/core/src/java/org/apache/solr/servlet/CoordinatorHttpSolrCall.java
+++ b/solr/core/src/java/org/apache/solr/servlet/CoordinatorHttpSolrCall.java
@@ -102,21 +102,25 @@ public class CoordinatorHttpSolrCall extends HttpSolrCall {
           try {
             createColl(syntheticCollectionName, solrCall.cores, confName);
           } catch (SolrException exception) {
-            //concurrent requests could have created the collection hence causing collection exists exception
+            // concurrent requests could have created the collection hence causing collection exists
+            // exception
             createException = exception;
           } finally {
             syntheticColl =
-                    zkStateReader.getClusterState().getCollectionOrNull(syntheticCollectionName);
+                zkStateReader.getClusterState().getCollectionOrNull(syntheticCollectionName);
           }
 
-          //then indeed the collection was not created properly, either by this or other concurrent requests
+          // then indeed the collection was not created properly, either by this or other concurrent
+          // requests
           if (syntheticColl == null) {
             if (createException != null) {
-              throw createException; //rethrow the exception since such collection was not created
+              throw createException; // rethrow the exception since such collection was not created
             } else {
               throw new SolrException(
-                      SolrException.ErrorCode.SERVER_ERROR,
-                      "Could not locate synthetic collection [" + syntheticCollectionName + "] after creation!");
+                  SolrException.ErrorCode.SERVER_ERROR,
+                  "Could not locate synthetic collection ["
+                      + syntheticCollectionName
+                      + "] after creation!");
             }
           }
         }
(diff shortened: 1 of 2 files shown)


noblepaul (approved):
LGTM

noblepaul:
@patsonluk I've made the changes to avoid race condition in core creation as well

solr/core/src/java/org/apache/solr/servlet/CoordinatorHttpSolrCall.java
```diff
@@ -208,9 +208,12 @@ private static void setMDCLoggingContext(String collectionName) {
   private static void addReplica(String syntheticCollectionName, CoreContainer cores) {
     SolrQueryResponse rsp = new SolrQueryResponse();
     try {
+      String coreName = syntheticCollectionName + "_" + "r1";
```
patsonluk, line 211: Do we want to include node name as a part of the core name? otherwise 2 coordinator nodes might use same name for the replica?

noblepaul: I thought about that. Core names do not have to be unique

patsonluk: Good to know! 👍🏼 

In this case we might want to always call `waitForState` as in https://github.com/apache/solr/pull/1762/files#diff-eedf409265fc219f98f193ae89d3f1b09df78fe49f70bb5b9eaa6c6ff46e6ac7R143 since `addReplica` can now return while the target replica is still under construction by another request.

patsonluk: Ah I think we do need to add node name, as the [AddReplicaCmd logic](https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/cloud/api/collections/AddReplicaCmd.java#L388) would iterate through all slices and throw exception on any core name matches.

Therefore, for the 1st coordinator node, it's fine -> the replica is created by the collection creation call, which follows the standard name, ie `.sys.COORDINATOR-COLL-conf_shard1_replica_n1`.

The 2nd coordinator node is also fine -> `.sys.COORDINATOR-COLL-conf_r1`.

However starting from the 3rd coordinator, it would fail with infinite loop, since it will skip adding replica (due to the check linked above), but `solrCall.getCoreByCollection(syntheticCollectionName, isPreferLeader)` would not be able to load such core (since it's not in 3rd node), hence causing the infinite loop.

Such can be reproduced by modifying node count [here](https://github.com/apache/solr/pull/1762/files#diff-13f37ffd1807b53dacbf21317f6f6d3114448ae2e85c90835fa2f368e31b3cbaR489) in the test case from 2 to 3

noblepaul: true. I'm not even sure why someone added that check

patsonluk: Thank you @noblepaul !

I added another small commit 45654e1c5bff66413956202cade3d1b75733be17 to ensure that the fix work:
1. Changed coordinator node count 2 -> 4
2. Verified the replica count on the synthetic collection

Also a minor change to the addReplica flow, that we always check for replica status afterwards (since `addReplica` might now return if exception is thrown and caught, and with replica status not always active, we could run into infinite call loop, this is a rather rare case, but it doesn't hurt to check.

tflobbe:
@noblepaul @patsonluk, This test has been failing very frequently since merged, did you have time to take a look?

patsonluk:
> @noblepaul @patsonluk, This test has been failing very frequently since merged, did you have time to take a look?

@tflobbe thanks for raising the concern, do u have any links to failures?

patsonluk:
@noblepaul I think setting the core name does not work as it would not avoid duplicated core on the first coordinator node. I printed out the replica list on failure and it shows:


```
[core_node2:{
  "core":".sys.COORDINATOR-COLL-conf_shard1_replica_n1",
  "leader":"true",
  "node_name":"127.0.0.1:49656_solr",
  "base_url":"https://127.0.0.1:49656/solr",
  "state":"active",
  "type":"NRT",
  "force_set_state":"false"}, core_node3:{
  "node_name":"127.0.0.1:49656_solr",
  "base_url":"https://127.0.0.1:49656/solr",
  "core":".sys.COORDINATOR-COLL-conf_127.0.0.1_49656_solr",
  "state":"active",
  "type":"NRT",
  "force_set_state":"false"}, core_node4:{
  "node_name":"127.0.0.1:49647_solr",
  "base_url":"https://127.0.0.1:49647/solr",
  "core":".sys.COORDINATOR-COLL-conf_127.0.0.1_49647_solr",
  "state":"active",
  "type":"NRT",
  "force_set_state":"false"}, core_node5:{
  "node_name":"127.0.0.1:49650_solr",
  "base_url":"https://127.0.0.1:49650/solr",
  "core":".sys.COORDINATOR-COLL-conf_127.0.0.1_49650_solr",
  "state":"active",
  "type":"NRT",
  "force_set_state":"false"}, core_node6:{
  "node_name":"127.0.0.1:49653_solr",
  "base_url":"https://127.0.0.1:49653/solr",
  "core":".sys.COORDINATOR-COLL-conf_127.0.0.1_49653_solr",
  "state":"active",
  "type":"NRT",
  "force_set_state":"false"}]
```

Any thoughts? 🤔

patsonluk:
@noblepaul  what do u think about this proposed fix? 😊  https://github.com/apache/solr/pull/1794

PR #1766, opened 2023-07-07
Pass verbose flag to solr start when using example

HoustonPutman:
Small ticket so no JIRA needed.


epugh (approved):
LGTM.   `RunExampleTool` feels like it could use some love in the future...

PR #1773, opened 2023-07-11
SOLR-16886: Don't commit multi-part uploads that have been aborted

tflobbe:
The `complete()` method is currently being called from the `close()`, which is called from try-with-resources

### DIFF
--- a/solr/modules/s3-repository/src/java/org/apache/solr/s3/S3OutputStream.java
+++ b/solr/modules/s3-repository/src/java/org/apache/solr/s3/S3OutputStream.java
@@ -29,7 +29,6 @@ import software.amazon.awssdk.core.exception.SdkException;
 import software.amazon.awssdk.core.sync.RequestBody;
 import software.amazon.awssdk.services.s3.S3Client;
 import software.amazon.awssdk.services.s3.model.CompletedPart;
-import software.amazon.awssdk.services.s3.model.MultipartUpload;
 import software.amazon.awssdk.services.s3.model.UploadPartRequest;
 import software.amazon.awssdk.services.s3.model.UploadPartResponse;
 
@@ -205,7 +204,7 @@ public class S3OutputStream extends OutputStream {
     void uploadPart(ByteArrayInputStream inputStream, long partSize) {
       if (aborted) {
         throw new IllegalStateException(
-                "Can't upload new parts on an MultipartUpload that was aborted. id '" + uploadId + "'");
+            "Can't upload new parts on an MultipartUpload that was aborted. id '" + uploadId + "'");
       }
       int currentPartNumber = completedParts.size() + 1;
 
--- a/solr/modules/s3-repository/src/java/org/apache/solr/s3/S3OutputStream.java
+++ b/solr/modules/s3-repository/src/java/org/apache/solr/s3/S3OutputStream.java
@@ -163,13 +163,17 @@ public class S3OutputStream extends OutputStream {
       return;
     }
 
-    if (multiPartUpload == null || !multiPartUpload.aborted) {
-      // flush first
-      uploadPart();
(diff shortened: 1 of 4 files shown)


solr/modules/s3-repository/src/java/org/apache/solr/s3/S3OutputStream.java
```diff
@@ -200,6 +202,10 @@ public MultipartUpload(String uploadId) {
     }
 
     void uploadPart(ByteArrayInputStream inputStream, long partSize) {
+      if (aborted) {
```
HoustonPutman, line 205: this could just return like `complete()` below. Any reason to error here and not there? It'd make the logic a bit simpler in `close()` as well.

tflobbe: It's true that this isn't consistent with the logic added in `close`. I kind of like throwing the error instead of just ignoring things because it otherwise hides the fact that the component is being misused (an upload after an abort). The reality is that none of the error paths should happen with the current code. I'm more inclined to throw the ISE exception in the `compoete()` if `aborted` than going the other way around.

>It'd make the logic a bit simpler in close() as well.

Maybe it's simpler if I check for `aborted` in `close()` too and handled separately? let me push a change and see

tflobbe:
> Nice catch. Is there a chance to cover it with some sort of unit test?

I was trying to see if it was possible with `s3mock`, but I can't find a way to introduce failures. I could do a Mockito test, or otherwise increase visibility to some methods for testing purposes

solr/modules/s3-repository/src/java/org/apache/solr/s3/S3OutputStream.java
HoustonPutman: yeah that makes sense to me.

solr/modules/s3-repository/src/test/org/apache/solr/s3/S3OutputStreamMockitoTest.java
```diff
@@ -0,0 +1,130 @@
...
+            .getCause()
+            .getMessage()
+            .contains("Can't upload new parts on a MultipartUpload that was aborted"));
+
+    verify(clientMock)
+        .createMultipartUpload((Consumer<CreateMultipartUploadRequest.Builder>) any());
+    verify(clientMock).uploadPart((UploadPartRequest) any(), (RequestBody) any());
+    verify(clientMock, never())
+        .completeMultipartUpload((Consumer<CompleteMultipartUploadRequest.Builder>) any());
+    verify(clientMock, times(2))
```
HoustonPutman, line 123: why does it call abort twice?

HoustonPutman (approved):
Looks good to me. The mockito tests are clean and test what we need!

solr/modules/s3-repository/src/test/org/apache/solr/s3/S3OutputStreamMockitoTest.java
tflobbe: The first time is after calling `write` for the first time (in line 102 in this test). The S3 client's `uploadPart` will fail, and after that our `S3OutputStream` will capture the exception and call abort. The second time is because we call `write` again in line 109 of this test, in this case, the exception i thrown by our `MultipartUpload` code (the `IllegalStateException`) because MPU has already been aborted. This will call abort again. This shouldn't be a problem, since one can call abort on the same ID multiple times (in fact, the javadocs say that one should, to ensure better cleanup of multipart objects, but this isn't really addressing that part)

HoustonPutman: Thanks for the explanation!

PR #1774, opened 2023-07-11
SOLR-16397: Create v2(JAX-RS) endpoints for RELOAD core

iamsanjay:

# Description

V2(JAX-RS) Endpoints for core admin op "RELOAD".

# Solution

There is already one class present in the codebase named as `ReloadCoreAPI.java`. New code changes is to make it using JAX-RS framework.

# Tests

Three test cases currently.
1. To test the valid response and status message.
2. Test Non Existent core exception and status code.
3. Test Unable to reload Exception and status code.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
+++ b/solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
@@ -89,7 +89,7 @@ public class CPUCircuitBreaker extends CircuitBreaker {
 
   @Override
   public String getErrorMessage() {
-    return "CPU Circuit Breaker triggered as seen CPU usage is above allowed threshold."
+    return "CPU Circuit Breaker triggered as seen CPU usage is above allowed threshold. "
         + "Seen CPU usage "
         + seenCPUUsage.get()
         + " and allocated threshold "
--- a/solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
+++ b/solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
@@ -89,7 +89,7 @@ public class MemoryCircuitBreaker extends CircuitBreaker {
 
   @Override
   public String getErrorMessage() {
-    return "Memory Circuit Breaker triggered as JVM heap usage values are greater than allocated threshold."
+    return "Memory Circuit Breaker triggered as JVM heap usage values are greater than allocated threshold. "
         + "Seen JVM heap memory usage "
         + seenMemory.get()
         + " and allocated threshold "
--- a/solr/core/src/test-files/solr/collection1/conf/solrconfig-pluggable-circuitbreaker.xml
+++ b/solr/core/src/test-files/solr/collection1/conf/solrconfig-pluggable-circuitbreaker.xml
@@ -79,7 +79,7 @@
   </query>
 
   <circuitBreaker class="solr.MemoryCircuitBreaker">
-    <double  name="threshold">75</double>
+    <double  name="threshold">99</double>
     <arr name="requestTypes">
       <str>update</str>
     </arr>


solr/core/src/test/org/apache/solr/handler/admin/api/ReloadCoreAPITest.java
```diff
@@ -0,0 +1,155 @@
...
+            new SolrException(
+                SolrException.ErrorCode.BAD_REQUEST, "No such core: " + NON_EXISTENT_CORE))
+        .when(mockCoreContainer)
+        .reload("demo");
+    final Response response =
+        target("/cores/" + NON_EXISTENT_CORE + "/reload").request().post(Entity.json(""));
+    final String responseJsonStr = response.readEntity(String.class).toString();
+    assertEquals(400, response.getStatus());
+    assertTrue(responseJsonStr.contains("No such core"));
+  }
```
iamsanjay, line 138: @gerlowskija I added test cases to check the valid and error response but for some reason where I am throwing exception inside the API through mock, the test cases are failing. 
Upon further inspection, When API class throws error then It goes somehow to `CatchAllExceptionMapper.toResponse()` and there when it tries to get "solrQueryResponse" object containerRequestContext, it returns null.

iamsanjay: PR still in draft mode! When the test cases will be fixed. I will change it.

solr/core/src/java/org/apache/solr/handler/admin/api/ReloadCoreAPI.java
```diff
@@ -17,52 +17,56 @@
...
 
 /**
- * V2 API for reloading an individual core.
+ * V2 API for reloading collections.
  *
- * <p>The new API (POST /v2/cores/coreName {'reload': {...}}) is equivalent to the v1
- * /admin/cores?action=reload command.
- *
- * @see ReloadCorePayload
+ * <p>The new API (POST /v2/core/coreName/reload {...}) is analogous to the v1
```
gerlowskija, line 39: [0] This comment needs updated to match the value in the `@Path` annotation below (i.e. `/v2/core` should be changed to `/v2/cores`)

Also, we can probably delete the `{...}` bit, since this request doesn't actually take a request body.

solr/core/src/java/org/apache/solr/handler/admin/api/ReloadCoreAPI.java
```diff
@@ -17,52 +17,56 @@
...
+      @Parameter(description = "The name of the core to snapshot.", required = true)
+          @PathParam("coreName")
+          String coreName) {
+    SolrJerseyResponse solrJerseyResponse = instantiateJerseyResponse(ReloadCoreResponse.class);
+    coreContainer.reload(coreName);
+    return solrJerseyResponse;
   }
 
-  public static class ReloadCorePayload implements ReflectMapWriter {}
+  public static class ReloadCoreResponse extends SolrJerseyResponse {}
```
gerlowskija, line 71: [-1] No need to create an empty Response class if this API doesn't return anything beyond what's in a normal `SolrJerseyResponse`; we can just have the method above return a SJR instance.

solr/core/src/java/org/apache/solr/handler/admin/api/ReloadCoreAPI.java
```diff
@@ -17,52 +17,56 @@
...
-
-    coreHandler.handleRequestBody(wrapParams(obj.getRequest(), v1Params), obj.getResponse());
+  @POST
+  @Produces({"application/json", "application/xml", BINARY_CONTENT_TYPE_V2})
+  @PermissionName(CORE_EDIT_PERM)
+  public SolrJerseyResponse reloadCore(
+      @Parameter(description = "The name of the core to snapshot.", required = true)
+          @PathParam("coreName")
+          String coreName) {
+    SolrJerseyResponse solrJerseyResponse = instantiateJerseyResponse(ReloadCoreResponse.class);
```
gerlowskija, line 66: [0] See my comment on L71 below about whether ReloadCoreResponse is really needed, or whether we can maybe use  `SolrJerseyResponse.class` as the method arg here.

solr/core/src/test/org/apache/solr/handler/admin/api/ReloadCoreAPITest.java
gerlowskija: Yeah, that's going to be pretty tough to mock out.  I think our best bet here is to test this at a slightly lower level (i.e. by calling ReloadCoreAPI.reloadCore directly and verifying the exception that it throws.  You've done similar stuff in other PRs, so I can use [your own code as an example](https://github.com/apache/solr/pull/1740/files#diff-ac75bc632270f1f0d082ebc4d82d4ffed7d162c2ec622aee7b38370fe960f2d2R86) 😛 )

solr/core/src/test/org/apache/solr/handler/admin/api/ReloadCoreAPITest.java
```diff
@@ -0,0 +1,155 @@
...
+  private static final String CORE_NAME = "demo";
+
+  @BeforeClass
+  public static void ensureWorkingMockito() {
+    assumeWorkingMockito();
+  }
+
+  @Override
+  public Application configure() {
+    forceSet(TestProperties.CONTAINER_PORT, "0");
```
gerlowskija, line 62: [0] I love these tests, but the stubbing you've had to do here seems painful.

I'd recommend we skip using JerseyTest here, and instead instantiate and call `ReloadCoreAPI` directly, like you've done in your [BACKUPCORE PR](https://github.com/apache/solr/pull/1740).

solr/solr-ref-guide/modules/configuration-guide/pages/coreadmin-api.adoc
```diff
@@ -318,11 +318,7 @@ http://localhost:8983/solr/admin/cores?action=RELOAD&core=techproducts
 
 [source,bash]
 ----
-curl -X POST http://localhost:8983/api/cores/techproducts -H 'Content-Type: application/json' -d '
-  {
-    "reload": {}
-  }
-'
+curl -X POST http://localhost:8983/api/cores/techproducts/reload -H 'Content-Type: application/json'
```
gerlowskija, line 321: [0] I'd have to check this to be certain, but we should be able to remove the `Content-type` header from this example, since this API ignores the request body.

gerlowskija (changes requested):
Looks great overall - I've suggested a few (mostly small) changes.

The biggest one being a revamp of how the tests are done.  Let me know if anything's unclear or if I can help at all.

Thanks again for taking this on!

PR #1777, opened 2023-07-12
SOLR-16856: Set autoSoftCommit for built-in configSets

HoustonPutman:
The default is 3 seconds.

https://issues.apache.org/jira/browse/SOLR-16856


tflobbe (approved):
LGTM!

PR #1780, opened 2023-07-12
SOLR-16889 Rate Limiter should stop processing on 429

stillalex:
https://issues.apache.org/jira/browse/SOLR-16889

# Description

PR for short cutting the request if Rate Limiter decides against processing. also takes care of cleaning up the request.

# Solution

# Tests

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
+++ b/solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
@@ -297,10 +297,6 @@ public abstract class ServletUtils {
     return (Tracer) req.getAttribute(SolrDispatchFilter.ATTR_TRACING_TRACER);
   }
 
-  private static RateLimitManager getRateLimitManager(HttpServletRequest req) {
-    return (RateLimitManager) req.getAttribute(SolrDispatchFilter.ATTR_RATELIMIT_MANAGER);
-  }
-
   protected static Span buildSpan(Tracer tracer, HttpServletRequest request) {
     if (tracer instanceof NoopTracer) {
       return NoopSpan.INSTANCE;
--- a/solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
+++ b/solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
@@ -197,39 +197,30 @@ public abstract class ServletUtils {
    * @param request The request to limit
    * @param response The associated response
    * @param limitedExecution code that will be traced
-   * @param trace a boolean that turns tracing on or off
    */
   static void rateLimitRequest(
       RateLimitManager rateLimitManager,
       HttpServletRequest request,
       HttpServletResponse response,
-      Runnable limitedExecution,
-      boolean trace)
+      Runnable limitedExecution)
       throws ServletException, IOException {
     boolean accepted = false;
-    boolean needsCleanup = true;
     try {
-      try {
-        accepted = rateLimitManager.handleRequest(request);
-      } catch (InterruptedException e) {
-        Thread.currentThread().interrupt();
-        throw new SolrException(ErrorCode.SERVER_ERROR, e.getMessage());
-      }
-
+      accepted = rateLimitManager.handleRequest(request);
       if (!accepted) {
         response.sendError(ErrorCode.TOO_MANY_REQUESTS.code, RateLimitManager.ERROR_MESSAGE);
         return;
       }
-      needsCleanup = false;
       // todo: this shouldn't be required, tracing and rate limiting should be independently
       // composable
-      traceHttpRequestExecution2(request, response, limitedExecution, trace);
+      traceHttpRequestExecution2(request, response, limitedExecution);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new SolrException(ErrorCode.SERVER_ERROR, e.getMessage());
     } finally {
-      if (needsCleanup) {
-        consumeInputFully(request, response);
-        SolrRequestInfo.reset();
-        SolrRequestParsers.cleanupMultipartFiles(request);
-      }
+      consumeInputFully(request, response);
+      SolrRequestInfo.reset();
+      SolrRequestParsers.cleanupMultipartFiles(request);
       if (accepted) {
         rateLimitManager.decrementActiveRequests(request);
       }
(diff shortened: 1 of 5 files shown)


solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
```diff
@@ -216,16 +217,19 @@ static void rateLimitRequest(
...
       }
+      needsCleanup = false;
       // todo: this shouldn't be required, tracing and rate limiting should be independently
       // composable
       traceHttpRequestExecution2(request, response, limitedExecution, trace);
     } finally {
+      if (needsCleanup) {
+        consumeInputFully(request, response);
+        SolrRequestInfo.reset();
+        SolrRequestParsers.cleanupMultipartFiles(request);
```
dsmiley, line 231: This looks wrong.... we won't reset SolrRequestInfo sometimes?  Why not?  Not doing so is dangerous; messing up access to SolrRequestInfo for future requests on the same thread.

And structurally this doesn't look right.  We shouldn't be adding additional callers to some of these methods.  Control flow should be well balanced logical and avoid calling methods from multiple spots under different conditions.  It's not that it's wrong necessarily but is hard to understand and reason about.

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
```diff
@@ -209,8 +208,8 @@ public void doFilter(
     Tracer t = getCores() == null ? GlobalTracer.get() : getCores().getTracer();
     request.setAttribute(Tracer.class.getName(), t);
     RateLimitManager rateLimitManager = coreService.getService().getRateLimitManager();
-    request.setAttribute(RateLimitManager.class.getName(), rateLimitManager);
```
dsmiley, line 212: It's not clear to me if the RateLimitManager was on the request merely out of convenience for the existing logic or if it was intentionally used for miscellaneous unforeseen use-cases (and you are now removing it).  Shrug.

dsmiley (changes requested):

solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
stillalex: thank you for taking a look. I agree with the control flow idea 100%. it is very difficult to work around code that is so poorly set in place.
I introduced this flag because ` traceHttpRequestExecution2` does resource cleanup "sometimes", take a look (I actually think the 'else' block in there is dead code, but it's not relevant to this PR). 
ideally this entire block "rate limiter method calling into trace method calling into mystery method" would be wrapped into a try/finally with correct cleanup but it's not so I added this flag which was "best effort cleanup" in case the trace method does not get called.
Also, is there an actual test that verifies correct resource cleanup? I am happy to extend it with this change.

If you have stronger opinions (proposing a proper refactoring) and are good for a review, I am happy to dive deeper, but so far there are not too many reviewers available.

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
stillalex: luckily we have very good tooling support to show where a class is used inside a project. I can 100% tell that this is not used anywhere else. Happy to be shown different. I think this was just a partial/unfinished plan to do something and it ended up in an awkward place where it is set as an attribute instead of simply passing the ref.
I took the liberty to remove some of the tech debt by cleaning up some code I saw was not clear, if this is putting too much burden on the review I will be happy to revert, let me know.

solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
dsmiley: I'm happy to review around this code (here I am) so restructuring I think is necessary first.  @gus-asf might -- he brought about I'd current structure.

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
dsmiley: Of course we can use Java tooling to look within Solr to see but this may be here for external / plugins.  In general we put objects on the request for plugins.  I'm not familiar with the RateLimitManager to consider if there would be a point for any plugin to want to get it from the request.  My intuition agrees with you.

stillalex: hmm. good point. I was quick to say it's not used in Solr, but you are of course right plugins are something I forgot about. still, the rate limiter doesn't strike me as being a good fit for a request attribute. I am ok with leaving this in if you think there is a risk of regression.

solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
stillalex: thank you for offering. I will try to clean this up and at the same time reduce collateral changes as much as possible.

stillalex:
Pushing a deeper refactoring. the cleanup will happen only once on a finally block inside the main entry method. I think the null tracer codepath is dead code, I don't see anywhere it could be null unless there is a bad test setup (so test must be fixed instead). in some sense this is needing a major split to servlet filters (rate limiting, tracing), but until that happens at least the code is clearer wrt. resource cleanup.

solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
```diff
@@ -197,35 +197,30 @@ static void configExcludes(PathExcluder excluder, String patternConfig) {
...
       }
       // todo: this shouldn't be required, tracing and rate limiting should be independently
       // composable
-      traceHttpRequestExecution2(request, response, limitedExecution, trace);
+      traceHttpRequestExecution2(request, response, limitedExecution);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new SolrException(ErrorCode.SERVER_ERROR, e.getMessage());
     } finally {
+      consumeInputFully(request, response);
```
dsmiley, line 221: Lets move these three lines somewhere else as this doesn't have to do with reate limiting (this method).  Perhaps generic doFilter.

dsmiley (approved):
Major improvement already!

stillalex:
as suggested, I moved the resource cleanup to `doFilter`. I also removed another layer of try/catch blocks.
please review carefully, I did my best to not introduce any bugs but I might have missed something in the process. thank you @dsmiley!

stillalex:
@dsmiley please let me know your thoughts on the latest version

PR #1790, opened 2023-07-18
Update v2-api.adoc: remove mention of /api/c

mkhludnev:
I haven't found this alias anywhere in code. When I checked 9.3 RC1 /api/c isn't handled either. Perhaps I miss something.

### DIFF
--- a/solr/solr-ref-guide/modules/configuration-guide/pages/v2-api.adoc
+++ b/solr/solr-ref-guide/modules/configuration-guide/pages/v2-api.adoc
@@ -39,7 +39,7 @@ Following are some v2 API URL paths and path prefixes, along with some of the op
 [width="100%",options="header",]
 |===
 |Path prefix |Some Supported Operations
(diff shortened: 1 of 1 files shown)


epugh:
I suspect the `/api/c/` pattern was something our old home grown code supported that we don't have in the new V2 approach...

epugh (approved):

gerlowskija:
> I suspect the /api/c/ pattern was something our old home grown code supported that we don't have in the new V2 approach...

Yep, pretty much.  Technically, the new v2 framework supports it, but we are moving away from it for consistency/intuitiveness reasons.  "Saving" a few characters in a URL path isn't really worth the downsides of making the path less readable to users, and obscuring whether a given API call is a core vs a collection operation.  So as we've been making APIs more REST-ful some of them have dropped the abbreviated "/c/" syntax.

This PR LGTM.

IMO the page could use some even more drastic changes; this is a good reminder for me to get to those for 9.4; thanks @mkhludnev !

gerlowskija (approved):

PR #1792, opened 2023-07-18
add support of OAuth 2.0/OIDC 'code with PKCE' flow (back-end)

laminelam:
https://issues.apache.org/jira/browse/SOLR-16897

This is the "back-end" part of '[code with PKCE](https://github.com/apache/solr/pull/1791)' contribution. 

The back-end code is mainly for configuration. This is where the different options are configured.

This PR adds _tokenEndpoint_ and _authorizationFlow_ options.

# Checklist

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -107,6 +107,8 @@ Improvements
 
 * SOLR-16927: Allow SolrClientCache clients to use Jetty HTTP2 clients (Alex Deparvu, David Smiley)
 
+* SOLR-16896, SOLR-16897: Add support of OAuth 2.0/OIDC 'code with PKCE' flow (Lamine Idjeraoui, janhoy, Kevin Risden)
+
 * SOLR-16879: Limit the number of concurrent expensive core admin operations by running them in a
   dedicated thread pool. Backup, Restore and Split are expensive operations.
   (Pierre Salagnac, David Smiley)
--- a/solr/core/src/java/org/apache/solr/servlet/LoadAdminUiServlet.java
+++ b/solr/core/src/java/org/apache/solr/servlet/LoadAdminUiServlet.java
@@ -16,11 +16,15 @@
  */
 package org.apache.solr.servlet;
 
+import com.google.common.net.HttpHeaders;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import org.apache.commons.io.output.CloseShieldOutputStream;
@@ -37,6 +41,8 @@ public final class LoadAdminUiServlet extends BaseSolrServlet {
   // check system properties for whether or not admin UI is disabled, default is false
   private static final boolean disabled =
       Boolean.parseBoolean(System.getProperty("disableAdminUI", "false"));
+  // list of comma separated URLs to inject into the CSP connect-src directive
+  public static final String SYSPROP_CSP_CONNECT_SRC_URLS = "solr.ui.headers.csp.connect-src.urls";
 
   @Override
   public void doGet(HttpServletRequest _request, HttpServletResponse _response) throws IOException {
@@ -60,20 +66,47 @@ public final class LoadAdminUiServlet extends BaseSolrServlet {
       if (in != null && cores != null) {
         response.setCharacterEncoding("UTF-8");
         response.setContentType("text/html");
+        String connectSrc = generateCspConnectSrc();
+        response.setHeader(
+            HttpHeaders.CONTENT_SECURITY_POLICY,
+            "default-src 'none'; base-uri 'none'; connect-src "
+                + connectSrc
+                + "; form-action 'self'; font-src 'self'; frame-ancestors 'none'; img-src 'self' data:; media-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self'; worker-src 'self';");
 
         // We have to close this to flush OutputStreamWriter buffer
         try (Writer out =
             new OutputStreamWriter(
                 CloseShieldOutputStream.wrap(response.getOutputStream()), StandardCharsets.UTF_8)) {
-          Package pack = SolrCore.class.getPackage();
           String html =
               new String(in.readAllBytes(), StandardCharsets.UTF_8)
-                  .replace("${version}", pack.getSpecificationVersion());
+                  .replace("${version}", getSolrCorePackageSpecVersion());
           out.write(html);
         }
       } else {
         response.sendError(404);
       }
     }
+  }
+
+  /**
+   * Retrieves the specification version of the SolrCore package.
+   *
+   * @return The specification version of the SolrCore class's package or Unknown if it's
+   *     unavailable.
+   */
+  private String getSolrCorePackageSpecVersion() {
+    Package pack = SolrCore.class.getPackage();
+    return pack.getSpecificationVersion() != null ? pack.getSpecificationVersion() : "Unknown";
+  }
+
+  /**
+   * Fetch the value of {@link #SYSPROP_CSP_CONNECT_SRC_URLS} system property, split by comma, and
+   * concatenate them into a space-separated string that can be used in CSP connect-src directive
+   */
+  private String generateCspConnectSrc() {
+    String cspURLs = System.getProperty(SYSPROP_CSP_CONNECT_SRC_URLS, "");
+    List<String> props = new ArrayList<>(Arrays.asList(cspURLs.split(",")));
+    props.add("'self'");
+    return String.join(" ", props);
   }
 }
--- a/solr/modules/jwt-auth/src/java/org/apache/solr/security/jwt/JWTAuthPlugin.java
+++ b/solr/modules/jwt-auth/src/java/org/apache/solr/security/jwt/JWTAuthPlugin.java
@@ -63,6 +63,7 @@ import org.apache.solr.security.AuthenticationPlugin;
 import org.apache.solr.security.ConfigEditablePlugin;
 import org.apache.solr.security.jwt.JWTAuthPlugin.JWTAuthenticationResponse.AuthCode;
 import org.apache.solr.security.jwt.api.ModifyJWTAuthPluginConfigAPI;
+import org.apache.solr.servlet.LoadAdminUiServlet;
 import org.apache.solr.util.CryptoKeys;
 import org.eclipse.jetty.client.api.Request;
 import org.jose4j.jwa.AlgorithmConstraints;
@@ -130,7 +131,9 @@ public class JWTAuthPlugin extends AuthenticationPlugin
           JWTIssuerConfig.PARAM_CLIENT_ID,
           JWTIssuerConfig.PARAM_WELL_KNOWN_URL,
           JWTIssuerConfig.PARAM_AUDIENCE,
-          JWTIssuerConfig.PARAM_AUTHORIZATION_ENDPOINT);
+          JWTIssuerConfig.PARAM_AUTHORIZATION_ENDPOINT,
+          JWTIssuerConfig.PARAM_TOKEN_ENDPOINT,
+          JWTIssuerConfig.PARAM_AUTHORIZATION_FLOW);
 
   private JwtConsumer jwtConsumer;
   private boolean requireExpirationTime;
@@ -280,8 +283,22 @@ public class JWTAuthPlugin extends AuthenticationPlugin
     }
 
     initConsumer();
+    registerTokenEndpointForCsp();
 
     lastInitTime = Instant.now();
+  }
+
+  /**
+   * Record Issuer token URL as a system property so it can be picked up and sent to Admin UI as CSP
+   */
+  protected void registerTokenEndpointForCsp() {
+    final String syspropName = LoadAdminUiServlet.SYSPROP_CSP_CONNECT_SRC_URLS;
+    String url = !issuerConfigs.isEmpty() ? getPrimaryIssuer().getTokenEndpoint() : null;
+    if (url != null) {
+      System.setProperty(syspropName, url);
+    } else {
+      System.clearProperty(syspropName);
+    }
   }
 
   /**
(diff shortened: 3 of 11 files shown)


risdenk (approved):

solr/modules/jwt-auth/src/java/org/apache/solr/security/jwt/JWTIssuerConfig.java
```diff
@@ -315,6 +327,24 @@ public JWTIssuerConfig setAuthorizationEndpoint(String authorizationEndpoint) {
...
+    this.tokenEndpoint = tokenEndpoint;
+    return this;
+  }
+
+  public String getAuthorizationFlow() {
+    return authorizationFlow;
+  }
+
+  public JWTIssuerConfig setAuthorizationFlow(String authorizationFlow) {
+    this.authorizationFlow = authorizationFlow;
```
janhoy, line 344: Should we not validate the flow value early and fail if not either of the two allowed values? I'm thinking we can also let this default to `implicit` here in the backend instead of in the frontend?

janhoy (commented):
Looks overall good.

Please do a stab on RefGuide docs. The ref guide should encourage the use of code flow and warn that implicit flow support is deprecated.

Perhaps we can also print a WARN log in backend if `implicit` is in use (either explicitly or implicitly) so that users upgrading to 9.4 without reading the release notes will at least get a warn log that they should switch.

solr/modules/jwt-auth/src/java/org/apache/solr/security/jwt/JWTAuthPlugin.java
```diff
@@ -847,9 +851,11 @@ protected String generateAuthDataHeader() {
     Map<String, Object> data = new HashMap<>();
     data.put(
         JWTIssuerConfig.PARAM_AUTHORIZATION_ENDPOINT, primaryIssuer.getAuthorizationEndpoint());
+    data.put(JWTIssuerConfig.PARAM_TOKEN_ENDPOINT, primaryIssuer.getTokenEndpoint());
```
janhoy, line 854: In the `X-Solr-AuthData` the keys should be "snake_case", so this should be `token_endpoint`, not `tokenEndpoint`. The frontend part will probably need to adapt to this as well.
```suggestion
    data.put("token_endpoint", primaryIssuer.getTokenEndpoint());
```

solr/modules/jwt-auth/src/java/org/apache/solr/security/jwt/JWTAuthPlugin.java
```diff
@@ -847,9 +851,11 @@ protected String generateAuthDataHeader() {
     Map<String, Object> data = new HashMap<>();
     data.put(
         JWTIssuerConfig.PARAM_AUTHORIZATION_ENDPOINT, primaryIssuer.getAuthorizationEndpoint());
```
janhoy, line 853:
```suggestion
        "authorization_endpoint", primaryIssuer.getAuthorizationEndpoint());
```
See comment in line below

janhoy: For this line I realize that it's been like this since 9.0, so let's not change that in this PR...

solr/modules/jwt-auth/src/java/org/apache/solr/security/jwt/JWTIssuerConfig.java
laminelam: I've added a validation test to _setAuthorizationFlow_. The reason I am not doing it earlier is:
- I am following the same logic/style than the other methods (like _setJwksUrl_)
- Adding it earlier would lead to too much verbose code for such basic test.

I have also set _implict_ as default. Will change the front code too.

solr/modules/jwt-auth/src/java/org/apache/solr/security/jwt/JWTAuthPlugin.java
laminelam: Okay, I am leaving as it is both _PARAM_AUTHORIZATION_ENDPOINT_ and _PARAM_TOKEN_ENDPOINT_  for now, is that what you meant?

laminelam:
> Looks overall good.
> 
> Please do a stab on RefGuide docs. The ref guide should encourage the use of code flow and warn that implicit flow support is deprecated.
> 
> Perhaps we can also print a WARN log in backend if `implicit` is in use (either explicitly or implicitly) so that users upgrading to 9.4 without reading the release notes will at least get a warn log that they should switch.

Thank you @janhoy for the review.
I am planning to add a new entry to the ref guide.
BTW, I have added a WARN message.

solr/modules/jwt-auth/src/java/org/apache/solr/security/jwt/JWTAuthPlugin.java
janhoy: Yea, let’s re-visit this later..

janhoy:
> BTW, I have added a WARN message.

Thanks, the validation and warn log looks good 👍

Will do a final review when refguide docs are in. Also remember CHANGES.txt.

janhoy (changes requested):
* Add refguide docs
* CHANGES entry
* Constant naming
* Backend part of CSP header

janhoy:
Eager to get this into 9.4. Let me know if we can help finalizing these two PRs

laminelam:
> Eager to get this into 9.4. Let me know if we can help finalizing these two PRs
@janhoy 
I should push a new commit sometime today. I have changed little bit your CSP patch, I just need to add a test class for LoadAdminUiServlet.

janhoy:
> > Eager to get this into 9.4. Let me know if we can help finalizing these two PRs
> 
> @janhoy I should push a new commit sometime today. I have changed little bit your CSP patch, I just need to add a test class for LoadAdminUiServlet.

Cool. In my patch i completely disabled CSP response for all other endpoints than the `AdminUIServlet`. I cannot see a reason to return that header for other endpoints. The only I can think of is if you have a solr package that exposes another UI on some path, it could be nice to protect that UI with CSP rules, so if it would be possible to have some fallback CSP header for all other requests perhaps?

solr/solr-ref-guide/modules/deployment-guide/pages/jwt-authentication-plugin.adoc
```diff
@@ -85,6 +85,8 @@ jwk                  ; As an alternative to `jwksUrl` you may provide a static J
 iss                  ; Unique issuer id as configured on the IdP. Incoming tokens must have a matching `iss` claim. Also used to resolve issuer when multiple issuers configured.      ; Auto configured if `wellKnownUrl` is provided
 aud                  ; Validates that the `aud` (audience) claim equals this string      ; Uses `clientId` if configured
 authorizationEndpoint; The URL for the Id Provider's authorization endpoint ; Auto configured if `wellKnownUrl` is provided
+tokenEndpoint; The URL for the Id Provider's token endpoint ; Auto configured if `wellKnownUrl` is provided
+authorizationFlow; Specifies the OAuth 2.0 flow to be used. Supported flows are 'implicit_flow' and 'code_pkce' (for authorization code with 'Proof Key for Code Exchange'). Note: 'implicit_flow' is deprecated and it is highly recommended to use 'code_pkce' instead. ; implicit_flow
```
janhoy, line 89: `implicit_flow` --> `implicit`?

solr/server/etc/jetty.xml
```diff
@@ -99,6 +100,7 @@
           </New>
         </Arg>
       </Call>
```
janhoy, line 102: Do we need a fallback CSP header for all API endpoints?

solr/solr-ref-guide/modules/deployment-guide/pages/jwt-authentication-plugin.adoc
laminelam: > `implicit_flow` --> `implicit`?

Good catch.
Fixed :-)

laminelam:
> > > Eager to get this into 9.4. Let me know if we can help finalizing these two PRs
> > 
> > 
> > @janhoy I should push a new commit sometime today. I have changed little bit your CSP patch, I just need to add a test class for LoadAdminUiServlet.
> 
> Cool. In my patch i completely disabled CSP response for all other endpoints than the `AdminUIServlet`. I cannot see a reason to return that header for other endpoints. The only I can think of is if you have a solr package that exposes another UI on some path, it could be nice to protect that UI with CSP rules, so if it would be possible to have some fallback CSP header for all other requests perhaps?

Yes I was thinking the same thing, we don't need this header for non browser targeted servlets (API endpoint, etc), so it will be just ignored by other endpoints. Now, I am not sure of a scenario where we would need a different UI servlet exposed on a different path. Let me think about it and get back to you hopefully by tomorrow.

laminelam:
> * Add refguide docs
> * CHANGES entry
> * Constant naming
> * Backend part of CSP header

Will the CHANGEs entry.

Here are the main changes to the patch:
- In line with the security best practice of granting minimal necessary permissions, adjusted the _registerTokenEndpointForCsp_() method to incorporate the precise URL for the /token endpoint rather than a more general host:port based URL.

- Thought providing the capability of passing the URLs as a comma-separated list is less confusing for the end-user/operator compared to using a prefix-based system property name. This approach also enables more concise implementation.

solr/server/etc/jetty.xml
janhoy: @rmuir You added the CSP stuff to `jetty.xml`. Do you see a reason to keep sending the header for any and all API requests to Solr that are not from a browser? I don't.

@chatman The only other UI I can think of being served by Solr backend are those that are part of a package installed by package manager. Those UIs are 100% stand alone, can we also let it be the UI author's responsibility to secure it with proper headers? I think so.

janhoy: If there's an easy way to provide a "fallback" if the header is not set, we could do that to avoid this discussion. I'm aware of Jetty's generic response rewrite plugins. We also have our `SolrDispatchFilter` but I find no single code point where one could manipulate response header before response is sent.

rmuir: > @rmuir You added the CSP stuff to `jetty.xml`. Do you see a reason to keep sending the header for any and all API requests to Solr that are not from a browser? I don't.

I set these headers across the board for all responses, pure API servers too. For something doing OAuth authorization code flow I would definitely set it.  I don't understand "not from a browser" as this flow usually involves one?

epugh:
We do have two packages "in the wild" that come with a UI, the YASA package https://github.com/yasa-org/yasa and Splainer, https://github.com/o19s/splainer/tree/main/solr-splainer-package.    Will these packages need to be updated?   

If we want to make this change, and that ripples down to packages, then we should do it now before more Packages are developed ;-).

We do have a test (commented out) that validates deploying Packages that might be useful: https://github.com/apache/solr/blob/main/solr/packaging/test/test_packages.bats#L69

solr/server/etc/jetty.xml
janhoy: Ok, turned out to be much simpler. Just leaving the `HeaderPatternRule` in `jetty.xml` worked fine. The `LoadAdminUiServlet` simply overwrite the header, and everything is just as it has been, no extra code. I pushed the fix, we can close this concern.

janhoy:
> We do have two packages "in the wild" that come with a UI

As stated anove, we never really needed to disable the old CSP header config in `jetty.xml`. Turns out the `LoadAdminUiServlet` simply replaces the header, which is nice.

solr/core/src/java/org/apache/solr/servlet/LoadAdminUiServlet.java
```diff
@@ -60,20 +66,49 @@ public void doGet(HttpServletRequest _request, HttpServletResponse _response) th
...
+
+  /**
+   * Retrieves the specification version of the SolrCore package.
+   *
+   * <p>This method is primarily introduced to facilitate unit testing
+   *
+   * @return The specification version of the SolrCore class's package or Unknown if it's
+   *     unavailable.
+   */
+  protected String getSolrCorePackageSpecVersion() {
```
janhoy, line 99: Precommit checks fail due to this. Change to private:
```suggestion
  private String getSolrCorePackageSpecVersion() {
```

laminelam: done

solr/CHANGES.txt
```diff
@@ -147,6 +149,8 @@ Bug Fixes
 * SOLR-16916: Use of the JSON Query DSL should ignore the defType parameter
   (Christina Chortaria, Max Kadel, Ryan Laddusaw, Jane Sandberg, David Smiley)
 
+* SOLR-16900: Add/fix wrong/missing "solr.home" to zkcli scripts  (Lamine Idjeraoui. Review by Eric Pugh & Jason Gerlowski)
```
janhoy, line 152: Unrelated edit?

solr/CHANGES.txt
```diff
@@ -107,6 +107,8 @@ Improvements
 
 * SOLR-16927: Allow SolrClientCache clients to use Jetty HTTP2 clients (Alex Deparvu, David Smiley)
 
+* SOLR-16897: Add support of OAuth 2.0/OIDC 'code with PKCE' flow - back-end  (Lamine Idjeraoui. Review and contribution by janhoy. Review by Kevin Risden)
```
janhoy, line 110: Simply list the contributors, no need to qualify what each one contributed.

Github will set author and add `Co-authored-by: John Doe <john@doe.com>` for code contributors in final merge message.

Also, no ned for two CHANGES entris even if you split it in two PRs with two commits. Suggest to add both JIRA numbers in same line:

```suggestion
* SOLR-16896, SOLR-16897: Add support of OAuth 2.0/OIDC 'code with PKCE' flow (Lamine Idjeraoui, janhoy, Kevin Risden)
```

solr/CHANGES.txt
laminelam: Yes not related to this PR.
But I forgot to add a change entry of another already merged PR, so I added it here to avoid a new PR just to add a change entry.
If you think it's not good idea never mind we can remove it.

janhoy: Yea, remove it from this PR, and do a separate commit (and PR) for the missing changes entry, with `SOLR-16900: ..` in the commit message - that will make it ping the JIRA issue to keep track of the commit history.

laminelam: done :)

janhoy (approved):
I fixed the merge conflict and CHANGES entry. Think this is good to go now. Great work!

janhoy:
I'm going to force-push a squashed git-history in order for full test suite to run. Will preserve you as commit author..

PR #1784, opened 2023-07-14
SOLR-16842: Make "version" just another Tool.

epugh:
https://issues.apache.org/jira/browse/SOLR-16842

# Description

Eliminate special case code in solrCLI by introducing VersionTool

# Solution
refactoring

# Tests
bats test and manual.   Pinging @Willdotwhite and @mkhludnev to verify the `solr.cmd` changes!

PR #1799, opened 2023-07-20
SOLR-16842: Remove reference to PostTool for branch_9x

patsonluk:
# Description

A merge from master to branch_9x includes reference to `PostTool` which is not available in 9x

Details in https://github.com/apache/solr/pull/1784#issuecomment-1642461497

# Solution

Removed the reference to `PostTool`. This branch is based on branch_9x


epugh:
Thank you @HoustonPutman and @patsonluk for this!

PR #1808, opened 2023-07-25
SOLR-16800: Move towards using the root of Solr as the -solrUrl in the CLI.

epugh:
https://issues.apache.org/jira/browse/SOLR-16800

# Description

In the future we'll have V2 API urls that are under /api, so having urls be /solr is a bit awkward.   Allow -solrUrl http://localhost:8983 to work, and warn on http://localhost:8983/solr.   

Some other cleanups as well.

# Solution

Use the SolrCLI.getSolrClient method through out the client Tool's.

# Tests

bats tests

# Checklist


solr/core/src/java/org/apache/solr/cli/SolrCLI.java
```diff
@@ -383,6 +383,11 @@ public static boolean exceptionIsAuthRelated(Exception exc) {
   }
 
   public static SolrClient getSolrClient(String solrUrl) {
+    // today we require all urls to end in /solr, however in the future we will need to support the
+    // /api url end point instead.
+    if (!solrUrl.endsWith(("/solr"))) {
```
gerlowskija, line 388: [0] Not sure whether this is intended or not, but if the provided URL was `http://localhost:8983/solr/` (not the ending '/') we'd still append`/solr`.  I wonder if there isn't some Java utility for adding path segments to a URL that can be a little smarter about some of the edge cases (handling adjacent forward slashes, etc.)

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
```diff
@@ -445,8 +450,31 @@ private static void printHelp() {
...
+   *
+   * @param solrUrl with /solr stripped off.
+   * @return the truncated if need solrUrl.
+   */
+  public static String resolveSolrUrl(String solrUrl) {
+    if (solrUrl != null) {
+      if (solrUrl.indexOf("/solr") > -1) { //
+        solrUrl = solrUrl.substring(0, solrUrl.indexOf("/solr"));
+        CLIO.out(
+            "The solrUrl parameter only needs to only point to the root of Solr ("
```
gerlowskija, line 466: [-0] I think this message would read more clearly if you dropped the second "only".

[0] Maybe add some indication in this message that Solr is able to auto-correct the URL for folks, so they don't see the warning and think the command has failed.  e.g.

> 'solrUrl' needn't include Solr's context-root (e.g. "/solr"); correcting from [old] to [new]

gerlowskija (approved):
LGTM, despite a few minor in-line suggestions.

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
epugh: So this kind of made me sad...  basically I was trying to write both code to handle legacy /solr type urls, but then, since everything TODAY requries a /solr url, then to inject it here.   Have we figured out how in the future to hit a V2 api that has an /api end point instead via a SolrClient?    My hope is once that is solved then we remove this logic.  
Having said that, if there is better code to use here, would love some of that!   I hope that before 10x gets released, this code will be eliminated, so I kind of just put this in "for now"....    I don't expect to backport this to 9x.

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
epugh: Love it... pushing up fix.

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
epugh: Ugh...  One test has "http://127.0.0.1:51902/uf_" as the URL.....

epugh: Oh wait...  why do tests now have solr urls that end in `/solr`???    Now I wonder if they also stand up Solr servers that don't use `/api` for the v2 api endpoint?

epugh:
This PR is held up on getting SOLR-16911, PR #1810 committed first....

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
epugh: okay, I've pulled in `main` that has the now hardcoded hostContext...

epugh:
@chatman I've got some optimizations to the Package tool if you want to look.  @gerlowskija I'd love another looksee now that hostContext has been sorted....

solr/core/src/test/org/apache/solr/cli/SolrCLITest.java
```diff
@@ -19,7 +19,14 @@
 import org.apache.solr.SolrTestCase;
 import org.junit.Test;
 
-public class SolrCliUptimeTest extends SolrTestCase {
+public class SolrCLITest extends SolrTestCase {
+  @Test
+  public void testResolveSolrUrl() {
+    assertEquals(SolrCLI.resolveSolrUrl("http://localhost:8983/solr"), "http://localhost:8983");
+    assertEquals(SolrCLI.resolveSolrUrl("http://localhost:8983"), "http://localhost:8983");
+    assertEquals(SolrCLI.resolveSolrUrl("http://localhost:8983/"), "http://localhost:8983");
```
gerlowskija, line 27: [0] Seeing these three cases makes me a little twitchy without `http://localhost:8983/solr/` to complete the set 😛

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
```diff
@@ -444,8 +453,34 @@ private static void printHelp() {
     print("Pass -help or -h after any COMMAND to see command-specific usage information,");
     print("such as:    ./solr start -help or ./solr stop -h");
   }
+
   /**
-   * Get the base URL of a live Solr instance from either the solrUrl command-line option from
+   * Strips off the end of solrUrl any /solr when a legacy solrUrl like http://localhost:8983/solr
+   * is used, and warns those users. In the future we'll have url's with /api as well.
+   *
+   * @param solrUrl with /solr stripped off.
```
gerlowskija, line 461: [Q] Are these Javadocs right?  It looks like the main point of this method is to do the "/solr"-stripping, but this Javadoc makes it sound like the `solrUrl` param is supposed to already be stripped?

gerlowskija (changes requested):
Few small nits, but looks mostly good.  Thanks for taking this on!

solr/core/src/test/org/apache/solr/cli/SolrCLITest.java
epugh: Happy to "de-twitch" you!

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
epugh: doh!

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
```diff
@@ -474,6 +509,8 @@ public static String resolveSolrUrl(CommandLine cli) throws Exception {
           solrUrl = ZkStateReader.from(cloudSolrClient).getBaseUrlForNodeName(firstLiveNode);
         }
       }
+    } else {
+      solrUrl = resolveSolrUrl(solrUrl);
```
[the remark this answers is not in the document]
epugh: Yeah, I redid it, took out some nesting to make it easier to read, and fix this...

gerlowskija (approved):
LGTM

PR #1810, opened 2023-07-28
SOLR-16911: Remove ability to change hostContext, keep it as /solr

epugh:
https://issues.apache.org/jira/browse/SOLR-16911

# Description

Discovered during work on on SolrCLI that we support different hostContext than `/solr`.   Remove that.

# Solution

# Tests

# Checklist


epugh:
Locally at least all the bats tests pass except for `test_placement_plugin.bats` who may actually be a bit flaky...

solr/core/src/test/org/apache/solr/cloud/LeaderElectionTest.java
```diff
@@ -258,8 +258,8 @@ public void testCancelElection() throws Exception {
     Thread.sleep(1000);
 
     String urlScheme = zkStateReader.getClusterProperty(URL_SCHEME, "http");
-    String url1 = Utils.getBaseUrlForNodeName("127.0.0.1:80_solr/1", urlScheme) + "/";
-    String url2 = Utils.getBaseUrlForNodeName("127.0.0.1:80_solr/2", urlScheme) + "/";
+    String url1 = Utils.getBaseUrlForNodeName("127.0.0.1:80_solr/1", urlScheme) + "/" + "1" + '/';
```
epugh, line 261: I have some concerns about how this test works....   I kind of worked around it, however the naming of the urls for leader election seemed "odd".

solr/solrj/src/java/org/apache/solr/common/util/Utils.java
```diff
@@ -712,7 +712,7 @@ public static String getBaseUrlForNodeName(
     }
     final String hostAndPort = nodeName.substring(0, _offset);
     final String path = URLDecoder.decode(nodeName.substring(1 + _offset), UTF_8);
```
dsmiley, line 714: "path" isn't needed now?  Or just leave this code as it was?

solr/test-framework/src/java/org/apache/solr/BaseDistributedSearchTestCase.java
```diff
@@ -189,40 +134,19 @@ public static void clearSolrDisableShardsWhitelist() throws Exception {
   }
 
   private static String getHostContextSuitableForServletContext() {
-    String ctx = System.getProperty("hostContext", "/solr");
-    if (ctx == null || ctx.isEmpty()) ctx = "/solr";
-    if (ctx.endsWith("/")) ctx = ctx.substring(0, ctx.length() - 1);
-    if (!ctx.startsWith("/")) ctx = "/" + ctx;
+    String ctx = "/solr";
+
     return ctx;
```
dsmiley, line 139:
```suggestion
    return "/solr";
```

Or just inline this thing!  It's just a constant string after all.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -2366,35 +2358,11 @@ public LeaderElector getOverseerElector() {
...
-    String out = in;
-    if (out.startsWith("/")) {
-      out = out.substring(1);
-    }
-    if (out.endsWith("/")) {
-      out = out.substring(0, out.length() - 1);
-    }
-    return out;
+  static String generateNodeName(final String hostName, final String hostPort) {
+    return hostName + ':' + hostPort + '_' + "solr";
```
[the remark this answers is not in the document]
HoustonPutman: Yeah if this happens it should be a separate jira and definitely only be Solr 10x. I think leaving it would be fine, but a discussion for a different PR anyways

solr/solrj/src/java/org/apache/solr/common/util/Utils.java
epugh: Good catch.  So, at one time we needed the path variable to decide if you had `/solr/mysubsolr` etc..  but now, we have just `/solr` so this can be eliminated actually!

solr/core/src/test/org/apache/solr/cloud/LeaderElectionTest.java
epugh: So, I actually am worried this entire idea may fail..  At least the way I read this test, when we do leaderelection, we name our nodes `{some_ip}/solr/1` and `{same_ip}/solr/2` and `{same_ip}/solr/3` and then something happens.   However, I don't know if that is ACTUALLY how it happens, or maybe just how this test was written, since I've never dug into leader election in any way....    I also was wondering if maybe we need some sort of `.bats` test of leader election to prove that this change didn't break leader election in the real world..  I mean, technically this LeaderElectionTest.java passes...  but...

henrik242 (approved):
lgtm

solr/core/src/test/org/apache/solr/cloud/LeaderElectionTest.java
dsmiley: @psalagnac WDYT of the changes to this test?

epugh: Thanks for tagging @psalagnac...   I think getting this resolved is the last thing that needs doing...  I am wondering if I need to create some sort of `.bats` test that validates the happy path of leaderelection...?

psalagnac: Thanks for pining me on this.

This is related to other David's  [comment](https://github.com/apache/solr/pull/1810/files/49fe85375b74cd76d7bc08bcc0d32781632bc0d6#r1279827024) on the path removal in `Utils.getBaseUrlForNodeName`.

The change in the test does not make sense to me since the core name is here twice, but it is only once in the result URL. We just want to build the core url from the node base URL. I think it should be like this:

```
    String url1 = Utils.getBaseUrlForNodeName("127.0.0.1:80_solr", urlScheme) + "/1/";
```

solr/solrj/src/java/org/apache/solr/common/util/Utils.java
psalagnac: If we don't use path anymore, this line could be removed.

epugh:
@dsmiley  @psalagnac all the tests pass.  Thoughts on comittting this?

dsmiley (approved):

PR #1816, opened 2023-07-30
SOLR-16360: Atomic update on boolean fields doesn't reflect when value starts with "1", "t" or "T"

rahulgoswami:
https://issues.apache.org/jira/browse/SOLR-16360

# Description

Added a change in toNativeType(Object val) method of BoolField.java to check for the first character of val and return true if it is '1', 't' or 'T'

# Solution

Added a change in toNativeType(Object val) method of BoolField.java to check for the first character of val and return true if it is '1', 't' or 'T'

# Tests

# Checklist


dsmiley:
Can you add a test please?

rahulgoswami:
On it!

justinrsweeney:
Are we able to reuse the code from [StrUtils](https://github.com/apache/solr/blob/main/solr/solrj/src/java/org/apache/solr/common/util/StrUtils.java#L259C32-L259C32) for this instead of duplicating?

rahulgoswami:
Added tests and incorporated Justin's comment.

solr/core/src/test/org/apache/solr/schema/BooleanFieldTest.java
```diff
@@ -17,6 +17,7 @@
 
 package org.apache.solr.schema;
 
+import java.util.Map;
 import org.apache.solr.SolrTestCaseJ4;
 import org.junit.BeforeClass;
 import org.junit.Test;
```
justinrsweeney, line 23: I think you could change this test to use solrconfig-tlog.xml instead of solrconfig-basic.xml. That way you could remove the change in this PR to solrconfig-basic.xml and keep that one simple.

rahulgoswami: Ah didn't know there's a solrconfig-tlog.xml. Still finding my way through the test suite. Makes sense. Incorporated.

justinrsweeney (approved):
👍 Looks good!

rahulgoswami:
The check failure seems like some issue with Crave.io, is it ? I can't see any failure specifics. Had the tests pass when running locally.

solr/core/src/java/org/apache/solr/schema/BoolField.java
```diff
@@ -211,7 +212,7 @@ public List<IndexableField> createFields(SchemaField field, Object value) {
   @Override
   public Object toNativeType(Object val) {
     if (val instanceof CharSequence) {
-      return Boolean.valueOf(val.toString());
+      return Boolean.valueOf(StrUtils.parseBoolean(val.toString()));
```
dsmiley, line 215: I think you could easily broaden the parameter type of parseBoolean to be CharSequence; then you're not calling val.toString.

dsmiley (commented):
If we're going to use StrUtils.parseBoolean here (makes sense), shouldn't we change `toInternal` to use it too?

solr/core/src/java/org/apache/solr/schema/BoolField.java
rahulgoswami: Thanks for pointing out the change in toInternal(). Makes sense. 

About the parameter type change in StrUtils.parseBoolean() to CharSeqeunce, is there a performance gain to be achieved there? With the proposed change, I'd have to typecast val to CharSequence while calling parseBoolean() inside toNativeType(). Alterntively, calling val.toString() should not have any overhead either since val would already be a String and the String class toString() implementation returns "this" anyway, so there is no extra object creation. We can avoid a change to StrUtils unless I am missing something. Thoughts welcome.

dsmiley: if "val" is not a String, then there is an avoidable String conversion.  Any way, unless someone bothers to benchmark this (which would be a very dubious use of anyone's time), we don't know for sure, so feel free to ignore my proposal.

rahulgoswami: Although I couldn't find evidence that val could be any other implementation of CharSequence, I agree with you it's one of those things not worth splitting hairs on. And it is definitely worth it being 100% sure rather than 99.99% in the interest of everyone's time :) Incorporating the comment.  Thanks.

dsmiley:
Can you please add an entry to CHANGES.txt under 9.4 (the next release)

rahulgoswami:
Would it make sense for this to be considered for 8.11.3 release next month ?

dsmiley:
Yes (if you want to back-port), but that shouldn't affect CHANGES.txt for now because that's a special situation.

rahulgoswami:
This is done. Awaiting review. Thanks.

PR #1824, opened 2023-08-03
SOLR-16415: ZK DistributedMap should reject slash in IDs.

dsmiley:
* asyncId should not have forward slashes in it
* SizeLimitedDistributedMap should be resilient to directories (from before)

https://issues.apache.org/jira/browse/SOLR-16415

From what I see, this will work for both Overseer & Distributed Admin API modes.  The latter is processed by DistributedApiAsyncTracker and it uses DistributedMap and that's what this PR touches.


dsmiley:
@patsonluk you may be interested based on your previous modifications

dsmiley:
I'll merge this Monday if I hear no further feedback.  We've been using this in production and it's been peer reviewed.

PR #1826, opened 2023-08-04
SOLR-16920: Test full package lifecycle

epugh:
https://issues.apache.org/jira/browse/SOLR-16920

# Description

Test the full lifecycle.

# Solution
Use the now will be supported into the future Packaging of Splainer in the test.   It HAS found a bug in the current PackageManager!

# Tests

the unit tests plus bats


epugh:
Okay, this is ready for Review!   @chatman I'd love a LGTM on the test and the bug fix ;-)

HoustonPutman (commented):
While I do understand the reasoning, I'm not a big fan of having an integration test rely on connection to the internet...

epugh:
> While I do understand the reasoning, I'm not a big fan of having an integration test rely on connection to the internet...

I totally suspected this would come up....   I also don't love that we are referencing some random package created by some random person too!   While the "right" thing is to maybe have some sort of test package that exercises all the aspects of package management, and that is maybe part of Apache Solr as well, I don't really have the energy/interest to go down that path.   Are you a 0 or a -1 on this PR?   I.e, can we live with it for now or do you think to merge we need to deal with it?

One more idea, what if we comment out the test, so at least we have it someone wants to run it, but it isn't in the regular path?

HoustonPutman:
> One more idea, what if we comment out the test, so at least we have it someone wants to run it, but it isn't in the regular path?

I'm +1 to this with a comment stating why the test is not enabled by default

PR #1827, opened 2023-08-04
SOLR-16916: JSON boolean queries when solrconfig defType is set to edismax

sandbergja:
https://issues.apache.org/jira/browse/SOLR-16916

# Description

As discussed on [this solr-users thread](https://lists.apache.org/thread/3t25858oh6wcf74j868nk8l9801v21q4), if your requestHandler has a default defType of "edismax", you are unable to send boolean queries using the JSON Query DSL syntax.  This PR resolves the issue.

# Solution

We took the approach recommended by @dsmiley [in the jira ticket](https://issues.apache.org/jira/browse/SOLR-16916) (thanks, @dsmiley !)

# Tests

We created a new JUnit test based on the existing JSON Query tests, but with a different solrconfig.xml.  We confirmed that without our fix, the test fails, and with the fix, the test passes.

# Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,6 @@ __pycache__
 # Ignore gradle wrapper
 gradle/wrapper/gradle-wrapper.jar
 .gradletasknamecache
+
+.vscode/
+.tool-versions
--- a/solr/core/src/test-files/solr/collection1/conf/solrconfig-tlog-edismax.xml
+++ b/solr/core/src/test-files/solr/collection1/conf/solrconfig-tlog-edismax.xml
@@ -1,0 +1,143 @@
+<?xml version="1.0" ?>
+
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one or more
+ contributor license agreements.  See the NOTICE file distributed with
+ this work for additional information regarding copyright ownership.
+ The ASF licenses this file to You under the Apache License, Version 2.0
+ (the "License"); you may not use this file except in compliance with
+ the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
+<config>
+
+  <luceneMatchVersion>${tests.luceneMatchVersion:LATEST}</luceneMatchVersion>
+
+  <directoryFactory name="DirectoryFactory" class="${solr.directoryFactory:solr.RAMDirectoryFactory}">
+    <!-- used to keep RAM reqs down for HdfsDirectoryFactory -->
+    <bool name="solr.hdfs.blockcache.enabled">${solr.hdfs.blockcache.enabled:true}</bool>
+    <int name="solr.hdfs.blockcache.blocksperbank">${solr.hdfs.blockcache.blocksperbank:1024}</int>
+    <str name="solr.hdfs.home">${solr.hdfs.home:}</str>
+    <str name="solr.hdfs.confdir">${solr.hdfs.confdir:}</str>
+    <str name="solr.hdfs.blockcache.global">${solr.hdfs.blockcache.global:true}</str>
+  </directoryFactory>
+  <schemaFactory class="ClassicIndexSchemaFactory"/>
+  <requestDispatcher>
+    <requestParsers />
+  </requestDispatcher>
+
+  <dataDir>${solr.data.dir:}</dataDir>
+
+  <xi:include href="solrconfig.snippet.randomindexconfig.xml" xmlns:xi="http://www.w3.org/2001/XInclude"/>
+
+  <!-- an update processor the explicitly excludes distrib to test
+       clean errors when people attempt atomic updates w/o it
+  -->
+  <updateRequestProcessorChain name="nodistrib" >
+   <processor class="solr.NoOpDistributingUpdateProcessorFactory" />
+   <processor class="solr.RunUpdateProcessorFactory" />
+ </updateRequestProcessorChain>
+
+  <requestHandler name="/select" class="solr.SearchHandler">
+    <lst name="defaults">
+      <str name="defType">edismax</str>
+      <str name="qf">id</str>
+    </lst>
+  </requestHandler>
+
+
+  <updateHandler class="solr.DirectUpdateHandler2">
+    <!-- autocommit pending docs if certain criteria are met -->
+    <autoCommit>
+      <maxSize>${solr.autoCommit.maxSize:}</maxSize>
+    </autoCommit>
+    <updateLog class="${solr.tests.ulog:solr.UpdateLog}">
+      <str name="dir">${solr.ulog.dir:}</str>
+      <str name="maxNumLogsToKeep">${solr.ulog.maxNumLogsToKeep:10}</str>
+      <str name="numRecordsToKeep">${solr.ulog.numRecordsToKeep:100}</str>
+      <int name="tlogDfsReplication">${solr.ulog.tlogDfsReplication:2}</int>
+    </updateLog>
+
+
+    <autoCommit>
+      <maxTime>${solr.autoCommit.maxTime:-1}</maxTime>
+      <openSearcher>false</openSearcher>
+    </autoCommit>
+
+    <autoSoftCommit>
+      <maxTime>${solr.autoSoftCommit.maxTime:-1}</maxTime>
+    </autoSoftCommit>
+  </updateHandler>
+
+  <updateRequestProcessorChain name="distrib-dup-test-chain-explicit">
+    <!-- explicit test using processors before and after distrib -->
+    <processor class="solr.RegexReplaceProcessorFactory">
+      <str name="fieldName">regex_dup_A_s</str>
+      <str name="pattern">x</str>
+      <str name="replacement">x_x</str>
+    </processor>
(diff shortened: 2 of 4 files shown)


dsmiley (commented):
It's apparent you have copy-pasted an existing test and existing configuration in all its details of legacy things, including suppressing codecs that don't even exist anymore.  The testing situation in Solr is sad; not your fault.  For every "good test", there are more than ten bad ones IMO.  I think it's too much to ask of non-committers to figure out this mess, so I won't ask.  There needs to be an easy way to trivially augment a baseline config from within a test source file, thus not polluting our test configs with yet another test config with only one line that is different from some other config.  The current situation is a tech debt nightmare. 

Looking back through my attempts at improving this, I did a little experimentation and came up with the following snippet to exemplify how to create a core with a baseline configuration and then augment it:

```
initCore("solrconfig-basic.xml", "schema-basic.xml");

try (var req = (SolrQueryRequestBase) req(CommonParams.QT, "/config")) {
      req.setContentStreams(
          List.of(
              new ContentStreamBase.StringStream(
                  "{ 'set-property':{"
                      + "'query.filterCache.size':100, "
                      + "}}")));
      assertQ(req);
    }
```
In this example I merely show how to initialize the filter cache (which your test doesn't need) but you could do things like register a request handler with edismax.  See https://solr.apache.org/guide/solr/latest/configuration-guide/config-api.html

.gitignore
```diff
@@ -43,3 +43,6 @@ __pycache__
 # Ignore gradle wrapper
 gradle/wrapper/gradle-wrapper.jar
 .gradletasknamecache
+
+.vscode/
+.tool-versions
```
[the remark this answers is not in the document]
christinach: @dsmiley thank you for reviewing. In our application dev environment we use [asdf](https://asdf-vm.com/) to manage all [tool-versions](https://asdf-vm.com/manage/configuration.html#tool-versions) definitions. Will add a comment.

dsmiley:
I can anticipate a problem with my suggestion; may or may not be an issue.  Sometimes tests use Solr home's and/or configSets _in-place_ (end up being in a build dir) in directories that we consider read-only.  Tests that manipulate configuration need to use a temp directory.  Gradle executed tests enforce the read-only rule but using your IDE will probably not enforce it.  Other than copying files to a temp dir, I don't have a simpler recommendation.  There's no hurry on this; I can try to work on it being easier.  Ultimately I want this to be the norm for tests.

dsmiley (commented):
Here's an example of a partly working source file that doesn't involve adding new test configs:
```

public class TestJsonRequest2 extends SolrTestCaseJ4 {

  @ClassRule
  public static final SolrClientTestRule solrClientTestRule = new EmbeddedSolrServerTestRule();

  public void test() throws Exception {

    final var configSet = LuceneTestCase.createTempDir();
    copyMinConf(configSet.toFile());

    solrClientTestRule.startSolr();
    solrClientTestRule
        .newCollection()
        .withConfigSet(configSet.toString())
        .create();

    final var client = solrClientTestRule.getSolrClient();

    client.request(new ConfigRequest("{"
        + "  'add-requesthandler':{"
        + "    'name':'/query',"
        + "    'class':'solr.SearchHandler',"
        + "    'defaults' : {'defType':'edismax'}"
        + "  }"
        + "}"));

    addDocs(client);

    doQuery(client);
  }
  
  private static void addDocs(SolrClient client) throws Exception {
    //client.deleteByQuery("*:*");
    client.add(sdoc("id", "1", "cat_s", "A", "where_s", "NY"));
    client.add(sdoc("id", "2", "cat_s", "B", "where_s", "NJ"));
    client.add(sdoc("id", "3"));
    client.commit();
    client.add(sdoc("id", "4", "cat_s", "A", "where_s", "NJ"));
    client.add(sdoc("id", "5", "cat_s", "B", "where_s", "NJ"));
    client.commit();
    client.add(sdoc("id", "6", "cat_s", "B", "where_s", "NY"));
    client.commit();
  }
  
  private static void doQuery(SolrClient client) throws Exception {
    final var jsonQuery =
        "{\"query\":{\"bool\":{\"should\":[{\"lucene\":{\"query\":\"id:1\"}}, \"id:2\"]}}, \"params\":{\"debug\":\"true\"}}";
    SolrTestCaseHS.assertJQ(client, params("json", jsonQuery, "qt", "/query"), "response/numFound==2");
  }

}
```

The schema probably doesn't match up with the expectations of the test.  But this test could be simplified further so as to not need a custom schema in order to verify the intended change.  With `edismax`, the parser will complain if "df" or "qf" isn't provided.  So merely testing a trivial query is adequate; no need for a bunch of docs to reason about.

solr/core/src/test/org/apache/solr/search/json/TestJsonRequestWithEdismaxDefType.java
```diff
@@ -0,0 +1,110 @@
...
+  }
+  
+  private static void doQuery(SolrClient client) throws Exception {
+    final var jsonQuery =
+        "{\"query\":{\"bool\":{\"should\":[{\"lucene\":{\"query\":\"id:1\"}}, \"id:2\"]}}}";
+    SolrTestCaseHS.assertJQ(client, params("json", jsonQuery, "qt", "/query"), "response/numFound==2");
+  }
+
+  // write an elevation config file to boost some docs
+  private void writeElevationConfigFile(File file, String query, String... ids) throws Exception {
```
dsmiley, line 94: Nothing calls this.  Why is it here?

solr/core/src/test/org/apache/solr/search/json/TestJsonRequestWithEdismaxDefType.java
```diff
@@ -0,0 +1,110 @@
...
+import org.apache.solr.util.ExternalPaths;
+import org.apache.solr.util.EmbeddedSolrServerTestRule;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import java.nio.file.Path;
+
+
+public class TestJsonRequestWithEdismaxDefType extends SolrTestCaseHS {
```
dsmiley, line 37: Can you please not subclass SolrTestCaseHS?  Try SolrTestCase

solr/core/src/test/org/apache/solr/TestGroupingSearch.java
```diff
@@ -360,7 +360,7 @@ public void testGroupingSimpleFormatArrayIndexOutOfBoundsException() throws Exce
 
     assertJQ(
         req("q", "*:*", "start", "1", "group", "true", "group.field", "id", "group.main", "true"),
-        "/response=={'numFound':3,'start':1,'numFoundExact':true,'docs':[{'id':'2'},{'id':'3'}]}");
+        "/response=={'numFound':4,'start':1,'numFoundExact':true,'docs':[{'id':'2'},{'id':'3'}]}");
```
dsmiley, line 363: What happened in this PR requiring this?

solr/core/src/test-files/solr/collection1/conf/solrconfig-tlog-edismax.xml
dsmiley, line 1: This file can now be removed

dsmiley (changes requested):
I still see the

.gitignore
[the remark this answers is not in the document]
dsmiley: Instead of listing the dot-files of all possible tools, maybe users of such tools such put them in their own ~/.gitignore ?

dsmiley (commented):
I was about to push a change but I think you need to enable a setting that allows committers to the apache/solr repo access to your PR branch on your fork

dsmiley:
Instructions:  https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

Somehow I think it can be set to automatic so you don't have to think about it henceforth, but I'm not sure.

About some changes I did:
* Don't use SolrTestCaseHS at all, even the assertJQ thing.  There is only one trivial assert so it's easy to have it assert the manual way.  Still... it's curious that the JSON path expression didn't work.  I dug into that but it's a distraction here.
* The configSet needs to be a copy because we modify it.  The ~3 lines for this are kinda ugly but it's for another time to improve it.
* Using a very minimal configSet.

There is still some tidying up for you to do based on my comments.  I also think the test data should be simplified because your test doesn't even use half of it.

sandbergja:
Thanks for all your help, @dsmiley !  The test is so much more readable now!  Should be ready for your re-review.

dsmiley:
I switched the test to subclass SolrTestCaseJ4 (which 99% of Solr tests ultimately extend from) as there were mysterious cross-test side-effects when I didn't.  I don't have time today to debug why that is the case.  The recent test run still shows one failure for `org.apache.solr.search.TestCoordinatorRole ` but it's a known flapper as indicated by http://fucit.org/solr-jenkins-reports/failure-report.html

I'll add a CHANGES.txt entry crediting you and merge fairly soon.  Proposed text:
* SOLR-16916: Use of the JSON Query DSL should ignore the defType parameter (Jane Sandberg, David Smiley)

sandbergja:
Thank you, @dsmiley !  Could you please change "(Jane Sandberg, David Smiley)" to "(Christina Chortaria, Max Kadel, Ryan Laddusaw, Jane Sandberg, David Smiley)"?  All four of us here worked on it.

dsmiley (approved):

noah-dbc:
Just a note to flag the comment on the jira issue that was left a couple of days ago. This seems to have broken some functionality in solr 9.4 (https://issues.apache.org/jira/browse/SOLR-16916)

PR #1829, opened 2023-08-06
SOLR-6853: Add slash support for Managed Resources (synonyms and stopwords)

xtenzQ:
https://issues.apache.org/jira/browse/SOLR-6853

# Description

When adding synonyms or stop words containing the slash `/` character through the REST API, it becomes impossible to later remove these words.

# Solution

Rather than only extracting the last one or two strings from the end of the URL, I parse it until the managed resource is located or there are no more slashes in the `resourceId` path.

# Tests

I added a couple of tests that ensure synonyms and stop words containing a slash inside them are removed.

# Checklist


solr/core/src/java/org/apache/solr/rest/RestManager.java
```diff
@@ -288,27 +288,31 @@ public void doInit(SolrQueryRequest solrRequest, SolrQueryResponse solrResponse)
...
-          String parentResourceId = resourceId.substring(0, lastSlashAt);
+        int lastSlashAt;
+        String parentResourceId;
+        String initialResourceId = resourceId;
+        // Check if we have a registered endpoint, going one level up each time...
+        do {
+          lastSlashAt = initialResourceId.lastIndexOf('/');
+          parentResourceId = resourceId.substring(0, lastSlashAt);
           log.info(
               "Resource not found for {}, looking for parent: {}", resourceId, parentResourceId);
```
janhoy, line 299: This could be DEBUG level

solr/core/src/java/org/apache/solr/rest/RestManager.java
```diff
@@ -288,27 +288,31 @@ public void doInit(SolrQueryRequest solrRequest, SolrQueryResponse solrResponse)
...
+            String errMsg =
+                String.format(
+                    Locale.ROOT,
+                    "%s does not support child resources!",
+                    managedResource.getResourceId());
+            throw new SolrException(ErrorCode.BAD_REQUEST, errMsg);
           }
+
+          childId = resourceId.substring(lastSlashAt + 1);
+          log.info("Found parent resource {} for child: {}", parentResourceId, childId);
```
janhoy, line 315: Put this as DEBUG level as well

janhoy (approved):
Looks ok to me, just some log level observation.
Have not spun it up or ran tests myself.

janhoy:
Also feel free to add a line to CHANGES.txt, giving yourself credit

janhoy:
@gerlowskija, @risdenk  added you as reviewers since you may know RestManager for context..

xtenzQ:
> Also feel free to add a line to CHANGES.txt, giving yourself credit

Is it ok if I added this ticket under bug section?

janhoy:
> > Also feel free to add a line to CHANGES.txt, giving yourself credit
> 
> Is it ok if I added this ticket under bug section?

Yes I believe that is appropriate

janhoy:
I believe there is no documentation needed since this is a bug fix?

@gerlowskija any comments before I merge?

janhoy (approved):

xtenzQ:
@janhoy I've begun to question whether my current solution is more of a workaround rather than an actual solution.

I've been contemplating recently how to create a universal solution for handling all special characters. I've come up with an idea to modify the deletion behavior. Instead of sending a synonym/stopword for deletion via a URL, it seems more effective to include it in the request body of the DELETE request.

What do you think?

xtenzQ:
@janhoy I've begun to question whether my current solution is more of a workaround rather than an actual resolution that eliminates the problem.

To be honest, there are certain aspects that I find problematic. The situation involves managed resources within the URI, leading us to decode the entire URI due to its path nature. This becomes an issue when special characters like +, %, :, or / are present in synonyms/stopwords, as they become integrated into the path. While I haven't encountered clients in my work using `+` or `%`, they have certainly used `/` for things like abbreviations such as `AC/DC`. While my current solution solves this problem, it admittedly appears inelegant from my perspective and looks like a workaround.

Here are a couple of potential solutions:

1. Decode only the portion of the URI that contains the synonym/stopword. However, this approach raises doubts, especially if the user submits a UTF-8 encoded request to the managed resource. Since I don't have much experience with REST, I'd appreciate input from the RestManager contributors on this matter.

2. Consider altering the way we handle the deletion of synonyms/stopwords. For instance, we could opt for using a PUT request instead of DELETE and embed the synonym/stopword within the request body.

epugh:
@xtenzQ I saw your comment about "workaround", and are you aware of the effort to roll out new "proper RESTful" api's for Solr?   These new V2 apis use the proper http verbs, so deleting a managed resource would be a POST, reading them a GET etc.   There is a spreadsheet of all the apis being worked on here: https://docs.google.com/spreadsheets/d/1HAoBBFPpSiT8mJmgNZKkZAPwfCfPvlc08m5jz3fQBpA/edit#gid=605017626

I notice, and tagging @gerlowskija  for confirmation, that the Managed Resources don't actually show up in the list of apis related to Schemas!!!   Maybe the right path forward is to leave the existing somewhat broken api's alone, and work on rolling out the new V2 style apis, which would avoid some of this encoding stuff...

janhoy:
Hi. First of all, I think it is better with a working solution than no solution. It would be nice with a bug fix for the current API, and then consider changing the API behavior in followup tickets.

I'd like @gerlowskija to weigh in on whether `DELETE foo/bar/my-resource-name` is prefered over `DELETE foo/bar -d '{"resource": "my-resource-name"}'`. I suspect that we want to favour the former, i.e. have the URL address the resource directly in the path.

Really, you can throw anything into a path component if it is properly encoded, `/foo/bar/%#?&` would ideally be passed by the user as `/foo/bar/%25%23%3F%26` and our backend would then decode the path component before using it.

I have not dived deep into this, but I think we should assume the user has properly encoded the URL component and then fix our backend to always decode??

janhoy:
The docs at https://solr.apache.org/guide/solr/latest/configuration-guide/managed-resources.html says
```bash
To delete a stop word, you would do:

curl -X DELETE "http://localhost:8983/solr/techproducts/schema/analysis/stopwords/english/foo"
```

Nowhere on the page does it mention URL-encoding the word. Let's add that to the docs in this PR?

However, according to the JIRA, it still does not work if the URL is properly encoded. So to get a generic solution for all special chars a more thorough review of the handling of DELETE request on the backend is needed.

> The situation involves managed resources within the URI, leading us to decode the entire URI due to its path nature.

Why would we decode the entire URI? If we know that user-data is encoded, we can do `urlPath.split('/')` and grab the last component as the word, then decode it to find the synonym.

xtenzQ:
> @xtenzQ I saw your comment about "workaround", and are you aware of the effort to roll out new "proper RESTful" api's for Solr? These new V2 apis use the proper http verbs, so deleting a managed resource would be a POST, reading them a GET etc. There is a spreadsheet of all the apis being worked on here: https://docs.google.com/spreadsheets/d/1HAoBBFPpSiT8mJmgNZKkZAPwfCfPvlc08m5jz3fQBpA/edit#gid=605017626
> 
> I notice, and tagging @gerlowskija for confirmation, that the Managed Resources don't actually show up in the list of apis related to Schemas!!! Maybe the right path forward is to leave the existing somewhat broken api's alone, and work on rolling out the new V2 style apis, which would avoid some of this encoding stuff...

Yeah I can imagine how much work it is. I sincerely apologize if my previous message unintentionally caused any offense. It was not my intention.

epugh:
> > @xtenzQ I saw your comment about "workaround", and are you aware of the effort to roll out new "proper RESTful" api's for Solr? These new V2 apis use the proper http verbs, so deleting a managed resource would be a POST, reading them a GET etc. There is a spreadsheet of all the apis being worked on here: https://docs.google.com/spreadsheets/d/1HAoBBFPpSiT8mJmgNZKkZAPwfCfPvlc08m5jz3fQBpA/edit#gid=605017626
> > I notice, and tagging @gerlowskija for confirmation, that the Managed Resources don't actually show up in the list of apis related to Schemas!!! Maybe the right path forward is to leave the existing somewhat broken api's alone, and work on rolling out the new V2 style apis, which would avoid some of this encoding stuff...
> 
> Yeah I can imagine how much work it is. I sincerely apologize if my previous message unintentionally caused any offense. It was not my intention.

Oh, I wasn't offended in any way! I love when folks find bugs in Solr and work on fixing them, like this one!.    The V2 effort is huge, and I could totally imagine that when we were putting the list of apis that need migration together, that this one was missed!   It's a community effort ;-)

janhoy:
> I could totally imagine that when we were putting the list of apis that need migration together, that this one was missed!

@epugh Well, this managed resource API is the only legacy V1 API that was ever close to RESTful in the first place (it used to be implemented with RESTlet), so its v2 version may not be that different in the end :) But let's keep this bugfix PR focused on fixing the bug for v1.

IIRC there have also been suggestions to deprecate and remove managed resources, assuming they are less frequently used and that the more common way of managing these is to push a new copy of the configset. I don't know how much used these are - the fact that this bug is 8 years old hints that the feature is not that popular?

xtenzQ:
> Why would we decode the entire URI? If we know that user-data is encoded, we can do urlPath.split('/') and grab the last component as the word, then decode it to find the synonym.

Yeah, I've been thinking on this too. I have to give it a try probably.

> IIRC there have also been suggestions to deprecate and remove managed resources, assuming they are less frequently used and that the more common way of managing these is to push a new copy of the configset. I don't know how much used these are - the fact that this bug is 8 years old hints that the feature is not that popular?

Well, having an API for this is quite nice! It might not be very popular, but overall, it works great. The issue with special characters isn't significant; we discovered it after a couple of years of consistent operation.

epugh:
Agreed that we should focus on this bugfix ;-), thanks @janhoy for pointing that out.   As far as what to do with this long term (configset versus managed resources), that is probably a good discussion to have!

risdenk:
I haven't spent much time looking into this but I noticed there were questions about `/` and url encoding. Be aware that `%2F` (url encoded `/`) is handled very differently between webservers, load balancers, and everything in between. I wrote two blogs on this when I went down this rabbit hole fixing this for Apache Knox. https://risdenk.github.io/2017/12/21/percent-2f-webhbase-knox-traefik-part-1.html

Even if `/` is url encoded, there is no guarantee that you see it as `%2F` or `/` on the backend and it could be interpreted as a url separator.

thomascorthals:
> IIRC there have also been suggestions to deprecate and remove managed resources, assuming they are less frequently used and that the more common way of managing these is to push a new copy of the configset. I don't know how much used these are - the fact that this bug is 8 years old hints that the feature is not that popular?

We let our team of content moderators manage synonym lists. Because we only apply them at query time, they can directly enhance the users' search experience that way. The mapping between our backend and this V1 API is very straightforward. (With a workaround for the encoding issue in the client and with the exception of slashes.)

janhoy:
So any suggestions on completing this?

The pragmatic side of me says merge a working bugfix, even if not optimal. 

The more thorugh side says dive deep to investigate whether Jetty and/or Solr code treats `%2F` consistently so we could rely on a more generic/clean solution.

I don't have spare cycles for the latter, so unless others volunteer, I'll merge in a few days.

xtenzQ:
@janhoy I'm not sure if I can improve my fix in some way. 

Also, I apologize for the late reply, I was on the move.

xtenzQ:
@janhoy @gerlowskija @risdenk any updates on the MR? 😄

solr/CHANGES.txt
epugh, line 1: could we bit just a bit more specific in changes?  what does "slash support" mean?  "support a slash character in a managed resource string"???

epugh:
@janhoy if you aren't available, I can shepher merging...

janhoy:
Please go ahead!

epugh:
@xtenzQ  I assume you want this back ported to the next Solr...   Figuring out how to handle CHANGES.txt ;-(

PR #1832, opened 2023-08-08
SOLR-16733: Fail 'test' tasks using JDK20 on Mac

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-16733

# Description

Many (> 700!) tests fail with this OS/JDK combination due to a JDK bug that crops up when the Security Manager is used.

Disabling the Security Manager in this case isn't ideal, but is really kindof necessary to test this OS/Java combination in any meaningful way.

# Solution

~~This PR makes a change to randomization.gradle to skip enabling the SecurityManager if the known problem OS/Java combination is present.  A warning is logged as a breadcrumb to avoid misleading users who wanted to run tests with SecurityManager, including a pointer to SOLR-16733.~~

Rather than disabling the Security Manager in these cases, feedback below has suggested failing the test run entirely before it really kicks off.  If JDK20+Mac+SecurityManager is fundamentally broken, we shouldn't fail loudly so folks don't get the wrong impression.  That is the approach this PR now takes.

# Tests

Manual testing - install JDK20 and run a command similar to:

```
export "RUNTIME_JAVA_HOME=<locationOfJdk20>" ; ./gradlew --info clean test --tests "TestRestoreCore"
```

The 'test' task should fail without this PR but pass with it incorporated.

# Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -127,6 +127,7 @@ apply from: file('gradle/java/javac.gradle')
 apply from: file('gradle/testing/defaults-tests.gradle')
 apply from: file('gradle/testing/randomization.gradle')
 apply from: file('gradle/testing/fail-on-no-tests.gradle')
+apply from: file('gradle/testing/fail-on-unsupported-jdk.gradle')
 apply from: file('gradle/testing/alternative-jdk-support.gradle')
 apply from: file('gradle/java/jar-manifest.gradle')
 
--- a/gradle/testing/fail-on-unsupported-jdk.gradle
+++ b/gradle/testing/fail-on-unsupported-jdk.gradle
@@ -1,0 +1,32 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+configure(rootProject) {
+  task ensureJdkSupported() {
+    doFirst {
+      if (System.getProperty("os.name").toLowerCase(Locale.ROOT).contains("mac") && rootProject.runtimeJavaVersion == JavaVersion.VERSION_20) {
+        throw new GradleException("Tests cannot be run with JDK20 on Mac; see SOLR-16733 for more details.")
+      }
+    }
+  }
+
+  allprojects {
+    tasks.withType(Test) {
+        dependsOn ":ensureJdkSupported"
+    }
+  }
+}
--- a/gradle/testing/randomization.gradle
+++ b/gradle/testing/randomization.gradle
@@ -201,24 +201,20 @@ allprojects {
         // Enable security manager, if requested. We could move the selection of security manager and security policy
         // to each project's build/ configuration but it seems compact enough to keep it here for now.
         if (Boolean.parseBoolean(testOptionsResolved["tests.useSecurityManager"])) {
-          if (System.getProperty("os.name").toLowerCase(Locale.ROOT).contains("mac") && rootProject.runtimeJavaVersion == JavaVersion.VERSION_20) {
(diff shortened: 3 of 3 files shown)


risdenk (commented):
I understand the rationale that tests should be passing - but this just seems like a terrible way to do this. We shouldn't even run tests this way on Mac. It gives a false sense of security that things are working when they really aren't. The tests failing are exactly what happens on Solr on Mac with JDK 20 today.

gerlowskija:
> this just seems like a terrible way to do this. We shouldn't even run tests this way on Mac. It gives a false sense of security that things are working when they really aren't.

I guess where I differ is on the "false sense of security" thing.  It'd be nearly impossible for users not to get the message: our release notes have warnings about this, the bug surfaces immediately upon starting Solr, etc.  Any admin doing any due-diligence is going to catch this right away.

I totally agree it's a bad user experience for the folks who have to hit the bug the hard way, but it's probably safer than doing something more comprehensive like disabling Security Manager for folks without their explicit intervention and then hoping they see whatever message we log about that.

If I'm being overly cautious and am the minority there, I'm happy to add some `bin/solr` level fix to this PR (assuming that's what you had in mind @risdenk).  But wanted to call out what seems like at least a potential security risk.

(And even with a `bin/solr` level change if we go that route, I think we'd still need the gradle-level change in this PR to keep our JUnit tests in line with "real" packaged deployments?)

risdenk:
> "false sense of security" thing

false sense of security - My poor wording since we are working around security manager. I mean that we shouldn't even run the tests at all. Disabling the security manager for tests and saying "hey look on JDK 20 on Mac tests pass" but we just hide the fact that we disable the default enabled security manager just to make tests pass seems wrong.

I would prefer if we didn't even run any tests or at least fail loudly and say Mac OS X + JDK 20 is unsupported or something like that.

I don't have anything in mind about bin/solr or other smarter ways to potentially make the user aware of this. I'm just not a fan of trying to make the tests pass just to make the tests pass. They are failing because JDK 20 + Mac + Jetty + security manager is broken - which is the default out of the box config for Solr.

gerlowskija:
I hear what you're getting at about JDK20+Mac+SM being fundamentally broken.  We don't want to give users a mistaken impression that there's not this big issue.

But doesn't disabling testing entirely for that JDK+OS "throw the baby out with the bathwater"?

Sure, the bug probably makes JDK20 a non-starter for a lot of folks, but I'd bet there's a decent chunk of users who don't really care about SM and would be happy to disable it to use JDK20.  Shouldn't we still gather test data for that set of users?  Surely there's a way to do that while still signaling clearly (in release notes, announce emails, etc.) that JDK20+Mac+SM has issues?

dsmiley:
Confusingly, we don't use the Java SecurityManager in tests for "Security".  We use it to enforce file system rules in our tests  to enforce best-practices.  This ensures we don't waste time root causing test instability due to some _other_ test that leaked state into a place seen by a _later_ test.  I think we should alter the security policy of the test config to only be about the file system and not bother maintaining all the other rules which are more toil than value.

janhoy:
Personally I switch to JDK11 or 17 when running tests on my Mac. The alternative is to fail loudly and ask the dev to choose JDK17.

There's a tiny risk that if more committers start using JDK20 for their dev work on a mac, someone may commit code that is incompatible with secMan rules, but won't be caught by tests. Although it would be caught in PR Crave tests om Linux...

Guess I'm +0.5 on this hack..

gerlowskija:
> I mean that we shouldn't even run the tests at all. Disabling the security manager for tests and saying "hey look on JDK 20 on Mac tests pass" but we just hide the fact that we disable the default enabled security manager just to make tests pass seems wrong.

> The alternative is to fail loudly and ask the dev to choose JDK17.

It sounds like there's some preference for truncating the build altogether when this combination is detected, rather than my initial idea of trying to preserve _some_ test results for curious JDK20+Mac users who don't care about Sec Manager.

That's fine by me - my primary interest here is in clearing up the build noise so regression-detection is feasible again.  I'll update the PR shortly.

gerlowskija:
Alright, I've updated the PR in line with my comment above and previous feedback.  With these changes, the whole test run now fails with:

```
$ export "RUNTIME_JAVA_HOME=/usr/local/Cellar/openjdk/20.0.1" ; ./gradlew --info clean test --tests "TestRestoreCore"
<snip>

* Where:
Script '/Users/gerlowskija/checkouts/solr/gradle/testing/fail-on-unsupported-jdk.gradle' line: 22

* What went wrong:
Execution failed for task ':ensureJdkSupported'.
> Tests cannot be run with JDK20 on Mac; see SOLR-16733 for more details.

<snip>

BUILD FAILED in 5s
```

If I can get a +1 from someone, I'll probably aim to merge in relatively short order so we can start to see how our tests look without the additional noise and since it sounds like there's at least some consensus on this approach.

HoustonPutman (approved):
I like this

PR #1835, opened 2023-08-10
SOLR-16927 Allow SolrClientCache clients to use Jetty HTTP2 clients

stillalex:
https://issues.apache.org/jira/browse/SOLR-16927

# Description

Allow SolrClientCache to work with Jetty http2 clients.

# Solution

This change is allowing the SolrClientCache users to use a Jetty http2 client, in parallel with the internal Solr http client.

All http client creation (both old and new) is moved into the SolrClientCache, this will make it a single source for all the streaming code. It will also fallback to Jetty http2 by default.
I also poked around the timeouts a bit, the current values based on system properties will overwrite the internal http client ones, which is bad when the client values are larger than the local ones.

I had to touch some of the streaming classes, as the http2 client needs to be properly closed, which was not the case everywhere.

I added a benchmark, but I still need to tweak it a bit before publishing numbers.

Just for future reference of what remains to be done in this class, most of the internal use of this cache is based on the UpdateShardHandler.getDefaultHttpClient (SOLR-16503) - the CoreContainer reference, so the complete refactoring is blocked until that transition completes.

# Tests

# Checklist

### DIFF
--- a/solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/SolrClientCache.java
+++ b/solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/SolrClientCache.java
@@ -50,11 +50,11 @@ public class SolrClientCache implements Serializable, Closeable {
   // Set the floor for timeouts to 60 seconds.
   // Timeouts cans be increased by setting the system properties defined below.
   private static final int MIN_TIMEOUT = 60000;
-  private static final int minConnTimeout = Math.max(
-      Integer.getInteger(HttpClientUtil.PROP_CONNECTION_TIMEOUT, MIN_TIMEOUT),
-      MIN_TIMEOUT);
-  private static final int minSocketTimeout = Math.max(
-      Integer.getInteger(HttpClientUtil.PROP_SO_TIMEOUT, MIN_TIMEOUT), MIN_TIMEOUT);
+  private static final int minConnTimeout =
+      Math.max(
+          Integer.getInteger(HttpClientUtil.PROP_CONNECTION_TIMEOUT, MIN_TIMEOUT), MIN_TIMEOUT);
+  private static final int minSocketTimeout =
+      Math.max(Integer.getInteger(HttpClientUtil.PROP_SO_TIMEOUT, MIN_TIMEOUT), MIN_TIMEOUT);
 
   private final Map<String, SolrClient> solrClients = new HashMap<>();
   private final HttpClient httpClient;
@@ -101,7 +101,8 @@ public class SolrClientCache implements Serializable, Closeable {
     return client;
   }
 
-  public static CloudHttp2SolrClient newCloudHttp2SolrClient(String zkHost, Http2SolrClient http2SolrClient) {
+  public static CloudHttp2SolrClient newCloudHttp2SolrClient(
+      String zkHost, Http2SolrClient http2SolrClient) {
     final List<String> hosts = List.of(zkHost);
     var builder = new CloudHttp2SolrClient.Builder(hosts, Optional.empty());
     // using internal builder to ensure the internal client gets closed
@@ -136,17 +137,19 @@ public class SolrClientCache implements Serializable, Closeable {
   @Deprecated
   private static void adjustTimeouts(SolrClientBuilder<?> builder, HttpClient httpClient) {
     builder.withHttpClient(httpClient);
-    int socketTimeout= Math.max(minSocketTimeout, builder.getSocketTimeoutMillis());
+    int socketTimeout = Math.max(minSocketTimeout, builder.getSocketTimeoutMillis());
     builder.withSocketTimeout(socketTimeout, TimeUnit.MILLISECONDS);
-    int connTimeout= Math.max(minConnTimeout, builder.getConnectionTimeoutMillis());
+    int connTimeout = Math.max(minConnTimeout, builder.getConnectionTimeoutMillis());
     builder.withConnectionTimeout(connTimeout, TimeUnit.MILLISECONDS);
   }
 
-  public static Http2SolrClient newHttp2SolrClient(String baseUrl, Http2SolrClient http2SolrClient) {
+  public static Http2SolrClient newHttp2SolrClient(
+      String baseUrl, Http2SolrClient http2SolrClient) {
     return newHttp2SolrClientBuilder(baseUrl, http2SolrClient).build();
   }
 
-  private static Http2SolrClient.Builder newHttp2SolrClientBuilder(String baseUrl, Http2SolrClient http2SolrClient) {
+  private static Http2SolrClient.Builder newHttp2SolrClientBuilder(
+      String baseUrl, Http2SolrClient http2SolrClient) {
     var builder = new Http2SolrClient.Builder(baseUrl);
     if (http2SolrClient != null) {
       builder = builder.withHttpClient(http2SolrClient);
--- a/solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/SolrClientCache.java
+++ b/solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/SolrClientCache.java
@@ -18,7 +18,6 @@ package org.apache.solr.client.solrj.io;
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.io.Serializable;
 import java.lang.invoke.MethodHandles;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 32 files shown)


stillalex:
looking for a review on this. @dsmiley, @epugh @joel-bernstein I hope you don't mind I added you. looking forward to your thoughts!

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/stream/CommitStream.java
```diff
@@ -55,6 +55,7 @@ public class CommitStream extends TupleStream implements Expressible {
   private TupleStream tupleSource;
 
   private transient SolrClientCache clientCache;
+  private transient boolean isCloseCache;
```
risdenk, line 58: Do we need this new boolean `isCloseCache`? Can't we just check if `clientCache == null` in the close method and set it to `clientCache = null` after `clientCache.close();`?

risdenk (approved):

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/stream/CommitStream.java
stillalex: you are right. there are 2 patterns of doing this in the PR (flag vs null check). I will switch to this one, it's cleaner.

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/stream/Facet2DStream.java
```diff
@@ -325,13 +322,7 @@ public void open() throws IOException {
...
       cloudSolrClient = cache.getCloudSolrClient(zkHost);
     } else {
-      final List<String> hosts = new ArrayList<>();
-      hosts.add(zkHost);
-      cloudSolrClient =
-          new CloudLegacySolrClient.Builder(hosts, Optional.empty())
-              .withSocketTimeout(30000, TimeUnit.MILLISECONDS)
-              .withConnectionTimeout(15000, TimeUnit.MILLISECONDS)
-              .build();
+      cloudSolrClient = SolrClientCache.newCloudHttp2SolrClient(zkHost, null);
```
dsmiley, line 325: This line is curious... it seems you have made a public static method on SolrClientCache for creating a CloudHttp2SolrClient, thus making this class a builder for anyone anywhere who wants such a client?  Shouldn't we let the builder of CloudSolrClient do such a job?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -829,6 +824,7 @@ private NamedList<Object> processErrorsAndResponse(
         // no processor specified, return raw stream
         NamedList<Object> rsp = new NamedList<>();
         rsp.add("stream", is);
+        rsp.add("responseStatus", httpStatus);
```
dsmiley, line 827: Maybe we should name this "httpStatus" and should use an integer so as to keep generic of the type of HttpClient we are using?

dsmiley (commented):
Thanks for working on this!

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/stream/Facet2DStream.java
stillalex: yes I agree with your concern. but I added it to replace all manual builder use from across the streaming module. there is no way to define visibility to module only, so this is what I ended up with. I prefer this single point of entry because it handles the timeout setting correctly, rather than repeating this code all over. the inconsistency was that some of the builders have increased timeouts, some don't, it's hard to understand what goes where in the streaming module.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
stillalex: this is an int, check a few lines above `int httpStatus = response.getStatus();` and yes this was exactly what I was trying to do, separate implementation details from the response :)

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/SolrClientCache.java
```diff
@@ -40,85 +43,131 @@
  *
  * <p>TODO: Cut this over to using Solr's new Http2 clients
  */
-public class SolrClientCache implements Serializable {
+public class SolrClientCache implements Serializable, Closeable {
```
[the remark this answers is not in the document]
stillalex: would removing it and seeing if any test fails be sufficient as a check if we can do this or not?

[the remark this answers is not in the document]
stillalex: > Maybe a subclass approach would work better

I am not convinced. the complexity comes from the streaming code that has a lot of ins and outs. ideally it would work both types of clients irrespective of where they come from (internal node communication or external apps). I don't think the code would be as clean, but I can definitely give it a go and see what fits better.

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/stream/CommitStream.java
stillalex: sorry @risdenk I misread your question. in this case no, you can't get rid of the flag because if the cache is not null you don't know if it came from the context or it was created locally. if it came from the context it's reused for multiple streams so you can't close it here, you have to close it at the parent location where it was created.

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/stream/Facet2DStream.java
stillalex: thinking about it a bit more, we can transition all streaming code to use 'client cache' instances instead of 'http clients' directly and this would allow all the builder code to be in one place with no public static methods.

dsmiley: Yeah I think that's best.
And if there is timeout initialization handling in SolrClientCache, I wonder why it's there; why isn't it in the builder of CloudSolrClient etc.?  SCC should just be a cache IMO.  Nothing much to it.  CC @joel-bernstein

stillalex: the way I read it, this class attempts to set a min timeout(~60s) irrespective of what was set on client originally ... and also in passing manages to reset the timeouts back to 60s if they are bigger (which is an unfortunate side effect I wanted to address with this change) - I'm thinking the default httpclient on the UpdateShardHandler might play a central role here.

stillalex: @dsmiley did a complete pass over all classes and moved everything to cache. please take a look.

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/SolrClientCache.java
```diff
@@ -40,85 +43,131 @@
  *
  * <p>TODO: Cut this over to using Solr's new Http2 clients
```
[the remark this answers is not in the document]
stillalex: line removed

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/SolrClientCache.java
```diff
@@ -40,85 +43,131 @@
...
+  private static final int MIN_TIMEOUT = 60000;
+  private static final int minConnTimeout =
       Math.max(
-          Integer.parseInt(System.getProperty(HttpClientUtil.PROP_SO_TIMEOUT, "60000")), 60000);
+          Integer.getInteger(HttpClientUtil.PROP_CONNECTION_TIMEOUT, MIN_TIMEOUT), MIN_TIMEOUT);
+  private static final int minSocketTimeout =
+      Math.max(Integer.getInteger(HttpClientUtil.PROP_SO_TIMEOUT, MIN_TIMEOUT), MIN_TIMEOUT);
+
+  private final Map<String, SolrClient> solrClients = new HashMap<>();
+  private final HttpClient httpClient;
```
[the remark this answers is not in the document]
stillalex: rename done

stillalex:
@joel-bernstein do you think you will have some time to review this PR?

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/stream/CommitStream.java
dsmiley: The name "isCloseCache" is clumsy IMO; I think "doCloseCache" is far clearer.

dsmiley (commented):
Other than a little rename, looks good.
Please don't force-push!
My concerns about the necessity of the timeouts can be deferred -- "progress not perfection".

stillalex:
sounds good. will rename and merge in the next few days. thanks @dsmiley!

PR #1849, opened 2023-08-17
SOLR-16941: Use smarter default Solr URL in CLI

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-16941

I've also updated one of the SSL tests to show that this works (defaulting the scheme at least)

### DIFF
--- a/solr/core/src/java/org/apache/solr/cli/SolrCLI.java
+++ b/solr/core/src/java/org/apache/solr/cli/SolrCLI.java
@@ -75,7 +75,7 @@ public class SolrCLI implements CLIO {
       TimeUnit.NANOSECONDS.convert(1, TimeUnit.MINUTES);
 
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
-  public static final String DEFAULT_SOLR_URL = "http://localhost:8983";
+
   public static final String ZK_HOST = "localhost:9983";
 
   public static final Option OPTION_ZKHOST =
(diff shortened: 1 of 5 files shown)


solr/core/src/java/org/apache/solr/cli/SolrCLI.java
```diff
@@ -97,13 +97,13 @@ public class SolrCLI implements CLIO {
...
               "Base Solr URL, which can be used to determine the zkHost if that's not known; defaults to: "
-                  + DEFAULT_SOLR_URL
+                  + getDefaultSolrUrl()
                   + '.')
           .build();
   public static final Option OPTION_VERBOSE =
       Option.builder("verbose").required(false).desc("Enable more verbose command output.").build();
 
-  // should this be boolean or just a otption?
+  // should this be boolean or just an option?
```
epugh, line 106: thank you for fixing this typo...  it' bugged me for a while.

solr/packaging/test/test_ssl.bats
```diff
@@ -54,8 +54,8 @@ teardown() {
   solr start -c
   solr assert --started https://localhost:8983/solr --timeout 5000
 
-  run curl --cacert "$ssl_dir/solr-ssl.pem" 'https://localhost:8983/solr/admin/collections?action=CREATE&collection.configName=_default&name=test&numShards=2&replicationFactor=1&router.name=compositeId&wt=json'
-  assert_output --partial '"status":0'
+  run solr create -c test -s 2
```
epugh, line 57: Did you mean this change?    I guess this is demonstrating that bin/solr create works with https?    When you define those variables?

epugh (commented):
LGTM.   This just touches bin/solr, what about bin/solr.cmd?

solr/packaging/test/test_ssl.bats
epugh: I looked, and I see another curl example elsewhere in the tests.  I actually like the example of using curl with the `solr-ssl.pem`, it helped me understand how to interact with Solr in SSL ;-)

HoustonPutman: Yep, it's to test the usage!

HoustonPutman:
> This just touches bin/solr, what about bin/solr.cmd?

Not exactly sure but from my surface level research, the windows variables do not need to be exported like they do in bash, so there's nothing to change. The bash changes are just to export the variables we use in the CLI.

epugh (approved):

HoustonPutman:
Ok fixed the issues. @epugh Going forward (10.0) the `package` tool should probably work like the rest of the tools, with regards to the Solr Port.

PR #1857, opened 2023-08-22
SOLR-16934: Give securityManager permission for client TLS

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-16934


solr/server/etc/security.policy
```diff
@@ -191,6 +191,12 @@ grant {
   permission java.io.FilePermission "${solr.jetty.truststore}", "read,write,delete,readlink";
   permission java.io.FilePermission "${solr.jetty.truststore}${/}-", "read,write,delete,readlink";
 
+  permission java.io.FilePermission "${javax.net.ssl.keyStore}", "read,write,delete,readlink";
```
risdenk, line 194: is write and delete needed? I'm assuming no and only read is needed.

I also don't know if the `/` directory version is needed. Isn't `javax.net.ssl.*Store` going to point directly to the file in question?

HoustonPutman: That's fair. I was just copying from above. Will fix.

risdenk (approved):

PR #1859, opened 2023-08-22
SOLR-16825: Migrate v2 definitions to 'api' module

gerlowskija:

# Description

SOLR-16825 added a new gradle module, 'api', which holds v2 API definitions as interfaces.  This allows us to generate an OAS (and SolrRequest implementations from that) as a part of the solrj build.

But these artifacts (the OAS and generated Java code), only cover the v2 APIs that have interfaces in the 'api' module.  We need to extract interfaces to live in 'api' for each v2 API in 'core' that doesn't already have one.
  
# Solution

This PR creates 'api' interfaces for a number of v2 APIs, allowing SolrRequest implementations to be generated for them.  The following APIs are covered in this PR:

- delete replica property
- delete replica
- delete (i.e. "clear") node
- delete collection snapshot
- delete collection backup

# Tests

Manual inspection of the generated OAS and SolrRequest implementations.

# Checklist


solr/api/src/java/org/apache/solr/client/api/model/Constants.java
```diff
@@ -0,0 +1,51 @@
...
+package org.apache.solr.client.api.model;
+
+public class Constants {
+
+  private Constants() {
+    /* Private ctor prevents instantiation */
+  }
+
+  /** A parameter to specify the name of the backup repository to be used. */
+  public static final String BACKUP_REPOSITORY = "repository";
```
gerlowskija, line 27: [0] The additions here are a real shame, as they're duplicates of constants that already exist in SolrJ.

Should we move some of the solrj "constants" files (i.e. CollectionAdminParams, CommonParams) into 'api' to avoid the duplication?  (To handle backcompat, in 9.x we could have a deprecated copy of each of these constants files in solrj that extends its 'api' module counterpart)

epugh (commented):
My eyes bleed a bit at all the change, but it LGTM.   I guess we have to accept the verbosity is critical to getting the nice things ;-)

epugh (approved):

gerlowskija:
Yeah, the diff is large for sure. 

I'd argue though that this is pretty deceptive though, re: "verbosity".  This commit adds lines in isolation sure, but if it ultimately lets us delete a bunch of existing SolrJ (and maybe even ref-guide) code that we've been maintaining by hand up to this point...well then the end result will be a huge "net-negative" in terms of LOC.

(In theory we could nuke the v1 SolrRequest classes in this PR, if we wanted to commit to removing them in 10.0.  I was reluctant to do that here though, since there's a lot of v2 api work remaining and the 10.0 timeline being a real unknown...) 

Anyway in the short term I def agree it's non-ideal.

PR #1860, opened 2023-08-22
SOLR-16265: Fix NPE for req in Http2SolrClient

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-16265

Fixing error seen below when passing a bad url to the client.

> ERROR: Cannot invoke "org.eclipse.jetty.client.api.Request.abort(java.lang.Throwable)" because "req" is null


stillalex (approved):

stillalex:
good catch @HoustonPutman!

PR #1861, opened 2023-08-22
SOLR-16945: Add multi-threaded warming support for CaffeineCache and QuerySenderListener

tpunder:
https://issues.apache.org/jira/browse/SOLR-16945

# Description

This adds opt-in multi-threaded warming support to CaffeineCache and QuerySenderListener.

# Solution

The implementation for both CaffeineCache and QuerySenderListener are basically the same. They both now make use of an `ExecutorUtil.MDCAwareThreadPoolExecutor` with the user configured number of threads to run any warming queries. If multi-threaded warming was not configured (or the number of threads is set to 0 or 1) then the existing single-threaded implementation is used. If a negative number of threads are specified then `Runtime.getRuntime().availableProcessors()` is used for the number of threads. The multi-threaded implementation is opt-in.

# Tests

I've run manual tests on this change to verify the single-threaded and multi-threaded code paths are working as expected based on logging messages showing the threads used for the warming queries and based on the total time it takes to run the warming with various numbers of threads.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/QuerySenderListener.java
+++ b/solr/core/src/java/org/apache/solr/core/QuerySenderListener.java
@@ -21,7 +21,16 @@ import static org.apache.solr.common.params.CommonParams.DISTRIB;
 import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import org.apache.lucene.util.NamedThreadFactory;
 import org.apache.solr.common.params.CommonParams;
+import org.apache.solr.common.params.EventParams;
+import org.apache.solr.common.util.ExecutorUtil;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.request.LocalSolrQueryRequest;
 import org.apache.solr.request.SolrQueryRequest;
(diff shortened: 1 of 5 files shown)


solr/core/src/java/org/apache/solr/core/QuerySenderListener.java
```diff
@@ -44,65 +53,131 @@ public QuerySenderListener(SolrCore core) {
...
     if (allLists == null) return;
-    for (NamedList<Object> nlst : allLists) {
+    final int threads = getThreadsParam();
+
+    if (0 == threads || 1 == threads || allLists.isEmpty()) { // Non-threaded code path
+      for (NamedList<Object> nlst : allLists) {
+        runQuery(newSearcher, currentSearcher, nlst);
+      }
+    } else { // Multi-threaded code path
+      final int nThreads = threads > 0 ? threads : Runtime.getRuntime().availableProcessors();
```
janhoy, line 71: The fallback where `threads <= 0` will never happen since the if clause above grabs `0 == threads`. Was the intention to auto-configure threads based on available CPUs if you configure a negative number, or 0?

tpunder: My intent is that negative numbers will default to using the number of available processors. However, I originally was not catching `0 == threads` but later decided that it made more sense to treat `0 == threads` as meaning "no threading". That behavior seems more consistent with the `facet.threads` parameter which has the similar behavior for a value of 0 and negative values.

tpunder: I'll add documentation to the Solr Guide for this behavior.

tpunder:
I've now included docs for the Reference Guide as part of this PR.

solr/solr-ref-guide/modules/configuration-guide/pages/caches-warming.adoc
```diff
@@ -343,6 +355,34 @@ A key best practice is to modify these defaults before taking your application t
...
 ====
 
+### Multi-threaded listener warming
+
+If you have a large list of queries you are running via the listeners you might want to enable multi-threading listener warming by specifying the `<threads>` element:
+
+[source,xml]
+----
+<listener event="newSearcher" class="solr.QuerySenderListener">
+  <threads>4</threads>
```
janhoy, line 365: Normally we use `<int name="threads">4</int>` style of namedList config many places in solrconfig. I suppose you get some type casting for free then? Personally I  am not opposed to a more custom XML syntax, just asking whether you considered that format as an alternative?

tpunder: You're absolutely right. That was a documentation typo on my part. It should be `<int name="threads">4</int>` which is that is what my local configuration files for testing are using. I've updated the PR.

PR #1866, opened 2023-08-28
SOLR-16825: Migrate API definitions to submodule, pt 2

gerlowskija:

# Description

[SOLR-16825](https://issues.apache.org/jira/browse/SOLR-16825) added a new gradle module, 'api', which holds v2 API definitions as interfaces. This allows us to generate an OAS (and SolrRequest implementations from that) as a part of the solrj build.

But these artifacts (the OAS and generated Java code), only cover the v2 APIs that have interfaces in the 'api' module. We need to extract interfaces to live in 'api' for each v2 API in 'core' that doesn't already have one.

# Solution

This PR creates 'api' interfaces for a number of v2 APIs, allowing SolrRequest implementations to be generated for them. The following APIs are covered in this PR:

- list-aliases
- list-incremental-backups
- list-collections
- list-configsets
- reload-collection
- various alias-property CRUD APIs

# Tests

Manual inspection of the generated OAS and SolrRequest implementations.

# Checklist


gerlowskija:
Kindof?

That approach works fine as far as JAX-RS and Solr are concerned.  In fact I'm sure you noticed it's how things were structured when these classes lived in 'core'.

The problem is code-generation.  As each SolrRequest class is being generated, it needs to be able to import the Java POJOs it relies on.  Right now our template has some pretty simple logic around imports - we essentially expect all POJOs to be in a single package without any static class nesting that'd change the import pattern. (see [here](https://github.com/apache/solr/blob/3cbb97549cef1664a1d6b01fd2d1367cefbb0665/solr/solrj/build.gradle#L135) and [here](https://github.com/apache/solr/blob/3cbb97549cef1664a1d6b01fd2d1367cefbb0665/solr/solrj/src/resources/java-template/api.mustache#L39))

(openapi-generator has the ability to generate model POJOs for users, and handles imports much more gracefully in this case.  We've decided _not_ to generate models though, since they'd essentially duplicate the POJOs that we already have.  The "import" problem is the main downside of that decision.)

Anyway, this feels like a fixable problem to me if it's worth it to us to dig in.  I spent a few hours tweaking the OpenAPI spec and `api.mustache` template without much success, but there's gotta be some fix out there.  I guess it just depends how strongly we care about being able to have POJOs nested the way @HoustonPutman described...

HoustonPutman:
Thanks for the explanation! In general it seems a little daunting to have them all separated and there, when it would be easier to work with them in the same place. Maybe it'll be worth it to look into sometime. Maybe it won't... But I'm not going to block you on it here.

gerlowskija:
Sounds good; I'll merge this in for now and we can revisit later if needed.

PR #1868, opened 2023-08-28
SOLR-15367 [9.x] Convert "rid" functionality into a default Tracer

stillalex:
https://issues.apache.org/jira/browse/SOLR-15367

# Description

**PR for 9.x branch**
This is a 9.x version of https://github.com/apache/solr/pull/1854 for evaluation.

Almost all changes relevant to OpenTracing are in SimplePropagator. I made an effort to minimize changes that are not on main branch. ideally we'd reach a state where this has working basic generation of traces and remains frozen on 9.x until 10.x starts releasing and everyone transitions automatically to the new OTEL-based impl.

Similar to main branch, this is enabled by default. this could be an issue for setups that already have tracing enabled, so I am ok with off by default if switching this off via system property is too much to handle in a minor upgrade.

# Solution

# Tests

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/util/tracing/SimplePropagator.java
+++ b/solr/core/src/java/org/apache/solr/util/tracing/SimplePropagator.java
@@ -1,0 +1,304 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.util.tracing;
+
+import io.opentracing.Scope;
+import io.opentracing.ScopeManager;
+import io.opentracing.Span;
+import io.opentracing.SpanContext;
+import io.opentracing.Tracer;
+import io.opentracing.Tracer.SpanBuilder;
+import io.opentracing.propagation.Format;
+import io.opentracing.propagation.TextMap;
+import io.opentracing.tag.Tag;
+import io.opentracing.util.GlobalTracer;
+import io.opentracing.util.ThreadLocalScopeManager;
+import java.lang.invoke.MethodHandles;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.concurrent.atomic.AtomicLong;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SimplePropagator {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  private static final String TRACE_HOST_NAME =
+      System.getProperty("trace_host_name", System.getProperty("host"));
+
+  static final String TRACE_ID = System.getProperty("TRACE_ID", "X-Trace-Id");
+
+  private static final AtomicLong traceCounter = new AtomicLong(0);
+
+  public static void load() {
+    GlobalTracer.registerIfAbsent(
+        () -> {
+          log.info("Always-on trace id generation enabled.");
+          return new SimplePropagatorTracer();
+        });
+  }
+
+  /**
+   * Tracer that only aims to do simple header propagation, tailored to how Solr works.
+   *
+   * <p>Heavily inspired from JaegerTracer, NoopTracer
+   */
+  static class SimplePropagatorTracer implements Tracer {
+
+    private final ScopeManager scopeManager = new ThreadLocalScopeManager();
+
+    @Override
+    public ScopeManager scopeManager() {
+      return scopeManager;
+    }
+
+    @Override
+    public Span activeSpan() {
+      return scopeManager.activeSpan();
+    }
+
+    @Override
+    public Scope activateSpan(Span span) {
+      return scopeManager.activate(span);
+    }
+
+    @Override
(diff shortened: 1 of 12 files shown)


stillalex:
@dsmiley raising this PR for a discussion on building an OpenTracing version of the always-on tracer that went into main branch based on OTEL.

solr/core/src/java/org/apache/solr/util/tracing/SimplePropagator.java
```diff
@@ -0,0 +1,304 @@
...
+    public SimplePropagatorSpanBuilder(ScopeManager scopeManager) {
+      this.scopeManager = scopeManager;
+    }
+
+    @Override
+    public Span start() {
+      if (parent != null) {
+        return new SimplePropagatorSpan(parent);
+      } else if (scopeManager.activeSpan() != null) {
+        return new SimplePropagatorSpan(scopeManager.activeSpan().context());
```
dsmiley, line 144: This is going to be the common case as more spans show up _within_ Solr.  Could SimplePropagatorSpan and SimplePropagatorSpanContext be combined into one?  Create no new object; the context could *be* the span if all we are doing is propagation.

I even wonder if we need a distinct builder?

solr/core/src/java/org/apache/solr/util/tracing/SimplePropagator.java
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    @Override
+    public void close() {}
+
+    @Override
+    public <C> void inject(SpanContext spanContext, Format<C> format, C carrier) {
+      if (!format.equals(Format.Builtin.HTTP_HEADERS)) {
+        throw new IllegalArgumentException("unsupported format " + format);
+      }
```
dsmiley, line 93: Is this constraint necessary?  What if we inject & extract to transfer to the Overseer?  (I have done this at work)

solr/core/src/java/org/apache/solr/util/tracing/SimplePropagator.java
```diff
@@ -0,0 +1,304 @@
...
+      return this;
+    }
+
+    @Override
+    public <T> Span setTag(Tag<T> arg0, T arg1) {
+      return this;
+    }
+  }
+
+  private static String newTraceId() {
```
dsmiley, line 301: nit: this method seems to hide down here after these long inner subclasses.  Above them is probably better organized.

solr/core/src/java/org/apache/solr/util/tracing/TraceUtils.java
```diff
@@ -32,7 +33,7 @@ public static void setDbInstance(SolrQueryRequest req, String coreOrColl) {
   }
 
   public static void ifNotNoop(Span span, Consumer<Span> consumer) {
-    if (span != null && !(span instanceof NoopSpan)) {
+    if (span != null && !(span instanceof NoopSpan) && !(span instanceof SimplePropagatorSpan)) {
```
dsmiley, line 36: Arguably, SimplePropagatorSpan should implement NoopSpan?

dsmiley (commented):
Thanks for bringing this to 9!

solr/core/src/java/org/apache/solr/util/tracing/TraceUtils.java
stillalex: ugh, of course, good one

solr/core/src/java/org/apache/solr/util/tracing/SimplePropagator.java
stillalex: good suggestion. I merged the 2 but it introduced some minor complexity dealing with the parent context, but not a big issue.

solr/core/src/java/org/apache/solr/util/tracing/SimplePropagator.java
```diff
@@ -139,11 +147,15 @@ public SimplePropagatorSpanBuilder(ScopeManager scopeManager) {
...
-        return new SimplePropagatorSpan(parent);
+        if (parent instanceof SimplePropagatorSpan) {
+          return (SimplePropagatorSpan) parent;
+        } else {
+          return NoopSpan.INSTANCE;
+        }
       } else if (scopeManager.activeSpan() != null) {
-        return new SimplePropagatorSpan(scopeManager.activeSpan().context());
+        return scopeManager.activeSpan();
       } else {
```
dsmiley, line 157: I think this is a double-lookup in a ThreadLocal?

stillalex: yes. good catch

solr/core/src/java/org/apache/solr/util/tracing/SimplePropagator.java
dsmiley: In practice at the moment, the format will *always* be HTTP since we control where we inject.  But if we get to add the Overseer as you have done in main branch, then it will mysteriously not propagate, and then maybe we'll "fix" it by removing these 3 lines.  I don't think need to care about what the format is; just try to inject best-effort.

stillalex: > I don't think need to care about what the format is; just try to inject best-effort.
I agree. the `inject` is easy enough, you just return early. the `extract` is now a `return NoopSpan.INSTANCE.context();` and this I'm not sure will work correctly.
not having some basic test coverage for this setup worries me a bit.

stillalex:
@dsmiley would like to merge this soon. do you think the context injection/extraction parts are correct now?
and apologies for the force merge, I wanted to rebase on top of other backports and ended up with this.

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
```diff
@@ -1341,5 +1349,10 @@ public Builder withDefaultClusterProperty(String key, String value) {
       cluster.put(key, value);
       return this;
     }
+
```
dsmiley, line 1352: Can you please add javadoc here; it may not be obvious what it does and why a test might want to set it.
I'm thinking maybe:
> Disables the default/built-in simple trace ID generation/propagation.  For example if you want to use a  MockTracer or a "real" Tracer, you'll need to call this so that it doesn't displace the Tracer you want to use.

dsmiley (approved):

stillalex:
thank you @dsmiley I really appreciate your diligence here. very happy to see this land on 9.x too!

PR #1869, opened 2023-08-28
logging, MDC: collection for action=DELETE

dsmiley:
Logging MDC "collection" is sometimes not set for certain admin API commands like DELETE.

----
Been using this for awhile.  I don't love that an assumption is made that perhaps isn't necessarily true?  Although I like it better than not setting MDC.  I wish there was an internal standardization at some layer that would *always* put the collection being operated on into a consistent spot, lower level than the surface layer API.

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/api/collections/OverseerCollectionMessageHandler.java
+++ b/solr/core/src/java/org/apache/solr/cloud/api/collections/OverseerCollectionMessageHandler.java
@@ -113,7 +113,9 @@ public class OverseerCollectionMessageHandler implements OverseerMessageHandler,
 
   @Override
   public OverseerSolrResponse processMessage(ZkNodeProps message, String operation) {
-    MDCLoggingContext.setCollection(message.getStr(COLLECTION));
+    // sometimes overseer messages have the collection name in 'name' field, not 'collection'
+    MDCLoggingContext.setCollection(
+        message.getStr(COLLECTION) != null ? message.getStr(COLLECTION) : message.getStr(NAME));
     MDCLoggingContext.setShard(message.getStr(SHARD_ID_PROP));
     MDCLoggingContext.setReplica(message.getStr(REPLICA_PROP));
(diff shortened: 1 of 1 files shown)


solr/core/src/java/org/apache/solr/cloud/api/collections/OverseerCollectionMessageHandler.java
```diff
@@ -113,7 +113,9 @@ public OverseerCollectionMessageHandler(
 
   @Override
   public OverseerSolrResponse processMessage(ZkNodeProps message, String operation) {
-    MDCLoggingContext.setCollection(message.getStr(COLLECTION));
+    // sometimes overseer messages have the collection name in 'name' field, not 'collection'
```
epugh, line 116: i ran into this with the CLI as well....   is it --name or --collection?

stillalex:
I think CreateCollectionCmd also uses 'NAME'' https://github.com/apache/solr/blob/22ac9105ea0382215c8fe558fc55977dc98aa14c/solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java#L111

stillalex (approved):

solr/core/src/java/org/apache/solr/cloud/api/collections/OverseerCollectionMessageHandler.java
```diff
@@ -113,7 +113,9 @@ public OverseerCollectionMessageHandler(
...
+    MDCLoggingContext.setCollection(
+        message.getStr(COLLECTION) != null ? message.getStr(COLLECTION) : message.getStr(NAME));
```
stillalex, line 118: [small nit] instead of `COLLECTION' which is coming from `CollectionAdminParams` is would be better to use `ZkStateReader.COLLECTION_PROP`.

dsmiley: I'll switch.

PR #1873, opened 2023-08-30
Removing List.toArray calls with pre-sized Arrays

renatoh:
Here the explanation from the static code analysis tool:
In older Java versions, using a pre-sized array was recommended, as the reflection call necessary to create an array of proper size was quite slow.
However, since late updates of OpenJDK 6, this call was intrinsified, making the performance of the empty array version the same, and sometimes even better, compared to the pre-sized version. Also, passing a pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray calls. This may result in extra nulls at the end of the array if the collection was concurrently shrunk during the operation.


madrob:
Which static code analysis reports on this? Can we add this inspection to our build so that we don't regress on it in the future?

renatoh:
@madrob 
The one build in to intellij, not sure if this can be added to the build, but I assume this would be picked up by any other static code analyzer too.
Screenshot 2023-08-30 at 20 49 22

renatoh:
@madrob I might have been wrong with the static code analysis, looks like it is even a compiler warning:
Screenshot 2023-08-30 at 21 04 30

Here a pretty comprehensive comparision, including JMH-Performance tests:
https://www.baeldung.com/java-collection-toarray-methods

renatoh:
i don't see whats why the Precommit is complaining about SpellCheckComponent.java, I did run tidy again, and no changes appear.

renatoh:
hmm, now JavaSpotless check is failing on a class I have not touched, this one:
 src/java/org/apache/solr/cloud/api/collections/OverseerCollectionMessageHandler.java

cpoerschke:
> hmm, now JavaSpotless check is failing on a class I have not touched, this one:
> src/java/org/apache/solr/cloud/api/collections/OverseerCollectionMessageHandler.java

Hi @renatoh - I think another merging in of the latest `origin/main` is needed to include the https://github.com/apache/solr/commit/3cbb97549cef1664a1d6b01fd2d1367cefbb0665 change in the https://github.com/apache/solr/commits/main commit history.

renatoh:
@cpoerschke I did that, and it shows me that my fork is up to date, but I am getting:
(#error: patch failed: solr/core/src/java/org/apache/solr/cli/AssertTool.java:35

cpoerschke:
> ... (#error: patch failed: solr/core/src/java/org/apache/solr/cli/AssertTool.java:35

Hmm, I saw on the dev-list some thread(s) w.r.t. the crave CI but didn't follow too closely about the details on what the current advice for handling it is. Sorry about that.

renatoh:
@cpoerschke do you know why patches are applied  e.g. "Applying: SOLR-16871: Synchronize on a larger block to avoid race condition in CoordinatorHttpSolrCall init" as a part of the build?

renatoh:
@cpoerschke I did another merge of origin/main, it looks like the checks have now passed, or haven't they?

renatoh:
@epugh is it possible that you have a look at tell me what's wrong with my commit? I don't what the issue is

cpoerschke:
https://lists.apache.org/thread/ybmp5xz3c9549opqgktdb4cxlpmdmwy4 (I think) is the dev list thread w.r.t. Crave.io CI patch application.

renatoh:
@epugh Hey Eric, could you please give me a hand with this PR? I do not understand what the issue is. 
Thanks

PR #1878, opened 2023-09-01
SOLR-16958: Fix spurious warning about LATEST luceneMatchVersion

colvinco:
https://issues.apache.org/jira/browse/SOLR-16958

Switch back to the use of `==` to check whether the supplied version is `LATEST` or the actual latest version (e.g. `9.7.0`).


colvinco:
Accidentally pushed to  main. I've reverted the commits on main and I'll create a new PR.

PR #1880, opened 2023-09-01
SOLR-16956: fix not being able to stop a solr by port number

epugh:
https://issues.apache.org/jira/browse/SOLR-16956
# Description

In doing some other testing, I noticed that `bin/solr stop -p 7574` didn't work, I get back the error message:

```
Found 2 Solr nodes running! Must either specify a port using -p or -all to stop all Solr nodes on this host.
```

# Solution

Set the PROVIDED_SOLR_PORT to the SOLR_PORT.

# Tests

extended bats tests

### DIFF
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -1301,7 +1301,7 @@ if [ $# -gt 0 ]; then
               exit 1
             fi
             SOLR_PORT="$2"
-            PROVIDED_SOLR_PORT=SOLR_PORT
+            PROVIDED_SOLR_PORT="${SOLR_PORT}"
             PASS_TO_RUN_EXAMPLE+=("-p" "$SOLR_PORT")
             shift 2
         ;;
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -1301,6 +1301,7 @@ if [ $# -gt 0 ]; then
               exit 1
             fi
             SOLR_PORT="$2"
+            PROVIDED_SOLR_PORT=SOLR_PORT
             PASS_TO_RUN_EXAMPLE+=("-p" "$SOLR_PORT")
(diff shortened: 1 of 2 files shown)


epugh:
@HoustonPutman  didn't you recently look at some port stuff?   COuld you review this bug fix?

epugh:
I tested on Branch_9x, and this issue is only on Main.....

solr/bin/solr
```diff
@@ -1301,6 +1301,7 @@ if [ $# -gt 0 ]; then
               exit 1
             fi
             SOLR_PORT="$2"
+            PROVIDED_SOLR_PORT=SOLR_PORT
```
cpoerschke, line 1304: Did you mean `${SOLR_PORT}` here instead?
```suggestion
            PROVIDED_SOLR_PORT="${SOLR_PORT}"
```

epugh:
Thanks @cpoerschke  for the feedback...  one thing I noticed in the BATS docs is that it suggests running shellcheck.   That isn't actually something I've ever done, and I wonder if it's the type of thing that should be run as part of precommit to make sure we keep both our bin/solr script and the various bats tests following best practices?

mkhludnev (approved):

cpoerschke:
> ... it suggests running shellcheck. That isn't actually something I've ever done, and I wonder if it's the type of thing that should be run as part of precommit ...

I've not yet run shellcheck either but running as part of precommit too sounds like a good idea.

epugh:
I am going to merge this without an entry in Changes.txt since it's a bugfix in 10, that doesn't appear to exist in 9...

HoustonPutman (commented):
Thanks for fixing this!

PR #1884, opened 2023-09-01
SOLR-16955 Tracing v2 apis breaks SecurityConfHandler

stillalex:
https://issues.apache.org/jira/browse/SOLR-16955

# Description

V2HttpCall reads the request stream to determine a better name for the tracing span. this is problematic as other code might bypass the operation cache. I am proposing to remove this part and rely on the method name.
Optionally we can add an attrib for the operation that can be set inside the Security handler to reflect what is being processed. - this is not included in the current PR.

# Solution

# Tests

# Checklist


gerlowskija (approved):
LGTM

solr/modules/opentelemetry/src/test/org/apache/solr/opentelemetry/BasicAuthIntegrationTracingTest.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /** See SOLR-16955 */
+  @Test
+  public void testSetupBasicAuth() throws Exception {
+    CloudSolrClient cloudClient = cluster.getSolrClient();
+    V2Request req =
+        new V2Request.Builder("/cluster/security/authentication")
+            .withMethod(SolrRequest.METHOD.POST)
+            .withPayload("{\"set-user\": {\"harry\":\"HarryIsCool\"}}")
```
dsmiley, line 98: what; no Map.of/toJSON?

dsmiley (approved):
Lovely test.

stillalex:
poking some more at the test I found the following interesting bit: the 'db.user' does a Principal#toString resulting in the following. this might be too much info to share in the telemetry data, what do others think?
```
db.user=BasicAuthPlugin [username=solr,pwd=*****]
```

solr/modules/opentelemetry/src/test/org/apache/solr/opentelemetry/BasicAuthIntegrationTracingTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+  private static final String COLLECTION = "collection1";
+  private static final String USER = "solr";
+  private static final String PASS = "SolrRocksAgain";
+  private static final String SECURITY_JSON =
+      Utils.toJSONString(
+          Map.of(
+              "authorization",
+              Map.of(
+                  "class",
```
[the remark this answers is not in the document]
stillalex: as much as I would love to say I came up with this, the truth is I lifted it from some existing Solr core test. but I agree it looks good hehe.

stillalex:
more changes!
 - added 'ops' tag (name can be changed of course). SecurityConfHandler will publish the list of operations as a list.
 - reduced user tag data to principal name only. I am ok to revert back to the old form if this poses an issue
 - updated the test to check the above

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
```diff
@@ -239,7 +239,13 @@ private void dispatch(
...
         log.debug("User principal: {}", request.getUserPrincipal());
       }
-      TraceUtils.setUser(span, String.valueOf(request.getUserPrincipal()));
+      final String principalName;
+      if (request.getUserPrincipal() != null) {
+        principalName = request.getUserPrincipal().getName();
+      } else {
+        principalName = null;
+      }
+      TraceUtils.setUser(span, String.valueOf(principalName));
```
dsmiley, line 248: I think we should be setting no "user" at all if there is no user principal

solr/modules/opentelemetry/src/test/org/apache/solr/opentelemetry/TestDistributedTracing.java
```diff
@@ -201,7 +201,7 @@ private void assertOneSpanIsChildOfAnother(List<SpanData> finishedSpans) {
     assertEquals(child.getTraceId(), parent.getTraceId());
   }
 
-  private List<SpanData> getAndClearSpans() {
+  static List<SpanData> getAndClearSpans() {
```
dsmiley, line 204: This is a hack, calling some method on some other test to manage global lifecycle of some singleton mock tracer.  Seems like we might ideally have a JUnit Rule to manage trace listening, where we might have this method?

solr/modules/opentelemetry/src/test/org/apache/solr/opentelemetry/BasicAuthIntegrationTracingTest.java
```diff
@@ -99,5 +104,12 @@ public void testSetupBasicAuth() throws Exception {
             .build();
     req.setBasicAuthCredentials(USER, PASS);
     assertEquals(0, req.process(cloudClient, COLLECTION).getStatus());
+
+    var finishedSpans = getAndClearSpans();
+    assertEquals(1, finishedSpans.size());
+    var span = finishedSpans.get(0);
+    assertEquals("post:/cluster/security/authentication", span.getName());
+    assertEquals("solr", span.getAttributes().get(TraceUtils.TAG_USER));
+    assertEquals(List.of("set-user"), span.getAttributes().get(TraceUtils.TAG_OPS));
```
dsmiley, line 113: Is it not possible to set the operation name later on with OTEL?  The action/operation is a huge part of what the span *is*; deserves to be a part of the operation name if we can.

stillalex: could you explain a bit better what you expect to see in the operation name? I thought having `post:/cluster/security/authentication` is in line with all other apis.
also one part that is different here is that it's not a single op, there is a list in the body, so I went with a list attribute and the previous approach of taking the first entry to set as a name seems fragile. I am open to suggestions on what we could do.

dsmiley: Perhaps `op1,op2,op3:post:/cluster/security/authentication` ?  AFAIK there is no standard here, so we can propose whatever syntax we think is useful.

An alternative would be to leave this be but always create another span at a lower level immediately in front of doing the operation.  At least then there would be a span per operation!

stillalex: re. `op1,op2,op3` I would avoid having the op name dynamic and of unknown length. it makes its usefulness in searching and grouping almost zero.

> An alternative would be to leave this be but always create another span at a lower level immediately in front of doing the operation. At least then there would be a span per operation!

this one makes more sense. will give it a go to see how it looks, but the added complexity of wrapping each operation in a span might be overkill for this part of the code. I see 2 aspects here: performance and auditability and I'm hard pressed to make a good argument for either at this level. but maybe I am missing something.

stillalex: ugh. spoke too soon. you can't break the ops apart, they are passed in as a list to the impl and processed in a loop. so I'd have to update each plugin separately to have one span per op.
as an aside, one thing I realized would be useful here is to have the plugin class that processed the list as an attrib.

dsmiley: Let's just keep the list then.  I suppose this is rather specific to this plugin; I am more interested in general/broad cases.

stillalex: I agree with the sentiment. this is a good opportunity to evaluate closely the patterns and make sure things are lean and clean. if we start adding otel data in more and more places, there is a real risk of attributes explosion.

stillalex:
thank you @dsmiley for the continued review, I am getting ready to merge.
one more aspect of this is the backport to 9.x. it will be extremely complicated to take the test as is. one easy option is to not backport the test, not ideal, but way less work. another option is to try to rewrite the test in the context of 9.x. this will be very involved and I am wondering if the risk of regressions is high enough to warrant the trouble.

dsmiley:
I'm very sympathetic to the pain of maintaining tests of distributed tracing!  Okay by me to skip such a test in 9x.  We don't test details of our logs... yet we do with tracing... :shrug:

stillalex:
I sat on this a bit and realized I would prefer adding a test to 9.x backport and ended up rewriting it so it fits with 9.x constructs. makes me feel a tiny bit more confident that this remains working

PR #1886, opened 2023-09-02
SOLR-16654: Improve test

dsmiley:
* remove 2 test config files
* revert "solr.tests.loadSolrXml" option recently added
* use EmbeddedSolrServerTestRule for initialization

https://issues.apache.org/jira/browse/SOLR-16654


solr/core/src/test/org/apache/solr/search/TestThinCache.java
```diff
@@ -26,13 +26,63 @@
...
 import org.apache.solr.util.stats.MetricUtils;
 import org.junit.BeforeClass;
+import org.junit.ClassRule;
 import org.junit.Test;
 
 /** Test for {@link ThinCache}. */
 public class TestThinCache extends SolrTestCaseJ4 {
 
+  @ClassRule public static EmbeddedSolrServerTestRule solrRule = new EmbeddedSolrServerTestRule();
+  public static final String SOLR_NODE_LEVEL_CACHE_XML =
```
dsmiley, line 40: String literal (vs a separate file) keeps the test with the config its testing.  Constant field allows it to be used in the "TestSolrXml" test too.
Looking forward to Java 17 where we could use multi-line String syntax!  (maybe worth requiring Java 17 for tests?)

solr/core/src/test/org/apache/solr/search/TestThinCache.java
```diff
@@ -26,13 +26,63 @@
...
+import org.apache.solr.util.TestHarness;
 import org.apache.solr.util.stats.MetricUtils;
 import org.junit.BeforeClass;
+import org.junit.ClassRule;
 import org.junit.Test;
 
 /** Test for {@link ThinCache}. */
 public class TestThinCache extends SolrTestCaseJ4 {
 
+  @ClassRule public static EmbeddedSolrServerTestRule solrRule = new EmbeddedSolrServerTestRule();
```
dsmiley, line 39: This is fairly new; there will be more rules available in time + randomization but not yet.

solr/core/src/test/org/apache/solr/search/TestThinCache.java
```diff
@@ -26,13 +26,63 @@
...
+    solrRule.startSolr(home);
+
+    Path configSet = createTempDir("configSet");
+    copyMinConf(configSet.toFile());
+    // insert a special filterCache configuration
+    Path solrConfig = configSet.resolve("conf/solrconfig.xml");
+    Files.writeString(
+        solrConfig,
+        Files.readString(solrConfig)
+            .replace(
```
dsmiley, line 69: This is a bit hacky I admit.  I would like to add a utility method that might look something like
 `updateXml(Path, String xpath, String newXml)` where the Xpath would be like "/config/query/filterCache".  ChatGPT wrote one for me :-)

dsmiley: When possible, the Config API can be used as is done in `org.apache.solr.search.json.TestJsonRequestWithEdismaxDefType#test`.  But "parentCacheName" isn't supported / is too exotic.  I hate that the Config API is a custom format and only partially implements what we can specify in XML.  I'm torn between embracing it and pretending it doesn't exist.

solr/core/src/test/org/apache/solr/search/TestThinCache.java
```diff
@@ -26,13 +26,63 @@
...
+                "</config>",
+                "<query>\n"
+                    + "<filterCache\n"
+                    + "      class=\"solr.ThinCache\"\n"
+                    + "      parentCacheName=\"myNodeLevelCacheThin\"\n"
+                    + "      size=\"5\"\n"
+                    + "      initialSize=\"5\"/>\n"
+                    + "</query></config>"));
+
+    solrRule.newCollection().withConfigSet(configSet.toString()).create();
```
dsmiley, line 79: the rule uses collection terminology even if it's really a core.  The point is consistent terminology.  Sometimes I think we should go the other way (standardize on "core") unless collection specific options are needed.  Shrug.

solr/core/src/test/org/apache/solr/search/TestThinCache.java
```diff
@@ -126,35 +176,4 @@ public void testInitCore() throws Exception {
     // for the other node-level cache, simply check that metrics are accessible
     assertEquals(0, nodeMetricsSnapshot.get("CACHE.nodeLevelCache/myNodeLevelCache.size"));
   }
-
-  @BeforeClass
```
dsmiley, line 130: Please never put lifecycle methods at the bottom!

solr/core/src/test/org/apache/solr/search/TestThinCache.java
```diff
@@ -26,13 +26,63 @@
...
+                    + "      parentCacheName=\"myNodeLevelCacheThin\"\n"
+                    + "      size=\"5\"\n"
+                    + "      initialSize=\"5\"/>\n"
+                    + "</query></config>"));
+
+    solrRule.newCollection().withConfigSet(configSet.toString()).create();
+
+    // legacy; get rid of this someday!
+    h = new TestHarness(solrRule.getCoreContainer());
+    lrf = h.getRequestFactory("/select", 0, 20);
```
dsmiley, line 83: I need to refresh the vision and discuss/debate it.  Your test here, like most Solr tests, uses `assertU`, `assertQ`, `req`, `doc`, which are implemented today on XML strings and uses TestHarness.  But the vision is to embrace SolrClient.  Nonetheless we need convenience methods to assert via XPath.  I could see a new base test class emerging that has methods named like this and that which are called with the same idioms yet actually have different return types  (definitely not XML strings).

dsmiley (commented):
Again, the SOLR-11872 Ditch TestHarness vision is incomplete and we need convenience utilities not yet in place but we can nonetheless reduce the proliferation of test configuration files as I show here.

solr/core/src/test/org/apache/solr/search/TestThinCache.java
epugh: I like seeing xml next ot the test too, helps me understand what is going on.  I use the tests as a "reference" to figure things out a lot.

solr/core/src/test/org/apache/solr/search/TestThinCache.java
epugh: Could we have patterns like that caught in precommit or lint checker?    These are the kinds of things that should be checked automatically for us.   not putting lifecycle methods at the bottom isn't an actual pattern that I knew about for example ;-).  We need more support in keeping formatting clean.

epugh (approved):
This does look nicer!   It would be great to figure out how to rally folks to make the conversion happen faster..   I hate that we still have SolrTestCase and SolrTestCaseHS for example...     I know some folks use linting tools that kick in when a file changes to fail the build.  So if you touch a test, well then you fail the build until it's migrated to these newer patterns..

magibney (approved):
Thanks for addressing this, and thanks especially for the extra explanation around what was done, and why.

solr/core/src/test/org/apache/solr/search/TestThinCache.java
stillalex: @dsmiley shouldn't this have also been `final`?

dsmiley: I suppose so.

PR #1896, opened 2023-09-07
SOLR-16954: Fix test failure in UPDATE circuit breaker

janhoy:
https://issues.apache.org/jira/browse/SOLR-16954

The tests ran fine both on localhost and github actions, but on Jenkins this one tripped during indexing content in `indexDocs()`. Must be some difference in memory handling, perhaps how many tests are ran in parallel in Jeknins vs GitHub actions/Crave, or the `-Xmx` setting??

Output from [failing test](https://ci-builds.apache.org/job/Solr/job/Solr-Check-main/7905/testReport/org.apache.solr.util/TestCircuitBreaker/classMethod/):
> Seen JVM heap memory usage 383.848.016 and allocated threshold 383.778.816

383Mb threshold suggests that the VM is ran with 512Mb RAM, which makes sense.

### DIFF
--- a/solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
+++ b/solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
@@ -89,7 +89,7 @@ public class CPUCircuitBreaker extends CircuitBreaker {
 
   @Override
   public String getErrorMessage() {
-    return "CPU Circuit Breaker triggered as seen CPU usage is above allowed threshold."
+    return "CPU Circuit Breaker triggered as seen CPU usage is above allowed threshold. "
         + "Seen CPU usage "
         + seenCPUUsage.get()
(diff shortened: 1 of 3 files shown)


janhoy:
I guess this failing test also begs the question whether the memory breaker is too trigger happy, tripping for a spike during a ms memory pressure. If we had an average memory pressure over N seconds that would be so much better. Does not dropwizard provide such, or would we need to roll our own?

cpoerschke (approved):

janhoy:
I think the intention of including the update breaker in the config was to exercise the initialization. So another approach could be to keep it but increase the threshold to 99% or something...

cpoerschke:
> ... keep it but increase the threshold to 99% or something...

I like that idea.

And if it still causes regular issues at 99% or so threshold then removal is still an option. Or then another option could be to sub-class the update handler, configure the sub-classed update handler and have it ignore the circuit breaker when some magic field is part of the test setup update requests.

PR #1899, opened 2023-09-07
SOLR-16966: Add a first-class cache for OrdinalMaps

magibney:
See: SOLR-16966


magibney:
Tests are not built out yet; I wanted first to solicit feedback on the general approach, particularly around allowing entries to age out by tracking access time in cache entry metadata.

magibney:
I considered adding the `regenKeepAlive` param to CaffeineCache more generally, but decided against it because:
1. it makes things somewhat more complicated, and
2. the problem that it addresses is quite particular to relatively small, low-turnover caches -- i.e., ordMapCache.

I'm hoping to commit this soon -- any feedback welcome!

magibney:
Proposed `CHANGES.txt` entry under `Improvements` for 9.4.0 release:
```
* SOLR-16966: Add a first-class cache for OrdinalMaps (Michael Gibney)
```

I plan to commit this early next week, pending feedback.

magibney:
The last three commits (through  b8124677d7c88b903b5f3c4fd24ca8a54596d215) introduce a subtle but substantial change that abstracts all the "metadata wrapper" accounting so that application code (SolrIndexSearcher, other users of caches) know nothing about any potential metadata wrappers around cache value entries.

The cleanest place to do this is from the regenerator; I've added `wrap()` and `unwrap()` methods to `CacheRegenerator`, with a default passthrough implementation. Regenerators that require value-entry metadata accounting (such as access timestamp, individual entry hit count, etc.) may override `wrap()` to return a simple `SolrCache<K, V>` view of an underlying `SolrCache<K, MetadataWrapper<V>>` cache. It is the _raw_ cache (unwrapped) that is autowarmed, providing the regenerator with the information it needs to make warming decisions, etc. But it is the _wrapped_ cache that is presented for all normal cache interactions.

This approach works equally well for implementing more nuanced cache metrics (histograms, dumping cache contents, etc.). The overall approach is described in the [javadocs here](https://github.com/apache/solr/blob/b8124677d7c88b903b5f3c4fd24ca8a54596d215/solr/core/src/java/org/apache/solr/search/MetaCacheRegenerator.java#L36-L62). It should be quite trivial to implement regenerators that  implement custom metrics.

PR #1905, opened 2023-09-08
SOLR-16968: The MemoryCircuitBreaker now uses average heap usage

janhoy:
https://issues.apache.org/jira/browse/SOLR-16968

Introduced a new utility class `AveragingMetricProvider` that will cache the average for a `double` value provided by the breaker. Uses a scheduled executor every 5 seconds, which is shared by all instances of this class. This same utility can be used by other CBs to achieve the same, but I started with `MemoryCircuitBreaker`.

As this requires some cleanup of executor threads, CircuitBreaker now implements `Closeable` and the registry will close all CBs on close, which is also called from `CoreContainer#doClose`. I also added some synchonization on the map, which isn't really necessary in normal runtime, but I figured useful for tests.

*NOTE* that this may conflict with other in-flight CB PRs...

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -1088,9 +1088,6 @@ public class SolrCore implements SolrInfoBean, Closeable {
       solrMetricsContext = coreMetricManager.getSolrMetricsContext();
       this.coreMetricManager.loadReporters();
 
-      // init pluggable circuit breakers
-      initPlugins(null, CircuitBreaker.class);
-
       if (updateHandler == null) {
         directoryFactory = initDirectoryFactory();
         recoveryStrategyBuilder = initRecoveryStrategyBuilder();
@@ -1114,6 +1111,9 @@ public class SolrCore implements SolrInfoBean, Closeable {
 
       // initialize core metrics
       initializeMetrics(solrMetricsContext, null);
+
+      // init pluggable circuit breakers, after metrics because some circuit breakers use metrics
+      initPlugins(null, CircuitBreaker.class);
 
       SolrFieldCacheBean solrFieldCacheBean = new SolrFieldCacheBean();
       // this is registered at the CONTAINER level because it's not core-specific - for now we
@@ -1763,6 +1763,17 @@ public class SolrCore implements SolrInfoBean, Closeable {
     log.info("CLOSING SolrCore {}", this);
 
     ExecutorUtil.shutdownAndAwaitTermination(coreAsyncTaskExecutor);
+
+    // Close circuit breakers that may have background threads, before metrics because some circuit
+    // breakers use metrics
+    try {
+      getCircuitBreakerRegistry().close();
+    } catch (Throwable e) {
+      log.error("Exception closing circuit breakers", e);
+      if (e instanceof Error) {
+        throw (Error) e;
+      }
+    }
 
     // stop reporting metrics
     try {
--- a/solr/core/src/java/org/apache/solr/util/circuitbreaker/AveragingMetricProvider.java
+++ b/solr/core/src/java/org/apache/solr/util/circuitbreaker/AveragingMetricProvider.java
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.util.circuitbreaker;
+
+import com.google.common.util.concurrent.AtomicDouble;
+import java.io.Closeable;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.apache.solr.common.util.ExecutorUtil;
+import org.apache.solr.common.util.SolrNamedThreadFactory;
+import org.apache.solr.logging.CircularList;
+
+/** Averages the metric value over a period of time */
+public class AveragingMetricProvider implements Closeable {
+  private final CircularList<Double> samplesRingBuffer;
+  private ScheduledExecutorService executor;
+  private final AtomicDouble currentAverageValue = new AtomicDouble(-1);
+
+  /**
+   * Creates an instance with an executor that runs every sampleInterval seconds and averages over
+   * numSamples samples.
+   *
+   * @param metricProvider metric provider that will provide a value
+   * @param numSamples number of samples to calculate average for
+   * @param sampleInterval interval between each sample
+   */
+  public AveragingMetricProvider(
+      MetricProvider metricProvider, int numSamples, long sampleInterval) {
+    this.samplesRingBuffer = new CircularList<>(numSamples);
+    executor =
+        Executors.newSingleThreadScheduledExecutor(
+            new SolrNamedThreadFactory(
+                "AveragingMetricProvider-" + metricProvider.getClass().getSimpleName()));
+    executor.scheduleWithFixedDelay(
+        () -> {
+          samplesRingBuffer.add(metricProvider.getMetricValue());
+          currentAverageValue.set(
+              samplesRingBuffer.toList().stream()
+                  .mapToDouble(Double::doubleValue)
+                  .average()
+                  .orElse(-1));
+        },
+        0,
+        sampleInterval,
+        TimeUnit.SECONDS);
+  }
+
+  /**
+   * Return current average. This is a cached value, so calling this method will not incur any
+   * calculations
+   */
+  public double getMetricValue() {
+    return currentAverageValue.get();
+  }
+
+  @Override
+  public void close() {
+    ExecutorUtil.shutdownAndAwaitTermination(executor);
+  }
+
+  /** Interface to provide the metric value. */
+  public interface MetricProvider {
+    double getMetricValue();
+  }
+}
--- a/solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerManager.java
+++ b/solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerManager.java
@@ -17,6 +17,7 @@
 
 package org.apache.solr.util.circuitbreaker;
 
+import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import org.apache.solr.common.util.NamedList;
 import org.slf4j.Logger;
@@ -73,6 +74,19 @@ public class CircuitBreakerManager extends CircuitBreaker {
     if (cpuEnabled) {
       cpuCB = new CPUCircuitBreaker();
       cpuCB.setThreshold(cpuThreshold);
+    }
+  }
+
+  @Override
+  public void close() throws IOException {
+    try {
+      if (memEnabled) {
+        memCB.close();
+      }
+    } finally {
+      if (cpuEnabled) {
+        cpuCB.close();
+      }
     }
   }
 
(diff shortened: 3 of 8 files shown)


solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -100,17 +123,11 @@ private long getCurrentMemoryThreshold() {
...
-    // NOTE: MemoryUsageGaugeSet provides memory usage statistics but we do not use them
-    // here since it will require extra allocations and incur cost, hence it is cheaper to use
-    // MemoryMXBean directly. Ideally, this call should not add noticeable
-    // latency to a query -- but if it does, please signify on SOLR-14588
-    return MEMORY_MX_BEAN.getHeapMemoryUsage().getUsed();
+  @Override
+  public void close() throws IOException {
+    if (averagingMetricProvider != null) {
+      averagingMetricProvider.close();
+      averagingMetricProvider = null;
```
janhoy, line 130: In case you register multiple Mem CBs, the first instance will setup the executor and the first one closed will close the common executor. Ideally the last instance closed should close the executor. Imagine two cores, both setting up such a CB, then you delete one of the cores, which will trigger close(). The other core will now have a stale value since the background thread updating it is no longer running.

Guess this is the downside of a shared object. Need to introduce some ref-counting or something?

solr/core/src/test/org/apache/solr/util/BaseTestCircuitBreaker.java
```diff
@@ -59,11 +62,17 @@ protected static void indexDocs() {
...
   }
 
   @After
   public void after() {
-    h.getCore().getCircuitBreakerRegistry().deregisterAll();
+    try {
+      h.getCore().getCircuitBreakerRegistry().deregisterAll();
+    } catch (IOException e) {
+      fail("Failed to unload circuit breakers");
+    }
```
cpoerschke, line 75:
```suggestion
    removeAllExistingCircuitBreakers();
```

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -43,8 +46,34 @@ public class MemoryCircuitBreaker extends CircuitBreaker {
...
+   * Constructor that allows override of sample interval for which the memory usage is fetched. This
+   * is provided for testing, not intended for general use.
+   *
+   * @param numSamples number of samples to calculate average for
+   * @param sampleInterval interval between each sample
+   */
+  protected MemoryCircuitBreaker(int numSamples, int sampleInterval) {
     super();
+    if (averagingMetricProvider == null || averagingMetricProvider.getRefcount() == 0) {
+      averagingMetricProvider =
```
cpoerschke, line 64: Wondering how this would work if two threads concurrently instantiated a circuit breaker e.g. two shards in the same JVM. Might an alternative be to

```
private static RefCounted<AveragingMetricProvider> averagingMetricProvider = ...
```

initialise always with the test constructor doing a replacement?

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
```diff
@@ -43,8 +46,34 @@ public class MemoryCircuitBreaker extends CircuitBreaker {
...
   private static final ThreadLocal<Long> allowedMemory = ThreadLocal.withInitial(() -> 0L);
 
+  /** Creates an instance which averages over 6 samples during last 30 seconds. */
   public MemoryCircuitBreaker() {
+    this(6, 5);
+  }
+
+  /**
+   * Constructor that allows override of sample interval for which the memory usage is fetched. This
+   * is provided for testing, not intended for general use.
```
cpoerschke, line 56:
```suggestion
   * is provided for testing, not intended for general use because the average metric provider implementation is the same for all instances of the class.
```

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerRegistry.java
```diff
@@ -97,4 +112,34 @@ public static String toErrorMessage(List<CircuitBreaker> circuitBreakerList) {
...
+                          }
+                          it.close();
+                        } catch (IOException e) {
+                          log.error("Failed to close circuit breaker", e);
+                          closeFailed.set(true);
+                        }
+                      }));
+      circuitBreakerMap.clear();
+      if (closeFailed.get()) {
+        throw new IOException("Failed to close one or more circuit breakers");
```
cpoerschke, line 141: minor: swapping out of the atomic boolean for integer could allow specific wording.
```suggestion
        throw new IOException("Failed to close " + closeFailedCounter.get() + " circuit breakers");
```

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerRegistry.java
```diff
@@ -97,4 +112,34 @@ public static String toErrorMessage(List<CircuitBreaker> circuitBreakerList) {
...
+                        try {
+                          if (log.isDebugEnabled()) {
+                            log.debug(
+                                "Closed circuit breaker {} for request type(s) {}",
+                                it.getClass().getSimpleName(),
+                                it.getRequestTypes());
+                          }
+                          it.close();
+                        } catch (IOException e) {
+                          log.error("Failed to close circuit breaker", e);
```
cpoerschke, line 135:
```suggestion
                          log.error("Failed to close circuit breaker " + it.getClass().getSimpleName(), e);
```

or

```suggestion
                          log.error("Failed to close circuit breaker " + it, e);
```

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerManager.java
```diff
@@ -91,6 +92,16 @@ public void init(NamedList<?> args) {
...
   }
 
+  @Override
+  public void close() throws IOException {
+    if (memEnabled) {
+      memCB.close();
+    }
+    if (cpuEnabled) {
+      cpuCB.close();
+    }
```
cpoerschke, line 102: edge case: `memCB.close()` throwing an exception and `cpuCB.close()` not getting called as a result. might not be worth extra complexity of handling since class is deprecated?

solr/core/src/java/org/apache/solr/util/circuitbreaker/AveragingMetricProvider.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.common.util.concurrent.AtomicDouble;
+import java.io.Closeable;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.apache.solr.common.util.SolrNamedThreadFactory;
+import org.apache.solr.logging.CircularList;
+
+/** Averages the metric value over a period of time */
+public class AveragingMetricProvider implements Closeable {
```
cpoerschke, line 28: Am surprised such a class doesn't yet exist.

solr/core/src/java/org/apache/solr/util/circuitbreaker/AveragingMetricProvider.java
```diff
@@ -0,0 +1,86 @@
...
+  @Override
+  public void close() {
+    try {
+      executor.shutdown();
+      if (!executor.awaitTermination(10, TimeUnit.SECONDS)) {
+        executor.shutdownNow();
+      }
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+    }
```
cpoerschke, line 79: Vaguely recall seeing some "shutdown util" or similar wrappers which do some things as a one function call, might not match exactly the logic here, haven't looked into.

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -1115,6 +1112,9 @@ private SolrCore(
       // initialize core metrics
       initializeMetrics(solrMetricsContext, null);
 
+      // init pluggable circuit breakers
```
cpoerschke, line 1115:
```suggestion
      // init pluggable circuit breakers, after metrics because some circuit breakers use metrics
```

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -1764,6 +1764,16 @@ private void doClose() {
 
     ExecutorUtil.shutdownAndAwaitTermination(coreAsyncTaskExecutor);
 
+    // Close circuit breakers that may have background threads
```
cpoerschke, line 1767:
```suggestion
    // Close circuit breakers that may have background threads, before metrics because some circuit breakers use metrics
```

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
janhoy: It's my first attempt using `RefCounted` API, so I may be holding it wrong.

janhoy:
@atris feel free to weigh in here. The intention is to get fewer "false positive" request rejections due to short spikes. Now there will be some delay (0-30s) from heavy memory pressure starts until the breaker starts tripping, and likewise some time before we allow requests again after cooling down. So if the client implements some kind of exponential backoff logic on receipt of HTTP 429 errors, the system will stabilize right under the memory threshold. If they don't, it will be an oscillating pattern on and off, but hopefully give Solr time to recover in between.

solr/core/src/java/org/apache/solr/util/circuitbreaker/AveragingMetricProvider.java
janhoy: I had a good look around Solr metrics classes and dropwizard. Even asked ChatGPT, and ended up with custom impl. There may be some commons-lib somewhere that does similar generic aggregation..?

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerManager.java
janhoy: Done

solr/core/src/java/org/apache/solr/util/circuitbreaker/AveragingMetricProvider.java
janhoy: So I found https://github.com/apache/solr/blob/main/solr/solrj/src/java/org/apache/solr/common/util/ExecutorUtil.java#L104 which is probably what you were thinking of. Will use it...

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
cpoerschke: No worries, I haven't yet used the `RefCounted` API either, perhaps something for me to learn about next week.

As an aside, just noticed the TODO in the `isTripped()` comments below and it sounds like the changes in this PR will have an effect similar to what is described there, maybe.

janhoy: Yes, let's remove that TODO, as averaging will have about the same effect. Also the isTripped method is now almost "free", as it just pulls the cached average.

solr/core/src/java/org/apache/solr/util/circuitbreaker/AveragingMetricProvider.java
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Creates an instance with an executor that runs every sampleInterval seconds and averages over
+   * numSamples samples.
+   *
+   * @param metricProvider metric provider that will provide a value
+   * @param numSamples number of samples to calculate average for
+   * @param sampleInterval interval between each sample
+   */
+  public AveragingMetricProvider(
```
sigram, line 42: Perhaps this could be replaced with one of the already existing implementations for EWMAs in the Dropwizard metrics (or metrics that use them, such as Meter).

janhoy: I had a look at Meter, but it seems very event/count oriented, not for smoothing out gauge values? Or can you show examples of such use somewhere?

sigram: Hmm, not sure what you mean by event/count oriented... it does the counting, true, but it also keeps track of `ExponentialMovingAverages`, which in turn uses a bunch of EWMAs. "Rate" in its terminology is exactly the moving average of values that are recorded into it by `EWMA.update(n)` method (or `Meter.mark(n)`)

janhoy: Had a look, but the `getRate(TimeUnit rateUnit)` implies events/second, so this API feels like the wrong tool. We're not measuring events per second, but average of a gauge over time. I'm sure it could be shoehorned in, but then I don't see a big gain compared to the simple ring buffer?

solr/core/src/java/org/apache/solr/util/circuitbreaker/MemoryCircuitBreaker.java
janhoy: So I made both constructor and close methods synchronized to safeguard against this.

janhoy:
Planning to merge this on friday...

PR #1912, opened 2023-09-11
Add full mTLS integration tests for Solr

HoustonPutman:
This adds full mTLS integration tests (WANT_CLIENT_AUTH and NEED_CLIENT_AUTH). The client/server certs are different, as are the truststore and keystore, so each is tested independently.

One test is broken, which will be fixed by https://issues.apache.org/jira/browse/SOLR-16970, so cannot merge this yet.

### DIFF
--- a/solr/packaging/test/test_ssl.bats
+++ b/solr/packaging/test/test_ssl.bats
@@ -48,10 +48,7 @@ teardown() {
   export SOLR_SSL_TRUST_STORE_PASSWORD=secret
   export SOLR_SSL_NEED_CLIENT_AUTH=false
   export SOLR_SSL_WANT_CLIENT_AUTH=false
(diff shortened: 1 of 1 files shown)


epugh:
I love it...     I wonder though if we should think about having a explicit "integration" tests subdir?   We use BATS for both "unit testing" our Bash scripts, but also now are using it more and more to run a integration or End 2 End test for Solr.   (For example, my adding a test to verify stand alone Solr replication!)...    Maybe for another ticket if we find that these bats tests start taking a REALLY long time.

tflobbe (approved):
Thanks @HoustonPutman, these are great tests. One thing I think would be useful would be to have at least an `add` and replicas in the collection that's created, to test the update side too (which as of today uses the Apache client, so a different code path). I don't think this should block this PR, since this is already a big improvement.

PR #1919, opened 2023-09-13
SOLR-16974: Global Circuit Breakers

janhoy:
https://issues.apache.org/jira/browse/SOLR-16974

Proof-of-concept with static map of global breakers.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -246,7 +246,7 @@ public class SolrCore implements SolrInfoBean, Closeable {
   private final ConfigSet configSet;
   // singleton listener for all packages used in schema
 
-  private final CircuitBreakerRegistry circuitBreakerRegistry = new CircuitBreakerRegistry();
+  private final CircuitBreakerRegistry circuitBreakerRegistry;
 
   private final List<Runnable> confListeners = new CopyOnWriteArrayList<>();
 
@@ -1072,6 +1072,7 @@ public class SolrCore implements SolrInfoBean, Closeable {
     final CountDownLatch latch = new CountDownLatch(1);
     try {
       this.coreContainer = coreContainer;
+      this.circuitBreakerRegistry = new CircuitBreakerRegistry(coreContainer);
       this.configSet = configSet;
       this.coreDescriptor = Objects.requireNonNull(coreDescriptor, "coreDescriptor cannot be null");
       this.name = Objects.requireNonNull(coreDescriptor.getName());
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -3179,6 +3179,9 @@ public class SolrCore implements SolrInfoBean, Closeable {
             type.getSimpleName() + "." + info.name, (SolrMetricProducer) o);
       }
       if (o instanceof CircuitBreaker) {
+        if (o instanceof SolrCoreAware) {
+          ((SolrCoreAware) o).inform(this);
+        }
         circuitBreakerRegistry.register((CircuitBreaker) o);
       }
       if (info.isDefault()) {
--- a/solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
@@ -72,6 +72,7 @@ import org.apache.solr.schema.ManagedIndexSchemaFactory;
 import org.apache.solr.schema.SimilarityFactory;
 import org.apache.solr.search.QParserPlugin;
 import org.apache.solr.update.processor.UpdateRequestProcessorFactory;
+import org.apache.solr.util.circuitbreaker.CircuitBreaker;
 import org.apache.solr.util.plugin.SolrCoreAware;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -814,6 +815,7 @@ public class SolrResourceLoader
         new Class<?>[] {
           // DO NOT ADD THINGS TO THIS LIST -- ESPECIALLY THINGS THAT CAN BE CREATED DYNAMICALLY
           // VIA RUNTIME APIS -- UNTIL CAREFULLY CONSIDERING THE ISSUES MENTIONED IN SOLR-8311
+          CircuitBreaker.class,
           CodecFactory.class,
           DirectoryFactory.class,
           ManagedIndexSchemaFactory.class,
--- a/solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
+++ b/solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
@@ -49,6 +49,11 @@ public class CPUCircuitBreaker extends CircuitBreaker implements SolrCoreAware {
     super();
   }
 
+  @Deprecated(since = "9.5")
+  public CPUCircuitBreaker(SolrCore core) {
+    this(core.getCoreContainer());
+  }
+
   public CPUCircuitBreaker(CoreContainer coreContainer) {
     super();
     this.cc = coreContainer;
--- a/solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
+++ b/solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
@@ -56,11 +56,6 @@ public class CPUCircuitBreaker extends CircuitBreaker implements SolrCoreAware {
   }
 
   @Override
-  public void init(NamedList<?> args) {
-    super.init(args);
-  }
-
-  @Override
   public boolean isTripped() {
     if (enabled == null) {
       double localSeenCPUUsage = calculateLiveCPUUsage();
--- a/solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
+++ b/solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
@@ -20,7 +20,6 @@ package org.apache.solr.util.circuitbreaker;
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.Metric;
 import java.lang.invoke.MethodHandles;
-import org.apache.solr.common.util.NamedList;
 import org.apache.solr.core.CoreContainer;
 import org.apache.solr.core.SolrCore;
 import org.apache.solr.metrics.SolrMetricManager;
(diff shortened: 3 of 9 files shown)


janhoy:
Next step would be to add some solr.xml parsing, say from a new `<circuitBreakers>` tag. The inner xml could be identical as the one used in solrconfig.xml, and the loop parsing them would simply register them in the global static context of the registry. Then I believe the POC would be ready :)

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerRegistry.java
```diff
@@ -62,9 +81,11 @@ public void deregisterAll() {
    */
   public List<CircuitBreaker> checkTripped(SolrRequestType requestType) {
     List<CircuitBreaker> triggeredCircuitBreakers = null;
+    final List<CircuitBreaker> breakersOfType = new ArrayList<>();
 
-    for (CircuitBreaker circuitBreaker :
-        circuitBreakerMap.getOrDefault(requestType, Collections.emptyList())) {
+    breakersOfType.addAll(circuitBreakerMap.getOrDefault(requestType, Collections.emptyList()));
+    breakersOfType.addAll(globalCircuitBreakerMap.getOrDefault(requestType, Collections.emptyList()));
+    for (CircuitBreaker circuitBreaker : breakersOfType) {
```
janhoy, line 88: I'm sure it would be possible to stream the two maps in a more efficient streaming fashion to avoid extra list, but this is just a POC :)

solr/core/src/test/org/apache/solr/util/TestGlobalCircuitBreaker.java
```diff
@@ -0,0 +1,71 @@
...
+/** Tests the pluggable circuit breaker implementation. The actual tests are in base class. */
+public class TestGlobalCircuitBreaker extends SolrTestCaseJ4 {
+  @BeforeClass
+  public static void setUpClass() throws Exception {
+    System.setProperty("filterCache.enabled", "false");
+    System.setProperty("queryResultCache.enabled", "false");
+    System.setProperty("documentCache.enabled", "true");
+
+    // Set a global update breaker for a low CPU, which will trip during indexing
+    System.setProperty(CircuitBreakerRegistry.SYSPROP_UPDATE_LOADAVG, "0.1");
```
janhoy, line 36: This prop causes a global loadavg-cb to be registered. I chose this since the new CPU CB is not supported on all JVMs and did not work in tests on my Mac.

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -3178,6 +3179,9 @@ public <T> T initPlugins(
             type.getSimpleName() + "." + info.name, (SolrMetricProducer) o);
       }
       if (o instanceof CircuitBreaker) {
+        if (o instanceof SolrCoreAware) {
+          ((SolrCoreAware) o).inform(this);
```
janhoy, line 3183: As this init mechanism requires a default no-arg constructor, had to make it SolrCoreAware to inject CC.

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
```diff
@@ -33,39 +35,47 @@
...
+        String msg =
+            "Initialization failure for CPU circuit breaker. Unable to get 'systemCpuLoad', not supported by the JVM?";
+        if (log.isErrorEnabled()) {
+          log.error(msg);
+        }
+        enabled = false;
+      } else {
+        enabled = true;
+      }
+    }
```
janhoy, line 78: When this CB is initialized by SolrCore it gets its CoreContainer from `inform()`, which happens after `init()`. Thus  had to move the early check of enabled true/false to here.

When initialized from static / global context, we still use the constructor with CC, since we don't have a SolrCore there to inject.

solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
```diff
@@ -814,6 +815,7 @@ public Path getInstancePath() {
         new Class<?>[] {
           // DO NOT ADD THINGS TO THIS LIST -- ESPECIALLY THINGS THAT CAN BE CREATED DYNAMICALLY
           // VIA RUNTIME APIS -- UNTIL CAREFULLY CONSIDERING THE ISSUES MENTIONED IN SOLR-8311
+          CircuitBreaker.class,
```
janhoy, line 818: Any objections to add CircuitBreaker to the list of SolrCoreAware classes? Solr's home grown IOC sucks but that's what we have...

janhoy:
I think this is ready for review

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
```diff
@@ -33,39 +35,47 @@
...
-
-    if (localSeenCPUUsage < 0) {
-      String msg =
-          "Initialization failure for CPU circuit breaker. Unable to get 'systemCpuLoad', not supported by the JVM?";
-      if (log.isErrorEnabled()) {
-        log.error(msg);
-      }
-      enabled = false;
-    }
   }
```
cpoerschke, line 61: Can remove `init` then in favour of super method only.

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
cpoerschke, line 78: Wondering if this could move into the `inform` method to avoid `Boolean` vs. `AtomicBoolean` wonderings for `enabled` in case of concurrent `isTripped` calls.

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerRegistry.java
```diff
@@ -111,43 +199,82 @@ public static String toErrorMessage(List<CircuitBreaker> circuitBreakerList) {
   }
 
   public boolean isEnabled(SolrRequestType requestType) {
-    return circuitBreakerMap.containsKey(requestType);
+    return circuitBreakerMap.containsKey(requestType)
+        || globalCircuitBreakerMap.containsKey(requestType);
```
cpoerschke, line 203:
```suggestion
    return globalCircuitBreakerMap.containsKey(requestType)
        || circuitBreakerMap.containsKey(requestType);
```

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerRegistry.java
```diff
@@ -111,43 +199,82 @@ public static String toErrorMessage(List<CircuitBreaker> circuitBreakerList) {
...
+   * @throws SolrException if any CB fails to close
+   */
+  private static void closeCircuitBreakers(List<CircuitBreaker> breakers) {
+    final AtomicInteger closeFailedCounter = new AtomicInteger(0);
+    breakers.forEach(
+        it -> {
+          try {
+            if (log.isDebugEnabled()) {
+              log.debug(
+                  "Closed circuit breaker {} for request type(s) {}",
```
cpoerschke, line 237:
```suggestion
                  "Closing circuit breaker {} for request type(s) {}",
```

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerRegistry.java
```diff
@@ -111,43 +199,82 @@ public static String toErrorMessage(List<CircuitBreaker> circuitBreakerList) {
...
+                  it.getRequestTypes());
+            }
+            it.close();
+          } catch (IOException e) {
+            if (log.isErrorEnabled()) {
+              log.error(
+                  String.format(
+                      Locale.ROOT,
+                      "Failed to close circuit breaker %s",
+                      it.getClass().getSimpleName()),
```
cpoerschke, line 248: In case of ambiguity e.g. a global search CB and a per-core update CB or so.
```suggestion
                      "Failed to close circuit breaker %s for request type(s) {}",
                      it.getClass().getSimpleName(),
                      it.getRequestTypes()),
```

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
```diff
@@ -33,39 +35,47 @@
...
-  public CPUCircuitBreaker(SolrCore core) {
+  public CPUCircuitBreaker() {
     super();
-    this.core = core;
+  }
+
+  public CPUCircuitBreaker(CoreContainer coreContainer) {
+    super();
+    this.cc = coreContainer;
   }
```
cpoerschke, line 56: Wondering if we need to worry about backwards compatibility here, maybe not maintaining it but upgrade-note-explaining the signature change perhaps? `CircuitBreakerRegistry` is tagged as experimental but this class is not i.e. in principle folks could have sub-classed a custom version of it or so.

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
janhoy: Indeed

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerRegistry.java
janhoy: Why?

solr/core/src/java/org/apache/solr/util/circuitbreaker/CPUCircuitBreaker.java
janhoy: Sure, will keep the old constructor and deprecate it.

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerRegistry.java
cpoerschke: If (and that is an if) one assumes that global circuit breakers would be more commonly configured (since they are easier to configure) than per-core ones then a successful lookup in the first map would save a lookup attempt in the second map.

cpoerschke (approved):

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerRegistry.java
janhoy: Hehe. This map will contain two request type enums `QUERY` and `UPDATE`, so they should be super fast anyway. And we don't have reason to believe that one is more used than the other at this point. So I'll leave it as is.

PR #1920, opened 2023-09-13
SOLR-16975: Remove loading of solr.xml from zookeeper (main branch only)

janhoy:
https://issues.apache.org/jira/browse/SOLR-16975

Solr will now fail and exit if it finds `/solr.xml` in zookeeper, with a message on how to remove it.

Have also cleaned up docs and some scripts that used `/solr.xml` as example of how to upload stuff to zookeeper.

The `MiniSolrCloudCluster` relied on uploading a provided solr.xml to ZK, now it instead copies the provided file to each Jetty.


solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
```diff
@@ -299,8 +299,9 @@ public MiniSolrCloudCluster(
             .withUrl(zkServer.getZkHost())
             .withTimeout(AbstractZkTestCase.TIMEOUT, TimeUnit.MILLISECONDS)
             .build()) {
-      if (!zkClient.exists("/solr/solr.xml", true)) {
-        zkClient.makePath("/solr/solr.xml", solrXml.getBytes(Charset.defaultCharset()), true);
+      // NOCOMMIT: Create chroot here, not sure why it is now necessary?
+      zkClient.makePath("/solr", false, true);
```
HoustonPutman, line 303: Hmm, you would hope the code in [SOLR-7642](https://issues.apache.org/jira/browse/SOLR-7642) would make this unnecessary... Or is it because of the `zkClient.MakePath()` code below? weird since that is the same code as was deleted for `solr.xml`...

janhoy: Earlier line 303 always created `/solr/solr.xml`, so when that is removed, and security is not enabled, we need to create `/solr` manually.

dsmiley (approved):

solr/core/src/java/org/apache/solr/core/NodeConfig.java
```diff
@@ -218,44 +211,36 @@ private NodeConfig(
...
-      log.warn(
-          "Solr property solr.solrxml.location is no longer supported. Will automatically load solr.xml from ZooKeeper if it exists");
-    }
     final SolrResourceLoader loader = new SolrResourceLoader(solrHome);
     initModules(loader, null);
     nodeProperties = SolrXmlConfig.wrapAndSetZkHostFromSysPropIfNeeded(nodeProperties);
+
+    // TODO: Only job of this block is to
+    //  delay starting a solr core to satisfy
+    //  ZkFailoverTest test case...
```
janhoy, line 225: Rewrite ZkFailoverTest to enforce a failed restart of cluster in a different way than relying on this sysprop hack...

janhoy:
So I fixed the last tests and back some ZK connection code to `NodeConfig`. It really does not belong there, but `ZkFailoverTest` uses it plus a sysProp to simulate cluster restart failure. It should be fixed but I don't want to tackle it here, so I added a TODO.

Final review welcome. @dsmiley ?

dsmiley (approved):
+1
Sad TODO there :-(

risdenk (approved):

HoustonPutman (approved):
Thanks Jan!

PR #1921, opened 2023-09-13
SOLR-16951: Cache client pkiAuth headers for a second

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-16951

Also increase the TTL from 5 seconds to 10.

This is a WIP for now. probably need to think some more about it.

### DIFF
--- a/solr/core/src/java/org/apache/solr/security/PKIAuthenticationPlugin.java
+++ b/solr/core/src/java/org/apache/solr/security/PKIAuthenticationPlugin.java
@@ -26,6 +26,7 @@ import java.security.InvalidKeyException;
 import java.security.Principal;
 import java.security.PublicKey;
 import java.security.SignatureException;
+import java.time.Duration;
 import java.time.Instant;
 import java.util.Base64;
 import java.util.List;
@@ -33,6 +34,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReference;
 import javax.servlet.FilterChain;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -92,7 +94,7 @@ public class PKIAuthenticationPlugin extends AuthenticationPlugin
   private final Map<String, PublicKey> keyCache = new ConcurrentHashMap<>();
   private final PublicKeyHandler publicKeyHandler;
   private final CoreContainer cores;
-  private static final int MAX_VALIDITY = Integer.getInteger("pkiauth.ttl", 5000);
+  private static final int MAX_VALIDITY = Integer.getInteger("pkiauth.ttl", 10000);
   private final String myNodeName;
   private final HttpHeaderClientInterceptor interceptor = new HttpHeaderClientInterceptor();
   private boolean interceptorRegistered = false;
@@ -403,12 +405,12 @@ public class PKIAuthenticationPlugin extends AuthenticationPlugin
             final Optional<String> preFetchedUser = getUserFromJettyRequest(request);
             if ("v1".equals(System.getProperty(SEND_VERSION))) {
               preFetchedUser
-                  .map(PKIAuthenticationPlugin.this::generateToken)
-                  .ifPresent(token -> request.header(HEADER, token));
+                  .map(PKIAuthenticationPlugin.this::getToken)
+                  .ifPresent(token -> request.headers(mutable -> mutable.add(HEADER, token)));
             } else {
               preFetchedUser
-                  .map(PKIAuthenticationPlugin.this::generateTokenV2)
-                  .ifPresent(token -> request.header(HEADER_V2, token));
+                  .map(PKIAuthenticationPlugin.this::getTokenV2)
+                  .ifPresent(token -> request.headers(mutable -> mutable.add(HEADER_V2, token)));
             }
           }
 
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/security/PKIAuthenticationPlugin.java
```diff
@@ -485,34 +487,75 @@ private Optional<String> getUser() {
...
+      this.token = token;
+    }
+  }
+
+  private volatile ConcurrentHashMap<String, AtomicReference<CachedToken>> cachedV1Tokens =
+      new ConcurrentHashMap<>();
+  private volatile ConcurrentHashMap<String, AtomicReference<CachedToken>> cachedV2Tokens =
+      new ConcurrentHashMap<>();
+
+  private static final Duration cacheExpiryTime = Duration.ofSeconds(1);
```
tflobbe, line 505: How about we make this relative to `MAX_VALIDITY`? like `MAX_VALIDITY * 0.5`, or `MAX_VALIDITY * 0.1` by default?

solr/core/src/java/org/apache/solr/security/PKIAuthenticationPlugin.java
```diff
@@ -485,34 +487,75 @@ private Optional<String> getUser() {
     }
   }
 
+  private static class CachedToken {
+    Instant generatedAt;
+    String token;
+
+    private CachedToken(Instant generatedAt, String token) {
+      this.generatedAt = generatedAt;
```
tflobbe, line 495: not a big deal, but if we just store the expiration time instead of the generation time we don't have to subtract every time

solr/core/src/java/org/apache/solr/security/PKIAuthenticationPlugin.java
```diff
@@ -485,34 +487,75 @@ private Optional<String> getUser() {
...
+  private volatile ConcurrentHashMap<String, AtomicReference<CachedToken>> cachedV1Tokens =
+      new ConcurrentHashMap<>();
+  private volatile ConcurrentHashMap<String, AtomicReference<CachedToken>> cachedV2Tokens =
+      new ConcurrentHashMap<>();
+
+  private static final Duration cacheExpiryTime = Duration.ofSeconds(1);
+
+  private String getToken(String usr) {
+    AtomicReference<CachedToken> tokenRef =
+        cachedV1Tokens.computeIfAbsent(usr, u -> new AtomicReference<>(generateToken(u)));
```
tflobbe, line 509: I'm wondering if we need to consider limiting the size of this map in some way? can the different users be ever growing (probably not with the OOTB authentication, but maybe with plugins?)

solr/core/src/java/org/apache/solr/security/PKIAuthenticationPlugin.java
```diff
@@ -485,34 +487,75 @@ private Optional<String> getUser() {
...
-  private String generateToken(String usr) {
+  private CachedToken generateToken(String usr) {
     assert usr != null;
     String s = usr + " " + System.currentTimeMillis();
     byte[] payload = s.getBytes(UTF_8);
     byte[] payloadCipher = publicKeyHandler.getKeyPair().encrypt(ByteBuffer.wrap(payload));
     String base64Cipher = Base64.getEncoder().encodeToString(payloadCipher);
     log.trace("generateToken: usr={} token={}", usr, base64Cipher);
-    return myNodeName + " " + base64Cipher;
+    return new CachedToken(Instant.now(), myNodeName + " " + base64Cipher);
```
stillalex, line 541: `Instant.now(),` could move into the CachedToken constructor, doesn't need to be specified here.

solr/core/src/java/org/apache/solr/security/PKIAuthenticationPlugin.java
stillalex: +1 to the above and maybe also add a method `isExpired()` to simplify the code

solr/core/src/java/org/apache/solr/security/PKIAuthenticationPlugin.java
```diff
@@ -485,34 +487,75 @@ private Optional<String> getUser() {
...
+      synchronized (tokenRef) {
+        if (tokenRef.get().generatedAt.isBefore(Instant.now().minus(cacheExpiryTime))) {
+          tokenRef.set(generateToken(usr));
+        }
+      }
+    }
+    return tokenRef.get().token;
+  }
+
+  private synchronized String getTokenV2(String usr) {
```
stillalex, line 520: the 2 methods `getToken` and  `getTokenV2` are very similar. would you consider unifying into a single one? (one that gets a map and a 'token generator' function)

solr/core/src/java/org/apache/solr/security/PKIAuthenticationPlugin.java
```diff
@@ -485,34 +487,75 @@ private Optional<String> getUser() {
...
+    AtomicReference<CachedToken> tokenRef =
+        cachedV2Tokens.computeIfAbsent(usr, u -> new AtomicReference<>(generateTokenV2(u)));
+    if (tokenRef.get().generatedAt.isBefore(Instant.now().minus(cacheExpiryTime))) {
```
stillalex, line 523: just wondering how this performs if it would be a single `ConcurrentHashMap#compute` method instead of all the locks and checks. is there a benchmark for this change?

HoustonPutman:
Closing this in favor of https://github.com/apache/solr/pull/3334

PR #1922, opened 2023-09-13
SOLR-16938 Auto configure tracer without a <tracerConfig> tag in solr…

stillalex:
….xml

https://issues.apache.org/jira/browse/SOLR-16938

# Description

Enable OTEL based on presence of system property. also use config params for otel init, but not if the respective system properties are already defined.

# Solution

# Tests

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/TracerConfigurator.java
+++ b/solr/core/src/java/org/apache/solr/core/TracerConfigurator.java
@@ -20,6 +20,7 @@ package org.apache.solr.core;
 import io.opentelemetry.api.trace.Tracer;
 import java.lang.invoke.MethodHandles;
 import java.util.Locale;
+import java.util.Map;
 import org.apache.solr.common.SolrException;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.util.plugin.NamedListInitializedPlugin;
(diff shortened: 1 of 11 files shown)


stillalex:
@janhoy if you have a few mins could you please take a look.
The proposed approach is to attempt to init otel if the `OTEL_SERVICE_NAME` property exists. I am open to suggestions here if there is anything better. if it will attempt and fail there is a log message that points to enabling module.

The other part of this PR is passing the init params to the `OtelTracerConfigurator` and using them to set some system properties if they are not already set. this should fill in the plugin config gap.

solr/core/src/java/org/apache/solr/core/TracerConfigurator.java
```diff
@@ -18,29 +18,78 @@
...
+      return false;
+    }
+    return getConfig("OTEL_SERVICE_NAME") != null;
+  }
+
+  private static String getConfig(String envName) {
+    String envValue = System.getenv().get(envName);
+    String sysName = envName.toLowerCase(Locale.ROOT).replace("_", ".");
+    String sysValue = System.getProperty(sysName);
+    return sysValue != null ? sysValue : envValue;
```
janhoy, line 93: This is duplicating `OtelTracerConfigurator#getEnvOrSysprop`. Perhaps you can extract some functionality into a common static utility?

janhoy (commented):
Looks like a way to go.
Will you attempt a ref-guide edit as well?

PR #1923, opened 2023-09-14
SOLR-16979: Randomize port number of Solr in BATS tests

janhoy:
https://issues.apache.org/jira/browse/SOLR-16979

* Assign a random `$SOLR_PORT` for bats tests (through env var)
* Make `solr -e techproducts` respect `$SOLR_PORT`. The `cloud` example uses `$SOLR_PORT+{0..3}` for each node.
* Set `SOLR_PID_DIR` explicitly, to make `solr stop -all` find the pid files
* Add a dummy cleanup test at the very end that explicitly shuts down solr on every known port

The last dummy test is perhaps not necessary, if we get the `solr stop -all` work in all cases, but wanted to make sure solr is always stopped in Jenkins even if a test fails. That has not worked very well until now.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -204,6 +204,8 @@ Other Changes
 
 * SOLR-16623: new SolrJettyTestRule for tests needing HTTP or Jetty. (David Smiley, Joshua Ouma)
 
+* SOLR-16979: BATS integration tests now mostly start solr instances on a randomly selected port, and do extra cleanup (janhoy)
+
 ==================  9.3.0 ==================
 
 Upgrade Notes
--- a/solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
@@ -44,6 +44,7 @@ import org.apache.commons.exec.Executor;
 import org.apache.commons.exec.OS;
 import org.apache.commons.exec.environment.EnvironmentUtils;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.SystemUtils;
 import org.apache.solr.client.solrj.SolrClient;
 import org.apache.solr.client.solrj.impl.CloudSolrClient;
 import org.apache.solr.client.solrj.impl.Http2SolrClient;
@@ -253,7 +254,9 @@ public class RunExampleTool extends ToolBase {
 
     boolean isCloudMode = cli.hasOption('c');
     String zkHost = cli.getOptionValue('z');
-    int port = Integer.parseInt(cli.getOptionValue('p', "8983"));
+    int port =
+        Integer.parseInt(
+            cli.getOptionValue('p', SystemUtils.getEnvironmentVariable("SOLR_PORT", "8983")));
     Map<String, Object> nodeStatus =
         startSolr(new File(exDir, "solr"), isCloudMode, cli, port, zkHost, 30);
 
@@ -436,7 +439,10 @@ public class RunExampleTool extends ToolBase {
 
     boolean prompt = !cli.hasOption("noprompt");
     int numNodes = 2;
-    int[] cloudPorts = new int[] {8983, 7574, 8984, 7575};
+    int defaultPort =
+        Integer.parseInt(
+            cli.getOptionValue('p', SystemUtils.getEnvironmentVariable("SOLR_PORT", "8983")));
+    int[] cloudPorts = new int[] {defaultPort, defaultPort + 1, defaultPort + 2, defaultPort + 3};
     File cloudDir = new File(exampleDir, "cloud");
     if (!cloudDir.isDirectory()) cloudDir.mkdir();
 
--- a/solr/packaging/build.gradle
+++ b/solr/packaging/build.gradle
@@ -238,6 +238,7 @@ task integrationTests(type: BatsTask) {
   def integrationTestOutput = "$buildDir/test-output"
   def solrHome = "$integrationTestOutput/solr-home"
   def solrTestFailuresDir = "$integrationTestOutput/failure-snapshots"
+  def solrPort = new Random().nextInt(10000, 50000)
 
   inputs.dir(distDir)
   outputs.dir(integrationTestOutput)
@@ -250,11 +251,18 @@ task integrationTests(type: BatsTask) {
 
     // TODO - if quiet then don't tee
     standardOutput = new TeeOutputStream(System.out, new FileOutputStream("$integrationTestOutput/test-output.txt"))
-
+    println("Runnin BATS tests with Solr base port ${solrPort}")
   }
 
   environment SOLR_TIP: distDir.toString()
   environment SOLR_HOME: solrHome
+  environment SOLR_PID_DIR: solrHome
+  environment SOLR_PORT: solrPort
+  environment SOLR2_PORT: solrPort + 1
+  environment SOLR3_PORT: solrPort + 2
+  environment SOLR_PORT_8983: 8983
+  environment SOLR_PORT_7574: 7574
+  environment ZK_PORT: solrPort + 1000
   environment SOLR_LOGS_DIR: "$solrHome/logs"
   environment TEST_OUTPUT_DIR: integrationTestOutput
   environment TEST_FAILURE_DIR: solrTestFailuresDir
--- a/solr/packaging/test/test_extraction.bats
+++ b/solr/packaging/test/test_extraction.bats
@@ -47,11 +47,11 @@ teardown() {
       "class": "solr.extraction.ExtractingRequestHandler",
       "defaults":{ "lowernames": "true", "captureAttr":"true"}
     }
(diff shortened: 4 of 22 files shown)


solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
```diff
@@ -436,7 +439,10 @@ protected void runCloudExample(CommandLine cli) throws Exception {
 
     boolean prompt = !cli.hasOption("noprompt");
     int numNodes = 2;
-    int[] cloudPorts = new int[] {8983, 7574, 8984, 7575};
+    int defaultPort =
+        Integer.parseInt(
+            cli.getOptionValue('p', SystemUtils.getEnvironmentVariable("SOLR_PORT", "8983")));
+    int[] cloudPorts = new int[] {defaultPort, defaultPort + 1, defaultPort + 2, defaultPort + 3};
```
epugh, line 445: Yay!  Finally have port numbers that make a bit more sense...

solr/packaging/build.gradle
```diff
@@ -250,11 +251,18 @@ task integrationTests(type: BatsTask) {
 
     // TODO - if quiet then don't tee
     standardOutput = new TeeOutputStream(System.out, new FileOutputStream("$integrationTestOutput/test-output.txt"))
-
+    println("Runnin BATS tests with Solr base port ${solrPort}")
```
epugh, line 254: "Running"

solr/packaging/test/test_zz_cleanup.bats
epugh, line 1: Do we need a comment in this test why it's called zz?   I'm assuming it's to run at the end ;-)

solr/packaging/build.gradle
```diff
@@ -250,11 +251,18 @@ task integrationTests(type: BatsTask) {
...
   }
 
   environment SOLR_TIP: distDir.toString()
   environment SOLR_HOME: solrHome
+  environment SOLR_PID_DIR: solrHome
+  environment SOLR_PORT: solrPort
+  environment SOLR2_PORT: solrPort + 1
+  environment SOLR3_PORT: solrPort + 2
+  environment SOLR_PORT_8983: 8983
+  environment SOLR_PORT_7574: 7574
```
epugh, line 264: not sure we need solr3 or solr_port8983 and solr_port7574....

epugh:
So this all makes sense.   One thing that I think we lose a bit is that I used to take the commands in the BATS script and run the to test out things locally in Solr, and so haveing the 8983 was convenient ;-).    

One thing is, if we move bats tests to NOT running in the 8983 port number, well, it might be nice for me to run locally 8983 solr, and then run the bats tests to see if I get it all working.   

Sometimes I get a setup of Solr (on 8983) that is donig what I want.  Now I want to write an integration tests.  So then I am constantly stopping the 8983, running the bats, and then starting 8983, as I work through it.

If we move the bats tests to run in the higher port numbers, then i can have my solr running on 8983, 8984 etc, and then bats ones are running on the 10000 and higher range???   

Do we then maybe NOT want the test_zz behavior?

solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
janhoy: Yep. Of course this will ruin tutorials out there that assume something. So perhaps we should make the default as they were, but if `SOLR_HOST` or `-p` is given, we override those defaults like here?

solr/packaging/test/test_zz_cleanup.bats
epugh: Feels like `stop -all` should stop everything that `solr status` finds!

solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
```diff
@@ -436,7 +439,10 @@ protected void runCloudExample(CommandLine cli) throws Exception {
 
     boolean prompt = !cli.hasOption("noprompt");
     int numNodes = 2;
-    int[] cloudPorts = new int[] {8983, 7574, 8984, 7575};
+    int defaultPort =
+        Integer.parseInt(
+            cli.getOptionValue('p', SystemUtils.getEnvironmentVariable("SOLR_PORT", "8983")));
```
risdenk, line 444: do we need systemutils? 

```
System.getenv().getOrDefault("SOLR_PORT", "8983");
```

solr/packaging/build.gradle
janhoy: Removed. There was a test somewhere that used a third port number (check the var), but perhaps it could use 1 or 2?

solr/packaging/test/test_extraction.bats
```diff
@@ -70,15 +70,15 @@ teardown() {
       "class": "solr.extraction.ExtractingRequestHandler",
       "defaults":{ "lowernames": "true", "captureAttr":"true"}
     }
-  }' 'http://localhost:8983/solr/content_extraction/config'
+  }' 'http://localhost:${SOLR_PORT}/solr/content_extraction/config'
```
risdenk, line 73: Should be double quotes to support SOLR_PORT expansion

solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
janhoy: Ah, I just tried `System.getEnv()`, will switch.

solr/packaging/build.gradle
```diff
@@ -238,6 +238,7 @@ task integrationTests(type: BatsTask) {
   def integrationTestOutput = "$buildDir/test-output"
   def solrHome = "$integrationTestOutput/solr-home"
   def solrTestFailuresDir = "$integrationTestOutput/failure-snapshots"
+  def solrPort = new Random().nextInt(10000, 50000)
```
risdenk, line 241: Future improvement idea: check that the ports chosen aren't actually in use by anything before the test(s). I like this improvement, but there is no guarantee that a run won't conflict with an already allocated port on the machine.

solr/packaging/build.gradle
HoustonPutman: We do need a Solr 3. The new mTLS tests spin up a third node. They could restart the second node instead, so your choice there.

HoustonPutman:
Thanks so much for taking this up @janhoy ! I've been meaning to, but always put it off. This will be great.

solr/packaging/test/test_zz_cleanup.bats
janhoy: Should probably spin that out in a new JIRA

solr/packaging/build.gradle
janhoy: Keeping as is for now.

solr/packaging/build.gradle
janhoy: True. Do you know a good Java library / API to check it (other than trying to listen to a socket and catch the exception)?

risdenk: not really - even https://stackoverflow.com/questions/2675362/how-to-find-an-available-port which is used elsewhere in Solr doesn't guarantee that port won't be used immediately after. Its an idea though

janhoy:
> One thing that I think we lose a bit is that I used to take the commands in the BATS script and run the to test out things locally in Solr, and so haveing the 8983 was convenient ;-).

So we could perhaps have a gradle `-Pbats.port=8983` option that could be passed to the tests to override?

solr/packaging/build.gradle
madrob: Why do we need three consecutive ports? Get three random ports from :0

janhoy: Yep, will test binding to zero now.

janhoy: That worked, except it picked 648xx once, and then solr complained that the embedded ZK on +1000 could not start. So I now use a while loop until we get an acceptable port. Then I still add +1 and +2 to the other nodes, since else we'd need to factor out the port finder as a function and I don't know how that looks like in gradle, and you'd still not be guaranteed unique ports, since it could suggest the same port twice :) Think this is about 100x better than what we had...

solr/packaging/test/test_zz_cleanup.bats
janhoy: So I simplified the test to only use `stop -all` and then assert the `solr status`. Let's have it bake for some time, and try to improve `bin/solr` separately.

solr/CHANGES.txt
```diff
@@ -204,6 +204,8 @@ Other Changes
 
 * SOLR-16623: new SolrJettyTestRule for tests needing HTTP or Jetty. (David Smiley, Joshua Ouma)
 
+* SOLR-16979: BATS integration tests now mostly start solr instances on a randomly selected port, and do extra cleanup (janhoy)
```
janhoy, line 207:
```suggestion
* SOLR-16979: BATS integration tests now start solr instances on a randomly selected port (janhoy)
```

janhoy:
I'll merge once I get one review approval.

HoustonPutman (approved):
Awesome change, and long overdue! Very well executed

PR #1924, opened 2023-09-14
SOLR-16981: bin/solr stop improvements

janhoy:
https://issues.apache.org/jira/browse/SOLR-16981

```
bin/solr stop -h

Usage: solr stop [-k key] [-p port] [-all] [-V]

  -k <key>      Stop key; default is solrrocks

  -p <port>     Specify the port the Solr HTTP listener is bound to

  -all          Find and stop all running Solr servers on this host

  -V/-verbose   Verbose messages from this script

  NOTE: To see if any Solr servers are running, do: solr status
```

The behavior with only `solr stop` is not clearly documented, but it can be seen from [script comments](https://github.com/apache/solr/blob/main/solr/bin/solr#L1469-L1470) that the intent is that a pure `solr stop` will check PIDs, and if exactly one is found, it will stop it. If more are found, it will abort and ask user to pass `-all`.

However, if the `SOLR_HOME` env var is set (as it e.g. is in docker), and the running solr is on another port, the `solr stop` command will output "No process found for Solr node running on 8983". Thus it is interpreted the same way as if `solr stop -p 8983` was passed. I feel this is confusing, and this PR fixes it by requiring an explicit `-p` switch for that behavior.

We can also explore how the `stop -all` command can have a fallback to use `ps` as the `status` command has, if no PIDs are found...

### DIFF
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -147,8 +147,6 @@ fi
 : "${SOLR_STOP_WAIT:=180}"
 : "${SOLR_START_WAIT:=$SOLR_STOP_WAIT}" # defaulting to $SOLR_STOP_WAIT for backwards compatibility reasons
 
-# Store whether a solr port was explicitly provided, for the "solr stop" command.
-PROVIDED_SOLR_PORT="${SOLR_PORT:-}"
 : "${SOLR_PORT:=8983}"
 export SOLR_PORT
 
@@ -412,7 +410,7 @@ function print_usage() {
     echo ""
   elif [ "$CMD" == "stop" ]; then
     echo ""
-    echo "Usage: solr stop [-k key] [-p port] [-V]"
+    echo "Usage: solr stop [-k key] [-p port] [-all] [-V]"
     echo ""
     echo "  -k <key>      Stop key; default is solrrocks"
     echo ""


janhoy:
You can reproduce as follows
```bash
docker run -ti -p 8983:8983 --rm solr:9.3 bash
solr@09e8908a61c6:/opt/solr-9.3.0$ solr start -p 3333
[...]
Waiting up to 180 seconds to see Solr running on port 3333 [/]  
Started Solr server on port 3333 (pid=64). Happy searching!

solr@09e8908a61c6:/opt/solr-9.3.0$ solr stop
No process found for Solr node running on port 8983
solr@09e8908a61c6:/opt/solr-9.3.0$ unset SOLR_PORT
solr@09e8908a61c6:/opt/solr-9.3.0$ solr stop
Sending stop command to Solr running on port 3333 ... waiting up to 180 seconds to allow Jetty process 64 to stop gracefully.
```

epugh:
this makes sense...    BTW, do we have the same issue on the windows side?

HoustonPutman:
> LGTM.
> 
> I like seeing the elimination of that werid "PROVIDED_SOLR_PORT" variable!

The variable is still being used. If you provide `-p` to `solr stop`, it will set it in the PROVIDED_SOLR_PORT variable.

> We can also explore how the stop -all command can have a fallback to use ps as the status command has, if no PIDs are found...

The package command already does this (which I don't think it should), so it would be easy to port that code over to the stop command. I don't even think it's worth keeping the PID code. We should just use `ps`.

janhoy:
> The variable is still being used. If you provide `-p` to `solr stop`, it will set it in the PROVIDED_SOLR_PORT variable.

Just realized that part of this was added/modified in https://github.com/apache/solr/pull/1849/files from August. The line I removed is not there in 9.3 but still the same behavior?

janhoy:
Thinking aloud. Not sure this is a bug, but rather a documentation issue. If I were to add docs for the current behavior, it would be:

```
Usage: solr stop [-k key] [-p port] [-all] [-V]

  If no options are given, the behavior depends on whether SOLR_PORT variable is set.
  If SOLR_PORT is not set, Solr will stop the running instance, unless more than one is running.
  If SOLR_PORT is set, Solr will stop solr running on that port.
```

epugh:
@janhoy I tried to make a comment "Your docs make sense...   honestly the use of SOLR_PORT isn't something I have ever used since it is buried in config file..    " and somehow clicked the wrong buttons!!!    And merged!   

I opened a new PR, but going to stop now so I don't make things worse!!!!!!!!!!!!!!!!!

PR #1929, opened 2023-09-15
Revert "SOLR-16981: bin/solr stop improvements"

epugh:
Reverts apache/solr#1924


janhoy:
I think the best is to just revert and continue work on the issue as normal until we decide what needs to be done.

PS: The "usage" text fix does not need to be reverted?

PR #1930, opened 2023-09-15
SOLR-16982: Trip a Circuit Breaker only for external requests

janhoy:
https://issues.apache.org/jira/browse/SOLR-16982

There may be more smart/robust ways of telling whether a request is "internal" or sharded, but this works :)


solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -343,4 +343,16 @@ public Collection<Api> getApis() {
...
+  /**
+   * Checks whether the given request is an internal request to a shard. We rely on the fact that an
+   * internal search request to a shard contains the param "isShard", and an internal update request
+   * to a shard contains the param "distrib.from".
+   *
+   * @return true if request is internal
+   */
+  public static boolean isInternalShardRequest(SolrQueryRequest req) {
+    return req.getParams().get("distrib.from") != null
+        || "true".equals(req.getParams().get("isShard"));
```
cpoerschke, line 356:
```suggestion
        || "true".equals(req.getParams().get(ShardParams.IS_SHARD));
```

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -343,4 +343,16 @@ public Collection<Api> getApis() {
...
+
+  /**
+   * Checks whether the given request is an internal request to a shard. We rely on the fact that an
+   * internal search request to a shard contains the param "isShard", and an internal update request
+   * to a shard contains the param "distrib.from".
+   *
+   * @return true if request is internal
+   */
+  public static boolean isInternalShardRequest(SolrQueryRequest req) {
+    return req.getParams().get("distrib.from") != null
```
cpoerschke, line 355:
```suggestion
    return req.getParams().get(DistributedUpdateProcessor.DISTRIB_FROM) != null
```

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -343,4 +343,16 @@ public Collection<Api> getApis() {
...
   }
+
+  /**
+   * Checks whether the given request is an internal request to a shard. We rely on the fact that an
+   * internal search request to a shard contains the param "isShard", and an internal update request
+   * to a shard contains the param "distrib.from".
+   *
+   * @return true if request is internal
+   */
+  public static boolean isInternalShardRequest(SolrQueryRequest req) {
```
cpoerschke, line 354: Perhaps this could even be protected and non-static i.e. by overriding handlers could also treat other requests as-if they are internal and so exempt them from circuit breaker logic? Hmm, though then `isInternalShardRequest` could be confusing as a name.
```suggestion
  protected boolean isInternalShardRequest(SolrQueryRequest req) {
```

janhoy: We have something called Rate Limiters in Solr too, and they rely on [some other mechanism](https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java#L82-L86) for determining internal vs external request, but I found no obvious way to re-use that here. So this param checking is a very naiive approach, but perhaps robust enough?

janhoy: > Perhaps this could even be protected and non-static

We reference this method from unit tests on the SolrQueryRequest, without an instance of a request hander, so this is a bit hard. It would of course be possible to create another instance method that delegates to this static one, but that can also be done later when need arises.

janhoy:
Plan to merge this on Friday..

solr/CHANGES.txt
```diff
@@ -109,6 +109,8 @@ Improvements
 
 * SOLR-15474: Make Circuit breakers individually pluggable (Atri Sharma, Christine Poerschke, janhoy)
 
+* SOLR-16982: Trip Circuit Breakers only for external requests (janhoy)
```
janhoy, line 112:
```suggestion
* SOLR-16982: Trip Circuit Breakers only for external requests (janhoy, Christine Poerschke)
```

PR #1931, opened 2023-09-15
SOLR-16978: Be case insensitive when parsing booleans from text

tflobbe:
The translation between V1 parameters into the new code based in V2 APIs is using `StrUtils.getBool(String param)` to parse from string. This `StrUtils.getBool` is intended to be more flexible than `Boolean.parseBoolean()`, and accepts things like `yes`/`no`, `on`/`off`, but it's case sensitive.

This PR makes it case insensitive. Note that this code is often used for reading configuration files, so those will also now be case insensitive when this is merged. 
An alternative could be to have a new method to be case insensitive and use that one for the API parameters, but really, if this code is supposed to be more flexible than `Boolean.parseBoolean()`, I think make it case insensitive is the right way to go.

I was also a bit surprised that this code uses `startsWith`, for the most part, which sounds strange to me (maybe intended to allow trailing whitespaces? I'm not sure), but that could be a discussion for a different PR (changing that would be a breaking change)

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -216,6 +216,8 @@ Other Changes
 
 * SOLR-16979: BATS integration tests now start solr instances on a randomly selected port (janhoy)
 
+* SOLR-16978: Be case insensitive when parsing booleans from text (Tomás Fernández Löbbe)
+
 ==================  9.3.0 ==================
 
 Upgrade Notes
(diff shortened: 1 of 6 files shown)


tflobbe:
@gerlowskija Any thoughts here? I'm on the fence. I don't really love the leniency with which we are reading boolean params here, it seems more useful for configuration files than it is for request parameters, but that's what other parts of the code are using, so I also want to consider consistency with other APIs. 
As far as I can tell, pre-Solr9 most Collections API parameters `ZkNodeProps`, which uses `Boolean.getBoolean` (case insensitive), so that's why they worked before and not now.

janhoy (approved):

PR #1933, opened 2023-09-16
SOLR-16397: Create v2(JAX-RS) endpoints for UNLOAD core

iamsanjay:

# Description

V2(JAX-RS) Endpoints for core admin op "UNLOAD".

# Solution

There is already one class present in the codebase named as UnloadCoreAPI.java which is using home-grown V2 api framework, this New code code changes will migrate it to JAX-RS framework.

# Tests

Added two test cases:

1. To test the valid response and status message.
2. Test Non Existent core exception and status code.

# Checklist


iamsanjay:
Getting below error, only if I'm using V2 Jersey endpoints.

> Core demo is not yet closed, waiting 100 ms before checking again

I looked into this more and found out below code snipped that got executed when calling Jersey endpoints, and that results in increase the variable named `refCount` -- core current usage count. 

https://github.com/iamsanjay/solr/blob/d772f112032c4a8648c97d688b792ccb6459a10e/solr/core/src/java/org/apache/solr/api/V2HttpCall.java#L171-L174

Now, the unload endpoints executes `SolrCore.closeAndWait()` which decreases the `refCount` and check if it's closed. Nonetheless, the refCount still 1 and thread got stuck in below loop.

https://github.com/iamsanjay/solr/blob/d772f112032c4a8648c97d688b792ccb6459a10e/solr/core/src/java/org/apache/solr/core/SolrCore.java#L1564-L1572

solr/solr-ref-guide/modules/configuration-guide/pages/coreadmin-api.adoc
```diff
@@ -439,7 +439,31 @@ The `UNLOAD` action removes a core from Solr.
 Active requests will continue to be processed, but no new requests will be sent to the named core.
 If a core is registered under more than one name, only the given name is removed.
 
-`admin/cores?action=UNLOAD&core=_core-name_`
+[.dynamic-tabs]
```
gerlowskija, line 442: [+1] Thanks for remembering the docs, looks great!

gerlowskija (changes requested):
> Core demo is not yet closed, waiting 100 ms before checking again

Hmm, I ran into a similar SolrCore ref-count issue on one other API.  Let me try to remember what I did back then and hopefully I'll have a fix to discuss here shortly.

----

Since you've been interested in helping with these PRs, I wanted to bring a recent change I made to your attention:

It turns out that if we put all of the JAX-RS annotations that describe the API on a Java interface that lives in a particular gradle module (instead of in the `core` module as we've done up to this point), then those annotations can be used to not just define the API, but also to drive client code generation!  We've already got this working in the Solr build to generate a Java client, and I've got ideas around Python, golang, and Javascript on the horizon as well - very very exciting IMO!

There's a few [developer docs](https://github.com/apache/solr/blob/main/dev-docs/apis.adoc#writing-jax-rs-apis) that describe how to create the interface with all our JAX-RS annotations, and we have a [few APIs](https://github.com/apache/solr/blob/9426902acb7081a2e9a1fa29699c5286459e1365/solr/api/src/java/org/apache/solr/client/api/endpoint/AddReplicaPropertyApi.java) that can serve as reference examples.  If you don't mind, I can make the necessary tweaks to this PR to get it to work with our code generation?

----

Otherwise, I've left a few other comments of things that we should probably change, but this looks pretty good once we get the ref-counting issue sorted out!

gerlowskija:
> I ran into a similar SolrCore ref-count issue on one other API. Let me try to remember what I did back then and hopefully I'll have a fix to discuss here shortly.

This should be fixed now.  Test it out and lmk if you're still able to reproduce.  See [the fix](https://github.com/apache/solr/pull/1933/commits/030c5e0c795bb8e150bc7f40a08ba8138fd9f4c6) here.

In short, when serving APIs, Solr attempts to distinguish whether an incoming request is trying to trigger core-specific or cluster-wide functionality.  This has always been an easy question to answer for v1 APIs because of the relatively limited set of paths that those APIs use.  But for some v2 APIs we occasionally have to "guess-and-check": we fetch the SolrCore (incrementing its ref-count), we try to find the API in the list that the SolrCore knows about, and then if that doesn't work we look elsewhere.  In this case - where we fetched a SolrCore thinking the API might live there, only to be wrong- it turns out we weren't properly decrementing the ref-count for the SolrCore.

> It turns out that if we put all of the JAX-RS annotations that describe the API on a Java interface that lives in a particular gradle module (instead of in the core module as we've done up to this point), then those annotations can be used to not just define the API, but also to drive client code generation! [...] There's a few [developer docs](https://github.com/apache/solr/blob/main/dev-docs/apis.adoc#writing-jax-rs-apis) that describe how to create the interface with all our JAX-RS annotations [...] If you don't mind, I can make the necessary tweaks to this PR to get it to work with our code generation?

I went ahead and made these changes.  With these changes, the build will now generate the following `SolrRequest` implementation "for free".  Very cool IMO

```
public static class UnloadCoreResponse extends JacksonParsingResponse<SolrJerseyResponse> {
    public UnloadCoreResponse() {
      super(SolrJerseyResponse.class);
    }
  }

  public static class UnloadCore extends SolrRequest<UnloadCoreResponse> {
    private final UnloadCoreRequestBody requestBody;
    private final String coreName;

    /**
     * Create a UnloadCore request object.
     *
     * @param coreName Path param -
     */
    public UnloadCore(String coreName) {
      super(
          SolrRequest.METHOD.valueOf("POST"),
          "/cores/{coreName}/unload".replace("{" + "coreName" + "}", coreName));

      this.coreName = coreName;
      this.requestBody = new UnloadCoreRequestBody();
      addHeader("Content-type", "application/json");
    }

    /**
     * @param deleteIndex If true, will remove the index when unloading the core.
     */
    public void setDeleteIndex(Boolean deleteIndex) {
      this.requestBody.deleteIndex = deleteIndex;
    }

    /**
     * @param deleteDataDir If true, removes the data directory and all sub-directories.
     */
    public void setDeleteDataDir(Boolean deleteDataDir) {
      this.requestBody.deleteDataDir = deleteDataDir;
    }

    /**
     * @param deleteInstanceDir If true, removes everything related to the core, including the index
     *     directory, configuration files and other related files.
     */
    public void setDeleteInstanceDir(Boolean deleteInstanceDir) {
      this.requestBody.deleteInstanceDir = deleteInstanceDir;
    }

    /**
     * @param async Request ID to track this action which will be processed asynchronously.
     */
    public void setAsync(String async) {
      this.requestBody.async = async;
    }

     // <snip> remainder of boilerplate definition removed
  }
```

----

To summarize this all - I think the only outstanding issue at this point is whether to change the RequestBody class over to use `Boolean` instead of `boolean`.  I can make that change if you don't object, or you're welcome to do it.

iamsanjay:
> I went ahead and made these changes. With these changes, the build will now generate the following SolrRequest implementation "for free". Very cool IMO

@gerlowskija Client code generation is cool indeed, I am going to explore it more. Always coming up with cool ways to improve V2 and inspire us, Love it. And, in this mission, With You!

gerlowskija:
Hi Sanjay - sorry for the delay circling back to this.  Thanks for making that `boolean`->`Boolean` switch.  This LGTM now.  I see there's still one or two tests failing: particularly `V2ApiIntegrationTest`.  Did you notice that failure?  Or maybe it passes for you and it's just something environmental for me?

Will poke at it when I can and see where things go!

gerlowskija:
Ended up finding a few fixes for the tests, and 'check' passes so I'm going to merge, thanks again Sanjay!

PR #1935, opened 2023-09-17
SOLR-15960 Unified use of system properties and environment variables

janhoy:

This is a first step, demonstrating how we can skip converting env.vars like `SOLR_FOO` in bin/solr and then passing them as `-Dsolr.foo`. Instead we just provide Solr with all the `SOLR_*` environment variables, and let the new `EnvUtils` class convert and set the proper system propertis. Our Solr code can then continue reading `System.getProperty("solr.foo")` as before.

We have some exceptions to the naming convention, and for those there is a properties file to map ENV -> prop key.

Some places in Solr we look for either env.var or sys.prop and pick the one that exists. This is no longer necessary, as you can just read the sys.prop and it will be set if the corresponding ENV was passed to Solr. 

This PR gives a few examples of cleanup that can be done, by removing explicit parsing of `SOLR_LOG_LEVEL` and `SOLR_MODULES` from start scripts. Tons of other places can also be simplified, but to keep this PR small, this is probably a nice first step, then other can follow in other PRs.


janhoy:
Tests pass, @epugh can you have a first quick review and tell if this helps your CLI work? After your first feedback I can invite more reviewers.

I feel the principle is well proven by the fact that I removed explicit parsing of `SOLR_MODULES` variable in `bin/solr`, and still `test_modules.bats` pass tests both for `-Dsolr.modules` and `export SOLR_MODULES`.

I want to keep the scope of this PR somewhat limited. Can follow up with new PRs that remove cruft from the scripts.

solr/core/src/java/org/apache/solr/util/EnvUtils.java
```diff
@@ -0,0 +1,235 @@
...
+        init(false);
+      }
+    } catch (IOException e) {
+      throw new SolrException(
+          SolrException.ErrorCode.INVALID_STATE, "Failed loading env.var->properties mapping", e);
+    }
+  }
+
+  /**
+   * Get all environment variables with SOLR_ prefix.
```
janhoy, line 64:
```suggestion
   * Get Solr's mutable copy of all environment variables.
```

solr/core/src/java/org/apache/solr/util/ModuleUtils.java
```diff
@@ -56,17 +55,9 @@ public static Path getModuleLibPath(Path solrInstallDirPath, String moduleName)
...
-      mods.addAll(StrUtils.splitSmart(modulesFromProps, ',', true));
-    }
-    String modulesFromEnv = System.getenv("SOLR_MODULES");
-    if (StrUtils.isNotNullOrEmpty(modulesFromEnv)) {
-      mods.addAll(StrUtils.splitSmart(modulesFromEnv, ',', true));
-    }
-    return mods.stream().map(String::trim).collect(Collectors.toSet());
+    return EnvUtils.getPropAsList("solr.modules", Collections.emptyList()).stream()
+        .map(String::trim)
+        .collect(Collectors.toSet());
```
janhoy, line 60: No need for the `trim()` here, as `getPropAsList()` already trims values.

solr/bin/solr
```diff
@@ -112,6 +112,13 @@ elif [ -r "$SOLR_INCLUDE" ]; then
   . "$SOLR_INCLUDE"
 fi
 
+# Export all SOLR variables so they are visible to Solr sub-process
+for var in $(compgen -e); do
```
epugh, line 116: `compgen` is new to me...    I had to look a bit to learn, really interesting how short it works.

solr/bin/solr
```diff
@@ -1345,12 +1352,10 @@ if [ $# -gt 0 ]; then
         ;;
         -v)
             SOLR_LOG_LEVEL=DEBUG
-            PASS_TO_RUN_EXAMPLE+=("-Dsolr.log.level=$SOLR_LOG_LEVEL")
```
epugh, line 1348: Love this!

solr/core/src/java/org/apache/solr/util/EnvUtils.java
```diff
@@ -0,0 +1,235 @@
...
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.stream.Collectors;
+import org.apache.solr.common.SolrException;
+import org.apache.solr.common.util.StrUtils;
+import org.apache.solr.common.util.Utils;
+
+/**
+ * This class is a unified provider of environment variables and system properties. It exposes a
+ * mutable copy of environment variables with. It also converts 'SOLR_FOO' variables to system
```
epugh, line 39: "with"?

solr/core/src/resources/EnvToSyspropMappings.properties
epugh, line 1: I ❤️  this file!    Now imagine if you could find all the possible system property settings in Solr in one place?   Versus hunting through ref guide and source code!!!    Maybe someday...

solr/bin/solr
janhoy: Me too, ChatGPT surprised me there 😉

epugh:
Would then the preferred access mode for getting properties be via `EnvUtils.getProp(SolrDispatchFilter.SOLR_DEFAULT_CONFDIR_ATTRIBUTE)` pattern?   I could imagine in a future ticket getting rid of the System.getPRoperty and adding that to our list of disallowed methods ;-).     One more thought, how did you build the list of nonstandard mappings?

solr/core/src/java/org/apache/solr/util/EnvUtils.java
```diff
@@ -0,0 +1,235 @@
...
+    String value = ENV.get(key);
+    if (value == null) {
+      return defaultValue;
+    }
+    return Long.parseLong(value);
+  }
+
+  /** Get an env var as boolean */
+  public static boolean getEnvAsBool(String key) {
+    return Boolean.parseBoolean(ENV.get(key));
```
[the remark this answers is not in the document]
janhoy: Yea, it is in use. Only difference is it is case sensitive, see #1931

janhoy:
> Would then the preferred access mode for getting properties be via `EnvUtils.getProp(SolrDispatchFilter.SOLR_DEFAULT_CONFDIR_ATTRIBUTE)` pattern? I could imagine in a future ticket getting rid of the System.getPRoperty and adding that to our list of disallowed methods ;-). One more thought, how did you build the list of nonstandard mappings?

Such moves can be considered for sure. The `EnvUtils.getProp()` method does not bring much to the table, but the other variants such as `getPropAsList()` etc are helpful, but still a convenience on top of the props. I may have duplicated similar logic from other Solr Util classes, so happy to consolidate if you know of something.

epugh (approved):
This looks like a great step in the right direction!   I don't yet have like a immediate "hey, this totally solved one of the CLI issues", but I bet it helps.   Bigger is that in the future, maybe folks can use either system properties or environment variables to configure Solr without it being a battle ;-).

janhoy:
Widening the review to get more perspectives.

risdenk (approved):

solr/core/src/resources/EnvToSyspropMappings.properties
```diff
@@ -0,0 +1,60 @@
...
+SOLR_SSL_OPTS=
+SOLR_SSL_TRUST_STORE=
+SOLR_SSL_TRUST_STORE_PASSWORD=
+SOLR_SSL_TRUST_STORE_TYPE=
+SOLR_SSL_WANT_CLIENT_AUTH=
+SOLR_START_WAIT=
+SOLR_STOP_WAIT=
+
+# Parking lot
+# TODO: These seemed to cause issues with the BATS ssl tests...
```
janhoy, line 50: I think these are consumed by Jetty, so probably not very useful to remove them from bin/solr, since Jetty probably need these before our app is even started?

solr/core/src/java/org/apache/solr/util/EnvUtils.java
```diff
@@ -0,0 +1,235 @@
...
+  static {
+    try {
+      Properties props = new Properties();
+      try (InputStream stream =
+          EnvUtils.class.getClassLoader().getResourceAsStream("EnvToSyspropMappings.properties")) {
+        props.load(new InputStreamReader(Objects.requireNonNull(stream), StandardCharsets.UTF_8));
+        for (String key : props.stringPropertyNames()) {
+          CUSTOM_MAPPINGS.put(key, props.getProperty(key));
+        }
+        init(false);
```
janhoy, line 55: One question I have is whether there may be any race condition here, between `EnvUtils` static init and potential `System.getProperty()` access in static context in other Solr classes, such as populating static final strings. Not sure how/whether we can control ordering.

Of course, if we ban `System.getProperty()` we could have `getProp()` and friends block until init is done.

janhoy: Added two commits to address race condition and changed three static var initializations to use EnvUtils.getProp - to enable env->prop mapping. Can U have another look @epugh ?

solr/core/src/resources/EnvToSyspropMappings.properties
```diff
@@ -13,10 +15,15 @@ SOLR_TIMEZONE=user.timezone
 SOLR_TIP=solr.install.dir
 SOLR_TIP_SYM=solr.install.symDir
 SOLR_WAIT_FOR_ZK=waitForZk
+ZK_CLIENT_TIMEOUT=zkClientTimeout
+ZK_CREATE_CHROOT=createZkChroot
+ZK_HOST=zkHost
 
 # These should not be mapped to system properties
```
epugh, line 41: why should these not be mapped to system properties?  wouldn't you want to be able to set some of these that way??

solr/core/src/java/org/apache/solr/pkg/PackageAPI.java
```diff
@@ -58,7 +59,7 @@
 /** This implements the public end points (/api/cluster/package) of package API. */
 public class PackageAPI {
   public final boolean enablePackages =
-      Boolean.parseBoolean(System.getProperty("enable.packages", "false"));
+      EnvUtils.getPropAsBool("enable.packages", false);
```
epugh, line 62: So much more compact!

epugh:
This is all looking really great @janhoy ...     If the process of getting this done everywhere is a big job, make other jira's and maybe split them up by area so others can help?

solr/core/src/resources/EnvToSyspropMappings.properties
janhoy: The ones I marked as no map are either SSL props that must anyway be set in script because they are read by Jetty (?), or variables that are temp variables in the script (SOLR_OPTS etc) that have no real use as sysProps, but are still exported since we export everything `SOLR_*`

janhoy:
> This is all looking really great @janhoy ... If the process of getting this done everywhere is a big job, make other jira's and maybe split them up by area so others can help?

For sure my plan. Just want as many eyes as possible on the first foundation to detect flaws. When this is in, it also opens up for simplifying documentation, e.g.

```
SOLR_OPTS="$SOLR_OPTS -Dsolr.allowUrls=http://localhost:8983,http://localhost:8984"
```
can be supported as:
```
SOLR_ALLOW_URLS=http://localhost:8983,http://localhost:8984
```
...in either `solr.in.sh` or as an environment before starting Solr.

We also need to 
* Document ENV/Props naming convention in dev-guide
* Deprecate use of all unstandard names e.g. `enable.packages` -->  add alias `solr.packages.enable`

solr/core/src/resources/EnvToSyspropMappings.properties
HoustonPutman: The client SSL options (`SOLR_SSL_CLIENT_*`, `SOLR_SSL_CHECK_PEER_NAME`) are not read by the Jetty Server, they are read instead by the SolrClients. They should be mapped.

You are missing `SOLR_SSL_CLIENT_HOSTNAME_VERIFICATION` for the non-mapped options. This is a server option, so it will be read by the jetty config on startup.

HoustonPutman (commented):
I love this work Jan!

HoustonPutman:
> For sure my plan. Just want as many eyes as possible on the first foundation to detect flaws. When this is in, it also opens up for simplifying documentation, e.g.

This is an awesome start, and will be a huge boon to the project (or at least those working on things like the operator 😅 ). If we make `System.getProperty()` a forbidden API, it also makes it so that devs have to support EnvVars by default going forward, which I love.

> * Deprecate use of all unstandard names e.g. `enable.packages` -->  add alias `solr.packages.enable`. Defining and looking up aliases could also be part of `EnvUtils`.

Yeah I was thinking that this code would be a great way of cleaning up and deprecating bad names of variables. (We just need to give a way for the operator to support 2 major versions, so remove the bad variables in Solr 10, but have them deprecated and replaced in a mid-version of Solr 9, like 9.5).


One question about your envVar exporting in bin/solr. Is that command compatible across unix varieties?

solr/core/src/resources/EnvToSyspropMappings.properties
janhoy: Ah, so as long as `EnvUtils` is in solr core, it cannot be used in SolrJ. Perhaps `EnvUtils` should be moved to SolrJ?

janhoy: So I added `SOLR_SSL_CLIENT_HOSTNAME_VERIFICATION` to non-mapped, and added the CLIENT mappings back, have a look.

janhoy:
> One question about your envVar exporting in bin/solr. Is that command compatible across unix varieties?

Yea, both `compgen` and `export` are built-in commands of bash, and the start script requires bash.

janhoy:
I added some refguide docs and some examples in `solr.in.sh/cmd`.

solr/bin/solr
```diff
@@ -112,6 +112,13 @@ elif [ -r "$SOLR_INCLUDE" ]; then
   . "$SOLR_INCLUDE"
 fi
 
+# Export variables we want to make visible to Solr sub-process
+for var in $(compgen -e); do
+  if [[ "$var" =~ ^(SOLR_.*|DEFAULT_CONFDIR|ZK_.*|GCS_BUCKET|GCS_.*|S3_.*|OTEL_.*|AWS_.*)$ ]]; then
```
janhoy, line 117: By exporting more patterns we make sure users can add such vars to `solr.in.sh` and still get them exported to solr app without prefixing with `export` in solr.in.sh.

janhoy:
Camel case support is now better. `getProp("solr.fooBar")` now falls back to `solr.foo.bar`. Also setting env `SOLR_FOO_BAR` will set both `solr.fooBar` due to the unstandard mapping, and `solr.foo.bar`.

This makes it easier for us to either continue using some of the camelCase props in our code, since CAMEL_CASE will find them, or to deprecate and support both over some time. We'll still need some mappings if we want better names.

epugh:
This is very exciting.   And imagine in the future we could provide sane default configurations for `development` versus `production` deploys of Solr!    Kind of like what we did with `bin/solr auth` recently and the sane `security.json` setup.

solr/core/src/java/org/apache/solr/util/EnvUtils.java
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  /** Set all environment variables */
+  public static synchronized void setEnvs(Map<String, String> env) {
+    ENV.clear();
+    ENV.putAll(env);
+  }
+
+  /** Get all Solr system properties as a sorted map */
+  public static SortedMap<String, String> getProps() {
```
dsmiley, line 133: I assume sorting is for making the results pretty if printed.  But can we assume the caller is going to print them?  Not sure why we need this a tall actually.

solr/core/src/java/org/apache/solr/util/EnvUtils.java
```diff
@@ -0,0 +1,279 @@
...
+    String value = getPropWithCamelCaseFallback(key);
+    return value != null ? value : defaultValue;
+  }
+
+  /**
+   * Get a property from given key or an alias key converted from CamelCase to dot separated.
+   *
+   * @return property value or value of dot-separated alias key or null if not found
+   */
+  private static String getPropWithCamelCaseFallback(String key) {
```
dsmiley, line 163: We should put cached answers in a ConcurrentHashMap so that string manipulations needn't be re-parsed/resolved on every fetch (calling camelCaseToDotSeparated).  The query path might reasonably want to call getProp and it'd be annoying to see that for obscure features.

dsmiley (commented):
Excellent developer quality of life type of improvement!  I wish this was done a long time ago!

As we seek embracing this, be aware that Boolean.getBoolean(string) does a system property lookup, and so will similar methods on the other classes corresponding with primitives.

solr/core/src/java/org/apache/solr/util/EnvUtils.java
```diff
@@ -0,0 +1,279 @@
...
+        : envName.toLowerCase(Locale.ROOT).replace("_", ".");
+  }
+
+  @SuppressWarnings("unchecked")
+  private static List<String> stringValueToList(String string) {
+    if (string.startsWith("[") && string.endsWith("]")) {
+      // Convert a JSON string to a List<String> using Noggit parser
+      return (List<String>) Utils.fromJSONString(string);
+    } else {
+      return StrUtils.splitSmart(string, ",", true).stream()
```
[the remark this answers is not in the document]
janhoy: Hmm, don't recall this backslach decoding being a deilberate choice. Would it make more sense to flip to decode=false?

solr/core/src/java/org/apache/solr/util/EnvUtils.java
```diff
@@ -0,0 +1,279 @@
...
+          .map(String::trim)
+          .collect(Collectors.toList());
+    }
+  }
+
+  private static synchronized void ensureInitialized() {
```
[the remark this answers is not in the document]
janhoy: Since we're working with static global methods and state here, the initialization order is not well defined (unless we use some IOC stuff). Say you have a class declaring `public static final String MY_VAR = EnvUtils.getProp("solr.foo")`. Now, if user sets `SOLR_FOO` as env.var, then we need to make sure `EnvUtils.init()` has already run.

There may be better ways to achieve this than busy-waiting, ideas?

solr/core/src/java/org/apache/solr/util/EnvUtils.java
janhoy: I don't have a strong defence, but it is convenient to work with both in actual CLI/UI, in debug print and in tests, especially since the list of vars is so long.

solr/core/src/java/org/apache/solr/util/EnvUtils.java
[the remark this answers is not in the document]
janhoy: I tested both and they both work, so not sure it makes much of a difference in our case.

janhoy:
Thanks for your review @dsmiley  and sorry for late response. I updated PR.

janhoy:
In 28400bb72fa4bf06fbf7824b08a79da2ee3b55b1 I removed yet 20 lines from bin/solr :)

solr/core/src/java/org/apache/solr/util/EnvUtils.java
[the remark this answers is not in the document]
janhoy: Hmm, this bit me, cause I call `getProp` from within `init()` itself, and then we got a deadlock so Solr would not start at all. So it is a bit fragile, we must make sure there is no circular calls with the same lock, or find another safe way for late or async initialization..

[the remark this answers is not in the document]
dsmiley: I'm not quite following.  A static initializer is guaranteed to run to completion before any other threads may proceed to call into it (they block).  And you have a static initializer calling init(false).  init(true) is only used by some tests; we can ignore synchronization for them.

solr/core/src/java/org/apache/solr/util/EnvUtils.java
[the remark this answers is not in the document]
dsmiley: My suggestion is to make it clear.  Boolean args are sadly unnamed at the call sight in java.

solr/core/src/java/org/apache/solr/util/EnvUtils.java
```diff
@@ -0,0 +1,304 @@
...
+
+  /**
+   * Get a property as string with a fallback value. All other getProp* methods use this.
+   *
+   * @param key property key, which treats 'camelCase' the same as 'camel.case'
+   * @param defaultValue fallback value if property is not found
+   * @param waitForInit if true, wait for initialization to complete before returning. When calling
+   *     from init(), this should be false to avoid a deadlock.
+   */
+  public static String getProp(String key, String defaultValue, boolean waitForInit) {
```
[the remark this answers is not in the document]
janhoy: Made it private for now and a dedicated method named "noBlocking" for internal use from within init()

solr/core/src/java/org/apache/solr/util/EnvUtils.java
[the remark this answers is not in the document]
janhoy: For internal use in this one class that's ok. But when another class calls one of the `getProp` methods that require initialization, we have to block.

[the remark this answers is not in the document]
dsmiley: We agree blocking is necessary.  I claim the way a JVM loads classes with a static initializer will accomplish that without code specific in this class to try to accomplish similarly.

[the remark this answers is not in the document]
janhoy: The JVM does not know about our `init()` method, and that e.g. `getProp()` depends on `init()` having ran. We could expclicitly load this `EnvUtils` class early by calling it  at the earliest possible time, and JVM would make sure it was initialized before others. But locking is safer.

janhoy:
@HoustonPutman Thanks to your BATS ssl tests, we caught a problem in this PR. Needed to exclude all `SOLR_SSL_*` vars from auto conversion to SysProp.

solr/core/src/java/org/apache/solr/util/EnvUtils.java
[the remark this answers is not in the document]
dsmiley: The JVM will block any thread from calling `getProp` (or accessing `EnvUtils` in any way) until `EnvUtils` is initialized.  Initialization of this class calls init().  See [section 12.4.2 of the Java Language Specification](https://docs.oracle.com/javase/specs/jls/se8/html/jls-12.html#jls-12.4.2), particularly references to "LC".  

I'm aware init() indirectly calls getProp but it proceeds because this thread holds the LC lock.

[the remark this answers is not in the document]
janhoy: Hum, I’m sure it did not work before adding the lock, but I see what the spec claims here, so I’ll have to give it a try by removing locks.. Would indeed simplify all.

janhoy:
Disabled locking and tests pass. I seem to recall that the GlobalCircuitBreaker tests would break. I’ll try some beasting over Christmas..

janhoy:
Could not make the CB tests fail with beasting. So I fully removed all locking code. We'll soon discover if there are issues.

dsmiley (approved):
Thanks for getting back to something simpler :-)

janhoy:
Thanks for the extra review. Will merge to main tomorrow.

epugh:
I went to check if Ref Guide was updated, and sure enough it was!

PR #1941, opened 2023-09-20
SOLR-16925 Broken JSON indentation using JacksonJsonWriter

noblepaul:


solr/core/src/java/org/apache/solr/response/JacksonJsonWriter.java
```diff
@@ -25,11 +25,17 @@
...
 /** A JSON ResponseWriter that uses jackson. */
 public class JacksonJsonWriter extends BinaryResponseWriter {
+  private static final char[] indentChars = new char[81];
+
+  static {
+    Arrays.fill(indentChars, ' ');
+  }
 
   protected final JsonFactory jsonfactory;
   protected static final PrettyPrinter pretty =
```
HoustonPutman, line 41:
```suggestion
  protected static final DefaultPrettyPrinter pretty =
```

HoustonPutman (changes requested):
I'm not really sure how this fixing the issue (I also tested and the issue is still there). We are already using a pretty printer, but the issue is that the pretty printer is stateful.

So basically we need to create a new instance of PrettyPrinter for every json serialization, which we can do by using the `gen.setPrettyPrinter(pretty.createInstance())` method if `pretty` is a `DefaultPrettyPrinter` object.

noblepaul:
`I also tested and the issue is still there`

I'm sure it did fix the issue. Your solution works just fine. I'll close this and approve that

noblepaul:
closing this in favor of #1942

PR #1942, opened 2023-09-20
SOLR-16925: Fix indentation for JacksonJsonWriter

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-16925

`PrettyPrinter` is stateful, so we need a new instance every time.


noblepaul (approved):
LGTM

PR #1943, opened 2023-09-20
SOLR-16960 Tests should sometimes run with a Tracer (not no-op)

stillalex:
https://issues.apache.org/jira/browse/SOLR-16960

# Description

Took a stab at randomizing the isRecording flag during tests so we get better coverage.

# Solution

# Tests

# Checklist

### DIFF
--- a/solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
+++ b/solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
@@ -1308,7 +1308,7 @@ public class MiniSolrCloudCluster {
      * with span creation without having to enable otel module
      */
     private void injectRandomRecordingFlag() {
-      boolean isRecording = LuceneTestCase.random().nextBoolean();
+      boolean isRecording = LuceneTestCase.rarely();
       TraceUtils.IS_RECORDING = (ignored) -> isRecording;
     }
   }
--- a/solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
+++ b/solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
@@ -1303,9 +1303,17 @@ public class MiniSolrCloudCluster {
     }
 
     /**
-     * It only makes sense to call this if we are using the alwaysOn tracer. this will randomize the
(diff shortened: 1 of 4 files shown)


solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
```diff
@@ -1299,5 +1301,15 @@ public Builder withTraceIdGenerationDisabled() {
...
       return this;
     }
+
+    /**
+     * It only makes sense to call this if we are using the alwaysOn tracer. this will randomize the
+     * Span::isRecording check for the alwaysOn tracer so we have coverage of all methods that deal
+     * with span creation without having to enable otel module
+     */
+    private void injectRandomRecordingFlag() {
+      boolean isRecording = LuceneTestCase.random().nextBoolean();
```
dsmiley, line 1311: I believe there is a "rarely()" method or something like that.  I think it should be rare that we add this and not merely 50%.

dsmiley (approved):
+1 thanks.

Progress over perfection, -- this is good, but ideally we would inject the propagating tracer and potentially this recording flag in an independent way that isn't specific to SolrCloud.  Tracing isn't specific to SolrCloud!  In the mean time, it'd be nice if there was a comment to that effect.  People learn systems by reading code; someone might think tracing is SolrCloud specific seeing this (and be wrong).

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
stillalex: good suggestion. will move to rarely()

stillalex:
> but ideally we would inject the propagating tracer and potentially this recording flag in an independent way that isn't specific to SolrCloud. Tracing isn't specific to SolrCloud! In the mean time, it'd be nice if there was a comment to that effect. People learn systems by reading code; someone might think tracing is SolrCloud specific seeing this (and be wrong).

we don't disagree. I want to make this available to all testing code, and it's easy enough to do so, the current code serves as an example. what is confusing is not having a clear indication of what is missing. do you want to move this to a better location? please suggest a better place. there is so much testing code around I usually get lost in the options and pick one that looks 'good enough'.

stillalex:
@dsmiley fyi. reworded the javadocs. will merge once the build completes.

anshumg (approved):

PR #1945, opened 2023-09-21
SOLR-17004: ZkStateReader waitForState should check clusterState before using watchers

risdenk:
https://issues.apache.org/jira/browse/SOLR-17004

# Description

Trying to get the leader with retry ends up using a lot of watches and locks on collection state watchers. This ends up being contentious and slows down processing. In most cases we can pull the leader right from the `clusterState` and fall back to the slower logic if the leader is not setup yet.

# Solution

Update waitForState to check the predicate against the known clusterState before falling back to the retry logic.

### DIFF
--- a/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
+++ b/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
@@ -936,7 +936,19 @@ public class ZkStateReader implements SolrCloseable {
   /** Get shard leader properties, with retry if none exist. */
   public Replica getLeaderRetry(String collection, String shard, int timeout)
       throws InterruptedException {
-    AtomicReference<DocCollection> coll = new AtomicReference<>();
+
+    // Try to get from cluster state without retry first, otherwise fallback to retry logic
+    if (clusterState != null) {
+      DocCollection docCollection = clusterState.getCollectionOrNull(collection);
+      if (docCollection != null) {
+        Replica leader = docCollection.getLeader(shard);
+        if (leader != null) {
+          log.debug("leader found for {}/{} to be {}", collection, shard, leader);
+          return leader;
+        }
+      }
+    }
+
     AtomicReference<Replica> leader = new AtomicReference<>();
     try {
       waitForState(
@@ -945,7 +957,6 @@ public class ZkStateReader implements SolrCloseable {
           TimeUnit.MILLISECONDS,
           (n, c) -> {
             if (c == null) return false;
-            coll.set(c);
             Replica l = getLeader(n, c, shard);
             if (l != null) {
               log.debug("leader found for {}/{} to be {}", collection, shard, l);
--- a/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
+++ b/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
@@ -1800,6 +1800,7 @@ public class ZkStateReader implements SolrCloseable {
       throw new AlreadyClosedException();
     }
 
+    // Check predicate against known clusterState before trying to add watchers
     if (clusterState != null) {
       Set<String> liveNodes = clusterState.getLiveNodes();
       DocCollection docCollection = clusterState.getCollectionOrNull(collection);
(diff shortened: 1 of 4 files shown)


risdenk:
I'd appreciate some thoughts on this. This doesn't fix the `setupRequest` being called twice, but should make the `setupRequest` method much faster by not having to lock and check for leader in one concurrenthashmap. Basically the normal happy path should be faster - if there are nodes going up/down/state changing then logic should fall back to the slower default.

risdenk:
Another potential improvement is to use caffeinecache for collectionwatches due to performance being way better on compute - https://github.com/ben-manes/caffeine/wiki/Benchmarks#compute

risdenk:
I ran into this reproducing failure - `gradlew test --tests LeaderElectionIntegrationTest.testSimpleSliceLeaderElection -Dtests.seed=7FE90E71BD77ABA3 -Dtests.locale=ga-Latn-IE -Dtests.timezone=Hongkong -Dtests.asserts=true -Dtests.file.encoding=UTF-8`

which looks to be an issue in the `LeaderElectionIntegrationTest` - it was passing previously because the leader retry was waiting for a zookeeper state change. The improvement I made checks the ZK state directly and there is no wait for a state change. 

[d1f1554](https://github.com/apache/solr/pull/1945/commits/d1f1554aab2aa945a6ae50e9c583a4773598e67b) improves the test to make sure Jetty actually shutdown before trying to find a new leader. There is no guarantee prior to this change that a leader change has taken place before Jetty has shutdown completely.

risdenk:
I fixed a similar race condition in [269d19b](https://github.com/apache/solr/pull/1945/commits/269d19b3a95b8de35130020facc8c47fab391c77) that makes sure there is a new leader before we try to find one again.

psalagnac:
This seems to be a promising performance improvement.

Now, I wonder if this should be extended to the generic `waitForState()` method instead of only `getLeaderWithRetry()`. We could check the predicate first once without creating watchers, and fallback to current mechanism if the predicate is not true.

patsonluk:
> This seems to be a promising performance improvement.
> 
> Now, I wonder if this should be extended to the generic `waitForState()` method instead of only `getLeaderWithRetry()`. We could check the predicate first once without creating watchers, and fallback to current mechanism if the predicate is not true.

+1 on this. I think in general we can quickly check the `clusterState` field, give that to the `predicate` and see if it's happy, if not then proceed down to the "install watch" path? 😊

risdenk:
Agreed thought about that as I was looking at this more. Will update tomorrow

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -1802,6 +1800,18 @@ public void waitForState(
       throw new AlreadyClosedException();
     }
 
+    // Check predicate against known clusterState before trying to add watchers
+    if (clusterState != null) {
+      Set<String> liveNodes = clusterState.getLiveNodes();
+      DocCollection docCollection = clusterState.getCollectionOrNull(collection);
+      if (liveNodes != null && docCollection != null) {
+        if (predicate.matches(liveNodes, docCollection)) {
+          log.debug("Found {} directly in clusterState", predicate);
```
psalagnac, line 1809: Probably the `toString()` output of the predicate won't be relevant most of the time. I'm not sure what is the usual practice in Solr for this.

psalagnac (approved):
Looks good to me.
I just did a minor comment on logging.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
risdenk: Hmmm so I see what you are saying @psalagnac. I took inspiration from line 1862 where log.debug is doing basically the same logic in the other wait for state.

risdenk:
I moved this over to https://issues.apache.org/jira/browse/SOLR-17004 since it really didn't change setupRequest at all.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
risdenk: I'm going to leave this as is for now.

risdenk:
I put changes on main for now. I plan on merging this once tests pass again locally. I'll let it sit on main for a few days then look at backporting. Not sure it will make 9.4.

PR #1950, opened 2023-09-22
SOLR-16994: facilitate date-related FieldValueFeature extensions (docValues=True only)

cpoerschke:
https://issues.apache.org/jira/browse/SOLR-16994

### DIFF
--- a/solr/modules/ltr/src/java/org/apache/solr/ltr/feature/FieldValueFeature.java
+++ b/solr/modules/ltr/src/java/org/apache/solr/ltr/feature/FieldValueFeature.java
@@ -136,7 +136,7 @@ public class FieldValueFeature extends Feature {
      * @param val value of the field
      * @return value after transformation
      */
-    protected long interprete_NumericDocValues_NumberType_DATE(long val) {
+    protected long readNumericDocValuesDate(long val) {
       return val;
     }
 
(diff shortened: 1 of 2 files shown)


cpoerschke:
Tagging @tomglk and @alessandrobenedetti as potential reviewers, and reviews from others welcome too, of course. Thank you!

stillalex:
just a comment from the peanut gallery (will leave the review for the experts).

this method `interprete_NumericDocValues_NumberType_DATE` feels very specific, but at the same time it doesn't have to be. on the surface it looks like you want to post-process the value. the default return from `readNumericDocValues ` is already the `docValues.longValue();`, so you don't really need the type check.
so just a minor suggestion: why not add a generic wrapper method (name is bad I know)
```
protected long postProcess(long val) {
  return val;
}
```

and just wrap any default return value in it
```
// just take the long value
return postProcess(docValues.longValue());
```

this way it's easy to override if needed.

dsmiley:
I've not seen underscores in method names in our codebase before.  I think you could call this method simply readNumericDocValuesDate.  I just added "Date" onto the existing method that calls it.

Alex's feedback looks good too; another way to view this.

cpoerschke:
Thanks @stillalex and @dsmiley for your feedback, both really resonated with me. I like `readNumericDocValuesDate` as a name, thanks for that suggestion!

During development I'd used `longToLong` as a temporary name (obviously can't use that for real) and wondered back and forth about making or not making it specific to the `NumberType.DATE` scenario. I saw a few challenges with a more generic approach:
* how to convey that for non-float and non-double doc value types the post-processing is available but for float and double doc value types it is not available?
* how to convey that for non doc value types the post-processing in not available? or how to make it available?
* if making post-processing available more widely, is there a use case for it and how to avoid confusion with the normalisation that can be used at the model level? -- https://solr.apache.org/guide/solr/latest/query-guide/learning-to-rank.html#normalizer

So having _doc values_ and _date_ in the name, somehow, addresses those challenges and also leaves the door open for `readNumericDocValues[Float,Double,...]` additions in future if there's a use case.

PR #1951, opened 2023-09-22
Port defaulting should be the same for all tools

HoustonPutman:
If Solr processes need to be found, port scanning will always be used.

I need to make a JIRA for this, but this has been exhausting, so it can wait till later.

Basically any Solr tool accepts `-p`, `-host` and `-urlScheme`, and can default the Solr URL from that. If a port is not provided, it will use `SOLR_PORT` if that is explicitly set beforehand. Otherwise, it will scan processes for a Solr Port. If more than one process is found, it will fail.

Similarly the tools (`stop -all`, `status`) that find all running Solr instances will use port scanning instead of PIDs.

This is somewhat back-incompat, for tools that use to use port scanning instead of `SOLR_PORT`, but I think this is acceptable. This can also wait for 10.0 if we want to, but I don't think that's necessary.

This is a WIP and likely needs more tests, and possibly there are some tools that I forgot.

### DIFF
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -148,7 +148,7 @@ fi
 : "${SOLR_START_WAIT:=$SOLR_STOP_WAIT}" # defaulting to $SOLR_STOP_WAIT for backwards compatibility reasons
 
 # Store whether a solr port was explicitly provided, for the "solr stop" command.
-PROVIDED_SOLR_PORT="${SOLR_PORT:-}"
+PROVIDED_SOLR_PORT_ENV="${SOLR_PORT:-}"
 : "${SOLR_PORT:=8983}"
 export SOLR_PORT
 
@@ -626,6 +626,7 @@ function jetty_port() {
 # run a Solr command-line tool using the SolrCLI class;
 # useful for doing cross-platform work from the command-line using Java
 function run_tool() {
+  export SOLR_PORT="$(choose_default_solr_port)"
 
   # shellcheck disable=SC2086
   "$JAVA" $SOLR_SSL_OPTS $AUTHC_OPTS ${SOLR_ZK_CREDS_AND_ACLS:-} ${SOLR_TOOL_OPTS:-} -Dsolr.install.dir="$SOLR_TIP" \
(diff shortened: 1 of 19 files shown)


janhoy:
Wow, huge effort. Love consistency! Will this need changes in `solr.cmd` as well? 🫣
Will try to circle back to this and take it for a spin. It's great having the bats tests to guard against regressions here!

epugh:
I absolutely love all this clean up!   I'll be honest, all the flavours of variables have been a bit confusing ;-)...   I think some accretion over time of concepts layered on concepts...

So one thing I'll ask, "Is any of this required?"...   I suspect that I don't quite understand how some folks use Solr...   I pretty much only ever run one Solr per VM/image with proxies in front etc, so in my world, Solr is always http, Solr is always localhost (127.0.01?), and Solr is always port 8983 when I am on that node.   I had actually been hoping to move in the direction of eliminating the whole port, schema, host variables, and just using a url everywhere....    I want to encourage us to be able to use the various commands like `bin/solr create` from anywhere, which would imply that you pretty much only ever do a `bin/solr [your_command] --solrUrl https://myserverincloud:50678 -u user:password` instead of the current way of looking up local variables.   This pattern may come from I generally stand up a solr as step one, and then externally, as step 2, interact with it.

Regardless, this is all amazing work!

solr/bin/solr
```diff
@@ -148,7 +148,7 @@ fi
 : "${SOLR_START_WAIT:=$SOLR_STOP_WAIT}" # defaulting to $SOLR_STOP_WAIT for backwards compatibility reasons
 
 # Store whether a solr port was explicitly provided, for the "solr stop" command.
-PROVIDED_SOLR_PORT="${SOLR_PORT:-}"
+PROVIDED_SOLR_PORT_ENV="${SOLR_PORT:-}"
```
epugh, line 151: is `_ENV` a pattern we use elsewhere in `bin/solr`?   For these variables?  If we have similar concepts, maybe we can get all of the variables following this pattern?

solr/bin/solr
```diff
@@ -626,6 +626,7 @@ function jetty_port() {
 # run a Solr command-line tool using the SolrCLI class;
 # useful for doing cross-platform work from the command-line using Java
 function run_tool() {
+  export SOLR_PORT="$(choose_default_solr_port)"
```
epugh, line 629: would `(choose_default_solr_port())` be possible to signal it's a function?   just a spitball...

solr/bin/solr
```diff
@@ -636,98 +637,76 @@ function run_tool() {
...
-    fi
+    echo -e "\nNo Solr nodes are running.\n" >&2
   fi
+}
 
-} # end get_status
-
-function run_package() {
-  runningSolrUrl=""
+function get_all_running_solr_ports() {
```
epugh, line 651: I love the breaking down these convulted functions into smaller ones, makes it easier to read and reason!   Hell, maybe someday we bats test the functions?

solr/bin/solr
```diff
@@ -636,98 +637,76 @@ function run_tool() {
...
-          if [ "$port" != "" ]; then
-            echo ""
-            echo "Solr process $ID running on port $port"
-            runningSolrUrl="$SOLR_URL_SCHEME://$SOLR_TOOL_HOST:$port/solr"
-            break
-            CODE=$?
-            echo ""
-          fi
-      done
+    # no-commit - check if the SOLR_TOOL_HOST is the localhost, otherwise don't do this
```
epugh, line 675: I never quite understood what `SOLR_TOOL_HOST` gave us that `SOLR_HOST` doesn't!

epugh: Assuming that you are running your tool on the same node as a Solr and therefore HAVE a `SOLR_HOST`.

solr/bin/solr
```diff
@@ -739,36 +718,31 @@ function stop_solr() {
...
+  echo -e "Sending stop command to Solr running on port $SOLR_PORT ... waiting up to $SOLR_STOP_WAIT seconds to allow Jetty process $SOLR_PID to stop gracefully."
+  # shellcheck disable=SC2086
+  "$JAVA" $SOLR_SSL_OPTS $AUTHC_OPTS ${SOLR_TOOL_OPTS:-} -jar "$DIR/start.jar" "STOP.PORT=$THIS_STOP_PORT" "STOP.KEY=$STOP_KEY" --stop || true
+    (loops=0
+    while true
+    do
+      # Check if a process is running with the specified PID.
+      # -o stat will output the STAT, where Z indicates a zombie
+      # stat='' removes the header (--no-headers isn't supported on all platforms)
+      # Note the space after '$('. It is needed to avoid confusion with special bash eval syntax
```
epugh, line 730: An this is why I ❤️  bash..   ;-).

solr/bin/solr
```diff
@@ -1017,7 +991,7 @@ if [[ "$SCRIPT_CMD" == "package" ]]; then
       esac
     done
   fi
-  run_package "$@"
+  run_tool package "$@"
```
epugh, line 994: thank you!   This has been on my backlog for a while to fix...    I think you can delete `run_package` now too!

solr/bin/solr
```diff
@@ -1143,17 +1132,7 @@ if [[ "$SCRIPT_CMD" == "auth" ]]; then
...
-      do
-        port=$(jetty_port "$ID")
-        if [ "$port" != "" ]; then
-          AUTH_PORT=$port
-          break
-        fi
-      done
-  fi
-  run_tool auth "${AUTH_PARAMS[@]}" -solrUrl "$SOLR_URL_SCHEME://$SOLR_TOOL_HOST:${AUTH_PORT:-8983}/solr" -authConfDir "$SOLR_HOME" $VERBOSE
+  run_tool auth "${AUTH_PARAMS[@]}" -authConfDir "$SOLR_HOME" $VERBOSE
```
epugh, line 1135: You can see where I was struggling to figure out how to eliminate the SOLR_URL_SCHEME and firends, and yet it was sneaking back in.   This at least is a lot shorter!

epugh: Also, I was noticing in `bin/solr auth` getting the warning about the solrUrl not needing `/solr` and this fixes it!

solr/core/src/java/org/apache/solr/cli/ApiTool.java
```diff
@@ -56,24 +56,38 @@ public List<Option> getOptions() {
             .argName("URL")
             .hasArg()
             .required(true)
-            .desc("Send a GET request to a Solr API endpoint.")
-            .build());
+            .desc("Send a GET request to a Solr API endpoint, or path to use with the default Solr URL.")
```
epugh, line 59: nice tweak to the tool, in a world where you can assume your solr!

solr/core/src/java/org/apache/solr/cli/ApiTool.java
```diff
@@ -56,24 +56,38 @@ public List<Option> getOptions() {
...
+        SolrCLI.OPTION_SOLRHOST,
+        SolrCLI.OPTION_URLSCHEME,
+        SolrCLI.OPTION_VERBOSE);
   }
 
   @Override
   public void runImpl(CommandLine cli) throws Exception {
     String response = null;
     String getUrl = cli.getOptionValue("get");
     if (getUrl != null) {
```
epugh, line 73: Do we need this check?   If the get parameter is marked required in the ocnfiguration, I would assume you wouldn't get to here if it is missing!

solr/core/src/java/org/apache/solr/cli/CreateTool.java
```diff
@@ -131,7 +134,7 @@ public void runImpl(CommandLine cli) throws Exception {
 
   protected void createCore(CommandLine cli, SolrClient solrClient) throws Exception {
     String coreName = cli.getOptionValue("name");
-    String solrUrl = cli.getOptionValue("solrUrl", SolrCLI.getDefaultSolrUrl());
+    String solrUrl = cli.getOptionValue("solrUrl", SolrCLI.getDefaultSolrUrl(cli));
```
epugh, line 137: Out of curiosity, if `SolrCLI.getDefaultSolrUrl(cli))` always takes a cli, should we make it be `cli.getDefaultSolrUrl()` instead?

solr/core/src/java/org/apache/solr/cli/PackageTool.java
```diff
@@ -310,6 +310,18 @@ private Pair<String, String> parsePackageVersion(String arg) {
   public List<Option> getOptions() {
     return List.of(
         SolrCLI.OPTION_SOLRURL,
+        // Cannot use the default OPTION_SOLRPORT, because -p is already taken by PARAMS
```
epugh, line 313: Yeah...   I wonder if we should make the slightly breaking chagne?   This feels like we are going to get bit by this in the future?  Or, do we think humans will never actually use teh `-p` parameter to mean a solr port, becuase it's a config file type thing...   I could see just forcing --package, cause how often do you type this command?  whereas -p is super common IF we think humans will use these parameters a lot.  <-- did that make sense?

solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
```diff
@@ -869,6 +857,7 @@ protected boolean isValidConfig(File configsetsDir, String config) {
   }
 
   protected Map<String, Object> getNodeStatus(String solrUrl, int maxWaitSecs) throws Exception {
+    solrUrl += "/solr";
```
epugh, line 860: bummer!  But we are in an awkward place where a solrUrl can end with /solr or with /api ;-)  So maybe this pattern becomes more common in varous tools...  Actually, are you sure you need this?  Ithought the status tool would deal with the solrURL.....

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
```diff
@@ -80,7 +80,7 @@ public class SolrCLI implements CLIO {
   public static final Option OPTION_ZKHOST =
       Option.builder("z")
           .longOpt("zkHost")
-          .argName("HOST")
```
epugh, line 83: nice

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
```diff
@@ -91,12 +91,45 @@ public class SolrCLI implements CLIO {
...
+                  + '.')
+          .build();
+  public static final Option OPTION_URLSCHEME =
+      Option.builder("urlScheme")
+          .longOpt("url-scheme")
+          .argName("urlScheme")
+          .hasArg()
+          .required(false)
+          .desc(
+              "Solr URL Scheme, which will be used, along with SOLR_TOOL_HOST and SOLR_PORT, to determine the Solr URL to connect to if a solrURL is not provided; defaults to: "
```
epugh, line 131: SOLR_HOST instead of SOLR_TOOL_HOST?  again, I don't quite get the difference ;-).

solr/packaging/test/test_create_collection.bats
```diff
@@ -50,6 +50,33 @@ teardown() {
...
+  export SOLR_PORT=
+  export SOLR_HOST=
+  export SOLR_TOOL_HOST=
+
+  run solr create -c "COLL_NAME" -p ${tmp_port} --host localhost
+  assert_output --partial "Created collection 'COLL_NAME'"
+  assert_output --partial "assuming solrUrl is http://localhost:${tmp_port}"
+}
+
+@test "create collection with port scanning" {
```
epugh, line 72: Is port scanning really a feature we want?   Do people just randomly launch solr on random port, and then scan to find it?   If poeple do, then great feature, and I LOVE the test...   Just kind of wondering about that use case...   (Look, I used the word "use case" ;-). )

solr/packaging/test/test_packages.bats
```diff
@@ -29,7 +29,12 @@ teardown() {
 }
 
 @test "package detects no running solr" {
-  # not sure this is actually a good thing..  we may not want this..
```
epugh, line 32: I kind of want to keep this comment...   There is an issue open to look at this.   Why do we check if solr is running on this command?

solr/packaging/test/test_status.bats
```diff
@@ -43,7 +43,6 @@ teardown() {
 @test "status shell script ignores passed in -solrUrl cli parameter from user" {
   solr start
   run solr status -solrUrl http://localhost:9999/solr
-  assert_output --partial "needn't include Solr's context-root"
```
epugh, line 46: ?  no longer germaine?   as in we're checking something that no longer needs checking because this is how it works?

HoustonPutman:
@epugh and @janhoy is this something we want to get done for Solr 10?

epugh:
This is the natural boundary! Also, it would be nice if we could get to done done on kicking around with our cli so we can have some long term patterns and aren't constantly changing the code.  This appears to go a long way in that direction.

janhoy:
Yea, consistency is good. When you say port scanning, what really goes on isn’t scanning all 64k ports but rather listing all processes, filtering on Java/solr processes, parsing -Djetty.port from the args. This was moved to the Java class SolrProcesses.

Sometimes we can add too much magic. Not sure if it’s right to not require a `-p` if we happen to find Solr running on some none standard port, just because it’s the only Solr process currently running on the host. That’s not deterministic if there used to be two Solrs and one is temporarily down.

I see lots of pr comments, not sure if I’m able to put much time in a thorough review..

HoustonPutman:
> Sometimes we can add too much magic. Not sure if it’s right to not require a -p if we happen to find Solr running on some none standard port, just because it’s the only Solr process currently running on the host. That’s not deterministic if there used to be two Solrs and one is temporarily down.

Yeah I very much agree with this sentiment. 

I will work on this a bit to simplify everything down to requiring -p or defaulting the same way

janhoy:
@HoustonPutman I looked through open PRs where I'm assigned as reviewer and found this. 
Lots has happend on the CLI space in parallel for the 10.0 release and after, and the PicoCLI effort will change it even more.

So having drifted a lot, it would be sad to see good things here just disappear.

Not saying you should prioritize this, but perhaps you can highlight some key improvements that you'd like to salvage from this PR, perhaps into a new one?

PR #1952, opened 2023-09-23
Clean up some deprecations and warnings in jwt-auth module

janhoy:
(Jo JIRA)

Basically followed recommendations from IDEA.
We still use Apache HTTPClient in some tests, did not cross that bridge now


epugh (approved):
LGTM.   I wish there were ways for these types of nice cleanups to get across the entire codebase faster.    And we update our linting stuff to prevent these patterns coming back....   I think back to that work that @risdenk did a while ago that just helped us so much...

PR #1954, opened 2023-09-24
SOLR-14496: Add Basic Auth support to SolrCLI

epugh:
https://issues.apache.org/jira/browse/SOLR-14496

# Description
Add a new commmand line parameter `-credentials` (sometimes known as `-u`) in the format of: `-u eric:password` to enable basic auth support.

# Solution

Lots and LOTS of sprinkling of the `credentials` parameter all over the code ;-(.   I thought there would be a much stronger single place to create a SolrClient, where I could add the credentials, and then it would all work.   However that seems to not be the case, we create them all over the place in many different mways.

I *did* discover the use of the magic variables in the solr config file that via some sort of "inject/systemproperty magic" allows basic auth to work on the actual solr box!   It's somehow enabled with `bin/solr auth` command...   See the bats test with how I hacked a workaround to NOT make that happen so I could test.   

# Tests

Bats!

# Checklist

- [x ] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [x ] I have created a Jira issue and added the issue ID to my pull request title.
- [x ] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [x ] I have developed this patch against the `main` branch.
- [ x] I have added tests for my changes.


solr/core/src/java/org/apache/solr/cli/DeleteTool.java
```diff
@@ -92,7 +94,13 @@ public void runImpl(CommandLine cli) throws Exception {
     SolrCLI.raiseLogLevelUnlessVerbose(cli);
     String solrUrl = SolrCLI.normalizeSolrUrl(cli);
 
-    try (var solrClient = SolrCLI.getSolrClient(solrUrl)) {
+    System.out.println("HEre are teh options:");
```
stillalex, line 97: weird casing on this message, and typo ;)

solr/core/src/java/org/apache/solr/cli/DeleteTool.java
```diff
@@ -102,13 +110,18 @@ public void runImpl(CommandLine cli) throws Exception {
   }
 
   protected void deleteCollection(CommandLine cli) throws Exception {
+    Http2SolrClient.Builder builder =
+        new Http2SolrClient.Builder()
+            .withIdleTimeout(30, TimeUnit.SECONDS)
+            .withConnectionTimeout(15, TimeUnit.SECONDS);
+    if (cli.hasOption("credentials")) {
```
stillalex, line 117: I think I wrote this 'if' a hundred times over. would it be simpler to have a `withOptionalBasicAuthCredentials` on the builder that would ignore null values instead of throwing some error? it would also allow for cleaner method chaining.

stillalex:
re. this pattern of parsing `user:password`

```
if (credentials != null && credentials.indexOf(":") > 0) {
      String[] credentialsArray = credentials.split(":");
```

what happens if the password itself contains `:` (I don't think the username is allowed to but I may be wrong). would it be a safer bet to break the string in 2 parts on the first occurrence?

solr/core/src/java/org/apache/solr/cli/DeleteTool.java
epugh: So, for some reason, for me, I am not getting the `-credentials` parameters in this tool.   I wish I had a junit test instead of just bats...   This is debugging...

solr/core/src/java/org/apache/solr/cli/DeleteTool.java
epugh: I would be up for that...   Definitly need some sort of better way of dealing with this.   Maybe some more refactoring and move the method to `SolrCLI` instead of it being all over the place on the individual tools?

epugh:
> re. this pattern of parsing `user:password`
> 
> ```
> if (credentials != null && credentials.indexOf(":") > 0) {
>       String[] credentialsArray = credentials.split(":");
> ```
> 
> what happens if the password itself contains `:` (I don't think the username is allowed to but I may be wrong). would it be a safer bet to break the string in 2 parts on the first occurrence?

Good call.    Maybe a util method somehwere that every class can use it?

solr/core/src/java/org/apache/solr/cli/DeleteTool.java
epugh: Take a looksee at what I did?   THoughts?

epugh:
> re. this pattern of parsing `user:password`
> 
> ```
> if (credentials != null && credentials.indexOf(":") > 0) {
>       String[] credentialsArray = credentials.split(":");
> ```
> 
> what happens if the password itself contains `:` (I don't think the username is allowed to but I may be wrong). would it be a safer bet to break the string in 2 parts on the first occurrence?
> 
> [edit] turns out the solution to this was very simple, use `credentials.split(":",2)`. this will force the split into 2 parts only.

your fix is nicer than mine!

solr/core/src/java/org/apache/solr/cli/DeleteTool.java
epugh: so, my unit test passes, but not the bats test.  sigh.

epugh:
Holy smokes, almost everything is building/running cleanly, and that crave failure of `org.apache.solr.cli.DeleteToolTest.testFailsToDeleteProtectedCollection` runs locally just fine..  (?)...    I am down to messing with the PackageTool for https, and then will be looking for reveiw and some thumbs up.

epugh:
Okay, I'd love some review on this, I think it's more or less ready!

stillalex:
the gradle precommit build is unhappy about something. did you already take a look at that?

solr/core/src/java/org/apache/solr/cli/ApiTool.java
```diff
@@ -57,27 +57,28 @@ public List<Option> getOptions() {
...
+        SolrCLI.OPTION_CREDENTIALS);
   }
 
   @Override
   public void runImpl(CommandLine cli) throws Exception {
     String response = null;
     String getUrl = cli.getOptionValue("get");
     if (getUrl != null) {
-      response = callGet(getUrl);
+      response = callGet(getUrl, cli.getOptionValue("credentials"));
```
stillalex, line 69: wondering if instead of hardcoding the name to `"credentials"` everywhere we could actually use `cli.getOptionValue(SolrCLI.OPTION_CREDENTIALS.getOpt())` instead.

solr/core/src/java/org/apache/solr/cli/AssertTool.java
```diff
@@ -123,6 +123,14 @@ public List<Option> getOptions() {
         Option.builder("e")
             .desc("Return an exit code instead of printing error message on assert fail.")
             .longOpt("exitcode")
+            .build(),
+        // u was taken, can we change that instead?
+        Option.builder("credentials")
```
stillalex, line 128: excuse the possibly dumb question. why is SolrCLI.OPTION_CREDENTIALS not available to use here?

solr/core/src/java/org/apache/solr/cli/PackageTool.java
```diff
@@ -342,6 +341,14 @@ public List<Option> getOptions() {
             .required(false)
             .desc("Don't prompt for input; accept all default choices, defaults to false.")
             .longOpt("noprompt")
+            .build(),
+        // u was taken, can we change that instead?
+        Option.builder("credentials")
```
stillalex, line 346: same as above, why not `SolrCLI.OPTION_CREDENTIALS`?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -1311,6 +1311,22 @@ public Builder withProxyConfiguration(
...
     }
+
+    /**
+     * Setup basic authentication from a string formatted as username:password. If the string is
+     * Null then it doesn't do anything.
+     *
+     * @param credentials The username and password formatted as username:password
+     * @return this Builder
+     */
+    public Builder withOptionalBasicAuthCredentials(String credentials) {
```
stillalex, line 1322: if there is no check for existence of `:` at least an `assert credentials.contains(":")` would be good

solr/core/src/java/org/apache/solr/cli/ExportTool.java
```diff
@@ -202,7 +209,16 @@ DocsSink getSink() {
     abstract void exportDocs() throws Exception;
 
     void fetchUniqueKey() throws SolrServerException, IOException {
-      solrClient = new CloudHttp2SolrClient.Builder(Collections.singletonList(baseurl)).build();
+      Http2SolrClient.Builder builder =
+          new Http2SolrClient.Builder()
+              .withIdleTimeout(30, TimeUnit.SECONDS)
```
stillalex, line 214: while it seems reasonable for Create and Delete to have the timeout values you used here, I am wondering if the Export is not too short? don't really have a good idea, just raising this for evaluation again.

solr/core/src/java/org/apache/solr/cli/DeleteTool.java
stillalex: sorry missed this message. looks good!

stillalex:
looks good to me. left a few minor notes.

solr/core/src/java/org/apache/solr/cli/ApiTool.java
epugh: i like that...   It is probably something we should do everywhere.   I always struggle with "forbiddenapis" stuff, but you could make a case that "cli.getOptionValue(String)" shouldn't be permitted...

solr/core/src/java/org/apache/solr/cli/AssertTool.java
epugh: Because the `u` short option is already used elsewhere in the assert tool....

solr/core/src/java/org/apache/solr/cli/PackageTool.java
epugh: likewise, the `u` already used...

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
epugh:
```
throw new IllegalStateException(
              "Invalid Authentication credentials. Either both username and password or none must be provided");
```
this appears to be a pattern in other builders...

epugh:
So, one thing I noticed is that the method `withOptionalBasicAuthCredentials` is ONLY on Http2SolrClient.Builder.  It isn't on the `HttpSolrClient builder`.  Or on some of the other variations like `LBHttpSolrClient` and `LB2HttpSolrClient`.   Maybe that's okay..   And if we need basicauth on more builders in the same manner, well, maybe thats for another talk.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
epugh: See what I added!

solr/core/src/java/org/apache/solr/cli/ExportTool.java
epugh: Good question.  I removed them since there are some defaults defined.....

solr/core/src/java/org/apache/solr/cli/AssertTool.java
stillalex: ah yeah, and I missed your comment too 👍

HoustonPutman:
Will this still support using `-Dbasicauth` and `-Dsolr.httpclient.builder.factory` in `SOLR_TOOL_OPTS`?

epugh:
> Will this still support using `-Dbasicauth` and `-Dsolr.httpclient.builder.factory` in `SOLR_TOOL_OPTS`?

It *should*, however, to be honest, I've never used those parameters...    Do we have any BATS tests that cover that?   I know you were working on some PR's in this area.    Whats the best way to test this?   Fire up local solr, run the bin/auth tool, and then confrim that not passing in `-u` still works?

epugh:
> `Dbasicauth`

I looked at `test_ssl.bats` and it has these lines:

```
  solr start -c
  solr auth enable -type basicAuth -credentials name:password
  solr assert --started https://localhost:${SOLR_PORT}/solr --timeout 5000
```

Notice the assert does not have the `-u` or `-credentials` parameter?   Yet we are successfully asserting the start?

epugh:
Still need to update the Ref Guide...  Hoping to get clarity on if we can re-purpose the `-u` parameter ;-)

epugh:
Ref Guide is updated!   I think I'm ready for final review on this....   Are there any other tools/cli items that *should* support Basic auth that I missed?

solr/solr-ref-guide/modules/deployment-guide/pages/solr-control-script-reference.adoc
```diff
@@ -577,6 +578,16 @@ This parameter is unnecessary if `ZK_HOST` is defined in `solr.in.sh` or `solr.i
...
 *Example*: `bin/solr healthcheck -z localhost:2181`
 
+`-u <user:pass>` or `--credentials <user:pass>`::
++
+[%autowidth,frame=none]
+|===
+|Optional | Default: none
+|===
++
+Basic authentication username and password seperated by a `:` character.
```
stillalex, line 588: `separated`?

solr/solr-ref-guide/modules/deployment-guide/pages/solr-control-script-reference.adoc
```diff
@@ -734,6 +745,16 @@ Unnecessary if `ZK_HOST` is defined in `solr.in.sh` or `solr.in.cmd`.
...
 Base Solr URL, which can be used in SolrCloud mode to determine the ZooKeeper connection string if that's not known.
 
+`-u <user:pass>` or `--credentials <user:pass>`::
++
+[%autowidth,frame=none]
+|===
+|Optional | Default: none
+|===
++
+Basic authentication username and password seperated by a `:` character.
```
stillalex, line 755: `separated`

solr/solr-ref-guide/modules/deployment-guide/pages/solr-control-script-reference.adoc
epugh: thanks!!!

stillalex:
are the 2 test failures actual problems?
```
  - org.apache.solr.cli.PostToolTest.testBasicRun (:solr:core)
    Test output: /tmp/src/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.cli.PostToolTest.txt
    Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.cli.PostToolTest.testBasicRun" -Ptests.jvms=96 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=28AE95C7AC0DDA9 -Ptests.file.encoding=ISO-8859-1

  - org.apache.solr.cli.DeleteToolTest.testFailsToDeleteProtectedCollection (:solr:core)
    Test output: /tmp/src/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.cli.DeleteToolTest.txt
    Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.cli.DeleteToolTest.testFailsToDeleteProtectedCollection" -Ptests.jvms=96 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=28AE95C7AC0DDA9 -Ptests.file.encoding=ISO-8859-1
```

epugh:
> are the 2 test failures actual problems?
> 
> ```
>   - org.apache.solr.cli.PostToolTest.testBasicRun (:solr:core)
>     Test output: /tmp/src/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.cli.PostToolTest.txt
>     Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.cli.PostToolTest.testBasicRun" -Ptests.jvms=96 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=28AE95C7AC0DDA9 -Ptests.file.encoding=ISO-8859-1
> 
>   - org.apache.solr.cli.DeleteToolTest.testFailsToDeleteProtectedCollection (:solr:core)
>     Test output: /tmp/src/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.cli.DeleteToolTest.txt
>     Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.cli.DeleteToolTest.testFailsToDeleteProtectedCollection" -Ptests.jvms=96 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=28AE95C7AC0DDA9 -Ptests.file.encoding=ISO-8859-1
> ```

I am afraid they may be..  I *think* it may be due to SSL???   However, the reproduction seed doesn't actually reproduce for me.

epugh:
```
2> SimplePostTool: FATAL: Connection error (is Solr running at https://127.0.0.1:54967/solr/aliasedCollection ?): javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
  2> 
  2> ERROR: System/Runtime.exit(2) or halt(2) calls are not allowed because they terminate the test runner's JVM.
  2> 
   >     java.lang.AssertionError: expected:<0> but was:<1>
   >         at __randomizedtesting.SeedInfo.seed([28AE95C7AC0DDA9:C0FC1E58719B75EA]:0)
   >         at org.junit.Assert.fail(Assert.java:89)
   >         at org.junit.Assert.failNotEquals(Assert.java:835)
   >         at org.junit.Assert.assertEquals(Assert.java:647)
   >         at org.junit.Assert.assertEquals(Assert.java:633)
   >         at org.apache.solr.cli.PostToolTest.testBasicRun(PostToolTest.java:99)
```

epugh:
I'd like to merge this next week, can I get another LGTM on this code?  Any other concerns?

stillalex (approved):

PR #1957, opened 2023-09-25
SOLR-16980 Connect to SOLR standalone with basic authentication

stillalex:
https://issues.apache.org/jira/browse/SOLR-16980

# Description

Adding support for passing basic auth details to cli command

# Solution

# Tests

# Checklist


epugh:
Awesome work @stillalex ....   Have you looked at https://github.com/apache/solr/pull/1954 (SOLR-14496) where I took a stab at adding credentials to the SolrCLI?   It would be great if we used the same patterns in both there AND here...   For example, I went with `-u eric:password` instead of seperate username and password params, mostly because curl does it that way!   

Sigh, in looking, looks like we use net.sourceforge.argparse4j instead of commons-cli...  Wish we picked one CLI arg parser for them all!

Lastly, are there any improvements here that could be reused in better handling of the CLI stuff?  Would that change any of this code?

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrClientFactory.java
epugh, line 1: Do we need this class?  Probably, but I wonder if it actually is required with all the other builders we have for Solr Clients?   Seems a shame to have "YASCB".   _yet another solr client builder_.....  ;-)

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrExporter.java
```diff
@@ -96,6 +96,20 @@ public class SolrExporter {
           + ARG_NUM_THREADS_DEFAULT
           + ".";
 
+  private static final String[] ARG_USER_FLAGS = {"-u", "--user"};
```
epugh, line 99: at least we have `--user` instead of `-user`, over on SolrCLI, we want to get to the `--` pattern!

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrExporter.java
```diff
@@ -242,6 +256,22 @@ public static void main(String[] args) {
         .setDefault(ARG_CLUSTER_ID_DEFAULT)
         .help(ARG_CLUSTER_ID_HELP);
 
+    parser
```
epugh, line 259: Curious if we love this parser or not for CLI args???   THoughts on this versus the commons-cli ?

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrExporter.java
```diff
@@ -266,6 +296,12 @@ public static void main(String[] args) {
         clusterId = defaultClusterId;
       }
 
+      if (!res.getString(ARG_USER_DEST).isEmpty() && !res.getString(ARG_PASSWORD_DEST).isEmpty()) {
+        String user = res.getString(ARG_USER_DEST);
+        String password = res.getString(ARG_PASSWORD_DEST);
+        scrapeConfiguration.withBasicAuthCredentials(user, password);
```
epugh, line 302: i ended up littering the https://github.com/apache/solr/pull/1954 code iwth lots of things that look like this too ;-).

solr/prometheus-exporter/src/test-files/solr/security.json
epugh, line 1: have you seen the new "sane defaults" security.json?  It declares four roles, index, search, admin, and superadmin.  I wonder if it would make sense to add a "monitor" role to that as part of this?  So you do:

```
bin/solr auth...
```
and now you are ready for Prometheus?

solr/prometheus-exporter/src/test/org/apache/solr/prometheus/scraper/SolrStandaloneScraperBasicAuthTest.java
```diff
@@ -0,0 +1,112 @@
...
+  private static ExecutorService executor;
+
+  private static String user = "solr";
+  private static String pass = "SolrRocks";
+
+  @BeforeClass
+  public static void setupSolrHome() throws Exception {
+    Path solrHome = LuceneTestCase.createTempDir();
+    Files.copy(
+        SolrTestCaseJ4.TEST_PATH().resolve("security.json"), solrHome.resolve("security.json"));
```
epugh, line 58: a thought...  that if security.json isn't special for your test, just use the one that now ships with Solr?

solr/prometheus-exporter/src/test/org/apache/solr/prometheus/scraper/SolrStandaloneScraperBasicAuthTest.java
```diff
@@ -0,0 +1,112 @@
...
+            .withBasicAuthCredentials(user, pass);
+    solrClient =
+        new SolrClientFactory(settings, scrapeConfiguration)
+            .createStandaloneSolrClient(solrRule.getBaseUrl());
+    executor =
+        ExecutorUtil.newMDCAwareFixedThreadPool(
+            25, new SolrNamedThreadFactory("solr-cloud-scraper-tests"));
+    solrScraper = new SolrStandaloneScraper(solrClient, executor, "test");
+
+    Helpers.indexAllDocs(solrClient);
```
epugh, line 84: don't love this class name?  Helpers feels very generic...   Also, we don't have the equivaletn of indexAllDocs somewhere already?

solr/prometheus-exporter/src/test/org/apache/solr/prometheus/scraper/SolrStandaloneScraperTest.java
```diff
@@ -47,42 +54,45 @@ public class SolrStandaloneScraperTest extends RestTestBase {
...
 
+  public static void createConf(Path configSet) throws IOException {
+    Path subHome = configSet.resolve("conf");
+    Files.createDirectories(subHome);
+
+    Path top = SolrTestCaseJ4.TEST_PATH().resolve("collection1").resolve("conf");
+    Files.copy(top.resolve("managed-schema.xml"), subHome.resolve("schema.xml"));
+    Files.copy(top.resolve("solrconfig.xml"), subHome.resolve("solrconfig.xml"));
+
+    Files.copy(top.resolve("stopwords.txt"), subHome.resolve("stopwords.txt"));
```
epugh, line 87: wish we had a different more minimal config that dind't drag along stopwords and synnymos....

stillalex: trouble with using the minimal config is that the data indexed here uses multi value strings heavily which is not allowed in the existing min config. it will fail indexing, so this is the best I could come up with

solr/prometheus-exporter/src/test/org/apache/solr/prometheus/scraper/SolrStandaloneScraperBasicAuthTest.java
stillalex: it only has a user+pass. if one already exists and is available to this module I can try to use it, sure

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrExporter.java
stillalex: I used what was there already :( I can take a stab at rewriting this cli code on top of commons-cli

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrClientFactory.java
stillalex: it wires the existing builders with params passed via cli, it didn't fee like a big problem to me.

epugh:
> thank you for the review @epugh!
> 
> > Awesome work @stillalex .... Have you looked at #1954 ([SOLR-14496](https://issues.apache.org/jira/browse/SOLR-14496)) where I took a stab at adding credentials to the SolrCLI? It would be great if we used the same patterns in both there AND here... For example, I went with -u eric:password instead of seperate username and password params, mostly because curl does it that way!
> 
> I have not, no. will take a look and try to make it consistent with #1954. `-u` style works too I don't have a strong preference here.
> 
> > Sigh, in looking, looks like we use net.sourceforge.argparse4j instead of commons-cli... Wish we picked one CLI arg parser for them all!
> 
> strange. did not even notice this until you said it. will try to see how complicated the refactor to commons-cli would be

Also, not wedded to commons-cli...  I think if the community likes argparse4j, then happy to go that way as well ;-)

solr/prometheus-exporter/src/test/org/apache/solr/prometheus/scraper/SolrStandaloneScraperTest.java
epugh: makes sense.  Maybe someday we figure out how many tests actually want stopwords, or which ones want stopwords, and only have those tests depend on it!

solr/prometheus-exporter/src/test/org/apache/solr/prometheus/scraper/SolrStandaloneScraperBasicAuthTest.java
epugh: Actually, the class name I didn't love was `Helpers`...  ;-)   I feel like that method `indexAllDocs` must exist already!

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrClientFactory.java
epugh: Maybe I just wished we could use one of the other factory/builders we have..  You are right that this probably isn't a big deal ;-)...

solr/prometheus-exporter/src/test/org/apache/solr/prometheus/scraper/SolrStandaloneScraperBasicAuthTest.java
stillalex: ah right. this was already there. I picked up the SolrStandaloneScraperTest test (cleaned it up a bit) and added basic auth to it. `indexAllDocs` will actually index a local copy of the solr examples files. I'm pretty sure this was not strictly needed for this, but yeah I didn't add any of it.

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrExporter.java
stillalex: I moved it all to 'credentials' similar to the example PR.

solr/prometheus-exporter/src/test/org/apache/solr/prometheus/scraper/SolrStandaloneScraperBasicAuthTest.java
stillalex: trying to find an example for this. would you happen to have one that I could use as a reference?

epugh: https://github.com/apache/solr/blob/main/solr/core/src/resources/security.json is the one that was introduced...

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrExporter.java
```diff
@@ -96,6 +96,13 @@ public class SolrExporter {
           + ARG_NUM_THREADS_DEFAULT
           + ".";
 
+  private static final String[] ARG_CREDENTIALS_FLAGS = {"-u", "--credentials"};
```
epugh, line 99: Loving it!

stillalex:
@epugh re. `commons-cli` vs `argparse4j` I want to take a stab at rewriting, but I would like to do it as a part of a different PR as an improvement. for this PR i am focusing on resolving a regression: not being able to use prometheus with basic auth. if that works for you I will create another Jira to track it so it doesn't get lost.

solr/prometheus-exporter/src/test/org/apache/solr/prometheus/scraper/SolrStandaloneScraperBasicAuthTest.java
stillalex: I thin the trouble is prometheus module does not depend on core (not even for testing)? bringing that in just for access to the security file might be overkill.

solr/prometheus-exporter/src/test-files/solr/security.json
epugh: yeah, a "metrics" role or a "monitor" role that has what you would need, that way you don't have to figure that stuff out just to use Prometheus...  (assuming there are some logical permissions you need)

solr/prometheus-exporter/src/test/org/apache/solr/prometheus/scraper/SolrStandaloneScraperBasicAuthTest.java
epugh: ah.. yeah.

stillalex: found a cleaner way. removed the file and I am recreating it as needed.

stillalex:
sure, created SOLR-16996 and linked to umbrella ticket.

PR #1963, opened 2023-09-28
Fix the admin UI green core-size graph on nodes screen

janhoy:
The green bar should be 100% wide for the largest core, and the other should be as wide as their relative size compared to the largest core.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -219,6 +219,8 @@ Bug Fixes
 
 * SOLR-16997: OTEL configurator NPE when SOLR_HOST not set (janhoy)
 
+* PR#1963: Fix the admin UI green core-size graph on nodes screen (janhoy)
+
 * SOLR-16980: Connect to SOLR standalone with basic authentication (Alex Deparvu)
 
 Dependency Upgrades
--- a/solr/webapp/web/js/angular/controllers/cloud.js
+++ b/solr/webapp/web/js/angular/controllers/cloud.js
@@ -422,10 +422,10 @@ var nodesSubController = function($scope, Collections, System, Metrics) {
 
               // These are the cores we _expect_ to find on this node according to the CLUSTERSTATUS
               var cores = nodes[node]['cores'];
-              var indexSizeTotal = 0;
+              var indexSizeMax = 0;
               var docsTotal = 0;
               var graphData = [];
-              for (coreId in cores) {
+              for (let coreId in cores) {
                 var core = cores[coreId];
                 if (core['shard_state'] !== 'active' || core['state'] !== 'active') {
                   // If core state is not active, display the real state, or if shard is inactive, display that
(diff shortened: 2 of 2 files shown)


epugh (approved):
LGTM.   The logic is actually more complex than you would have thought!  So I can see how this bug snuck in...

beegotsy:
Hi, I just ran into this same issue today (I’m still on Solr<9.4.0).

While preparing a patch I [searched Jira for related issues](https://issues.apache.org/jira/browse/SOLR-8207?jql=(text%20~%20%22indexSizeTotal%22%20OR%20text%20~%20%22sizeInBytes%22%20OR%20text%20~%20%22Metrics%22)%20AND%20component%20IN%20(AdminGUI%2C%20%22Admin%20Interface%22%2C%20AdminUI%2C%20%22Admin%20UI%22)) and found the original ticket: [SOLR‑8207](https://issues.apache.org/jira/browse/SOLR-8207).
In [the comment where the green bar was first implemented](https://issues.apache.org/jira/browse/SOLR-8207?focusedCommentId=16448243), the intention was for the green bar to show how much space a core uses **relative to the total index size** on the node (with a still‑present flaw that the total is computed only from live cores, so disk used by dead cores is not included).

From the original patch:
https://patch-diff.githubusercontent.com/raw/apache/lucene-solr/pull/360.patch
```
+              var cores = nodes[node]['cores'];
+              var indexSizeTotal = 0;
+              var graphData = [];
+              if (cores) {
+                for (coreId in cores) {
+                  var core = cores[coreId];
+                  var keyName = "solr.core." + core['core'].replace('_', '.').replace('_', '.');
+                  var nodeMetric = m.metrics[keyName];
+                  var size = nodeMetric['INDEX.sizeInBytes'];
+                  core['sizeInBytes'] = size;
+                  core['size'] = bytesToSize(size);
+                  core['sizeLabel'] = core['core'].replace('_shard', '_s').replace(/_replica_./, 'r');
+                  indexSizeTotal += size;
+                }
+                for (coreId in cores) { // here the cicle was correctly separated
+                  var core = cores[coreId];
+                  var graphObj = {};
+                  graphObj['label'] = core['sizeLabel'];
+                  graphObj['size'] = core['sizeInBytes'];
+                  graphObj['sizeHuman'] = core['size'];
+                  graphObj['pct'] = (core['sizeInBytes'] / indexSizeTotal) * 100; // here the bar was correlated to the total index size
+                  graphData.push(graphObj);
+                }
+              }
+              cores.sort(function (a, b) {
+                return b.sizeInBytes - a.sizeInBytes
+              });
+              graphData.sort(function (a, b) {
+                return b.size - a.size
+              });
+              nodes[node]['graphData'] = graphData;
+              nodes[node]['sizeInBytes'] = indexSizeTotal;
+              nodes[node]['size'] = bytesToSize(indexSizeTotal);
```

At some later point (I haven’t checked the exact file history), these two `for` loops were merged, and that’s when the bug appeared: `indexSizeTotal` contained the size of the first core, so that first core was always shown as 100%; and the later cores were no longer showing the correct percentage.

This PR changes the semantics slightly so that the bar is now scaled relative to the **largest** core on the node instead of the total size of all cores.

For our use case (we’re running over 3000 collections), this new behavior actually makes the visualization more useful and readable. Thanks for fixing this.

PR #1964, opened 2023-09-29
SOLR-11535: Fix race condition in singleton-per-collection StateWatcher creation

magibney:
ZkStateReader is supposed to register a singleton StateWatcher per collection name, but there's a race condition that can lead to multiple statewatchers being registered, recreating themselves in perpetuity, and accumulating indefinitely. We've seen up to ~16x accumulation in real-world deployments, leading to state update floods upon large-scale cluster operations (e.g. cluster restart). It's difficult to be 100% certain, but we strongly suspect that in the most pathological cases this can actually lead to nodes seizing up on shutdown as they're forced to handle a flood of incoming messages from other nodes (up/down/etc.), with many thousands of redundant zk callback threads.

It is again difficult to say with 100% certainty, but several months' worth of circumstantial evidence around cluster restarts suggests that this change has drastically reduced overall latency (up to ~2x), and increased consistency/reliability of cluster restarts. We have seen no problems introduced with clusters running this patch.


solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -1329,8 +1336,9 @@ public void process(WatchedEvent event) {
         return;
       }
 
-      if (!collectionWatches.watchedCollections().contains(coll)) {
-        // This collection is no longer interesting, stop watching.
+      StatefulCollectionWatch scw = collectionWatches.statefulWatchesByCollectionName.get(coll);
+      if (scw == null || scw.associatedWatcher != this) {
+        // Collection no longer interesting, or we have been replaced by a different watcher.
```
magibney, line 1341: The substance of the fix is in these three lines. Since StateWatcher instance do a deferred check against collectionWatches in order to determine whether to exit, it's possible that by the time a StateWatcher checks here, the watch it was created to serve has been removed, and a new watch inserted (and new StateWatcher created) before this check. In that case it's possible to accumulate persistent StateWatcher instances that all notify callbacks on the same event.

The fix is for the StateWatcher to check whether it is the instance responsible for the current entry in collectionWatches, and exit if not. The rest of the changes are basically just to ensure that The new StatefulCollectionWatch.associatedWatcher field is never null once it's been inserted in the map.

magibney:
Having lived with these changes for a bit, I plan to commit this within the next day or two, pending feedback. Proposed CHANGES.txt entry:
```
* SOLR-11535: Fix race condition in singleton-per-collection StateWatcher creation (Michael Gibney)
```
I'm inclined to put this under "Bugs". Even though it's a bug that often doesn't often manifest in a visible way, the essence of this issue is that ZkStateReader has been leaking redundant `StateWatcher`s that, once present, re-register themselves in perpetuity as long as the collection is still watched ... which definitely sounds to me like a bug.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -1679,19 +1698,21 @@ public static String getCollectionPath(String coll) {
...
   public void registerCore(String collection) {
-    AtomicBoolean reconstructState = new AtomicBoolean(false);
+    AtomicReference<StateWatcher> newWatcherRef = new AtomicReference<>();
     collectionWatches.compute(
         collection,
         (k, v) -> {
           if (v == null) {
-            reconstructState.set(true);
-            v = new StatefulCollectionWatch();
+            StateWatcher stateWatcher = new StateWatcher(collection);
```
noblepaul, line 1706: Can we replace this with
can we replace this with
```
 v = new StatefulCollectionWatch(new StateWatcher(collection));
 newWatcherRef.set(v.associatedWatcher);
```

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -1763,26 +1784,29 @@ public void registerCollectionStateWatcher(
...
+  public void registerDocCollectionWatcher(
+      String collection, DocCollectionWatcher docCollectionWatcher) {
+    AtomicReference<StateWatcher> newWatcherRef = new AtomicReference<>();
     collectionWatches.compute(
         collection,
         (k, v) -> {
           if (v == null) {
-            v = new StatefulCollectionWatch();
-            watchSet.set(true);
+            StateWatcher stateWatcher = new StateWatcher(collection);
```
noblepaul, line 1794: same here too
```
            v = new StatefulCollectionWatch(new StateWatcher(collection));
            newWatcherRef.set(v.associatedWatcher);
```

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
magibney: Curious, is there a reason to prefer this beyond style? Style-wise I somewhat prefer _not_ writing to a volatile variable and then reading back from it ... but because the `StatefulCollectionWatch` instance is not accessible from other threads at this point there's no functional issue with it.

PR #1965, opened 2023-09-29
SOLR-16924: Refactor requestApplyUpdate step in action=RESTORE

julia-maimone:
https://issues.apache.org/jira/browse/SOLR-16924

# Description

Refactor action=RESTORE to transition UpdateLog to ACTIVE after each replica is restored in RestoreCore call, instead of making calls to RequestApplyUpdate (which effectively transitions UpdateLog to ACTIVE state) on each newly restored replica.
The purpose of this is to simplify and clarify steps of restore. The RequestApplyUpdate op not only buffers updates, it also transitions state (the only part necessary post restoring replica). 

# Solution

Remove calls to RequestApplyUpdate operation on each newly restored replica, and instead complete transition of UpdateLog to ACTIVE state in each RestoreCore op after data is restored to replica.

# Tests

Have run all existing unit tests in solr core module. 
This is a refactor, existing tests for Backup/Restore are sufficient.

# Checklist

- [ X] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.


solr/core/src/java/org/apache/solr/handler/admin/api/RestoreCoreAPI.java
```diff
@@ -146,6 +147,10 @@ private void doRestore(String coreName, RestoreCoreRequestBody requestBody) thro
           .getZkController()
           .getShardTerms(cd.getCollectionName(), cd.getShardId())
           .ensureHighestTermsAreNotZero();
+
+      // transitions state of update log to ACTIVE
+      UpdateLog updateLog = core.getUpdateHandler().getUpdateLog();
+      updateLog.applyBufferedUpdates();
```
dsmiley, line 153: I'd add a null check on this updateLog just in case there is none.  In practice it's always there for SolrCloud but there are lots of null checks elsewhere for it and I like to think we could run SolrCloud without insisting on it (albeit expert).

dsmiley (changes requested):

dsmiley (approved):
Proposed CHANGES.txt:
> SOLR-16924: RESTORECORE now sets the UpdateLog to ACTIVE state instead of requiring a separate REQUESTAPPLYUPDATES call in Collection restore.  (Julia Lamoine, David Smiley)

A backwards-compatibility issue occurred to me.  If during a Solr upgrade, the node issuing this command (the Overseer) is upgraded before the nodes it talks to, then it won't issue REQUESTAPPLYUPDATES.  I'll bring this up on the dev list.  I think this is fine in a major version upgrade but unsure if it's significant enough in a minor one.  It would probably be a transient issue; the restore would fail and ultimately would need to be retried.

dsmiley:
I'll be pushing this to main with a CHANGES.txt and upgrade note for getting to 10.  In 9x I will omit the RestoreCmd part but will add a comment about it disappearing in 10.

dsmiley:
Going to merge soon.  9x port will exclude the RestoreCmd part so we retain backwards-compatibility for in-place upgrades within the 9x line.

BTW I'm so glad to see Crave working so well -- the GitHub Action in entirety took 8m which is amazing.  There is still room for improvement.

PR #1986, opened 2023-10-06
SOLR-17015: Test and fix bug in MoveReplicaCmd for PRS collections

magibney:
https://issues.apache.org/jira/browse/SOLR-17015

The extant test for AbstractMoveReplicaTestBase didn't cover PRS at all, and was also a bit naive in the sense that it tested replica moves on a static system, with no concurrent updates coming in.

This PR enhances the main test to concurrently index and commit to the collection as the MoveReplicas are happening, pausing enough to verify that no docs have been dropped in the process. It also randomly enables/disabled PRS for the test run.

The more robust testing uncovered an optimization in ActiveReplicaWatcher (to work around behavior that's fixed by #1964) that assumes non-PRS-enabled collections. This optimization ended up causing PRS moves to fail in some circumstances (reliably triggered by the enhanced test). This PR removes that optimization, which after merging #1964 should be irrelevant anyway.

Unsurprisingly, the more intensive concurrent indexing and query load increases the latency of the test, in the worst case ~2x, but often actually slightly _faster_ for PRS, and on average ~20% slower (based on beasting the test).

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/ActiveReplicaWatcher.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ActiveReplicaWatcher.java
@@ -44,8 +44,6 @@ public class ActiveReplicaWatcher implements CollectionStateWatcher {
   private final List<String> solrCoreNames = new ArrayList<>();
   private final List<Replica> activeReplicas = new ArrayList<>();
 
-  private int lastZkVersion = -1;
-
   private SolrCloseableLatch latch;
 
   /**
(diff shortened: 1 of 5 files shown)


magibney:
I'm planning to merge this soon (after #1964). The main code fixes a clear bug that only affects PRS. The test code is considerably more robust in stressing conditions that could cause moves to fail. There's little point to letting this bake on main, since the (minimal) core code changes are definitely solid. The only risk from merging this is spurious test failures. I'll keep an eye on builds after committing this and figure out how to proceed if we indeed hit any new test failures.

Proposed CHANGES.txt entry:
```
* SOLR-17015: Test and fix bug in MoveReplicaCmd for PRS collections (Michael Gibney)
```

dsmiley:
@magibney if this is ready to go; why not merge?

PR #1938, opened 2023-09-19
SOLR-16989: Optimize and consolidate reuse of DocValues iterators for value retrieval

magibney:
See: SOLR-16989

PR #1996, opened 2023-10-10
SOLR-17022: Support for glob patterns for fields in Export handler, Stream handler and with SelectStream streaming expression

justinrsweeney:
https://issues.apache.org/jira/browse/SOLR-17022

# Description

This PR provides support for using glob patterns for fields in Export handler, Stream handler and with the SelectStream streaming expression. This creates similar ease of use for these handlers as we already have for the Select handler. Given how Solr export handler works, the glob pattern will be applied to only match fields that support stream, which are fields with docvalues enabled.

# Solution

This solution uses similar methods to the Select handler to part glob patterns into matching fields. We then create a FieldWriter for each matching field to return the necessary data.

# Tests

Implemented tests to ensure that we create the correct set of fields as a results of glob pattern matching.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
+++ b/solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
@@ -894,8 +894,7 @@ public class ExportWriter implements SolrCore.RawWriter, Closeable {
         SchemaField schemaField = searcher.getSchema().getField(fi.getName());
         if (fieldsProcessed.add(fi.getName())
             && schemaField.hasDocValues()
-            && (!(schemaField.getType() instanceof SortableTextField)
-                || schemaField.useDocValuesAsStored())) {
+            && schemaField.useDocValuesAsStored()) {
           expandedFields.add(schemaField);
         }
       }
--- a/solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
+++ b/solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
@@ -33,7 +33,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
-import org.apache.commons.io.FilenameUtils;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.LeafReader;
 import org.apache.lucene.index.LeafReaderContext;
@@ -58,6 +57,7 @@ import org.apache.solr.common.PushWriter;
 import org.apache.solr.common.params.CommonParams;
 import org.apache.solr.common.params.SolrParams;
 import org.apache.solr.common.params.StreamParams;
+import org.apache.solr.common.util.GlobPatternUtil;
 import org.apache.solr.common.util.JavaBinCodec;
 import org.apache.solr.core.SolrCore;
 import org.apache.solr.metrics.SolrMetricsContext;
@@ -890,7 +890,7 @@ public class ExportWriter implements SolrCore.RawWriter, Closeable {
       Set<String> fieldsProcessed,
       List<SchemaField> expandedFields) {
     for (FieldInfo fi : searcher.getFieldInfos()) {
-      if (FilenameUtils.wildcardMatch(fi.getName(), fieldPattern)) {
+      if (GlobPatternUtil.matches(fieldPattern, fi.getName())) {
         SchemaField schemaField = searcher.getSchema().getField(fi.getName());
         if (fieldsProcessed.add(fi.getName())
             && schemaField.hasDocValues()
--- a/solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
+++ b/solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
@@ -28,12 +28,9 @@ import java.io.PrintWriter;
 import java.lang.invoke.MethodHandles;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.TreeSet;
-import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.LeafReader;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.index.SortedDocValues;
@@ -57,7 +54,6 @@ import org.apache.solr.common.PushWriter;
 import org.apache.solr.common.params.CommonParams;
 import org.apache.solr.common.params.SolrParams;
 import org.apache.solr.common.params.StreamParams;
-import org.apache.solr.common.util.GlobPatternUtil;
 import org.apache.solr.common.util.JavaBinCodec;
 import org.apache.solr.core.SolrCore;
 import org.apache.solr.metrics.SolrMetricsContext;
@@ -81,6 +77,7 @@ import org.apache.solr.schema.SortableTextField;
 import org.apache.solr.schema.StrField;
 import org.apache.solr.search.DocValuesIteratorCache;
 import org.apache.solr.search.SolrIndexSearcher;
+import org.apache.solr.search.SolrReturnFields;
 import org.apache.solr.search.SortSpec;
 import org.apache.solr.search.SyntaxError;
 import org.slf4j.Logger;
@@ -126,7 +123,7 @@ public class ExportWriter implements SolrCore.RawWriter, Closeable {
   private int priorityQueueSize;
   StreamExpression streamExpression;
   StreamContext streamContext;
-  FieldWriter[] fieldWriters;
+  List<FieldWriter> fieldWriters;
   int totalHits = 0;
   FixedBitSet[] sets = null;
   PushWriter writer;
@@ -298,7 +295,7 @@ public class ExportWriter implements SolrCore.RawWriter, Closeable {
     }
 
     try {
-      fieldWriters = getFieldWriters(fields, req.getSearcher());
+      fieldWriters = getFieldWriters(fields, req);
     } catch (Exception e) {
       writeException(e, writer, true);
       return;
@@ -478,7 +475,7 @@ public class ExportWriter implements SolrCore.RawWriter, Closeable {
   }
 
   void writeDoc(
-      SortDoc sortDoc, List<LeafReaderContext> leaves, EntryWriter ew, FieldWriter[] writers)
+      SortDoc sortDoc, List<LeafReaderContext> leaves, EntryWriter ew, List<FieldWriter> writers)
       throws IOException {
     int ord = sortDoc.ord;
     LeafReaderContext context = leaves.get(ord);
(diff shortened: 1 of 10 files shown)


risdenk (changes requested):

solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
```diff
@@ -27,9 +27,14 @@
 import java.io.PrintWriter;
 import java.lang.invoke.MethodHandles;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.TreeSet;
+import org.apache.commons.io.FilenameUtils;
```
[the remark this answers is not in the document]
justinrsweeney: This isn't a new dependency in core, this is already used elsewhere today: https://github.com/apache/solr/blob/e45be355e2b0fde5f5c80265f1baade024e25e48/solr/core/src/java/org/apache/solr/search/SolrReturnFields.java#L581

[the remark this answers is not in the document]
risdenk: Ideally we don't pull in more dependencies than needed. I think its weird to use `FilenameUtils` when we are basically just looking at simple globs. 

There was a push to limit dependencies where possible and commons-io was one that just kept getting used all over the place.

epugh:
This seems like a very useful feature!   I also like the establishing parity in how things work.  Will you make sure to update ref guide?

solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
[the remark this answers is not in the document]
justinrsweeney: I updated to create a GlobPatternUtil class that uses Java NIO file pattern matching to avoid using an additional library or having to implement our own glob pattern matching.

solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
```diff
@@ -844,4 +844,61 @@ public String getMessage() {
...
+      String fieldPattern,
+      SolrIndexSearcher searcher,
+      Set<String> fieldsProcessed,
+      List<SchemaField> expandedFields) {
+    for (FieldInfo fi : searcher.getFieldInfos()) {
+      if (GlobPatternUtil.matches(fieldPattern, fi.getName())) {
+        SchemaField schemaField = searcher.getSchema().getField(fi.getName());
+        if (fieldsProcessed.add(fi.getName())
+            && schemaField.hasDocValues()
+            && (!(schemaField.getType() instanceof SortableTextField)
```
justinrsweeney, line 897: Wanted to highlight this because @magibney and I were discussing this previously. The way this is implemented for the Export handler here actually deviates from how glob patterns work for the Select handler.

The Select handler will not match any fields where `useDocValuesAsStored=false` for glob patterns, those fields must be explicitly provided in the field list.

For the purposes of Export I did not follow that pattern and instead will match any fields that have docvalues enabled with the exception of SortableTextField which require `useDocValuesAsStored=true` to be used in Export in other places in the code.

My reasoning here is that there is a performance hit for getting non-stored, docvalues enabled fields in the Select handler that doesn't seem to be to be as impactful in the Export handler.

Open to other opinions on this topic, should we:
1. Have glob patterns for fields in Export handler return any fields that are able to be exported
2. Match the behavior in the Select handler and only return fields that match the pattern, have docvalues enabled AND have `useDocValuesAsStored=true`

solr/solrj/src/java/org/apache/solr/common/util/GlobPatternUtil.java
```diff
@@ -0,0 +1,11 @@
+package org.apache.solr.common.util;
```
risdenk, line 1: Looks like missing header

solr/solrj/src/test/org/apache/solr/common/util/TestGlobPatternUtil.java
```diff
@@ -0,0 +1,17 @@
+package org.apache.solr.common.util;
```
risdenk, line 1: Looks like missing header

risdenk (changes requested):
Thanks @justinrsweeney looks much cleaner

solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
magibney: As @justinrsweeney foreshadowed, I am in favor of parity in how this is handled between select and export.

The purpose of udvas as I understand it is not simply to mitigate performance issues in select (and in fact, performance should be comparable between select and export as of #1938), but also as a config option to allow the configuration of docValues for a field without requiring that those values be returned when a field matches a glob pattern.

There are plenty of use cases where, e.g., a derivative string field is configured solely for the purpose of sort, faceting, [or other purpose that would make even less sense to return on export]. It should be possible to configure a field that's intended for one of the many other purposes docValues serve, without forcing it to be returned when the field name matches a glob pattern. Note also, if both `stored=true` and `useDocValuesAsStored=true`, stored fields will be used for select and docValues will be used for export.

I think what I'm missing is: how does this help usability? Are there cases where one would want docValues to be used for field value retrieval, but could not simply set that field to `useDocValuesAsStored=true`? And udvas defaults to true anyway (for all fields except SortableTextField) iirc.

I'm curious to hear others' thoughts on this as well!

solr/solrj/src/java/org/apache/solr/common/util/GlobPatternUtil.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package org.apache.solr.common.util;
+
+import java.nio.file.FileSystems;
+import java.nio.file.Paths;
+
+public class GlobPatternUtil {
+
+  public static boolean matches(String pattern, String input) {
+    return FileSystems.getDefault().getPathMatcher("glob:" + pattern).matches(Paths.get(input));
```
dsmiley, line 25: This is bizarre just for some filename wildcard support.  If there is a reason we use FileSystems then a comment is necessary.

solr/solrj/src/java/org/apache/solr/common/util/GlobPatternUtil.java
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.common.util;
+
+import java.nio.file.FileSystems;
+import java.nio.file.Paths;
+
+public class GlobPatternUtil {
```
dsmiley, line 22: All public classes need a javadoc

dsmiley (commented):
I'm surprised to see that this isn't using SolrReturnFields, which is already doing field glob expansion.  I know it's in solr-core and you seem to need something in solrj-streaming but we could move it.

justinrsweeney:
> I'm surprised to see that this isn't using SolrReturnFields, which is already doing field glob expansion. I know it's in solr-core and you seem to need something in solrj-streaming but we could move it.

SolrReturnFields was using `FilenameUtils.wildcardMatch()` from commons-io and this just uses a very similar method but not requiring the third party library and allows solrj-streaming, SolrReturnFields, and ExportWriter to all use the same mechanism to match glob patterns.

solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
justinrsweeney: I updated this PR to mimic the same behavior as the select handler, where `useDocValuesAsStored=false` fields will not be returned unless specifically requested. I'm in agreement that consistency here is better.

solr/solrj/src/java/org/apache/solr/common/util/GlobPatternUtil.java
dsmiley: I took a look at how it's implemented.  If only we could call ZipUtils.toRegexPattern but the class is package protected.  It's a shame to recompile the glob on each call to matches!

solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
```diff
@@ -844,4 +844,60 @@ public String getMessage() {
...
+   * @param searcher an index search to access schema info
+   * @param fieldsProcessed the set of field names already processed to avoid duplicating
+   * @param expandedFields the list of fields to add expanded field names into
+   */
+  private void getGlobFields(
+      String fieldPattern,
+      SolrIndexSearcher searcher,
+      Set<String> fieldsProcessed,
+      List<SchemaField> expandedFields) {
+    for (FieldInfo fi : searcher.getFieldInfos()) {
```
dsmiley, line 892: There can be many fieldInfo, and you're looping over a "matches" call that is going to internally build a regex each time.  Maybe you should first do the hasDocValues etc. checks so we do this matches check last?

solr/core/src/java/org/apache/solr/handler/export/ExportWriter.java
```diff
@@ -844,4 +844,60 @@ public String getMessage() {
...
+      if (GlobPatternUtil.matches(fieldPattern, fi.getName())) {
+        SchemaField schemaField = searcher.getSchema().getField(fi.getName());
+        if (fieldsProcessed.add(fi.getName())
```
dsmiley, line 895: This line might add to fieldsProcessed, yet exclude the field because doesn't "hasDocValues".  This looks suspicious to me.

dsmiley (commented):
My concern is duplication of logic between SolrReturnFields.  If you think it's not an issue or is impossible to reconcile, then I trust you.

justinrsweeney:
> My concern is duplication of logic between SolrReturnFields. If you think it's not an issue or is impossible to reconcile, then I trust you.

I finally got back around to this. With the latest change, ExportWriter is using SolrReturnFields which in turn is using GlobPatternUtil so there is a consistent approach there.

justinrsweeney:
@risdenk Pinging for re-review when you have a chance

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/stream/SelectStream.java
```diff
@@ -308,6 +327,13 @@ public Tuple read() throws IOException {
       workingForEvaluators.put(fieldName, original.get(fieldName));
       if (selectedFields.containsKey(fieldName)) {
         workingToReturn.put(selectedFields.get(fieldName), original.get(fieldName));
+      } else {
+        for (String globPattern : selectedFieldGlobPatterns) {
+          if (GlobPatternUtil.matches(globPattern, fieldName)) {
```
dsmiley, line 332: Shouldn't SelectStream also use SolrReturnFields and not use lower level GlobPattern stuff (it's something SRF can handle)?

Disclaimer: I haven't looked at this PR in a long time.

justinrsweeney: `solrj-streaming` does not currently have a dependency on `core`, in fact currently I think `core` depends on `solrj-streaming`. I didn't want to refactor SolrReturnFields to live elsewhere given the scope of this PR so not using that.

dsmiley: Oh right; of course.

dsmiley (approved):

LuckyFrost:
@justinrsweeney, hello, I'm not sure, but it seems your change causes a significant performance hit with prefixed wildcard fl

https://lists.apache.org/thread/vbwnjxprl6s1qy0t1jzfcw8hprg1gvzh

magibney:
It's also possible this could be related to [SOLR-16989](https://issues.apache.org/jira/browse/SOLR-16989) -- sorry I missed the mailing list thread, I'll follow up there.

PR #1998, opened 2023-10-10
SOLR-17019: ZkCli should create subpaths when necessary

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17019

### DIFF
--- a/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ClusterProperties.java
+++ b/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ClusterProperties.java
@@ -220,7 +220,7 @@ public class ClusterProperties {
           @SuppressWarnings({"rawtypes"})
           Map properties = new LinkedHashMap();
           properties.put(propertyName, propertyValue);
-          client.create(
+          client.makePath(
               ZkStateReader.CLUSTER_PROPS, Utils.toJSON(properties), CreateMode.PERSISTENT, true);
         }
       } catch (KeeperException.BadVersionException | KeeperException.NodeExistsException e) {
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -260,6 +260,8 @@ Bug Fixes
 
 * SOLR-17009: json.wrf parameter ignored in JacksonJsonWriter (Alex Deparvu)
 
+* SOLR-17019: ZkCli should create subpaths when necessary (Houston Putman)
+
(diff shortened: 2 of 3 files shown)


risdenk (approved):
Do we need to worry about ACLs on these created znodes?

HoustonPutman:
> Do we need to worry about ACLs on these created znodes?

Good question! So zkCli.sh take the `SOLR_ZK_CREDS_AND_ACLS` envVar, just like the `bin/solr` tool commands do. So both should use the same ACLs when starting the Solr ZK Client.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ClusterProperties.java
```diff
@@ -220,7 +220,7 @@ public void setClusterProperty(String propertyName, Object propertyValue) throws
           @SuppressWarnings({"rawtypes"})
           Map properties = new LinkedHashMap();
           properties.put(propertyName, propertyValue);
-          client.create(
+          client.makePath(
```
stillalex, line 223: minor: this doesn't seem to need the change, given that it's a constant.

stillalex (approved):

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ClusterProperties.java
HoustonPutman: This is what necessitated the change. the `chroot` that the user is using might not exist yet.

epugh:
I don't know if this matters at all, but at some point, I'd like to see what `zkCli.sh` does be subsumed into `bin/solr zk` commands.   Obviously with the appropriate deprecation process...

HoustonPutman:
> I don't know if this matters at all, but at some point, I'd like to see what `zkCli.sh` does be subsumed into `bin/solr zk` commands. Obviously with the appropriate deprecation process...

Yeah, I'm not a fan of the separate utilities. One path for all this stuff would be much better. Anyways, this would need to be fixed either way.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ClusterProperties.java
stillalex: yep. wrote the comment and like 10 mins later...oh yeah... thanks for confirming :)

PR #1510, opened 2023-03-30
WIP JDK 17

risdenk:
https://issues.apache.org/jira/browse/SOLR-XXXXX

PR #1999, opened 2023-10-10
SOLR-17023: Use Modern NLP Models via ONNX and Apache OpenNLP with Solr

epugh:
https://issues.apache.org/jira/browse/SOLR-17023

# Description

Code and BATS test that demonstrates downloading a sentiment classification model from Huggingface, converting it to Onnx model, uploading the model to Solr's FileStore, and then configuring a collection and associated pipeline to use the model for enriching content.  

# Solution

New DocumentCategorizerUpdateProcessorFactory that interacts with the model. 

# Tests

BATS test.   Check this PR out and run: ```./gradlew integrationTests --tests test_opennlp.bats```

# Checklist

- [x ] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [x ] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [ x] I have developed this patch against the `main` branch.
- [ ] Look at GPU support
- [ ] Integrate with Films demo.
- [ ] Can we have BATS tests that depend on external resources, or do I check in the ONNX model instead?   (make a stupid one???)
- [ ] Write a unit test?
- [ ] Provide both a label but ALSO a score!


epugh:
Check it out @jzonthemtn

solr/modules/analysis-extras/build.gradle
```diff
@@ -19,6 +19,13 @@ apply plugin: 'java-library'
 
 description = 'Additional analysis components'
 
+configurations.all {
+  resolutionStrategy {
+    force 'org.apache.opennlp:opennlp-tools:2.2.0'
+    force 'org.apache.opennlp:opennlp-dl:2.2.0'
+  }
+}
+
```
risdenk, line 28: Why do you need for force a resolution? This shouldn't be necessary at all...

risdenk (changes requested):

solr/modules/analysis-extras/build.gradle
epugh: We need https://github.com/apache/lucene/pull/448 to be merged to get us to 2.2, or we ahve to work around Lucene...   Is there a better way?   This is what Jeff and I hacked in..

versions.props
```diff
@@ -70,3 +70,4 @@ org.semver4j:semver4j=5.2.1
 org.slf4j:*=2.0.9
 org.xerial.snappy:snappy-java=1.1.10.5
 software.amazon.awssdk:*=2.20.155
+org.apache.opennlp:opennlp-dl=2.2.0
```
risdenk, line 73: try `org.apache.opennlp:opennlp*=2.2.0`

solr/modules/analysis-extras/build.gradle
risdenk: try `org.apache.opennlp:opennlp*=2.2.0` in versions.props and remove this from here. It should force the upgrade to 2.2.0.

epugh: lots of errors when I took out the `resolutionStrategy`:

```
* What went wrong:
Could not determine the dependencies of task ':solr:modules:analysis-extras:compileJava'.
> Could not resolve all task dependencies for configuration ':solr:modules:analysis-extras:compileClasspath'.
   > Could not resolve org.apache.opennlp:opennlp-tools:2.2.0.
     Required by:
         project :solr:modules:analysis-extras
         project :solr:modules:analysis-extras > org.apache.opennlp:opennlp-dl:2.2.0
      > Cannot find a version of 'org.apache.opennlp:opennlp-tools' that satisfies the version constraints:
           Dependency path 'org.apache.solr:analysis-extras:10.0.0-SNAPSHOT' --> 'org.apache.opennlp:opennlp-tools:2.2.0'
           Constraint path 'org.apache.solr:analysis-extras:10.0.0-SNAPSHOT' --> 'org.apache.solr:core:10.0.0-SNAPSHOT' (apiElements) --> 'org.apache.opennlp:opennlp-tools:1.9.4' because of the following reason: Computed from com.palantir.consistent-versions' versions.lock in solr-root
           Constraint path 'org.apache.solr:analysis-extras:10.0.0-SNAPSHOT' --> 'org.apache.solr:solrj:10.0.0-SNAPSHOT' (apiElements) --> 'org.apache.opennlp:opennlp-tools:1.9.4' because of the following reason: Computed from com.palantir.consistent-versions' versions.lock in solr-root
           Dependency path 'org.apache.solr:analysis-extras:10.0.0-SNAPSHOT' --> 'org.apache.lucene:lucene-analysis-opennlp:9.8.0' (compile) --> 'org.apache.opennlp:opennlp-tools:1.9.1'

```

also, feel free to push up changes to this branch ;-)

risdenk: make sure to run `./gradlew --write-locks` to make sure it takes into account `versions.props` changes.

epugh:
Thanks @risdenk , your suggestions worked.   One slight wrinkle that I don't know how to deal with is that the `com.microsoft.onnxruntime:onnxruntime_gpu:1.14.0 ` file doesn't work on OSX, only linux and windows.  So if you look at the BATS test, you'll see some code where I replace it with `com.microsoft.onnxruntime:onnxruntime:1.14.0`.   I don't really know how to handle that going forward..  I mean, swapping a magic jar is probably a bad idea ;-).

epugh:
```
./gradlew integrationTests --tests test_opennlp.bats
```

epugh:
```
> Task :solr:packaging:integrationTests
Running BATS tests with Solr base port 53065
1..1
ok 1 Check lifecycle of sentiment classification in 20000ms
```

This is after uploading the model to the PackageStore and reference it in there!

versions.props
```diff
@@ -70,3 +70,4 @@ org.semver4j:semver4j=5.2.1
 org.slf4j:*=2.0.9
 org.xerial.snappy:snappy-java=1.1.10.5
 software.amazon.awssdk:*=2.20.155
+org.apache.opennlp:opennlp*=2.2.0
```
risdenk, line 73: Should be in order alphabetically...

solr/packaging/test/test_opennlp.bats
```diff
@@ -0,0 +1,110 @@
...
+
+teardown() {
+  # save a snapshot of SOLR_HOME for failed tests
+  save_home_on_failure
+}
+
+@test "Check lifecycle of sentiment classification" {
+  
+  # GPU versions is linux and windows only, not OSX.  So swap jars.
+  rm -f ${SOLR_TIP}/modules/analysis-extras/lib/onnxruntime_gpu-1.14.0.jar
```
risdenk, line 42: Why not just always run on cpu? Why need to test if there is a capable GPU or not? Could be on a VM without a GPU too even w/ linux and windows...

solr/packaging/test/test_opennlp.bats
```diff
@@ -0,0 +1,110 @@
...
+  # restore 
+  #curl --insecure -o ${SOLR_TIP}/modules/analysis-extras/lib/onnxruntime-1.14.0.jar https://repo1.maven.org/maven2/com/microsoft/onnxruntime/onnxruntime/1.14.0/onnxruntime-1.14.0.jar
+  cp /Users/epugh/Documents/projects/solr-epugh/onnxruntime-1.14.0.jar ${SOLR_TIP}/modules/analysis-extras/lib/
```
risdenk, line 45: This is using your path...

risdenk (changes requested):

solr/packaging/test/test_opennlp.bats
epugh: for the purposes of this, I totally agree..   I think I need some Gradle help to figure this out ;-(   I've banged at it with no luck.

solr/packaging/test/test_opennlp.bats
epugh: yeah...    I was in the back of David Smiley's car in the wilds of New Brunswick Canada while I was working on this and didn't want to burn up my data plan ;-).    Likewise I need to figure out a better apporoach with the model since it's 600 mb....   I don't know if I can find a toy one that is super small, or if I add docs to download it from hugging facet etc.    Longer term, I doin't even think this bats test gets committed the way it is because of all the external dependencies..   Though..  if I had a super small model that I could check into Git, maybe that owuld be okay?  Like we have a binary package for some of the package tests..

solr/packaging/test/test_opennlp.bats
jzonthemtn: Just CPU is definitely a better choice here. And just a note, ONNX Runtime does not currently support Mac silicon so this will only work on x64 machines.

epugh:
@jzonthemtn  did open https://issues.apache.org/jira/browse/OPENNLP-1515, but that will require a release of OpenNLP before we can piggy back on it

versions.props
epugh: I think updating the Lucene dependency is a PR on you @jzonthemtn  right?  (I actually have done no commits to Lucene, so wouldn't be comfortable attempting to update opennlp there).   And yeah, then in Lucene 9.9, we could remove this dependency here...

epugh:
I think the first draft is done, and it "works".  There are some caveats and things to be figured out, like the `onnxruntime_gpu` versus `onnxruntime` issue.   Based on my work so far, we definitly need tutorials, there are too many moving pieces.   The PackageStore is great, at least as far as I've worked with it, but it does expect to load the data in memory as it moves it around, so we need to figure that out.   Right now we only get sentiment labels back, and we really need sentiment scores, i.e a number.  I don't know if that can be a 0 to 1, of if it has to be 1,2,3,4,5 labels.   I'd like to experiment with the saem thing as a copyField into a text analyzer, so you copy "review" into "review_sentiment_score" and get a number there.   Or, do we have a streaming expression you run that updates the review_sentiment_score for all docs that don't have it with an atomic update?

solr/packaging/test/test_opennlp.bats
epugh: if there is a workaround, please let me know...!

rzo1: Release is done. New artifacts should be available soon ;-) (not depending on gpu anymore)

epugh: Thank you!!!

solr/modules/analysis-extras/src/java/org/apache/solr/update/processor/DocumentCategorizerUpdateProcessorFactory.java
```diff
@@ -0,0 +1,566 @@
...
+        final double[] result = documentCategorizerDL.categorize(new String[] {fullText});
+
+        // Add the categories to the list and return it.
+        // Just take the top category for now.
+        // TODO: Allow for a threshold value for returning categories.
+
+        List<Pair<String, String>> classifications = new ArrayList<>();
+
+        String bestCategory = documentCategorizerDL.getBestCategory(result);
+        if (log.isInfoEnabled()) {
```
jzonthemtn, line 535: I don't think calling `isInfoEnabled()` here is needed but that's super minor.

jzonthemtn:
Looks like a great first step! Glad that OpenNLP 2.3.1 helped move it along.

epugh:
> Looks like a great first step! Glad that OpenNLP 2.3.1 helped move it along.

I did a community demo yesterday, and it went well.    Having 2.3.1 meant I could remove some ugly moving of Jars!  Which made the demo more compelling.

epugh:
@cpoerschke when I demoed this code at the last community meetup, @gerlowskija asked why not to commit it, and I didn't have a super great reason.   I'd love your thoughts on this PR since you played some with ONNX as well..  Is there anything here you think needs changing before its get merged?  I'd love to get the ONNX stuff in and unblock your work...

solr/licenses/onnxruntime-LICENSE-MIT.txt
```diff
@@ -0,0 +1,21 @@
+MIT License
```
cpoerschke, line 1: Looking up https://github.com/microsoft/onnxruntime/blob/v1.15.0/LICENSE here I noticed there's a 1.15.1 and 1.16.x now too, wondering about 1.15.0 vs. the later ones then.

solr/modules/analysis-extras/src/java/org/apache/solr/update/processor/DocumentCategorizerUpdateProcessorFactory.java
```diff
@@ -0,0 +1,566 @@
...
+
+    Object modelParam = args.remove(MODEL_PARAM);
+    if (null == modelParam) {
+      throw new SolrException(SERVER_ERROR, "Missing required init param '" + MODEL_PARAM + "'");
+    }
+    if (!(modelParam instanceof CharSequence)) {
+      throw new SolrException(SERVER_ERROR, "Init param '" + MODEL_PARAM + "' must be a <str>");
+    }
+    model = modelParam.toString();
+    System.out.println("In OpenNLP doccat - model: " + model);
```
cpoerschke, line 132:
```suggestion
    log.info("In OpenNLP doccat - model: " + model);
```

solr/licenses/onnxruntime-LICENSE-MIT.txt
cpoerschke: part-answering own question: OpenNLP 2.3.1 uses onnxruntime 1.15.0 - https://github.com/apache/opennlp/blame/opennlp-2.3.1/pom.xml#L176

epugh: I *think* we don't specify the version of onnx, so maybe we poke them to update?

solr/modules/analysis-extras/src/java/org/apache/solr/update/processor/DocumentCategorizerUpdateProcessorFactory.java
epugh: took this and did it sliglty different!

cpoerschke:
> Looking at these build failures and the error message generated, it appears that it may be caused by us using Java 11 and OpenNLP being compiled with Java 17?? ... Is this a deal breaker for this PR?

Hmm, interesting. So we have:
* OpenNLP as minimum Java17 as you mention -- https://github.com/apache/opennlp/blob/opennlp-2.3.1/pom.xml#L167
* Lucene as minimum Java11 -- https://github.com/apache/lucene/blob/releases/lucene/9.9.1/build.gradle#L75-L76
* Solr as minimum Java11 -- https://github.com/apache/solr/blob/releases/solr/9.4.1/build.gradle#L88

So if the classes here were built as an independent [plugin](https://solr.apache.org/guide/solr/latest/configuration-guide/solr-plugins.html) (with minimum Java17) and then deployed (with the relevant dependencies) into a Solr setup running Java17 with the original Solr artefacts (built with Java11) -- I wonder if that would work?

Also noting that https://github.com/apache/lucene/pull/579 bumped Lucene to Java17 on `main` branch i.e. presumably then a future Lucene10 will be minimum Java17 version.

rzo1:
> So if the classes here were built as an independent [plugin](https://solr.apache.org/guide/solr/latest/configuration-guide/solr-plugins.html) (with minimum Java17) and then deployed (with the relevant dependencies) into a Solr setup running Java17 with the original Solr artefacts (built with Java11) -- I wonder if that would work?

I guess this should work.

solr/licenses/onnxruntime-LICENSE-MIT.txt
rzo1: Next opennlp will have 1.16.3 (or higher)

epugh:
Feels like what we should be doing is having Solr 10 target Java 17 since Lucene 10 will require it, and then this code goes on Solr 10, but not on Solr 9.  This lets us have some more time to experiment with out dealing with the headaches of supporting an official release in the 9.x line (backcompat and the rest)....??

cpoerschke:
> Feels like what we should be doing is having Solr 10 target Java 17 since Lucene 10 will require it, and then this code goes on Solr 10, but not on Solr 9. This lets us have some more time to experiment with out dealing with the headaches of supporting an official release in the 9.x line (backcompat and the rest)....??

I concur. Also a nice motivation for targeting Java 17 i.e. specific example of functionality that it would unlock. And in the meantime "independent plugin" approaches remain a possibility in the community, perhaps even in the https://github.com/apache/solr-sandbox if someone wanted to pursue that (haven't checked how that is built, just kinda "name dropping" `solr-sandbox` here).

epugh:
Lucene 9 requires older version of Java than the minimum required version that OpenNLP requires. That means that this PR is pending a release of Lucene 10, and the adoption of Lucene 10 by Solr.   It would be interesting to think about if there was a way for Solr `main` branch to somehow depend on the Lucene `main` branch release that jumps the minimum Java versions all around, and would allow this PR to be merged.

versions.props
```diff
@@ -49,6 +49,7 @@ org.apache.httpcomponents:httpmime=4.5.14
 org.apache.kerby:*=1.0.1
 org.apache.logging.log4j:*=2.21.0
 org.apache.lucene:*=9.9.2
```
cpoerschke, line 51: Temporarily within this pull request (pre-merge) we could change this to a Lucene 10 prerelease based on the _"Update Lucene prerelease"_ steps in https://github.com/apache/solr/blob/main/help/dependencies.txt ...

... might be worth waiting though until Lucene 9.10 is out and https://issues.apache.org/jira/browse/SOLR-17157 has upgraded Solr to use it i.e. then `solr/main` will be closer to `lucene/main` than it is right now.

cpoerschke:
> ... It would be interesting to think about if there was a way for Solr `main` branch to somehow depend on the Lucene `main` branch release that jumps the minimum Java versions all around, and would allow this PR to be merged.

Technically I guess Solr `main` could continue to depend on whatever Lucene version and just jumping up the minimum Java version for Solr `main` to 17 would be sufficient? With all the ups-and-downs of `main` and `branch_9x` having different minimums.

risdenk:
@cpoerschke https://github.com/apache/solr/pull/1510 might be helpful here. I have a few wip prs for newer jdks

epugh:
this is still valid ;-)

epugh:
I updated this PR with main, to see what happened, and we're closer.    Things that I think are still holding us back:

1) Solr Main is NOT on Lucene 10, which means we have to override the version of OpenNLP that Lucene uses (maybe somehow?).   However when #3053 gets in that should deal with it.
2) Need to write a JUnit test for `DocumentCategorizerUpdateProcessorFactory` (oops!)

solr/modules/analysis-extras/src/java/org/apache/solr/update/processor/DocumentCategorizerUpdateProcessorFactory.java
```diff
@@ -0,0 +1,569 @@
...
+    final FieldNameSelector srcSelector = getSourceSelector();
+    return new UpdateRequestProcessor(next) {
+
+      DocumentCategorizerDL documentCategorizerDL = null;
+
+      {
+        // Initialize the categorizer.
+        FileStore fs = req.getCoreContainer().getFileStore();
+
+        var path = solrHome.resolve(ClusterFileStore.FILESTORE_DIRECTORY);
```
epugh, line 435: @gerlowskija this is the line I was referenceing, where the below `modelFile` and `vocabFile` feel like they are not using the ClusterFileStore properly....

epugh:
Yep, need to wait for Lucene 10, otherwise we get some unit test failures:

```
gradlew test --tests TestOpenNLPExtractNamedEntitiesUpdateProcessorFactory.testExtractFieldRegexReplaceAll -Dtests.seed=F5DD0B40AC590A66 -Dtests.locale=pt-GW -Dtests.timezone=PRT -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```

```
>     java.lang.NoSuchMethodError: 'opennlp.tools.util.Span[] opennlp.tools.sentdetect.SentenceDetectorME.sentPosDetect(java.lang.String)'
   >         at __randomizedtesting.SeedInfo.seed([F5DD0B40AC590A66:4351D7F53AC9F7A4]:0)
   >         at org.apache.lucene.analysis.opennlp.tools.NLPSentenceDetectorOp.splitSentences(NLPSentenceDetectorOp.java:41)
   >         at org.apache.lucene.analysis.opennlp.OpenNLPSentenceBreakIterator.setText(OpenNLPSentenceBreakIterator.java:199)
   >         at org.apache.lucene.analysis.util.SegmentingTokenizerBase.reset(SegmentingTokenizerBase.java:89)

```

epugh:
> @epugh Just curious on the status of this pull request.

Since we are giving a talk on it at C/C NA in a few months, we need to get it in!   

More seriously, I think what stalled it was the Lucene 10 need... and the `DocumentCategorizerUpdateProcessorFactory` test...   

Are you interested in giving it a bit of a run through with `main` and see if it's mergable, and then I'd be happy to start on the docs side of things?

epugh:
Looks like lucene is on OpenNLP 2.5.4, and we are on a older version...

epugh:
@jzonthemtn I just did some more updating to dependencies.   However, I think we still have some inconsistencies by not being on Lucene 10 that cause some tests to fail.   For example: `./gradlew test --tests TestOpenNLPExtractNamedEntitiesUpdateProcessorFactory.testExtractFieldRegexReplaceAll` produces:

```
 >     java.lang.NoSuchMethodError: 'opennlp.tools.util.Span[] opennlp.tools.sentdetect.SentenceDetectorME.sentPosDetect(java.lang.String)'
   >         at __randomizedtesting.SeedInfo.seed([5E06D84F7820115D:AFA47BD30AF8C6DB]:0)
   >         at org.apache.lucene.analysis.opennlp.tools.NLPSentenceDetectorOp.splitSentences(NLPSentenceDetectorOp.java:41)
   >         at org.apache.lucene.analysis.opennlp.OpenNLPSentenceBreakIterator.setText(OpenNLPSentenceBreakIterator.java:199)
   >         at org.apache.lucene.analysis.util.SegmentingTokenizerBase.reset(SegmentingTokenizerBase.java:89)
   >         at org.apache.lucene.analysis.opennlp.OpenNLPTokenizer.reset(OpenNLPTokenizer.java:94)
   >         at org.apache.solr.update.processor.OpenNLPExtractNamedEntitiesUpdateProcessorFactory$1.extractTypedNamedEntities(OpenNLPExtractNamedEntitiesUpdateProcessorFactory.java:635)
   >         at org.apache.solr.update.processor.OpenNLPExtractNamedEntitiesUpdateProcessorFactory$1.processAdd(OpenNLPExtractNamedEntitiesU
```

Any thoughts on how to fix it?

versions.props
epugh: Bumped to the latest that lucene supports now!

solr/licenses/onnxruntime-LICENSE-MIT.txt
epugh: okay, I think we are NOW on the latest that Lucene 10.2 --> OpenNLP --> Solr supports.

epugh:
> @jzonthemtn I just did some more updating to dependencies. However, I think we still have some inconsistencies by not being on Lucene 10 that cause some tests to fail. For example: `./gradlew test --tests TestOpenNLPExtractNamedEntitiesUpdateProcessorFactory.testExtractFieldRegexReplaceAll` produces:
> 
> ```
>  >     java.lang.NoSuchMethodError: 'opennlp.tools.util.Span[] opennlp.tools.sentdetect.SentenceDetectorME.sentPosDetect(java.lang.String)'
>    >         at __randomizedtesting.SeedInfo.seed([5E06D84F7820115D:AFA47BD30AF8C6DB]:0)
>    >         at org.apache.lucene.analysis.opennlp.tools.NLPSentenceDetectorOp.splitSentences(NLPSentenceDetectorOp.java:41)
>    >         at org.apache.lucene.analysis.opennlp.OpenNLPSentenceBreakIterator.setText(OpenNLPSentenceBreakIterator.java:199)
>    >         at org.apache.lucene.analysis.util.SegmentingTokenizerBase.reset(SegmentingTokenizerBase.java:89)
>    >         at org.apache.lucene.analysis.opennlp.OpenNLPTokenizer.reset(OpenNLPTokenizer.java:94)
>    >         at org.apache.solr.update.processor.OpenNLPExtractNamedEntitiesUpdateProcessorFactory$1.extractTypedNamedEntities(OpenNLPExtractNamedEntitiesUpdateProcessorFactory.java:635)
>    >         at org.apache.solr.update.processor.OpenNLPExtractNamedEntitiesUpdateProcessorFactory$1.processAdd(OpenNLPExtractNamedEntitiesU
> ```
> 
> Any thoughts on how to fix it?

Hooray!  Now that we are on Lucene 10.2 this passed!!!!!!!!!!!!!!!!

jzonthemtn:
Cloning to build locally and check out

epugh:
> Cloning to build locally and check out

We really need a unit test!   Not just a bats test to get this to done....

epugh:
@jzonthemtn I fixed the bats test...   Now seeing this from Solr...  Any ideas?

```
2025-08-16 14:22:34.083 ERROR (qtp89448984-32-localhost-15) [c:COLL_NAME s:shard1 r:core_node2 x:COLL_NAME_shard1_replica_n1 t:localhost-15] o.d.d.DocumentCategorizerDL Unload to perform document classification inference => java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of "java.util.Map.get(Object)" is null
	at opennlp.dl.doccat.DocumentCategorizerDL.tokenize(DocumentCategorizerDL.java:295)
java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of "java.util.Map.get(Object)" is null
	at opennlp.dl.doccat.DocumentCategorizerDL.tokenize(DocumentCategorizerDL.java:295) ~[?:?]
	at opennlp.dl.doccat.DocumentCategorizerDL.categorize(DocumentCategorizerDL.java:140) ~[?:?]
	at org.apache.solr.update.processor.DocumentCategorizerUpdateProcessorFactory$1.classify(DocumentCategorizerUpdateProcessorFactory.java:529) ~[?:?]
	at org.apache.solr.update.processor.DocumentCategorizerUpdateProcessorFactory$1.processAdd(DocumentCategorizerUpdateProcessorFactory.java:494) ~[?:?]
	at org.apache.solr.update.processor.UpdateRequestProcessor.processAdd(UpdateRequestProcessor.java:54) ~[?:?]
```

Also, we need to remove use of java.io.File...

epugh:
@jzonthemtn looks like your last commit means it's actually even EASIER to use Onnx based models?  Is this a new feature of huggingface in general?    If so, it's a lovely win!

rzo1:
Wonder if we should put the onxx on nightlies.a.o to avoid a cross dependency in CI to huggingface?

epugh:
> Wonder if we should put the onxx on nightlies.a.o to avoid a cross dependency in CI to huggingface?

I love your question!   Honestly, this is kind of uncharted territory in Solr, we've never had large-ish models that we used as part of tests or other infrastructure before.   Do you have some advice on how other projects have done this?   If you have some suggestions (and even better some code fixes) to push to this PR?

rzo1:
We are using nightlies.a.o. for our datatsets (4g b, compressed) used for our long running evaluation tests in OpenNLP. They are downloaded and used in our nightly eval runs on Jenkins CI.

Projects can just interact with nightlies.a.o. via WebDAV. We did ask for guidance because some of the datatsets are not ASLv2, but consensus was, that nightlies would be ok for it...

jzonthemtn:
> @jzonthemtn looks like your last commit means it's actually even EASIER to use Onnx based models? Is this a new feature of huggingface in general? If so, it's a lovely win!

No need to do the onnx conversion when there's already an onnx model available! So can simply and not need the Python dependencies.

+1 to @rzo1's suggestion about (eventually) putting the files on nightlies.

epugh:
I think instead of "eventually", @rzo1  got it done!  I just merged his PR to this PR

epugh:
@rzo1  any ideas on the redirect I am getting on nightlys?   it says I need to go to another site, but wget follow redirects not working or curl -l..

you can run the script via `./gradlew iTest --tests test_opennlp.bats`

rzo1:
@epugh Yes. The issue was that the original file was a redirect and **not** the actual model. I have replaced the onxx on nightlies (it's now about 161 MB), might need some time until the cache refreshes: https://nightlies.apache.org/solr/opennlp-dl-integration/ 

```
[   ] [ORIGIN](https://nightlies.apache.org/solr/opennlp-dl-integration/ORIGIN)                  2025-09-10 16:08  275   
[   ] [model_quantized.onnx](https://nightlies.apache.org/solr/opennlp-dl-integration/model_quantized.onnx)    2025-09-12 15:58  161M  
[TXT] [vocab.txt](https://nightlies.apache.org/solr/opennlp-dl-integration/vocab.txt)               2025-09-10 16:08  851K
```

Can you re-trigger the build? I am unable to run it locally due to missing `wget` on my osx ;-)

epugh:
> @epugh Yes. The issue was that the original file was a redirect and **not** the actual model. I have replaced the onxx on nightlies (it's now about 161 MB), might need some time until the cache refreshes: https://nightlies.apache.org/solr/opennlp-dl-integration/
> 
> ```
> [   ] [ORIGIN](https://nightlies.apache.org/solr/opennlp-dl-integration/ORIGIN)                  2025-09-10 16:08  275   
> [   ] [model_quantized.onnx](https://nightlies.apache.org/solr/opennlp-dl-integration/model_quantized.onnx)    2025-09-12 15:58  161M  
> [TXT] [vocab.txt](https://nightlies.apache.org/solr/opennlp-dl-integration/vocab.txt)               2025-09-10 16:08  851K
> ```

thank you!   Any tips on how to check if the cache refreshes?

rzo1:
Can you re-trigger the build? I am able to download it from via `curl` locally now.

PR #2006, opened 2023-10-12
Support ./gradlew modernizer

risdenk:

A long time ago I was looking at ways to improve the build process to detect out of date usages. This adds `./gradlew modernizer` that will find things that can be replaced.

based on https://github.com/andygoossens/gradle-modernizer-plugin

I went through all the failures and added some exclusions to make things clean. Mostly simple stuff.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,6 +28,7 @@ plugins {
   id 'net.ltgt.errorprone' version '3.1.0' apply false
   id 'com.diffplug.spotless' version '6.5.2' apply false
   id 'com.github.node-gradle.node' version '7.0.1' apply false
+  id 'com.github.andygoossens.modernizer' version '1.9.0' apply false
 }
 
 apply from: file('gradle/globals.gradle')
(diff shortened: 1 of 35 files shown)


stillalex (approved):
looks good. but mainly charset related changes?

risdenk:
Yup @stillalex this is just finding MOSTLY charset related changes right now. There are some disabled - https://github.com/apache/solr/pull/2006/files#diff-9257ac5900027c3c54700c25c646e201f7fcec13c7f39ecedd288f99c0ecb5f2R30. I don't know if this is worth it. We have cleaned up a lot of older usages with errorprone already.

cpoerschke (approved):

risdenk:
Interestingly errorprone https://github.com/google/error-prone/releases/tag/v2.23.0 might have just addressed the charset stuff so this PR might not be necessary.

stillalex:
interesting. if they add this then yeah, I agree this change is on longer useful.

risdenk:
Yea https://github.com/apache/solr/pull/2035 covers MOST of this.

PR #2003, opened 2023-10-11
SOLR-17027: Removing list.toArray calls with pre sized arrays - core

renatoh:
https://issues.apache.org/jira/browse/SOLR-17027

see discussing on old PR:
https://github.com/apache/solr/pull/1873

PR #2007, opened 2023-10-14
SOLR-17027: Removing list.toArray calls with pre sized arrays - part 2

renatoh:
https://issues.apache.org/jira/browse/SOLR-17027
fixing all List.toArray calls which use a pre-sized Array, use empty Array instead

Follow up to https://github.com/apache/solr/pull/2003


renatoh:
@risdenk 
I cleaned up the remaining toArray calls with the same issue, please have a look

risdenk (approved):

risdenk:
Thanks @renatoh

PR #2012, opened 2023-10-15
SOLR-17030: Open TransactionLog for extension.

bruno-roustant:
TransactionLog is already extended by HdfsTransactionLog. But the extension basically relies on an empty protected constructor, and HdfsTransactionLog re-implements everything.
We want to open more TransactionLog to extension so that an extending class could leverage the logic present in TransactionLog, and override mainly how output/input streams are opened. This extension will be used by an EncryptionTransactionLog in solr-sandbox.

No functional change, only some refactoring.

The change consists in introducing OutputStreamOpener and ChannelInputStreamOpener so that an extending class can replace the defaults by a custom implementation (e.g. stream openers supporting encryption).
Also fix the various flush() calls to ensure the delegate streams are also flushed.

### DIFF
--- a/solr/core/src/java/org/apache/solr/update/TransactionLog.java
+++ b/solr/core/src/java/org/apache/solr/update/TransactionLog.java
@@ -18,6 +18,7 @@ package org.apache.solr.update;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.invoke.MethodHandles;
 import java.nio.ByteBuffer;
@@ -63,17 +64,18 @@ import org.slf4j.LoggerFactory;
  */
 public class TransactionLog implements Closeable {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
-  private boolean debug = log.isDebugEnabled();
-  private boolean trace = log.isTraceEnabled();
+  private static final boolean debug = log.isDebugEnabled();
+  private static final boolean trace = log.isTraceEnabled();
 
   public static final String END_MESSAGE = "SOLR_TLOG_END";
 
   long id;
-  Path tlog;
-  FileChannel channel;
-  OutputStream os;
+  protected Path tlog;
+  protected FileChannel channel;
+  protected OutputStream os;
   // all accesses to this stream should be synchronized on "this" (The TransactionLog)
   protected FastOutputStream fos;
+  protected ChannelInputStreamOpener channelInputStreamOpener;
   int numRecords;
   public boolean isBuffer;
 
@@ -103,6 +105,10 @@ public class TransactionLog implements Closeable {
                   + "; try implementing ObjectResolver?");
         }
       };
+
+  protected static final OutputStreamOpener OUTPUT_STREAM_OPENER = (channel, position) -> Channels.newOutputStream(channel);
+
+  protected static final ChannelInputStreamOpener CHANNEL_INPUT_STREAM_OPENER = ChannelFastInputStream::new;
 
   public class LogCodec extends JavaBinCodec {
 
@@ -168,9 +174,18 @@ public class TransactionLog implements Closeable {
   }
 
   TransactionLog(Path tlogFile, Collection<String> globalStrings, boolean openExisting) {
+    this(tlogFile, globalStrings, openExisting, OUTPUT_STREAM_OPENER, CHANNEL_INPUT_STREAM_OPENER);
+  }
+
+  protected TransactionLog(Path tlogFile,
+                           Collection<String> globalStrings,
+                           boolean openExisting,
+                           OutputStreamOpener outputStreamOpener,
+                           ChannelInputStreamOpener channelInputStreamOpener) {
     boolean success = false;
     try {
       this.tlog = tlogFile;
+      this.channelInputStreamOpener = channelInputStreamOpener;
 
       if (debug) {
         log.debug(
@@ -190,14 +205,14 @@ public class TransactionLog implements Closeable {
 
         long start = Files.size(tlog);
         channel = FileChannel.open(tlog, StandardOpenOption.READ, StandardOpenOption.WRITE);
-        os = Channels.newOutputStream(channel);
-        fos = new FastOutputStream(os, new byte[65536], 0);
-
         if (start > 0) {
           readHeader(null);
+        }
+        os = outputStreamOpener.open(channel, start);
+        fos = new FastOutputStream(os, new byte[65536], 0);
+        if (start > 0) {
           channel.position(start);
-          fos.setWritten(start); // reflect that we aren't starting at the beginning
-          assert fos.size() == channel.size();
+          setWrittenCount(start);
         } else {
           addGlobalStrings(globalStrings);
         }
@@ -213,7 +228,7 @@ public class TransactionLog implements Closeable {
                 StandardOpenOption.READ,
                 StandardOpenOption.WRITE,
                 StandardOpenOption.CREATE_NEW);
-        os = Channels.newOutputStream(channel);
+        os = outputStreamOpener.open(channel, 0);
         fos = new FastOutputStream(os, new byte[65536], 0);
 
         addGlobalStrings(globalStrings);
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/update/TransactionLog.java
```diff
@@ -190,14 +205,14 @@ public boolean writePrimitive(Object val) throws IOException {
 
         long start = Files.size(tlog);
         channel = FileChannel.open(tlog, StandardOpenOption.READ, StandardOpenOption.WRITE);
-        os = Channels.newOutputStream(channel);
-        fos = new FastOutputStream(os, new byte[65536], 0);
-
         if (start > 0) {
           readHeader(null);
+        }
```
bruno-roustant, line 210: Change the order: first read the header, and then open the output stream.
This does not change this TransactionLog behavior, but it allows an extending class to read some metadata in the header and use it to open the output stream.

solr/core/src/java/org/apache/solr/update/TransactionLog.java
```diff
@@ -239,6 +254,20 @@ public boolean writePrimitive(Object val) throws IOException {
   // for subclasses
   protected TransactionLog() {}
 
+  /**
+   * Sets the counter of written data in the {@link FastOutputStream} view of the log file,
+   * to reflect that we aren't starting at the beginning.
+   */
+  protected void setWrittenCount(long fileStartOffset) throws IOException {
+    fos.setWritten(fileStartOffset);
```
bruno-roustant, line 262: If the extending class inserts some metadata in the file header, it needs here to adjust the offsets according to the metadata size.

risdenk (approved):

solr/core/src/java/org/apache/solr/update/TransactionLog.java
```diff
@@ -527,18 +562,18 @@ public Object lookup(long pos) {
       // make sure any unflushed buffer has been flushed
       synchronized (this) {
         // TODO: optimize this by keeping track of what we have flushed up to
-        fos.flushBuffer();
+        fos.flush();
```
stillalex, line 565: there is a difference between `FastOutputStream` `flush` and `flushBuffer` methods. have you verified this does not introduce some performance regression due to calling the flush method?

solr/core/src/java/org/apache/solr/update/TransactionLog.java
```diff
@@ -987,4 +1022,28 @@ public String toString() {
...
+  /** Opens {@link ChannelFastInputStream} from {@link FileChannel}. */
+  protected interface ChannelInputStreamOpener {
+
+    /**
+     * Opens a {@link ChannelFastInputStream} to read a {@link FileChannel}.
+     *
+     * @param position The initial read position of the {@link OutputStream} view of the {@link
+     *     FileChannel}.
+     */
+    ChannelFastInputStream open(FileChannel channel, long position) throws IOException;
```
stillalex, line 1047: the previous impl (calling `new ChannelFastInputStream`) did not throw IOException. introducing this new signature changes the flow a bit and I was curious if this was necessary. as in, instead of adding `throws ...` just take care of the IOException in the new impl you want to introduce.
what I am concerned about is the below change (https://github.com/apache/solr/pull/2012/files#diff-754514ecac361b4d69e8e7c9948f59befe8986afe643562ccb7f192af923f666R1932-R1933) where now you can have an UncheckedIOException bubble up, instead of the previous behavior of simply logging.

stillalex:
before merging could you run a few rounds of the CloudIndexing benchmark (https://github.com/apache/solr/blob/main/solr/benchmark/src/java/org/apache/solr/bench/index/CloudIndexing.java#L51)? I would like to see if we can check for performance regressions.
the benchmark is not perfect by any means, and I would like to see if there are improvements that can be made to it so we can more easily use it for cases like these.

solr/core/src/java/org/apache/solr/update/TransactionLog.java
bruno-roustant: Right, the methods are different. But the choice here of using flushBuffer is questionable.
(in addition, perf benchmark confirms no difference)

Line 565, fos.flushBuffer() followed by opening an input stream on the channel to read it. It works fine because fos delegates to a Channel output stream with no buffering. When fos.flushBuffer() writes to its delegate out, it writes directly to the channel. So calling fos.flush() which calls the delegate out.flush() is a no op. If now we want to support any delegate output stream with OutputStreamOpener.open(), then we need to call fos.flush().

Line 615, fos.flushBuffer() is called in the finish method to finish the transaction log. We should call flush() to ensure all intermediate output buffers are flushed.

Line 735, fos.flushBuffer() is called just before reading the corresponding input stream fis. Again, this works because the current delegate output stream has no buffering.

Line 876, fos.flushBuffer() is called just before asserting fos.size == channel.size(). Again, this works because the current delegate output stream has no buffering.

solr/core/src/java/org/apache/solr/update/TransactionLog.java
bruno-roustant: I added `throws IOException` to `getReader()` and `getSortedReader()` first for consistency with `getReverseReader()` which already throws IOE. I want the caller to know if there is an issue when creating a reader, and up to the caller to decide what to do.
Then, in `UpdateLog.doReplay()`, we have the choice of either catch and log and return, or throw an unchecked exception which is caught anyway in `UpdateLog.run()` to log it.

bruno-roustant:
Thanks Alex for the review and questions. Good point with the indexing benchmark.
So I ran it to confirm there is no perf impact.

I ran the benchmark twice with the current code (baseline):

```
# Run complete. Total time: 00:25:22

Benchmark               (nodeCount)  (numReplicas)  (numShards)  (preGenerate)  (useSmallDocs)   Mode  Cnt     Score      Error  Units
CloudIndexing.indexDoc            4              1            4          50000            true  thrpt    4  8167.548 ± 5407.518  ops/s
CloudIndexing.indexDoc            4              1            4          50000           false  thrpt    4   664.634 ±  207.693  ops/s
CloudIndexing.indexDoc            4              3            4          50000            true  thrpt    4  3153.489 ± 2220.162  ops/s
CloudIndexing.indexDoc            4              3            4          50000           false  thrpt    4   196.909 ±  168.691  ops/s

```
```
# Run complete. Total time: 00:26:35

Benchmark               (nodeCount)  (numReplicas)  (numShards)  (preGenerate)  (useSmallDocs)   Mode  Cnt     Score      Error  Units
CloudIndexing.indexDoc            4              1            4          50000            true  thrpt    4  7850.734 ± 7853.495  ops/s
CloudIndexing.indexDoc            4              1            4          50000           false  thrpt    4   718.769 ±  144.427  ops/s
CloudIndexing.indexDoc            4              3            4          50000            true  thrpt    4  3121.904 ± 1488.081  ops/s
CloudIndexing.indexDoc            4              3            4          50000           false  thrpt    4   208.965 ±  279.114  ops/s

```

And I ran the benchmark twice with the new code:

```
# Run complete. Total time: 00:25:34

Benchmark               (nodeCount)  (numReplicas)  (numShards)  (preGenerate)  (useSmallDocs)   Mode  Cnt     Score      Error  Units
CloudIndexing.indexDoc            4              1            4          50000            true  thrpt    4  7987.105 ± 7236.395  ops/s
CloudIndexing.indexDoc            4              1            4          50000           false  thrpt    4   720.472 ±  240.961  ops/s
CloudIndexing.indexDoc            4              3            4          50000            true  thrpt    4  3337.063 ± 2887.649  ops/s
CloudIndexing.indexDoc            4              3            4          50000           false  thrpt    4   199.760 ±  288.116  ops/s

```
```
# Run complete. Total time: 00:26:27

Benchmark               (nodeCount)  (numReplicas)  (numShards)  (preGenerate)  (useSmallDocs)   Mode  Cnt     Score      Error  Units
CloudIndexing.indexDoc            4              1            4          50000            true  thrpt    4  7851.275 ± 5788.400  ops/s
CloudIndexing.indexDoc            4              1            4          50000           false  thrpt    4   741.718 ±  158.800  ops/s
CloudIndexing.indexDoc            4              3            4          50000            true  thrpt    4  3034.854 ± 2308.357  ops/s
CloudIndexing.indexDoc            4              3            4          50000           false  thrpt    4   210.943 ±  254.704  ops/s

```

solr/core/src/java/org/apache/solr/update/TransactionLog.java
stillalex: ok I see what you are saying. it's all handled in `#run` method (this code is very confusing to read, some exceptions are handled locally some bubble up). I am ok with your changes, as you point out the exception is handled one level up so nothing changes in the flow.

stillalex (approved):

stillalex:
thank you for giving the benchmarks a go

solr/core/src/java/org/apache/solr/update/TransactionLog.java
stillalex: this makes sense to me. thank you for all the details.

PR #2016, opened 2023-10-16
SOLR-15782: Declare container-level plugins in solr.xml

pvcnt:
https://issues.apache.org/jira/browse/SOLR-15782

# Description

Currently, container-level plugins can only be declared installed via calls to /cluster/plugin APIs. In some situations (especially corporate ones having tight security requirements), it is not possible to mutate the configuration of a cluster at runtime. Cluster should rather be configured fully at startup, and any change in config requires a re-deployment.

# Solution

This PR makes it possible to declare container-level plugins in solr.xml. Those plugins will be installed when the system starts, and can later be updated/uninstalled using plugin APIs if needed. By design, the current implementation **works even in non-cloud installations**, which would allow to use container-level plugins in such situations.

This raises two questions:
* Do we envision having non-cloud container-level plugins, as this PR opens the way to? I am tempted to say it can be useful, but I don't know the original intention behind those plugins.
* There is a mixed terminology of container plugins and cluster plugins, both in the docs and in the code. Related to my previous point, "container plugins" hints at the possibility of non-cloud plugins, while "cluster plugins" gets a rid of this hypothesis.

# Tests

Added a new test class for this.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
+++ b/solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
@@ -68,6 +68,8 @@ import org.slf4j.LoggerFactory;
  * removing / replacing the plugins according to the updated configuration obtained from {@link
  * ContainerPluginsApi#plugins(Supplier)}.
  *
+ * <p>It also handles plugins declared in solr.xml, which are loaded at startup.
+ *
  * <p>Plugins instantiated by this class may implement zero or more {@link Api}-s, which are then
  * registered in the CoreContainer {@link ApiBag}. They may be also post-processed for additional
  * functionality by {@link PluginRegistryListener}-s registered with this class.
@@ -120,6 +122,33 @@ public class ContainerPluginsRegistry implements ClusterPropertiesListener, MapW
   public ContainerPluginsRegistry(CoreContainer coreContainer, ApiBag apiBag) {
     this.coreContainer = coreContainer;
     this.containerApiBag = apiBag;
+    loadFromNodeConfig();
+  }
+
+  private void loadFromNodeConfig() {
+    // Load plugins defined in solr.xml. Those plugins are then considered exactly like
+    // other plugins, e.g., they can be updated or uninstalled.
+    for (PluginInfo pluginInfo : coreContainer.getNodeConfig().getContainerPlugins()) {
+      Map<String, Object> info = new HashMap<>();
+      info.put("name", pluginInfo.name);
+      info.put("class", pluginInfo.className);
+      String version = pluginInfo.attributes.get("version");
+      if (null != version) {
+        info.put("version", version);
+      }
+      String pathPrefix = pluginInfo.attributes.get("path-prefix");
+      if (null != pathPrefix) {
+        info.put("path-prefix", pathPrefix);
+      }
+      if (pluginInfo.initArgs.size() > 0) {
+        info.put("config", pluginInfo.initArgs.toMap(new HashMap<>()));
+      }
+      try {
+        addOrUpdatePlugin(Diff.ADDED, pluginInfo.name, new PluginMetaHolder(info));
+      } catch (Exception exp) {
+        log.error("Invalid pluginInfo configuration", exp);
+      }
+    }
   }
 
   @Override
@@ -167,6 +196,7 @@ public class ContainerPluginsRegistry implements ClusterPropertiesListener, MapW
     }
   }
 
+  /** Update the list of container plugins from configuration stored in ZK. */
   @SuppressWarnings("unchecked")
   public synchronized void refresh() {
     Map<String, Object> pluginInfos;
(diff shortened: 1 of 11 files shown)


solr/core/src/test/org/apache/solr/api/ContainerPluginsRegistryTest.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Map;
+import org.apache.solr.SolrTestCaseJ4;
+import org.apache.solr.cluster.placement.plugins.AffinityPlacementConfig;
+import org.apache.solr.core.CoreContainer;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+/** Tests for {@link ContainerPluginsRegistry} */
+public class ContainerPluginsRegistryTest extends SolrTestCaseJ4 {
```
pvcnt, line 13: There is an existing `TestContainerPlugin` class that tests the same area of code, but in a purely Solr Cloud context. Whether or not to merge both depends on whether or not we want to go with cluster or container plugins imho.

solr/core/src/test/org/apache/solr/core/TestSolrXml.java
```diff
@@ -78,6 +78,7 @@ public void testAllInfoPresent() throws IOException {
     assertEquals("info handler class", "testInfoHandler", cfg.getInfoHandlerClass());
     assertEquals(
         "config set handler class", "testConfigSetsHandler", cfg.getConfigSetsHandlerClass());
+    assertEquals("cores locator class", "testCoresLocator", cfg.getCoresLocatorClass());
```
pvcnt, line 81: I did not see this test in a previous PR of mine, piggy backing here.

solr/solrj/src/java/org/apache/solr/client/solrj/request/beans/PluginMeta.java
```diff
@@ -48,6 +48,7 @@ public PluginMeta copy() {
     result.name = name;
     result.klass = klass;
     result.version = version;
+    result.pathPrefix = pathPrefix;
```
pvcnt, line 51: Looks like it was forgotten.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -776,6 +775,7 @@ private void loadInternal() {
     ClusterEventProducerFactory clusterEventProducerFactory = new ClusterEventProducerFactory(this);
     clusterEventProducer = clusterEventProducerFactory;
 
+    containerPluginsRegistry = new ContainerPluginsRegistry(this, containerHandlers.getApiBag());
```
pvcnt, line 778: Need to instantiate only when CoreContainer has a NodeConfig ready.

dsmiley:
> Do we envision having non-cloud container-level plugins, as this PR opens the way to? I am tempted to say it can be useful, but I don't know the original intention behind those plugins.

I think so!  There shouldn't be anything SolrCloud/ZK specific about this PR.  As I'm sure you know, Solr runs in standalone and has a solr.xml and there are already specific named plugins that work there.  These are "node plugins", albeit of a specific expected type for a specific purpose.  Like CoresLocator, which I think you are familiar with :-). All classes in NodeConfig/solr.xml are a "node plugin" by virtue of a pluggable class specified in this spot.  You are generalizing this to have more for more general purposes.  And for new handlers as a possible purpose.  That's how I think of all this.

dsmiley (commented):
I'd like to help on the documentation side.  cluster-plugins.adoc is kind of a challenge.  One part of it is an HTTP API for CRUD verbs on these plugins when they are managed in a cluster way (stored in ZK).  Another part of it is documenting that certain plugin types (1) exist (some like ClusterSingleton are not documented anywhere else at all), and (2) are registered in this way (not in other ways).  It's disharmonious to have this dichotomy of plugins that have always been in solr.xml (and even new ones that showed up yesterday like CoresLocator) who have been true node plugins all along, and in parallel have a separate world of node/cluster plugins that are declared in a different way in solr.xml (don't look like the others there), or you can use the "cluster plugin API" runtime alternative.   Open-source can be messy; contributions from different people, different ideas, and then later we see a confusing mess emerge.  From experience, documenting/explaining something like I'm trying to do here exposes a problem.  Need to think about this.  I wish it could all be unified.

solr/core/src/test/org/apache/solr/api/ContainerPluginsRegistryTest.java
pvcnt: Yes, I agree, that is why I did not reuse the existing test class, considering that the feature I was adding was not specific to Solr cloud. I would like as well to have another PR for the various renamings, if this one gets through.

solr/solrj/src/java/org/apache/solr/client/solrj/request/beans/PluginMeta.java
pvcnt: It was making my test to fail, because `ContainerPluginsRegistry#getPlugin` calls `copy` and `pathPrefix` was lost.

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
```diff
@@ -120,6 +123,33 @@ public void unregisterListener(PluginRegistryListener listener) {
...
+      if (null != pathPrefix) {
+        info.put("path-prefix", pathPrefix);
+      }
+      if (pluginInfo.initArgs.size() > 0) {
+        info.put("config", pluginInfo.initArgs.toMap(new HashMap<>()));
+      }
+      try {
+        addOrUpdatePlugin(Diff.ADDED, pluginInfo.name, new PluginMetaHolder(info));
+      } catch (Exception exp) {
+        log.error("Invalid pluginInfo configuration", exp);
```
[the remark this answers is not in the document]
pvcnt: If I remember well, it should technically never happen. The only checked exception is thrown in the constructor of `PluginMetaHolder` where the JSON is parsed, but here JSON is built manually. I would have liked to get a rid of that step btw, but I was not comfortable with changing the existing behaviour too much.

solr/core/src/java/org/apache/solr/core/SolrXmlConfig.java
```diff
@@ -659,6 +660,17 @@ private static PluginInfo[] getBackupRepositoryPluginInfos(List<ConfigNode> cfg)
...
+  private static PluginInfo[] getNodePluginInfos(List<ConfigNode> cfg) {
+    if (cfg.isEmpty()) {
+      return new PluginInfo[0];
+    }
+    PluginInfo[] configs = new PluginInfo[cfg.size()];
+    for (int i = 0; i < cfg.size(); i++) {
+      configs[i] = new PluginInfo(cfg.get(i), "NodePlugin", true, true);
+    }
+    return configs;
+  }
```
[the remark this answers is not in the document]
pvcnt: Sure, I copy/pasted this from elsewhere. 🙄

solr/solr-ref-guide/modules/configuration-guide/pages/cluster-plugins.adoc
```diff
@@ -278,3 +276,34 @@ curl -X POST -H 'Content-type: application/json' -d '{
     }'
   http://localhost:8983/api/cluster/plugin
 ----
+
+== Plugin Configuration in solr.xml
+
+Plugins can also be installed by creating `<plugin>` elements in `solr.xml`.
+This might be preferred in some situations, such as enforcing immutable infrastructure.
+
+TIP: See also the section xref:configuration-guide:configuring-solr-xml.adoc[] for more information about the `solr.xml` file, where to find it, and how to edit it.
```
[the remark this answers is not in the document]
pvcnt: I was unsure, because I had the impression that the trend was to split the documentation about `solr.xml` in more specialised pages. For example, metric configuration is already described on its own page.

solr/solr-ref-guide/modules/configuration-guide/pages/cluster-plugins.adoc
```diff
@@ -88,24 +88,22 @@ A JSON map of additional plugin configuration parameters.
 Plugins that implement `ConfigurablePlugin` interface will be initialized with a
 plugin-specific configuration object deserialized from this map.
 
-Example plugin configuration:
+Example plugin with configuration:
 
-[source,bash]
+[source,json]
 ----
-curl -X POST -H 'Content-type: application/json' -d '{
```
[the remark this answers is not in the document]
pvcnt: I did it because I was trying to fit in the same page documentation of plugins configured at node- and cluster-level. This command was shown when describing what a plugin looks like in JSON, before the cluster API was even presented. What matters here is hence the contents of the JSON, the curl command appeared as noise to me - and a forward reference to some concept not already presented.

The way I was seeing it, documentation should present three things:
1. A general introduction of node plugins, and what their properties are.
2. How to configure them in solr.xml
3. How to configure them with cluster API

pvcnt:
Superseded by #2071

PR #2020, opened 2023-10-17
SOLR-17034: Hitting /solr// ends up with HttpSolrCall NPE

risdenk:
https://issues.apache.org/jira/browse/SOLR-17034


dsmiley (approved):
I assume you tested manually.  A trivial test somewhere would be nice.  Could use LogListener to detect the NPE then lack of it after fix.

epugh (approved):
LGTM

risdenk:
Yea I looked at adding a test for this but couldn't find a good place. I think its pretty obvious that `origCoreName` can be null since a few lines later is on line 1181 is:

```
if (origCorename != null) {
```

so there is obviously a need for this check in normal cases.

PR #2021, opened 2023-10-17
SOLR-17036: Lazy create VersionBucket in the update log to free memory.

bruno-roustant:
Modify VersionInfo to not create the large array of VersionBucket at initialization. Instead start with a hash map, and replace it by the array if it grows beyond the threshold. In addition, BucketVersion instances are created lazily.

Empty cores require much less memory.
Cores with no update will not fill the map.
Cores with less updates before they are closed will stay with a small sized map.


solr/core/src/java/org/apache/solr/update/VersionInfo.java
```diff
@@ -103,14 +113,14 @@ public VersionInfo(UpdateLog ulog, int nBuckets) {
...
-    buckets = new VersionBucket[BitUtil.nextHighestPowerOfTwo(nBuckets)];
-    for (int i = 0; i < buckets.length; i++) {
-      if (versionBucketLockTimeoutMs > 0) {
-        buckets[i] = new TimedVersionBucket();
-      } else {
-        buckets[i] = new VersionBucket();
-      }
-    }
+    numBuckets = BitUtil.nextHighestPowerOfTwo(nBuckets);
+    // Instead of directly creating the large bucket array which takes 1.5 MB per SolrCore,
```
dsmiley, line 117: state the assumption of 65k buckets

dsmiley (commented):
It'd be nice to take this opportunity to document a bit more what's going on at a higher level.  Like, what is a "version bucket" anyway; what's it for.  Why all these locks.  Maybe you don't know; if not no problem.  My understanding is that Solr needs to lock a document ID to do atomic operations for that ID to prevent other updates for that same doc ID that might happen concurrently.  And we don't want to have an Object or Lock per doc ID as doing so could be challenging; maybe GC issue to create and remove not sure.  So we use a striped lock technique with a hash of the doc ID capping the possible locks we may need, and we keep them in memory. I would love it if someone could check me on this and embellish it.

bruno-roustant:
Here is my understanding.

VersionBucket
It serves 2 purposes.
First, it is a lock object to lock atomic operations on a doc ID. It is a bucket because it is used to lock operations of all the doc IDs which hash falls into this bucket.
Second, it keeps the highest version which is the max of the versions of the docs in the bucket. This is an optimization only when the leader forwards the update request to another replica (there could be an option to not store this long if only the leader updates and never forwards). When adding a doc with version v, to compare first with the highest version vh of the bucket. If v > vh, then we know the doc version is ordered, without having to look in the transaction log or index for the precise indexed doc version.

Why 65536 VersionBucket?
For both VersionBucket goals, the more buckets there are, the more precise is the locking and the highest optimization. The drawback is the memory usage.
In SOLR-XXX, the number of buckets was studied. If there are not enough buckets, it happens that update threads are blocked waiting on the same lock when an update operation takes a long time. With a large number of buckets, the probability of updates locking the same bucket, within the duration of a long operation, is sufficiently low to not impact perf.

Can we manage buckets dynamically?
For the locking aspect, yes, but there is more synchronization required if we support bucket removal. If we only support lazy creation, we only need to synchronize when creating a bucket.
For the highest version optimization, partially, we cannot remove the highest value if it is different than the common "seed" value.

solr/core/src/java/org/apache/solr/update/VersionBucket.java
```diff
@@ -31,14 +31,41 @@
...
 public class VersionBucket {
-  public long highest;
 
+  private long highest;
+
+  /**
+   * @param highest the initial value of this bucket highest version.
+   */
+  public VersionBucket(long highest) {
+    setHighestIfGreater(highest);
```
stillalex, line 41: if I am reading the code correctly, this will be a noop because the `setHighestIfGreater ` method checks `if (highest != 0)` which coming from the constructor will always be false.

solr/core/src/java/org/apache/solr/update/VersionInfo.java
```diff
@@ -205,9 +215,59 @@ public VersionBucket bucket(int hash) {
...
+    return getBucketFromArray(slot);
+  }
 
-    int slot = hash & (buckets.length - 1);
-    return buckets[slot];
+  private VersionBucket getBucketFromMap(int slot) {
+    VersionBucket bucket = bucketMap.get(slot);
+    if (bucket == null) {
+      synchronized (bucketsSync) {
+        int index = bucketMap.indexOf(slot);
```
stillalex, line 229: is this a potential NPE? if a thread is waiting on the lock while in the background the map turns into an array. thread gets the lock, does a `bucketMap.indexOf` but the bucketMap is null by the transformation.

solr/core/src/java/org/apache/solr/update/VersionInfo.java
```diff
@@ -284,12 +344,28 @@ public Long getMaxVersionFromIndex(IndexSearcher searcher) throws IOException {
...
+      } else {
+        for (VersionBucket bucket : bucketArray) {
+          if (bucket != null) {
+            synchronized (bucket) {
+              bucket.setHighestIfGreater(highestVersion);
+            }
+          }
+        }
       }
+      highestVersionSeed = Math.max(highestVersion, highestVersionSeed);
```
stillalex, line 368: honestly trying to wrap my head around the use of bucket's internal version. why wouldn't all the buckets get updated to this `max(highestVersion, highestVersionSeed)` instead of `highestVersion`?

solr/core/src/java/org/apache/solr/update/VersionInfo.java
bruno-roustant: Yes, good catch.

solr/core/src/java/org/apache/solr/update/VersionInfo.java
bruno-roustant: That would be actually the same, as highest can only increase in VersionBucket.
Here highestVersionSeed is updated with the max to seed the next VersionBucket to create. If highestVersionSeed was set to highestVersion, it could potentially decrease.

bruno-roustant:
Actually, based on an a separate discussion with David, I'll also propose a second PR simplified with the lazy creation of the bucket array, without the map transition.

dsmiley:
My feedback to Bruno today was overall the this Map upgrade to array (the array we have today) seems kind of pointless; sort of inevitable (assuming there is indexing) so why have this earlier Map incarnation.  The actual sought-after benefit that would be nice would be to not have the array at all if there is no indexing here.  So just a lazy creation of what exists today would be nice.

If all we needed was a lock for the ID, we could use just a ConcurrentHashSet of the IDs to represent holding a lock.  But clearly there is a "version" being tracked here, which I confess I don't understand the purpose behind.

bruno-roustant:
I read the code more. The VersionBucket is also used by non-leaders to wait for dependent updates in the case of in-place update. So it's not only a lock for atomic access.

bruno-roustant:
I removed the Map transition from this PR. The VersionBucket array is created lazily.

solr/core/src/java/org/apache/solr/update/VersionInfo.java
```diff
@@ -284,11 +297,20 @@ public Long getMaxVersionFromIndex(IndexSearcher searcher) throws IOException {
...
+          highestVersionSeed = Math.max(highestVersion, highestVersionSeed);
+          return;
+        }
+      }
+    }
+    for (VersionBucket bucket : buckets) {
+      // Should not happen, but synchronize in case other threads are calling
+      // updateHighest on the version bucket.
+      synchronized (bucket) {
+        bucket.setHighestIfGreater(highestVersion);
```
dsmiley, line 313: should the synchronization move to be on setHighestIfGreater ?

solr/core/src/java/org/apache/solr/update/VersionBucket.java
```diff
@@ -31,14 +31,41 @@
...
+   * Sets the highest version if the provided value is greater than the current bucket value. The
+   * caller must synchronize on this bucket when calling this method.
+   */
+  public void setHighestIfGreater(long val) {
+    if (val > highest) {
+      this.highest = val;
     }
   }
 
+  public long getHighest() {
```
dsmiley, line 65: If writers must synchronize, it's probably true that readers should synchronize?  Or make this method synchronized?

dsmiley (commented):
Looks good!  I think there's some confusion as to whose responsibility it is to synchronize.  It would be clearer/safer if VersionBucket's methods were synchronized.  Or assert [Thread.holdsLock(this)](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Thread.html#holdsLock(java.lang.Object)) say.

solr/core/src/java/org/apache/solr/update/VersionInfo.java
```diff
@@ -284,11 +297,20 @@ public Long getMaxVersionFromIndex(IndexSearcher searcher) throws IOException {
...
+    if (buckets == null) {
+      synchronized (bucketsSync) {
+        if (buckets == null) {
+          // Update the highest version seed to use if/when buckets are created.
+          highestVersionSeed = Math.max(highestVersion, highestVersionSeed);
+          return;
+        }
+      }
+    }
+    for (VersionBucket bucket : buckets) {
```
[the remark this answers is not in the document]
bruno-roustant: An "else" on which "if" block?

solr/core/src/java/org/apache/solr/update/VersionBucket.java
bruno-roustant: I'll use a volatile. This fixes a potential multi-thread memory visibility bug in the existing code.
Actually it's another hint that makes me wonder whether this version optimization has any perf improvement impact. I'll run the CloudIndexing benchmark to try to see something when I disable it.

dsmiley: The methods we're talking about are really trivial.  And the concurrency concern is low probability.  I wouldn't worry about a negative perf impact that you hurt anything here.

bruno-roustant:
Ok, here is a big change in the PR. After thorough tests and benchmarks, I propose to simply remove the "highest version" optimization.
This is only an optimization. So I tested by deactivating it and it works (tests pass) and it does not reduce perf (SolrIndexing benchmark). Actually it seems it even improve perf in some bench cases, maybe due to the removal of unneeded methods to get the max version from the index.

This new commit removes the long "highest" from VersionBucket, and also removes all the methods in VersionInfo and UpdateLog to compute the max version. By reading the code and by running the tests, I'm convinced they were used only by tests and for an unnecessary optimization. In the end, we also reduce by 500 KB the memory required by VersionInfo (for 64K VersionBucket), going down to 1 MB (lazy allocated in this PR).

dsmiley:
Hey way to go Bruno!  Thus essentially all we need is a lock per ID (not even necessarily "bucket"-ed); the version was an optimization and doesn't seem to be worth its complexity? @markrmiller -- really looking for your feedback.  If true, I'd love to take a stab at a large simplification further from here in which there aren't buckets either :-). (I've been thinking of this; how to do it simply)

bruno-roustant:
I'm going to merge this in 2 days.

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -507,29 +504,19 @@ private boolean doVersionAdd(
...
           } else {
             // if we aren't the leader, then we need to check that updates were not re-ordered
-            if (bucketVersion != 0 && bucketVersion < versionOnUpdate) {
-              // we're OK... this update has a version higher than anything we've seen
-              // in this bucket so far, so we know that no reordering has yet occurred.
-              bucket.updateHighest(versionOnUpdate);
-            } else {
-              // there have been updates higher than the current update. we need to check
-              // the specific version for this id.
-              Long lastVersion = vinfo.lookupVersion(cmd.getIndexedId());
```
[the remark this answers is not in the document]
bruno-roustant: I agree this must have been the objective of the max version per bucket. I also agree that, when the optimization triggers, it avoids the vinfo.lookupVersion().
What I questioned is the frequency of when the optimization triggered, as it occurs when the doc version is higher than the max version of all docs in the bucket. That's why I ran multiple times the indexing benchmark to see if there is a difference with and without the optimization.
My take on this is that the gain (apparently very low to none from the benchmark) is not worth the complexity and the additional memory per Core.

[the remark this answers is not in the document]
bruno-roustant: Looking forward to your follow up!

PR #2023, opened 2023-10-18
SOLR-17039: Entropy calculation fails in Docker due to missing 'bc' cmd

janhoy:
https://issues.apache.org/jira/browse/SOLR-17039

@chatman


HoustonPutman:
Huh very cool, I learn more about bash every day!

risdenk (approved):

janhoy:
> Huh very cool, I learn more about bash every day!

And I learnt it from ChatGPT by giving it the old if statement and asking it to get rid of "bc" :-) I could of course also have used a plain `if [[ $ratio -lt 25 ]]` but I think the double parens were nice

PR #2025, opened 2023-10-18
SOLR-13748: Add support for mm parameter to bool query parser

AndreyBozhko:
https://issues.apache.org/jira/browse/SOLR-13748

# Description

Added support for `mm` parameter to {!bool} query parser

# Tests

Added test suite for BoolQParserPlugin

# Checklist


AndreyBozhko:
Since the similar parameter for DisMax/eDisMax query parsers is also named `mm`, perhaps it makes sense to keep the naming consistent?

HoustonPutman (approved):
This looks great to me! Very complete PR 🙂

AndreyBozhko:
Thanks for the review @HoustonPutman, feel free to merge this whenever you have time.

solr/core/src/java/org/apache/solr/search/join/FiltersQParser.java
```diff
@@ -61,6 +61,9 @@ protected BooleanQuery parseImpl() throws SyntaxError {
     for (Map.Entry<QParser, Occur> clause : clauses.entrySet()) {
       builder.add(unwrapQuery(clause.getKey().getQuery(), clause.getValue()), clause.getValue());
     }
+
+    builder.setMinimumNumberShouldMatch(minShouldMatch());
```
mkhludnev, line 65: Not sure why. MinShouldMatch matters only for SHOULD clauses, but never for FILTER ever.

solr/solr-ref-guide/modules/query-guide/pages/other-parsers.adoc
```diff
@@ -75,6 +75,16 @@ However, unlike `must`, the score of filter queries is ignored.
...
 
+`mm`::
++
+[%autowidth,frame=none]
+|===
+|Optional |Default: `0`
+|===
++
+The number of optional clauses that must match. By default, no optional clauses are necessary for a match
+(unless there are no required clauses). If this parameter is set, then the specified number of `should` clauses is required.
```
mkhludnev, line 86: It's probably worth to clarify that when there are no required clauses it works like mm=1.

mkhludnev (approved):
Thanks. Pls consider comments

solr/solr-ref-guide/modules/query-guide/pages/other-parsers.adoc
AndreyBozhko: Thanks for the suggestion @mkhludnev, I updated the description.

solr/core/src/java/org/apache/solr/search/join/FiltersQParser.java
AndreyBozhko: Could it be because FILTER clauses are considered required, and never optional?

Also, if one needs to use, say, three clauses in the filter query and make sure that at least two match, - it should be possible to construct such filter query as `{!bool filter=...}`, where the only FILTER clause is itself a boolean query with three SHOULD clauses and mm=2.

mkhludnev: > Could it be because FILTER clauses are considered required, and never optional?

indeed [Like MUST](https://lucene.apache.org/core/9_4_1/core/org/apache/lucene/search/BooleanClause.Occur.html#FILTER) 

> 
> Also, if one needs to use, say, three clauses in the filter query and make sure that at least two match, - it should be possible to construct such filter query as `{!bool filter=...}`, where the only FILTER clause is itself a boolean query with three SHOULD clauses and mm=2.

perhaps, 

1. wrap a such of bool of should via single clause filter. 
2. wipe the score of that `{!bool should=... mm=2}^=0`
3. or wipe scores of underneath filters via `{!bool should=A:b^=0  should=C:d^=0 should=E:f^=0 mm=2}`

NB: I'm not sure whether parser is capable to parse those options above. If it isn't, one can trick it with local param referencing.

mkhludnev:
Thanks @AndreyBozhko I'll check it next week

HoustonPutman (approved):

AndreyBozhko (commented):
Thanks, LGTM.

PR #2038, opened 2023-10-23
SOLR-17054: Remove unused and duplicate code in DistributedZkUpdateProcessor

pvcnt:
https://issues.apache.org/jira/browse/SOLR-17054

# Description

There is dead code and duplicate code in DistributedZkUpdateProcessor. It makes understanding this already complex class more difficult.

# Solution

Dead code and duplicate code has been removed, using automated refactoring tools available in Intellij whenever possible, to reduce the risk of errors.

# Tests

No new code.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/update/processor/DistributedZkUpdateProcessor.java
+++ b/solr/core/src/java/org/apache/solr/update/processor/DistributedZkUpdateProcessor.java
@@ -213,8 +213,8 @@ public class DistributedZkUpdateProcessor extends DistributedUpdateProcessor {
         params.set(DISTRIB_UPDATE_PARAM, DistribPhase.TOLEADER.toString());
         params.set(COMMIT_END_POINT, "leaders");
         params.set(
-                DISTRIB_FROM,
-                ZkCoreNodeProps.getCoreUrl(zkController.getBaseUrl(), req.getCore().getName()));
+            DISTRIB_FROM,
+            ZkCoreNodeProps.getCoreUrl(zkController.getBaseUrl(), req.getCore().getName()));
         cmdDistrib.distribCommit(cmd, useNodes, params);
         issuedDistribCommit = true;
       }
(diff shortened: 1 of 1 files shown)


solr/core/src/java/org/apache/solr/update/processor/DistributedZkUpdateProcessor.java
```diff
@@ -212,13 +212,11 @@ public void processCommit(CommitUpdateCommand cmd) throws IOException {
         useNodes = nodes;
         params.set(DISTRIB_UPDATE_PARAM, DistribPhase.TOLEADER.toString());
         params.set(COMMIT_END_POINT, "leaders");
-        if (useNodes != null) {
```
pvcnt, line 215: Was always non-null.

solr/core/src/java/org/apache/solr/update/processor/DistributedZkUpdateProcessor.java
```diff
@@ -786,52 +784,7 @@ protected List<SolrCmdDistributor.Node> setupRequest(
         // that means I want to forward onto my replicas...
         // so get the replicas...
         forwardToLeader = false;
-        String leaderCoreNodeName = leaderReplica.getName();
```
pvcnt, line 789: Was exactly the same as `getReplicaNodesForLeader`, except for the timeout (has been added as an argument of the method).

solr/core/src/java/org/apache/solr/update/processor/DistributedZkUpdateProcessor.java
```diff
@@ -883,8 +836,7 @@ private void checkReplicationTracker(UpdateCommand cmd) {
     }
   }
 
-  private List<SolrCmdDistributor.Node> getCollectionUrls(
-      String collection, EnumSet<Replica.Type> types, boolean onlyLeaders) {
```
pvcnt, line 887: `onlyLeaders` was always true therefore `types` was not used.

bruno-roustant: I looked at the git history.
This code comes from a refactor in 2019 of DistributedUpdateProcessor to extract the code specific to cloud. At that time this method signature already contained the boolean.
Looking earlier, I saw that the boolean was added in 2018 in a commit “Make massive improvements to the tests”. It added the boolean and modified the single call to pass true. Well, I believe it was a perf optimization to always limit to leaders. At that time the intent was clearly to ignore the dead code, and I think the dead code should have been removed from the method instead of adding this boolean.
So after looking at the context, I’m +1 to remove this dead code.

bruno-roustant:
Could you create a JIRA issue for this PR to give it more visibility?
I agree it's only a cleanup, but since I investigated to be sure of the dead code removal, for me it means it's not fully trivial so I expect to have a JIRA in this case.

pvcnt:
@bruno-roustant JIRA ticket has been created: https://issues.apache.org/jira/browse/SOLR-17054

bruno-roustant:
Thanks Vincent, let's merge it!

PR #2040, opened 2023-10-24
SOLR-16907: Reserve some 8.x security permission names

janhoy:
https://issues.apache.org/jira/browse/SOLR-16907


risdenk (approved):

uschindler:
In the issue you wanted to add a warning of an unknown permission occurred. Should we add this, too?

janhoy:
> In the issue you wanted to add a warning of an unknown permission occurred. Should we add this, too?

Agree. The updated PR now instead of maintaining a list of old "reserved" permission names, now adds a validation when parsing a permission that is not a predefined one. The requirement is that it must contain one of these keys: "method", "path", "params". I believe it is correct to assume that every custom permission must contain one of those, am I correct.

In other words, a permission like this would be invalid:
```json
{ "name": "hello", "role": "admin", "collection": "test", "index": 1 }
```

solr/core/src/java/org/apache/solr/security/Permission.java
```diff
@@ -160,7 +164,12 @@ public String toString() {
...
   }
 
+  static final Set<String> predefinedPermissionAllowedKeys =
+      Set.of("collection", "role", NAME, "index");
+  static final Set<String> customPermissionAdditionalKeys = Set.of("method", "path", "params");
   static final Set<String> knownKeys =
-      Set.of("collection", "role", "params", "path", "method", NAME, "index");
+      Stream.concat(
+              predefinedPermissionAllowedKeys.stream(), customPermissionAdditionalKeys.stream())
+          .collect(Collectors.toSet());
```
uschindler, line 173: Should better be `Collectors.toUnmodifiableSet`.

solr/core/src/java/org/apache/solr/security/Permission.java
```diff
@@ -52,18 +53,21 @@ static Permission load(Map<?, ?> m) {
...
-      disAllowed.remove("collection"); // allowed keys for well-known permissions
-      disAllowed.remove("index");
-      for (String s : disAllowed) {
+      for (String s : customPermissionAdditionalKeys) {
         if (m.containsKey(s))
           throw new SolrException(
               SolrException.ErrorCode.BAD_REQUEST,
               s + " is not a valid key for the permission : " + name);
       }
+    } else if (customPermissionAdditionalKeys.stream().noneMatch(m.keySet()::contains)) {
```
uschindler, line 62: It might be better to iterate over stream of `m.keySet()` if it is generally smaller in size. I don't know the details but to me it looks like `m.size()` is generally small!?

uschindler (approved):
Looks fine and much better as it is fully dynamic.

This also helps people who have spelling errors in their permissions, not only legacy names.

I just added 2 small comments. I am not sure about the order on how to intersect the sets. The smaller one should be iterated on, the larger one queries. Guava has a nice view for this, but we wanted to get rid of guava.

uschindler:
I font know internals of how the permissions work, so I only checked the code under the assumption that you know what a correct permission name looks like. Maybe cross check with @risdenk

solr/core/src/java/org/apache/solr/security/Permission.java
janhoy: I tried the other way around but IntelliJ complained about "suspicious" code, and somehow it did not work either. There are only a few values anyway and this code is run once during parsing of security.json, so leaving as is.

uschindler: Funny. Suspicious. 😜

uschindler: You could also use: `m::containsKey`

uschindler:
Thanks for the change, looks fine.

uschindler (approved):

solr/core/src/java/org/apache/solr/security/Permission.java
```diff
@@ -52,18 +53,21 @@ static Permission load(Map<?, ?> m) {
...
-      disAllowed.remove("index");
-      for (String s : disAllowed) {
+      for (String s : customPermissionAdditionalKeys) {
         if (m.containsKey(s))
           throw new SolrException(
               SolrException.ErrorCode.BAD_REQUEST,
               s + " is not a valid key for the permission : " + name);
       }
+    } else if (customPermissionAdditionalKeys.stream().noneMatch(m::containsKey)) {
+      // Custom permissions must contain one of the additional keys to be valid
```
janhoy, line 63: @risdenk Do you agree with the above assumption, i.e. that a custom permission MUST have either a "path", "method" or "params" key? Perhaps the rule could be simplified to say it must have a "path" key?

risdenk: based on https://solr.apache.org/guide/solr/latest/deployment-guide/rule-based-authorization-plugin.html#custom-permissions - pretty much all parameters are optional. `path` is not required. 

it looks like `collection` is missing as one of the options to `customPermissionAdditionalKeys`

janhoy: Yea, not sure the docs are precise here. How can you configure a custom permission without a `path`? Like the below one, it does not make sense as it does not define what endpoint to protect. You'd need a `"path": "/select"` for that:

```json
{ "name": "myColl", "role": "admin", "collection": "test" }
```

And for a custom admin request (no collection), it makes no sense to specify "method" or "params" without a "path":

```json
{ "name": "myAdmin", "role": "admin", "method": "GET" }
```

Thus I believe we must require a `path` config for custom perms.

janhoy: So I changed it to require "path" only, will see if tests pass now.

janhoy:
@risdenk, @noblepaul Please have another look. Now the logic is
* First look for well-known permission name (as before)
* If **not** a well-known name, and `path` is not specified, fail with error that custom permission is mis-configured

This will both catch bad custom permissions, mis-spelled well-known permissions and use of old well-known permissions that no longer exist in a newer version.

So if my assumption is true that it makes no sense to have a custom permission without a "path" key, then this should be good.

epugh (approved):
LGTM.  This kind of feature makes it easier for users to start "doing the right thing" when it comes to security!

solr/core/src/java/org/apache/solr/security/Permission.java
risdenk: @janhoy if we are making path required (or at least making it explicit now) the docs should be updated as well?

risdenk (approved):

PR #2042, opened 2023-10-26
Capture build scans on ge.apache.org to benefit from deep build insights

clayburn:
# Description

@dsmiley - This is the similar PR to https://github.com/apache/lucene/pull/12293

This PR publishes a build scan for every CI build on Jenkins and GitHub Actions and for every local build from an authenticated Apache committer. The build will not fail if publishing fails.

The build scans of the Apache Solr project are published to the Gradle Enterprise instance at [ge.apache.org](https://ge.apache.org/), hosted by the Apache Software Foundation and run in partnership between the ASF and Gradle. This Gradle Enterprise instance has all features and extensions enabled and is freely available for use by the Apache Solr project and all other Apache projects.

This pull request enhances the functionality of publishing build scans to the publicly available [scans.gradle.com](https://scans.gradle.com/) by instead publishing build scans to [ge.apache.org](https://ge.apache.org/). On this Gradle Enterprise instance, Apache Solr will have access not only to all of the published build scans but other aggregate data features such as:

- Dashboards to view all historical build scans, along with performance trends over time
- Build failure analytics for enhanced investigation and diagnosis of build failures
- Test failure analytics to better understand trends and causes around slow, failing, and flaky tests

Please let me know if there are any questions about the value of Develocity or the changes in this pull request and I’d be happy to address them.

# Checklist

### DIFF
--- a/.github/workflows/bin-solr-test.yml
+++ b/.github/workflows/bin-solr-test.yml
@@ -15,6 +15,9 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    env:
+      GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GE_ACCESS_TOKEN }}
+
     steps:
     # Setup
     - uses: actions/checkout@v2
--- a/.github/workflows/docker-test.yml
+++ b/.github/workflows/docker-test.yml
@@ -20,6 +20,7 @@ jobs:
     env:
       SOLR_DOCKER_IMAGE_REPO: github-pr/solr
       SOLR_DOCKER_IMAGE_TAG: ${{github.event.number}}
+      GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GE_ACCESS_TOKEN }}
 
     steps:
     # Setup
--- a/.github/workflows/gradle-precommit.yml
+++ b/.github/workflows/gradle-precommit.yml
@@ -11,6 +11,9 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    env:
+      GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GE_ACCESS_TOKEN }}
+
     steps:
     # Setup
     - uses: actions/checkout@v2
--- a/.github/workflows/solrj-test.yml
+++ b/.github/workflows/solrj-test.yml
@@ -14,6 +14,9 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    env:
+      GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GE_ACCESS_TOKEN }}
+
     steps:
     # Setup
     - uses: actions/checkout@v2
--- a/gradle/ge.gradle
+++ b/gradle/ge.gradle
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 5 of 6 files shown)


clayburn:
@dsmiley - I do not know the best way to address the Crave workflow. In order for it publish a build scan, it needs to have the access key where the build is running. I'm not sure it is a good idea to send the org-level access key to the Crave hosted runner, and I don't know enough about Crave to know how to do this. But it would be valuable to capture the build from that runner as well. What do you think?

dsmiley:
If the build scan is a file that can be generated from the Gradle build (without access to GE), then this file could be moved back to the GitHub Action runner machine (which does have access to GE) to send it to GE.

clayburn:
@dsmiley - this is a good idea. While we don't have first-class, documented support for moving the build scan data around on disk, I think it _might_ be possible. I will investigate and get back to you.

dsmiley:
I see this enables [the build cache](https://docs.gradle.org/current/userguide/build_cache.html) by default, which hopefully is a big positive benefit, at the expense of some local disk and risks of a misconfigured build leading to accidental caching.  Note that this topic has been raised before: https://issues.apache.org/jira/browse/SOLR-15603 which defaults to false via commented out part of `gradle.properties`.  Solr's build (and Lucene) generates on first build from a [template](https://github.com/apache/solr/blob/main/gradle/template.gradle.properties) which mentions `org.gradle.caching`.  That should probably be removed entirely if we're going to enable build caching in this new `ge.gradle` file.

Even if the Crave integration is lacking; I don't think that should hold this up.

clayburn:
> I see this enables [the build cache](https://docs.gradle.org/current/userguide/build_cache.html) by default, which hopefully is a big positive benefit, at the expense of some local disk and risks of a misconfigured build leading to accidental caching. Note that this topic has been raised before: [issues.apache.org/jira/browse/SOLR-15603](https://issues.apache.org/jira/browse/SOLR-15603) which defaults to false via commented out part of gradle.properties. Solr's build (and Lucene) generates on first build from a [template](https://github.com/apache/solr/blob/main/gradle/template.gradle.properties?rgh-link-date=2023-11-01T21%3A46%3A36Z) which mentions org.gradle.caching. That should probably be removed entirely if we're going to enable build caching in this new ge.gradle file.

This change actually will not. You still need the `org.gradle.caching=true` property for the configuration here to take effect. When that is true, then the configuration in this PR will be applied.

> Even if the Crave integration is lacking; I don't think that should hold this up.

I've done a little investigation on this so far, but without success. It would involve pulling files out of Gradle User Home back to the GitHub Actions runner. The `crave pull` command doesn't seem to let me access $HOME, which makes sense. I still have some avenues to explore, but I would not let that hold up this PR.

clayburn:
> It'd be nice if the target path for the scan could be configurable.

As of now, the storage of the previous scan is really an implementation detail to support the `buildScanPublishPrevious` command. It would be helpful in this case though, and I will bring it up internally to see if it makes sense as a feature.

> What's involved in having tests balance better across runners? Would that make sense in a separate PR?

Essentially, it is [enabling Test Distribution](https://docs.gradle.com/enterprise/test-distribution/#gradle_enabling) and [configuring local executors](https://docs.gradle.com/enterprise/test-distribution/#gradle_local_executors). I would recommend a separate PR though, since the change is logically different, and so we can use the data accumulated from this one to better inform how many executors to configure.

dsmiley:
Cool; I'll merge tomorrow if I don't get any +1 sooner.

janhoy (approved):
Lgtm, exciting!

PR #2044, opened 2023-10-26
Extract duplicate code and move it to CollectionHandlingUtils

renatoh:
I came across some duplicate code which I moved to the utils-class


renatoh:
@risdenk  Hey Kevin, would you mind having a look at this PR? thanks!

solr/core/src/java/org/apache/solr/cloud/OverseerConfigSetMessageHandler.java
```diff
@@ -98,6 +99,8 @@ public OverseerSolrResponse processMessage(ZkNodeProps message, String operation
         log.error("ConfigSet: {} operation: {} failed", configSetName, operation, e);
       }
 
+      addExceptionToNamedList(operation, e, results);
+
       results.add("Operation " + operation + " caused exception:", e);
```
epugh, line 104: Do we still need this after adding the `addExceptionToNamedList` call?

epugh (changes requested):
thansk for chasing this down...  i added two questions...

solr/core/src/java/org/apache/solr/cloud/api/collections/CollectionHandlingUtils.java
```diff
@@ -450,6 +450,26 @@ static void processResponse(
     }
   }
 
+  static void logFailedOperation(final Object operation, final Exception e, final String collName) {
+    if (collName == null) {
+      log.error("Operation {} failed", operation, e);
+    } else {
+      log.error("Collection {}}, operation {} failed", collName, operation, e);
```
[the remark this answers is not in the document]
renatoh: To my defense, this was already within the original code, I extracted it to a method without noticing the issue, good catch!

solr/core/src/java/org/apache/solr/cloud/OverseerConfigSetMessageHandler.java
renatoh: Nope, I had forgotten to remove the original code in this instance. I did remove it

PR #2051, opened 2023-10-31
SOLR-17026: Fix releaseBuild without signing

HoustonPutman:
Tested this out a bit locally with and without signing. Love that we can get rid of the “-x signJarsPublication” requirement for non signed builds.

Need to make sure that the artifacts are still teated to make sure that the signatures exist in the full smoke tester.


risdenk (approved):

HoustonPutman:
Ok I tested the smoke tester with and without `--not-signed` using artifacts built with and without `-Psign -PuseGpg`. The smoke test failed and succeeded accordingly (succeeded with `--not-signed` and without `-Psign`, and without `--not-signed` and with `-Psign`, and failed the other two combinations).

This should be good to go.

PR #2052, opened 2023-10-31
SOLR-16911: Purge last hostContext references

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-16911

# Description

SOLR-16911 removed almost all of the references to a variable called 'hostContext' used by Jetty to set a base path for v1 requests.  However, one reference remained.

# Solution

This PR removes the last removable hostContext reference from 'main': an unused parameter to a MiniSolrCloudCluster method.

# Tests

N/A - small refactor. Existing tests continue to pass.

# Checklist

### DIFF
--- a/solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
+++ b/solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
@@ -447,14 +447,12 @@ public class MiniSolrCloudCluster {
   /**
    * Start a new Solr instance
    *
-   * @param hostContext context path of Solr servers used by Jetty
    * @param extraServlets Extra servlets to be started by Jetty
    * @param extraRequestFilters extra filters to be started by Jetty
    * @return new Solr instance
    */
   public JettySolrRunner startJettySolrRunner(
       String name,
-      String hostContext,
       SortedMap<ServletHolder, String> extraServlets,
       SortedMap<Class<? extends Filter>, String> extraRequestFilters)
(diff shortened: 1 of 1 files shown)


gerlowskija:
FYI @epugh

dsmiley (approved):

epugh (approved):
One question..   Is the "startJettySolrRunner" method actually used?  I expected that if you removed that parameter, then you'd be removing it from the caller classes too?   I wonder if that method is even needed???

gerlowskija:
> Is the "startJettySolrRunner" method actually used? [...] I wonder if that method is even needed???

It's not used currently.  I thought about deleting it, but wasn't sure whether it might be there for the benefit of other projects that consume solr-test-framework.  Happy to remove it though if you think it's not helpful in that context?

gerlowskija:
Alright - I'll remove it.  If the signature was less complex I'd be worried about deleting a method that users might be relying on in their tests, but looking at it, it'd take a pretty advanced user to invoke it.

It does raise an interesting question about whether backcompat even applies to solr-test-framework.  As someone who's used it in 3rd-party projects (e.g. [spark-solr](https://github.com/lucidworks/spark-solr/blob/master/pom.xml#L532)) it is a pain when breaking changes happen.  But at the same time, it probably doesn't require quite as rigid a backcompat-policy as a true client library like SolrJ.

epugh:
Yeah...   I am not looking for more stasis and difficulty evolving Solr by establishing more "back compat" expectations...

epugh:
🚢 it!

PR #2057, opened 2023-11-01
SOLR-17108: Prometheus Exporter: do not throw NullPointerException.

dsmiley:
A minor improvement to avoid throwing an NPE and instead more clearly log an error.

I've seen NPEs when trying to use "queryResponse" if we can't reach Solr, and I've noticed the same NPE reported elsewhere like the dev list.


dsmiley:
Proposed in 9.5 bug fix:

SOLR-17108: Prometheus Exporter: don't throw NPE; gracefully handle no metrics

I'll merge in January.

PR #2058, opened 2023-11-01
SOLR-17063: Do not retain log param references in LogWatcher

magibney:
https://issues.apache.org/jira/browse/SOLR-17063


mkhludnev (approved):
assuming it's covered by existing tests.

PR #1904, opened 2023-09-08
SOLR-16961: Remove deprecated CircuitBreakerManager

janhoy:
https://issues.apache.org/jira/browse/SOLR-16961

Draft PR for eventually removing deprecated code from main. Will not merge until other in-flight CB efforts have landed..

PR #2059, opened 2023-11-01
SOLR-16961 Remove deprecated CircuitBreakerManager in 10.0

janhoy:
https://issues.apache.org/jira/browse/SOLR-16961


cpoerschke (approved):

janhoy:
Thanks for review, will merge tomorrow (friday)

janhoy:
Ok, I forgot that I already had #1904 already open with almost exactly same changes. Well, double work there 🤣 In that PR I chose to keep the Base test-class in order to minimize merge conflicts for future CB backports. But I wonder if git manages to do the right thing, understanding that the `BaseTestCircuitBreaker` class was renamed `TestCircuitBreaker`?

cpoerschke:
> Ok, I forgot that I already had #1904 already open with almost exactly same changes. ...

Ah, I had a vague "wasn't there a prior draft or wasn't this further down in the PR list" sense but didn't look beyond that. Thanks for finding the other PR!

> ... But I wonder if git manages to do the right thing, understanding that the `BaseTestCircuitBreaker` class was renamed `TestCircuitBreaker`?

I _think_ git will manage to do the right thing often in this situation, based on some sort of partial understanding of how git works underneath and the occasional "i'm surprised this worked" reaction in this kind of situation, but I'm no expert.

janhoy:
Yea, I might diff the patches from these two PRs and see if I miss anything on this latest one...

janhoy:
The only real diff is that #1904 had configurable error code, https://github.com/apache/solr/pull/1904/files#diff-e040eb54da15dcb4051d543df5be72f979fbc2f0f62fb217353f1ef90cb6f61dR71-R80

janhoy:
For some reason, #1904 was not added to the JIRA issue by the bot, else I would have seen it immediately when browsing my open JIRS for work to complete. I think we should migrate from JIRA to github issues like Lucene did.

cpoerschke:
> The only real diff is that #1904 had configurable error code ...

Without the configurability seems simpler but given that `setErrorCode` is not annotated as deprecated (or final), need we retain it then or to deprecate it before removal?

janhoy:
> > The only real diff is that #1904 had configurable error code ...
> 
> Without the configurability seems simpler but given that `setErrorCode` is not annotated as deprecated (or final), need we retain it then or to deprecate it before removal?

The [CircuitBreaker class](https://github.com/apache/solr/blob/branch_9x/solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreaker.java) does not have any `setErrorCode()` method, it was added in the other PR.

I added it in this PR too, this time through sysprop `solr.circuitbreaker.errorcode` so users can actually configure it (globally).

solr/core/src/test/org/apache/solr/util/TestCircuitBreakers.java
```diff
@@ -66,26 +76,39 @@ protected static void indexDocs() {
...
+                System.setProperty(
+                    CircuitBreaker.SYSPROP_SOLR_CIRCUITBREAKER_ERRORCODE, String.valueOf(code));
+              }
+              SolrException ex =
+                  expectThrows(
+                      SolrException.class,
+                      () -> {
+                        h.query(req("name:\"john smith\""));
+                      });
+              assertEquals(code == -1 ? 429 : code, ex.code());
```
cpoerschke, line 109: question: when could the code be `-1` based on the list above?
```suggestion
              assertEquals(code == -1 ? SolrException.ErrorCode.TOO_MANY_REQUESTS.code : code, ex.code());
```

solr/core/src/test/org/apache/solr/util/TestCircuitBreakers.java
```diff
@@ -66,26 +76,39 @@ protected static void indexDocs() {
...
-        () -> {
-          h.query(req("name:\"john smith\""));
-        });
+    List.of(
+            SolrException.ErrorCode.TOO_MANY_REQUESTS.code,
+            SolrException.ErrorCode.SERVICE_UNAVAILABLE.code,
+            SolrException.ErrorCode.BAD_REQUEST.code)
+        .forEach(
+            code -> {
+              if (code != SolrException.ErrorCode.TOO_MANY_REQUESTS.code) {
```
cpoerschke, line 99: setting to "too many requests" should also work, right? could we maybe use `null` to indicate do-not-set?
```suggestion
              if (code != null) {
```

solr/core/src/test/org/apache/solr/util/TestCircuitBreakers.java
```diff
@@ -66,26 +76,39 @@ protected static void indexDocs() {
...
 
     h.getCore().getCircuitBreakerRegistry().register(circuitBreaker);
 
-    expectThrows(
-        SolrException.class,
-        () -> {
-          h.query(req("name:\"john smith\""));
-        });
+    List.of(
+            SolrException.ErrorCode.TOO_MANY_REQUESTS.code,
```
cpoerschke, line 94:
```suggestion
            null,
            SolrException.ErrorCode.TOO_MANY_REQUESTS.code,
```

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreaker.java
```diff
@@ -62,6 +64,16 @@ public CircuitBreaker() {}
...
   public abstract String getErrorMessage();
 
+  /**
+   * Get http error code, defaults to {@link SolrException.ErrorCode#TOO_MANY_REQUESTS} but can be
+   * overridden with system property {@link #SYSPROP_SOLR_CIRCUITBREAKER_ERRORCODE}
+   */
+  public static SolrException.ErrorCode getErrorCode() {
+    return SolrException.ErrorCode.getErrorCode(
+        Integer.getInteger(
+            SYSPROP_SOLR_CIRCUITBREAKER_ERRORCODE, SolrException.ErrorCode.TOO_MANY_REQUESTS.code));
```
cpoerschke, line 74: edge case: wondering about noisily failing if the property is specified but invalid e.g. `-Dsolr.circuitbreaker.errorcode=SERVICE_UNAVAILABLE` or some such misunderstanding.

janhoy: You'd get an exception from either the `getInteger` call or the `getErrorCode` call. But only during runtime. We could of course have a static context initialization, but I put init here mainly to be able to write a test...

solr/core/src/test/org/apache/solr/util/TestCircuitBreakers.java
janhoy: Sorry, confusion here. I now went back to using -1 as "default", as `null` is not allowed in the list.

solr/core/src/test/org/apache/solr/util/TestCircuitBreakers.java
```diff
@@ -66,26 +76,39 @@ protected static void indexDocs() {
...
 
-    expectThrows(
-        SolrException.class,
-        () -> {
-          h.query(req("name:\"john smith\""));
-        });
+    List.of(
+            -1,
+            SolrException.ErrorCode.SERVICE_UNAVAILABLE.code,
+            SolrException.ErrorCode.BAD_REQUEST.code)
```
cpoerschke, line 96:
```suggestion
            -1,
            SolrException.ErrorCode.TOO_MANY_REQUESTS.code,
            SolrException.ErrorCode.SERVICE_UNAVAILABLE.code,
            SolrException.ErrorCode.BAD_REQUEST.code)
```

solr/core/src/test/org/apache/solr/util/TestCircuitBreakers.java
cpoerschke:
```suggestion
              assertEquals(code == -1 ? SolrException.ErrorCode.TOO_MANY_REQUESTS.code : code, ex.code());
```

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreaker.java
```diff
@@ -62,6 +64,16 @@ public CircuitBreaker() {}
   /** Get error message when the circuit breaker triggers */
   public abstract String getErrorMessage();
 
+  /**
+   * Get http error code, defaults to {@link SolrException.ErrorCode#TOO_MANY_REQUESTS} but can be
```
cpoerschke, line 68: CI flags this, not sure what the solution is, maybe fully qualified?
```suggestion
   * Get http error code, defaults to {@link org.apache.solr.common.SolrException.ErrorCode#TOO_MANY_REQUESTS} but can be
```

janhoy: No error in IDE, but I re-phrased it without a link ..

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreaker.java
janhoy: So in this case the end user / client would receive HTTP 500 error messages from the exception bubbling up I suppose. I could add a `getErrorCode()` call to the constructor to fail during initialization too?

cpoerschke (approved):

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreaker.java
cpoerschke: Sorry, I missed this comment. The exception bubbling up could be confusing for clients and operationally an exception or at least logging on startup could be helpful?

solr/core/src/test/org/apache/solr/util/TestCircuitBreakers.java
```diff
@@ -66,26 +76,40 @@ protected static void indexDocs() {
...
+                System.setProperty(
+                    CircuitBreaker.SYSPROP_SOLR_CIRCUITBREAKER_ERRORCODE, String.valueOf(code));
+              }
+              SolrException ex =
+                  expectThrows(
+                      SolrException.class,
+                      () -> {
+                        h.query(req("name:\"john smith\""));
+                      });
+              assertEquals(code == -1 ? SolrException.ErrorCode.TOO_MANY_REQUESTS.code : code, ex.code());
```
cpoerschke, line 110: tidy says:
```suggestion
              assertEquals(
                  code == -1 ? SolrException.ErrorCode.TOO_MANY_REQUESTS.code : code, ex.code());
```

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreaker.java
janhoy: So I learnt something: `Integer.getInteger()` will return the default-value also if the system property is set but is not an integer. And `getErrorCode` will return `UNKNOWN` (0) for unsupported codes such as `999`. To fail loudly we must catch both, and I'll push a solution.

janhoy:
Latest changes
* Fail-early for invalid system property values in `solr.circuitbreaker.errorcode`
* Parsing Integer from sysprop is now strict, not falling back to default 429 if not integer
* Add test for validating that invalid sysprop input causes exception during initialization of CB

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreaker.java
```diff
@@ -54,14 +57,48 @@ public void init(NamedList<?> args) {
...
+   */
+  public static SolrException.ErrorCode getExceptionErrorCode() {
+    return errorCode;
+  }
+
+  private static SolrException.ErrorCode resolveExceptionErrorCode() {
+    int intCode = SolrException.ErrorCode.TOO_MANY_REQUESTS.code;
+    if (System.getProperty(SYSPROP_SOLR_CIRCUITBREAKER_ERRORCODE) != null) {
+      try {
+        intCode = Integer.parseInt(System.getProperty(SYSPROP_SOLR_CIRCUITBREAKER_ERRORCODE));
```
cpoerschke, line 83: subjective
```suggestion
    String strCode = System.getProperty(SYSPROP_SOLR_CIRCUITBREAKER_ERRORCODE);
    if (strCode != null) {
      try {
        intCode = Integer.parseInt(strCode);
```

solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreaker.java
```diff
@@ -54,14 +57,48 @@ public void init(NamedList<?> args) {
...
+    SolrException.ErrorCode errorCode = SolrException.ErrorCode.getErrorCode(intCode);
+    if (errorCode != SolrException.ErrorCode.UNKNOWN) {
+      return errorCode;
+    } else {
+      throw new SolrException(
+          SolrException.ErrorCode.SERVER_ERROR,
+          String.format(
+              Locale.ROOT,
+              "Invalid error code %s specified for circuit breaker system property %s.",
+              System.getProperty(SYSPROP_SOLR_CIRCUITBREAKER_ERRORCODE),
```
cpoerschke, line 97:
```suggestion
              strCode,
```

cpoerschke (approved):

cpoerschke (approved):

cpoerschke (approved):

cpoerschke (approved):

PR #2062, opened 2023-11-03
SOLR-17065: Absorb the Cross DC project

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17065

- [x] Absorb the CrossDC Consumer/Producer/Common
- [x] Have the build work
  - [x] Fix the compilation warning and uncomment the lines from `javac.gradle`
- [x] Have the unit tests work
- [x] Integrate with Solr packaging (deduplicating dependency jars and all)
- [ ] Add an integration test
- [ ] Add a docker test?
- [x] Add documentation in the repo
- [x] Add ref guide documentation
- [x] Make sure the smoke tests work
- [x] Rename the configurations so that they make sense with the whole project


HoustonPutman:
We can't upgrade to kafka 3.8.0, because of https://issues.apache.org/jira/browse/KAFKA-17227, but 3.8.1 will be released in early october and we should be good to upgrade to that.

gradle/testing/randomization.gradle
```diff
@@ -199,7 +199,12 @@ allprojects {
 
         // Enable security manager, if requested. We could move the selection of security manager and security policy
         // to each project's build/ configuration but it seems compact enough to keep it here for now.
-        if (Boolean.parseBoolean(testOptionsResolved["tests.useSecurityManager"])) {
+        def useSecurityManager = Boolean.parseBoolean(testOptionsResolved["tests.useSecurityManager"]);
```
dsmiley, line 202: @HoustonPutman, I would prefer we have generalized mechanisms instead of one-off peculiarities, and we have them.  See globals.gradle propertyOrDevant and its friends.  Is that stuff sufficient?  If we need to toggle settings globally, we already have gradle.properties

gradle/validation/validate-log-calls.gradle
```diff
@@ -118,7 +118,7 @@ class ValidateLogCallsTask extends DefaultTask {
 
     if (hasPlus) {
       cause = "hasPlus: " + hasPlus
-      violation = true
+      violation = level != "error"
```
dsmiley, line 121: What's this about?  Could use a comment.

solr/modules/cross-dc/build.gradle
```diff
@@ -0,0 +1,70 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+apply plugin: 'java-library'
+
+description = 'Cross-DC Replication Plugins'
+
+ext {
+  // Use of the security manager is not supported on windows.
```
dsmiley, line 25: Why is this concern specific to cross-dc?  We use Mockito throughout Solr tests.

solr/solrj-zookeeper/build.gradle
```diff
@@ -37,7 +37,7 @@ dependencies {
     implementation 'org.apache.httpcomponents:httpclient'
     implementation 'org.apache.httpcomponents:httpcore'
 
-    implementation('org.apache.zookeeper:zookeeper', {
```
dsmiley, line 40: Why was this done? It suggests any consumer of solrj-zookeeper would likely want to use ZK APIs directly.  But I don't think that's true.

solr/test-framework/src/java/org/apache/solr/SolrTestCaseJ4.java
```diff
@@ -2158,7 +2158,8 @@ public static File getFile(String name) {
       return file;
     }
     throw new RuntimeException(
-        "Cannot find resource in classpath or in file-system (relative to CWD): " + name);
+        "Cannot find resource in classpath or in file-system (relative to CWD): "
```
dsmiley, line 2161: shouldn't the exception have the name as-given.

HoustonPutman: I think it was much easier for me to debug whatever error was happening when I changed to this. I'm not opposed to put it back.

solr/modules/cross-dc/build.gradle
HoustonPutman: It's because of a feature of mockito that spins up new jvms I think?
We spent a ton of time trying to get it to work, to no avail.
Since the security manager is going away anyways, it's not the worst thing.

gradle/testing/randomization.gradle
HoustonPutman: Are you talking about the `testOptionsResolved` thing? If so it's all over this randomization grade file. I have no idea why it's used but I just added the ability for a project to override whatever it is.

gradle/validation/validate-log-calls.gradle
HoustonPutman: Yeah, this was a misunderstanding of the log.error() function, and the fact that the last parameter is treated as an exception and not a parameter to the string format. I can undo this

solr/solrj-zookeeper/build.gradle
HoustonPutman: It doesn't necessarily mean that, it means that the users would likely want to use ZK classes. This is true because in particular because of `Stat`, `KeeperException` and `CreateMode`.

Why I added it in this PR? I'm not really sure.

dsmiley: Wouldn't a user pretty be looking only for only ZkClientClusterStateProvider and even then only indirectly through SolrJ's reflection instantiation of it as a result of passing ZK URLs to CloudSolrClient.Builder?  I understand the number of public classes in basically any module is large so there indeed exists some interfaces of miscellaneous transitive dependencies.  But in practice, I think it's best to be conservative with class path exposure based on typical use (subjective, I know).  They can always add it if they need to.

gradle/testing/randomization.gradle
dsmiley: I glossed over testOptionsResolved but glad you pointed this out now.  This is a better example than those I listed.  Basically, my point is why don't you go enhance randomization.gradle to compute testOptionsResolved with a new ability for a gradle module/project change tweak *whatever* test options instead of making useSecurityManager a special case.  I'm actually surprised, given all the flexibility here, that it's not _already_ possible.  I think it is actually... notice it calls propertyOrDefault which in turn consults project properties.

PR #2069, opened 2023-11-08
SOLR-16974: Documentation and changes entry for global CB

janhoy:
https://issues.apache.org/jira/browse/SOLR-16974


janhoy:
@cpoerschke I was a bit quick merging the main PR. So here is a follow-up with changes entry and refguide docs. If you find other things missing as well, let's include it here. I'm awaiting the 9x backport #2070 until this is done.

janhoy:
This is how the ref guide page will look like in the guide:
[Circuit Breakers __ Apache Solr Reference Guide.pdf](https://github.com/apache/solr/files/13306062/Circuit.Breakers.__.Apache.Solr.Reference.Guide.pdf)

solr/bin/solr
```diff
@@ -1426,8 +1426,6 @@ if [ -n "${SOLR_CIRCUITBREAKER_QUERY_LOADAVG:-}" ]; then
   SOLR_OPTS+=("-Dsolr.circuitbreaker.query.loadavg=$SOLR_CIRCUITBREAKER_QUERY_LOADAVG")
 fi
 
-echo "SOLR_OPTS is now: ${SOLR_OPTS[*]}"
-
```
cpoerschke, line 1430: Good catch about this one. Any thoughts about changes in `solr.cmd` similar to the `solr` changes at lines 1409-1427 above?

solr/solr-ref-guide/modules/deployment-guide/pages/circuit-breakers.adoc
```diff
@@ -32,8 +32,25 @@ Setting the `shards.tolerant=true` parameter on requests can help with graceful
 circuit breaker thresholds are reached on some nodes. See the <<shards.tolerant Parameter>> for details.
 
 == Circuit Breaker Configurations
-All circuit breaker configurations are listed as independent `<circuitBreaker>` entries in `solrconfig.xml` as shown below.
-A circuit breaker can register itself to trip for query requests and/or update requests. By default only search requests are affected. A user may register multiple circuit breakers of the same type with different thresholds for each request type.
+Circuit breakers can be configured globally for the entire node, or for each collection individually, or a combination. Per-collection circit breakers take precedence over global circuit breakers.
```
cpoerschke, line 35:
```suggestion
Circuit breakers can be configured globally for the entire node, or for each collection individually, or a combination. Per-collection circuit breakers are checked before global circuit breakers.
```

Both per-collection and global CBs are checked, right? Hmm, not quite sure how to best convey that though.

solr/solr-ref-guide/modules/deployment-guide/pages/circuit-breakers.adoc
```diff
@@ -32,8 +32,25 @@ Setting the `shards.tolerant=true` parameter on requests can help with graceful
...
+
+A circuit breaker can register itself to trip for query requests and/or update requests. A user may register multiple circuit breakers of the same type with different thresholds for each request type.
```
cpoerschke, line 37:
```suggestion
A circuit breaker can register itself to be checked for query requests and/or update requests. A user may register multiple circuit breakers of the same type with different thresholds for each request type.
```

or

```suggestion
A circuit breaker can register itself for query requests and/or update requests. A user may register multiple circuit breakers of the same type with different thresholds for each request type.
```

solr/solr-ref-guide/modules/deployment-guide/pages/circuit-breakers.adoc
```diff
@@ -52,13 +69,20 @@ The main configuration for this circuit breaker is controlling the threshold per
...
 ----
 <circuitBreaker class="org.apache.solr.util.circuitbreaker.MemoryCircuitBreaker">
  <double name="threshold">75</double>
 </circuitBreaker>
 ----
 
+.Global in `solr.in.sh`
+[source,bash]
+----
+SOLR_CIRCUITBREAKER_QUERY_HEAP=75
```
cpoerschke, line 83:
```suggestion
SOLR_CIRCUITBREAKER_QUERY_MEM=75
```

solr/bin/solr
janhoy: Yea, good catch. It would be covered by #1935 but unless that is merged first, we'll definitely need it...

solr/solr-ref-guide/modules/deployment-guide/pages/circuit-breakers.adoc
janhoy: What I meant with "take precedence" is that if you have a QUERY CB for CPU configured both globally and for collection1, then it is the one from collection1 that will be used / override the global one.

janhoy: I committed another re-word of the paragraph: "Per-collection circuit breakers are checked before global circuit breakers, and if there is a conflict, the per-collection circuit breaker takes precedence."

cpoerschke: Hmm, perhaps I'm misunderstanding the code then ...

... so `getCombinedMap` handles the scenario where both per-collection and global are present by concatenating the lists and the per-collection ones are placed before the global ones: https://github.com/apache/solr/blob/e740123546c7f20c6728df42d493e11f93dc8c79/solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerRegistry.java#L267-L279

... `checkTripped` checks all the circuit breakers of the request type in question but if a tripping one is found it still continues to check the remaining ones because they could have tripped too: https://github.com/apache/solr/blob/e740123546c7f20c6728df42d493e11f93dc8c79/solr/core/src/java/org/apache/solr/util/circuitbreaker/CircuitBreakerRegistry.java#L168-L181

So my understanding then is that there is ordering in the checks but no precedence or conflict as such because both per-collection and global circuit breakers are always checked?

cpoerschke: Thanks for the example, that helps with thinking through the scenarios :-)

> ... e.g. QUERY MEM 90% globally and QUERY MEM 70% for the collection, then a request to that collection will trip at 70%, not 95, since the "local" value takes precedence. ...

That makes me wonder actually, should we consider what happens if someone (mis)configures

> ... e.g. QUERY MEM 70% globally and QUERY MEM 90% for the collection ...

i.e. in that case it will still trip at 70% and the 90% threshold is either a mistake or misunderstanding, or possibly some sort of transitional addition before the global threshold is removed or something.

Or put another way, it's always the lowest threshold that takes precedence?

janhoy: > Or put another way, it's always the lowest threshold that takes precedence?

I think most users will choose either configuring globally or per core. I really only see utility (from a protect server from over-heating perspective) in the global config. But if a user chooses to do both, my thinking was that they should be allowed to do it in any fashion they like, which includes shooting themselves in the foot by allowing higher or lower values for a particular collection than the global. There may be reasons for both, and I don't think we should dictate what they can do.

The only environment where I can see it would be beneficial to prevent collections from overriding would be in multi tenant environments where you as cluster owner want to maintain full control. But IMO such requirements should come from those kind of users and it can be provided as a future config option to either disable collection-level CB entirely or fine grained precedence rules of some kind. I feel it is premature to guess those needs in this first feature introduction, so keeping it simple.

cpoerschke: How about keeping the existing logic and illustrating in the docs what a global/local combination would typically look like e.g.

```
- ... Per-collection circuit breakers are checked before global circuit breakers, and if there is a conflict, the per-collection circuit breaker takes precedence. ...
- ... Per-collection circuit breakers are checked before global circuit breakers, and if there is a conflict, the per-collection circuit breaker takes precedence. Typically, any per-collection circuit breaker thresholds are set lower than global thresholds. ...
```

janhoy: Done

cpoerschke (approved):

cpoerschke:
Hmm, not sure what's up with the precommit failure. Looks unrelated and unable to reproduce locally.

janhoy:
Yea, reported it in https://github.com/apache/solr/pull/2086#issuecomment-1823635256

PR #2072, opened 2023-11-09
SOLR-16397: Modify v2 'renamecore' API to be more REST-ful

iamsanjay:

# Description

V2(JAX-RS) Endpoints for core admin op "RENAME".

# Solution

This commit changes the v2 "renamecore" API to be more in line with the REST-ful design we're targeting for Solr's v2 APIs.

Following these changes, the v2 API now appears as:
  `POST /api/cores/coreName/rename {...}` 

Although not shown above, the 'rename' command specifier has been removed from the request body. This commit also converts the API to the new JAX-RS framework.

# Tests

Already few test cases are present which are modified as per the new changes.

# Checklist


solr/core/src/java/org/apache/solr/handler/admin/CoreAdminOperation.java
```diff
@@ -172,12 +175,14 @@ public enum CoreAdminOperation implements CoreAdminOp {
       RENAME,
       it -> {
         SolrParams params = it.req.getParams();
-        String name = params.required().get(CoreAdminParams.OTHER);
-        String cname = params.required().get(CoreAdminParams.CORE);
```
iamsanjay, line 176: As we are moving our code from CoreAdminOperation to API. I think it would be great to remove this validation from here as we are already testing required parameter using `ensureRequiredParameterProvided` method in APIs.

solr/core/src/java/org/apache/solr/handler/admin/CoreAdminOperation.java
```diff
@@ -172,12 +175,14 @@ public enum CoreAdminOperation implements CoreAdminOp {
...
-
-        if (cname.equals(name)) return;
-
-        it.handler.coreContainer.rename(cname, name);
+        final String cname = params.get(CoreAdminParams.CORE);
+        final var renameCoreRequestBody = new RenameCoreRequestBody();
+        renameCoreRequestBody.to = params.get((CoreAdminParams.OTHER));
+        RenameCoreApi renameCoreApi =
+            new RenameCore(
+                it.handler.coreContainer, it.handler.getCoreAdminAsyncTracker(), it.req, it.rsp);
```
iamsanjay, line 183: Should I use `var` in place for `RenameCoreApi`?

solr/core/src/test/org/apache/solr/handler/admin/CoreAdminOperationTest.java
```diff
@@ -167,23 +167,29 @@ public void testRenameUnexpectedFailuresResultIn500SolrException() {
 
   @Test
   public void testRenameMissingCoreParamResultsIn400SolrException() {
+    final var coreAdminHandler = mock(CoreAdminHandler.class);
+    final var callInfo1 = new CoreAdminHandler.CallInfo(coreAdminHandler, mockRequest, null, null);
+
```
iamsanjay, line 172: Earlier the code exists at the start If all the required parameters are not passed. 
Now that validation is removed from here and CoreAdminOperation relies on RenameCore class for it,  test started failing as passed mock does not contain the Handler object. That's why I created new mock and passed it. Also Let me know If you want to change the variable name.

gerlowskija: Thanks for calling these out.

> That's why I created new mock and passed it

The mock makes sense IMO 👍 

You likely already know this, but I'll mention it as an FYI, since it's a common gotcha in writing Solr tests: any tests that use mocks need to have a `assumeWorkingMockito` call in a `@BeforeClass` method.  It looks like we've already got in this class, so no need to change anything here!  Just wanted to put the FYI out there.

> Let me know If you want to change the variable name.

Haha, 'callInfo1' isn't the best name, maybe.  I probably would've chosen something like `renameCallInfo` or 'callInfoWithMockHandler`.  (That latter option makes it clear why we're not using the 'callInfo' like all the other tests.). But the test methods are small and clear, so it's not a big deal, ultimately.  So it's up to you.

solr/solr-ref-guide/modules/configuration-guide/pages/coreadmin-api.adoc
```diff
@@ -349,9 +349,19 @@ This parameter is required in v1, and part of the url in the v2 API.
 
 The `RENAME` action changes the name of a Solr core.
 
+[.dynamic-tabs]
```
gerlowskija, line 352: [+1] Thanks as always for your attention to the docs!

But....

[-1] I definitely see what you were going for in creating a different "Parameter" section for each API version.  It's tough otherwise to be clear about which parameter names are used in v1 vs. v2.  But in handling similar cases for other APIs, we've found a pattern that does this a bit more concisely without (I think) any loss of clarity.

Essentially what we've been doing is just to put the v1 and v2 parameter names in the same "list item" separated by a comma, and with the API version for each in parenthesis.  See the shards/shardNames parameter entry [here](https://solr.apache.org/guide/solr/latest/deployment-guide/collection-management.html#create-parameters) for an example.

Would you mind merging the "Parameters" sections here to be closer to that example?  (It's a bit of a hard thing to describe.  If I wasn't clear enough or you're not quite sure what I'm getting at, lmk I can push up a commit to show what I mean.)

gerlowskija (changes requested):
Hi @iamsanjay - thanks again for taking on another of these API changes!

The large majority here looks great, but I suggested a doc change we might want to make before merging and a few other nitpicks.

Let me know if you disagree or are unsure what I meant at any point, and thanks again!

gerlowskija:
Thanks again for the PR @iamsanjay !  Really starting to crank through these now - it's awesome!

PR #2076, opened 2023-11-15
SOLR-17076: Optimize `OrderedNodePlacementPlugin#getAllReplicasOnNode`

patsonluk:
https://issues.apache.org/jira/browse/SOLR-17076

# Description

`OrderedNodePlacementPlugin#getAllReplicasOnNode` can be slow in a cluster with lot of replicas. The effect could compound with new collection creation with many shards. 

# Solution

Introduced a new field `allReplicas` which keeps track of all the replicas added/removed in the Plugin. Return a copy of that set instead for `getAllReplicasOnNode`.

Also added a new convenient method `getAllReplicaCount` if only the count is needed

# Tests

A minor change, rely on existing unit test case

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cluster/placement/plugins/OrderedNodePlacementPlugin.java
+++ b/solr/core/src/java/org/apache/solr/cluster/placement/plugins/OrderedNodePlacementPlugin.java
@@ -514,6 +514,7 @@ public abstract class OrderedNodePlacementPlugin implements PlacementPlugin {
                 (shard, reps) -> {
                   if (reps.remove(replica)) {
                     hasReplica.set(true);
+                    allReplicas.remove(replica);
                   }
                   return reps.isEmpty() ? null : reps;
                 });
@@ -521,7 +522,6 @@ public abstract class OrderedNodePlacementPlugin implements PlacementPlugin {
           });
       if (hasReplica.get()) {
         removeProjectedReplicaWeights(replica);
-        allReplicas.remove(replica);
       }
     }
 
--- a/solr/core/src/java/org/apache/solr/cluster/placement/plugins/OrderedNodePlacementPlugin.java
+++ b/solr/core/src/java/org/apache/solr/cluster/placement/plugins/OrderedNodePlacementPlugin.java
@@ -403,7 +403,9 @@ public abstract class OrderedNodePlacementPlugin implements PlacementPlugin {
   public abstract static class WeightedNode implements Comparable<WeightedNode> {
     private final Node node;
     private final Map<String, Map<String, Set<Replica>>> replicas;
-    private final Set<Replica> allReplicas; //a flattened list of all replicas, computing from the map could be costly
+
+    // a flattened list of all replicas, as computing from the map could be costly
+    private final Set<Replica> allReplicas;
 
     public WeightedNode(Node node) {
       this.node = node;
--- a/solr/core/src/java/org/apache/solr/cluster/placement/plugins/OrderedNodePlacementPlugin.java
+++ b/solr/core/src/java/org/apache/solr/cluster/placement/plugins/OrderedNodePlacementPlugin.java
@@ -403,10 +403,12 @@ public abstract class OrderedNodePlacementPlugin implements PlacementPlugin {
   public abstract static class WeightedNode implements Comparable<WeightedNode> {
     private final Node node;
     private final Map<String, Map<String, Set<Replica>>> replicas;
+    private final Set<Replica> allReplicas; //a flattened list of all replicas, computing from the map could be costly
 
     public WeightedNode(Node node) {
       this.node = node;
       this.replicas = new HashMap<>();
+      this.allReplicas = new HashSet<>();
     }
 
     public Node getNode() {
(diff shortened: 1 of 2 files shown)


patsonluk:
```
  - org.apache.solr.cloud.BasicDistributedZk2Test.test (:solr:core)
    Test output: /Users/patson/src/cowpath-solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.cloud.BasicDistributedZk2Test.txt
    Reproduce with: gradlew :solr:core:test --tests "org.apache.solr.cloud.BasicDistributedZk2Test.test" -Ptests.jvms=5 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=5A95C1706F927E37 -Ptests.file.encoding=ISO-8859-1
```

solr/core/src/java/org/apache/solr/cluster/placement/plugins/OrderedNodePlacementPlugin.java
```diff
@@ -515,6 +521,7 @@ public final void removeReplica(Replica replica) {
           });
       if (hasReplica.get()) {
         removeProjectedReplicaWeights(replica);
+        allReplicas.remove(replica);
```
magibney, line 524: Now that we're separately tracking replicas in a top-level map (and assuming that this is all accessed by a single thread), I think this method could be simplified to:
```java
if (allReplicas.remove(replica)) {
  // [the nested removal logic]
}
```

solr/core/src/java/org/apache/solr/cluster/placement/plugins/OrderedNodePlacementPlugin.java
```diff
@@ -404,20 +404,25 @@ public abstract static class WeightedNode implements Comparable<WeightedNode> {
...
     public Node getNode() {
       return node;
     }
 
     public Set<Replica> getAllReplicasOnNode() {
-      return replicas.values().stream()
-          .flatMap(shard -> shard.values().stream())
-          .flatMap(Collection::stream)
-          .collect(Collectors.toSet());
+      return new HashSet<>(allReplicas);
```
magibney, line 421: based on how this is being used currently, there's no need to wrap this in a new HashSet. The method is public, so you never know who might call it and what their assumptions might be; but you are replacing `Collectors.toSet()`, which specifically makes no guarantees about mutability/immutability, so I'd be inclined to just do `Collections.unmodifableSet(allReplicas)`, taking this opportunity to make the API more conservative, without breaking any backcompat.

Also, if the motivating concern is "many replicas", the cost of instantiating a new HashMap that we don't really need is not worth it.

magibney (changes requested):
Suggested a couple of changes, but overall this looks good to me!

solr/core/src/java/org/apache/solr/cluster/placement/plugins/OrderedNodePlacementPlugin.java
patsonluk: Good catch @magibney ! I'm wondering if these 2 statements should just placed in https://github.com/apache/solr/blob/d8902468b8b292cb37ee3e1ba8d1a724752ab9a8/solr/core/src/java/org/apache/solr/cluster/placement/plugins/OrderedNodePlacementPlugin.java#L516 directly?

Perhaps I have overlooked some edge cases which using the AtomicBoolean hasReplica was necessary ?

solr/core/src/java/org/apache/solr/cluster/placement/plugins/OrderedNodePlacementPlugin.java
patsonluk: As discussed, the current fix is to use the most defensive approach which does not modify the behavior at all. The minor concern with `unmodifableSet` (which I did use at firs) was that the "view" could still mutate after this method returns and this could break certain expectation of the caller.

Otherwise I don't have any objection using `unmodifableSet`. Perhaps @HoustonPutman can share some thoughts? 😊

solr/core/src/java/org/apache/solr/cluster/placement/plugins/OrderedNodePlacementPlugin.java
patsonluk: After i read the code again, I'm not 100% sure about the `hasReplica` flag. 

However, it does make the logic a bit more isolated and easier to follow - iterate the list first, figure out if the replica exists, if it does, deal with it AFTEr finishing the traversal. In a logical level, it's pretty clean and avoid any concurrent modification exception (though the map/list being iterated on is private field anyway).

Hm...I simply move the allReplicas.remove to within the `if (reps.remove(replica))` block now, cause i do think it's the best if the operations (add/remove) on the 2 collections stay close to each other.

Anyway, I think these are very minor concerns? 😊 either ways are fine

dsmiley:
Just a little reminder... when squash-merging, remember to edit the commit message so it's cleaned up to reflect the final result in totality; it's almost always necessary.  Should not refer to running "tidy", for example.

PR #2078, opened 2023-11-15
SOLR-10653: leader to replica updates fail on UUIDField

mkhludnev:
https://issues.apache.org/jira/browse/SOLR-10653

reproducer for `Error while creating field 'uuid{type=uuid,properties=indexed,stored,omitNorms,omitTermFreqAndPositions,useDocValuesAsStored,uninvertible}' from value 'java.util.UUID:ded2baf5-d8f5-497e-a4bc-2f9085e09798'`

sure, uuid is grabbed by javabin codec. But I assume we are not going to create a separate tag for UUID in javabin wire format. Or we are? So, I tend to blame UUID.toObject(), it seems to me it just should be a string, or there's some compatibility problem.

The worst this, I see is that remote (replica's) update error is not propagated to client! I'm shocked. 

 It's just a reproducer, that's why it's a draft.

### DIFF
--- a/solr/core/src/java/org/apache/solr/schema/FieldType.java
+++ b/solr/core/src/java/org/apache/solr/schema/FieldType.java
@@ -301,23 +301,18 @@ public abstract class FieldType extends FieldProperties {
     String val;
     try {
       val = toInternal(value.toString());
-    } catch (RuntimeException e) {
+    }
+    catch (SolrException se) {
+      throw se; // fall through BAD_REQUEST TODO test
+    }
+    catch (RuntimeException e) {
       throw new SolrException(
-          SolrException.ErrorCode.SERVER_ERROR, // TODO Seems wrong. It's BAD_REQUEST!
+          SolrException.ErrorCode.SERVER_ERROR,
           "Error while creating field '" + field + "' from value '" + value + "'",
           e);
     }
     if (val == null) return null;
 
-    /*org.apache.lucene.document.FieldType newType = new org.apache.lucene.document.FieldType();
-    newType.setTokenized(field.isTokenized());
-    newType.setStored(field.stored());
-    newType.setOmitNorms(field.omitNorms());
-    newType.setIndexOptions(field.indexed() ? getIndexOptions(field, val) : IndexOptions.NONE);
-    newType.setStoreTermVectors(field.storeTermVector());
-    newType.setStoreTermVectorOffsets(field.storeTermOffsets());
-    newType.setStoreTermVectorPositions(field.storeTermPositions());
-    newType.setStoreTermVectorPayloads(field.storeTermPayloads());*/
     return createField(field.getName(), val, field);
   }
 
--- a/solr/core/src/java/org/apache/solr/schema/FieldType.java
+++ b/solr/core/src/java/org/apache/solr/schema/FieldType.java
@@ -301,11 +301,9 @@ public abstract class FieldType extends FieldProperties {
     String val;
     try {
       val = toInternal(value.toString());
-    }
-    catch (SolrException se) {
-      throw se; // fall through BAD_REQUEST TODO test
-    }
-    catch (RuntimeException e) {
+    } catch (SolrException se) {
+      throw se; //  BAD_REQUEST to fall through
+    } catch (RuntimeException e) {
       throw new SolrException(
           SolrException.ErrorCode.SERVER_ERROR,
           "Error while creating field '" + field + "' from value '" + value + "'",
--- a/solr/core/src/java/org/apache/solr/schema/FieldType.java
+++ b/solr/core/src/java/org/apache/solr/schema/FieldType.java
@@ -303,7 +303,7 @@ public abstract class FieldType extends FieldProperties {
       val = toInternal(value.toString());
     } catch (RuntimeException e) {
       throw new SolrException(
-          SolrException.ErrorCode.SERVER_ERROR,
+          SolrException.ErrorCode.SERVER_ERROR, // TODO Seems wrong. It's BAD_REQUEST!
           "Error while creating field '" + field + "' from value '" + value + "'",
           e);
     }
--- a/solr/core/src/java/org/apache/solr/schema/UUIDField.java
+++ b/solr/core/src/java/org/apache/solr/schema/UUIDField.java
@@ -92,21 +92,4 @@ public class UUIDField extends StrField {
   public String toInternal(UUID uuid) {
     return uuid.toString().toLowerCase(Locale.ROOT);
   }
-
-  @Override
-  public String toObject(IndexableField f) {
-    // it fails reading from stored field
-    // UUID.fromString(toExternal(f));
-    return toExternal(f);
-  }
-
-  @Override
-  public Object toNativeType(Object val) {
-    // it fails reading from translog (perhaps) ??
-    //
-    //if (val instanceof CharSequence) {
-    //  return UUID.fromString(val.toString());
-    //}
-    return val;
-  }
 }
--- a/solr/core/src/java/org/apache/solr/schema/UUIDField.java
+++ b/solr/core/src/java/org/apache/solr/schema/UUIDField.java
@@ -94,15 +94,19 @@ public class UUIDField extends StrField {
   }
 
   @Override
-  public UUID toObject(IndexableField f) {
-    return UUID.fromString(toExternal(f));
+  public String toObject(IndexableField f) {
+    // it fails reading from stored field
+    // UUID.fromString(toExternal(f));
+    return toExternal(f);
   }
 
   @Override
   public Object toNativeType(Object val) {
-    if (val instanceof CharSequence) {
-      return UUID.fromString(val.toString());
-    }
+    // it fails reading from translog (perhaps) ??
+    //
+    //if (val instanceof CharSequence) {
+    //  return UUID.fromString(val.toString());
+    //}
     return val;
   }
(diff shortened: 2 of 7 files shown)


mkhludnev:
UUIDField throws `BAD_REQUEST` but in log I see  `o.a.s.s.HttpSolrCall 500 Exception`

mkhludnev:
It fails when we have leader and replica, when committed document is updated, leader updates a replica
`Remote error message: ERROR: [doc=1] Error adding field 'uuid'='java.util.UUID:fb4e3bbc-8fc6-46e4-8f65-f7e0cc9989dc' `
Then leader goes to recover replica 
`RecoveringCoreTermWatcher Start recovery on core_node4 because core's term is less than leader's term`
`o.a.s.u.DefaultSolrCoreState Running recovery`
Then update request complete with success. 
but selecting from replica (leader:false) fails to see this update. 
I think it's a severe error when failed replica update, falls back to shard recovery, fails recovery, but return success to client. Why?

solr/core/src/test/org/apache/solr/update/UUIDAtomicUpdateTest.java
```diff
@@ -0,0 +1,153 @@
...
+import org.apache.solr.common.SolrDocument;
+import org.apache.solr.common.SolrInputDocument;
+import org.apache.solr.common.params.ModifiableSolrParams;
+import org.apache.solr.common.util.NamedList;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class UUIDAtomicUpdateTest extends SolrCloudTestCase {
```
dsmiley, line 22: Our naming policy would be Uuid, even for acronyms.  (Yes; much code before now was done in ignorance of the policy or predated it).

dsmiley (commented):
Glad you could reproduce this!

mkhludnev:
I removed both of UUID instantiation cases since javabin have no type tag for UUID. 
It can be introduced in a separate ticket for sake of performance.

solr/core/src/java/org/apache/solr/schema/UUIDField.java
```diff
@@ -92,17 +92,4 @@ public String toInternal(String val) {
   public String toInternal(UUID uuid) {
     return uuid.toString().toLowerCase(Locale.ROOT);
   }
-
```
mkhludnev, line 95: javabin can't transfer UUID, no sense to create it :(

solr/core/src/java/org/apache/solr/schema/FieldType.java
```diff
@@ -301,6 +301,8 @@ public IndexableField createField(SchemaField field, Object value) {
     String val;
     try {
       val = toInternal(value.toString());
+    } catch (SolrException se) {
```
mkhludnev, line 304: it's worth to keep for other types as well.

solr/core/src/test/org/apache/solr/update/UuidAtomicUpdateTest.java
```diff
@@ -0,0 +1,155 @@
...
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class UuidAtomicUpdateTest extends SolrCloudTestCase {
+  private static final String COMMITTED_DOC_ID = "1";
+  private static final String UNCOMMITTED_DOC_ID = "2";
+  private static final String COLLECTION = "collection1";
+  private static final int NUM_SHARDS = 1;
+  private static final int NUM_REPLICAS = 2;
```
mkhludnev, line 27: replica is crucial.

solr/modules/extraction/src/test/org/apache/solr/handler/extraction/TestXLSXResponseWriter.java
```diff
@@ -38,8 +38,10 @@
 import org.apache.solr.search.SolrReturnFields;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
 
+@Ignore("Crave tests fails on headless jvm")
```
mkhludnev, line 44: I'll remove this `@Ignore` before commit.

dsmiley: Maybe you don't recall but we use the string "nocommit" to convey that.  `gradlew precommit` will fail so we won't forget.

solr/core/src/test/org/apache/solr/update/UuidAtomicUpdateTest.java
```diff
@@ -0,0 +1,167 @@
...
+  public void testUpdateUncommittedUuidField() throws Exception {
+    atomicSetValue(UNCOMMITTED_DOC_ID, "uuid", uncommittedUuidAfter);
+    ensureFieldHasValues(UNCOMMITTED_DOC_ID, "uuid", uncommittedUuidAfter);
+    commit();
+    ensureFieldHasValues(UNCOMMITTED_DOC_ID, "uuid", uncommittedUuidAfter);
+  }
+
+  private static void commit() throws IOException, SolrServerException {
+    final UpdateRequest committedRequest = new UpdateRequest();
+    committedRequest.commit(cluster.getSolrClient(), COLLECTION);
```
dsmiley, line 111:
```suggestion
    cluster.getSolrClient().commit(collection);
```

solr/core/src/test/org/apache/solr/update/UuidAtomicUpdateTest.java
```diff
@@ -0,0 +1,167 @@
...
+    final UpdateRequest committedRequest = new UpdateRequest();
+    committedRequest.deleteByQuery("*:*");
+    committedRequest.commit(cluster.getSolrClient(), COLLECTION);
+  }
+
+  private static void atomicSetValue(String docId, String fieldName, Object value)
+      throws Exception {
+    final SolrInputDocument doc = new SolrInputDocument();
+    doc.setField("id", docId);
+    Map<String, Object> atomicUpdateRemoval = new HashMap<>(1);
```
dsmiley, line 125: minor: use Map.of

solr/core/src/test/org/apache/solr/update/UuidAtomicUpdateTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+  @Override
+  @Before
+  public void setUp() throws Exception {
+    super.setUp();
+    String committedUuidBefore = UUID.randomUUID().toString();
+    final SolrInputDocument committedDoc =
+        sdoc("id", COMMITTED_DOC_ID, "title_s", "title_1", "uuid", committedUuidBefore);
+    final UpdateRequest committedRequest = new UpdateRequest().add(committedDoc);
+    committedRequest.commit(cluster.getSolrClient(), COLLECTION);
```
dsmiley, line 68: Minor: in general, I see you creating UpdateRequest and saving to a named variable and then later calling commit or process.  But you can chain the call; no need to declare a variable and come up with a name or use "final".

solr/core/src/test/org/apache/solr/update/UuidAtomicUpdateTest.java
```diff
@@ -0,0 +1,167 @@
...
+    ensureFieldHasValues(UNCOMMITTED_DOC_ID, "uuid", uncommittedUuidAfter);
+  }
+
+  private static void commit() throws IOException, SolrServerException {
+    final UpdateRequest committedRequest = new UpdateRequest();
+    committedRequest.commit(cluster.getSolrClient(), COLLECTION);
+  }
+
+  @AfterClass
+  public static void afterAll() throws Exception {
```
dsmiley, line 115: Why bother?  The test is about to be shut down.

solr/core/src/test/org/apache/solr/update/UuidAtomicUpdateTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+    CollectionAdminRequest.createCollection(COLLECTION, "conf", NUM_SHARDS, NUM_REPLICAS)
+        .process(cluster.getSolrClient());
+
+    cluster.waitForActiveCollection(COLLECTION, NUM_SHARDS, NUM_REPLICAS * NUM_SHARDS);
+  }
+
+  @Override
+  @Before
+  public void setUp() throws Exception {
```
dsmiley, line 62: no clear of existing data first?  It's hard to reason how this might not be necessary given the particularly high sensitivity this test has to documents that are added but not committed.  I'd explain this in a comment if there's a good reason.  Otherwise, why not just clean all to be sure we know what the state is.

dsmiley (commented):
Mostly just minor comments; nice job on the fix & test!  I wonder if there are other field types that could fall prey to this risk?  I could imagine (eventually) making this test repeated to test other field types.  No need to do this now!

solr/core/src/test/org/apache/solr/update/UuidAtomicUpdateTest.java
```diff
@@ -0,0 +1,167 @@
...
+    super.setUp();
+    String committedUuidBefore = UUID.randomUUID().toString();
```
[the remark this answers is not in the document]
mkhludnev: statics are named `...After` to detect change between _before_ and _after_.

mkhludnev:
Last call! Going, going..

PR #2084, opened 2023-11-17
SOLR-17077: Don't cancel twice when rejoining shard leader election

psalagnac:
https://issues.apache.org/jira/browse/SOLR-17077

# Description

When we invoke `REJOINELECTION` command on a replica, the replica must first leave its current election position by deleting the sequential ephemeral node in Zookeeper. This is done by invoking `cancelElection()` method on the election context.

This issue is `cancelElection()` method is invoked twice. Once in `ZkController` and once in `retryElection()` in the leader elector. Consequently we try to delete the Zookeeper leader election node twice. The second attempt always fails since the node was already removed by the first attempt.

# Solution

Remove the call to `cancelElection()` from ZkController. Also remove the creation of a new election context here, since a new one will be created in `retryElection()`.

Second commit on the pull request is to remove useless parameters `shardId` and `electionNode` from the `REJOINELECTION` command, that are not used anymore.

We have this change in production for 6 months and spotted no issue.

# Tests

No test added.
I made sure all leader election tests still pass.

# Checklist

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/request/beans/RejoinLeaderElectionPayload.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/request/beans/RejoinLeaderElectionPayload.java
@@ -26,13 +26,9 @@ public class RejoinLeaderElectionPayload implements ReflectMapWriter {
   // ones are meant to be required without that being specified on the v1 API or elsewhere
   @JsonProperty public String collection;
 
-  @JsonProperty public String shard;
-
   @JsonProperty public String coreNodeName;
 
   @JsonProperty public String core;
-
-  @JsonProperty public String electionNode;
 
   @JsonProperty public Boolean rejoinAtHead;
 }
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -20,6 +20,8 @@ Improvements
 
 * SOLR-16536: Replace OpenTracing instrumentation with OpenTelemetry (Alex Deparvu, janhoy)
 
+* SOLR-17077: When a shard rejoins leader election, leave previous election only once to save unneeded calls to Zookeeper. (Pierre Salagnac)
+
 Optimizations
 ---------------------
 (No changes)
--- a/solr/core/src/java/org/apache/solr/cloud/ElectionContext.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ElectionContext.java
@@ -72,6 +72,12 @@ public abstract class ElectionContext implements Closeable {
 
   public void joinedElectionFired() {}
 
+  /**
+   * Create a new instance when retrying the election.
+   *
+   * <p>This carries over all election parameters, but not current status (mostly the {@link
+   * SyncStrategy} is not copied for shared leader election.
+   */
   public ElectionContext copy() {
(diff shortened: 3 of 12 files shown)


solr/solrj/src/java/org/apache/solr/client/solrj/request/beans/RejoinLeaderElectionPayload.java
```diff
@@ -26,13 +26,9 @@ public class RejoinLeaderElectionPayload implements ReflectMapWriter {
   // ones are meant to be required without that being specified on the v1 API or elsewhere
   @JsonProperty public String collection;
 
-  @JsonProperty public String shard;
```
dsmiley, line 29: Is this backwards-compatible? I'm not super familiar with this, but from what I can see, there are some inputs that are no longer needed.  It seems they weren't really used before (yes?).  What would happen if a client supplies it?

CC @gerlowskija you might know if this produces a failure or not

dsmiley (commented):
An additional side benefit is a bit less code :-)

solr/solrj/src/java/org/apache/solr/client/solrj/request/beans/RejoinLeaderElectionPayload.java
psalagnac: Thanks for pointing this @dsmiley.
I did a test locally and it seems to me any additional property in the Json payload here is simply ignored. Since this change is only removing some properties, I don't think we can have backwards-compatibility issue.

dsmiley:
@psalagnac can you add a CHANGES.txt entry please?  In "Other" or "Improvements" depending on your point of view.  Try to both be brief and communicate the scope/impact -- kind of hard sometimes ;-)

psalagnac:
Thanks.
I added an entry in "Improvements", but not sure whether it should go to bug "Bug Fixes". Maybe the code should have been like the from the beginning.

dsmiley (approved):

dsmiley:
By entering this CHANGES.txt into the 10.0 release section, it means you believe it should not be back ported to 9x.  Is that what you meant or was that accidental?  I'll move it and backport if you wish.

psalagnac:
Yes, just main is fine.
I want to follow-up on this work with other cleanups in leader election code. I don't think it is worth to backport them to 9x.

mtchorz:
@psalagnac @dsmiley are there any chances that the changes could be back ported to next 9.x version? We are facing similar issues with solr 9.5

psalagnac:
I don't see any reason this could not be ported.
@dsmiley I'm not sure of the process here. Is a new Jira required?

epugh:
> I don't see any reason this could not be ported. @dsmiley I'm not sure of the process here. Is a new Jira required?

If a backport isn't a major effort then generally we don't have new JIRA's.  You can just backport it (and update the CHANGES.txt!)

mtchorz:
Are there any chances that this will fix will be released as a part of next, 9.7 release? Or we will need to wait for next one?

dsmiley:
It's a minor bug.  Do you think it's has worse consequences than as-reported?
Any way, it's not going to make 9.7 (will be released probably Monday) but I'll backport after your response so it makes whatever release happens next.

PR #2089, opened 2023-11-23
clarify .github/workflows display name: gradle precommit--->check w/ Java 11

cpoerschke:
To make it clearer what the same command for running locally is.


dsmiley:
On my mind here is wondering why we aren't simply calling `precommit` here?  Did we once have this but changed our minds (why?)

janhoy (approved):

PR #2099, opened 2023-11-27
SOLR-15484 JWTAuthPluginIntegrationTest - dynamically generate the self-signed certificate using the localhost loopback address

jdyer1:
https://issues.apache.org/jira/browse/SOLR-15484

# Description

# Solution

# Tests

# Checklist


jdyer1:
This is an attempt to help the frequent failures with JWTAuthPluginIntegrationTest.  I see a past comment on SOLR-15484 suggested generating the certificate based on the loopback address rather than using the static file.  Here is an attempt a this.  I have no idea if this is a real help;  I do not know how to set up my system so that the test will fail without the fix.

solr/modules/jwt-auth/src/test/org/apache/solr/security/jwt/JWTAuthPluginIntegrationTest.java
```diff
@@ -116,6 +122,12 @@ public static void afterClass() throws Exception {
     if (mockOAuth2Server != null) {
       mockOAuth2Server.shutdown();
     }
+    if (tempDir != null) {
+      Files.walk(tempDir)
+              .sorted(Comparator.reverseOrder())
+              .map(Path::toFile)
+              .forEach(java.io.File::delete);
```
janhoy, line 129: I think you can replace all this with `tempDir.toFile().deleteOnExit()`?

janhoy (approved):
Thanks for giving this some love, I've spent many hours wrestling this.
Perhaps the only way to know if we can un-bad-apple this test is to merge it to main and let it bake?

solr/modules/jwt-auth/src/test/org/apache/solr/security/jwt/JWTAuthPluginIntegrationTest.java
```diff
@@ -25,16 +25,14 @@
...
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.KeyStore;
-import java.util.Base64;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
epugh, line 35: Check your code editor settings, I think we spell out all the imports ;-)

solr/modules/jwt-auth/src/test/org/apache/solr/security/jwt/KeystoreGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+    private static final String PASS_PHRASE = "secret";
+
+    public void generateKeystore(Path existingKeystore, Path newKeystore, String cn) {
+        KeyStore ks = null;
+        try(FileInputStream fis = new FileInputStream(existingKeystore.toFile())) {
+            ks = KeyStore.getInstance(KeyStore.getDefaultType());
+            ks.load(fis, PASS_PHRASE.toCharArray());
+            ks.setCertificateEntry(cn, selfSignCertificate(cn));
+
+        } catch (KeyStoreException |CertificateException |NoSuchAlgorithmException | IOException e) {
```
epugh, line 57: white space formatting?   Does the `gradlew tidy` fix up these thigns?

epugh (changes requested):
two small formatting things.   Very cool improvement ;-)

jdyer1:
@epugh  Thank you for your review.  I expected IntelliJ to automatically use the project's formatting preferences but it does not seem to be the case.  But `gradlew tidy` did the trick.

@janhoy I wish I knew how to set up an environment that would return other than "localhost" but yes, perhaps just hope for the best is all we can get in this situation.  I do not think I can use "deleteOnExit" because the directory is not empty.  I did change the cleanup not to use a forbidden api though!

epugh (approved):

solr/modules/jwt-auth/src/test/org/apache/solr/security/jwt/JWTAuthPluginIntegrationTest.java
janhoy: Have you tried? I think it will clean up even non-empty dirs

janhoy:
Regarding the `usedUndeclaredArtifacts` you can add a special entry in grade file for it, look in dev-docs/ for details.

janhoy:
The other test failure looks like some Java security manager thing;

```
  2> 3803 INFO  (SUITE-JWTAuthPluginIntegrationTest-seed#[F234B11BA9F49CDF]-worker) [n: c: s: r: x: t:] o.a.s.u.ErrorLogMuter Creating ErrorLogMuter-regex-2 for ERROR logs matching regex: ignore_exception
   >     java.lang.InternalError: cannot create instance of org.bouncycastle.jcajce.provider.digest.GOST3411$Mappings : java.security.AccessControlException: access denied ("java.security.SecurityPermission" "putProviderProperty.BC")
```
There is a security polity file for tests, perhaps it needs some extra entry for the BC stuff?

solr/modules/jwt-auth/src/test/org/apache/solr/security/jwt/JWTAuthPluginIntegrationTest.java
janhoy: I pushed the simplification using `deleteOnExit`. We use it lots of other places. If there is indeed some issue with it, feel free to revert.

janhoy:
I'm tempted to merge this to main, still as BadAlle, and see whether [failure rate](http://fucit.org/solr-jenkins-reports/failure-report.html) comes down from 42% during the coming days. I'm uncertain what your finding about `localhost.localdomain` actually means.

solr/modules/jwt-auth/src/test/org/apache/solr/security/jwt/KeystoreGenerator.java
```diff
@@ -0,0 +1,116 @@
...
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
+import org.bouncycastle.operator.ContentSigner;
+import org.bouncycastle.operator.OperatorCreationException;
+import org.bouncycastle.operator.jcajce.JcaContentSignerBuilder;
+
+public class KeystoreGenerator {
+
+  private static final String PASS_PHRASE = "secret";
+
+  public void generateKeystore(Path existingKeystore, Path newKeystore, String cn) {
```
janhoy, line 58: @jdyer1 We may be able to simplify this PR by using mock-oauth2-server's [new SSL capability,](https://github.com/navikt/mock-oauth2-server#enabling-https) which includes ssl cert auto creation. So if we find a way to detect the correct local hostname to use in the cert, we can initialize the mock-server with this code:

```java
    String myHostname = "localhost";
    Ssl selfSignedSslLocalhost =
        new Ssl(new SslKeystore("", SslKeystore.Companion.generate(myHostname, "")));
    OAuth2Config config =
        new OAuth2Config(
            false,
            null,
            new OAuth2TokenProvider(),
            Collections.emptySet(),
            new MockWebServerWrapper(selfSignedSslLocalhost));
    MockOAuth2Server mockOAuth2Server = new MockOAuth2Server(config);
```

jdyer1:
@janhoy  With my last update, I am cautiously optimistic this PR may help us with the failure rate.  With this change, the test passes even if I change `localhost` to `localhost.localdomain` in my `/etc/hosts` file.  (I am trying this as a modest attempt to mimic the environmental conditions when it fails in our CI.)  However, carefully consider the changes I am making to `JWTIssuerConfig.java`.  Here I am assuming the host name check bypass for `localhost` exists merely to help unit tests with self-signed certificates.  Perhaps we are also helping users who try this functionality on their laptops?  I think with my change it will work for either case, but maybe we also want to exempt `InetAddress.getLoopbackAddress().getHostName()` as well (as far as I can tell this will always return `localhost`).

It would be much better if we could use the Mock library's self-signed certificate generation, as you suggest.  This would greatly streamline this test and keep the focus were it belongs!  Unfortunately with the short amount of time I spent I was unable to make it work.

janhoy:
Yes, disabling the name check will of course work for the localhost case. Normally the plugin will always connect to an external IDP to fetch JWKs or well-known, and then it will enable validation of cert CN. A pity we won't be able to test that here, but it is far better with a working test that tests everything else, than having full coverage.

Not sure of the consequences of exempting canonicalHostName and not hostName. Perhaps safest to check both?

Wrt simplifying the generation, we can leave that as a follow up, perhaps there is a need for a PR upstream first.

janhoy:
+1, feel free to merge when you wish

PR #2101, opened 2023-11-28
SOLR-17060 CoreContainer#create may deadlock with concurrent requests for metrics

stillalex:
https://issues.apache.org/jira/browse/SOLR-17060

# Description

Posting a proposal for SOLR-17060, I am not entirely sure this will work correctly.
The approach is to identify the problematic window (core still in init phase) and avoid enabling the metrics. based on this idea I extended the existing `isClosed()` (`refCount.get() <= 0`) check to verify that the core is opened instead (`refCount.get() <= 1`).

The test provided is a best-effort, I could not capture the deadlock, but I did get some exception which may indicate some error in opening the searcher but I am not familiar enough with this code to tell for sure.

Open to feedback on both fix and test - even if to say that this approach is not good in identifying and fixing the deadlock.

# Solution

# Tests

```
./gradlew beast -p solr/core --tests SolrCoreTest.testCoreInitDeadlockMetrics -Ptests.dups=200
```

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -64,7 +64,6 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
-import java.util.function.BooleanSupplier;
 import java.util.function.Consumer;
 import java.util.stream.StreamSupport;
 import org.apache.commons.io.file.PathUtils;
@@ -1343,17 +1342,8 @@ public class SolrCore implements SolrInfoBean, Closeable {
         true,
         "size",
         Category.INDEX.toString());
-
-    // using only isClosed() might result in deadlock due to metrics trying to eagerly open a
-    // new searcher during core init phase
-    BooleanSupplier isNotOpen =
-        () -> {
-          boolean isPreSearcherInit = newSearcherCounter.getCount() == 0;
-          return isPreSearcherInit || isClosed();
-        };
-
     parentContext.gauge(
-        () -> isNotOpen.getAsBoolean() ? parentContext.nullNumber() : getSegmentCount(),
+        () -> isReady() ? getSegmentCount() : parentContext.nullNumber(),
         true,
         "segments",
         Category.INDEX.toString());
@@ -1907,6 +1897,18 @@ public class SolrCore implements SolrInfoBean, Closeable {
   /** Whether this core is closed. */
   public boolean isClosed() {
     return refCount.get() <= 0;
+  }
+
+  /**
+   * Whether this core is ready. Sometimes using {@link #isClosed()} might result in a deadlock due
+   * to the caller trying to eagerly open a new searcher via {@link #withSearcher(IOFunction)}
+   * during core init phase.
+   *
+   * <p>Like for example concurrently polling metrics collection via the registered gauges in {@link
+   * #initializeMetrics(SolrMetricsContext, String)}
+   */
+  public boolean isReady() {
+    return !isClosed() && newSearcherCounter.getCount() > 0;
   }
 
   private Collection<CloseHook> closeHooks = null;
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -64,6 +64,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.BooleanSupplier;
 import java.util.function.Consumer;
 import java.util.stream.StreamSupport;
 import org.apache.commons.io.file.PathUtils;
@@ -1332,18 +1333,29 @@ public class SolrCore implements SolrInfoBean, Closeable {
         true,
         "indexDir",
         Category.CORE.toString());
+
+    BooleanSupplier isNotOpen =
+        () -> {
+          // using isClosed() is insufficient and might result in deadlock due to metrics trying to
+          // open a new searcher during core init phase (when refCount == 1)
+          return getOpenCount() <= 1;
+        };
+
     parentContext.gauge(
-        () -> isClosed() ? parentContext.nullNumber() : getIndexSize(),
+        () -> isNotOpen.getAsBoolean() ? parentContext.nullNumber() : getIndexSize(),
         true,
         "sizeInBytes",
         Category.INDEX.toString());
     parentContext.gauge(
-        () -> isClosed() ? parentContext.nullNumber() : getSegmentCount(),
+        () -> isNotOpen.getAsBoolean() ? parentContext.nullNumber() : getSegmentCount(),
         true,
         "segments",
         Category.INDEX.toString());
     parentContext.gauge(
-        () -> isClosed() ? parentContext.nullString() : NumberUtils.readableSize(getIndexSize()),
+        () ->
+            isNotOpen.getAsBoolean()
+                ? parentContext.nullString()
+                : NumberUtils.readableSize(getIndexSize()),
         true,
         "size",
         Category.INDEX.toString());
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -64,7 +64,6 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
-import java.util.function.BooleanSupplier;
 import java.util.function.Consumer;
 import java.util.stream.StreamSupport;
 import org.apache.commons.io.file.PathUtils;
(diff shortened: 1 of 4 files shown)


dsmiley (commented):
Who is calling SolrCore.initializeMetrics?  It's hard to tell; it's an override of a base abstraction so is being called generically somewhere.  I'm wondering if we can delay initializeMetrics until the SolrCore constructor finishes.

ahubold:
> Who is calling SolrCore.initializeMetrics? It's hard to tell; it's an override of a base abstraction so is being called generically somewhere. I'm wondering if we can delay initializeMetrics until the SolrCore constructor finishes.

It's called directly from the constructor at https://github.com/apache/solr/blob/0056e82a30d6239af4a80726527ca912a5ca3d31/solr/core/src/java/org/apache/solr/core/SolrCore.java#L1113

I was also thinking about whether that could be moved further down. But I don't know much about this code, and if there's a reason for that call at that time.

stillalex:
> I was also thinking about whether that could be moved further down.

I don't think that will help. there are some async things happening so simply moving the code would not provide a stronger guarantee.

> I'm wondering if we can delay initializeMetrics until the SolrCore constructor finishes.

delaying the metrics was what I was going for. but would we need to delay all metrics or just the problematic ones? there is a 'latch' object that hints to provide some safety against events firing too early but I could not hook into this code due to complexity. https://github.com/apache/solr/pull/2101/files#diff-3bf44f923643744a743460bb0f64301618dd7529a602d945fe2f7f193d4cdde0R1166-R1168

dsmiley:
Thanks for clarifying Andreas.  I propose initializing metrics at the very end of the constructor, which is when we can say the SolrCore is reasonably constructed.

> simply moving the code would not provide a stronger guarantee.

Guarantee of what?  Maybe you mean guarantee that there isn't some indirect call via SolrCore that isn't immediately obvious but registers metrics that might need a Searcher?  Okay.

> delaying the metrics was what I was going for. but would we need to delay all metrics or just the problematic ones?

The calls to initializeMetrics() both on SolrCore and on solrFieldCacheBean.

The CountDownLatch "latch" in the constructor is interesting; notice it's closed at the very end.  "latch" was a poor name; ambiguous (it's like calling an int "integer").  We might generalize that slightly to a field that conveys the state of the SolrCore being constructed.  Call it "constructedLatch".  Then gate the metrics gauges on that & (and !closed) instead of refCount. I don't understand how `refCount <= 1` is a solution; there can be many simultaneous references (say 10) and we should expose values then.

dsmiley:
BTW I'd be happy to throw up a proposal PR with my ideas, taking your test :-)

stillalex:
sure @dsmiley go for it

stillalex:
> I don't understand how refCount <= 1 is a solution; there can be many simultaneous references (say 10) and we should expose values then.

read the PR again, you may be parsing it in reverse (I opted for keeping the 'isClosed' style to minimize changes and get the discussion going). this replaces 'isClosed' with a better state check something that reflect 'isOpen'.

dsmiley:
I played a bit with my idea but it wasn't working out; perhaps because internal complexities around searcher lifecycle.

I still don't understand why `getOpenCount() <= 1;` could be correct because a metrics request isn't going to increment the core count.  Maybe we should change the initial value to 0 and increment at the end of the constructor; albeit it will then appear as "closed" during this time?  A value of "-1" could be used to differentiate.  I went to a test, MetricsHandlerTest.test and set a breakpoint to inspect its response, which is only partially validated.  There are no validations for, say, "segments" (exposed as INDEX.segments), but if you look at the response in a debugger, there is a null value.  It should instead be the number of segments.  Obviously we have a testing gap here.

stillalex:
> I played a bit with my idea but it wasn't working out; perhaps because internal complexities around searcher lifecycle.

I hear you. as an aside if (at least) 3 people are trying to make sense of this code and can't - this is a pretty strong signal it needs some proper refactoring.

> I still don't understand why getOpenCount() <= 1; could be correct because a metrics request isn't going to increment the core count. Maybe we should change the initial value to 0 and increment at the end of the constructor; albeit it will then appear as "closed" during this time? A value of "-1" could be used to differentiate.

I used this as an indirect indicator that the init part is completed. metrics do not increase open count, they only expose some lazy computations which can trigger the deadlocking flow. as far as I can read this code the 'open count > 1' (strictly bigger) signals the 'open' method has been called, which is post-init phase. I find the 'open count = 1 by default' confusing too but I avoided messing around too much with it because that was not the core issue being addressed.

the way I understand it:
 - init phase -- open count = 1
 - "open" / "close" calls will increase and decrease open count
 - final 'close' will set open count to 0 and actually perform resource cleanup (doClose)

If the open count is not to your liking (I am not married to it, it was just a proposal) I think you had some thoughts on alternatives. it's just a matter of getting the timing right for the concurrent calls to the metrics, where simply moving the code below a few lines is not sufficient.

> There are no validations for, say, "segments" (exposed as INDEX.segments), but if you look at the response in a debugger, there is a null value. It should instead be the number of segments. Obviously we have a testing gap here.

I think some nulls can creep in if you call the metrics before they are made available in the metrics context. otherwise I agree there might be gaps here, and we can look at improving existing tests.

stillalex:
the failing test was caused by my change, so I had to reduce the open count check to this metric only. I had extended it eagerly to all 3 metrics for some idea of consistency, but that didn't play well with the PlacementPluginIntegrationTest. I think there is some underlying expectation in the placement plugin test to have access to the metrics even if the core is not actively open and my change made it so there are no metrics available which broke the test.

dsmiley:
I do like the refCount==-1 during initialization idea I came up with but am unsure what you think.  I think this would address the matter and be clear to reason about.

stillalex:
> I do like the refCount==-1 during initialization idea I came up with but am unsure what you think. I think this would address the matter and be clear to reason about.

thanks for clarifying, I am good with this idea! good to see consensus on approaching this from the refCount side. I will try to give this a go in the following days, unless you are already working on a different PR.

stillalex:
took a swing at it and could not get it working based on refCount only. I don't think it can be done without major changes (happy to be shown otherwise). the trouble I had was: the init code checks for isClosed before opening a new searcher so in order for init to complete isClosed needs to return true, but same isClosed method for the same init window needs to return false so that the metrics do not become active.
so instead of going and changing the hot path for creating a new searcher I went back to my PR and poked at the condition again and came up with another idea: what if we use an existing metric (newSearcherCounter) to track if any number of searchers have been created since the init phase. anything > 0 means init has happened so it is safe to activate the metrics. this also has the benefit that (unlike the refCount) does not decrease.

dsmiley (commented):
I like it.  But I do think we need a new test to show that "segments" can return a non-null value.  I think it should work now but I observed it didn't work on your first try.  Basically, a new assertion; maybe a few lines to the existing metrics test.

I think it's hard to tell when we can simply look at `isClosed()` vs `isNotOpen()`.  Maybe let's have a new method on the core like `isReady()`?

stillalex:
> But I do think we need a new test to show that "segments" can return a non-null value. I think it should work now but I observed it didn't work on your first try. Basically, a new assertion; maybe a few lines to the existing metrics test.

agreed. I added a few trivial assertions to the test.

> I think it's hard to tell when we can simply look at isClosed() vs isNotOpen(). Maybe let's have a new method on the core like isReady()?

introduced `isReady()` and a best effort explanation as javadocs. please take a look and let me know if it can be explained better :)

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -1899,6 +1899,18 @@ public boolean isClosed() {
     return refCount.get() <= 0;
   }
 
+  /**
+   * Whether this core is ready. Sometimes using {@link #isClosed()} might result in a deadlock due
+   * to the caller trying to eagerly open a new searcher via {@link #withSearcher(IOFunction)}
+   * during core init phase.
+   *
+   * <p>Like for example concurrently polling metrics collection via the registered gauges in {@link
+   * #initializeMetrics(SolrMetricsContext, String)}
```
dsmiley, line 1908: Personally, I think these docs are overly prescriptive based on our first need for this.  Really, the docs can be as simple as: "The core is ready for use.  It is not initializing or closing/closed.

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -1899,6 +1899,18 @@ public boolean isClosed() {
...
+   */
+  public boolean isReady() {
+    return !isClosed() && newSearcherCounter.getCount() > 0;
```
dsmiley, line 1911: also check that newSearcherCounter isn't null

dsmiley (commented):
Getting real close here; let's add a CHANGES.txt comment as well.  Let's get this into 9.4.1.

I wonder if "withSearcher" should add an assert (or actual check) for `isReady()`?

solr/core/src/java/org/apache/solr/core/SolrCore.java
stillalex: good catch

stillalex:
finally had some time to look at the failures. turns out the counter idea was not the best because metrics can be disabled (seems the case for the failing replication tests where the counter is `org.apache.solr.metrics.MetricSuppliers$NoOpCounterSupplier$NoOpCounter@53f7aac1`) so the `isReady` check will fail.
I am running out of ideas, so I introduced a dedicated variable to track if the init is completed.

dsmiley (approved):

PR #2103, opened 2023-11-29
SOLR-16952: add DenseVectorField to DocsStreamer.KNOWN_TYPES

cpoerschke:
https://issues.apache.org/jira/browse/SOLR-16952

The demo script was started from the awesome https://github.com/apache/solr/blob/main/solr/example/films/README.md content and is for illustrative purposes only and not to be committed.

### How is `DocsStreamer.KNOWN_TYPES` used?

https://github.com/apache/solr/blob/releases/solr/9.4.0/solr/core/src/java/org/apache/solr/response/DocsStreamer.java#L201-L205 it determines if a number of a string value is returned:
* number via https://github.com/apache/solr/blob/releases/solr/9.4.0/solr/core/src/java/org/apache/solr/schema/DenseVectorField.java#L301-L317 or https://github.com/apache/solr/blob/releases/solr/9.4.0/solr/core/src/java/org/apache/solr/schema/FloatPointField.java#L93-L109
* string via https://github.com/apache/solr/blob/releases/solr/9.4.0/solr/core/src/java/org/apache/solr/schema/FieldType.java#L374-L384

### How is `DocsStreamer.getValue` called?

* https://github.com/apache/solr/blob/releases/solr/9.4.0/solr/core/src/java/org/apache/solr/response/DocsStreamer.java#L185
* https://github.com/apache/solr/blob/releases/solr/9.4.0/solr/core/src/java/org/apache/solr/response/BinaryResponseWriter.java#L120
* https://github.com/apache/solr/blob/releases/solr/9.4.0/solr/core/src/java/org/apache/solr/response/BinaryResponseWriter.java#L85


solr/core/src/test/org/apache/solr/response/DocsStreamerTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+public class DocsStreamerTest extends SolrTestCaseJ4 {
+
+  @BeforeClass
+  public static void beforeClass() throws Exception {
+    System.setProperty("enable.update.log", "false"); // schema12 doesn't support _version_
+    initCore("solrconfig.xml", "schema12.xml");
+  }
+
+  public void testDenseVectorField() throws Exception {
```
cpoerschke, line 34: #2107 to first add this test with existing behaviour -- converting this PR to draft until that is merged.

solr/core/src/java/org/apache/solr/response/DocsStreamer.java
```diff
@@ -221,6 +222,7 @@ public static Object getValue(SchemaField sf, IndexableField f) {
     KNOWN_TYPES.add(LongPointField.class);
     KNOWN_TYPES.add(DoublePointField.class);
     KNOWN_TYPES.add(FloatPointField.class);
+    KNOWN_TYPES.add(DenseVectorField.class); // DenseVectorField extends FloatPointField
```
epugh, line 225: A nit, but do we need this comment?   Why is it important to know that it extends FloatPointField?

solr/core/src/test/org/apache/solr/response/DocsStreamerTest.java
epugh: I love the test!

solr/CHANGES.txt
```diff
@@ -136,6 +136,8 @@ Bug Fixes
 * SOLR-17093: Collection restore API command now returns "requestid" when executed asynchronously like other APIs
   (Tomás Fernández Löbbe)
 
+* SOLR-16952: In distributed i.e. multi-shard cloud mode returned dense vector 'fl' fields are now a list of numbers instead of strings. (Qais Qadri, Christine Poerschke)
```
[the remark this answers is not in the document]
cpoerschke: My understanding is that this does not impact the search -- in fact one can do a non-vector search and request the vector field in the `fl` -- the search happens on the shards and the merging is based on scores. The type in the response being non-numeric would impact on clients using that part of the response _and_ from re-reading the https://lists.apache.org/thread/xnfrxvm0p8yknrrszcqn0dd0vfg3wtvj user mailing list thread is appears that it _might_ also impact debug/explain functionality, which from vague memory would run on the federating node rather than on the shard nodes.

Also tagging @alessandrobenedetti here for awareness and I'll also reply on the mailing list thread.

PR #2111, opened 2023-12-01
SOLR-17093: Collection restore API returns requestid when executed asynchronously

tflobbe:
Most APIs return a "requestid" field with the request id when executed asynchronously. [SolrJ even relies on this field to return](https://github.com/apache/solr/blob/7eceece76771bdcd71245bcdfd5c0ae99ec980ee/solr/solrj/src/java/org/apache/solr/client/solrj/request/CollectionAdminRequest.java#L218), but now this is returning null for the restore API.
This API adds this field to the response when present.

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/admin/api/RestoreCollectionAPI.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/api/RestoreCollectionAPI.java
@@ -144,10 +144,14 @@ public class RestoreCollectionAPI extends BackupAPIBase {
       throw remoteResponse.getException();
     }
 
+    if (requestBody.async != null) {
+      response.requestId = requestBody.async;
+      return response;
+    }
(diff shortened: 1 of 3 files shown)


epugh (approved):
LGTM.

solr/core/src/java/org/apache/solr/handler/admin/api/RestoreCollectionAPI.java
```diff
@@ -144,10 +144,14 @@ public SubResponseAccumulatingJerseyResponse restoreCollection(
       throw remoteResponse.getException();
     }
 
+    if (requestBody.async != null) {
+      response.requestId = requestBody.async;
+      return response;
```
mariemat, line 149: I am curious to know if there is any reason to return here and skip the 2 lines below?

PR #2112, opened 2023-12-02
GitHub: tests-via-crave.yml should use self-hosted runner

dsmiley:
We now have self-hosted runners by Crave.io.  Only Crave based GHA actions should use it.  More improvements to come later.

### DIFF
--- a/.github/workflows/tests-via-crave.yml
+++ b/.github/workflows/tests-via-crave.yml
@@ -9,7 +9,7 @@ jobs:
   test:
     name: Run Solr Tests using Crave.io resources
 
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
 
     steps:
     - uses: actions/checkout@v3


dsmiley:
@uvatbc , this PR used the new self-hosted runner!
https://foss.crave.io/app/#/build/info/74162?team=5
Some observations:
* The whole check took 12m.  The underlying build (Gradle command) took 5m45s.  The build was in queue 3m39s.  Most of the rest of the time was doing a git checkout.  
* I see "User Machine Info:MacBook-Air-M2" in Crave... I thought we changed that.  Oh well.

It'd be nice to enhance this PR to end the checkout step given that Crave already has the checkout.

mkhludnev (approved):

dsmiley:
Finally; we're in much better shape.  The build took 8m and the vast majority of that was all in our build (Gradle etc.) which is under our control to improve further; not doing "git clone" or whatever.

PR #2118, opened 2023-12-05
SOLR-16986: log cpuTime in addition to QTime

gbellaton:
https://issues.apache.org/jira/browse/SOLR-16986

# Description

This change logs cpuTime and localCpuTime in addition to QTime.
While QTime is the user time spend for a given request (difference between the query processing start time and end time),
cpuTime is the actual time spent by the cpu processing the request.
localCpuTime is the time spent by each local distributed request or by a top non distributed queries.
cpuTime is only computed for distributed queries and is the sum of all localCpuTime spent by all distributed queries.
It is expected to have sum(localCpuTimes=cpuTime

# Solution

CpuTime is measured using java.lang.management.ThreadMXBean in RequestHandler and aggregated in SearchHandler 
for distributed queries.
# Tests

No tests added
# Checklist


solr/core/src/java/org/apache/solr/util/ThreadStats.java
```diff
@@ -0,0 +1,94 @@
...
+  public static final String SHARDS_CPU_TIME = "cpuTime";
+  public static final String LOCAL_CPU_TIME = "localCpuTime";
+  public static final String ENABLE_CPU_TIME = "solr.enableCpuTime";
+
+  private static ThreadMXBean THREAD_MX_BEAN;
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  static {
+    try {
+      java.lang.management.ThreadMXBean threadBean = ManagementFactory.getThreadMXBean();
```
dsmiley, line 45: Why the full class name here?

dsmiley (commented):
No test?  You might look at TestDistributedSearch for this.

I would like the test to show that the response looks the same no matter distributed search or not.  This is how Solr works in general; the caller shouldn't have to know/care.  The distinction of "cpuTime" vs "localCpuTime" is maybe not needed / too much info -- can we simply have "cpuTime" and document how it's calculated?

solr/core/src/java/org/apache/solr/util/ThreadStats.java
gbellaton: No good reason. I will fix this.

gbellaton:
I think that the distinction between cpuTime and localCpuTime is useful in the logs.
It allows to generate statistics from the logs that make the difference between distributed queries and shard level queries.
However you are right saying that this is internal matter and should not appear to the caller.
What would you think about keeping the logs as currently done but changing the response to always and only use cpuTime ?

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -212,6 +216,14 @@ public abstract void handleRequestBody(SolrQueryRequest req, SolrQueryResponse r
 
   @Override
   public void handleRequest(SolrQueryRequest req, SolrQueryResponse rsp) {
+    ThreadStats cpuStats = null;
+    if (publishCpuTime == null) {
+      publishCpuTime = Boolean.getBoolean(ThreadStats.ENABLE_CPU_TIME);
+    }
```
dsmiley, line 222: Can't this be added to the field's declaration instead of here?  And would then be a primitive `boolean`

dsmiley: If it's then protected, could be accessed by SearchHandler, removing the field there.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -599,6 +602,28 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
+                            .getResponse()
+                            .get(SolrQueryResponse.RESPONSE_HEADER_KEY);
+                if (header != null) {
+                  Long shardCpuTime = (Long) header.get(ThreadStats.CPU_TIME);
+                  if (shardCpuTime != null) {
+                    totalShardCpuTime += shardCpuTime;
+                  }
+                }
+              }
+            }
```
dsmiley, line 626: could you extract out a method for this to try to keep handleRequestBody less long?
Gate by "publishCpuTime" as well?

solr/core/src/java/org/apache/solr/util/ThreadStats.java
dsmiley: hasn't changed

solr/core/src/test/org/apache/solr/CpuTimeTestCase.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.solr.client.solrj.response.QueryResponse;
+import org.apache.solr.cloud.SolrCloudTestCase;
+import org.apache.solr.common.SolrInputDocument;
+import org.apache.solr.common.util.NamedList;
+import org.apache.solr.util.ThreadStats;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+/** Test that cpuTime is generated a */
+public class CpuTimeTestCase extends SolrCloudTestCase {
```
dsmiley, line 33: We should avoid SolrCloudTestCase if the feature in question isn't actually related to SolrCloud.  Distributed search works without it and was implemented before SolrCloud.  A SolrCloud test will always be more heavyweight.

Consider looking at SolrJettyTestRule, a new test mechanism to manage a Solr/Jetty instance with test lifecycle.

solr/core/src/test/org/apache/solr/TestCpuTimeSearch.java
```diff
@@ -0,0 +1,140 @@
...
+  @AfterClass
+  public static void cleanup() throws Exception {
+    if (null != collection1) {
+      collection1.close();
+      collection1 = null;
+    }
+    if (null != collection2) {
+      collection2.close();
+      collection2 = null;
+    }
```
dsmiley, line 88: the test rule will auto-close these things; you can simply set to null.  Even that could be done in a test infrastructure test rule with reflection but we don't have one that does that (would be cool!)

solr/core/src/test/org/apache/solr/TestCpuTimeSearch.java
```diff
@@ -0,0 +1,140 @@
...
+    int size = results.size();
+    assertEquals("should have 2 results", 2, size);
+    NamedList<Object> header = response.getHeader();
+    List<Object> localCpuTimes = header.getAll(ThreadStats.LOCAL_CPU_TIME);
+    List<Object> cpuTimes = header.getAll(ThreadStats.CPU_TIME);
+    assertEquals("localCpuTime should not have values", 0, localCpuTimes.size());
+    assertEquals("cpuTime should only have one value", 1, cpuTimes.size());
+    long cpuTime = (long) cpuTimes.iterator().next();
+    assertTrue("cpuTime (" + cpuTime + ") should be positive", cpuTime >= 0);
+  }
```
dsmiley, line 139: we could also assert that cpuTimes >= sum(localCpuTimes)

dsmiley (commented):
Overall, nice work!

solr/core/src/test/org/apache/solr/TestCpuTimeSearch.java
gbellaton: only cpuTimes is returned so it is not possible to make this check from the client time.
We decided earlier that it was better not to let the client knows if the query was distributed

solr/core/src/java/org/apache/solr/util/ThreadStats.java
```diff
@@ -0,0 +1,94 @@
...
+ * {@link java.lang.management.ThreadMXBean}.
+ *
+ * <p>Calling code should create an instance of this class when starting the operation, and then can
+ * get the {@link #getCpuTimeMs()} at any time thereafter.
+ */
+public class ThreadStats {
+  private static final long UNSUPPORTED = -1;
+  public static final String CPU_TIME = "cpuTime";
+  public static final String LOCAL_CPU_TIME = "localCpuTime";
+  public static final String ENABLE_CPU_TIME = "solr.enableCpuTime";
```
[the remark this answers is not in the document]
dsmiley: @janhoy or "solr.metrics.cputime.enabled" ? Or "logcputime" instead of "cputime".  Monitoring is a broad topic, and what's computed here is more like a metric.  On the other hand, this addition is not part of the metrics API but I suppose it's still reasonable to use that category.
https://cwiki.apache.org/confluence/display/SOLR/System+property+naming+structure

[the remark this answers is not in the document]
janhoy: Yea, not a fan of using "metric" if it is not related to the metric feature. But if this property really changes what is logged to Solr's log, why not use `solr.log.cputime`?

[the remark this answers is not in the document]
dsmiley: +1 to `solr.log.cputime`

dsmiley (approved):
I think this is ready!  Oh, just need a CHANGES.txt entry, probably under "New Features" and refer to the system property name.  Proposed entry text:

> SOLR-16986: Log CPU time per request via solr.log.cputime (Gilles Bellaton, David Smiley)

gbellaton:
Should I add the New Feature in the 9.5.0 section of Changes.txt

dsmiley:
Yes, 9.5

dsmiley:
There's an intermittent test failure (~50%) that reproducibly fails (100%) with the randomization seed.  I ignored the test for now, and also incorporated a small improvement RE client lifecycle.  I tried to debug it to understand the source of the problem but can't see anything other than a connection closing.  Perhaps the true bug is some exception that sadly gets gobbled up and the connection is closed -- a wild guess.

dsmiley:
I should revert this unless we get the test figured out.  Otherwise the state of this JIRA is unclear.

PR #2125, opened 2023-12-07
SOLR-16880: Add OAS to 'assembleRelease' task

gerlowskija:
This adds our OAS to the release artifacts gathered by the 'assembleRelease' task in gradle.  It produces a checksum for the OAS file and stores the file in a separate 'oas' directory (similar to the 'changes' directory which already exists for Changes.html).

One remaining piece here is signing - this commit attempts to produce an '*.asc' signature for the OAS when signing is enabled, but doesn't quite have it working yet.

https://issues.apache.org/jira/browse/SOLR-16880

# Description

Solr has, for some time, produced an OpenAPI spec ("OAS") describing its v2 APIs that it uses to (e.g.) generate Solr clients in various languages.  But while it uses the file internally, Solr doesn't (yet) expose the file to users who may wish to do similar things.

# Solution

This PR proposes that we publish our openapi spec as a release artifact akin to the various Changes.html files that we also put in artifacts.apache.org.  It does this by modifying the gradle tasks (esp. 'assembleRelease') in the 'distribution' gradle module.

# Tests

Manual testing via the `buildAndPushRelease.py` script and the smoke tester.  

# Checklist


gerlowskija:
One TODO remaining, if we choose to go this route, is to produce an "*.asc" signature for the OAS file.  The PR currently contains some logic that tries to do this, but it's still not working for whatever reason.  Something to work through at some point...

gerlowskija:
Alright, worked through the signing issues with some help (Thanks @HoustonPutman !).  This should be functional and ready to go, assuming it's the route we want to take.

solr/api/build.gradle
```diff
@@ -28,7 +28,7 @@ ext {
     jsClientDir = "${buildDir}/generated/js"
     pythonClientDir = "${buildDir}/generated/python"
     openApiSpecDir = "${buildDir}/generated/openapi"
-    openApiSpecFile = "${project.openApiSpecDir}/openapi.json"
+    openApiSpecFile = "${project.openApiSpecDir}/solr-${version}-openapi.json"
```
dsmiley, line 31: For JARs, the version is at the end.  Maybe we should do similar?  Just a thought.

dsmiley:
Does anything test this, or must it be manually inspected?  Basically, how do we have confidence that this is basically working?

janhoy (commented):
LGTM. Agree a better name is `solr-openapi-x.y.z`. And I'd wish for `.yml` which I find easier to read and diff, but that's mainly taste.

Wrt test, the smoke tester would do something obviously.

gerlowskija:
> Agree a better name is solr-openapi-x.y.z

OK, that makes sense.  Will make that change later this morning and merge.

PR #2126, opened 2023-12-07
SOLR-17096: Cluster Singleton plugin support in solr.xml

pjmcarthur:
https://issues.apache.org/jira/browse/SOLR-17096

# Description
It should be possible to install and manage the configuration of a cluster singleton plugin in an immutable deployment of SolrCloud. Currently, the only method to manage cluster singleton plugin configs is by using the `ContainerPluginApi` edit APIs, and this is not always possible, especially in the case of immutable SolrCloud deployments.

# Solution

This PR changes the `ContainerPluginRegistry` to use a pluggable source for plugin configs. The pluggable source is determined by a system property, `solr.clusterPluginsSource`, or by a declaration in solr.xml.

The default source is `ZkContainerPluginsSource`, which allows plugins to be added, removed or updated at runtime using the `ContainerPluginApi`. This matches the existing behavior.

A new OOTB plugin implementation is `NodeConfigClusterPluginsSource`. This plugin supports managing cluster singleton plugins via declarations in solr.xml, and does not support the edit APIs of `ContainerPluginApi`; these APIs are not registered in the case that node config is used as the plugin source.

An example of solr.xml that supports cluster singleton declarations would be
```
<clusterPluginsSource class="org.apache.solr.api.NodeConfigClusterPluginsSource"/>
<clusterSingleton name="my.plugin.name" class="my.plugin.class">
  <config>...</config>
</clusterSingleton>
```

# Tests

New tests are added in `NodeConfigContainerPluginsSourceTest` that spin up clusters and Core containers using the node config plugin source, and verify that container plugin cluster singleton declarations in solr.xml are added correctly to the `ContainerPluginsRegistry`.

Node configuration tests have been added to TestSolrXml

# Checklist


dsmiley (commented):
Remember to update TestSolrXml as well.

Overall, I like this!
I think there is a big issue with naming.  In JIRA, I advocated for using "Node" nomenclature instead of "Container", at least for anything the user sees (users don't know about CoreContainer; that's an internal thing and a name we regret).  But we can punt on renaming until the very end of this PR with respect to class names, to ease review.

dsmiley:
Oh I totally missed an important concern that I raised with @pvcnt 's previous PR #2016.  The configuration is very generic (the specific plugins are effectively untyped in solr.xml).  I took this back to the parent JIRA with my comment here: https://issues.apache.org/jira/browse/SOLR-15782?focusedCommentId=17782415&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17782415

solr/core/src/test/org/apache/solr/api/NodeConfigContainerPluginsSourceTest.java
```diff
@@ -0,0 +1,287 @@
...
+                "</solr>",
+                NODE_CONFIG_PLUGINS_SOURCE_XML + SingletonNoConfig.xmlConfig() + "</solr>"))
+        .addConfig(
+            "conf", TEST_PATH().resolve("configsets").resolve("cloud-minimal").resolve("conf"))
+        .configure();
+  }
+
+  /** Verifies that a cluster singleton config declared in solr.xml is loaded into the registry */
+  public void testOneClusterSingleton() {
+    CoreContainer cc = newCoreContainer(solrXml(SingletonNoConfig.xmlConfig()));
```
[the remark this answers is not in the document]
pjmcarthur: Yes, I can get rid of newCoreContainer here. The question of whether or not to use EmbeddedSolrServerTestRule brings up an interesting point for discussion that I think was also raised by @pvcnt; the embedded server it creates isn't ZK aware, and so it doesn't have the /cluster/plugins APIs registered. This obviously makes sense when the only source for plugin configs is ZK, but if NodeConfig is also a source, should the /cluster/plugin API be available in non-cloud mode?
I'm tempted to say no, at least for now, especially because the cluster singleton plugins I want to declare only apply to cloud mode anyway, and theoretically its easier to enable this in future (enable /cluster/plugins in non-cloud mode) than it is to undo it.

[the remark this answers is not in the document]
dsmiley: >  if NodeConfig is also a source, should the /cluster/plugin API be available in non-cloud mode?

I would prefer it not.  Does that question impact the work here (harder / easier)?  I understand a "cluster singleton" implies SolrCloud, but namely a request handler does not.  I know that's out of scope of this PR.  Request handlers are one of the most popular plugin types to customize.

solr/core/src/test-files/solr/solr-50-all.xml
```diff
@@ -76,4 +76,9 @@
     <repository name="local" class="a.b.C" default="true"/>
   </backup>
 
+  <clusterPluginsSource class="org.apache.solr.api.NodeConfigClusterPluginsSource"/>
+
+  <clusterSingleton name="testSingleton1" class="my.singleton.class"/>
+  <clusterSingleton name="testSingleton2" class="my.other.singleton.class"/>
```
pjmcarthur, line 82: I changed the element name here to `clusterSingleton`, but at present there is nothing that actually enforces the constraint that the declared class implements the `ClusterSingleton` interface. Theoretically, this would instantiate and manage the provided class in a manner that is consistent with it's class hierarchy, and not the `clusterSingleton` tag given here.

If we consider the equivalent operation of adding a plugin using the /cluster/plugin API, there is no part of the request that identifies the plugin as a cluster singleton.

If we go with this naming, should there also be a check somewhere that the class actually implements `ClusterSingleton`?

dsmiley: Enforcement would be ideal but not essential.

solr/core/src/java/org/apache/solr/core/NodeConfig.java
```diff
@@ -929,8 +941,8 @@ public NodeConfig build() {
           metricsConfig,
           cachesConfig,
           tracerConfig,
-          containerPluginsSource,
-          containerPlugins,
+          clusterPluginsSource,
+          clusterSingletonPlugins,
```
dsmiley, line 945: In NodeConfig, do we really need each supported type separated out?  It's only "clusterSingleton" now but there are others that will be added.  I could see simply an array of them or maybe a map keyed by name/type and value is the array.

Come to think of it, I wonder if we could migrate all PluginInfo used in NodeConfig to use this Map!  This would certainly majorly cut down on NodeConfig's Builder!  If we do such a thing, it'd probably be best in a separate PR (before or after).

solr/core/src/java/org/apache/solr/core/SolrXmlConfig.java
```diff
@@ -654,11 +655,28 @@ private static PluginInfo[] getClusterSingletonPluginInfos(List<ConfigNode> node
...
+    if (!duplicateNames.isEmpty()) {
+      throw new SolrException(
+          SolrException.ErrorCode.SERVER_ERROR,
+          "Multiple clusterSingleton sections with name '"
+              + String.join("', '", duplicateNames)
+              + "' found in solr.xml");
     }
-    return configs;
+
+    return plugins.toArray(new PluginInfo[nodes.size()]);
```
dsmiley, line 679: our codebase stopped the practice of pre-sized; just use 0

solr/core/src/test/org/apache/solr/api/NodeConfigClusterPluginsSourceTest.java
```diff
@@ -34,107 +31,94 @@
...
-      MapWriter config = apiInfo.getInfo().config;
-      Map<String, Object> configMap = new HashMap<>();
-      config.toMap(configMap);
-      assertEquals(cfgVal, configMap.get("cfgInt"));
-    } finally {
-      cc.shutdown();
-    }
+  public void testClusterSingletonsRegistered() {
+
+    CoreContainer cc = cluster.getOpenOverseer().getCoreContainer();
```
dsmiley, line 67: just use index 0; not the overseer.  We don't care who is the overseer.

solr/core/src/java/org/apache/solr/core/NodeConfig.java
dsmiley: CC @janhoy

dsmiley (commented):
Nice test, by the way.  I'm really glad you tested configurability.  These plugins are configured internally very unusually compared to the rest of Solr.

@pvcnt could you also take a look at this too?  You feared this approach might be too much work, I recall.

@noblepaul this is the start of a move to support these plugins in solr.xml   

There is a decision to make on naming... "ClusterPlugins" vs "ContainerPlugins" vs "NodePlugins".  And where to document such plugins when they can be configured in two places.  I tend to think solr.xml documentation should be primary and only minimally document in the cluster plugin API (where these are currently documented).

pvcnt:
@dsmiley I’ll have a more detailed look after my holidays. But one thing we should consider is that this PR follows a different design than what I did for replica placement plugins. For the latter, what goes inside solr.xml is applied as a first layer, and then cluster plugins as overrides. Here it’s one or the other, if my understanding is correct.

I believe this PR is heading towards a promising direction, but we should make sure that cluster singletons and replica placement plugins are at least configured the same.

pjmcarthur:
@pvcnt @dsmiley I think there should be a follow-up PR that addresses the replica placement plugin. My understanding is that from a user's perspective, the configuration of solr.xml wrt. the placement plugin would remain as it is, but the internal mechanics of instantiating the RPP factory would change to align with this PR. I would be able to commit to doing this follow-up PR if it is deemed the right path forward.

pvcnt:
I don't know when Solr 9.5 is going to be released, but IMHO we should make sure that it is not released with two different ways of configuring a cluster plugin in solr.xml, or even not released with a new way of configuring a cluster plugin in solr.xml that would be changing in Solr 9.6 (hence breaking backwards compatibility, etc.). I would hence be in favour of either doing both in a single PR, or starting by refactoring replica placement plugins. It would be a good exercice to validate that the approach can be extended to other cluster plugins.

The more philosophical question I have is about the behaviour. Do we want to be able to choose between solr.xml and plugin API (this PR) or to layer the latter on top of the former (my previous PR). I don't have a strong opinion on this, but that is a structuring question I believe.

dsmiley:
Agreed that 9.5 should not be released without reconciling the two approaches.

I think I would prefer to revert SOLR-17079 as it exists today, then merge this PR SOLR-17096 (after whatever changes in the review process), then do a quick (?) follow-up PR for SOLR-17079 to return the essential functionality building on what's here.  Do you think that hypothetical PR would be straight-forward?  You raised concerns a while back that what Paul is doing here would maybe be too much work, if I recall.

I don't think it's important/necessary to layer them as you did for replica placement.  That adds greater support requirements (testing that both can be used coherently) and nobody has asked for that.

pvcnt:
I believe it would still be interesting to have at least two different plugin types to validate that the generic approach is really generic, and what it takes to add a new type of plugin.

Another remark, it seems that there is no validation that a `<clusterSingleton/>` tag does indeed defines a cluster singleton. I believe the following would work and define a replica placement plugin:

```xml
<clusterSingleton name=".placement-plugin" class="org.apache.solr.cluster.placement.plugins.AffinityPlacementFactory"/>
```

This is why in SOLR-17079 we decided to go with an ad-hoc solution (i.e., one for every possible cluster plugin) instead of the generic solution. Otherwise, we are back to my initial proposal which is to have an untype `<plugin/>` tag in `solr.xml`, which is essentially what this PR does.

solr/core/src/java/org/apache/solr/core/NodeConfig.java
janhoy: You know I like simplifications. So as long as they are handled in the same generic way, it would be a welcome change. The devil could be in the details when one dives into the task though, so let's postpone it to a separate investigation.

solr/core/src/java/org/apache/solr/core/SolrXmlConfig.java
```diff
@@ -165,6 +166,9 @@ public static NodeConfig fromConfig(
     if (cloudConfig != null) configBuilder.setCloudConfig(cloudConfig);
     configBuilder.setBackupRepositoryPlugins(
         getBackupRepositoryPluginInfos(root.get("backup").getAll("repository")));
+    configBuilder.setClusterPluginsSource(getPluginInfo(root.get("clusterPluginsSource")));
```
dsmiley, line 169: There's a bigger question of naming.  I propose to not make a big naming change today or even very soon; can be deferred to Solr 10.  The internal class names you have chosen are fine in the mean time; they look relatively obvious relative to the existing class names, and that's a good thing.

But the user visible aspect should ideally be a name we don't think we're going to change.  I think that's only the  "clusterPluginsSource".  I can see why you are making a class name configurable but realistically the user has a boolean choice to make.  And ideally they wouldn't even have to explicitly make it -- simply define any such plugin in solr.xml and voila -- you get the NodeConfigClusterPluginsSource internally.  If we want a user to make an explicit choice, then perhaps we can piggyback off of an existing system properly `disable.configEdit` defined in `org.apache.solr.handler.SolrConfigHandler#CONFIGSET_EDITING_DISABLED_ARG` as a general flag for wether config edits are possible.  That flag is not publicly documented yet but we could do so.  It currently only applies to editing configSet at runtime.

dsmiley: note `org.apache.solr.handler.SolrConfigHandler#configEditing_disabled` is a public static final, but for testability you can change that.  If we expand the use as I suggest, it even deserves a more global location, like on NodeConfig perhaps (even if not actually configured in solr.xml).

dsmiley:
> I believe it would still be interesting to have at least two different plugin types to validate that the generic approach is really generic, and what it takes to add a new type of plugin.

I don't think it'll be *totally* generic; a given plugin type might best have some adjustments to create the PluginInfo.  For example, the ".placement-plugin" name should be added to the PluginInfo when the XML element is `<placementPlugin>`.  While that could be done in SolrXmlConfig (reading solr.xml), such an idiosyncrasy feels specific to the nature of the cluster edit API/JSON  that maybe could be in another spot (unsure).

>  it seems that there is no validation that a <clusterSingleton/> tag does indeed defines a cluster singleton.

I don't recall ever making a mistake of that nature, probably also because the class name is nearly always indicative of the type, so it'd look wrong if it was wrong as well.  Validation would be nice but not a requirement.  I leave the choice to validate or not up to Paul.

pvcnt:
The point I am raising about validation has actually already been raised by Paul there: https://github.com/apache/solr/pull/2126/files#r1424690508 I think it would be a mistake to not enforce the a `<clusterSingleton/>` actually defines a cluster singleton and not a replica placement plugin. Otherwise, what is the point of having a type tag, besides providing some documentation?

Again, with the current implementation, you can perfectly declare a replica placement plugin (or any kind of cluster plugin really) inside a `<clusterSingleton/>` and it will just work.

solr/core/src/java/org/apache/solr/core/SolrXmlConfig.java
dsmiley: Explicitly toggling could be deferred to a separate PR if desired.

pjmcarthur:
I will add the validation.

Regarding piggybacking on the existing `disable.configEdit` property, I like the idea, but I have one concern. 
If there are any Solr deployments out there that set `disable.configEdit` to true and also have cluster plugins installed, they will have issues after this change. The property value of `true` would cause the node config plugin source to be used instead of the ZK plugin source, and any plugins that were configured would not be loaded.

Maybe this isn't a real problem, and its unlikely that anyone would configure Solr in this way. You mention the property is not documented, so that could benefit us. I'm not sure how confident we can be about how safe it really is though. I think for that reason I slightly prefer the implicit selection of the pluigin source based on the presence of `clusterSingleton` declarations in solr.xml, but I'm ok to use the `disable.configEdit` property if we think its safe.

dsmiley:
Let's imagine someone out there has set `disable.configEdit` and is also posting to `/api/cluster/plugin`.  After the proposed change here, a call to that API is going to get an error (or _should_) telling them it's disabled.  So they know, and can act.  They aren't stuck; they have options.  Like don't set `disable.configEdit` (easiest) or don't POST to `/api/cluster/plugin` (use solr.xml instead).

solr/core/src/java/org/apache/solr/core/NodeConfig.java
```diff
@@ -223,17 +221,23 @@ private NodeConfig(
 
     if (this.clusterSingletonPlugins != null
         && this.clusterSingletonPlugins.length > 0
```
dsmiley, line 223: Do we need fields for clusterSingletonPlugins specifically?  Remember we're soon expanding this to other types like replica placement and hopefully more.  Could there be a Map of node/cluster/container plugins instead, in other words?  This way you're not singling out one specific type here in this check.

solr/core/src/java/org/apache/solr/core/NodeConfig.java
```diff
@@ -223,17 +221,23 @@ private NodeConfig(
...
+          "clusterSingleton section found in solr.xml but the property "
+              + CONFIG_EDITING_DISABLED_ARG
+              + " is set to false. clusterSingleton plugins may only be declared in solr.xml with immutable configs.");
     }
 
     setupSharedLib();
     initModules();
   }
 
+  public static boolean isImmutableConfigSet() {
```
dsmiley, line 237: My proposal is that this *not* be configSet specific.  So simply rename it to be more general.  Also FYI note that upstream there's a new sys-prop/env-var system and a dev list discussion.

solr/core/src/java/org/apache/solr/handler/SolrConfigHandler.java
```diff
@@ -104,9 +105,6 @@
 public class SolrConfigHandler extends RequestHandlerBase
     implements SolrCoreAware, PermissionNameProvider {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
-  public static final String CONFIGSET_EDITING_DISABLED_ARG = "disable.configEdit";
```
dsmiley, line 107: @janhoy if you didn't notice, we're elevating this very internal undocumented sys prop to node level to increase the scope beyond config editing of a core live to also include disabling cluster config edit.  Obviously the name of this sys prop is poor and your new efforts to standardize suggest we should take the opportunity to have a nicer name?

solr/core/src/java/org/apache/solr/api/ClusterPluginsSourceConfigurator.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.solr.core.NodeConfig;
+import org.apache.solr.core.SolrResourceLoader;
+import org.apache.solr.util.plugin.NamedListInitializedPlugin;
+
+/**
+ * Loads the {@link ClusterPluginsSource} depending on the declared implementation. The default
+ * implementation is {@link ZkClusterPluginsSource}, but can be overridden by a system property, or
+ * by declaring the implementation class in solr.xml
+ */
+public class ClusterPluginsSourceConfigurator implements NamedListInitializedPlugin {
```
dsmiley, line 29: Extending NamedListInitializedPlugin means we have an init() method.  But what becomes of those args?

solr/core/src/java/org/apache/solr/handler/SolrConfigHandler.java
dsmiley: Sorry for the back & forth here... I had thought let's use `disable.configEdit` but I don't want to shine a light (document) a badly named config, especially at a time when a config naming standardization effort is underway.

Absent further feedback, I think we should use a **new** property, like `solr.cluster.plugin.mutable` using `EnvUtils`.  It's scoped to just the `/cluster/plugin` API and thus doesn't have a debatable scope (too broad, like all config editing, which should get more agreement/care).  Such a narrow scope also suggests moving the setting to ContainerPluginsRegistry or ZkClusterPluginsSource.

solr/core/src/java/org/apache/solr/core/SolrXmlConfig.java
```diff
@@ -651,6 +651,27 @@ private static PluginInfo[] getBackupRepositoryPluginInfos(List<ConfigNode> cfg)
...
+          && !replicaPlacementFactory.name.equals(PlacementPluginFactory.PLUGIN_NAME)) {
+        throw new SolrException(
+            SolrException.ErrorCode.SERVER_ERROR,
+            "The replicaPlacementFactory name attribute must be "
+                + PlacementPluginFactory.PLUGIN_NAME);
+      }
+      clusterPlugins.add(replicaPlacementFactory);
+    }
+
+    return clusterPlugins.toArray(new PluginInfo[0]);
```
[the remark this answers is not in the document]
pjmcarthur: The interface to NodeConfig and its Builder uses native arrays, at least for the existing backupRepository plugins. I think if we return a list here, we would immediately have to convert it back to an array anyway, unless we change the Builder. I would rather not do that in this PR.

pjmcarthur:
@pvcnt FYI - This PR has been updated to handle the replicaPlacementFactory declared in solr.xml in the same way as cluster singleton plugins.

solr/core/src/java/org/apache/solr/core/NodeConfig.java
```diff
@@ -211,6 +215,17 @@ private NodeConfig(
     if (null == this.solrHome) throw new NullPointerException("solrHome");
     if (null == this.loader) throw new NullPointerException("loader");
 
+    if (this.clusterPlugins != null
+        && this.clusterPlugins.length > 0
+        && !ClusterPluginsSourceConfigurator.resolveClassName()
```
pvcnt, line 220: Can this be done later in CoreContainer, and rely on a method on the `ClusterPluginsSource` interface rather than a comparison of the class name?

pvcnt, line 220: Can this be done later in CoreContainer, and rely on a method on the ClusterPluginsSource interface rather than a comparison of the class name?

solr/core/src/java/org/apache/solr/core/NodeConfig.java
```diff
@@ -211,6 +215,17 @@ private NodeConfig(
...
+            .equals(NodeConfigClusterPluginsSource.class.getName())) {
+      throw new SolrException(
```
pvcnt, line 222: Maybe it can be a warning instead of an exception? Not entirely sure, but maybe people will have those in `solr.xml`, change the source later and want to keep them. Or maybe they could be injected into the cluster API as defaults at some point.

solr/core/src/java/org/apache/solr/core/SolrXmlConfig.java
```diff
@@ -647,6 +649,68 @@ private static PluginInfo[] getBackupRepositoryPluginInfos(List<ConfigNode> cfg)
...
+    if (nodes == null || nodes.isEmpty()) {
+      return new PluginInfo[0];
+    }
+
+    List<PluginInfo> plugins =
+        nodes.stream()
+            .map(n -> new PluginInfo(n, n.name(), true, true))
+            .collect(Collectors.toList());
+
+    // Cluster plugin names must be unique
```
pvcnt, line 684: Aren't plugin names supposed to be unique globally, instead of per plugin type? If so, this should be moved outside of this method specific to ClusterSingleton. We should also verify for reserved names, i.e., that a name does not start with a dot.

solr/core/src/java/org/apache/solr/api/NodeConfigClusterPluginsSource.java
```diff
@@ -0,0 +1,106 @@
...
+   */
+  private static String getPluginName(final PluginInfo pluginInfo) {
+
+    if (pluginInfo.name == null) {
+      if (pluginInfo.type.equals("replicaPlacementFactory")) {
+        return PlacementPluginFactory.PLUGIN_NAME;
+      }
+    }
+
+    return pluginInfo.name;
```
pvcnt, line 104: This can return null, and likely generate an NPE later on (?).

solr/core/src/test/org/apache/solr/cluster/placement/impl/PlacementPluginIntegrationTest.java
```diff
@@ -124,25 +124,6 @@ public void testConfigurationInSystemProps() {
     cc.shutdown();
   }
 
-  @Test
-  public void testConfigurationInSolrXml() {
```
pvcnt, line 128: Has this test be moved elsewhere? It is the only end-to-end (well, kindof) verification that the placement plugin configured in solr.xml is indeed taking into account in CoreContainer.

pvcnt (commented):
I like the approach, and feel much more comfortable now that replica placement plugins have been integrated.

solr/core/src/java/org/apache/solr/api/NodeConfigClusterPluginsSource.java
pjmcarthur: It should not be possible to return null here. The name attribute for cluster plugins plugin is mandatory and validated in SolrXmlConfig.
There are some exceptions for built in plugins, like the replica placement plugin, for which it isn't necessary to specify a name in solr.xml because the name is implied. For those cases only, the name is set here.

solr/core/src/test/org/apache/solr/cluster/placement/impl/PlacementPluginIntegrationTest.java
pjmcarthur: Yes, see `NodeConfigPlacementPluginTest`. The test requires a ZK aware SolrCloud now because it uses the ContainerPluginRegistry.

solr/core/src/java/org/apache/solr/core/NodeConfig.java
pjmcarthur: I don't think we'd be doing users any favors by allowing/encouraging ambiguity in configs, and ignoring some of it based on settings. IMO the exception is correct here.

solr/core/src/java/org/apache/solr/core/SolrXmlConfig.java
dsmiley: Indeed, it seems to be that Container/Cluster plugins have names in one namespace, but that's weird -- Solr doesn't work that way anywhere else.  I'd rather fix that limitation in a follow-up -- ContainerPluginsRegistry ought to have a a map keyed by type.

solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java
```diff
@@ -75,6 +73,8 @@
 public class ContainerPluginsRegistry implements ClusterPropertiesListener, MapWriter, Closeable {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
+  public static final String MUTABLE_CLUSTER_PLUGINS = "solr.cluster.plugin.mutable";
```
dsmiley, line 76: Fresh information: https://cwiki.apache.org/confluence/display/SOLR/System+property+naming+structure
Based on this audit and proposals (not yet formalized, granted), I think "solr.cluster.plugin.edit.enabled" would be the right property name here... ehh @janhoy ?

janhoy: Yea, one can argue how far the naming standard should go. I was mostly concerned with the lowercase dot separated aspect, with a controlled set of top-level categories. Guess it is nice-to-have to also agree on "enabled/disabled", "edit/mutable", "allowlist/includelist" etc

dsmiley:
There seems to be a test failure.  Once that's worked out, I plan to merge!

dsmiley:
The failure is unrelated; see https://github.com/apache/solr/pull/2204 for a fix.

PR #2145, opened 2023-12-12
SOLR-15484: use latest version of bouncycastle libraries

jdyer1:
There is now a new failure for `JWTAuthPluginIntegrationTest` [1] because the version of BouncyCastle we are using parses dates using the default locale.  (see `org.bouncycastle.asn1.x509.Time`) This is fixed with the latest version (`bc*-jdk15on:1.70` > `bc*-jdk18on:1.77`).  This I believe puts us on a newer version than what is used by `MockOAuth2Server` but best I can tell there are no other usages.  I think this will be a harmless upgrade.

-----
[1] https://jenkins.thetaphi.de/job/Solr-main-Linux/15897/testReport/junit/org.apache.solr.security.jwt/JWTAuthPluginIntegrationTest/classMethod/


janhoy (approved):
LGTM

janhoy:
I get a gradle warning that the NOTICE file is not in use. I believe it is required only for ASL and BSD licenses, not for MIT. I don't think it hurts to keep it around though.

PR #2149, opened 2023-12-12
Improve jsClient building for gradle caching

HoustonPutman:
Running "gradle clean" should no longer be necessary when changing branches.


gerlowskija (approved):
LGTM - sorry for the hassle!

PR #2154, opened 2023-12-14
SOLR-17104: Protect lib directories from being written to

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17104

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -294,6 +294,8 @@ public class CoreContainer {
   private SolrPackageLoader packageLoader;
 
   private final Set<Path> allowPaths;
+  // These paths do not allow write-access
+  private final List<Path> protectedPaths;
 
   private final AllowListUrlChecker allowListUrlChecker;
 
@@ -430,6 +432,7 @@ public class CoreContainer {
       }
     }
     this.allowPaths = allowPathBuilder.build();
+    this.protectedPaths = config.getProtectedDirectories();
 
     this.allowListUrlChecker = AllowListUrlChecker.create(config);
   }
@@ -657,6 +660,7 @@ public class CoreContainer {
     distributedCollectionCommandRunner = Optional.empty();
     allowPaths = null;
     allowListUrlChecker = null;
+    protectedPaths = null;
   }
 
   public static CoreContainer createAndLoad(Path solrHome) {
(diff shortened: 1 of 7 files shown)


epugh:
Is this the only place where this is needed?  Or are there lots of directories similar?

HoustonPutman:
Do you mean the restore code or the lib directories?

For the restore code, there might be more places we are writing user provided files, but I can't think of it.

As for the lib directories, this is all of them off of my head, but I could be missing a few.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1645,6 +1650,34 @@ public Set<Path> getAllowPaths() {
...
   }
 
+  /**
+   * Checks that the given path is not protected (i.e. can accept writes). Protected paths generally
+   * contain library directories that Solr loads jars and classes from.
+   *
+   * @param pathToAssert path to check
+   * @throws SolrException if path is contained inside a library folder
+   */
+  public void assertPathNotProtected(Path pathToAssert) throws SolrException {
```
dsmiley, line 1660: Please move this to SolrPaths.  CoreContainer has a ton of responsibilities; it doesn't need yet another one.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1645,6 +1650,34 @@ public Set<Path> getAllowPaths() {
...
+    pathToAssert = pathToAssert.toAbsolutePath();
+    List<Path> protectedPathWithCores =
+        new ArrayList<>(protectedPaths.size() + solrCores.getNumAllCores());
+    protectedPathWithCores.addAll(protectedPaths);
+    solrCores.getCoreDescriptors().stream()
+        .map(CoreDescriptor::getInstanceDir)
+        .map(p -> p.resolve("lib").toAbsolutePath())
+        .forEach(protectedPathWithCores::add);
+
+    for (Path path : protectedPathWithCores) {
```
dsmiley, line 1670: Why the separation here with intermediate "protectedPathWithCores list; why not end with forEach and do the check?

solr/core/src/java/org/apache/solr/core/NodeConfig.java
```diff
@@ -455,7 +460,32 @@ private void setupSharedLib() {
...
+      // Add in modules and tools
+      protectedDirectories.add(ModuleUtils.getModulesPath(getSolrInstallDir()).toAbsolutePath());
+      protectedDirectories.add(getSolrInstallDir().resolve("prometheus-exporter").toAbsolutePath());
+
+      // Add in server library folders
+      protectedDirectories.add(
+          getSolrInstallDir().resolve("server").resolve("lib").toAbsolutePath());
+      protectedDirectories.add(
+          getSolrInstallDir().resolve("server").resolve("solr-webapp").toAbsolutePath());
+    }
```
dsmiley, line 487: maybe assert that these actually exist?  If one day we change a name or move something, the assert will help us keep this up to date.

solr/core/src/java/org/apache/solr/core/NodeConfig.java
```diff
@@ -455,7 +460,32 @@ private void setupSharedLib() {
...
+    // Start with the shared lib directories
+    protectedDirectories.addAll(sharedLibPaths);
+
+    if (getSolrInstallDir() != null) {
+      // Add in modules and tools
+      protectedDirectories.add(ModuleUtils.getModulesPath(getSolrInstallDir()).toAbsolutePath());
+      protectedDirectories.add(getSolrInstallDir().resolve("prometheus-exporter").toAbsolutePath());
+
+      // Add in server library folders
+      protectedDirectories.add(
```
gerlowskija, line 483: [0] I'm a little surprised we don't have static constants for these directories somewhere.  Maybe we should...

solr/core/src/java/org/apache/solr/handler/SnapShooter.java
```diff
@@ -104,6 +104,7 @@ private void initialize(
     this.snapshotName = snapshotName;
     if ("file".equals(location.getScheme())) {
       solrCore.getCoreContainer().assertPathAllowed(Paths.get(location));
+      solrCore.getCoreContainer().assertPathNotProtected(Paths.get(location));
```
gerlowskija, line 107: [0] I worry a bit that folks might forget one or the other of these calls.  Maybe it'd be worth refactoring into a single utility method that optionally invokes the not-protected logic in the case of writes.  e.g.

```
public enum AccessType { READ, WRITE, UNKNOWN; }

public void assertPathAllowed(Path toCheck, AccessType access) {
    ...
    if (access == AccessType.WRITE) {
        assertPathNotProtected(...);
    }
}

// Note the 'private' here.
private void assertPathNotProtected(Path toCheck) {
}
```

gerlowskija (approved):
I left a few suggestions, mostly on how the path-checking is exposed.  But whether those make sense to implement or not, this is a huge improvement and LGTM!

PR #2158, opened 2023-12-15
Add missing log4j2.xml to test-files in solrj-streaming module

jdyer1:
I was confused because I could not get `solr.log` to output when running tests from the `solrj-streaming` module.  This is due to a missing `log4j2.xml` file which ought to have been included in the `test-files` directory.  This PR adds it, copied from the `solrj` module.


mkhludnev (approved):
Well. OK. But how it write into solr.log if it writes to stderr?

jdyer1:
@mkhludnev  With this log4j file added, if I run the test with `-Ptests.verbose=true`, solr.log gets outputted to stdout.

PR #2161, opened 2023-12-18
SOLR-17105, SOLR-17366: Require a location for local backups

HoustonPutman:
[SOLR-17105](https://issues.apache.org/jira/browse/SOLR-17105) & [SOLR-17366](https://issues.apache.org/jira/browse/SOLR-17366)


solr/CHANGES.txt
```diff
@@ -23,6 +23,8 @@ Improvements
 
 * SOLR-17077: When a shard rejoins leader election, leave previous election only once to save unneeded calls to Zookeeper. (Pierre Salagnac)
 
+* SOLR-17105: A base "location" is no required when using LocalFileSystemRepository for backup/restore. (Houston Putman, David Smiley, janhoy)
```
janhoy, line 26:
```suggestion
* SOLR-17105: A base "location" is now required when using LocalFileSystemRepository for backup/restore. (Houston Putman, David Smiley, janhoy)
```

solr/core/src/test/org/apache/solr/cloud/api/collections/BackupRestoreApiErrorConditionsTest.java
```diff
@@ -72,7 +78,8 @@ public static void setUpClass() throws Exception {
     final RequestStatusState backupState =
         CollectionAdminRequest.backupCollection(COLLECTION_NAME, BACKUP_NAME)
             .setRepositoryName(VALID_REPOSITORY_NAME)
-            .setLocation(validBackupLocation)
+            .setLocation(fullBackupDir.toString())
+            .setLocation(validBackupLocation.toString())
```
janhoy, line 82: Why two setLocation in a row?

janhoy (commented):
Looks nice. I did not try to fully understand the URI/Path juggling in the `getBackupLocation()` method - those can get pretty confusing.. Especially the fact that URI's `getPath` is not a `Path` but the string "path" part of the URI :)

solr/core/src/java/org/apache/solr/core/backup/repository/LocalFileSystemRepository.java
```diff
@@ -57,6 +57,9 @@ public void init(NamedList<?> args) {
     baseLocation =
         Optional.<String>ofNullable(getConfigProperty(CoreAdminParams.BACKUP_LOCATION))
             .map(Path::of)
+            .or(
+                () -> Optional.ofNullable(System.getProperty("solr.backups.path")).map(Path::of)
```
janhoy, line 61: Is it not possible to have multiple backup repos defined at the same time? Thus, this prop should probably qualify which one, e.g. `solr.backups.localfs.path`?

HoustonPutman: So I think the LocalFileSystemRepository is the only one that is implicitly created. (All others have to be explicitly mentioned in the solr.xml). That's why this would be necessary in the first place...

epugh:
I found trying out the local backup super confusing, docs did not make it clear what to do...  never quite got it.. maybe this being explicit would help!

HoustonPutman:
This might be worth getting into Solr 10, since its not back-compatible.

PR #2172, opened 2024-01-02
SOLR-17109: Give security manager explicit read access to sharedLib

tflobbe:


solr/server/etc/security.policy
```diff
@@ -219,6 +219,9 @@ grant {
   permission java.io.FilePermission "${solr.allowPaths}", "read,write,delete,readlink";
   permission java.io.FilePermission "${solr.allowPaths}${/}-", "read,write,delete,readlink";
 
+  permission java.io.FilePermission "${solr.sharedLib}", "read,readlink";
```
janhoy, line 222: Note that `solr.sharedLib` may contain a comma separated list of paths, which may not be supported in policy files. Is there a way to define policy programmatically?

tflobbe: True, it would fail with multiple paths (Note that the `allowPaths` just above is the same though). Another thing I don't really like is that these system properties are just "defaults", since the values are really just to populate `solr.xml`.
May be good enough though? Maybe the solution is to make it easy/documented how/when to update the policy for advanced usecases. Last I checked it wasn't easy to use a different policy (or amend one).
On the other hand, I don't know how much effort does the security manager configuration deserves, given it's deprecated.

janhoy: Yea, it will go away, so agree on that. One option could be to write a SolrCLI tool to generate/update the policy file. That way, `bin/solr` could run a tool that generates correct policy file based on whatever env/prop active and then start the main Solr process.

dsmiley: Furthermore, in a containerized world, thus with lots of read-only parts of the file system except for "/var", the point of enforcing file system checks in production via a security manager feels overly redundant -- too much of a pain to maintain.

dsmiley (approved):

epugh:
Can someone explain the future of security manager in Java to me?   My understanding is that it is going away (and no replacement??)...   So, wouldn't it be better just to deprecate the security manager now, and NOT rely on it?

epugh:
Also, @dsmiley  looks like you approved this change, can we go ahead and get it merged?

tflobbe:
I'm fine with merging this. @janhoy's comment is valid though, it would cause issues if `solr.sharedLib` has more than one directory

epugh:
> I'm fine with merging this. @janhoy's comment is valid though, it would cause issues if `solr.sharedLib` has more than one directory

do you want me to do it?   Or I can leave it up to you.  I just love seeing PR's and tickets close ;-)

epugh:
I wish we had some bats style tests for these things ;-).   I'm going to merge this when the tests finish.

epugh:
https://issues.apache.org/jira/browse/SOLR-17109

PR #2175, opened 2024-01-03
SOLR-16203: Initialize schema plugins loaded by SPI name

janhoy:
https://issues.apache.org/jira/browse/SOLR-16203

### DIFF
--- a/solr/core/src/java/org/apache/solr/schema/IndexSchema.java
+++ b/solr/core/src/java/org/apache/solr/schema/IndexSchema.java
@@ -45,12 +45,17 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.analysis.CharFilterFactory;
 import org.apache.lucene.analysis.DelegatingAnalyzerWrapper;
+import org.apache.lucene.analysis.TokenFilterFactory;
+import org.apache.lucene.analysis.TokenizerFactory;
 import org.apache.lucene.index.IndexableField;
 import org.apache.lucene.queries.payloads.PayloadDecoder;
 import org.apache.lucene.search.similarities.Similarity;
 import org.apache.lucene.util.BytesRef;
+import org.apache.lucene.util.ResourceLoaderAware;
 import org.apache.lucene.util.Version;
+import org.apache.solr.analysis.TokenizerChain;
 import org.apache.solr.common.ConfigNode;
 import org.apache.solr.common.MapSerializable;
 import org.apache.solr.common.SolrDocument;
@@ -652,6 +657,7 @@ public class IndexSchema {
       loadCopyFields(rootNode);
 
       postReadInform();
+      fieldTypes.values().forEach(this::informResourceLoaderAwareObjectsForFieldType);
 
     } catch (SolrException e) {
       throw new SolrException(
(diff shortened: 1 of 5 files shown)


janhoy:
First attempt in efadf94dc6ac87027ef708c71c58d3848e7a9a0d:

Naïve solution, explicitly calling `inform()` in `FieldTypePluginLoader`. The two tests from JIRA are added to this PR to demonstrate. 

There may be a more elegant place to do the init. We risk double inform if not using SPI since we don't check that here. Perhaps there is some method somewhere that could be called on the entire chain instead of in each init(), such as the one in `ManagedIndexSchema#informResourceLoaderAwareObjectsInChain()`. I did not dive deep here. Open to suggestions from folks who know the code better.

janhoy:
Second attempt:

The first attempt did not work for packages, since `FieldTypePluginLoader#init` was too early in the cycle to initialize analyzers from packages.

Managed schema handled the missing `inform` call after SPI already, so moved the method `informResourceLoaderAwareObjectsForFieldType` from `ManagedIndexSchema` to `IndexSchema` and use it at the end of `IndexSchema#readSchema`. All tests pass.

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
```diff
@@ -1991,6 +1997,37 @@ public IndexSchema deleteCopyFields(Map<String, Collection<String>> copyFields)
     throw new SolrException(ErrorCode.SERVER_ERROR, msg);
   }
 
+  /** Informs analyzers used by a fieldType. */
+  protected void informResourceLoaderAwareObjectsForFieldType(FieldType fieldType) {
```
uschindler, line 2001: does this need to be protected? The inner method is private, so why is this one public? A method reference in lambdas does not need to be visible in any way.

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
```diff
@@ -1991,6 +1997,37 @@ public IndexSchema deleteCopyFields(Map<String, Collection<String>> copyFields)
...
+    // must inform any sub-components used in the
+    // tokenizer chain if they are ResourceLoaderAware
+    if (!fieldType.supportsAnalyzers()) return;
+
+    Analyzer indexAnalyzer = fieldType.getIndexAnalyzer();
```
uschindler, line 2006: maybe do something like:

```java
Stream.of(fieldType.getIndexAnalyzer(), fieldType.getQueryAnalyzer(), noIdeaHowToHandleThis.getMultiTermAnalyzer()).filter(a -> a instanceof TokenizerChain)
  .distinct().foreach(a -> informResourceLoaderAwareObjectsInChain((TokenizerChain) a);
```

uschindler: maybe this one works:

```java
var stream = Stream.of(fieldType.getIndexAnalyzer(), fieldType.getQueryAnalyzer());
if (fieldType instanceof TextField) {
  stream = Stream.concat(stream, Stream.of(((TextField) fieldType).getMultiTermAnalyzer());
}
stream.filter(a -> a instanceof TokenizerChain)
  .distinct().foreach(a -> informResourceLoaderAwareObjectsInChain((TokenizerChain) a);
```

uschindler: (the distinct is by equals, but by default they use system's equals)

uschindler: null check is not needed because of `TokenizerChain.class::isInstance` (returns false on `null`)

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
janhoy: This is still also called from `ManagedIndexSchema` class which is a subclass.

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
janhoy: Note that this entire method is *moved* over from `ManagedIndexSchema`, and I have not modified it or attempted to optimize it. May we leave improvements to anoter PR?

uschindler: Ah, you're right. I just thought al little bit of deduplication of code would be fine.

It's up to you to apply that, but I can make a separate PR.

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
uschindler: This is not ideal. If you move the call to this private method into `postReadInfom()` of *this* class (`IndexSchema`), then the subclass would also call it and it could be private.
Here it is called in readSchema *after* `postReadInform()`. If it is moved there both implementations automatically use it.

uschindler (changes requested):

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
```diff
@@ -652,6 +657,7 @@ protected void readSchema(ConfigSetService.ConfigResource is) {
       loadCopyFields(rootNode);
 
       postReadInform();
+      fieldTypes.values().forEach(this::informResourceLoaderAwareObjectsForFieldType);
```
uschindler, line 660: move this line into the `postReadInform()` method called before.

solr/core/src/java/org/apache/solr/schema/ManagedIndexSchema.java
```diff
@@ -1307,37 +1299,6 @@ protected void postReadInform() {
     }
```
uschindler, line 1299: remove this whole method, too

uschindler (changes requested):

solr/core/src/java/org/apache/solr/schema/IndexSchema.java
janhoy: I'll close this comment and defer to new PR after this landed

janhoy:
Thanks for nice proposals @uschindler , the method is now private and code cleaner

uschindler (approved):

PR #2180, opened 2024-01-06
Auto-label PRs with labeler action

janhoy:
The project has some PR labels defined, but we don't use them because it is boring to manually label things, see https://github.com/apache/solr/pulls

This PR implements https://github.com/actions/labeler github action. It will make it easier to quickly discern the PRs and filter on area of interest.

The configuration will assign labels based on what paths have modifications. See `.github/labeler.yml` for labels and rules.

I borrowed some label names from lucene, e.g. `tool:build` and `module:XXX`.
Other labels are not prefixed such as `bats-tests`, `example`, `admin-ui`, `solrj`. I feel those should be prefixed in some way too, ideas welcome.

See screenshot below for a test I did in a fork:
Skjermbilde 2024-01-06 kl  00 37 27

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,0 +1,12 @@
+name: "Pull Request Labeler"
+on:
+  - pull_request_target
+
+jobs:
+  labeler:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/labeler@v5
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -10,3 +10,4 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/labeler@v5
+      
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,166 @@
+# Add 'dependencies' label to any PR updating dependencies
+dependencies:
+  - changed-files:
+      - any-glob-to-any-file:
+        - versions.props
+        - versions.lock
+        - solr/licenses/**
+
+# Add 'documentation' label to any changes within ref-guide or dev-docs
+documentation:
+  - changed-files:
+      - any-glob-to-any-file:
+        - dev-docs/**
+        - solr/solr-ref-guide/**
+        - solr/documentation/**
+        - help/**
+        - solr/README.adoc
+        - CONTRIBUTING.md
+        - README.md
+
+# Add 'example' label to examples
+example:
+  - changed-files:
+      - any-glob-to-any-file:
+          - solr/example/**
+
+# Add 'bats-tests' label
+bats-tests:
+  - changed-files:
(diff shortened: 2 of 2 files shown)


.github/workflows/labeler.yml
```diff
@@ -0,0 +1,12 @@
...
+  - pull_request_target
+
+jobs:
+  labeler:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/labeler@v5
```
MarcusSorealheis, line 12:
```suggestion
      - uses: actions/labeler@v5
      
```

MarcusSorealheis (approved):
This is a helpful PR and will enable to more improvements. The only issue I saw was the last line.

dsmiley:
Cool!

janhoy:
I changed `tool:docker` to `docker`. Would you prefer plain labels `build` instead of `tool:build` and `scripts` instead of `tool:scripts`?

epugh:
This is super exciting...  I am thinking about other areas of Solr that specific groups of folks might want to monitor, and therefore might want to have a label to cue them.   `streaming` is one area that some folks have some knowledge for reviewing...   And we can add more in the future.  For example maybe we want a label around `tests`....

Does prefixing do something special like give you another filter?  I don't mind `bats-test` or `example` as a label...

epugh (approved):
LGTM.

janhoy:
> Does prefixing do something special like give you another filter? I don't mind `bats-test` or `example` as a label...

No, it's more a convention to namespace different types of tags. E.g. if we want `type:bug`, `type:feature` that could be separate from the component labels. But I'm too ok with both styles. It is also possible in GH to rename a tag at a later time, so we can add name-spacing later if we want to.

janhoy:
Thanks, let's adjust with follow-up PRs if necessary

alessandrobenedetti:
I am late on the train, but I really like this initiative.
I've been labelling PRs from time to time, especially to track progress in the vector search space, but I admit I was doing it manually and periodically, with no rigour.

This is a much better way!
Thanks @janhoy for your effort!

I was wondering if it makes sense to add cascade rules that label also depending on the package?
I would love to see labels such as indexing:UpdateRequestProcessor, indexing:updateHandler, querying:MoreLikeThis etc

Should we create a different issue/PR to discuss this?

janhoy:
> I was wondering if it makes sense to add cascade rules that label also depending on the package?

That would be nice too. And probably possible  due to the folder structure of the source code. Please have a look in the labeler docs if you want to attempt a few of those. The “module:name” labels are kind of same thing already. Should those labels have some prefix, such as “feature:xxx” to distinguish from the broader labeling? Or “category:xxx”?

janhoy:
@alessandrobenedetti Btw since the first iteration is merged already, please open a new PR for your suggestion.

gerlowskija:
Joining this discussion late, but this is really cool. Thanks Jan!

PR #2184, opened 2024-01-08
Introduce stale bot

janhoy:
Following Lucene's lead and adding a stale bot that will label open, non-draft PRs as `Stale` after 60 days of inactivity. It is almost identical fork of Lucene's config: https://github.com/apache/lucene/blob/main/.github/workflows/stale.yml

I re-wrote the stale-message text a bit. I don't think random contributors can tag developers as reviewer, so I encouraged instead tagging people in a comment or emailing the dev-list.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -29,7 +29,8 @@ jobs:
 
           stale-pr-message: >
             This PR has not had activity in the past 2 weeks, labeling it as stale.
-            If the PR is waiting for review, notify the dev@solr.apache.org list.
+            Any new activity will remove the stale label. To attract more reviewers, tag
+            someone or notify the dev@solr.apache.org mailing list.
             Thank you for your contribution!
 
           # TODO: Adjust budget and debug after initial testing
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,37 @@
+# This workflow warns of PRs that have had no activity for a specified amount of time.
+#
+# For more information, see https://github.com/actions/stale
+name: Mark stale pull requests
+
+on:
+  # Run every day at 00:00 UTC
+  schedule:
+    - cron: '0 0 * * *'
+  # Or run on demand
+  workflow_dispatch:
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+
+    steps:
+      - uses: actions/stale@v9
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+          days-before-pr-stale: 14
+          days-before-issue-stale: -1   # we don't use issues
+          days-before-close: -1         # don't close stale PRs/issues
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,39 @@
...
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+          days-before-pr-stale: 14
+          days-before-issue-stale: -1   # we don't use issues
+          days-before-close: -1         # don't close stale PRs/issues
+          exempt-draft-pr: true         # don't mark draft PRs as stale
+          stale-pr-label: "stale"       # label to use when marking as stale
+
+          stale-pr-message: >
+            This PR has not had activity in the past 2 weeks, labeling it as stale.
```
cpoerschke, line 32: Maybe shortening "has not had" to "had no" and use "visible" to signal that maybe someone looked but did not (yet) comment?
```suggestion
            This PR had no visible activity in the past 2 weeks, labeling it as stale.
```

.github/workflows/stale.yml
```diff
@@ -0,0 +1,39 @@
...
+            Any new activity will remove the stale label. To attract more reviewers, tag
```
cpoerschke, line 33: And wondering about pull requests with merge conflicts e.g. to suggest or not to suggest to try and resolve any merge conflicts. Maybe that would be a different kind of bot though.
```suggestion
            Any new activity will remove the stale label. To attract more reviewers, please tag
```

janhoy: I don't think it matters whether there are merge conflicts or not. If a PR is stale, it should be quite obvious what next steps are? Or is it unclear? A contributor should aim to get all lights green, and then ask for review. But if the contributor needs help she can of course ask for help at any time.

cpoerschke (approved):

.github/workflows/stale.yml
```diff
@@ -0,0 +1,39 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+
+    steps:
+      - uses: actions/stale@v9
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+          days-before-pr-stale: 14
```
janhoy, line 25: I wonder if we can should consider 30 days stale time to avoid some noise over holidays etc.

And what if we sent an automated monthly email to dev@ with list of stale PS. That would let both committers and contributors triage and decide whether to close or contribute.

janhoy: Changed to 30 days. 

The monthly email would be a separate bot/script in a different GH action. It would need to use a pre-accepted email address as sender to be allowed to post to dev@ list, and a SMTP server to send from.

janhoy:
Being on the conservative side, I upped to 60 days before stale, and a daily budget of 10 PRs labeled.
This should create a minimum of friction and accomodate for the fact that in the Solr project, some development cycles are just slow without being stale. If noone touched a PR for 60 days, it's hard to argue that it is active, so a nudge by the bot may help drive some of those that just fell off the radar further.

PR #2185, opened 2024-01-09
Various asf.yaml updates

janhoy:
I browsed the [asf.yaml docs](https://cwiki.apache.org/confluence/pages/viewpage.action?spaceKey=INFRA&title=git+-+.asf.yaml+features), and found several new capabilities related to GitHub that I believe we should add:

* `protected_branches` - our main branch should at least be protected from force-push
* `protected_tags` - prevent deletion of `release*` tags
* `autolink_jira` - make references to short-form SOLR JIRA issues in PRs clickable
* `collaborators` - list of github user names to allow full PR edit karma

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -18,6 +18,18 @@ github:
     merge:  false
     rebase: false
 
+  protected_branches:
+    main: {}
+
+  protected_tags:
+    - "releases/*"
+
+  autolink_jira:
+    - SOLR
+
+  collaborators:
+    - solrbot
+
 notifications:
   commits:      commits@solr.apache.org
   issues:       issues@solr.apache.org
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -18,8 +18,15 @@ github:
     merge:  false
     rebase: false
 
+  # TODO: Add to this list for each new minor release
   protected_branches:
     main: {}
+    branch_9_0: {}
+    branch_9_1: {}
+    branch_9_2: {}
+    branch_9_3: {}
+    branch_9_4: {}
+    branch_9x: {}
 
   protected_tags:
     - "releases/*"


.asf.yaml
```diff
@@ -18,6 +18,18 @@ github:
...
+    main: {}
+
+  protected_tags:
+    - "releases/*"
+
+  autolink_jira:
+    - SOLR
+
+  collaborators:
+    - solrbot
```
cpoerschke, line 31: > * `collaborators` - list of github user names to allow full PR edit karma

Am curious if that would cover its own PRs only or others' pull requests too, and/or when it (solrbot) would have a need to use full PR edit.

janhoy: The list of users here are allowed full edit rights for all issue/PRs, so it will be users who are not yet committers, but we trust them to do triage. Solrbot would need this right in order to set a label on the PRs it opens. In [`renovate-config.js`](https://github.com/solrbot/renovate-github-action/blob/main/solr/renovate-config.js) we could [configure it to set a `dependencies` label ](https://docs.renovatebot.com/configuration-options/#addlabels) and even labels detailing what kind of dependency it is (e.g. test vs compile). But that is disabled as of now as solrbot does not have that karma.

We could imagine future bots to assign reviewers to PRs intelligently, and such a bot would also be added here. But primarily it would be cool to get 3-5 non-committer developers to volunteer to help with PRs.

janhoy:
Also, I consider `required_conversation_resolution`, which will require all conversations to be resolved before merging is possible. I have seen too quick merges of looong PRs where the author missed an open conversation.

.asf.yaml
```diff
@@ -18,6 +18,18 @@ github:
     merge:  false
     rebase: false
 
+  protected_branches:
+    main: {}
```
dsmiley, line 22: Is there a way to add `branch_*`?

dsmiley (approved):

.asf.yaml
janhoy: I cannot see it documented. But we could easily enumerate all of them and let the release wizard add new branches to the list?

janhoy: Also, `.asf.yaml` does not support glob format (see https://the-asf.slack.com/archives/CBX4TSBQ8/p1668009454564229?thread_ts=1667901002.937809&cid=CBX4TSBQ8), but will probably be added once GH api supports it.

PR #2186, opened 2024-01-09
Correct the log.error to include the file name instead of hard-coded …

haldmour-atypon:
Correct the log.error to include the file name instead of the hard-coded value.

`log.error("Error loading external value source + fileName + {}{}", var33, otherErrors < 10 ? "" : "\tSkipping future errors for this file.");`


uschindler (approved):
See discussion in #2186

PR #2189, opened 2024-01-12
Remove dot-files not needed and update .gitignore

dsmiley:
These dot-files seem like not worth keeping.

One of them was referenced in `.gitignore` so that brought my attention to remove some things there as well.  Maybe could be another PR; I can split if asked.  If someone uses less common tools (e.g. "syncthinghing"), they can add ignore rules for their own checkout, or put them in a common file (like in your home dir) and refer to it via adding a `core.excludesFile` git config reference.  That could even go in `~/.gitconfig`


.gitignore
```diff
@@ -11,8 +11,6 @@ build/
 **/*.ipr
 **/*.iws
 out/
-#    IntelliJ creates this folder, ignore.
-/dev-tools/missing-doclet/out/
```
dsmiley, line 15: obsoleted by the `out/` added a year later on the line above

.gitignore
```diff
@@ -24,22 +22,20 @@ out/
 
 .editorconfig
 
-# syncthing
-.stignore
```
dsmiley, line 28: @markrmiller  added this one once

.inferconfig
dsmiley, line 1: Was an interesting tool used by Lift, but Lift is gone

mkhludnev (approved):

.gitignore
janhoy: I see no harm in this, as it reduces the chance of any future contributor accidentally committing this file if they use such tool. A few extra lines in gitignore is tolerable. We’re not adding every possible tool in here, but devs will add such as they experience the pain with a certain tool.

dsmiley:
Will merge in 24 hours or so

PR #2190, opened 2024-01-12
GH Labeler action: Category labels

janhoy:
Followup from #2180 

Add GH labels based on java packages, e.g. for a package `org.apache.solr.foo[.xxx]`, assign label `cat:foo`. There are 23 new cat labels defined in total.

Also added label `tests` when any test is touched.

Most PRs will get two more labels due to this. Large refactor PRs or e.g. changing `javax.` to `jakarta.` would touch large parts of the code base and may thus get a large number of labels assigned.

This is still just a proposal. I'm open to suggestions for how to group some packages together into larger topics to obtain fewer categories. I also contemplated using different naming than the package name, e.g. "requesthandler" instead of "handler", but then there is something simplistic about that 1:1 relation too... 

As long as we're using this tool, we are constrained to the capabilities, i.e. looking at file paths. No fancy java-knowhow here...

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -224,6 +224,9 @@ cat:index:
       - any-glob-to-any-file:
           - '**/org/apache/solr/index/**'
           - '**/org/apache/solr/hdfs/index/**'
+          - '**/org/apache/solr/update/**'
+          - '**/org/apache/solr/hdfs/update/**'
+          - '**/org/apache/solr/scripting/update/**'
 
 cat:logging:
   - changed-files:
@@ -267,10 +270,3 @@ cat:servlet:
   - changed-files:
       - any-glob-to-any-file:
           - '**/org/apache/solr/servlet/**'
-
-cat:update:
-  - changed-files:
-      - any-glob-to-any-file:
-          - '**/org/apache/solr/update/**'
-          - '**/org/apache/solr/hdfs/update/**'
-          - '**/org/apache/solr/scripting/update/**'
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -181,8 +181,7 @@ tests:
           - '**/src/test/**'
           - '**/src/test-files/**'
 
-# Add category labels based on java pacakges
-# In essence, for a package org.apache.solr.foo[.xxx], assign label cat:foo
+# Add category labels loosely based on java packages
 cat:analysis:
   - changed-files:
       - any-glob-to-any-file:
@@ -197,10 +196,6 @@ cat:cloud:
   - changed-files:
       - any-glob-to-any-file:
           - '**/org/apache/solr/cloud/**'
-
-cat:cluster:
-  - changed-files:
-      - any-glob-to-any-file:
           - '**/org/apache/solr/cluster/**'
 
 cat:backup:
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -170,3 +173,135 @@ module:sql:
...
+  - changed-files:
+      - any-glob-to-any-file:
+          - '**/org/apache/solr/cli/**'
+
+cat:cloud:
+  - changed-files:
+      - any-glob-to-any-file:
+          - '**/org/apache/solr/cloud/**'
+
+cat:cluster:
```
dsmiley, line 201: I don't think this is really distinct from SolrCloud

.github/labeler.yml
```diff
@@ -170,3 +173,135 @@ module:sql:
...
+  - changed-files:
+      - any-glob-to-any-file:
+          - '**/org/apache/solr/filestore/**'
+
+cat:handler:
+  - changed-files:
+      - any-glob-to-any-file:
+          - '**/org/apache/solr/handler/**'
+
+cat:highlight:
```
dsmiley, line 227: too feature specific to warrant a category?

.github/labeler.yml
```diff
@@ -170,3 +173,135 @@ module:sql:
...
+          - '**/org/apache/solr/spelling/**'
+
+cat:update:
+  - changed-files:
+      - any-glob-to-any-file:
+          - '**/org/apache/solr/update/**'
+          - '**/org/apache/solr/hdfs/update/**'
+          - '**/org/apache/solr/scripting/update/**'
+
+cat:util:
```
dsmiley, line 303: Is such a generic category useful?

dsmiley (commented):
We could bike-shed so much of this.  I question if so many are useful.

.github/labeler.yml
janhoy: Probably not, will consolidate with 'cloud'

.github/labeler.yml
janhoy: Agree. I propose to fold [`highlight`, `query`, `request`, `response`, `spelling`] ==> `search`

.github/labeler.yml
janhoy: Agree, will remove this label. edits to only a util file will then get no label.

.github/labeler.yml
```diff
@@ -170,3 +173,104 @@ module:sql:
...
+  - changed-files:
+      - any-glob-to-any-file:
+          - '**/org/apache/solr/core/**'
+
+cat:filestore:
+  - changed-files:
+      - any-glob-to-any-file:
+          - '**/org/apache/solr/filestore/**'
+
+cat:handler:
```
janhoy, line 217: Can handler go in `search` as well?

.github/labeler.yml
```diff
@@ -170,3 +173,104 @@ module:sql:
...
+  - changed-files:
+      - any-glob-to-any-file:
+          - '**/org/apache/solr/security/**'
+
+cat:servlet:
+  - changed-files:
+      - any-glob-to-any-file:
+          - '**/org/apache/solr/servlet/**'
+
+cat:update:
```
janhoy, line 271: Perhaps `index` can go into `update`. The `index` package is mostly lucene merge policies...

janhoy:
Did some consolidation, now down to 9 new category labels:
> api, cli, cloud, index, metrics, packagemanager, schema, search, security

janhoy:
@dsmiley WDYT about latest iteration? Should we simplify it into even fewer categories?

dsmiley:
LGTM @janhoy

dsmiley (approved):

PR #2191, opened 2024-01-12
ReleaseWizard commands to protect new minor branch

janhoy:
As a new `branch_X_Y` branch is created, it should be added to list of protected branches.

These commands are not tested, so please do a thorough manual review.


HoustonPutman (commented):
We could use `yq` to do this automatically, but this works for now.

Also this needs to go to the main branch, not the release branch necessarily, right?

janhoy:
Main and 9x since the 9.5 release will use wizard from branch 9x

PR #2192, opened 2024-01-12
SOLR-15960 Move EnvUtils to Solrj common

janhoy:
SOLR-15960

I wanted to let the system property lookup done during variable expansion in xml files (e.g. solr.xml `${solr.foo.bar:xxx}`) take advantage of `EnvUtils`, but realized that that code resides in the "common" module of solrj, which all other modules depend on.

So in this PR I move `EnvUtils` to solrj-common.

PS: Since the original feature is still not released (9.5), we don't need a new JIRA.


epugh (approved):
LGTM.   Someday we'll figure out a `solr/server` and `solr/solrj` dependent on `solr/core` ;-).

PR #2193, opened 2024-01-12
SOLR-15960 Cut over System.getProperty() to EnvUtils for modules

janhoy:

This is the first in a series of PRs to cut over use of `System.getProperty()` to `EnvUtils.getProp()`. This PR does it for the modules. Followup PRs will target core.

I'm re-using same JIRA since it is not yet released, and this is part of the same effort.


solr/modules/s3-repository/src/java/org/apache/solr/s3/S3BackupRepositoryConfig.java
```diff
@@ -59,7 +58,7 @@ public S3StorageClient buildClient() {
   }
 
   private static String getStringConfig(NamedList<?> config, String property) {
-    String envProp = System.getenv().get(toEnvVar(property));
+    String envProp = EnvUtils.getProp(property);
```
janhoy, line 61: No need for `toEnvVar(prop)` here (which does `solr.foo.bar -> SOLR_FOO_BAR`) since we can rely on the fact that `EnvUtils.getProp` will obtain `solr.foo.bar` even when set as env `SOLR_FOO_BAR`.

epugh (approved):
LGTM.   I sort of wish it was the full `getProperty`.....

solr/modules/s3-repository/src/java/org/apache/solr/s3/S3BackupRepositoryConfig.java
epugh: Nice!

janhoy:
> LGTM. I sort of wish it was the full `getProperty`.....

We can rename `getProp -> getProperty` if you feel it is better. The code has still not left the building in a release...

epugh:
> > LGTM. I sort of wish it was the full `getProperty`.....
> 
> We can rename `getProp -> getProperty` if you feel it is better. The code has still not left the building in a release...

It matches the System.getProperty pattern, and just seems like one less thing to mentally parse...

janhoy:
> > > LGTM. I sort of wish it was the full `getProperty`.....
> > 
> > 
> > We can rename `getProp -> getProperty` if you feel it is better. The code has still not left the building in a release...
> 
> It matches the System.getProperty pattern, and just seems like one less thing to mentally parse...

It has crossed my mind. Can be a followup to this..

PS: Do you think that `EnvUtils` is a covering name, or would `NodeEnv` or `SolrEnv` or similar be better?

epugh:
> > > > LGTM. I sort of wish it was the full `getProperty`.....
> > > 
> > > 
> > > We can rename `getProp -> getProperty` if you feel it is better. The code has still not left the building in a release...
> > 
> > 
> > It matches the System.getProperty pattern, and just seems like one less thing to mentally parse...
> 
> It has crossed my mind. Can be a followup to this..
> 
> PS: Do you think that `EnvUtils` is a covering name, or would `NodeEnv` or `SolrEnv` or similar be better?

I think `EnvUtils` is just fine.  It lets you cover lots of environment related stuff!

epugh:
The latest commits are looking great....   Thanks for the rename!

PR #2194, opened 2024-01-12
SOLR-15960: Rename getProp as getProperty

janhoy:

More familiar name. Also renamed `getProps() -> getProperties()`


janhoy:
This can be done before #2193

epugh (approved):
Loving it!

PR #2197, opened 2024-01-14
Improving array list and map init

renatoh:
Small code improvements how the Set and ArrayList are initlaized


solr/core/src/java/org/apache/solr/handler/component/SearchComponent.java
```diff
@@ -114,18 +112,17 @@ public void initializeMetrics(SolrMetricsContext parentContext, String scope) {
...
-    map.put(FacetModule.COMPONENT_NAME, FacetModule.class);
-    map.put(MoreLikeThisComponent.COMPONENT_NAME, MoreLikeThisComponent.class);
-    map.put(StatsComponent.COMPONENT_NAME, StatsComponent.class);
-    map.put(DebugComponent.COMPONENT_NAME, DebugComponent.class);
-    map.put(RealTimeGetComponent.COMPONENT_NAME, RealTimeGetComponent.class);
-    map.put(ExpandComponent.COMPONENT_NAME, ExpandComponent.class);
-    map.put(TermsComponent.COMPONENT_NAME, TermsComponent.class);
-
-    standard_components = Collections.unmodifiableMap(map);
+    standard_components =
```
pvcnt, line 115: With this, there is even no need anymore for the `static` block.

renatoh: good point, I removed the static-block and directly initialize the Map.
I did a quick search, it looks like many static-blocks could be removed by making use of the colleciton-factory methods

renatoh: @pvcnt Are you able to merge my PR?

pvcnt: No I can't, sorry. :(

renatoh:
@risdenk would you mind having  a look at the small PR? thanks

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -117,7 +117,7 @@ public class SearchHandler extends RequestHandlerBase
   private SolrCore core;
 
   protected List<String> getDefaultComponents() {
-    ArrayList<String> names = new ArrayList<>(8);
+    ArrayList<String> names = new ArrayList<>(9);
```
epugh, line 120: is there any way to get rid of this magic number?   This seems very error prone.....?

solr/core/src/java/org/apache/solr/handler/component/SearchComponent.java
```diff
@@ -111,21 +109,16 @@ public void initializeMetrics(SolrMetricsContext parentContext, String scope) {
...
-    map.put(StatsComponent.COMPONENT_NAME, StatsComponent.class);
-    map.put(DebugComponent.COMPONENT_NAME, DebugComponent.class);
-    map.put(RealTimeGetComponent.COMPONENT_NAME, RealTimeGetComponent.class);
-    map.put(ExpandComponent.COMPONENT_NAME, ExpandComponent.class);
-    map.put(TermsComponent.COMPONENT_NAME, TermsComponent.class);
-
-    standard_components = Collections.unmodifiableMap(map);
-  }
+  public static final Map<String, Class<? extends SearchComponent>> standard_components =
+      Map.of(
```
epugh, line 113: kind of a nit pick, but the variable should be `standardComponents` ?    We really need more code quality checks for things like this globaly...

epugh (changes requested):
Thanks for making these changes and here are some comments!

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
renatoh: @epugh thinking about it again, it does not make too much sense to use the factory method List.of just so that we need to adapt code for dealing with a read-only list. What I would suggest instead is just initialize the ArrayList without a size at all. The default size then is applied, which is 10, we do not really save anything create a  list with an initialize size < 10.

solr/core/src/java/org/apache/solr/handler/component/SearchComponent.java
dsmiley: agreed

dsmiley (approved):
I'm inclined to merge this in a couple days if I don't hear back to the contrary.  It's nice as-is; the feedback is minor.  Thanks for contributing @renatoh !

renatoh:
@dsmiley Thanks for picking this up!

PR #2202, opened 2024-01-17
SOLR-17119: Exception swallowing in ContainerPluginsApi causes some config validation errors to be ignored

ycallea:
https://issues.apache.org/jira/browse/SOLR-17119

# Description

When registering a [ConfigurablePlugin](https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/api/ConfigurablePlugin.java) through the `/cluster/plugin` API using the [V2Request](https://github.com/apache/solr/blob/main/solr/solrj/src/java/org/apache/solr/client/solrj/request/V2Request.java) client, exceptions raised within the `void configure(T cfg)` method are swallowed in [ContainerPluginsApi](https://github.com/apache/solr/blob/b5db179164919df657845f5e71f9fe0f0a1c86d8/solr/core/src/java/org/apache/solr/handler/admin/ContainerPluginsApi.java#L153-L156).

This behavior can be troublesome as the exceptions raised in the `configure(...)` method are usually validation errors on the configuration to be applied, which are therefore ignored and allow for invalid plugin configuration to become effective.

# Solution

A small change has been implemented in `ContainerPluginsApi` to properly include the exceptions raised while configuring a configurable cluster plugin in the response's payload errors.

A minor code clean-up on the modified method have also been included to this pull request.

# Tests

- A [new test](https://github.com/apache/solr/pull/2202/files#diff-2bf86d66bf1beff376a584c216d4da601f680078de3a86639c427daf6e3a876eR381-R411) has been written to illustrate the priorly failing scenario and validate the changes.
- Existing tests continue to pass.

# Checklist

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -83,7 +83,8 @@ New Features
 
 Improvements
 ---------------------
-(No changes)
+* SOLR-17119: When registering or updating a ConfigurablePlugin through the `/cluster/plugin` API,
+  config validation exceptions are now propagated to the callers. (Yohann Callea)
 
 Optimizations
 ---------------------
--- a/solr/core/src/java/org/apache/solr/handler/admin/ContainerPluginsApi.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/ContainerPluginsApi.java
@@ -17,15 +17,12 @@
 
 package org.apache.solr.handler.admin;
 
-import static org.apache.lucene.util.IOUtils.closeWhileHandlingException;
-
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import org.apache.solr.api.AnnotatedApi;
 import org.apache.solr.api.ClusterPluginsSource;
 import org.apache.solr.api.Command;
 import org.apache.solr.api.ContainerPluginsRegistry;
@@ -130,28 +127,24 @@ public class ContainerPluginsApi {
   }
 
   private void validateConfig(PayloadObj<PluginMeta> payload, PluginMeta info) throws IOException {
-    if (info.klass.indexOf(':') > 0) {
-      if (info.version == null) {
-        payload.addError("Using package. must provide a packageVersion");
-        return;
+    if (info.klass.indexOf(':') > 0 && info.version == null) {
+      payload.addError("Using package. must provide a packageVersion");
+      return;
+    }
+
+    final List<String> errs = new ArrayList<>();
+    final ContainerPluginsRegistry.ApiInfo apiInfo =
+        coreContainer.getContainerPluginsRegistry().createInfo(payload.getDataMap(), errs);
+
+    if (errs.isEmpty()) {
+      try {
+        apiInfo.init();
+      } catch (Exception e) {
+        log.error("Error instantiating plugin ", e);
+        errs.add(e.getMessage());
       }
     }
-    List<String> errs = new ArrayList<>();
-    ContainerPluginsRegistry.ApiInfo apiInfo =
-        coreContainer.getContainerPluginsRegistry().createInfo(payload.getDataMap(), errs);
-    if (!errs.isEmpty()) {
-      for (String err : errs) payload.addError(err);
-      return;
-    }
-    AnnotatedApi api = null;
-    try {
-      apiInfo.init();
-    } catch (Exception e) {
-      log.error("Error instantiating plugin ", e);
-      errs.add(e.getMessage());
-      return;
-    } finally {
-      closeWhileHandlingException(api);
-    }
+
+    errs.forEach(payload::addError);
   }
 }
--- a/solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
+++ b/solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
@@ -545,18 +545,14 @@ public class TestContainerPlugin extends SolrCloudTestCase {
       implements ConfigurablePlugin<ValidatableConfig> {
     @Override
     public void configure(ValidatableConfig cfg) {
-      cfg.validate();
+      if (!cfg.willPassValidation) {
+        throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "invalid config");
+      }
     }
   }
 
   public static class ValidatableConfig implements ReflectMapWriter {
     @JsonProperty public boolean willPassValidation;
-
-    public void validate() {
-      if (!willPassValidation) {
-        throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "invalid config");
-      }
-    }
   }
 
   private Callable<V2Response> getPlugin(String path) {
--- a/solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
+++ b/solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
@@ -47,6 +47,7 @@ import org.apache.solr.client.solrj.request.beans.PluginMeta;
 import org.apache.solr.client.solrj.response.V2Response;
 import org.apache.solr.cloud.ClusterSingleton;
 import org.apache.solr.cloud.SolrCloudTestCase;
+import org.apache.solr.common.SolrException;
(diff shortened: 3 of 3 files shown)


solr/core/src/java/org/apache/solr/handler/admin/ContainerPluginsApi.java
```diff
@@ -134,29 +131,25 @@ public void update(PayloadObj<PluginMeta> payload) throws IOException {
...
+    if (errs.isEmpty()) {
+      try {
+        apiInfo.init();
+      } catch (Exception e) {
+        log.error("Error instantiating plugin ", e);
+        errs.add(e.getMessage());
+      }
     }
+
+    errs.forEach(payload::addError);
```
pvcnt, line 152: It doesn't change the outcome, but maybe it could be done in an `else` block? It's difficult to explain why, but I find this statement difficult to parse in my head in this current form.

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
```diff
@@ -509,6 +541,24 @@ public void m2(SolrQueryRequest req, SolrQueryResponse rsp) {
...
+    @Override
+    public void configure(ValidatableConfig cfg) {
+      cfg.validate();
+    }
+  }
+
+  public static class ValidatableConfig implements ReflectMapWriter {
+    @JsonProperty public boolean willPassValidation;
+
+    public void validate() {
```
pvcnt, line 555: Can this be merged into `ConfigurablePluginWithValidation#configure`? Again, just to make the code easier to follow by avoiding one jump.

solr/core/src/java/org/apache/solr/handler/admin/ContainerPluginsApi.java
```diff
@@ -134,29 +131,25 @@ public void update(PayloadObj<PluginMeta> payload) throws IOException {
...
     }
-    AnnotatedApi api = null;
-    try {
-      apiInfo.init();
-    } catch (Exception e) {
-      log.error("Error instantiating plugin ", e);
-      errs.add(e.getMessage());
-      return;
-    } finally {
-      closeWhileHandlingException(api);
```
pvcnt, line 158: Is this method (missing in the modified version) what was causing the exception to be swallowed?

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
ycallea: Sure, it can definitely be implemented this way :

```java
public static class ConfigurablePluginWithValidation implements ConfigurablePlugin<ValidatableConfig> {
  @Override
  public void configure(ValidatableConfig cfg) {
    if (!cfg.willPassValidation) {
      throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "invalid config");
    }
  }
}

public static class ValidatableConfig implements ReflectMapWriter {
  @JsonProperty public boolean willPassValidation;
}

```

Is it what you have in mind ?

solr/core/src/java/org/apache/solr/handler/admin/ContainerPluginsApi.java
ycallea: The exception was swallowed here because the error message was mistakenly only added to the local `errs` list (line 155) but never included in the response's payload.

The removal of the call to `closeWhileHandlingException(...)` is a minor code clean-up.
In this context, the purpose of this method was to close the `AnnotatedApi` declared above (line 150), but as it is never instantiated this finally block always resulted in a no-op.

I therefore removed both this unused local variable declaration and the finally block which was dedicated to closing this resource.

solr/core/src/java/org/apache/solr/handler/admin/ContainerPluginsApi.java
ycallea: I chose to implement it this way so we are only adding errors in the response's payload (`payload.addError(...)`) at a single place.

We could however also do it as such, to keep the same flow as before my change if it improves readability : 

```java
final List<String> errs = new ArrayList<>();
final ContainerPluginsRegistry.ApiInfo apiInfo =
    coreContainer.getContainerPluginsRegistry().createInfo(payload.getDataMap(), errs);

if (errs.isEmpty()) {
  try {
    apiInfo.init();
  } catch (Exception e) {
    log.error("Error instantiating plugin ", e);
    payload.addError(e.getMessage());
  }
} else {
  errs.forEach(payload::addError);
}
```

ycallea: On a side note, the way the `createInfo(...)` method (and by extension the `ApiInfo` constructor) is currently implemented feels a bit unintuitive to be honest, you have to pass it a reference to a list that will be populated with some error messages, should any validation error occur in here.

I wanted to keep the changes from this pull request small enough to only remediate the identified bug, but I feel it would be worth refactoring as part of another pull request.

In my opinion, we should just throw exceptions instead of relying on this errors' list (see [ContainerPluginsRegistry.java#L329-L413](https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/api/ContainerPluginsRegistry.java#L329-L413)), it would simplify the code and ease errors management when constructing `ApiInfo` objects.

pvcnt: Oh, I didn't get it that only appending to `errs` had no effect (which is what you are fixing). I take back my comment then, it's fine as it is!

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
pvcnt: Yes. Very minor comment, I just thought it would slightly improve the readability of the test. ;)

ycallea: Thanks, I updated the code accordingly in https://github.com/apache/solr/pull/2202/commits/729b14b191f79fc99759f94cdd3a45a09e7ffa2c

pvcnt (approved):

dsmiley:
Can a CHANGES.txt entry be proposed here?  I suppose this is an "Improvement", though I can sympathize why you filed the JIRA issue as a bug.

bruno-roustant (approved):
Thanks @ycallea

ycallea:
> Can a CHANGES.txt entry be proposed here? I suppose this is an "Improvement", though I can sympathize why you filed the JIRA issue as a bug.

I agree classifying it as a bug could be debatable.
I don't have much history on the initial implementation so I can't tell if this behavior was intended or not, thus I'm fine classifying it as an improvement if it makes more sense.

I added a `CHANGES.txt` entry proposal in https://github.com/apache/solr/pull/2202/commits/ad6c5958326971b2e77d1ff0fa3211ca448c3998, for now it has been added as an improvement for the 9.5.0 release.
Let me know if you need me to make some changes to the proposed description, category and/or release 🙂

solr/CHANGES.txt
```diff
@@ -140,6 +140,9 @@ Improvements
 
 * SOLR-17096: solr.xml now supports declaring clusterSingleton plugins (Paul McArthur, David Smiley)
 
+* SOLR-17119: When registering a ConfigurablePlugin through the `/cluster/plugin` API using the
```
dsmiley, line 143: Please move to 9.6

solr/CHANGES.txt
```diff
@@ -140,6 +140,9 @@ Improvements
...
+  V2Request client, config validation exceptions are now propagated to the callers (Yohann Callea)
```
dsmiley, line 144: Suggest rewording "the V2Request client" to "SolrJ with V2Request".   (people know SolrJ is a client and this word sets better context to users who have no clue what V2Request even is).

Curious; is"V2Request" actually pertinent to the bug?  Would someone using plain HTTP see the problem?

ycallea: It would indeed also affect users querying the API in plain HTTP without SolrJ.
I agree that the proposed description seems to imply that the issue only exist when calling the API through SolrJ, I updated it to be less specific.

I also moved the change to 9.6.0 as requested in the previous comment.

Thanks again for your guidance 🙂

PR #2206, opened 2024-01-18
SOLR-17121: Fix SchemaCodecFactory to get PostingsFormat and DocValues from field.

bruno-roustant:
This fixes [SOLR-13968](https://issues.apache.org/jira/browse/SOLR-13968) and adds a test case.
Thanks @dsmiley for spotting this!


dsmiley (approved):
+1 LGTM.

I suppose there may be some maintenance to this test when we upgrade to Lucene 10 but should be trivial, and not really avoidable.

dsmiley:
CHANGES.txt needed, of course

PR #2214, opened 2024-01-23
SOLR-17120: handle null value when merging partials

eukaryote:
https://issues.apache.org/jira/browse/SOLR-17120

Avoid a `NullPointerException` that can occur under some circumstances when performing multiple partial updates of the same document in separate requests using `commitWithin`.

# Description

When merging older partial updates into the doc for newer partial updates, the existing solr code would generate a `NullPointerException` under some circumstances when it tried to iterate over a `null` value.

# Solution

I replaced the loop that started with `for (Object val : olderDoc.getFieldValues(fieldName))` with code that first assigns the value to a temporary variable and then either iterates over the value (performing the same action as  previously on each value if it is non-nulll) or calls `newerDoc.addField(fieldName, null)` instead if the value is `null`.

# Tests

No tests yet, but I was able to regularly reproduce the error before, and have been running a patched version of a `solr-core-9.4.0-SNAPSHOT.jar` file that I created with this change against the 9.4.0 version and haven't been able to reproduce the error.

# Checklist

21 tests failed, all of which were in `org.apache.solr.s3.S3BackupRepositoryTest` or `org.apache.solr.s3.S3OutputStreamTest`, which my change doesn't seem related to at all. EDIT: After I removed my aws configs and ran the tests again, they all passed.


solr/CHANGES.txt
```diff
@@ -164,6 +164,9 @@ Bug Fixes
 
 * SOLR-17090: The v2 "delete alias" API no longer errantly returns a 405 status code (Jason Gerlowski)
 
+* SOLR-17120: Fix NullPointerException in UpdateLog.applyOlderUpdates that can occur if there are multiple partial
+  updates of the same document in separate requests using commitWithin. (Calvin Smith, Christine Poerschke)
+
```
cpoerschke, line 169: Tentatively adding for 9.5.0 here but we can move it to 9.6.0 if it's too late for the 9.5.0 release.

cpoerschke (approved):

gerlowskija:
> 21 tests failed, all of which were in org.apache.solr.s3.S3BackupRepositoryTest or org.apache.solr.s3.S3OutputStreamTest

Solr has some test-flakiness issues, and the S3 tests are common culprits.  Getting tests to run reliably is a bit of a dark-art.  We try to get around this as best we can by fixing flakiness when we find it, running multiple iterations, etc.

TL;DR - I wouldn't worry about these failures too much; whoever ends up merging this can validate that tests run for them 👍

solr/CHANGES.txt
gerlowskija: +1 to target 9.5.0 - seems like a pretty safe change!

gerlowskija (approved):
LGTM - do you want to handle merging @cpoerschke , or should I?  Don't want to step on your toes here...

PR #2218, opened 2024-01-23
SOLR-17130: edismax-matchalldocs-optimization

risdenk:
https://issues.apache.org/jira/browse/SOLR-17130

# Description

This unwraps the case of using edismax query parser and `q=*:*`. Prior to this change, `*:*` was being changed to `+*:*` which wraps the `MatchAllDocsQuery` with a boolean query. We can unwrap the boolean query since its not necessary if the `MatchAllDocsQuery` is the only clause.

I found this by looking at https://issues.apache.org/jira/browse/SOLR-14765 and wondering why when using `edismax` the optimizations were skipped even though the query was `*:*`. The boolean wrapped `MatchAllDocsQuery` means that some of the optimization logic is skipped - https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java#L1623

# Tests

Added tests to confirm that if `*` or `*:*` is the query then Solr returns just a `MatchAllDocsQuery` instead of a boolean wrapped `MatchAllDocsQuery`.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/ExtendedDismaxQParser.java
+++ b/solr/core/src/java/org/apache/solr/search/ExtendedDismaxQParser.java
@@ -187,28 +187,25 @@ public class ExtendedDismaxQParser extends QParser {
       query.add(f, BooleanClause.Occur.SHOULD);
     }
 
-    //
-    // create a boosted query (scores multiplied by boosts)
-    //
     Query topQuery = QueryUtils.build(query, this);
+
+    // If topQuery is a boolean query, unwrap the boolean query to check if it is just
+    // a MatchAllDocsQuery. Using MatchAllDocsQuery by itself enables later optimizations
+    BooleanQuery topQueryBoolean = (BooleanQuery) topQuery;
+    if (topQueryBoolean.clauses().size() == 1) {
+      Query onlyQuery = topQueryBoolean.clauses().get(0).getQuery();
+      if (onlyQuery instanceof MatchAllDocsQuery) {
+        topQuery = onlyQuery;
+      }
+    }
+
+    // create a boosted query (scores multiplied by boosts)
     List<ValueSource> boosts = getMultiplicativeBoosts();
     if (boosts.size() > 1) {
       ValueSource prod = new ProductFloatFunction(boosts.toArray(new ValueSource[0]));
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/search/ExtendedDismaxQParser.java
```diff
@@ -199,6 +199,18 @@ public Query parse() throws SyntaxError {
       topQuery = FunctionScoreQuery.boostByValue(topQuery, boosts.get(0).asDoubleValuesSource());
     }
 
+    // If topQuery is a boolean query, unwrap the boolean query to check if it is just
+    // a MatchAllDocsQuery. Using MatchAllDocsQuery by itself enables later optimizations
+    if (topQuery instanceof BooleanQuery) {
+      BooleanQuery topQueryBoolean = (BooleanQuery) topQuery;
+      if (topQueryBoolean.clauses().size() == 1) {
+        Query onlyQuery = topQueryBoolean.clauses().get(0).getQuery();
+        if (onlyQuery instanceof MatchAllDocsQuery) {
```
risdenk, line 208: There are probably other cases besides `MatchAllDocsQuery` that could be unwrapped here. Another example being `ConstantScoreQuery`.

solr/core/src/java/org/apache/solr/search/ExtendedDismaxQParser.java
```diff
@@ -199,6 +199,18 @@ public Query parse() throws SyntaxError {
       topQuery = FunctionScoreQuery.boostByValue(topQuery, boosts.get(0).asDoubleValuesSource());
     }
 
+    // If topQuery is a boolean query, unwrap the boolean query to check if it is just
```
dsmiley, line 202: I recommend moving this to above the section right above here since FunctionScoreQuery may wrap the BooleanQuery or whatever `topQuery` happens to be.

risdenk: Addressed in 81ca321

dsmiley (approved):

epugh:
Seems like this could be merged?  I'm happy to do it if you'all would like me to.

dsmiley:
Suggested CHANGES.txt Optimization:
* SOLR-17130: Optimized edismax when `*:*` is used (Kevin Risden)

PR #2219, opened 2024-01-24
SOLR-15781: Document v2 API syntax conventions

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-15781

# Description

Currently, decisions around what our v2 API should look like are scattered across a large number of JIRA tickets, PR reviews, and spreadsheets.  There's no one place giving an overview of the desired syntax (beyond perhaps the general guidance in SIP-16 to move towards "REST" where possible).

# Solution

This commit attempts to address this by introducing docs describing some of the conventions used in our v2 API.

Room for improvement remains: parameters conventions remain a big gap, as does sourcing to the original rationale behind many of these decisions.  But this PR provides a place to build off of moving forward.

# Tests

N/A - doc change only.

# Checklist

- [x] I have added documentation to the dev-docs.

### DIFF
--- a/dev-docs/apis.adoc
+++ b/dev-docs/apis.adoc
@@ -51,28 +51,28 @@ Separating the API "definition" and "implementation" in this way allows us to on
 
 Writing a new v2 API may appear daunting, but additions in reality are actually pretty simple:
 
-1. *Agree on Endpoint Syntax*: Before implementing, developers should generate consensus around what the new API will look like.  General guidelines and conventions for the v2 API syntax are described <<v2-api-conventions.adoc,here>>.
-2. *Create POJO ("Plain Old Java Object") classes as needed to represent the API's request and response*:
+. *Agree on Endpoint Syntax*: Before implementing, developers should generate consensus around what the new API will look like.  General guidelines and conventions for the v2 API syntax are described <<v2-api-conventions.adoc,here>>.
+. *Create POJO ("Plain Old Java Object") classes as needed to represent the API's request and response*:
 ** POJOs are used to represent both the body of the API request (for some `POST` and `PUT` endpoints), as well as the response from the API.
 ** Re-use of existing classes here is preferred where possible.  A library of available POJOs can be found in the `org.apache.solr.client.api.model` package of the `api` gradle project.
 ** POJO class fields are typically "public" and annotated with the Jackson `@JsonProperty` annotations to allow serialization/deserialization.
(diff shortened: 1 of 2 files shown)


dev-docs/apis.adoc
```diff
@@ -51,27 +51,28 @@ Separating the API "definition" and "implementation" in this way allows us to on
 
 Writing a new v2 API may appear daunting, but additions in reality are actually pretty simple:
 
-1. *Create POJO ("Plain Old Java Object") classes as needed to represent the API's request and response*:
+1. *Agree on Endpoint Syntax*: Before implementing, developers should generate consensus around what the new API will look like.  General guidelines and conventions for the v2 API syntax are described <<v2-api-conventions.adoc,here>>.
```
epugh, line 54: In markdown you can do:

```
1. First item
1. Second item
1. Third item
1. Fourth item 
```

Does that work in adoc?  would that make life easier?

dev-docs/v2-api-conventions.adoc
```diff
@@ -0,0 +1,76 @@
...
+Solr's "modify collection" functionality uses partial update semantics, but the v2 API uses `PUT` instead of `PATCH` due to the relative obscurity of the latter.
+
+For use within the v2 API, the four "popular" HTTP methods have the following semantics and implications:
+
+* `GET` - used for non-mutating (i.e. "read only") requests. Most often used to list elements of a particular resource type, or fetch information about about a specific named resource.
+* `POST` - used for non-idempotent resource modifications.
+* `PUT` - used for idempotent resource modifications.
+* `DELETE` - Used to delete or cleanup resource
+
+== Exceptional Cases - "Command" APIs
```
epugh, line 69: Do we need to also call out the "index" special case, of being a collection or core in this doc?

epugh (changes requested):
Two comments...

dev-docs/apis.adoc
gerlowskija: It definitely works in adoc.  The question I'd ask though is whether Github is smart enough to do it as well when it's previewing these asciidoc files.

That's the main way we expect these to be consumed, right?  Through the Github UI?  (i.e. we don't actually build these asciidoc files into any thing analogous to the ref-guide?)

I'll push the change and see how Github handles it; if GH doesn't do-the-right-thing I'll revert.

gerlowskija: Looks like GH does do the right thing 👍

PR #2227, opened 2024-01-28
SOLR-17068: Resolve mish mash of bin/post and bin/solr post references in favour of bin/solr post.

epugh:
https://issues.apache.org/jira/browse/SOLR-17068

# Description

Removing old bin/post references.

# Solution

Clean ups through out docs and code base.

# Tests

Bats and unit tests.

# Checklist

- [X ] I have reviewed the guidelines for [How to Contribute](https://github.com/apache/solr/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.
- [X ] I have created a Jira issue and added the issue ID to my pull request title.
- [ X] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [ X] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)


dsmiley:
Maybe I'm being grumpy but I like to think of the unfortunate addition of "post" to bin/solr as an implementation detail out of convenience to make multi-platform and general supportability easier for bin/post.  Does this mean we don't want users to use bin/post either?  I think not.

gerlowskija (commented):
The large majority of the changes here make sense to me - we're moving away from `bin/post` in 10.0 so we want to shift documentation and other usage in that direction.  That all fits the title of the PR.

But I'm seeing some other changes here too - loosening some `Option` declarations in PostTool.java, re-adding support for the 'c' parameter.  Can you add some context around these changes?

Are those follow-on changes from some prior PR?  Are they only done here to accommodate some of the new `bin/solr post` call sites?

epugh:
Thanks for asking about the other changes.   So what I was finding was that the `-c` usage was everywhere, and that we even have explicit support now in the SolrCLI to look up S`OLR_PORT`, `SOLR_HOST`, `SOLR_SCHEME` from your environment, which is there to make the `-c` option work better, and that forcing everyone to use `-url` was just too big of a change...   I decided that eliminating the `-c` was actually a step backwards for the user experience, especially when OTHER commands support it...    It definitely made reading the ref guide docs much harder...  to support `-c`, then I had to loosen the `Option`.    The other changes are bug fixes that i found while actually testing out the code!

gerlowskija (approved):
OK, makes sense.  Do we need to document that change of strategy somewhere other than SOLR-17068?  Maybe that's good enough 🤷 

Anyway, LGTM.

epugh:
> Maybe I'm being grumpy but I like to think of the unfortunate addition of "post" to bin/solr as an implementation detail out of convenience to make multi-platform and general supportability easier for bin/post. Does this mean we don't want users to use bin/post either? I think not.

I think that is EXACTLY what we want.  no more `bin/post` and no more `bin/postlogs` either...   Neighter of them supports basic auth for example...   Neighter of them picks up SOLR_PORT, SOLR_HOST, etc variables...    each has weird special logic to make them executable...       Now, in the future, if someone wants to make a better CLI thatn bin/solr, or break it up into bin/server and bin/client, sure...  but yeah, in the short term, no more bin/post and no more bin/postlogs.

gerlowskija:
Yeah, I thought that was the whole point - `bin/solr post` gets us a bunch of things that `bin/post` doesn't, we don't want to maintain them both indefinitely, so we should standardize and deprecate.

Is there a reason we _wouldn't_ converge across docs and test-usage in this case?

PR #2229, opened 2024-01-29
SOLR-17066: Add GenericSolrRequest.setRequiresCollection

gerlowskija:

# Description

SOLR-17066 added a 'defaultCollection' field to each SolrClient implementation, similar to the one that has long been in use for SolrJ's "cloud" clients.  This default collection (or core) is used on a request-by-request basis to build the complete HTTP path, based on the value of SolrRequest.requiresCollection().

This is a particular challenge for GenericSolrRequest though, which can be used to make both collection-agnostic and collection-aware requests.

# Solution

This commit adds a GenericSolrRequest setter, `setRequiresCollection(boolean)`, to allow GSR users to specify which requests they would like the client-level default collection to be used on.  It updates a number of GSR usages in the project to be a bit more explicit about whether they're making a collection/core/index request, or not.

GSR.requiresCollection defaults to `false`, so this PR only touches those uses that _do_ requires a collection/core.
 
# Tests

# Checklist


solr/core/src/java/org/apache/solr/cli/ExportTool.java
```diff
@@ -620,7 +622,7 @@ boolean exportDocsFromCore() throws IOException, SolrServerException {
 
         try (SolrClient client = SolrCLI.getSolrClient(baseurl, credentials)) {
           expectedDocs = getDocCount(replica.getCoreName(), client, query);
-          GenericSolrRequest request;
+          QueryRequest request;
```
epugh, line 625: I like that you were able to move away from the `GenericSolrRequest` here!

solr/core/src/java/org/apache/solr/packagemanager/PackageManager.java
```diff
@@ -237,8 +237,10 @@ public Map<String, SolrPackageInstance> getPackagesDeployed(String collection) {
       NamedList<Object> result =
           solrClient.request(
               new GenericSolrRequest(
-                  SolrRequest.METHOD.GET,
-                  PackageUtils.getCollectionParamsPath(collection) + "/PKG_VERSIONS"));
+                      SolrRequest.METHOD.GET,
+                      PackageUtils.getCollectionParamsPath(collection) + "/PKG_VERSIONS")
+                  .setRequiresCollection(
+                      false) /* Making a collection request, but already baked into path */);
```
epugh, line 243: interesting, is this a place we could try and improve things to not have this logic?   Seems  a shame to have a collection request and state 'but we don't require it'....?

solr/core/src/java/org/apache/solr/packagemanager/PackageUtils.java
```diff
@@ -155,9 +155,12 @@ public static String getFileFromJarsAsString(List<Path> jars, String filename) {
   }
 
   /** Returns JSON string from a given Solr URL */
-  public static String getJsonStringFromUrl(SolrClient client, String path, SolrParams params) {
+  public static String getJsonStringFromUrl(
```
epugh, line 158: I wonder if it would be simpler to just have "getJsonStringFromCollection" and "getJsonStringFromSolr" or something?   After all, it's not a generic "getJsonStringFromUrl" where I can pass in any URL, despite what the name of the method appears to suggest...!    Then we could skip passing a magic boolean around...

solr/core/src/test/org/apache/solr/security/BasicAuthIntegrationTest.java
```diff
@@ -330,7 +330,7 @@ public void testBasicAuth() throws Exception {
       assertPkiAuthMetricsMinimums(3, 3, 0, 0, 0, 0);
 
       // Query that succeeds
-      GenericSolrRequest req = new GenericSolrRequest(SolrRequest.METHOD.GET, "/select", params);
+      final var req = new QueryRequest(params);
```
epugh, line 333: Yay!

solr/solrj/src/java/org/apache/solr/client/solrj/request/GenericSolrRequest.java
epugh, line 1: ooh, docs!

solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java
```diff
@@ -1013,7 +1013,8 @@ public void testGetRawStream() throws Exception {
                 getBaseUrl() + "/debug/foo", DEFAULT_CONNECTION_TIMEOUT, DEFAULT_CONNECTION_TIMEOUT)
             .build()) {
       GenericSolrRequest req =
-          new GenericSolrRequest(SolrRequest.METHOD.GET, "/select", params("q", "*:*"));
+          new GenericSolrRequest(SolrRequest.METHOD.GET, "/select", params("q", "*:*"))
```
epugh, line 1016: Could this be replaced with a `QueryRequest` instead?

epugh (commented):
Some feedback.  What jumped out at me is that we really abuse `GenericSolrRequest` all over the place.   I also am hoping that as more of the V2 api's come online, we could reduce the places that we use `GenericSolrRequest`, kind of like how in some places you could swap in a `QueryRequest` in this PR!

gerlowskija:
> What jumped out at me is that we really abuse `GenericSolrRequest` all over the place. I also am hoping that as more of the V2 api's come online, we could reduce the places that we use `GenericSolrRequest`, kind of like how in some places you could swap in a `QueryRequest` in this PR!

Yeah, I was thinking the same thing.  There's a few times we use GenericSolrRequest as a way to hit a custom test endpoint or something like that, but that's definitely the exception and not the rule.  90% of these GSR usages are bandaids covering up the fact that we're missing SolrJ coverage for particular APIs.

epugh:
> > What jumped out at me is that we really abuse `GenericSolrRequest` all over the place. I also am hoping that as more of the V2 api's come online, we could reduce the places that we use `GenericSolrRequest`, kind of like how in some places you could swap in a `QueryRequest` in this PR!
> 
> Yeah, I was thinking the same thing. There's a few times we use GenericSolrRequest as a way to hit a custom test endpoint or something like that, but that's definitely the exception and not the rule. 90% of these GSR usages are bandaids covering up the fact that we're missing SolrJ coverage for particular APIs.
> 
> I think these invocations would be a great place to start sneaking in some of our v2 SolrRequest objects, as you suggested. (Would be a great newdev ticket!)

Totally agree on getting some of the new v2 objects in!   Also, I wonder if looking at uses of GenericSolrRequest would provide insight into which v2 apis should be prioritized?

solr/core/src/java/org/apache/solr/packagemanager/PackageManager.java
gerlowskija: We could - basically that'd involve changing `PackageUtils.getCollectionParamsPath` to return a value of `/config/params` instead of the current value `/api/collections/someCollectionName/config/params`.  The method has 8 or 10 call sites, so we'd have to restructure the logic in each of those places.

It's not that big of a lift; I'm happy to go that route if it'd be your preference?  Lmk.

That said, I have a slight preference for skipping this now.  Mostly because, as you mentioned - this code really shouldn't be using GSR in any case.  No point spending time reworking the path-building logic now, if down the line we're gonna replace all this code anyways with a more appropriate SolrRequest impl.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java
gerlowskija: I didn't try it earlier because I wasn't sure whether this was related to any of the `/debug/foo` stuff above, but I'll give it a shot and update if the test passes...

gerlowskija: Hey it worked - good catch!

solr/core/src/java/org/apache/solr/packagemanager/PackageUtils.java
gerlowskija: Yeah, the 'magic boolean' is less than ideal for sure.  Done!

solr/core/src/java/org/apache/solr/packagemanager/PackageManager.java
epugh: Do we have JIRA tickets to track removing GSR?   I think my concern is that we forget about this follow up refactoring and it never happens.   Although, maybe we *would* fix it by just having the better v2 SolrRequest impl.  Let's keep moving forward!  And not have this PR grow too much!

solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java
epugh: yay!

epugh (approved):
Thanks for responding to my comments.  :shipit: !

solr/core/src/java/org/apache/solr/packagemanager/PackageManager.java
gerlowskija: We didn't, but we do now: https://issues.apache.org/jira/browse/SOLR-17136

PR #2232, opened 2024-01-30
SOLR-17137 Add configuration for SSL between Solr and Prometheus exporter.

eivinhb:
I have used the same env-var/jvm properties as expected in Solr. This is simply because we have Solr and Prometheus exporter running in same Pod in k8s.

https://issues.apache.org/jira/browse/SOLR-17137

# Description

Enable TLS in http2Solj-client while talking to Solr with ssl enabled.

# Solution

I have expanded the command line arguments with a param flag to enable ssl. I then pick up the same dockumented environment-variables that are expected by Solr to handle TLS.

# Tests
Tested locally against a Solr with TLS.

# Checklist

- [ x] I have reviewed the guidelines for [How to Contribute](https://github.com/apache/solr/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.
- [x ] I have created a Jira issue and added the issue ID to my pull request title.
- [x ] I have developed this patch against the `main` branch.
- [x ] I have run `./gradlew check`.


solr/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrExporter.java
```diff
@@ -103,6 +104,13 @@ public class SolrExporter {
   private static final String ARG_CREDENTIALS_HELP =
       "Specify the credentials in the format username:password. Example: --credentials solr:SolrRocks";
 
+  private static final String[] ARG_SSL_FLAGS = {"-ssl", "--ssl-enabled"};
+  private static final String ARG_SSL_METAVAR = "SSL_ENABLED";
+  private static final String ARG_SSL_DEST = "ssl_enabled";
+  private static final boolean ARG_SSL_DEFAULT = false;
+  private static final String ARG_SSL_HELP =
+      "Enable tls to Solr. Expects following env varables: SOLR_SSL_KEY_STORE, SOLR_SSL_KEY_STORE_PASSWORD, SOLR_SSL_TRUST_STORE, SOLR_SSL_TRUST_STORE_PASSWORD. Example: --ssl-enabled true";
```
epugh, line 112: is TLS normally capitalized?   "Enable TLS connection to Solr.".

epugh (changes requested):
This makes sense to me.   Can you please update the `solr-exporter.cmd` as well?   Also, it would be good to update the Ref Guide page so folks learn about this capability https://solr.apache.org/guide/solr/latest/deployment-guide/monitoring-with-prometheus-and-grafana.html.

Lastly, are you familiar with the `bats` tests?   https://github.com/apache/solr/tree/main/solr/packaging/test.   These are great for testing that the system properties/environment variables are being appropriately set.    In the not too distant future I want to refactor Prometheus exporter to use commons-cli like Solr itself does, and having some bats tests would be good...

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrExporter.java
draperunner: Good point. I have updated the text with your suggestion and also fixed the typo in "varables".

eivinhb:
@epugh I have amended the commit and added documentation and fixed an error in the cmd-file as well. 
I looked into the .bats tests, and I cant find any tests for the exporter at all. Did you want me to create those tests?

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrExporter.java
```diff
@@ -329,4 +355,8 @@ private static MetricsConfiguration loadMetricsConfiguration(String configPath)
       throw new RuntimeException(e);
     }
   }
+
```
epugh, line 358: Somewhere we have code that lets you look up these settings from system variable or environment variable or solr configuration file..  That may be only applicable to main Solr code base however...!

epugh (approved):
This looks good.   Longer term, I'd like to get the CLI to using commons-cli, since that is what the rest of Solr uses.   I could also imagine down the road moving this to SolrCLI, though that would require you to install Solr core jars to power the Solr CLI.   

What I would LOVE for THIS commit is a `test_prometheus.bats` file that would fire up a solr in SSL, and fire up the exporter and verify things work.  That would give us confidence that the shell scripts etc all work...

solr/prometheus-exporter/src/java/org/apache/solr/prometheus/exporter/SolrExporter.java
eivinhb: Yeah. The exporter has very little to do with Solr, except being shipped in the same docker (not slim) image.  I can't find any dependency to Solr core. Thus I belive that this has a completely separate configuration. 
That is why I opted to use the same name for environment variables that is used in Solr to ease the configuration of this in a production environment, eg. Kubernetes.

When I try and run `./gradlew integrationTests --tests test_ssl.bats` it fails for me (exit 1), giving a novice like me in Solr development little confidence to create more tests based on them.
`test_start_solr.bats` works fine though.
I might try and rebase on the latest from main again.

epugh: what environment are you in?   I suppose a `test_prometheus.bats` without SSL would be good enough to start with.

I've been mulling over, should the .bats tests be moved from somewhat "unit style for scripts" type tests to a more heavyweight integration test..  i.e, fire up a SSL encyrpted Solr with SSL Zookeeper with Basic Auth enabled, and then test various things like Prometheus exporter ;-).

HoustonPutman: There's a docker test that tests it without SSL, though it obviously could be added to bats as well

epugh: oh, then I guess that covers it right?

epugh: I see them now...   I wonder if would help if the docker tests used .bats?   looks like a lot of custom logic!

eivinhb: I have a MacPro M3. nodejs-errorlogs are always a pain imho.

eivinhb: `curl: (7) Failed to connect to localhost port 3128 after 0 ms: Couldn't connect to server`

Hmm. That is my proxy... I guess curl pics up on that.

eivinhb: `unset $HTTPS_PROXY` 
Now the ssl tests run fine. Let me see what I can do with prometheus exporter.

epugh:
I triggered a build, which should pick up the new .bats test.  Worth adding one more test that shows solr doing somethign and prometheus picking up that data?

eivinhb:
Yeah, I had plans for that, but had to go to bed. 
I have now updated the test to make it scrape solr and verify that i does. I also squashed the commits into one and added the jira id to commit head. It looks like that is a convention in your commit history.

epugh:
@eivinhb this looks great.  How do you want to be credited in `solr/CHANGES.txt`.  I'm thinking this is an improvement under 9.7...  "Enable SSL protected Solr to communicate with Prometheus exporter.  (Your name via Eric Pugh)".

eivinhb:
Great! My name is "Eivind Bergstøl". I think that what you suggest is nice. :) I might argue that it is the other way around:

`Enable Prometheus exporter to communicate with SSL protected Solr. (Eivind Bergstøl via Eric Pugh)`

eivinhb:
Since I forked this under an organisation there is no way to give you the permission to the PR. Should I edit the file in a separate commit or the same one?

epugh:
thank you fro this!

PR #2235, opened 2024-01-31
SOLR-16391: Tweak v2 'CREATECORE' API to be more REST-ful

iamsanjay:

# Description

V2(JAX-RS) Endpoints for core admin op "create" core.

# Solution

Updated the core creation endpoint to use JAX-RS library. The plan is to use the existing v2 enpoints `/api/cores`, however, for now different one is using `/api/cores/{coreName}/create`, meanwhile working on resolving issue related to former path. 

# Tests

1. Test valid create core command.
2. Test creating core with existent configSet.
3. Test creating core with non existent configSet.
4. Test report error simultaneously firing two creating core commands.

# Checklist


iamsanjay:
Not related to this change!
> java.lang.AssertionError: expected same:<org.apache.solr.request.SolrRequestInfo@922a9b2> was not:<null>
>    >         at __randomizedtesting.SeedInfo.seed([BF8CAE2E1F3CCF8F]:0)
>    >         at org.junit.Assert.fail(Assert.java:89)
>    >         at org.junit.Assert.failNotSame(Assert.java:829)
>    >         at org.junit.Assert.assertSame(Assert.java:772)
>    >         at org.junit.Assert.assertSame(Assert.java:783)
>    >         at org.apache.solr.request.TestSolrRequestInfo.lambda$testThreadPool$0(TestSolrRequestInfo.java:64)
>    >         at org.apache.solr.common.util.ExecutorUtil$MDCAwareThreadPoolExecutor.lambda$execute$0(ExecutorUtil.java:294)
>    >         at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
>    >         at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)

```
public void testThreadPool() throws InterruptedException {
    final SolrRequestInfo info =
        new SolrRequestInfo(
            new LocalSolrQueryRequest(h.getCore(), params()), new SolrQueryResponse());
    AtomicInteger counter = new AtomicInteger();

    SolrRequestInfo.setRequestInfo(info);
    ExecutorUtil.MDCAwareThreadPoolExecutor pool =
        new ExecutorUtil.MDCAwareThreadPoolExecutor(
            1, 1, 1, TimeUnit.SECONDS, new ArrayBlockingQueue<>(1));
    AtomicBoolean run = new AtomicBoolean(false);
    pool.execute(
        () -> {
          final SolrRequestInfo poolInfo = SolrRequestInfo.getRequestInfo();
          assertSame(info, poolInfo);
          info.addCloseHook(counter::incrementAndGet);
          run.set(true);
        });
    if (random().nextBoolean()) {
      pool.shutdown();
    } else {
      pool.shutdownNow();
    }
    SolrRequestInfo.clearRequestInfo();
    SolrRequestInfo.reset();

    pool.awaitTermination(1, TimeUnit.MINUTES);
    assertTrue(run.get());
    assertEquals("hook should be closed only once", 1, counter.get());
    assertNull(SolrRequestInfo.getRequestInfo());
  }
```

Above test failed during the build run, failed at 
```
assertSame(info, poolInfo);
```
I believe before the Runnable got executed, other line `SolrRequestInfo.clearRequestInfo();` which would pop SolrRequestInfo from the stack and When finally the Runnable started, it failed at  `assertSame(info, poolInfo)`.

iamsanjay:
The fix would be to move below code after the `pool.awaitTermination()` call but before the assert `assertNull(SolrRequestInfo.getRequestInfo())`

SolrRequestInfo.clearRequestInfo();
SolrRequestInfo.reset();

solr/core/src/java/org/apache/solr/handler/admin/api/CreateCore.java
```diff
@@ -0,0 +1,146 @@
...
+      SolrQueryResponse rsp) {
+    super(coreContainer, coreAdminAsyncTracker, req, rsp);
+    this.objectMapper = SolrJacksonMapper.getObjectMapper();
+  }
+
+  @Override
+  @PermissionName(PermissionNameProvider.Name.CORE_EDIT_PERM)
+  public SolrJerseyResponse createCore(
+      CreateCoreRequestBody createCoreRequestBody, String coreName) {
+    final Map<String, Object> requestBodyMap =
```
gerlowskija, line 69: [0] I wrote and have used `Utils.reflectToMap()` to do similar things for other Solr APIs.  I'm not an expert on either Jackson or reflection - so I can't really tell whether this is the same or different, better or worse than how `reflectToMap` works.  Do you happen to know?  If using Jackson directly in this way is better, maybe it'd make sense to reimplement `reflectToMap` to do the better thing and we could call it here.

Alternately if you just weren't aware of `Utils.reflectToMap`, maybe the quickest path forward is to just use reflectToWrite here, and then we can tackle evaluating and potentially reimplementing `Utils.reflectToMap` as a separate JIRA ticket later.

solr/core/src/test/org/apache/solr/handler/admin/api/CreateCoreApiTest.java
```diff
@@ -0,0 +1,153 @@
...
+          try {
+            createCore.createCore(createCoreRequestBody, CREATE_CORE_NAME);
+          } catch (Exception e) {
+            exp.set(e);
+          }
+        };
+
+    Thread firstCoreCreationThread = new Thread(coreCreationCmd);
+    Thread secCoreCreationThread = new Thread(coreCreationCmd);
+    firstCoreCreationThread.start();
```
gerlowskija, line 123: [0] Might simplify this code slightly to create a ExecutorService via `ExecutorUtil`, rather than directly managing threads?  Just a suggestion...

Also you're likely aware of this, but this setup doesn't guarantee that the core creations happen "simultaneously".  "First" might be finished by the time that "second" is started.  It's still a good test to have IMO, just might need a different name.

gerlowskija (changes requested):
Looks good overall, though I think the API syntax should be tweaked slightly.

Lmk what you think!

epugh:
Seems a shame to lose this one!

PR #2236, opened 2024-01-31
SOLR-17144: Do not keep an active thread per core

psalagnac:
https://issues.apache.org/jira/browse/SOLR-17144

# Description

Each core in Solr spawns a thread when the core is opened. If we have thousands of unused cores, it means we have thousands of waiting threads, wasting some resources (fd, memory).

# Solution

Instead of keeping this thread always active, span it on-demand with a TTL of 60 seconds.

# Tests

No new tests added for this. Just checked basic non-regression.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -2059,7 +2059,7 @@ public class SolrCore implements SolrInfoBean, Closeable {
   // Single threaded pool, with a time-to-keep of 60 seconds
   final ExecutorService searcherExecutor =
       ExecutorUtil.newMDCAwareSingleLazyThreadExecutor(
-          new SolrNamedThreadFactory("searcherExecutor"));
+          new SolrNamedThreadFactory("searcherExecutor"), 60L, TimeUnit.SECONDS);
   private int onDeckSearchers; // number of searchers preparing
   // Lock ordering: one can acquire the openSearcherLock and then the searcherLock, but not
   // vice-versa.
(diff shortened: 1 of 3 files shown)


dsmiley (approved):
Great!  Can you please add a CHANGES.txt entry for Solr 9.6 under optimizations?  Something like "SOLR-17144: Close searcherExecutor thread/core after 1 minute (Pierre Salagnac)

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -2056,8 +2056,10 @@ public UpdateHandler getUpdateHandler() {
   private final ArrayDeque<RefCounted<SolrIndexSearcher>> _searchers = new ArrayDeque<>();
   private final ArrayDeque<RefCounted<SolrIndexSearcher>> _realtimeSearchers = new ArrayDeque<>();
 
+  // Single threaded pool, with a time-to-keep of 60 seconds
   final ExecutorService searcherExecutor =
-      ExecutorUtil.newMDCAwareSingleThreadExecutor(new SolrNamedThreadFactory("searcherExecutor"));
+      ExecutorUtil.newMDCAwareSingleLazyThreadExecutor(
+          new SolrNamedThreadFactory("searcherExecutor"));
```
cpoerschke, line 2062: minor: mentioning the 60s in the comment shows/assumes `newMDCAwareSingleLazyThreadExecutor` implementation detail, could potentially avoid that (and allow other future use) by adding  arguments to the method/call e.g.

```suggestion
          new SolrNamedThreadFactory("searcherExecutor"), 60L, TimeUnit.SECONDS);
```

psalagnac: Thanks for the suggestion. Added parameters for time to keep.

PR #2237, opened 2024-02-01
SOLR-17140 - provide extensible query limit concept.

gus-asf:
https://issues.apache.org/jira/browse/SOLR-17140

This replaces SolrQueryTimeoutImpl.java with a much simpler SolrTimeLimit and provides a QueryLimits class to allow addition of additional types of limits. (see also https://issues.apache.org/jira/browse/SOLR-17138)


solr/core/src/java/org/apache/solr/search/SolrQueryTimeLimit.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.solr.request.SolrQueryRequest;
+
+public class SolrQueryTimeLimit implements QueryTimeout {
+
+  private final long timeoutAt;
+
+  /**
+   * Create an object to represent a time limit for the current request.
+   *
+   * @param req A sollr request that has a value for {@code timeAllowed}
```
sigram, line 33: typo `sollr`

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -210,6 +212,10 @@ public void addCloseHook(Closeable hook) {
     }
   }
 
+  public QueryLimits getLimits() {
+    return (QueryLimits) req.getContext().computeIfAbsent(LIMITS_KEY, (k) -> new QueryLimits(req));
```
sigram, line 216: This may throw NPE if used in a wrong place - HttpSolrCall.sendRemoteQuery() uses the other constructor for SolrRequestInfo, where req is null.

solr/core/src/java/org/apache/solr/search/QueryLimits.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+package org.apache.solr.search;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.lucene.index.QueryTimeout;
+import org.apache.solr.request.SolrQueryRequest;
+
+/**
+ * Represents the limitations on the query. These limits might be wall clock time, cpu time memory,
```
sigram, line 25: missing comma.

solr/core/src/java/org/apache/solr/search/QueryLimits.java
```diff
@@ -0,0 +1,46 @@
...
+ * or other resource limits. Exceeding any specified limit will cause {@link #shouldExit()} to
+ * return true the next time it is checked (it may be checked in either Lucene code or Solr code)
+ */
+public class QueryLimits implements QueryTimeout {
+  private final List<QueryTimeout> limits = new ArrayList<>();
```
sigram, line 30: We could specify the size, we know how many entries there can be at most.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -725,7 +730,7 @@ protected void search(List<LeafReaderContext> leaves, Weight weight, Collector c
       // So we need to make a new IndexSearcher instead of using "this".
       new IndexSearcher(reader) { // cheap, actually!
         void searchWithTimeout() throws IOException {
-          setTimeout(queryTimeout.makeLocalImpl());
+          setTimeout(requestInfo.getLimits());
           super.search(leaves, weight, collector); // FYI protected access
           if (timedOut()) {
             throw new TimeAllowedExceededFromScorerException("timeAllowed exceeded");
```
sigram, line 736: Hmm, perhaps we should change this message to "allowed limit exceeded". Which suggests that we should be able to tell user what limit has been exceeded if multiple limits have been specified... We could add a utility method to `QueryLimits` to format a string telling what kind of limit has been tripped, based on the current values of `shouldExit` of each impl.

solr/core/src/java/org/apache/solr/search/SolrQueryTimeLimit.java
```diff
@@ -0,0 +1,59 @@
...
+   *     should be validated with {@link #hasTimeLimit(SolrQueryRequest)} prior to constructing this
+   *     object
+   */
+  public SolrQueryTimeLimit(SolrQueryRequest req) {
+    // reduce by time already spent
+    long reqTimeAllowed = req.getParams().getLong(CommonParams.TIME_ALLOWED, -1L);
+
+    if (reqTimeAllowed == -1L) {
+      throw new IllegalArgumentException(
+          "Check for limit with hasTimeLimit(req) before creating a SolrQuerTimeLimit");
```
sigram, line 44: typo

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
gus-asf: Yeah I need to understand better how req is populated.

solr/core/src/java/org/apache/solr/search/QueryLimits.java
```diff
@@ -0,0 +1,90 @@
...
+  public String limitStatusMessage() {
+    StringBuilder sb = new StringBuilder();
+    boolean first = true;
+    for (QueryTimeout limit : limits) {
+      if (first) {
+        first = false;
+        sb.append("Query limits:");
+      }
+      sb.append("[");
+      sb.append(limit.getClass().getName());
```
sigram, line 75: Maybe getShortName() ?

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -210,6 +217,12 @@ public void addCloseHook(Closeable hook) {
     }
   }
 
+  public QueryLimits getLimits() {
+    return req == null
+        ? QueryLimits.NONE
+        : (QueryLimits) req.getContext().computeIfAbsent(LIMITS_KEY, (k) -> new QueryLimits(req));
```
sigram, line 223: I think this is still incorrect... Since it uses the latest SRI from the stack it will completely ignore the already existing QueryLimits created at the bottom of the stack, and create a new instance to track just this part of the processing - whereas we want to track the resource consumption across the whole end-to-end request processing, beginning with the first SRI (at the bottom of the stack) allocated to this query.

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -75,6 +77,9 @@ public static void setRequestInfo(SolrRequestInfo info) {
     } else if (stack.size() > MAX_STACK_SIZE) {
       assert false : "SolrRequestInfo Stack is full";
       log.error("SolrRequestInfo Stack is full");
+    } else if (!stack.isEmpty() && info.req != null) {
+      // if req is null limits will be an empty instance with no limits anyway.
```
sigram, line 81: Maybe add a note that all SRI-s on the stack inherit the first instance of the limits.

solr/core/src/java/org/apache/solr/search/QueryLimits.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.lucene.index.QueryTimeout;
+import org.apache.solr.request.SolrQueryRequest;
+
+/**
+ * Represents the limitations on the query. These limits might be wall clock time, cpu time, memory,
+ * or other resource limits. Exceeding any specified limit will cause {@link #shouldExit()} to
+ * return true the next time it is checked (it may be checked in either Lucene code or Solr code)
+ */
+public class QueryLimits implements QueryTimeout {
+  private final List<QueryTimeout> limits =
```
dsmiley, line 32: Where do more limits get added?  I only see the time one.

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
dsmiley: agreed

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -210,6 +216,32 @@ public void addCloseHook(Closeable hook) {
...
 
+  /**
+   * This call creates the QueryLimits object and any required implementations of {@link
+   * org.apache.lucene.index.QueryTimeout}. Any code before this call will not be subject to the
+   * limitations set on the request. Note that calling {@link #getLimits()} has the same effect as
+   * this method.
+   *
+   * @see #getLimits()
+   */
+  public void initQueryLimits() {
```
dsmiley, line 227: why public; seems internal

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -972,7 +978,8 @@ private DocSet getAndCacheDocSet(Query query) throws IOException {
     }
 
     DocSet answer;
-    if (SolrQueryTimeoutImpl.getInstance().isTimeoutEnabled()) {
+    SolrRequestInfo requestInfo = SolrRequestInfo.getRequestInfo();
+    if (requestInfo != null && requestInfo.getLimits().isTimeoutEnabled()) {
```
dsmiley, line 982: The previous pattern had the nice characteristic that it's a one-liner.  We could put the use of SolrRequestInfo on a limits class convenience method?  Not a big deal anyway.

dsmiley (commented):
Overall -- looks nice!  I like tying it to SolrRequestInfo.  I'm not sure it needs to be baked into SRI but I can see you needed the inheritance characteristic modification so okay.

At Salesforce, we've extended this concept to also apply to indexing and one could imagine maybe someone using it in other cases.  The "Query" in the name is unfortunate... can we call this RequestLimits instead of QueryLimits?

I'd like to imagine one day someone improving query cancellation as currently implemented be refactored into using this framework.  As such it wouldn't be a "limit" but I suppose the wording is fine as it's the most common implementation.  Really; we're supporting cancellation of a query.  Maybe it's a "limit" or something else.  And the basic framework can work on anything; doesn't even need to be a "query".

solr/core/src/java/org/apache/solr/search/QueryLimits.java
```diff
@@ -0,0 +1,90 @@
...
+   */
+  public QueryLimits(SolrQueryRequest req) {
+    if (hasTimeLimit(req)) {
+      limits.add(new SolrQueryTimeLimit(req));
+    }
+  }
+
+  @Override
+  public boolean shouldExit() {
+    return limits.stream().anyMatch(QueryTimeout::shouldExit);
```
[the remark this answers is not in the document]
dsmiley: or support a chain implementation similar to URPs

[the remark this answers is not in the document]
gus-asf: Sure, will do, though I would expect compiler/jvm to optimize either.

solr/core/src/java/org/apache/solr/search/QueryLimits.java
gus-asf: https://issues.apache.org/jira/browse/SOLR-17141  && https://issues.apache.org/jira/browse/SOLR-17150

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
gus-asf: had been called elsewhere as getLimits(), agree can be private now.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
gus-asf: in some cases there's an assert for requestInfo != null, rather than a null check but that assert is not appropriate in all cases.

gus-asf:
> At Salesforce, we've extended this concept to also apply to indexing and one could imagine maybe someone using it in other cases. The "Query" in the name is unfortunate... can we call this RequestLimits instead of QueryLimits?
> 

This hinges on and (and implements) QueryTimeout so for the moment it's pretty solidly query focused. When we expand the functionality the name can change then?

gus-asf:
Feedback addressed, tests pass locally and crave runs in past that have failed on crave have been ZK or recovery related... This is holding up other work so I'm merging it.

PR #2239, opened 2024-02-01
SOLR-17146: Add DelegatingBackupRepository and alternative checksum verification

bruno-roustant:
Description at https://issues.apache.org/jira/browse/SOLR-17146

1- Add an optional initialization argument to disable the checksum verification (enabled by default) in each BackupRepository implementation.

2- Introduce an abstract DelegatingBackupDirectory to extend when we need to perform different/additional logic on top of a delegate BackupDirectory. For example, verify the checksum differently. Also, the DelegatingBackupDirectory could inject some initialization arguments when the BackupRepositoryFactory creates its delegate.
Also make the existing TrackingBackupRepository (for tests) extend DelegatingBackupDirectory.

3- Add a test case to AbstractBackupRepositoryTest to ensure all implementations support disabling the checksum verification. And add a test for LocalFileSystemDirectory (not tested currently) that also extends AbstractBackupRepositoryTest.


solr/core/src/java/org/apache/solr/core/backup/repository/LocalFileSystemRepository.java
```diff
@@ -97,7 +100,7 @@ public URI resolve(URI baseUri, String... pathComponents) {
   public void createDirectory(URI path) throws IOException {
     Path p = Path.of(path);
     if (!Files.exists(p, LinkOption.NOFOLLOW_LINKS)) {
-      Files.createDirectory(p);
```
bruno-roustant, line 100: The new LocalFileSystemRepositoryTest revealed a small bug here :)
The method contract is to create all the directories needed, including the parents.

solr/core/src/java/org/apache/solr/core/backup/repository/BackupRepositoryUtil.java
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * Copy a file from a source {@link Directory} to a destination {@link Directory} without
+   * verifying the checksum.
+   *
+   * @param sourceDir The source directory hosting the file to be copied.
+   * @param sourceFileName The name of the file to be copied.
+   * @param destDir The destination directory to copy the file to.
+   * @param destFileName The name of the copied file at destination.
+   */
+  public static void copyFileNoChecksum(
```
gerlowskija, line 38: [0] This doesn't seem too radically different from other methods that live on `BackupRepository` itself as "default" interface implementations or static methods...is there a particular reason you have this live in a separate 'util' class? (Maybe it's easier to test this way?)

I don't have an opinion on where this lives one way or another; I just want to make sure I understand how you've got things laid out...

solr/modules/hdfs/src/java/org/apache/solr/hdfs/backup/repository/HdfsBackupRepository.java
```diff
@@ -204,6 +213,17 @@ public void copyIndexFileFrom(
     }
   }
 
+  @Override
```
gerlowskija, line 216: [-1] It's a downer that this implementation has to live in pretty much every BackupRepository implementation.  (Currently I think this is forced by the visibility of `shouldVerifyChecksum`, which is read by each individual BR implementation)

The BR interface has an `init(NamedList)` method - maybe we could read `verifyChecksums` in that base method, and avoid duplicating a bunch of this stuff on each implementation?

If it's not possible, fair enough, but it's worth trying before we commit this IMO

gerlowskija (changes requested):
I like the overall approach, but had concerns about duplication in some of the BackupRepository implementations (see comments inline).

That said if changes aren't possible there, I def understand.  Happy to approve as soon as you've at least replied to my comments there, and there's some docs for the new property.

solr/core/src/java/org/apache/solr/core/backup/repository/BackupRepositoryUtil.java
bruno-roustant: I thought this method was enough different from the others in the BR interface to not have its place there. And it is static, so I don't think it could be overridden. But now that you suggest, I'll move it inside BR.

solr/core/src/java/org/apache/solr/core/backup/repository/FilterBackupRepository.java
```diff
@@ -0,0 +1,159 @@
...
+import java.net.URI;
+import java.util.Collection;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.IOContext;
+import org.apache.lucene.store.IndexInput;
+import org.apache.solr.common.util.NamedList;
+import org.apache.solr.core.backup.Checksum;
+
+/** Delegates to another {@link BackupRepository}. */
+public class FilterBackupRepository implements BackupRepository {
```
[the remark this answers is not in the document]
bruno-roustant: "Filter" means the same pattern as "Delegating" here. Actually there are more "Filter" names in Lucene than "Delegating", that's why I kept it, but I find "Delegating" a bit clearer, so I'll rename "Delegating".

solr/modules/hdfs/src/java/org/apache/solr/hdfs/backup/repository/HdfsBackupRepository.java
bruno-roustant: I agree, this method should be shared, but BackupRepository is an interface, so I don't see how we share the sholdVerifyCheckum field. Unless we introduce an abstract class in-between, something like AbstractBackupRepository, it that what you mean?

solr/core/src/java/org/apache/solr/core/backup/repository/LocalFileSystemRepository.java
```diff
@@ -153,17 +142,6 @@ public void copyIndexFileFrom(
     }
   }
 
-  @Override
```
gerlowskija, line 153: [+1] This looks awesome; thanks for changing this!

solr/solr-ref-guide/modules/deployment-guide/pages/backup-restore.adoc
```diff
@@ -422,7 +433,7 @@ LocalFileSystemRepository accepts the following configuration option:
 |===
 +
 A valid file path (accessible to Solr locally) to use for backup storage and retrieval.
-Used as a fallback when user's don't provide a `location` parameter in their Backup or Restore API commands
+Used as a fallback when user's don't provide a `location` parameter in their Backup or Restore API commands.
```
gerlowskija, line 436: [0] There shouldn't be any apostrophe in "user's".

solr/solr-ref-guide/modules/deployment-guide/pages/backup-restore.adoc
```diff
@@ -401,6 +401,17 @@ Any children under the `<repository>` tag are passed as additional configuration
...
+By default, all the repository implementations verify the integrity of the index files before they are copied to the destination. However, it is possible to disable this integrity check by setting the optional configuration property `verifyChecksum`.
+
+`verifyChecksum`::
++
+[%autowidth,frame=none]
+|===
+|Optional |Default: `true`
+|===
++
+Defines whether the backup repository should check the index files integrity before copying them to the destination. Set `false` to disable the checksum verification, for example to verify the integrity in a different way.
```
gerlowskija, line 413: [0] Is it worth calling out the encryption scenario explicitly here?  I can imagine users reading this and wondering why they would ever want to disable such a safety check...

(If we're intentionally NOT mentioning encryption because it's a 'solr-sandbox' thing, I guess I understand that as well.  Just figured I'd check)

gerlowskija (approved):
LGTM - left a few comments on your docs, but nothing that should hold up merging if you happen to disagree.

bruno-roustant:
Thanks @gerlowskija for the review!

ozlerhakan:
Hi there! I'm wondering how this feature can be disabled in a collection managed by the Solr operator @HoustonPutman? Since the current Solr operator does not have this initialization parameter in `SolrCloud.spec.backupRepositories`, my first guess is to rely on a solr.xml file with the necessary backup definitions and use IAM in the SolrCloud object, right?

PR #2240, opened 2024-02-02
SOLR-17066: Switch HttpSolrClient away from coreURLs, pt 3

gerlowskija:

# Description

SOLR-17066 introduced a `withDefaultCollection` builder method that can be used in lieu of providing a core URL as a "base".  (Core URLs are not ideal as they prevent clients from making core-agnostic requests.)

However, Solr's own client usage still needs modified to use this new builder method.

# Solution

# Tests

This PR tackles this for all remaining usages of HttpSolrClient in Solr's codebase, including solrj's tests and the test-framework module.

# Checklist


solr/core/src/test/org/apache/solr/cloud/TestCloudDeleteByQuery.java
```diff
@@ -157,18 +157,17 @@ public static void createMiniSolrCloudCluster() throws Exception {
       assertNotNull("could not find URL for " + shardName + " replica", passiveUrl);
 
       if (shardName.equals("shard1")) {
-        S_ONE_LEADER_CLIENT = getHttpSolrClient(leaderUrl + "/" + COLLECTION_NAME + "/");
-        S_ONE_NON_LEADER_CLIENT = getHttpSolrClient(passiveUrl + "/" + COLLECTION_NAME + "/");
+        S_ONE_LEADER_CLIENT = getHttpSolrClient(leaderUrl, COLLECTION_NAME);
+        S_ONE_NON_LEADER_CLIENT = getHttpSolrClient(passiveUrl, COLLECTION_NAME);
       } else if (shardName.equals("shard2")) {
-        S_TWO_LEADER_CLIENT = getHttpSolrClient(leaderUrl + "/" + COLLECTION_NAME + "/");
```
epugh, line 163: I love not having to think about "do I need a trailing / or not"!

solr/core/src/test/org/apache/solr/update/TestInPlaceUpdatesDistrib.java
```diff
@@ -1146,7 +1146,9 @@ private void reorderedDBQsResurrectionTest() throws Exception {
       // the resurrection can happen from there (instead of the leader)
       update.setParam(
           DistributedUpdateProcessor.DISTRIB_FROM,
-          ((HttpSolrClient) NONLEADERS.get(1)).getBaseURL());
+          ((HttpSolrClient) NONLEADERS.get(1)).getBaseURL()
+              + "/"
+              + NONLEADERS.get(1).getDefaultCollection());
```
epugh, line 1151: this actually helps you understand what we are doing in the DISTRIB_FROM by seeing the call to `getDefaultCollection`.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttpSolrClient.java
```diff
@@ -122,10 +122,14 @@ protected HttpSolrClient makeSolrClient(String server) {
...
-        httpSolrClientBuilder.withBaseSolrUrl(server).withHttpClient(httpClient);
-        httpSolrClientBuilder.withConnectionTimeout(connectionTimeoutMillis, TimeUnit.MILLISECONDS);
-        httpSolrClientBuilder.withSocketTimeout(soTimeoutMillis, TimeUnit.MILLISECONDS);
-
+        httpSolrClientBuilder
+            .withBaseSolrUrl(server)
+            .withHttpClient(httpClient)
+            .withConnectionTimeout(connectionTimeoutMillis, TimeUnit.MILLISECONDS)
+            .withSocketTimeout(soTimeoutMillis, TimeUnit.MILLISECONDS);
+        if (defaultCollection != null) {
```
epugh, line 130: would this have worked if you just blinding passed in `defaultCollection` and it happened to be null?   I/e with out this check?  Though maybe it's clearer to see the logic with the excplicit check...

solr/test-framework/src/java/org/apache/solr/SolrTestCaseJ4.java
```diff
@@ -2655,14 +2656,25 @@ public static HttpClient getHttpClient(String url) {
...
+
   /**
    * This method creates a basic HttpSolrClient. Tests that want to control the creation process
    * should use the {@link org.apache.solr.client.solrj.impl.Http2SolrClient.Builder} class directly
+   *
+   * @param url the base URL of a Solr node. Should <em>not</em> include a collection or core name.
+   * @param defaultCoreName the name of a core that the created client should default to when making
+   *     core-aware requests
    */
   public static HttpSolrClient getHttpSolrClient(String url, String defaultCoreName) {
```
epugh, line 2679: I wonder if a check for the situation where the defaultCoreName is in the url?   and throw an exception?

epugh (commented):
LGTM.   Nothing jumped out as a "must fix".    Huge amount of work.   Thanks for the docs that are added too!

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttpSolrClient.java
gerlowskija: > would this have worked if you just blindly passed in `defaultCollection` and it happened to be null?

Yes and no.

The `withDefaultCollection` setter is resilient enough to handle null-values.  So in the general case that'd be just fine.

But the reason we check explicitly in this case is because a user might have specified a default-coll on their httpSolrClientBuilder (but not on the LB Builder directly).  In that scenario we'd be overwriting the value they provided with 'null'.

You'll notice this is the same pattern we follow with request-writer and a few other objects here - we're essentially just doing out best to make sure that we don't accidentally clobber values that were provided on the HttpSolrClientBuilder.

solr/test-framework/src/java/org/apache/solr/SolrTestCaseJ4.java
gerlowskija: On `branch_9x`, afaict, there's no correct way to detect when a URL contains a core name.  I actually caused a bunch of test failures [earlier this week](https://issues.apache.org/jira/browse/SOLR-17066?focusedCommentId=17814783&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17814783) by not realizing this.  The root of the problem is that 9.x releases allow users to use arbitrary hostContext's or context-root's or whatever - in such a world there's no way to look at the URL and determine a collection name from anything else.

That said - on `main` where the context-root is "locked in" I 100% agree we should have some check for this.  But IMO it should go in the client ctors directly, so that it gets triggered in **all** cases.

solr/core/src/test/org/apache/solr/cloud/AliasIntegrationTest.java
```diff
@@ -968,8 +968,7 @@ private void searchSeveralWays(
       // HttpSolrClient
       JettySolrRunner jetty = cluster.getRandomJetty(random());
       if (random().nextBoolean()) {
-        try (SolrClient client =
-            getHttpSolrClient(jetty.getBaseUrl().toString() + "/" + collectionList)) {
+        try (SolrClient client = getHttpSolrClient(jetty.getBaseUrl().toString(), collectionList)) {
```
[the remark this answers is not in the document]
epugh: Any chance on selling you on this idea ;-)???

epugh:
This all LGTM....

gerlowskija:
> gradlew :solr:core:test --tests "org.apache.solr.cloud.TestPrepRecovery.testLeaderNotResponding" -Ptests.jvms=96

> [...] -Ptests.jvms=96 [...]

🤯 Man, easy to forget sometimes that these "Crave.io" machines are just absolute monsters.

Thanks for pointing out the failure @iamsanjay .  According to fucit, TestPrepRecovery is pretty flaky - it's been failing on between 5 and 10% of runs since November or so.  So my guess is that it's not directly caused by anything in this PR.  But I'll kick off some 'beast' runs on it locally and try to find out.  Will update this ticket shortly.

(If you're not familiar with "[fucit](https://fucit.org/solr-jenkins-reports/failure-report.html)", it's a really helpful test-aggregator that Chris Hostetter (who goes by "Hoss man") runs for us.  It gathers the test-reports from a bunch of Jenkins and other CI builds, and keeps track of how healthy or unhealthy each individual test has been historically.  It links out to the Jenkins builds for each failure so you can look at specific failures. 
 Really really helpful stuff if you haven't seen it.)

PR #2242, opened 2024-02-02
Demonstrate a test framework bug

jdyer1:
There seems to be a subtle bug in our test framework where it is sometimes possible for an unauthorized user to add document(s). This happens when both the client and server run in the same JVM and CloudSolrClient is set up to use Parallel Updates on the client (as with `MiniSolrCloudCluster`).   This is a potential test-only bug, and to my knowledge no existing tests are affected.  Yet if left unsolved, this may cause misleading test results in the future.

CloudSolrClient sets up a `MDCAwareThreadPoolExecutor` to perform parallel updates on the client JVM.  Prior to executing a parallel update, `ExecutorUtil#setServerThreadFlag` is set on the Client JVM's thread(s) from the pool.  Such a flag normally would not matter, because this happens on the client JVM and there this flag has no meaning.  But our unit tests can run both client and server in the same JVM.  This confuses `PKIAuthenticationPlugin` (server-side) which sees it as an inter-node request.

Although this can allow an unauthorized user to add a document, it does happen on commit.  `CloudSolrClient` does not use parallel updates to request a commit and these always fail when appropriate.  This could lead to an unreliable test case that passes or fails only when run slowly enough for an autocommit to finish before comparing documents.

This PR contains a (demonstration-only) failing unit test.  See `TestAuthorizationWhenClientInSameJvm`.  This PR adds a flag to `MiniSolrCloudCluster` to let tests disable parallel updates on the client.  If this is deemed an adequate solution, it would be best to default this flag to `false` with a comment warning of the downsides.

This issue touches on our efforts to reduce the size of the SolrJ client.  It seems to me that because we also use SolrJ for inter-node requests that we've perhaps introduced code on this client library that rightfully belong in the server only.  Specifically, it seems dubious that we include `ExecutorUtil` in the client library.

### DIFF
--- a/solr/core/src/test/org/apache/solr/security/TestAuthorizationWhenClientInSameJvm.java
+++ b/solr/core/src/test/org/apache/solr/security/TestAuthorizationWhenClientInSameJvm.java
@@ -1,0 +1,191 @@
+package org.apache.solr.security;
+
+import java.util.Arrays;
+import java.util.Map;
+import org.apache.solr.client.solrj.request.CollectionAdminRequest;
+import org.apache.solr.client.solrj.request.QueryRequest;
+import org.apache.solr.client.solrj.request.UpdateRequest;
+import org.apache.solr.client.solrj.response.QueryResponse;
+import org.apache.solr.cloud.SolrCloudTestCase;
+import org.apache.solr.common.params.MapSolrParams;
+import org.apache.solr.common.util.Utils;
+import org.junit.After;
+import org.junit.Test;
+
+/**
+ * This tries to ensure that our test framework reliably mimics the real world with regard to
+ * Authorization, even though our tests use the same JVM for both client and server.
+ *
+ * <p>NOCOMMIT: This test takes a few seconds to run, adds little value, and is for demonstration
+ * purposes only. We could reverse the order in CloudAuthStreamTest#testSanityCheckAuth and get the
+ * same coverage.
+ */
+public class TestAuthorizationWhenClientInSameJvm extends SolrCloudTestCase {
+
+  @Test
+  public void testReadOnlyUserCannotUpdate() {
+    // This fails because of a test framework bug.
+    // When CloudSolrClient is used with parallel updates turned on,
+    // with the client and server both running in the same JVM
+    // The PKI Authentication Plugin gets confused and thinks our update
+    // it is an inter-node request.
+
+    // Setting up the cluster with parallel updates enabled (the default)
+    setupCluster(true);
+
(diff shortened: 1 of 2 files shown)


janhoy:
Thanks for being proactive on this. I think too it is scary that SolrJ is hybrid client/server code that risks containing server specific code and dependencies. Your workaround of skipping parallel updates for auth tests is ok to support and document. It sounds like a harder task to re-architect the inter-node traffic auth to be less "magic".

jdyer1:
Thank you for looking at this @janhoy.  My belief here is that if server-specific code in `ExecutorUtil` was moved out of SolrJ, this would completely solve the problem.  In particular I would like to propose moving the `isServerPool` ThreadLocal to a server-side class.  This would mean that `MDCAwareThreadPoolExecutor` also would need to have both server-side and client-side versions.

epugh:
@jdyer1 I'd love to see auth testing used a lot more across Solr and this seems like the kind of thing that if not fixed would trip people up in the future.   What can I do to help?   Your proposal seems reasonable, but maybe an email to the dev list to get more eyes on it?

jdyer1:
@epugh As mentioned earlier in the PR, we can paper over this by disabling Parallel Updates for auth tests.  But the real problem in my opinion is we have allowed SolrJ to become a server-side library that clients also use.  I would like to move `MDCAwareThreadPoolExecutor` into a server-side jar and replace it with a client-specific Executor in SolrJ.

As for raising this on the Dev list, I always thought anyone interested in participating in decisions like this would also be subscribed to PR updates.  If you or others think my proposal here has merit, I can try and work up a PR for it so it can be evaluated further.

epugh:
They *should* be subscribed to the PR, but sometimes emailing dev@ get's some more eyes, especialy if you aren't sure which direction to proceed or even if you should proceed.   Your reasoning makes sense to me but again I'm not a deep Java person, so the ramifications of moving `MDCAwareThreadPoolExecutor` are beyond me ;-).

dsmiley:
> My belief here is that if server-specific code in ExecutorUtil was moved out of SolrJ, this would completely solve the problem. In particular I would like to propose moving the isServerPool ThreadLocal to a server-side class.

+1

But I don't want lots of code duplication for `MDCAwareThreadPoolExecutor`.  Maybe the isServerPool determination should be redesigned.  Namely, The creator thread of a MDCAwareThreadPoolExecutor; if it's a "server thread", then so is the thread pool and thus its threads likewise.  And startup CoreContainer as a server thread too.

PR #2244, opened 2024-02-05
SOLR-17141: Create CpuQueryLimit implementation

sigram:
Details in Jira. This builds on top of changes in #2237.


solr/core/src/test-files/solr/configsets/query-limits/conf/schema.xml
dsmiley, line 1: Please don't add more test configs unless we truly need it!  See `org.apache.solr.SolrTestCaseJ4#copyMinConf(java.io.File)` and `HideStackTraceTest` for an example of how to use it.

solr/core/src/test/org/apache/solr/search/ExpensiveSearchComponent.java
```diff
@@ -0,0 +1,58 @@
...
+package org.apache.solr.search;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import org.apache.lucene.tests.util.LuceneTestCase;
+import org.apache.lucene.tests.util.TestUtil;
+import org.apache.solr.handler.component.ResponseBuilder;
+import org.apache.solr.handler.component.SearchComponent;
+
+public class ExpensiveSearchComponent extends SearchComponent {
```
dsmiley, line 26: Looking forward to seeing this on https://solr.cool/ :-)

solr/core/src/test/org/apache/solr/search/TestCpuQueryTimeLimit.java
```diff
@@ -0,0 +1,126 @@
...
+import org.apache.solr.cloud.CloudUtil;
+import org.apache.solr.cloud.MiniSolrCloudCluster;
+import org.apache.solr.cloud.SolrCloudTestCase;
+import org.apache.solr.util.ThreadCpuTime;
+import org.junit.Assume;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TestCpuQueryTimeLimit extends SolrCloudTestCase {
```
dsmiley, line 33: Is this really SolrCloud related?  Notice how TestCpuTimeSearch does some distributed search yet doesn't need SolrCloud to do it.  I'm trying to keep our tests leaner.

dsmiley (changes requested):
Thanks for working on this.  Seems useful reporting or API uses cases but maybe not for a SERP with a tight timeAllowed.

solr/core/src/java/org/apache/solr/search/QueryLimits.java
```diff
@@ -41,11 +43,15 @@ private QueryLimits() {}
...
    * @param req the current SolrQueryRequest.
+   * @param threadCpuTime current thread CPU time monitor.
    */
-  public QueryLimits(SolrQueryRequest req) {
+  public QueryLimits(SolrQueryRequest req, ThreadCpuTime threadCpuTime) {
     if (hasTimeLimit(req)) {
       limits.add(new SolrQueryTimeLimit(req));
     }
+    if (hasCpuLimit(req)) {
+      limits.add(new CpuQueryTimeLimit(req, threadCpuTime));
```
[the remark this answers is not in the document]
sigram: It's still not too late to rename the whole family, for example:
* `SolrQueryTimeLimit` -> `TimeAllowedLimit`
* `CpuQueryTimeLimit` -> `CpuAllowedLimit`
* `MemQueryLimit` -> `MemAllowedLimit`

WDYT?

solr/core/src/test/org/apache/solr/search/TestCpuQueryTimeLimit.java
sigram: I'm not sure if this pseudo-cluster covers the same code paths when making distributed calls.

dsmiley: I wouldn't choose SolrCloudTestCase merely because you are not sure of this.  Remember that distributed-search existed before SolrCloud!

dsmiley: Thinking of the new SolrClientTestRule stuff; it'd be cool if basic sharded search/indexing could be done to allow a test like the one here to switch between Standalone & SolrCloud trivially.  I'm hoping to get there but the work isn't done yet.  FYI: https://issues.apache.org/jira/browse/SOLR-11872

solr/core/src/java/org/apache/solr/search/QueryLimits.java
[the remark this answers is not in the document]
gus-asf: Actually, I like the idea of tracking the parameter names.

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -250,17 +253,17 @@ public void handleRequest(SolrQueryRequest req, SolrQueryResponse rsp) {
...
-        Optional<Long> cpuTime = cpuStats.getCpuTimeMs();
+      if (publishCpuTime) {
+        Optional<Long> cpuTime = threadCpuTime.getCpuTimeMs();
         if (cpuTime.isPresent()) {
           // add CPU_TIME if not already added by SearchHandler
           NamedList<Object> header = rsp.getResponseHeader();
           if (header != null) {
-            if (header.get(ThreadStats.CPU_TIME) == null) {
-              header.add(ThreadStats.CPU_TIME, cpuTime.get());
+            if (header.get(ThreadCpuTime.CPU_TIME) == null) {
```
gus-asf, line 262: Is this correct? When would this get called twice for the same `rsp`? if so it appears that we would ignore a second invocation (but the `addToLog` below doesn't ignore it so they will differ?

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -250,17 +253,17 @@ public void handleRequest(SolrQueryRequest req, SolrQueryResponse rsp) {
       long elapsed = timer.stop();
       metrics.totalTime.inc(elapsed);
 
-      if (cpuStats != null) {
-        Optional<Long> cpuTime = cpuStats.getCpuTimeMs();
+      if (publishCpuTime) {
+        Optional<Long> cpuTime = threadCpuTime.getCpuTimeMs();
         if (cpuTime.isPresent()) {
           // add CPU_TIME if not already added by SearchHandler
           NamedList<Object> header = rsp.getResponseHeader();
```
gus-asf, line 260: I see lots of places where this is checked for null, but I'm wondering why it's allowed to return null (not a topic  for this ticket however)

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -250,17 +253,17 @@ public void handleRequest(SolrQueryRequest req, SolrQueryResponse rsp) {
       long elapsed = timer.stop();
       metrics.totalTime.inc(elapsed);
 
-      if (cpuStats != null) {
-        Optional<Long> cpuTime = cpuStats.getCpuTimeMs();
+      if (publishCpuTime) {
```
gus-asf, line 256: It looks like the present implementation cannot avoid logging cpu time if cpuAllowed is used?

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -622,8 +622,8 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
       } while (nextStage != Integer.MAX_VALUE);
 
       if (publishCpuTime) {
-        rsp.getResponseHeader().add(ThreadStats.CPU_TIME, totalShardCpuTime);
-        rsp.addToLog(ThreadStats.CPU_TIME, totalShardCpuTime);
+        rsp.getResponseHeader().add(ThreadCpuTime.CPU_TIME, totalShardCpuTime);
```
gus-asf, line 625: No null check here?

gus-asf (changes requested):
I think the logging and limiting features are too tightly coupled still... It should be possible to set a timeout without logging the cpu time or vice-a-versa

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -76,7 +77,7 @@ public abstract class RequestHandlerBase
 
   private PluginInfo pluginInfo;
 
-  protected boolean publishCpuTime = Boolean.getBoolean(ThreadStats.ENABLE_CPU_TIME);
+  protected boolean publishCpuTime = Boolean.getBoolean(ThreadCpuTime.ENABLE_CPU_TIME);
```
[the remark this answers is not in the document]
sigram: They are not. It's true that `ThreadCpuTimer` will be created on first call to `SolrRequestInfo.getThreadCpuTimer()`, whoever first calls it (either due to this sysprop being true, or due to the `cpuAllowed` query param). However, after that the logging will happen only if the sysprop is true, again regardless of the `cpuAllowed` query param, and the limiting will happen regardless of the logging sysprop.

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
sigram: I believe the idea was to make sure the total CPU time is added to response from any handler, not just the `SearchHandler` where it's calculated differently as a sum of per-shard times.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
sigram: This can never be null due to the assignment that depends on the same condition in the if() around line 221.

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
sigram: Yeah, I don't like this either. `ThreadCpuTimer.isSupported()` is the only condition for this to return false and it should be enough to check it.

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
sigram: I don't think that's the case - see above comment.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
sigram: Do you mean for the responseHeader?

solr/core/src/test/org/apache/solr/search/TestCpuQueryTimeLimit.java
sigram: I changed the latest version of the test to use SolrJettyTestRule.

solr/core/src/test/org/apache/solr/search/TestCpuAllowedLimit.java
```diff
@@ -0,0 +1,182 @@
...
+                "class=\"solr.SearchHandler\">",
+                "class=\"solr.SearchHandler\">\n"
+                    + "    <arr name=\"first-components\">\n"
+                    + "      <str>expensiveSearchComponent</str>\n"
+                    + "    </arr>\n"));
+    return configSet.resolve("conf");
+  }
+
+  @Test
+  public void testDistribLimit() throws Exception {
```
dsmiley, line 103: no distributed-search  (e.g. via "shards")?  Not mandatory to test this but if I thought that's why you wanted SolrCloudTestCase.

solr/core/src/java/org/apache/solr/search/QueryLimits.java
```diff
@@ -16,14 +16,13 @@
...
+import static org.apache.solr.search.TimeAllowedLimit.hasTimeLimit;
 
 import java.util.ArrayList;
 import java.util.List;
 import org.apache.lucene.index.QueryTimeout;
 import org.apache.solr.request.SolrQueryRequest;
-import org.apache.solr.util.ThreadCpuTime;
 
 /**
  * Represents the limitations on the query. These limits might be wall clock time, cpu time, memory,
```
dsmiley, line 28: Maybe this becomes "RequestLimits" or I'm pushing my luck with my suggestions :-)

dsmiley (commented):
Nice work!

solr/core/src/java/org/apache/solr/search/QueryLimits.java
sigram: Haha ... it's not a user-facing API, we can always rename it when we want to generalize it.

solr/core/src/test/org/apache/solr/search/TestCpuAllowedLimit.java
sigram: Hmm... yes, that would be useful. `TestCpuTimeSearch` has an example, I'm going to use it (though it's concerning that the distrib test there is Ignored).

sigram: Reverted back to SolrCloudTestCase, the other method was too unstable.

solr/core/src/test/org/apache/solr/search/TestCpuQueryTimeLimit.java
sigram: Changed it back to SolrCloudTestCase, it was too unstable for distrib multi-shard query.

solr/core/src/test/org/apache/solr/search/TestCpuAllowedLimit.java
```diff
@@ -20,32 +20,61 @@
...
   }
 
-  @AfterClass
-  public static void teardown() {
-    System.clearProperty(ThreadCpuTimer.ENABLE_CPU_TIME);
+  @BeforeClass
+  public static void setup() throws Exception {
+    System.setProperty(ThreadCpuTimer.ENABLE_CPU_TIME, "true");
+    Path configset = createConfigSet();
+    configureCluster(2).addConfig("conf", configset).configure();
```
dsmiley, line 67: Do we actually need 2 nodes to test this?  (Trying to keep tests as lean as needed to test what they need to test; no more)

dsmiley (approved):
Overall looks good to me; I just have one small comment.

solr/core/src/test/org/apache/solr/search/TestCpuAllowedLimit.java
sigram: Probably not at the moment... My thinking was that if/when we want to add more complex tests that trigger termination at various stages in the distrib processing then it's best to also include inter-node communication in the setup. We're not there yet so I can reduce it to 1 node.

dsmiley (approved):
I've found it to be *very* rare where we *actually* need more than one node to test functionality.

gus-asf (approved):

PR #2247, opened 2024-02-06
SOLR-17148: Fixing Config API overlay property enabling or disabling cache

iamsanjay:
https://issues.apache.org/jira/browse/SOLR-17148

# Description

Config API is not able to enable or disable the cache. 

# Solution

Please refer to the ticket https://issues.apache.org/jira/browse/SOLR-17148

# Tests

Five test cases added
1. Cache enabled in SolrConfig.XML returns valid CacheConfig
2. Cache disable in SolrConfigXML returns null CacheConfig
3. Cache enabled in SolrConfig but disable in ConfigOverlay returns null CacheConfig
4. Cache disabled in SolrConfig but enabled in ConfigOverlay returns Valid CacheConfig.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/CacheConfig.java
+++ b/solr/core/src/java/org/apache/solr/search/CacheConfig.java
@@ -28,7 +28,6 @@ import java.util.function.Supplier;
 import org.apache.solr.common.ConfigNode;
 import org.apache.solr.common.MapSerializable;
 import org.apache.solr.common.util.CollectionUtil;
-import org.apache.solr.common.util.StrUtils;
 import org.apache.solr.core.PluginInfo;
 import org.apache.solr.core.SolrConfig;
 import org.apache.solr.core.SolrResourceLoader;
@@ -81,7 +80,7 @@ public class CacheConfig implements MapSerializable {
 
   public static Map<String, CacheConfig> getMultipleConfigs(
       SolrResourceLoader loader, SolrConfig solrConfig, String configPath, List<ConfigNode> nodes) {
-    if (nodes == null || nodes.size() == 0) return new LinkedHashMap<>();
+    if (nodes == null || nodes.isEmpty()) return new LinkedHashMap<>();
     Map<String, CacheConfig> result = CollectionUtil.newHashMap(nodes.size());
     for (ConfigNode node : nodes) {
       if (node.boolAttr("enabled", true)) {
@@ -94,12 +93,7 @@ public class CacheConfig implements MapSerializable {
   }
 
   public static CacheConfig getConfig(SolrConfig solrConfig, ConfigNode node, String xpath) {
-    if (!node.exists() || !"true".equals(node.attributes().get("enabled", "true"))) {
-      Map<String, Object> m = solrConfig.getOverlay().getEditableSubProperties(xpath);
-      if (m == null) return null;
-      List<String> parts = StrUtils.splitSmart(xpath, '/');
-      return getConfig(solrConfig, parts.get(parts.size() - 1), Collections.emptyMap(), xpath);
-    }
+    if (!node.boolAttr("enabled", true) || !node.exists()) return null;
     return getConfig(solrConfig, node.name(), node.attributes().asMap(), xpath);
   }
 
@@ -178,9 +172,9 @@ public class CacheConfig implements MapSerializable {
   }
 
   @Override
-  public Map<String, Object> toMap(Map<String, Object> map) {
-    // TODO: Should not create new HashMap?
-    return new HashMap<>(args);
+  public Map<String, Object> toMap(Map<String, Object> argsMap) {
+    argsMap.putAll(args);
+    return argsMap;
   }
 
   public String getNodeName() {
--- a/solr/core/src/java/org/apache/solr/search/CacheConfig.java
+++ b/solr/core/src/java/org/apache/solr/search/CacheConfig.java
@@ -80,7 +80,9 @@ public class CacheConfig implements MapSerializable {
 
   public static Map<String, CacheConfig> getMultipleConfigs(
       SolrResourceLoader loader, SolrConfig solrConfig, String configPath, List<ConfigNode> nodes) {
-    if (nodes == null || nodes.isEmpty()) return new LinkedHashMap<>();
+    if (nodes == null || nodes.isEmpty()) {
+      return new LinkedHashMap<>();
+    }
     Map<String, CacheConfig> result = CollectionUtil.newHashMap(nodes.size());
     for (ConfigNode node : nodes) {
       if (node.boolAttr("enabled", true)) {
@@ -93,7 +95,9 @@ public class CacheConfig implements MapSerializable {
   }
 
   public static CacheConfig getConfig(SolrConfig solrConfig, ConfigNode node, String xpath) {
-    if (!node.boolAttr("enabled", true) || !node.exists()) return null;
+    if (!node.boolAttr("enabled", true) || !node.exists()) {
+      return null;
+    }
     return getConfig(solrConfig, node.name(), node.attributes().asMap(), xpath);
   }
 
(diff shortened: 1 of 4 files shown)


solr/core/src/java/org/apache/solr/search/CacheConfig.java
```diff
@@ -81,7 +80,7 @@ public void setRegenerator(CacheRegenerator regenerator) {
 
   public static Map<String, CacheConfig> getMultipleConfigs(
       SolrResourceLoader loader, SolrConfig solrConfig, String configPath, List<ConfigNode> nodes) {
-    if (nodes == null || nodes.size() == 0) return new LinkedHashMap<>();
+    if (nodes == null || nodes.isEmpty()) return new LinkedHashMap<>();
```
epugh, line 83: i prefer to nest these in `{ }` instead of single line return..

solr/core/src/java/org/apache/solr/search/CacheConfig.java
```diff
@@ -94,12 +93,7 @@ public static Map<String, CacheConfig> getMultipleConfigs(
   }
 
   public static CacheConfig getConfig(SolrConfig solrConfig, ConfigNode node, String xpath) {
-    if (!node.exists() || !"true".equals(node.attributes().get("enabled", "true"))) {
-      Map<String, Object> m = solrConfig.getOverlay().getEditableSubProperties(xpath);
-      if (m == null) return null;
-      List<String> parts = StrUtils.splitSmart(xpath, '/');
-      return getConfig(solrConfig, parts.get(parts.size() - 1), Collections.emptyMap(), xpath);
-    }
+    if (!node.boolAttr("enabled", true) || !node.exists()) return null;
```
epugh, line 96: liewise here nest the return...

epugh (changes requested):
Hoping @hossman will review this as well..   Couple of formatting nits..

solr/core/src/java/org/apache/solr/search/CacheConfig.java
```diff
@@ -94,11 +95,8 @@ public static Map<String, CacheConfig> getMultipleConfigs(
   }
 
   public static CacheConfig getConfig(SolrConfig solrConfig, ConfigNode node, String xpath) {
-    if (!node.exists() || !"true".equals(node.attributes().get("enabled", "true"))) {
-      Map<String, Object> m = solrConfig.getOverlay().getEditableSubProperties(xpath);
-      if (m == null) return null;
-      List<String> parts = StrUtils.splitSmart(xpath, '/');
-      return getConfig(solrConfig, parts.get(parts.size() - 1), Collections.emptyMap(), xpath);
```
iamsanjay, line 101: Here, Earlier If the cache is disabled in SolrConfig.xml then If it has overlay properties, only those will be used to create the cache object. 

However, I think overlay is only meant to override the old values. As per the new changes added, If cache is disabled in the solrConfig.xml and enabled via overlay API, and nothing new added in overlay, then It's going to use the old attributes from the SolrConfig.XML

```
<documentCache
      enabled=false
      size="399"
      initialSize="99"
      autowarmCount="0"/>
```
And after enabling it via overlay API, the documentCache will be initialized using `size="399" and initialSize="99"`

iamsanjay: One more thing, earlier If `!node.exists()` returns false, we would check whether if it has Overlay or not. But I believe we do not need to check for Overlay anymore. Because when we call getConfig, we always check if it's present in Overlay and then pass object wrapped in OverlaidConfigNode which returns true by default, only return false, If it's empty node.

solr/core/src/test/org/apache/solr/core/TestSolrConfigHandler.java
```diff
@@ -1011,4 +1011,61 @@ public static LinkedHashMapWriter getRespMap(String path, RestTestHarness restHa
...
+    MapWriter confMap = getRespMap("/admin/metrics", restTestHarness);
+    assertNull(
+            confMap._get(
+                    asList("metrics", "solr.core.collection1", "CACHE.searcher.documentCache"), null));
+    restTestHarness.setServerProvider(oldProvider);
+  }
+  public void testSetPropertyEnableAndDisableCache() throws Exception {
+    RESTfulServerProvider oldProvider = restTestHarness.getServerProvider();
+    // Enabling Cache
+    String payload = "{\n" + " 'set-property' : { 'query.documentCache.enabled': true} \n" + " }";
```
epugh, line 1046: i think, for readablity, can we remove the extra `\n` formatting, or, if we want it to be multi line, then format it over mulitple liens?  (I look forward to the version of Java that lets us have multiple line strings!!!)

solr/core/src/test/org/apache/solr/core/TestSolrConfigHandler.java
```diff
@@ -1011,4 +1011,61 @@ public static LinkedHashMapWriter getRespMap(String path, RestTestHarness restHa
...
+        confMap._get(
+            asList("metrics", "solr.core.collection1", "CACHE.searcher.documentCache"), null));
+  }
+
+  public void testSetPropertyCacheSize() throws Exception{
+    RESTfulServerProvider oldProvider = restTestHarness.getServerProvider();
+    // Changing cache size
+    String payload = "{\n" + " 'set-property' : { 'query.documentCache.size': 399} \n" + " }";
+    restTestHarness.setServerProvider(oldProvider);
+    runConfigCommand(restTestHarness, "/config", payload);
```
epugh, line 1032: not sure I understand the behavior of the "oldProvider"?  Looks like we get it and then set it again?

solr/core/src/test/org/apache/solr/core/TestSolrConfigHandler.java
```diff
@@ -1011,4 +1011,61 @@ public static LinkedHashMapWriter getRespMap(String path, RestTestHarness restHa
...
+  }
+  public void testSetPropertyEnableAndDisableCache() throws Exception {
+    RESTfulServerProvider oldProvider = restTestHarness.getServerProvider();
+    // Enabling Cache
+    String payload = "{\n" + " 'set-property' : { 'query.documentCache.enabled': true} \n" + " }";
+    restTestHarness.setServerProvider(oldProvider);
+    runConfigCommand(restTestHarness, "/config", payload);
+    MapWriter overlay = getRespMap("/config/overlay", restTestHarness);
+    assertEquals("true", overlay._getStr("overlay/props/query/documentCache/enabled", null));
+    restTestHarness.setServerProvider(() -> getBaseUrl());
```
epugh, line 1051: like wise not sure we need this?

epugh (changes requested):
looks like progress!

solr/core/src/test/org/apache/solr/core/TestSolrConfigHandler.java
iamsanjay: Actually in-between I changed it. So just reset it back to old behaviour. 
` restTestHarness.setServerProvider(() -> getBaseUrl());`

solr/core/src/java/org/apache/solr/search/CacheConfig.java
```diff
@@ -177,8 +177,8 @@ public SolrCache newInstance() {
 
   @Override
   public Map<String, Object> toMap(Map<String, Object> argsMap) {
-    argsMap.putAll(args);
-    return argsMap;
+    // TODO: Should not create new HashMap?
```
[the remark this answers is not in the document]
iamsanjay: I can't say for sure, which is why I decided to revert it back. I found the pull request that introduced it; it needed more careful inspection.
https://issues.apache.org/jira/browse/SOLR-14155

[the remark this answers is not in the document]
epugh: Thanks...    Appreciate that.!

PR #2248, opened 2024-02-06
SOLR-13350: Multithreaded search

chatman:
Abandoning previous PRs to create this new one.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -49,6 +49,8 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Function;
@@ -175,7 +177,11 @@ public class CoreContainer {
 
   final SolrCores solrCores;
 
-  public static class CoreLoadFailure {
+    public Executor getCollectorExecutor() {
+      return collectorExecutor;
+    }
+
+    public static class CoreLoadFailure {
 
     public final CoreDescriptor cd;
     public final Exception exception;
@@ -277,6 +283,8 @@ public class CoreContainer {
   private final ObjectCache objectCache = new ObjectCache();
 
   public final NodeRoles nodeRoles = new NodeRoles(System.getProperty(NodeRoles.NODE_ROLES_PROP));
+
+  private final ExecutorService collectorExecutor;
 
   private final ClusterSingletons clusterSingletons =
       new ClusterSingletons(
@@ -431,6 +439,9 @@ public class CoreContainer {
     this.allowPaths = allowPathBuilder.build();
 
     this.allowListUrlChecker = AllowListUrlChecker.create(config);
+
+    this.collectorExecutor = ExecutorUtil.newMDCAwareCachedThreadPool(6,
+        new SolrNamedThreadFactory("searcherCollector"));
   }
 
   @SuppressWarnings({"unchecked"})
@@ -656,6 +667,7 @@ public class CoreContainer {
     distributedCollectionCommandRunner = Optional.empty();
     allowPaths = null;
     allowListUrlChecker = null;
+    collectorExecutor = null;
   }
 
   public static CoreContainer createAndLoad(Path solrHome) {
@@ -1247,6 +1259,7 @@ public class CoreContainer {
     }
 
     ExecutorUtil.shutdownAndAwaitTermination(coreContainerAsyncTaskExecutor);
+    ExecutorUtil.shutdownAndAwaitTermination(collectorExecutor);
     ExecutorService customThreadPool =
         ExecutorUtil.newMDCAwareCachedThreadPool(new SolrNamedThreadFactory("closeThreadPool"));
 
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -177,11 +177,11 @@ public class CoreContainer {
 
   final SolrCores solrCores;
 
-    public Executor getCollectorExecutor() {
-      return collectorExecutor;
-    }
+  public Executor getCollectorExecutor() {
+    return collectorExecutor;
+  }
 
-    public static class CoreLoadFailure {
+  public static class CoreLoadFailure {
 
     public final CoreDescriptor cd;
     public final Exception exception;
@@ -440,8 +440,9 @@ public class CoreContainer {
 
     this.allowListUrlChecker = AllowListUrlChecker.create(config);
 
-    this.collectorExecutor = ExecutorUtil.newMDCAwareCachedThreadPool(6,
-        new SolrNamedThreadFactory("searcherCollector"));
+    this.collectorExecutor =
+        ExecutorUtil.newMDCAwareCachedThreadPool(
+            6, new SolrNamedThreadFactory("searcherCollector"));
   }
 
   @SuppressWarnings({"unchecked"})
--- a/solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
+++ b/solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
@@ -26,11 +26,13 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
@@ -51,10 +53,12 @@ import org.apache.lucene.index.StoredFieldVisitor;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.index.Terms;
 import org.apache.lucene.index.TermsEnum;
+import org.apache.lucene.search.BooleanClause;
 import org.apache.lucene.search.BooleanClause.Occur;
 import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.search.CollectionStatistics;
 import org.apache.lucene.search.Collector;
+import org.apache.lucene.search.CollectorManager;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.Explanation;
 import org.apache.lucene.search.FieldDoc;
@@ -63,6 +67,7 @@ import org.apache.lucene.search.LeafCollector;
 import org.apache.lucene.search.MatchAllDocsQuery;
 import org.apache.lucene.search.MultiCollector;
 import org.apache.lucene.search.Query;
+import org.apache.lucene.search.QueryVisitor;
 import org.apache.lucene.search.Scorable;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.ScoreMode;
@@ -87,6 +92,7 @@ import org.apache.lucene.store.Directory;
 import org.apache.lucene.util.Bits;
 import org.apache.lucene.util.BytesRef;
 import org.apache.lucene.util.FixedBitSet;
+import org.apache.lucene.util.automaton.ByteRunAutomaton;
 import org.apache.solr.common.SolrException;
 import org.apache.solr.common.SolrException.ErrorCode;
 import org.apache.solr.common.params.ModifiableSolrParams;
@@ -108,6 +114,7 @@ import org.apache.solr.response.SolrQueryResponse;
 import org.apache.solr.schema.IndexSchema;
 import org.apache.solr.schema.SchemaField;
 import org.apache.solr.search.facet.UnInvertedField;
+import org.apache.solr.search.join.GraphQuery;
 import org.apache.solr.search.stats.StatsCache;
 import org.apache.solr.search.stats.StatsSource;
 import org.apache.solr.uninverting.UninvertingReader;
@@ -335,7 +342,7 @@ public class SolrIndexSearcher extends IndexSearcher implements Closeable, SolrI
       boolean reserveDirectory,
       DirectoryFactory directoryFactory)
       throws IOException {
-    super(wrapReader(core, r));
+    super(wrapReader(core, r), core.getCoreContainer().getCollectorExecutor());
 
     this.path = path;
     this.directoryFactory = directoryFactory;
@@ -1799,7 +1806,7 @@ public class SolrIndexSearcher extends IndexSearcher implements Closeable, SolrI
     int last = len;
     if (last < 0 || last > maxDoc()) last = maxDoc();
     final int lastDocRequested = last;
-    int nDocsReturned;
+    int nDocsReturned = 0;
     int totalHits;
     float maxScore;
     int[] ids;
@@ -1858,7 +1865,6 @@ public class SolrIndexSearcher extends IndexSearcher implements Closeable, SolrI
 
       buildAndRunCollectorChain(qr, query, collector, cmd, pf.postFilter);
 
-      nDocsReturned = 0;
       ids = new int[nDocsReturned];
       scores = new float[nDocsReturned];
       totalHits = numHits[0];
@@ -1867,10 +1873,17 @@ public class SolrIndexSearcher extends IndexSearcher implements Closeable, SolrI
       qr.setNextCursorMark(cmd.getCursorMark());
       hitsRelation = Relation.EQUAL_TO;
     } else {
+      log.info("calling from 2, query: "+query.getClass()); // nocommit
+      MTCollectorQueryCheck allowMT = new MTCollectorQueryCheck();
+      query.visit(allowMT);
+      TopDocs topDocs;
+      if (pf.postFilter != null || cmd.getSegmentTerminateEarly() || cmd.getTimeAllowed() > 0
+              || !allowMT.allowed()) {
+        log.debug("skipping collector manager");
       final TopDocsCollector<?> topCollector = buildTopDocsCollector(len, cmd);
       MaxScoreCollector maxScoreCollector = null;
       Collector collector = topCollector;
-      if ((cmd.getFlags() & GET_SCORES) != 0) {
+      if (needScores) {
         maxScoreCollector = new MaxScoreCollector();
         collector = MultiCollector.wrap(topCollector, maxScoreCollector);
       }
@@ -1878,7 +1891,7 @@ public class SolrIndexSearcher extends IndexSearcher implements Closeable, SolrI
           buildAndRunCollectorChain(qr, query, collector, cmd, pf.postFilter).scoreMode();
 
       totalHits = topCollector.getTotalHits();
-      TopDocs topDocs = topCollector.topDocs(0, len);
+      topDocs = topCollector.topDocs(0, len);
       if (scoreModeUsed == ScoreMode.COMPLETE || scoreModeUsed == ScoreMode.COMPLETE_NO_SCORES) {
         hitsRelation = TotalHits.Relation.EQUAL_TO;
       } else {
@@ -1886,7 +1899,7 @@ public class SolrIndexSearcher extends IndexSearcher implements Closeable, SolrI
       }
       if (cmd.getSort() != null
           && cmd.getQuery() instanceof RankQuery == false
-          && (cmd.getFlags() & GET_SCORES) != 0) {
+          && needScores) {
         TopFieldCollector.populateScores(topDocs.scoreDocs, this, query);
       }
       populateNextCursorMarkFromTopDocs(qr, cmd, topDocs);
@@ -1896,6 +1909,37 @@ public class SolrIndexSearcher extends IndexSearcher implements Closeable, SolrI
               ? (maxScoreCollector == null ? Float.NaN : maxScoreCollector.getMaxScore())
               : 0.0f;
       nDocsReturned = topDocs.scoreDocs.length;
+
+      } else {
+        log.info("using CollectorManager");
+        SearchResult searchResult = searchCollectorManagers(len, cmd, query, true, needScores, false);
+        Object[] res = searchResult.result;
+        TopDocsResult result = (TopDocsResult) res[0];
+
+        totalHits = result.totalHits;
+        topDocs = result.topDocs;
+
+        if (res.length > 1) {
+          MaxScoreResult result2 = (MaxScoreResult) res[1];
+          maxScore = totalHits > 0 ? result2.maxScore : 0.0f;
+        } else {
+          maxScore = Float.NaN;
+        }
+
+        populateNextCursorMarkFromTopDocs(qr, cmd, topDocs);
+        if (cmd.getSort() != null && !(cmd.getQuery() instanceof RankQuery) && needScores) {
+          TopFieldCollector.populateScores(topDocs.scoreDocs, this, query);
+        }
+        nDocsReturned = topDocs.scoreDocs.length;
+        ScoreMode scoreModeUsed = searchResult.scoreMode;
+        if (scoreModeUsed == ScoreMode.COMPLETE || scoreModeUsed == ScoreMode.COMPLETE_NO_SCORES) {
+          hitsRelation = TotalHits.Relation.EQUAL_TO;
+        } else {
+          hitsRelation = topDocs.totalHits.relation;
+        }
+
+      }
+
       ids = new int[nDocsReturned];
       scores = (cmd.getFlags() & GET_SCORES) != 0 ? new float[nDocsReturned] : null;
       for (int i = 0; i < nDocsReturned; i++) {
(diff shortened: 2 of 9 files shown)


solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -440,8 +440,9 @@ public CoreContainer(NodeConfig config, CoresLocator locator, boolean asyncSolrC
 
     this.allowListUrlChecker = AllowListUrlChecker.create(config);
 
-    this.collectorExecutor = ExecutorUtil.newMDCAwareCachedThreadPool(6,
-        new SolrNamedThreadFactory("searcherCollector"));
+    this.collectorExecutor =
+        ExecutorUtil.newMDCAwareCachedThreadPool(
+            6, new SolrNamedThreadFactory("searcherCollector"));
```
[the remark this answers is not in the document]
chatman: Thanks @cpoerschke for your review! This is something I'm still wondering about. I'll post a comment on JIRA about the number of threads.

solr/core/src/test/org/apache/solr/search/TestFiltering.java
```diff
@@ -92,21 +93,34 @@ public void testLiveDocsSharing() throws Exception {
...
+        assertEffectivelySame(set.getFixedBitSet(), live.getFixedBitSet());
       }
 
     } finally {
       req.close();
     }
   }
 
+  /** If the a XOR b == 0, then both a & b are effectively the same bitset */
+  private void assertEffectivelySame(FixedBitSet a, FixedBitSet b) {
```
dsmiley, line 112: Can you explain why assertSame is no longer a working in this test?  I remember working on this with Yonik.  The very point of this test is to test that we use the actual instance of live docs (notice the test method's name)

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1867,35 +1873,79 @@ public ScoreMode scoreMode() {
...
-        hitsRelation = TotalHits.Relation.EQUAL_TO;
       } else {
-        hitsRelation = topDocs.totalHits.relation;
-      }
-      if (cmd.getSort() != null
-          && cmd.getQuery() instanceof RankQuery == false
-          && (cmd.getFlags() & GET_SCORES) != 0) {
-        TopFieldCollector.populateScores(topDocs.scoreDocs, this, query);
+        if (log.isInfoEnabled()) {
+          log.info("using CollectorManager");
```
dsmiley, line 1919: more debugging code

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1910,6 +1960,203 @@ public ScoreMode scoreMode() {
...
+  static class DocSetResult {
+    final DocSet docSet;
+
+    public DocSetResult(DocSet docSet) {
+      this.docSet = docSet;
+    }
+  }
+
+  static class SearchResult {
+    final Object[] result;
```
dsmiley, line 2151: type Object?

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1910,6 +1960,203 @@ public ScoreMode scoreMode() {
...
+    }
+    for (Collector collector : firstCollectors) {
+      if (collector != null) {
+        if (scoreMode == null) {
+          scoreMode = collector.scoreMode();
+        } else if (scoreMode != collector.scoreMode()) {
+          scoreMode = ScoreMode.COMPLETE;
+        }
+      }
+    }
```
dsmiley, line 2101: I've seen this code at least twice in this patch, if I'm not mistaken.  Could be a utility method.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1867,35 +1873,79 @@ public ScoreMode scoreMode() {
...
-      Collector collector = topCollector;
-      if ((cmd.getFlags() & GET_SCORES) != 0) {
-        maxScoreCollector = new MaxScoreCollector();
-        collector = MultiCollector.wrap(topCollector, maxScoreCollector);
+      if (log.isInfoEnabled()) {
+        log.info("calling from 2, query: {}", query.getClass());
       }
-      ScoreMode scoreModeUsed =
-          buildAndRunCollectorChain(qr, query, collector, cmd, pf.postFilter).scoreMode();
+      MTCollectorQueryCheck allowMT = new MTCollectorQueryCheck();
```
dsmiley, line 1879: Bunch of code being added in two places looking similar.  Have you thought of attempting to factor out a common approach?

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1979,36 +2226,84 @@ public ScoreMode scoreMode() {
...
+          }
+        } else {
+          if (res.length > 1) {
+            DocSetResult result2 = (DocSetResult) res[1];
+            set = result2.docSet;
+          }
+        }
+
+        populateNextCursorMarkFromTopDocs(qr, cmd, topDocs);
+        //        if (cmd.getSort() != null && !(cmd.getQuery() instanceof RankQuery) &&
```
dsmiley, line 2294: Forgot some logic or...?

solr/core/src/java/org/apache/solr/search/SolrMultiCollectorManager.java
```diff
@@ -0,0 +1,150 @@
...
+import org.apache.lucene.search.Collector;
+import org.apache.lucene.search.CollectorManager;
+import org.apache.lucene.search.FilterScorable;
+import org.apache.lucene.search.LeafCollector;
+import org.apache.lucene.search.MultiCollector;
+import org.apache.lucene.search.Scorable;
+import org.apache.lucene.search.ScoreMode;
+
+/**
+ * A {@link CollectorManager} implements which wrap a set of {@link CollectorManager} as {@link
```
dsmiley, line 33:
```suggestion
 * A {@link CollectorManager} implementation which wrap a set of {@link CollectorManager} as {@link
```

solr/core/src/java/org/apache/solr/search/SolrMultiCollectorManager.java
```diff
@@ -0,0 +1,150 @@
...
+ * MultiCollector} acts for {@link Collector}.
+ */
+public class SolrMultiCollectorManager
+    implements CollectorManager<SolrMultiCollectorManager.Collectors, Object[]> {
```
dsmiley, line 37: Hmm; Object?  Do we not know the type?

solr/solr-ref-guide/modules/indexing-guide/examples/stemdict.txt
dsmiley, line 1: Looks like you removed a symlink?

dsmiley (commented):
Seems a draft with all the logging and some commented code but anyway it's nice to see this!

solr/core/src/java/org/apache/solr/search/SolrMultiCollectorManager.java
```diff
@@ -0,0 +1,150 @@
...
+      final List<Collector> reducableCollector = new ArrayList<>(size);
+      for (Collectors collectors : reducableCollectors)
+        reducableCollector.add(collectors.collectors[i]);
+      results[i] = collectorManagers[i].reduce(reducableCollector);
+    }
+    return results;
+  }
+
+  /** Wraps multiple collectors for processing */
+  public class Collectors implements Collector {
```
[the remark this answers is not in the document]
cpoerschke: > not sure yet if these inner classes need to be public.

https://github.com/apache/solr/pull/2248/commits/3e66068f1405ed36a2c45dc529645706d9fec0ea to reduce visibility -- `implements CollectorManager<SolrMultiCollectorManager.Collectors, Object[]>` says `SolrMultiCollectorManager.Collectors has private access in SolrMultiCollectorManager` if it's private but package visibility is okay and `LeafCollectors` can be private.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1806,7 +1813,7 @@ private void getDocListNC(QueryResult qr, QueryCommand cmd) throws IOException {
     int last = len;
```
cpoerschke, line 1813: existing code: personally I would find more `final` use in this method helpful, #2258 proposes to add it, I'm happy to resolve any merge conflicts that this may cause for the pull request here.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1874,35 +1881,74 @@ public ScoreMode scoreMode() {
...
       }
-      populateNextCursorMarkFromTopDocs(qr, cmd, topDocs);
 
-      maxScore =
-          totalHits > 0
-              ? (maxScoreCollector == null ? Float.NaN : maxScoreCollector.getMaxScore())
-              : 0.0f;
-      nDocsReturned = topDocs.scoreDocs.length;
       ids = new int[nDocsReturned];
       scores = (cmd.getFlags() & GET_SCORES) != 0 ? new float[nDocsReturned] : null;
```
cpoerschke, line 1953: Well spotted about reuse of `needScores` above in two places, here it can be used too I think.
```suggestion
      scores = needScores ? new float[nDocsReturned] : null;
```

I've opened #2262 to change that separately for other places in this file too. Am happy to resolve any merge conflicts that may arise for this pull request here.

dsmiley:
Observation: I don't think we need the concurrent DocSet builder (the one from Netflix).  Couldn't we build segment level FixedBitSets (no safety issue) and then at the end combine into a master FixedBitSet so that we have our DocSet?  It could even be made to be long-aligned, and thus the final aggregation is just copying longs with System.arraycopy (no doc iteration! or long shifting).  The boundary long would be XOR.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
cpoerschke: https://github.com/apache/solr/pull/2248/commits/c69006a4ca6625307e17fb0c272b08eebacd9a69 and https://github.com/apache/solr/pull/2248/commits/1ec21d4274fb10eaeec2fed04093b91cf1e88c72 factored something out. Not thought much about naming in doing so and not specifically in response to this comment but just from my code reading/learning/reviewing perspective.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
cpoerschke: The second place is https://github.com/apache/solr/blob/74c05e520abe85f056fe440976dbbd6a812b6386/solr/core/src/java/org/apache/solr/search/SolrMultiCollectorManager.java#L86-L97 and it's similar but not identical i.e. here we have a `null` check on the collector.

And I note that there is https://github.com/apache/lucene/blob/releases/lucene/9.9.2/lucene/core/src/java/org/apache/lucene/search/MultiCollector.java#L104-L121 logic very similar (but again not identical) i.e. if there was (say) a

```
public static ScoreMode scoreMode(Collector... collectors)
```

available in Lucene then maybe that could be used in both places (after filtering out any nulls in `firstCollectors` here.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1917,6 +1963,203 @@ public ScoreMode scoreMode() {
...
+
+                for (LeafReaderContext reader : leaf.leaves) {
+                  numDocs += reader.reader().maxDoc();
+                }
+              } else {
+                numDocs = maxDoc();
+              }
+              log.error("new docset collector for {} max={}", numDocs, maxDoc());
+
+              return new ThreadSafeBitSetCollector(bits, maxDoc);
```
cpoerschke, line 2084: Not sure yet if a `firstCollectors[2] = ...` was intended here or if not we can reduce to `new Collector[2]` above it seems.

solr/core/src/java/org/apache/solr/search/SolrMultiCollectorManager.java
cpoerschke: Lucene has a `public class MultiCollectorManager implements CollectorManager<Collector, Object[]>` i.e. https://github.com/apache/lucene/blob/releases/lucene/9.9.2/lucene/core/src/java/org/apache/lucene/search/MultiCollectorManager.java and I'm curious if there'd be a way to code share with that, though so far I can't see a way there.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1910,6 +1954,190 @@ public ScoreMode scoreMode() {
...
+    }
+    for (Collector collector : firstCollectors) {
+      if (collector != null) {
+        if (scoreMode == null) {
+          scoreMode = collector.scoreMode();
+        } else if (scoreMode != collector.scoreMode()) {
+          scoreMode = ScoreMode.COMPLETE;
+        }
+      }
+    }
```
cpoerschke: I think this needs to happen after the `super.search` call since at this point here the collectors will still all be `null`? https://github.com/apache/solr/pull/2248/commits/2501f7b06f697f4a8b84610c6d5cfe586217fc4a to tentatively move it.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1731,6 +1738,19 @@ private void getDocListC(QueryResult qr, QueryCommand cmd) throws IOException {
...
+    if (cmd.getSort() != null && !(cmd.getQuery() instanceof RankQuery) && needScores) {
+      TopFieldCollector.populateScores(topDocs.scoreDocs, this, query);
+    }
+    if (scoreModeUsed == ScoreMode.COMPLETE || scoreModeUsed == ScoreMode.COMPLETE_NO_SCORES) {
+      return TotalHits.Relation.EQUAL_TO;
+    } else {
+      return topDocs.totalHits.relation;
+    }
+  }
+
```
cpoerschke, line 1753: Proposing to factor out a method like this -- see #2280 -- which can then be called from all four places. I'm happy to resolve any merge conflicts that may arise here if #2280 is merged first.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2254,7 +2541,7 @@ protected void sortDocSet(QueryResult qr, QueryCommand cmd) throws IOException {
     }
 
     // bit of a hack to tell if a set is sorted - do it better in the future.
-    boolean inOrder = set instanceof BitDocSet || set instanceof SortedIntDocSet;
+    // boolean inOrder = set instanceof BitDocSet || set instanceof SortedIntDocSet;
```
cpoerschke, line 2544: Good spot! Opened #2281 to remove both the comment and the variable.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
cpoerschke: https://github.com/apache/solr/pull/2248/commits/ab895b101a0626f0d393d2564e289d31ddb6cc63 factors out a method here in Solr or perhaps the Lucene logic becomes available via https://github.com/apache/lucene/pull/13126 in future. WDYT?

cpoerschke:
> Observation: I don't think we need the concurrent DocSet builder (the one from Netflix). Couldn't we build segment level FixedBitSets (no safety issue) and then at the end combine into a master FixedBitSet so that we have our DocSet? It could even be made to be long-aligned, and thus the final aggregation is just copying longs with System.arraycopy (no doc iteration! or long shifting). The boundary long would be XOR.

@dsmiley - is something like https://github.com/apache/solr/pull/2248/commits/34c7db09e08693a79d87eaa6e43ad6f5aacbc342 what you had in mind here? For starters modelled on https://github.com/apache/lucene/blob/releases/lucene/9.10.0/lucene/test-framework/src/java/org/apache/lucene/tests/search/FixedBitSetCollector.java i.e. no `System.arraycopy` etc. as yet.

@chatman - i hope you don't mind me adding commits to this PR.

dsmiley:
@cpoerschke  https://github.com/apache/solr/commit/34c7db09e08693a79d87eaa6e43ad6f5aacbc342 No; that is a single FixedBitSet and as such you've basically changed this to Solr's existing default handling, more or less.  But that's not thread-safe!

cpoerschke:
> @cpoerschke [34c7db0](https://github.com/apache/solr/commit/34c7db09e08693a79d87eaa6e43ad6f5aacbc342) No; that is a single FixedBitSet and as such you've basically changed this to Solr's existing default handling, more or less. But that's not thread-safe!

Hmm, but each `FixedBitSet` is in a `FixedBitSetCollector` and multiple `FixedBitSetCollector` objects are created via the `CollectorManager.newCollector()` calls and then merged via the `CollectorManager.reduce()` call into a separate joint `FixedBitSet`? Anyhow, I can revert the commit, that's what version control is for sometimes.

cpoerschke:
Commit reverted, though I still have doubts i.e. maybe the commit message was just unhelpful i.e. _"replace ThreadSafeBitSet[Collector] with FixedBitSet[Collector]"_ is mechanical whereas a more behavioural commit message would have been the longer _"replace one ThreadSafeBitSet and multiple ThreadSafeBitSetCollector(s) with multiple FixedBitSetCollector(s) containing one FixedBitSet each, to be merged into one FixedBitSet at the end"_ or similar.

dsmiley:
Sorry; I didn't look at your change enough!  I agree that it should work, and is probably better than having a concurrent ThreadSafeBitSet.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -445,6 +447,17 @@ public CoreContainer(NodeConfig config, CoresLocator locator, boolean asyncSolrC
             cfg.getIndexSearcherExecutorThreads(), // thread count
             cfg.getIndexSearcherExecutorThreads(), // queue size
             new SolrNamedThreadFactory("searcherCollector"));
+    ((ExecutorUtil.MDCAwareThreadPoolExecutor) collectorExecutor).setRejectedExecutionHandler(new RejectedExecutionHandler() {
```
dsmiley, line 450: I think that's an anti-pattern or broken and isn't what I meant in JIRA.  We could use a SynchronousQueue (with fairness) if we want to block for a thread -- probably what we should do.  FYI that queue is the default for `Executors.newCachedThreadPool()`.    The "caller runs" behavior I meant could be done via an ExecutorService delegate that catches RejectedException and simply runs the Runnable.

cpoerschke:
> ... Couldn't we build segment level FixedBitSets (no safety issue) and then at the end combine into a master FixedBitSet so that we have our DocSet? It could even be made to be long-aligned, and thus the final aggregation is just copying longs with System.arraycopy (no doc iteration! or long shifting). The boundary long would be XOR.

Please could you share in more detail what you had in mind here? E.g.
* Did `FixedBitSet` refer literally to the https://github.com/apache/lucene/blob/releases/lucene/9.10.0/lucene/core/src/java/org/apache/lucene/util/FixedBitSet.java class or was it meant conceptually i.e. an existing or new class similar to but different from `FixedBitSet`?
* Is there existing similar code that does the long-aligning and final aggregation like that elsewhere?

With the current implementation https://github.com/apache/solr/blob/bec277ab17293aee8fe44ab65264ddb0f6b697d5/solr/core/src/java/org/apache/solr/search/MultiThreadedSearcher.java#L194-L204 I can see
* that the `maxDoc` is excessive not least because `this.docBase` is added for every document in the segment, and
* that conceptually aggregation can be done via copying,
* but beyond that I haven't yet connected the ideas shared above.

Thanks!

dsmiley:
Yes I mean *the* FixedBitSet (otherwise I just would have said a bit set).  No I don't know of similar code doing this, and I don't expect it's in Lucene/Solr already.  Lucene is very segment oriented so there's no need for an entire index bit set view.  Solr has forever clung to index level bit sets, for better/worse (I think worse).
I'm tempted to help but not sure I have the time.  It's not a blocker any way; the code can be improved.

chatman:
Currently the TestQueryLimits.java test is failing with this branch. I'm looking into this. Any ideas, please, @sigram?
My intuition in that the execution isn't breaching the limits in the test, and hence full results are coming for QueryComponent, instead of partial results.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
noblepaul: > The "caller runs" behavior I meant could be done via an ExecutorService delegate that catches RejectedException and simply runs the Runnable.

I don't think  "caller runs" is a good solution. If there are multiple rejected tasks because all threads are busy, the caller thread can only run one task, what will happen to the other tasks? Meanwhile the already completed tasks may free up other threads. So, IMHO, the current code with a bigger queue count should be fine

dsmiley: Have you considered SynchronousQueue instead of a big queue?  I'm not sure what value the queue has for this use-case.

noblepaul: Actually, a SynchronousQueue is good enough too. 

So, the main thread will automatically wait for a pool thread to be available for executing the tasks

cpoerschke:
> Yes I mean _the_ FixedBitSet (otherwise I just would have said a bit set). No I don't know of similar code doing this, and I don't expect it's in Lucene/Solr already. Lucene is very segment oriented so there's no need for an entire index bit set view. Solr has forever clung to index level bit sets, for better/worse (I think worse). I'm tempted to help but not sure I have the time. It's not a blocker any way; the code can be improved.

Okay, I've got the basic outline of the idea I think:

```
    final FixedBitSet bitsetA = new FixedBitSet(numBits);
    bitsetA.set(...);

    final int skipWords = 8;
    final int skipBits = skipWords * 64;

    final FixedBitSet bitsetB = new FixedBitSet(numBits - skipBits);
    bitsetB.set(... - skipBits);

    final long[] storedBits = new long[FixedBitSet.bits2words(numBits)];
    System.arraycopy(bitsetA.getBits(), 0, storedBits, 0, bitsetA.getBits().length);
    System.arraycopy(bitsetB.getBits(), 0, storedBits, skipWords, bitsetB.getBits().length);

    final FixedBitSet bitsetC = new FixedBitSet(storedBits, numBits);
```

And then as a next step I'll have a go at applying that to the collecting code here on this PR.

dsmiley:
I figure that's somewhat pseudocode... so I don't want to over-critique that but it should have the XOR of the boundary long in-between each `arraycopy`.  In practice you will likely loop the FixedBitSets, skipping the first in order to XOR the last long of the previous FixedBitSet with the first long of the current one.

cpoerschke:
So the https://github.com/apache/solr/pull/2248/commits/fdfef9f734d091db7298d3493681d7f07f50c043 commit starts to explore using segment level bit sets ...

... but actually I don't _yet_ quite see what exactly that would 'buy' since conceptually then we'd need to maintain multiple sets of `maxDoc - docBase` size and in that case why not just have one set of `maxDoc` then? ...

... unless we can _somehow_ size the segment level bit sets to be the size of the segments?

dsmiley:
Definitely the segment bit sets need to be sized to the segment, plus no more than 63 bits for long-alignment.

solr/core/src/java/org/apache/solr/search/MultiThreadedSearcher.java
```diff
@@ -0,0 +1,415 @@
...
+    }
+
+    @Override
+    public void collect(int doc) throws IOException {
+      FixedBitSet bitSet = this.bitSets.getLast();
+      final int idx = this.skipBits.getLast() + doc;
+
+      final int numWords = FixedBitSet.bits2words(idx + 1); // +1 to ensure minimum 1 word
+
+      if (bitSet == null) {
```
cpoerschke, line 210: I'm thinking lazy allocation makes sense since for some segments maybe nothing will ever be collected ...

... however `+1` here for segment bitset sizing seems too basic ...

... didn't (yet) find a number-of-documents-in-a-segment value ... `maxDoc / numberOfLeaves` might a way to initially approximate but once that's exceeded the question of how to upsize arises again ...

... thoughts?

solr/core/src/java/org/apache/solr/search/MultiThreadedSearcher.java
```diff
@@ -0,0 +1,415 @@
...
+    }
+
+    void update(FixedBitSet allBitSet) {
+      final long[] allBits = allBitSet.getBits();
+      for (int bs_idx = 0; bs_idx < this.bitSets.size(); ++bs_idx) {
+        final FixedBitSet itBitSet = this.bitSets.get(bs_idx);
+        if (itBitSet != null) {
+          final int skipWords = this.skipWords.get(bs_idx);
+          final long[] itBits = itBitSet.getBits();
+          for (int idx = 0; idx < itBits.length; ++idx) {
```
cpoerschke, line 232:
```suggestion
          for (int idx = 0; idx < itBits.length && skipWords + idx < allBits.length; ++idx) {
```

cpoerschke:
> Definitely the segment bit sets need to be sized to the segment, plus no more than 63 bits for long-alignment.

https://github.com/apache/solr/pull/2248/commits/b1ac97649081aed058a431896f236756100066d1 is a first take on segment bit sets.

the no-more-than-63-bits-for-long-alignment seems easy, how to size the segment bit sets needs a bit more thought.

solr/core/src/java/org/apache/solr/search/MultiThreadedSearcher.java
```diff
@@ -0,0 +1,415 @@
...
+
+    public MaxScoreResult(float maxScore) {
+      this.maxScore = maxScore;
+    }
+  }
+
+  static class FixedBitSetCollector extends SimpleCollector {
+    private final LinkedList<FixedBitSet> bitSets = new LinkedList<>();
+    private final LinkedList<Integer> skipWords = new LinkedList<>();
+    private final LinkedList<Integer> skipBits = new LinkedList<>();
```
cpoerschke, line 192: CI says _warning: [JdkObsolete] It is very rare for LinkedList to out-perform ArrayList or ArrayDeque. Avoid it unless you're willing to invest a lot of time into benchmarking. Caveat: LinkedList supports null elements, but ArrayDeque does not._

I chose `LinkedList` for the initial implementation here because the `getLast()/removeLast()/addLast()` are nice-to-read but we can change to a different type later on.

chatman:
bq. I think that's an anti-pattern or broken and isn't what I meant in JIRA. We could use a SynchronousQueue (with fairness) if we want to block for a thread -- probably what we should do. FYI that queue is the default for Executors.newCachedThreadPool(). The "caller runs" behavior I meant could be done via an ExecutorService delegate that catches RejectedException and simply runs the Runnable.

@dsmiley instead of using a rejected tasks execution handler, I went with @noblepaul 's suggestion of having a reasonably large queue for the threadpool (number of threads * 1000). Beyond this, if tasks are submitted, it is okay to reject them. We can revisit these limits later as well.

chatman:
Thanks @cpoerschke and @dsmiley. I've merged to main, where this can bake for some days before merging to branch_9x. If there are any major outstanding issues, or any changes needed to documentation or default behaviour, we can take it up in another PR.

dsmiley:
Merging to main was unexpected to me because of the healthy code review happening here didn't conclude.  Next time please announce an intention to do so in a couple days.  I would have given this another look!  For example I *thought* the DocSet merging aspect was still tentative... I was awaiting another comment from Christine.  And I was looking forward to checking out the thread pool aspect again.

dsmiley:
> This pull request is closed, but the jira/solr-13350 branch has unmerged commits.

Not sure what didn't make it

chatman:
> Merging to main was unexpected to me because of the healthy code review happening here didn't conclude.

I didn't want this to languish for a long time, and there were large PRs in the waiting that could affect this PR. Example: https://github.com/apache/solr/pull/2382

>  I would have given this another look! 

...

>  And I was looking forward to checking out the thread pool aspect again.

Please feel free to look at it, would be happy to address all loose ends, if any.

cpoerschke:
> Merging to main was unexpected to me because of the healthy code review happening here didn't conclude. ...

I was surprised by this merge too and my initial thought was that it might have been an accident and well that happens sometimes and we can just revert the commit and resume on a new PR then.

gus-asf:
The discussion on this is long, so maybe I've missed it, but the actual merged code has introduced the possibility (though I suspect it might never happen) of a non-numeric Max Score...

````
    public float getMaxScore(int totalHits) {
      if (totalHits > 0) {
        for (Object res : result) {
          if (res instanceof MaxScoreResult) {
            return ((MaxScoreResult) res).maxScore;
          }
        }
        return Float.NaN;    <<<<<<<<<<<<<<<<<<<<<<
      } else {
        return 0.0f;
      }
    }
````
Did I miss discussion of this?

File: solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java [hunk 21]
```diff
@@ -335,7 +335,7 @@ public SolrIndexSearcher(
       boolean reserveDirectory,
       DirectoryFactory directoryFactory)
       throws IOException {
-    super(wrapReader(core, r));
+    super(wrapReader(core, r), core.getCoreContainer().getCollectorExecutor());
```
cpoerschke, line 338: This change here is not controlled by the request level `multiThreaded=true|false` flag. #2570 explores how node-level opt-out could be supported.

solr/core/src/java/org/apache/solr/search/MultiThreadedSearcher.java
```diff
@@ -0,0 +1,423 @@
...
+    if (needTopDocs) {
+
+      collectors.add(new TopDocsCM(len, cmd, firstCollectors, firstTopDocsCollectorIndex));
+    }
+    if (needMaxScore) {
+      collectors.add(new MaxScoreCM(firstCollectors, firstMaxScoreCollectorIndex));
+    }
+    if (needDocSet) {
+      int maxDoc = searcher.getRawReader().maxDoc();
+      log.error("raw read max={}", searcher.getRawReader().maxDoc());
```
cpoerschke, line 94: Am unclear on why we log an error here. Also since the overall bit set is now computed as union of per-segment bit sets, perhaps there is no need to 'know' the max doc and it could be computed from the per-segment bit sets?

solr/core/src/test/org/apache/solr/search/TestFiltering.java
cpoerschke: It seems the change is no longer needed now, #2589 proposes to revert it.

solr/core/src/test/org/apache/solr/search/TestCpuAllowedLimit.java
```diff
@@ -185,7 +185,9 @@ public void testDistribLimit() throws Exception {
                 "stages",
                 "prepare,process",
                 "cpuAllowed",
-                "50"));
+                "50",
+                "multiThreaded",
+                "false"));
```
cpoerschke, line 190: #2590 proposes to explicitly not (yet) support multi-threading when query limits are used, and with that then the test change here could be reverted.

solr/core/src/test/org/apache/solr/search/TestQueryLimits.java
```diff
@@ -76,7 +76,17 @@ public void testQueryLimits() throws Exception {
...
+                  "q",
+                  "id:*",
+                  "sort",
+                  "id asc",
+                  "facet",
+                  "true",
+                  "facet.field",
+                  "val_i",
+                  "multiThreaded",
+                  "false"));
```
cpoerschke, line 89: #2590 proposes to explicitly not (yet) support multi-threading when query limits are used, and with that then the test change here could be reverted.

solr/core/src/java/org/apache/solr/search/MultiThreadedSearcher.java
dsmiley: logging: surely it was temporary during dev to ensure a log message could be seen easily.  It's easier than logging at trace and reconfiguring logging.  This must be fixed.

cpoerschke: oops, looks like this wasn't fixed yet: https://issues.apache.org/jira/browse/SOLR-17454

cpoerschke: #2720 to remedy.

PR #2256, opened 2024-02-12
Build: ge.gradle add tags & projectId=solr

dsmiley:
Add more Gradle Enterprise build metadata.

### DIFF
--- a/gradle/ge.gradle
+++ b/gradle/ge.gradle
@@ -22,6 +22,7 @@ def isCIBuild = System.getenv().keySet().any { it ==~ /(?i)((JENKINS|HUDSON)(_\w
 
 gradleEnterprise {
     server = "https://ge.apache.org"
+    projectId = "solr"
 
     buildScan {
         capture { taskInputFiles = true }
--- a/gradle/ge.gradle
+++ b/gradle/ge.gradle
@@ -22,7 +22,6 @@ def isCIBuild = System.getenv().keySet().any { it ==~ /(?i)((JENKINS|HUDSON)(_\w
 
 gradleEnterprise {
     server = "https://ge.apache.org"
-    projectId = "solr"
 
     buildScan {
         capture { taskInputFiles = true }
--- a/gradle/ge.gradle
+++ b/gradle/ge.gradle
@@ -15,10 +15,15 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


dsmiley:
@clayburn I'm trying to enhance the metadata in Develocity/GE for the Solr builds.  However I see no scans since Feb 7th as of this writing, which is weird.  Wouldn't I at least see the results of this very PR?:
https://ge.apache.org/scans?search.rootProjectNames=*solr*&search.timeZoneId=America%2FNew_York&search.usernames=runner

clayburn:
If you are wanting your project to show as `solr` instead of `solr-root`, you should instead set the `rootProject.name` in your `gradle.properties` file to `solr`. Setting a project id in the Gradle Enterprise configuration is a different feature entirely, which ge.apache.org is not  setup to use well yet.

Your build count looks like it has [remained steady](https://ge.apache.org/scans/trends?search.rootProjectNames=*solr*&search.startTimeMax=1707890355524&search.startTimeMin=1706594400000&search.timeZoneId=America%2FChicago) over the past two weeks. You have a `runner` username in the filter you shared, which implies GitHub Actions to me. Your build count with GitHub Actions appears consistently sporadic for some time: https://ge.apache.org/scans/trends?search.relativeStartTime=P90D&search.rootProjectNames=*solr*&search.timeZoneId=America%2FChicago&search.usernames=runner

dsmiley:
Thanks Clay.  I've removed the foiled attempt at removing "solr-root" again.  But I had done so earlier in this PR's history and at no point has this PR published a build scan.  Yes I'm aware the Jenkins integration seems to be working fine.  And since your post, some other PR or two is publishing but not others; not this one.  Is there a way to troubleshoot why that might be?

clayburn:
I believe it is because these builds are coming from forks, which will not have access to `secrets.GE_ACCESS_TOKEN` as it is an organization level secret under the Apache org. This is something we hope to support in the future with a different workflow (where the build runs in the fork, but the scan is uploaded in a separate privileged step).

dsmiley:
Aaaaaah!  I will move my code to a branch then.  I may need to close this PR to do that.

> where the build runs in the fork, but the scan is uploaded in a separate privileged step

Would this also be useful for external build machines (like Crave.io as we discussed)?

clayburn:
Yes, the concept would be the same. It just deals with a few implementation specific details regarding where the build scan data is stored on the build runner, moving it to the privileged runner, and publishing it from there

dsmiley:
Replaced by #2264

PR #2259, opened 2024-02-13
SOLR-599 New Solr Client Using Java 11+ java.net.http.HttpClient

jdyer1:
https://issues.apache.org/jira/browse/SOLR-599

This PR adds `HttpSolrClientJdkImpl`, a Solr client using the built-in Java 11+ Http Client rather than an external dependency.  This is intended to be a drop-in replacement for `Http2SolrClient` and/or `HttpSolrClient` for client applications running either Http/1.1 or Http/2.  This is not intended for Solr's server-side inter-node requests.  This new client is aimed at those users wishing to reduce transitive dependencies in their applications.

This PR factors out a new abstract superclass, `Http2SolrClientBase`, from which both `Http2SolrClient` and `HttpSolrClientJdkImpl` extend.  A similar refactoring is made to the corresponding unit tests.

Should it be desirable to include this new client in a future SolrJ release, a possible follow-up would be to split the SolrJ artifact.  A "core" artifact could be published having fewer transitive dependencies with additional artifacts per client library dependency.  This would allow downstream applications to include only those transitive dependencies they need, without resorting to build tool exclusion mechanisms.

Although I am including this as a potential solution for [SOLR-599](https://issues.apache.org/jira/browse/SOLR-599), this client is only "lightweight" in the sense it does not require additional client dependencies.  This client is intended to support most features included with the existing clients.

There is one TODO in the `HttpSolrClientJdkImpl` to possibly find a more memory-efficient way to handle POST requests.  There is a NOCOMMIT in `solr-tests.policy`.  I could not figure out how to properly set the permissions for `java.net.URLPermission`, resorting to `AllPermission` for the time being.


janhoy:
This is so cool! Let’s make it the new default client in a solrj-core artifact?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientJdkImpl.java
```diff
@@ -0,0 +1,443 @@
...
+    String method = resp.request() == null ? null : resp.request().method();
+    int status = resp.statusCode();
+    String reason = "" + status;
+    InputStream is = resp.body();
+    return processErrorsAndResponse(
+        status, reason, method, parser, is, mimeType, encoding, isV2ApiRequest(solrRequest), url);
+  }
+
+  @Override
+  public void close() throws IOException {
```
dsmiley, line 272: Don't we just need to do instanceof here and call it?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientJdkImpl.java
```diff
@@ -0,0 +1,443 @@
...
+  }
+
+  private static final Pattern MIME_TYPE_PATTERN = Pattern.compile("^(.*?)(?:;| |$)");
+
+  private String contentTypeToMimeType(String contentType) {
+    Matcher mimeTypeMatcher = MIME_TYPE_PATTERN.matcher(contentType);
+    return mimeTypeMatcher.find() ? mimeTypeMatcher.group(1) : null;
+  }
+
+  private static final Pattern CHARSET_PATTERN = Pattern.compile("(?i)^.*charset=(.*)?(?:;| |$)");
```
dsmiley, line 249: Prior to "charset" we can expect a space, I think.  And the dot-star you are matching should not be whitespace or a semicolon; then you could eliminate the non-capturing group that follows.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientJdkImpl.java
```diff
@@ -0,0 +1,443 @@
...
+      String encoded =
+          basicAuthCredentialsToAuthorizationString(
+              solrRequest.getBasicAuthUser(), solrRequest.getBasicAuthPassword());
+      reqb.header("Authorization", encoded);
+    } else if (basicAuthAuthorizationStr != null) {
+      reqb.header("Authorization", basicAuthAuthorizationStr);
+    }
+  }
+
+  private static final Pattern MIME_TYPE_PATTERN = Pattern.compile("^(.*?)(?:;| |$)");
```
dsmiley, line 242: Wouldn't it be simpler to replace the lazy match with a normal one where we don't match "." (anything) but instead constrain to not whitespace or semicolon?  Then you could eliminate the non-capturing group afterwards as well.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClientBase.java
```diff
@@ -0,0 +1,385 @@
...
+import org.apache.solr.client.solrj.request.RequestWriter;
+import org.apache.solr.client.solrj.request.V2Request;
+import org.apache.solr.common.SolrException;
+import org.apache.solr.common.params.CommonParams;
+import org.apache.solr.common.params.ModifiableSolrParams;
+import org.apache.solr.common.util.ContentStream;
+import org.apache.solr.common.util.NamedList;
+import org.apache.solr.common.util.Utils;
+
+public abstract class Http2SolrClientBase extends SolrClient {
```
dsmiley, line 49: There isn't anything here that is fundamentally about Http2; right?  Then lets drop the "2".

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClientBase.java
```diff
@@ -0,0 +1,385 @@
...
+
+  protected static final String DEFAULT_PATH = "/select";
+  protected static final Charset FALLBACK_CHARSET = StandardCharsets.UTF_8;
+  private static final List<String> errPath = Arrays.asList("metadata", "error-class");
+
+  /** The URL of the Solr server. */
+  protected final String serverBaseUrl;
```
dsmiley, line 56: please add a getter for this; it's rather common to want this.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
```diff
@@ -0,0 +1,193 @@
...
+  protected ExecutorService executor;
+  protected CookieStore cookieStore;
+  protected String proxyHost;
+  protected int proxyPort;
+  protected boolean proxyIsSocks4;
+  protected boolean proxyIsSecure;
+
+  public HttpSolrClientBuilderBase() {}
+
+  protected abstract <B extends Http2SolrClientBase> B build(Class<B> type);
```
dsmiley, line 48: I think "C" would be a better type variable name because this *is* the client.  In another comment you'll see I suggest adding another type variable for the Builder, which would naturally be "B".

dsmiley (changes requested):
Great stuff here!  I like how you moved lots of stuff to base classes thoughtfully.  Also like that some of the tests are re-used.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientJdkImpl.java
```diff
@@ -0,0 +1,443 @@
...
+import org.apache.solr.client.solrj.util.ClientUtils;
+import org.apache.solr.common.SolrException;
+import org.apache.solr.common.params.ModifiableSolrParams;
+import org.apache.solr.common.util.ContentStream;
+import org.apache.solr.common.util.NamedList;
+import org.apache.solr.common.util.ObjectReleaseTracker;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class HttpSolrClientJdkImpl extends Http2SolrClientBase {
```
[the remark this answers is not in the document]
jdyer1: Your name is at least as good as mine.  I renamed it.  Maybe we could consider someday have some kind of factory so users do not need to think as much about which implementation best meets their needs.  This would also make the implementation names matter less.

solr/solrj/src/java/org/apache/solr/client/solrj/embedded/SSLConfigBase.java
```diff
@@ -0,0 +1,75 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.solr.client.solrj.embedded;
+
+public class SSLConfigBase {
```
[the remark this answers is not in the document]
jdyer1: Ok, I added some very minimal Javadocs to this class.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientJdkImpl.java
```diff
@@ -0,0 +1,443 @@
...
+    if (builder.sslContext != null) {
+      b.sslContext(builder.sslContext);
+    }
+    if (builder.executor != null) {
+      b.executor(builder.executor);
+    }
+    if (builder.proxyHost != null) {
+      if (builder.proxyIsSocks4) {
+        log.warn(
+            "Socks4 is likely not supported by this client.  See https://bugs.openjdk.org/browse/JDK-8214516");
```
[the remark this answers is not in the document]
jdyer1: I was on the fence with this one.  I chose the log warning over the exception because (1) `Http2SolrClient` already does the same thing for `keyStoreReloadIntervalSecs`.  and (2) JDK-8214516 _could_ maybe get resolved in such a way that a future JDK could work with socks2 without any changes on our end.   But then again, looking at the last (5 year old) comment on that ticket, this doesn't seem likely!

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientJdkImpl.java
```diff
@@ -0,0 +1,443 @@
...
+        try {
+          resp.body().close();
+        } catch (Exception e1) {
+          // ignore
+        }
+      }
+    }
+  }
+
+  private static class PreparePostPutRequestResult {
```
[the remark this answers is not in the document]
jdyer1: Well, I really wish Java had tuples for this purpose.  I always hate creating a little class just so my private method can return two things.  To be less misleading I renamed it `PreparePostPutRequestReturnValue`.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientJdkImpl.java
jdyer1: Wouldn't the user need to start their application with an "--adds-opens" in order to be able to get into the implementation class?  If there is a way to do this without placing such burdens on users, then I'd agree.  I think the unit test at least demonstrates that everything is cleaned up after this falls out of scope, after a full gc.  That is the best we can do until we require Jdk21, no?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientJdkImpl.java
jdyer1: I simplified it and added a unit test.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientJdkImpl.java
jdyer1: I simplified it and added a unit test.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClientBase.java
jdyer1: This was missed when factoring out the base class.  I went ahead and move that getter from `Http2SolrClient` to here.

jdyer1:
I am reminded here that we have a **NOCOMMIT** in `solr-tests.policy`.  I am worried here that the only way to properly set the `java.net.URLPermission` is to enumerate all possible URL's (with ports) that might be hit during the test.  In this case, we might need to generate the policy file in the @beforeClass method.  I am hoping someone can suggest something easier!

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientJdkImpl.java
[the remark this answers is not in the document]
dsmiley: You've thought about it more than me; I defer to your judgement.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientJdkImpl.java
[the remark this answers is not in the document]
dsmiley: See `org.apache.solr.common.util.Pair`

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -239,16 +244,16 @@ private void setBasicAuthHeader(SolrRequest<?> solrRequest, HttpRequest.Builder
...
+  private static final Pattern MIME_TYPE_PATTERN = Pattern.compile("[^;]*");
 
   private String contentTypeToMimeType(String contentType) {
     Matcher mimeTypeMatcher = MIME_TYPE_PATTERN.matcher(contentType);
-    return mimeTypeMatcher.find() ? mimeTypeMatcher.group(1) : null;
+    return mimeTypeMatcher.find() ? mimeTypeMatcher.group() : null;
   }
 
-  private static final Pattern CHARSET_PATTERN = Pattern.compile("(?i)^.*charset=(.*)?(?:;| |$)");
+  private static final Pattern CHARSET_PATTERN = Pattern.compile("(?i)^.*charset=(.*)$");
```
dsmiley, line 254: I was thinking maybe `(?i)[; ]charset=([^;]*)`  but not important

solr/solrj/src/test/org/apache/solr/client/solrj/impl/DebugServlet.java
```diff
@@ -125,6 +126,11 @@ private void recordRequest(HttpServletRequest req, HttpServletResponse resp) {
     setParameters(req);
     setQueryString(req);
     setCookies(req);
+    try {
+     requestBody = req.getInputStream().readAllBytes();
```
dsmiley, line 130: some tidy formatting to do, I see

solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java
```diff
@@ -36,34 +36,32 @@
...
+public class Http2SolrClientTest extends Http2SolrClientTestBase {
 
   @Override
   protected String expectedUserAgent() {
     return "Solr[" + Http2SolrClient.class.getName() + "] 2.0";
   }
 
   @Override
-  protected <B extends HttpSolrClientBuilderBase> B builder(
-      String url, int connectionTimeout, int socketTimeout, Class<B> type) {
```
dsmiley: I'm on the fence of you removing `Class<B> type` for the builder.  I *think* this would enable you to remove all those SuppressWarning lines.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java
```diff
@@ -335,7 +327,7 @@ public void testCollectionParameters() throws IOException, SolrServerException {
 
   @Test
   public void testQueryString() throws Exception {
-    testQueryString(Http2SolrClient.class, Http2SolrClient.Builder.class);
+    super.testQueryString();
   }
```
dsmiley, line 331: I *think* this isn't needed, as long as testQueryString in the superclass is marked as a test like this one.

dsmiley (commented):
The builder change for generics turned out very nice :-)

What do you think of removing the "2" in some of these class names, leaving some in place for backwards-compatibility for now?

BTW I'm on vacation till mid next week so I'm slow to review.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClientBase.java
jdyer1: I removed the "2' from both base class names.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClientBase.java
```diff
@@ -371,6 +371,10 @@ public boolean isV2ApiRequest(final SolrRequest<?> request) {
     return request instanceof V2Request || request.getPath().contains("/____v2");
   }
 
+  public String getBaseURL() {
```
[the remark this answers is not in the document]
jdyer1: This was copied over from `Http2SolrClient` so to maintain backwards-compatibility we should keep the spelling the same.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
```diff
@@ -24,7 +24,8 @@
 import org.apache.solr.client.solrj.ResponseParser;
 import org.apache.solr.client.solrj.request.RequestWriter;
 
-public abstract class HttpSolrClientBuilderBase {
+public abstract class HttpSolrClientBuilderBase<
+    B extends HttpSolrClientBuilderBase<?, ?>, C extends Http2SolrClientBase> {
```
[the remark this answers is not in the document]
jdyer1: I did remove the "2" from the name.  However, any classes here I am naming `*Base` were factored out of something specific to `Http2SolrClient`.  I know there might have been some copy/paste when the various other clients were written but I did not check to see if we could create an abstract builder that applies to all of the clients.  That might be a worthy follow-up project though.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -75,9 +77,16 @@ protected HttpSolrClientJdkImpl(String serverBaseUrl, HttpSolrClientJdkImpl.Buil
     if (builder.sslContext != null) {
       b.sslContext(builder.sslContext);
     }
+
     if (builder.executor != null) {
-      b.executor(builder.executor);
+      this.executor = builder.executor;
+      this.shutdownExecutor = false;
+    } else {
+      this.executor = Executors.newCachedThreadPool(Executors.defaultThreadFactory());
```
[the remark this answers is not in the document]
jdyer1: I initially wanted to more-or-less mimic what `java.net.util.http.HttpClient` does if you do not pass in an Executor, and only reluctantly create my own because we need access to it for the pipes.  But deferring to `ForbiddenAPIs` I did switch it to the `ExecutorUtil` version (sorry I accidently pushed WIP here).  

I will mention here that I am a little concerned that `ExecutorUtil` has some things in it that really do not belong to the client, that exist only for inter-node requests.  For this reason I was hoping to avoid it.  See my other PR for one case where this could possibly cause trouble.  https://github.com/apache/solr/pull/2242

solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java
jdyer1: I will say that writing a generic builder in Java is no easy task!  Maybe I'm wrong on this but I do not think we can avoid unchecked casts here unless we require the user to pass in the target class as an additional parameter on each method.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java
jdyer1: I initially did it that way but then found that it confused IntelliJ.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientJdkImpl.java
```diff
@@ -0,0 +1,443 @@
...
+    if (isMultipart(streams)) {
+      throw new UnsupportedOperationException("This client does not support multipart.");
+    }
+
+    if (contentWriter != null) {
+      // TODO:  There is likely a more memory-efficient way to do this!
+      ByteArrayOutputStream baos = new ByteArrayOutputStream();
+      contentWriter.write(baos);
+      byte[] bytes = baos.toByteArray();
+      return new PreparePostPutRequestResult(
```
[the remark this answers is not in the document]
jdyer1: I have the Piped Stream modifications more-or-less implemented.  However, I have another NOCOMMIT here as clearly I am missed something!

solr/solrj/src/java/org/apache/solr/client/solrj/embedded/SSLConfigBase.java
```diff
@@ -40,35 +52,75 @@ public SSLConfigBase(
...
+   */
   public void setClientAuth(boolean clientAuth) {
     this.clientAuth = clientAuth;
   }
 
-  /** All other settings on this object are ignored unless this is true */
+  /**
+   * All other settings on this object are ignored unless this is true
+   *
+   * @return boolean
```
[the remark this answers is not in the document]
jdyer1: I removed `SSLConfigBase` and reverted `SSLConfig` back.  My initial thought is the base class could be shared between the two clients but it did not work out that way.  Now we need not worry about the javadoc.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -188,33 +210,52 @@ private HttpResponse<InputStream> doPutOrPost(
...
-      ByteArrayOutputStream baos = new ByteArrayOutputStream();
-      contentWriter.write(baos);
-      byte[] bytes = baos.toByteArray();
-      bodyPublisher = HttpRequest.BodyPublishers.ofByteArray(bytes);
+      source = new PipedOutputStream();
+      PipedInputStream sink = new PipedInputStream(source);
+      bodyPublisher = HttpRequest.BodyPublishers.ofInputStream(() -> sink);
+      // NOCOMMIT - this does not work.  The call to "client.send" sometimes will throw an
+      // IOException
+      // because our thread closes the outputstream too early.  But if we wait to close it until
```
dsmiley, line 220: Hmm; doesn't make sense to me.  Writer should close to signify there's no more data.  The reader will eventually see this and return -1 bytes read to signify EOF.  I reviewed the code for that as well.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -188,33 +210,52 @@ private HttpResponse<InputStream> doPutOrPost(
...
     decorateRequest(reqb, solrRequest);
     reqb.uri(new URI(url + "?" + queryParams));
-    return client.send(reqb.build(), HttpResponse.BodyHandlers.ofInputStream());
+    HttpResponse<InputStream> response =
+        client.send(reqb.build(), HttpResponse.BodyHandlers.ofInputStream());
+    try {
+      source.close();
+    } catch (Exception e1) {
+      // ignore
+    }
```
dsmiley, line 257: I don't think this is necessary; source should already be closed.  But a concern is that, what if send() or other code fails... we want that executor thread to cancel/stop.  I think a Future should be kept so that we can cancel it in a try-finally.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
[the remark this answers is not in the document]
janhoy: It would be good to have a clearer distinction between the solrj classes that are meant for end user consumption vs the ones that are only ever to be used by the server.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -64,7 +66,7 @@ public class HttpSolrJdkClient extends HttpSolrClientBase {
   private static final String USER_AGENT =
       "Solr[" + MethodHandles.lookup().lookupClass().getName() + "] 1.0";
 
-  private HttpClient client;
+  protected HttpClient client;
```
dsmiley, line 69: minor, but can we call this `httpClient` to make it clearer wherever we use it what sort of client we're talking about?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -108,6 +118,29 @@ protected HttpSolrJdkClient(String serverBaseUrl, HttpSolrJdkClient.Builder buil
     }
     this.client = b.build();
     updateDefaultMimeTypeForParser();
+
+    // This is a workaround for the case where the first request using
```
dsmiley, line 122: Awesome comment/explanation.
Should this ping be gated on http2 only?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -210,50 +243,50 @@ private HttpResponse<InputStream> doPutOrPost(
...
-      source.close();
-    } catch (Exception e1) {
-      // ignore
+      response = client.send(reqb.build(), HttpResponse.BodyHandlers.ofInputStream());
+    } catch (IOException | InterruptedException | RuntimeException e) {
+      if (contentWritingFuture != null) {
+        contentWritingFuture.cancel(true);
+      }
+      throw e;
     }
```
dsmiley: It's okay to cancel a task that has already completed.  So we can simplify this to move the logic to a finally.  No need to catch exceptions.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/HttpSolrJdkClientTest.java
```diff
@@ -278,6 +281,15 @@ public void testUpdateDefault() throws Exception {
 
   @Test
   public void testUpdateXml() throws Exception {
+    testUpdateXml(false);
+  }
+
+  @Test
+  public void testUpdateXmlWithHttp11() throws Exception {
+    testUpdateXml(true);
+  }
```
dsmiley: Could we parameterize the test instead so that the whole test runs in both modes?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
```diff
@@ -38,7 +37,7 @@ public abstract class HttpSolrClientBuilderBase<
   protected Set<String> urlParamNames;
   protected Integer maxConnectionsPerHost;
   protected ExecutorService executor;
-  protected CookieStore cookieStore;
+  protected boolean useHttp1_1 = Boolean.getBoolean("solr.http1");
```
dsmiley: I like it but I can't help but notice that no other field here has a default that is defined here.  So _maybe_ this new setting should be initialized elsewhere?

solr/solrj/src/test/org/apache/solr/client/solrj/impl/HttpSolrJdkClientTest.java
dsmiley: Or just use randomization -- easy and adequate even if a bit annoying to debug a failure.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
jdyer1: I removed this NOCOMMIT and removed the unnecessary second close on the OutputStream.  I also changed it to hold onto the Future so as to cancel on exception.   I also made various other improvements.

The OutputStream was being closed early, but only on non-secure http2 requests.  If using TLS or forcing Http1.1, it always worked.  The problem was that with the first request, the JDK Http Client sends it as Http1.1 and requests an upgrade.  When that first request happens to be a POST with a chunked body, it evidently sends the body with the upgrade request.  Our Jetty server, however, does not allow a body with an upgrade request, so it replies with a GOAWAY frame.  The workaround I found is to issue a "ping" request prior to the first real one.  This lets the JDK Client know if our server supports Http2 in advance and upgrade requests are avoided after that.

I have two regrets with this round of changes: first, I cannot figure out a reliable way to write a unit test for the error handling when it calls "cancel" on the Future.  Second, I am calling "ping" in the constructor, which I felt was a lesser evil than having to check a volatile boolean with each request to see if "ping" was called yet.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -0,0 +1,453 @@
...
+              60,
+              TimeUnit.SECONDS,
+              queue,
+              new SolrNamedThreadFactory(this.getClass().getSimpleName()));
+      this.shutdownExecutor = true;
+    }
+    b.executor(this.executor);
+
+    if (builder.useHttp1_1) {
+      b.version(HttpClient.Version.HTTP_1_1);
```
dsmiley, line 105: If you expressly set the version to HTTP 2, can we avoid the ping altogether?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -0,0 +1,453 @@
...
+    validateGetRequest(solrRequest);
+    reqb.GET();
+    decorateRequest(reqb, solrRequest);
+    reqb.uri(new URI(url + "?" + queryParams));
+    return client.send(reqb.build(), HttpResponse.BodyHandlers.ofInputStream());
+  }
+
+  private HttpResponse<InputStream> doPutOrPost(
+      String url,
+      boolean isPut,
```
dsmiley, line 220: instead of a boolean, can you use the enum reference?  Boolean args are not very clear.  And this would allow you to combine both case statements back in the switch.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
jdyer1: This was moved from `Http2SolrClient` so I left it as-is.  I initially did not see any value in supporting this in the new client.  I reconsidered after realizing the stream-cancelling bug was http/2-specific.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/HttpSolrJdkClientTest.java
jdyer1: I just read up a little on Junit4 Parameterized tests, and well, it seems so much more complicated than what I have done here.  Yes, randomization will make the test suite run some ms faster, but I'd rather better know that nothing broke with 1 green test run.  In my opinion while randomization is very useful we perhaps overuse it sometimes.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
jdyer1: yes, good _catch_ here.  I changed it.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
jdyer1: I changed the name as you suggested.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
jdyer1: Not sure it buys much, but I went ahead and made this change.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -0,0 +1,449 @@
...
+    //
+    // By sending a ping before any real requests occur, the client
+    // knows if Solr can accept Http/2, and no additional
+    // upgrade requests will be sent.
+    //
+    // See https://bugs.openjdk.org/browse/JDK-8287589
+    // See https://github.com/jetty/jetty.project/issues/9998#issuecomment-1614216870
+    //
+    try {
+      ping();
```
janhoy, line 140: > Second, I am calling "ping" in the constructor, which I felt was a lesser evil than having to check a volatile boolean with each request to see if "ping" was called yet.

It would be somewhat unexpected to have the client start sending calls to the server by just constructing it. May it even cause real issues since startup order between a customer client application and the solr server would then become significant?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
dsmiley: I would rather we avoid the ping at the expense of the possibility that a minority of users would be required to configure HTTP 1.1 explicitly on their client.  After all, we know Solr supports Http/2 and it's been out there for quite some time.  I can understand why the JDK's HttpClient defaults the way it does (kind of in the middle) but we know how we deploy Solr for the majority of users.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -268,13 +258,14 @@ private HttpResponse<InputStream> doPutOrPost(
       bodyPublisher = HttpRequest.BodyPublishers.noBody();
     }
 
+    decorateRequest(reqb, solrRequest);
     if (method == SolrRequest.METHOD.PUT) {
-      reqb.PUT(bodyPublisher);
+      reqb.method("PUT", bodyPublisher);
     } else {
-      reqb.POST(bodyPublisher);
+      reqb.method("POST", bodyPublisher);
```
dsmiley, line 265: could use one line (no if-else) via method.toString to get the string of the method

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -287,6 +278,64 @@ private HttpResponse<InputStream> doPutOrPost(
     return response;
   }
 
+  private volatile boolean headRequested;
+  private volatile boolean headSucceeded;
```
dsmiley, line 282: volatile isn't needed because it's only accessed by one method (which ought to be mentioned in a comment so this invariant holds)

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
jdyer1: We can't tell it to always use Http/2.  Although this feature was requested (see https://bugs.openjdk.org/browse/JDK-8287589) it does not look like the JDK client will support this any time soon.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
jdyer1: I think these booleans need to go away anyhow.  My reasoning here is we cannot guarantee the client will behave the same every time with this 1-shot try with a HEAD request.  What if there is a network issue?  Then that client instance is stuck on Http/1.1 for its lifetime for these requests.

I think it is probably best to issue the HEAD request every time, to guarantee consistency.  This only affects POST requests with a chunked body (updates, not queries).  It only will be needed if not TLS, and if the user does not force Http/1.1.  I do not think it adds anymore overhead than if it always issued an upgrade request with no body (as the Eclipse maintainers seem to think it ought to.)  We can put a note about this in the Javadoc and if none of these options are workable for any user, they always have `Http2SolrClient` to fall back on.  

Progress, not Perfection, eh?

dsmiley: Does a "HEAD request every time" -- do you mean when the client is opened/instantiated?  Fine; agree to "progress not perfection".  It can be improved later driven by actual vs perceived problems.

janhoy: I'm ok with HEAD every time as a start. Then perhaps we can make it more clever over time. And if/when JDK improves, we can detect JDK version and conditionally force HTTP/2.

This also means we can avoid sending ping in constructor, right?

jdyer1: By "every time" I mean with every POST request that meets the criteria.  (chunked body; no TLS; not forcing Http/1.1).  This generally will mean every Solr Update will incur both an Http HEAD and an Http POST.  This should be no different if the JDK Client behaved as the Jetty maintainers think it should:  to incur both an Empty Upgrade then the Actual Request.  It is true most of the time the Http HEAD request is unnecessary.  However, I do not know how we know _when_ it is unnecessary.  So at the expense of performance, we get it to work correctly on a consistent basis.  

Maybe for a Version Two, we could get into retry logic, where if it fails, then try a HEAD request and then try the POST once again.  But the more I think about what it would take to do this right the less motivated I feel to go down this path.

dsmiley: I'd much prefer the boolean return to ensure we do this once and not every time.  Any issues with that?

jdyer1: The issue with doing it just once is it cannot guarantee the client will behave the same every time.  If we try once with the first request, we are at the mercy of luck as to whether or not it could fail for some unrelated reason.  Then the client is downgraded to Http/1.1 for its lifetime (which could be the life of the application).  I can see future bug reports along the lines of "sometimes, after application restart, my solr client downgrades to Http/1.1 for an unknown reason and then I have to restart again to fix it."

I think the main use-case for chunked body POST requests are index updates, and most users would batch these, so the cost of an additional HEAD request might be negligible?  

Maybe I am missing something here, perhaps there is a better way to ensure consistency without adding (much) cost?

dsmiley: I don't think anybody would notice an HTTP 2 to HTTP 1.1 fallback.  It works.

jdyer1: I left this as it is and it only will complete a HEAD request the very first time a POST request with a chunked body is executed.  If the HEAD request succeeds it lets the jdk client send upgrade requests from then on.  If it fails, it forces http/1.1 for all POST requests with a chunked body.  I did add an INFO logging message so users might get a clue of what is going on in these instances.

None of this applies if the request uses TLS.  For secure requests, Jetty and the JDK are in agreement on the correct way to negotiate the protocol so all of this is unnecessary.  Of course, if the user requests to always downgrade to http/1.1 this log is skipped as well.

jdyer1:
Today I did my best in covering all the possible security manager permissions in *solr-tests.policy*.  The *NOCOMIT* with the "all permission" is gone.  I will be surprised though once we have CI runs with this if we do not see failures due to the policy file not handling some other variation of "localhost".

Besides this, I will try to write something up for the reference guide and then if there are no objections I would like to merge this to main soon, and ?perhaps maybe to branch_9x as well?

gradle/testing/randomization/policies/solr-tests.policy
dsmiley, line 1: This is only for tests; must we be so specific on the URLs?

I'd like to remind us why we even use the security manager for tests in the first place.  It's for the file system to tame tests.  Makes sense; a clever solution.  I never recall anyone advocating that more than that is important.  Yet it was a slippery slope of "hey why not" for all these other things.  But it's plain here that this is burdensome for us to maintain; not to mention Oracle tells us it's all going away anyhow.  I'll raise this on the dev list.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -64,6 +62,12 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * A SolrClient implementation that communicates to a Solr server using the built-in Java 11+ Http
+ * Client. This client is targeted for those users who wish to minimize application dependencies.
+ * This client will connect to solr using Http/2 but can seamlessly downgrade to Http/1.1 when
+ * connecting to Solr hosts running on older versions.
+ */
 public class HttpSolrJdkClient extends HttpSolrClientBase {
```
dsmiley, line 71: If I suggested the "Jdk" between Solr & Client then I misspoke; I recall HttpJdkSolrClient or "JdkHttpSolrClient" (I like the second) because "SolrClient" is the important base type here.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -278,64 +285,74 @@ private HttpResponse<InputStream> doPutOrPost(
...
    *
-   *    We only try once, and if it fails, we downgrade to Http/1
+   * <p>We only try once, and if it fails, we downgrade to Http/1
    *
    * @param url the url with no request parameters
    * @return true if success
    */
-  private synchronized boolean maybeTryHeadRequest(String url) {
-    if(headRequested) {
+  private boolean maybeTryHeadRequest(String url) {
```
dsmiley, line 307: Would protected scope be useful to allow a user to change?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -278,64 +285,74 @@ private HttpResponse<InputStream> doPutOrPost(
...
       return headSucceeded;
     }
+
     URI uriNoQueryParams;
     try {
       uriNoQueryParams = new URI(url);
-    } catch(URISyntaxException e) {
-      headSucceeded = false;
+    } catch (URISyntaxException e) {
+
```
dsmiley, line 326: remove blank

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
```diff
@@ -278,64 +285,74 @@ private HttpResponse<InputStream> doPutOrPost(
...
+    } catch (InterruptedException ie) {
       Thread.currentThread().interrupt();
       headSucceeded = false;
     } finally {
+
+      // The HEAD request is tried only once.  All future requests will skip this check.
       headRequested = true;
+
+      if (!headSucceeded) {
+        log.info("All insecure POST requests with a chunked body will use http/1.1");
```
dsmiley, line 349: Would replacing "insecure" with "HTTP" or "unencrypted" or "non-SSL" be fine?  To me, there's something about the word "insecure" that refers to overall security posture, thus implying the user is negligent here which is not necessarily the case.

dsmiley (commented):
+1 for main & 9x; basically looks ready to me notwithstanding the naming choice.

Thanks for your efforts here!

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
jdyer1: Yes, this is a good idea.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
jdyer1: I can change this to _unencrypted_.  Let's try not to offend our users, eh?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrJdkClient.java
dsmiley: Cool; that was my second preference.

gradle/testing/randomization/policies/solr-tests.policy
janhoy: According to [docs](https://docs.oracle.com/en%2Fjava%2Fjavase%2F21%2Fdocs%2Fapi%2F%2F/java.base/java/net/URLPermission.html), it looks like a `-` acts as a recursive wildcard. I also think you could use wildcard for method, e.g.

```
permission "java.net.URLPermission" "https://127.0.0.1:*/solr/-", "*:*";
```

jdyer1: Yes, the dash wildcard works.  I stared at that document so long and do not know how I missed that line.  I appreciate your mentioning this and will commit this change to simplify the policy file.  It reduces it to just 9 lines.  I listed out the methods out of a desire to minimize use of wildcards where I could without needing to add lines to the file.

dsmiley:
@jdyer1  Lets use a PR to collaborate on workaroundToReleaseThreads_noClosableUntilJava21.  Getting all threads and interrupting each is a hack I'd like to try and avoid.  I don't recall any test doing that despite us having a crazy number of tests.

dsmiley:
Instead of simply interrupting the threads, shouldn't we also join on them?
Should that be in an AfterClass -- is it a problem between each test?
note: you don't need to gather a Set to then iterate; you can call `forEach` on the stream to do whatever you want to do.  I don't think using a Set (for uniqueness) is necessary here either.
What is the GC accomplishing?

I was looking again at all those test overrides that call super.  I removed them locally and I was able to run at the CLI and IntelliJ and it ran all the intherited tests.  This pattern is done elsewhere like SolrExampleTests

PR #2260, opened 2024-02-13
solr.xml: honor plugin enable=true|false

dsmiley:
Plugins can be disabled via `enable="false"` but unfortunately it's up to the caller of creating a PluginInfo to remember to consult this flag.  This PR addresses plugin disablement in solr.xml stuff.  No test, I know... it's not like every possible other case is tested :-).  Nonetheless I will manually test one of these soon that I have an interest in disabling via config before merging the PR.  Seems so minor, maybe no CHANGES.txt.


pjmcarthur (approved):

cpoerschke:
Wondering about having similar at lines 646 and 805 in this file.

dsmiley:
I forgot to push the BackupRepository case yesterday; did it now.  Converting to Java streams shows very nice results there too :-)

cpoerschke (approved):

PR #2262, opened 2024-02-14
NO JIRA: remove duplicate '(cmd.getFlags() & GET_SCORES) != 0' calls

cpoerschke:
@chatman included a few as part of #2248 and looking further there's a few more, so separating them out here. No JIRA ticket or CHANGES.txt entry needed I think.


gerlowskija:
> No JIRA ticket or CHANGES.txt entry needed I think
+1 - it'd be overkill here IMO

gerlowskija (approved):
LGTM - nice little cleanup!

PR #2263, opened 2024-02-14
Clarify split parameter usage

radu-gheorghe:
It's a minor change that we tested (if you just do `split=true` it seems to apply to all fields, which is rarely what you want).


HoustonPutman (approved):

HoustonPutman:
This should be reflected in the official 9.5 docs after tonight!

radu-gheorghe:
Oh, wow. That's fast :) Thanks!

PR #2271, opened 2024-02-15
SOLR-17160: Time based tracking of core admin requests

psalagnac:
https://issues.apache.org/jira/browse/SOLR-17160

# Description

Core admin completed/failed requests are tracked in-memory, but we only track a maximum of 100 requests by dropping the oldest ones. If for any reason we drop a request before the client (an other Solr node) fetched the status, we will get unexpected errors at the top level command.

# Solution

Instead of the hard limit of 100 tracked requests, we track all requests with a expiration timeout. This is a trade-off to not keep memory forever and to be sure we won't lose status in the case of many incoming requests.

The change adds two different timeouts:
- an hour for every requests (overridable with property `"solr.admin.requests.running.timeout.minutes"`)
- 5 minutes for completed/failed requests once the status is already retrieved by client (overridable with property `"solr.admin.requests.completed.timeout.minutes"`)

# Tests

Added a unit test in `CoreAdminHandlerTest` to check expired requests are not tracked in-memory anymore.

Also checks we don't have any issue for collection backup with very high numbers of shards.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/admin/CoreAdminHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/CoreAdminHandler.java
@@ -28,13 +28,13 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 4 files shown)


dsmiley (commented):
Instead of the manual cache management, can't we use CaffeineCache as is used elsewhere in Solr?  CC @bruno-roustant

PR #2274, opened 2024-02-16
SOLR-6572: Trim whitespace in XML nodes in solrconfig.xml

janhoy:
https://issues.apache.org/jira/browse/SOLR-6572

This PR is based on the patch uploaded to JIRA issue, modified to apply to main.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -110,7 +110,7 @@ Bug Fixes
 
 * SOLR-17148: Fixing Config API overlay property enabling or disabling the cache (Sanjay Dutt, hossman, Eric Pugh)
 
-* SOLR-6572: Extra leading or trailing whitespace in XML nodes in solrconfig.xml not supported
+* SOLR-6572: Extra leading or trailing whitespace in XML nodes in solrconfig.xml now ignored
   (Fredrik Rødland, Anurag Sharma via janhoy)
 
 Dependency Upgrades
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -110,6 +110,9 @@ Bug Fixes
 
 * SOLR-17148: Fixing Config API overlay property enabling or disabling the cache (Sanjay Dutt, hossman, Eric Pugh)
 
+* SOLR-6572: Extra leading or trailing whitespace in XML nodes in solrconfig.xml not supported
+  (Fredrik Rødland, Anurag Sharma via janhoy)
+
 Dependency Upgrades
 ---------------------
 (No changes)
--- a/solr/solrj/src/java/org/apache/solr/common/util/DOMUtil.java
+++ b/solr/solrj/src/java/org/apache/solr/common/util/DOMUtil.java
@@ -321,7 +321,7 @@ public class DOMUtil {
       case Node.CDATA_SECTION_NODE: /* fall through */
       case Node.COMMENT_NODE: /* fall through */
       case Node.PROCESSING_INSTRUCTION_NODE: /* fall through */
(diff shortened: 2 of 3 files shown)


epugh (approved):
LGTM...

Another example of why we should not do our own xml parsing!

solr/CHANGES.txt
```diff
@@ -110,6 +110,9 @@ Bug Fixes
 
 * SOLR-17148: Fixing Config API overlay property enabling or disabling the cache (Sanjay Dutt, hossman, Eric Pugh)
 
+* SOLR-6572: Extra leading or trailing whitespace in XML nodes in solrconfig.xml not supported
```
cpoerschke, line 113:
```suggestion
* SOLR-6572: Extra leading or trailing whitespace in XML nodes in solrconfig.xml is now ignored
```

or

```suggestion
* SOLR-6572: Extra leading or trailing whitespace in XML nodes in solrconfig.xml is now supported by being ignored
```

cpoerschke (approved):

janhoy:
Unfortunately several tests now fail due to the change. Meaning that the `DOMUtil` class is probably also used outside of config parsing.

```
  - org.apache.solr.update.processor.CloneFieldUpdateProcessorFactoryTest.testCloneCombinations (:solr:core)
  - org.apache.solr.update.processor.FieldMutatingUpdateProcessorTest.testComprehensive (:solr:core)
  - org.apache.solr.update.processor.FieldMutatingUpdateProcessorTest.testConcatExplicitWithDelim (:solr:core)
  - org.apache.solr.spelling.suggest.TestAnalyzedSuggestions.classMethod (:solr:core)
  - org.apache.solr.spelling.suggest.TestFuzzyAnalyzedSuggestions.classMethod (:solr:core)
  - org.apache.solr.spelling.suggest.TestBlendedInfixSuggestions.classMethod (:solr:core)
  - org.apache.solr.spelling.suggest.TestFreeTextSuggestions.classMethod (:solr:core)
  - org.apache.solr.spelling.suggest.TestHighFrequencyDictionaryFactory.classMethod (:solr:core)
  - org.apache.solr.spelling.suggest.TestPhraseSuggestions.classMethod (:solr:core)
  - org.apache.solr.spelling.suggest.TestFileDictionaryLookup.classMethod (:solr:core)
  - org.apache.solr.spelling.suggest.TestAnalyzeInfixSuggestions.classMethod (:solr:core)
```

It could be that this is only in our test code, have not dived deep. Anyone knows?

janhoy:
The failure in `TestFreeTextSuggestions` is due to the need to configure a SPACE as separator, see https://github.com/apache/solr/blob/main/solr/core/src/test-files/solr/collection1/conf/solrconfig-phrasesuggest.xml#L152

So with the general `strip()` the space is reduced to empty string and we get an IAOOB exception. Options:

* Only trim leading/trailing newline
* Leave as is and make code reading config from XML more robust and do a local `trim()` where crucial, e.g. if the value is a URI?

cpoerschke:
> ... Leave as is and make code reading config from XML more robust and do a local ...

Intuitively I'd prefer that, not for the leave-as-is reason but the parser parsing on an as-is basis seems clearest and any handling of spaces and trimming etc. seems more of an application level logic thing i.e. for the code reading the config.

janhoy:
The spec https://www.w3.org/TR/xml/#sec-white-space seems to suggest that white space handling is configurable, and that trimming leading/trailing whitespace stripping is the default. Again, would be much better to just rely on some OOTB XML lib than hacking away on this ourselves. I'll close this as not being the solution.

PR #1114, opened 2022-10-23
Link Solr Operator for learning on Solr in Kubernetes, and better exp…

jebnix:
…lain the relevancy of docker-networking

https://issues.apache.org/jira/browse/SOLR-XXXXX

# Description

The Docker networking page seems to have a pretty old description. While this page is still relevant (mainly for local development), it is important to link Solr Operator to the subject. In general, it seems like a pretty blog kind of a page, therefore I've also removed the "I" and "I'll" words, and replaced them with "We" and "We'll" which seems more professional.

# Solution

described

# Tests

No

# Checklist

Please review the following and check all that apply:

- [x] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [x] I have created a Jira issue and added the issue ID to my pull request title.
- [x] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [x] I have developed this patch against the `main` branch.
- [x] I have run `./gradlew check`.
- [ ] I have added tests for my changes.
- [x] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

PR #2277, opened 2024-02-18
Clean up docker-faq.adoc

janhoy:
Spinoff from #1114 which will clean up One sub chapter, while this cleans up the rest of the file.

### DIFF
--- a/solr/solr-ref-guide/modules/deployment-guide/pages/docker-faq.adoc
+++ b/solr/solr-ref-guide/modules/deployment-guide/pages/docker-faq.adoc
@@ -18,140 +18,34 @@
 
 == How do I persist Solr data and config?
 
-Your data is persisted already, in your container's filesystem.
-If you `docker run`, add data to Solr, then `docker stop` and later `docker start`, then your data is still there.
-The same is true for changes to configuration files.
-
-Equally, if you `docker commit` your container, you can later create a new
-container from that image, and that will have your data in it.
-
-For some use-cases it is convenient to provide a modified `solr.in.sh` file to Solr.
-For example to point Solr to a ZooKeeper host:
-
-[source,bash]
-----
-docker create --name my_solr -P solr
-docker cp my_solr:/opt/solr/bin/solr.in.sh .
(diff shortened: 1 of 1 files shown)


epugh (approved):
LGTM..    This looks so much nicer..  It felt like developer notes and now seems more FAQ..    I always struggled to understand them...

janhoy:
> It felt like developer notes and now seems more FAQ

I actually think the FAQ text copied over from the old `docker-solr` project was written even before our `Dockerfile` had the `VOLUME` declaration, and before data folder was moved to `/var/solr`. So lots of doc rot there.

syats:
The fact that this PR is not yet in the productive documentation https://solr.apache.org/guide/solr/latest/deployment-guide/docker-faq.html as of 2023-03-07 is a source of confusion! The current documentation does not reflect the move of the data folder!

epugh:
I don't believe changes to ref guides go immediately (unlike the website changes)...    We don't have a "quick push this change to productioin ref guide" that isn't tied to a release.   Hopefully 9.6 will be out soon, and I don't think this will justify a 9.5.x release...

janhoy:
Hmm, if I'm not mistaken, a merge of this fix to `branch_9_5` may actually update the 9.5 guide. Let me try.

epugh:
If this works...    I wonder if we should try an establish a stronger pattern for doc fixes going `main` --> `branch_9x` and --> `branch_9_5` ?   Though it is a whole nother step....

janhoy:
I did a push, let's wait for daily build of https://ci-builds.apache.org/job/Solr/job/solr-reference-guide-official/ - I think it will kick off in an hour or so.

janhoy:
The theory behind the guide build is that there is not supposed to be any functional change within a minor version, so any fixes pushed to `branch_9_5` are by definition bug fixes, including fixes to ref guide. This makes it safe to build the guide immediately even without a code release.

We could even use this to push "Errata" entries to the guide after we discover bugs in a release that will not be fixed in an upcoming release...

HoustonPutman:
> Hmm, if I'm not mistaken, a merge of this fix to `branch_9_5` may actually update the 9.5 guide. Let me try.

You are not mistaken!

PR #2279, opened 2024-02-19
SOLR-17166 cloud.sh can't put solr.xml in zk anymore

gus-asf:
https://issues.apache.org/jira/browse/SOLR-17166

### DIFF
--- a/dev-tools/scripts/cloud.sh
+++ b/dev-tools/scripts/cloud.sh
@@ -41,8 +41,7 @@
 # By default the script sets up a local Solr cloud with 4 nodes, in a local
 # directory with ISO date as the name. A local zookeeper at 2181 or the
 # specified port is presumed to be available, a new zk chroot is used for each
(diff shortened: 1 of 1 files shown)


epugh (approved):

epugh:
I wonder if some of what cloud.sh does could be delegated to other scripts that get more testing?  To make it less brittle?

janhoy (approved):

gus-asf:
> I wonder if some of what cloud.sh does could be delegated to other scripts that get more testing? To make it less brittle?

Some things to think about: 

- It's only intended for dev testing (so bin/solr is not a place for it I think). 
- It's dependent on knowing where to find a checkout of solr
- It's got routines that clear out data from a zk root (dangerous)

So these things are part of why it's on it's own. If we have other dev testing support it could be coordinated with that, but IMHO this script should never come anywhere near a production anything.

PR #2283, opened 2024-02-19
SOLR-17066: Switch "LB" clients away from core URLs

gerlowskija:

# Description

Bootstrapping Solr clients with core URLs has several downsides: it prevents the created client from making core-agnostic requests, it prevents the client from making both v1 and v2 requests, etc.  For all these reasons, SOLR-17066 has been moving our client implementations away from using these "core URLs" and towards using a "default collection" instead.

**Solr's "Load balancing" client is an interesting case though.**  A common (maybe even the primary?) use-case of the "LB" client is to round-robin requests across a set of replicas.  Each replica in a set has a unique core name, and distinguishing between them is important to ensure uniform load distribution.  So users can't rely on the "default collection" feature to meet their needs.

Any attempt to move the LBClients away from "core URLs" must also retain a way for them to specify a distinct set of replicas/cores.

# Solution

This PR solves this problem by creating a new LBClient-specific abstraction called an "endpoint".  Endpoints are thin wrappers around a Solr base URL and optional core name, and users now provide these "Endpoint" objects when creating their LBClient, instead of passing raw String URLs.

Using a strongly typed object instead of raw String URLs allows core-based clients to still make core-agnostic requests as needed, while still supporting the "target a set of distinct replicas" use case.

# Tests

Changes to `TestLBHttpSolrClient` and `TestLBHttp2SolrClient`.  New unit tests for the `Endpoint` class.  Existing tests continue to pass.

# Checklist

- [x] I have added Javadoc documentation


gerlowskija:
**Question** Is this change worth the added complexity?  I'm excited to move away from core URLs, but want to make sure the net effect is still positive given some of the complexity this pulls in.  Curious what others think. 

If we do go this route, this change should target `main`/10.0 **only**; I don't think 9.x is an option, since we still support arbitrary "context root"s on that branch.  (Though of course, we'll need to make some sort of branch_9x change to deprecate the LBClient method signatures that this PR modifies)

PR #2285, opened 2024-02-19
ReqSupplier interface not actually needed, so simplify code

epugh:
# Description

The `ReqSupplier` interface isn't actually needed, we only have one implementation, and it lends some more indirection for the reader to wade through.

This is not end user facing so no JIRA needed.

# Solution

Remove it.

# Tests

There are no `ReqSupplier` specific tests, so removing it and relying on existing tests to confirm it works!


dsmiley (approved):

gerlowskija (commented):
Belated "LGTM" - thanks @epugh !

PR #2289, opened 2024-02-20
SOLR-17156: introducing RandomNoReverseMergePolicyFactory

mkhludnev:
healing BJQParserTest

https://issues.apache.org/jira/browse/SOLR-17156

### DIFF
--- a/solr/test-framework/src/java/org/apache/solr/util/RandomNoReverseMergePolicyFactory.java
+++ b/solr/test-framework/src/java/org/apache/solr/util/RandomNoReverseMergePolicyFactory.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.util;
+
+import org.apache.lucene.index.MergePolicy;
+import org.apache.solr.index.MergePolicyFactory;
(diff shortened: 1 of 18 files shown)


mkhludnev:
It needs to be spread across other block join tests as well.

janhoy (approved):
Cool!

mkhludnev:
The scope has blown up. Are there other tests rely on block alignment?

solr/test-framework/src/java/org/apache/solr/util/RandomNoReverseMergePolicyFactory.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+package org.apache.solr.util;
+
+import org.apache.lucene.index.MergePolicy;
+import org.apache.solr.index.MergePolicyFactory;
+import org.apache.solr.index.MergePolicyFactoryArgs;
+
+/** A {@link MergePolicyFactory} for {@link RandomMergePolicy}. */
+public final class RandomNoReverseMergePolicyFactory extends MergePolicyFactory {
```
dsmiley, line 24: I don't understand how this prevents "ReverseMergePolicy"

mkhludnev: Here we go 
https://github.com/apache/solr/pull/2289/files/d24bfff1fe1aaa11a666c8f7188895ce23eed128#diff-681b5f32ae2e1f988d11d5112305b5c71e08b3c3abee5879db27a9b91d63b398R46

mkhludnev: @dsmiley thanks for the guidance. I'll check it out.

mkhludnev:
@dsmiley thanks for the guidance. I'll check it out.

mkhludnev:
> SystemPropertiesRestoreRule instead. See its constructor that takes the key and value. 

I'm ok with this proposal @dsmiley. But I can't find any usage of two args constructor, so far all tests use `systemSetPropertySolrTestsMergePolicyFactory`. As far as I understand `SystemPropertiesRestoreRule` rule already applied and restores props thus, I don't need to bother about restoring in `@AfterClass` - ok. 
Then if I use this rule, I need to copy-paste prop name from `systemSetPropertySolrTestsMergePolicyFactory` that doesn't seem good for me. if this bi-arg constructor works, should I use `RuleChain` or just `@Rule`?     

Note: I changed `SolrTestCase` just to fix its' subclasses which don't inherit `SolrTestCase4J`.

dsmiley (approved):
Nice

mkhludnev:
I think it's test only change, thus it doesn't deserves CHANGES entry

PR #2293, opened 2024-02-22
SOLR-17176: Log history V2 API doesn't serialize properly

magibney:
fix `LogWatcher.unmodifiable()` to be backed by a full-featured Map

`SolrDocument.getFieldValuesMap()` provides a feature-poor Map instance that also does undesired extra manipulations on values.

https://issues.apache.org/jira/browse/SOLR-17176

### DIFF
--- a/solr/core/src/test/org/apache/solr/logging/TestLogWatcher.java
+++ b/solr/core/src/test/org/apache/solr/logging/TestLogWatcher.java
@@ -16,17 +16,31 @@
  */
 package org.apache.solr.logging;
 
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.StringWriter;
 import java.lang.invoke.MethodHandles;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import org.apache.solr.SolrTestCaseJ4;
+import org.apache.solr.common.MapWriter;
 import org.apache.solr.common.SolrDocument;
 import org.apache.solr.common.SolrDocumentList;
+import org.apache.solr.common.params.ModifiableSolrParams;
 import org.apache.solr.common.util.TimeSource;
+import org.apache.solr.request.SolrQueryRequest;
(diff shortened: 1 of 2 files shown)


solr/core/src/test/org/apache/solr/logging/TestLogWatcher.java
```diff
@@ -104,4 +121,56 @@ public void testLog4jWatcher() throws InterruptedException {
       oldMessages.add(msg);
     }
   }
+
+  /**
+   * Here we validate that serialization works as expected for several different methods. Ideally we
+   * would use actual serialization from Jersey/Jackson, since this is what really happens in V2
+   * APIs. But this is simpler, and should give us roughly equivalent assurances.
+   */
```
magibney, line 129: Does this seem reasonable? Am I missing a straightforward and/or better way to directly validate Jersey/Jackson serialization? cc @gerlowskija

magibney:
I guess just because the results are +/- analogous to SolrDocuments results, in terms of serialized output. numFound, etc. For the purpose of this PR I was trying to leave that unchanged. I think the baggage with serialization is less from the "SolrDocuments" approach _per se_, and more around the exact way in which the initial PR went about constructing the SolrDocuments to be immutable (since they are reused across multiple requests).

The biggest problem with the initial PR really was its use of `doc.getFieldValueMap()`, which afaict (I belatedly realize) seems not designed to be used outside of a test context.

solr/core/src/test/org/apache/solr/logging/TestLogWatcher.java
gerlowskija: Apologies for not seeing this sooner @magibney 

I like what you have here, for a "unit" level test.

The alternative would be to test this higher up - say, by using SolrJettyTestCase and using SolrJ to hit the v2 logging API.  That'd test the serialization by virtue of testing the whole endpoint.

But I like your more-specific approach here better 👍

magibney: No problem, and thanks for the review/suggestion! I'll probably go ahead and commit as-is; have to make sure this gets in 9.6.

PR #2297, opened 2024-02-24
SOLR-16780: Prevent ZK subcommands from being run without the bin/solr zk prefix

epugh:
https://issues.apache.org/jira/browse/SOLR-16780

# Description

bin/solr cp would run, but the params wouldn't be properly formatted and you get an error.

# Solution

Trap the use of the zk subcommands and give user appropriate message.

# Tests

Bats

### DIFF
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -873,6 +873,12 @@ if [[ "$SCRIPT_CMD" == "delete" ]]; then
   exit $?
 fi
 
+# Prevent any zk subcommands from going through with out invoking zk command
+if [[ "$SCRIPT_CMD" == "upconfig" || $SCRIPT_CMD == "downconfig" || $SCRIPT_CMD == "cp" || $SCRIPT_CMD == "rm" || $SCRIPT_CMD == "mv" || $SCRIPT_CMD == "ls" || $SCRIPT_CMD == "mkroot" ]]; then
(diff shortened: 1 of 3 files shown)


solr/bin/solr
```diff
@@ -873,6 +873,12 @@ if [[ "$SCRIPT_CMD" == "delete" ]]; then
   exit $?
 fi
 
+# Prevent any zk subcommands from going through with out invoking zk command
+if [[ "$SCRIPT_CMD" == "upconfig" || $SCRIPT_CMD == "downconfig" || $SCRIPT_CMD == "cp" || $SCRIPT_CMD == "rm" || $SCRIPT_CMD == "mv" || $SCRIPT_CMD == "ls" || $SCRIPT_CMD == "mkroot" ]]; then
```
janhoy, line 877: Feels a bit of duplication, for every sub command of any command we need to add it here to forbid it?
Cannot this be handled in SolrCLI, or is the main-command handled by shell and sub-commands by Java?

epugh: Yeah, I did not like it.  main is shell and sub commands are java...  so thats why `bin/solr cp` actually is executed in java code.   Now, if we move this all over to more how the `bin/solr package <subcommand>` then we wouldn't need this.

janhoy (approved):
Let's get this small improvement in and then hope to push more cli parsing logic pushed over in Java land over time..

PR #2304, opened 2024-02-26
SOLR-17160: Time based tracking of core admin requests with Caffeine cache

psalagnac:
https://issues.apache.org/jira/browse/SOLR-17160

This is a simpler version of #2271, that now uses Caffeine cache instead of manually dropping old requests.

# Description

Core admin completed/failed requests are tracked in-memory, but we only track a maximum of 100 requests by dropping the oldest ones. If for any reason we drop a request before the client (an other Solr node) fetched the status, we will get unexpected errors at the top level command.

# Solution

Instead of the hard limit of 100 tracked requests, we track all requests in a Caffeine cache, configured to drop old requests.
This is a trade-off to not keep memory forever and to be sure we won't lose status in the case of many incoming requests.

The change adds two different timeouts:
- an hour for every requests (overridable with property `"solr.admin.requests.running.timeout.minutes"`)
- 5 minutes for completed/failed requests once the status is already retrieved by client (overridable with property `"solr.admin.requests.completed.timeout.minutes"`)

# Tests

Added a unit test in `CoreAdminHandlerTest` to check expired requests are not tracked in-memory anymore.

Also checks we don't have any issue for collection backup with very high numbers of shards.

# Checklist


solr/core/src/java/org/apache/solr/handler/admin/CoreAdminHandler.java
```diff
@@ -497,42 +533,19 @@ public void submitAsyncTask(TaskObject taskObject) throws SolrException {
...
-      }
-    }
-
-    private void ensureTaskIdNotInUse(String taskId) throws SolrException {
-      if (getRequestStatusMap(RUNNING).containsKey(taskId)
-          || getRequestStatusMap(COMPLETED).containsKey(taskId)
-          || getRequestStatusMap(FAILED).containsKey(taskId)) {
+    private void addTask(TaskObject taskObject) {
+      // Ensure task ID is not already in use
+      if (requestStatusCache.asMap().containsKey(taskObject.taskId)) {
```
dsmiley, line 538: I think there's a race where this check is false for 2 threads and both get entered into the cache via put.  If you instead call `get` (with lambda to supply the new value) , you can check if the returned value is the new taskObject or is an existing one.

solr/core/src/java/org/apache/solr/handler/admin/api/GetNodeCommandStatus.java
```diff
@@ -51,25 +51,24 @@ public GetNodeCommandStatus(
...
+
+    if (status == null) {
       requestStatusResponse.status = "notfound";
       requestStatusResponse.msg = "No task found in running, completed or failed tasks";
+    } else {
+      requestStatusResponse.status = status;
+
+      if (taskObject.getStatus().equals(COMPLETED)) {
+        requestStatusResponse.response = taskObject.getRspObject();
+        requestStatusResponse.response = taskObject.getOperationRspObject();
```
dsmiley, line 66: I'm confused; you are setting the response twice?

solr/core/src/java/org/apache/solr/handler/admin/api/GetNodeCommandStatus.java
```diff
@@ -51,25 +51,24 @@ public GetNodeCommandStatus(
...
+    TaskObject taskObject = coreAdminAsyncTracker.getAsyncRequestForStatus(requestId);
+    String status = taskObject != null ? taskObject.getStatus() : null;
+
+    if (status == null) {
       requestStatusResponse.status = "notfound";
       requestStatusResponse.msg = "No task found in running, completed or failed tasks";
+    } else {
+      requestStatusResponse.status = status;
+
+      if (taskObject.getStatus().equals(COMPLETED)) {
```
dsmiley, line 64: can reference `status`

solr/core/src/test/org/apache/solr/handler/admin/CoreAdminHandlerTest.java
```diff
@@ -528,4 +534,39 @@ public void testNonexistentCoreReload() throws Exception {
         e.getMessage());
     admin.close();
   }
+
+  @Test
+  public void testTrackedRequestExpiration() throws Exception {
```
dsmiley, line 539: wonderful test

solr/core/src/java/org/apache/solr/handler/admin/CoreAdminHandler.java
```diff
@@ -546,6 +559,8 @@ public static class TaskObject {
       final Callable<SolrQueryResponse> task;
       public String rspInfo;
       public Object operationRspInfo;
+      private volatile String status;
+      private volatile boolean polled;
```
dsmiley, line 563: Could use javadoc.  From observing you code, this is set to true once the status has been requested AND it is no longer RUNNING.  Maybe a different name like `polledAfterCompletion` would be clearer.

solr/core/src/java/org/apache/solr/handler/admin/api/GetNodeCommandStatus.java
psalagnac: Oh, good catch!
Something is definitely wrong here, but this is existing code. I did not notice that.
I think this deserves a a fix by itself. Values for `msg` and `response` in the V2 API response are not consistent depending on the status. And it seems to me there not consistent with V1 either.
I think this was introduced with #2144.

Maybe I should put this PR on hold and fix this first?

dsmiley: Ah; didn't know it was existing code.  Do in whichever order you want; it doesn't really matter.

solr/core/src/java/org/apache/solr/handler/admin/CoreAdminHandler.java
psalagnac: Good catch, thanks.
I'm also adding a test for this since the logic is now slightly more complicated (a simple test, can't easily check the race).

solr/core/src/java/org/apache/solr/handler/admin/CoreAdminHandler.java
```diff
@@ -497,42 +534,23 @@ public void submitAsyncTask(TaskObject taskObject) throws SolrException {
...
-          || getRequestStatusMap(FAILED).containsKey(taskId)) {
+      // If we get a different task instance, it means one was already in the cache with the
+      // same name. Just reject the new one.
+      if (taskInCache != taskObject) {
         throw new SolrException(
             ErrorCode.BAD_REQUEST, "Duplicate request with the same requestid found.");
       }
+
+      taskObject.status = RUNNING;
+      requestStatusCache.put(taskObject.taskId, taskObject);
```
dsmiley, line 549: The call to `get` with the lambda above (L539) will insert it atomically, so you shouldn't try to put it *again* here, thus maybe having some races.  Generally try to avoid double operations with a cache -- use one (atomic) operation.  Also, you should set the status to RUNNING before then so that a race will be sure to see RUNNING.

dsmiley: In the lambda providing the new taskObject, you could set to RUNNING there.

solr/core/src/java/org/apache/solr/handler/admin/CoreAdminHandler.java
```diff
@@ -448,25 +459,51 @@ public static class CoreAdminAsyncTracker {
...
     public CoreAdminAsyncTracker() {
-      HashMap<String, Map<String, TaskObject>> map = new HashMap<>(3, 1.0f);
-      map.put(RUNNING, Collections.synchronizedMap(new LinkedHashMap<>()));
-      map.put(COMPLETED, Collections.synchronizedMap(new LinkedHashMap<>()));
-      map.put(FAILED, Collections.synchronizedMap(new LinkedHashMap<>()));
-      requestStatusMap = Collections.unmodifiableMap(map);
+      this(
+          Ticker.systemTicker(),
+          TimeUnit.MINUTES.toNanos(
+              EnvUtils.getEnvAsLong("solr.admin.requests.running.timeout.minutes", 60L)),
```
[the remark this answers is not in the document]
psalagnac: Thanks. I wasn't aware of the two different methods.

solr/core/src/java/org/apache/solr/handler/admin/CoreAdminHandler.java
psalagnac: Good catch. I forgot to remove the call to `put` after the last change.

solr/core/src/java/org/apache/solr/handler/admin/CoreAdminHandler.java
```diff
@@ -436,7 +436,8 @@ public static class CoreAdminAsyncTracker {
      * we're not supposed to hit it. This is just a protection to grow in memory too much when
      * receiving an abusive number of admin requests.
      */
-    private static final int MAX_TRACKED_REQUESTS = 10_000;
+    private static final int MAX_TRACKED_REQUESTS =
+        EnvUtils.getPropertyAsInteger("solr.admin.requests.running.max", 10_000);
```
dsmiley, line 440: As this is only for "async" requests, maybe the name should incorporate this limit like 'solr.admin.async.max". ("requests" and "running" being kind of implicit maybe)?

solr/core/src/java/org/apache/solr/handler/admin/CoreAdminHandler.java
```diff
@@ -546,7 +553,6 @@ private void addTask(TaskObject taskObject) {
       }
 
       taskObject.status = RUNNING;
```
dsmiley, line 555: Isn't it no longer necessary to mark as RUNNING here since it's already done so in the lambda?

psalagnac: Thanks. Forgot once again to remove old line :facepalm:

dsmiley:
WDYT of this proposed CHANGES.txt under Improvements:
> SOLR-17160: Core Admin "async" request status tracking is no longer capped at 100; it's 10k.  Statuses are now removed 5 minutes after the read of a completed/failed status.   (Pierre Salagnac)

dsmiley:
Perhaps an additional sentence to highlight that this helps collection backup/restore and other operations scale to 100+ shards.

dsmiley:
I handled CHANGES.txt myself.  I'll merge this sometime this weekend.  Looking forward to seeing this one done!

psalagnac:
> I handled CHANGES.txt myself. I'll merge this sometime this weekend. Looking forward to seeing this one done!

Thanks for handling this @dsmiley. I was off for some time and unable to do any change.

PR #2311, opened 2024-02-27
SOLR-6572 Trim whitespace for shards.tolerant and leaderUrl

janhoy:
https://issues.apache.org/jira/browse/SOLR-6572

WIP


epugh:
Seems fine...   Worth committing?

epugh (approved):
LGTM.  I mean, I hate that we have to trim things on a one by one basis, it just seems like something "we shouldn't ahve to thing about"..  But, I guess if I am going to complain about whitespace, I should also argue about the existence of Null in Java and NullPointerExceptions too!

janhoy:
Yea, looks quite safe. Won't fix any whitespace issues in solrconfig.xml but this does not seem to have been a huge issue these last 10 years either. Will merge.

PR #2313, opened 2024-02-27
SOLR-16932: expose `AsyncTracker`'s max outstanding async req limit config in `Http2SolrClient`

tuannh982:
addressing https://issues.apache.org/jira/browse/SOLR-16932

# Description

expose `AsyncTracker`'s max outstanding async req limit config in `Http2SolrClient`

# Solution

# Tests

# Checklist

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
@@ -1287,9 +1287,7 @@ public class Http2SolrClient extends SolrClient {
       return this;
     }
 
-    /**
-     * The maximum outstanding async requests, default is 1000 requests
-     */
+    /** The maximum outstanding async requests, default is 1000 requests */
     public Builder withMaxOutstandingRequests(int max) {
       this.maxOutstandingAsyncRequests = max;
       return this;
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
@@ -1287,7 +1287,10 @@ public class Http2SolrClient extends SolrClient {
       return this;
     }
 
-    /** The maximum outstanding async requests, default is 1000 requests */
+    /**
+     * Specify the maximum number of concurrent asynchronous requests the client can send. The
+     * default value is 1000
+     */
     public Builder withMaxOutstandingRequests(int max) {
       this.maxOutstandingAsyncRequests = max;
       return this;
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
@@ -146,7 +146,7 @@ public class Http2SolrClient extends SolrClient {
 
   protected RequestWriter requestWriter = new BinaryRequestWriter();
   private List<HttpListenerFactory> listenerFactory = new ArrayList<>();
-  private final AsyncTracker asyncTracker = new AsyncTracker();
+  private final AsyncTracker asyncTracker;
 
   /** The URL of the Solr server. */
   private final String serverBaseUrl;
@@ -175,6 +175,12 @@ public class Http2SolrClient extends SolrClient {
     }
 
     this.idleTimeoutMillis = builder.idleTimeoutMillis;
+
+    if (builder.maxOutstandingAsyncRequests != null) {
+      this.asyncTracker = new AsyncTracker(builder.maxOutstandingAsyncRequests);
+    } else {
+      this.asyncTracker = new AsyncTracker(AsyncTracker.DEFAULT_MAX_OUTSTANDING_REQUESTS);
+    }
 
     if (builder.httpClient != null) {
       this.httpClient = builder.httpClient;
@@ -1017,19 +1023,21 @@ public class Http2SolrClient extends SolrClient {
   }
 
   private static class AsyncTracker {
-    private static final int MAX_OUTSTANDING_REQUESTS = 1000;
+    private static final int DEFAULT_MAX_OUTSTANDING_REQUESTS = 1000;
 
     // wait for async requests
     private final Phaser phaser;
+    private final int maxOutstandingRequests;
     // maximum outstanding requests left
     private final Semaphore available;
     private final Request.QueuedListener queuedListener;
     private final Response.CompleteListener completeListener;
 
-    AsyncTracker() {
+    AsyncTracker(int maxOutstandingRequests) {
       // TODO: what about shared instances?
+      this.maxOutstandingRequests = maxOutstandingRequests;
       phaser = new Phaser(1);
-      available = new Semaphore(MAX_OUTSTANDING_REQUESTS, false);
+      available = new Semaphore(maxOutstandingRequests, false);
       queuedListener =
           request -> {
             phaser.register();
@@ -1048,7 +1056,7 @@ public class Http2SolrClient extends SolrClient {
 
     int getMaxRequestsQueuedPerDestination() {
       // comfortably above max outstanding requests
-      return MAX_OUTSTANDING_REQUESTS * 3;
+      return maxOutstandingRequests * 3;
     }
 
     public void waitForComplete() {
@@ -1065,6 +1073,7 @@ public class Http2SolrClient extends SolrClient {
     private Long connectionTimeoutMillis;
     private Long requestTimeoutMillis;
     private Integer maxConnectionsPerHost;
+    private Integer maxOutstandingAsyncRequests;
     private String basicAuthAuthorizationStr;
     private boolean useHttp1_1 = Boolean.getBoolean("solr.http1");
     private Boolean followRedirects;
@@ -1275,6 +1284,14 @@ public class Http2SolrClient extends SolrClient {
      */
     public Builder withMaxConnectionsPerHost(int max) {
       this.maxConnectionsPerHost = max;
+      return this;
+    }
+
+    /**
+     * The maximum outstanding async requests, default is 1000 requests
+     */
+    public Builder withMaxOutstandingRequests(int max) {
+      this.maxOutstandingAsyncRequests = max;
       return this;
     }
 


solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -1278,6 +1287,12 @@ public Builder withMaxConnectionsPerHost(int max) {
       return this;
     }
 
+    /** The maximum outstanding async requests, default is 1000 requests */
```
jdyer1, line 1290: Follow standard javadoc convention here.

jdyer1 (commented):
This seems like a straightforward change, but I am curious about the problem you are encountering that motivates it?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
tuannh982: oops I don't know why `gradlew tidy` do collapse the comment, let me update the comment to make it longer than 1 line

tuannh982:
> This seems like a straightforward change, but I am curious about the problem you are encountering that motivates it?

We just need to throttle the incoming request to Solr server. The async request method throttle value is hard-coded (1000) so we have to wrap the sync request method in Future and also have a semaphore to limit the maximum concurrency.

tuannh982:
@jdyer1 can you re-review this?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -1278,6 +1287,15 @@ public Builder withMaxConnectionsPerHost(int max) {
       return this;
     }
 
+    /**
```
jdyer1, line 1290: This should have both a "@param" and a "@return" line as part of the javadoc, with a few words describing each.

jdyer1 (commented):
I think if this was to be considered to be merged, we first need unit test coverage on the new builder method and that the new value makes it way to the built `AsyncTracker`,  and that the default is used when not specificed.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -1017,19 +1023,21 @@ public String getBaseURL() {
   }
 
   private static class AsyncTracker {
-    private static final int MAX_OUTSTANDING_REQUESTS = 1000;
+    private static final int DEFAULT_MAX_OUTSTANDING_REQUESTS = 1000;
```
dsmiley, line 1026: While we're at it, this would be a good use of `EnvUtils` like "solr.solrj.async.max.outstanding.requests"

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -1278,6 +1287,15 @@ public Builder withMaxConnectionsPerHost(int max) {
       return this;
     }
 
+    /**
+     * Specify the maximum number of concurrent asynchronous requests the client can send. The
+     * default value is 1000
+     */
+    public Builder withMaxOutstandingRequests(int max) {
```
pvcnt, line 1294: If I'm not mistaken it's actually a maximum amount per host. Could we make it explicit, e.g., `maxOutstandingRequestsPerHost`?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -1065,6 +1073,7 @@ public static class Builder {
     private Long connectionTimeoutMillis;
     private Long requestTimeoutMillis;
     private Integer maxConnectionsPerHost;
+    private Integer maxOutstandingAsyncRequests;
```
pvcnt, line 1076: `maxOutstandingAsyncRequests` is inconsistent with `withMaxOutstandingRequests` (about the async part). I think we can drop the async part, since even synchronous requests use the threshold returned by `getMaxRequestsQueuedPerDestination`.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -1017,19 +1023,21 @@ public String getBaseURL() {
...
     private final Request.QueuedListener queuedListener;
     private final Response.CompleteListener completeListener;
 
-    AsyncTracker() {
+    AsyncTracker(int maxOutstandingRequests) {
       // TODO: what about shared instances?
+      this.maxOutstandingRequests = maxOutstandingRequests;
       phaser = new Phaser(1);
-      available = new Semaphore(MAX_OUTSTANDING_REQUESTS, false);
+      available = new Semaphore(maxOutstandingRequests, false);
```
pvcnt, line 1040: Not related to your review, but this semaphore is not working at all. `queuedListener` is called after the request may have been rejected, which essentially makes it useless.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
magibney: 💯 -- it seems like this PR is motivated by a "custom solr client" use case? But unless we make this configurable in a regular deployment (as would be achieved by `EnvUtils`), there's no way to configure the clients that are used for intra-cluster communication, which was the concern that initially motivated the filing of SOLR-16932.

In fact, an `EnvUtils`-type approach was the _only_ approach taken in [the patch we've been using to mitigate this issue locally](https://github.com/cowpaths/fullstory-solr/pull/124).

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
magibney: I'm all but certain that the semaphore _is_ working -- in fact I think it's the only thing that practically needed to be changed in our [local mitigation](https://github.com/cowpaths/fullstory-solr/pull/124). Notably, that mitigation changed `getMaxRequestsQueuedPerDestination()` to return a constant (3k -- effectively unchanged from current `main`) independent of the configuration of `maxOutstandingRequests`. There's some [relevant discussion](https://github.com/cowpaths/fullstory-solr/pull/124#discussion_r1294915329) there that explores the distinction between `maxOutstandingRequests` as enforced by the `available` Semaphore (across all destinations), vs. `getMaxRequestsQueuedPerDestination()` which, as the method name suggests, is per destination.

pvcnt: I have already observed in production exceptions thrown because the internal queue of the HTTP2 client was full. It should not have been the case if the semaphore was working correctly, should it? When I was debugging the code, I found out that the `queuedListener` is called after the request had a change to get rejected, which means that the semaphore does not have a chance to protected this from happening. Or am I missing the point of this semaphore?

magibney: It's been a while since I was in this code, but iirc (and thinking back to the "relevant discussion" comment linked above), there are _2_ queues: 1 is a per-destination buffer (before send), the other is the actual "request enqueued to be sent" queue, which is the one to which the semaphore applies, and covers _all_ destinations.

So I suspect that what happened was that the second queue (to which the semaphore applies) is actually throttling the request, which then causes the buffers to back up to the point where the requests get proactively rejected, which is what you would have been seeing in production.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
HoustonPutman: That sounds good, but I'd remove the `solrj` part of the sys prop

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -1048,7 +1056,7 @@ private static class AsyncTracker {
 
     int getMaxRequestsQueuedPerDestination() {
```
HoustonPutman, line 1057: Yeah I'm pretty confused here why the async tracker would be a global queue, and it's maximum is used to determine the max requests queued per destination?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
HoustonPutman: well I don't think this is per-host. The Async tracker doesn't do any per-host logic, does it?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
pvcnt: This is used to set Jetty's `maxRequestsQueuedPerDestination`, which is indeed per host. But the AsyncTracker itself has no per-host logic. However, it depends in which context it is used. When used by the LbSolrClient (notably in the context of the HttpShardHandler, but only when used directly by any client), there is effectively one Http2SolrClient created per host, so this matches. In other contexts, I believe there is a mismatch and it doesn't do what it claims to do.

magibney: Yes. again following on my earlier dive into this, my conclusion was that `maxRequestsQueuedPerDestination` is close to irrelevant as a user-configurable value. This is counterintuitive because when we actually get an exception, the exception is because this queue has overflowed. But there's no actual throttling behavior implemented by this queue, and it really makes more sense to think of the per-host setting as a buffer, to _very_ temporarily hold requests in between them being submitted for sending, and them actually being sent.

It's illustrative to consider that whether or not these two values are independently configurable, you could max out the configured global outstanding request limit with requests to a _single_ host, which would cause the `maxRequestsQueuedPerDestination` queues to overflow for _all_ hosts, throwing exceptions, but without any requests actually having been sent to any host other than the host the single host that consumed all the configured global semaphore permits. IMO in this context a configurable per destination "limit" is not really useful for anything, and is certainly confusing.

The [approach I'd recommend](https://github.com/cowpaths/fullstory-solr/pull/124#discussion_r1294915329) is to have the per-destination limit be hardcoded (it's useless at best, misleading at worst), and clarify that the configurable value affects connection throttling over _all_ hosts:

>I think the right way to think about maxRequestsQueuedPerDestination is: How many requests (per destination) is it reasonable to buffer, figuring "I'll send this as soon as I can" vs failing fast and sending an error code to the client indicating that we're too backed up right now. I could even see just making this a static value (e.g., the 3k that it has effectively been up to this point). I think I was thrown by the "comfortably above max outstanding requests" comment though, implying some relationship there. Tbh I can't think of any reason why this should even necessarily be "above" max outstanding requests, let alone "comfortably above" ...

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
magibney: Yeah, it's really 2 things:
1. throttles maximum number of outstanding requests across all hosts
2. the number of requests per host that will be buffered waiting to send, once 1. hits its limit.

2. does buffering, but no actual throttling, and IMO it's not useful. If we proceed with making the "per host" part hardcoded (removing 2.), then the javadocs here for `withMaxOutstandingRequests(int)` will be completely accurate. And as it stands I'd say they're essentially accurate, practically speaking, since the per-host logic that they do influence is more of an internal detail.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
magibney: afaict this is in fact standard javadoc formatting, judging by existing patterns for this project anyway. I've observed that it's not uncommon, at least for cases where  the arg and return value are obvious from the text description, for `@param` and `@return` to be omitted, and for shorter javadocs to be collapsed to a single line. No need to fight w/ `gradlew tidy` here 🙂

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
dsmiley: Isn't the "solrj" part important though?  The component name following "solr" is important to categorize this setting.  Simply omitting "solrj" would make "async" the category, which is a bad name IMO.  I think "solrj" is a good category name and it should be chosen.  See https://cwiki.apache.org/confluence/display/SOLR/System+property+naming+structure
CC @janhoy

dsmiley:
_Commenting to prevent auto-close._

janhoy:
Was this PR left to die on purpose or should it have a exemt-stale label? Do we have more insight of this issue now two years down the road?

PR #2316, opened 2024-02-28
[SOLR-17187] Add possibility to supply a custom commitPollInterval for TLOG/PULL replicas

tboeghk:
https://issues.apache.org/jira/browse/SOLR-17187

# Description

In TLOG/PULL replica setups with longer commit intervals it would be great to configure the interval to check for new commits manually to tune to your setup. Currently on the PULL replica the interval to check for new commits is 
[computed using the commit / soft commit settings](https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java#L83-L87)

# Solution

You can now supply a custom `commitPollInterval` in your `updateHandler` definition. If supplied, this takes precendence over the one calculated from hard/soft commit settings.

```xml
<updateHandler class="solr.DirectUpdateHandler2">

    <!-- For TLOG replicas, updateLog must be enabled -->
    <updateLog>
        <str name="dir">${solr.ulog.dir:}</str>
    </updateLog>

    <!-- custom poll interval fix -->
    <commitPollInterval>00:00:10</commitPollInterval>
</updateHandler>
```

# Tests

None :-/

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
@@ -74,14 +74,18 @@ public class ReplicateFromLeader {
               "SolrCore not found:" + coreName + " in " + CloudUtil.getLoadedCoreNamesAsString(cc));
         }
       }
+
       SolrConfig.UpdateHandlerInfo uinfo = core.getSolrConfig().getUpdateHandlerInfo();
+      String customPollInterval =
+          core.getSolrConfig().get("updateHandler").get("commitPollInterval").txt();
       String pollIntervalStr = "00:00:03";
+      String calculatedPollIntervalString = determinePollInterval(uinfo);
+
       if (System.getProperty("jetty.testMode") != null) {
         pollIntervalStr = "00:00:01";
-      }
-
-      String calculatedPollIntervalString = determinePollInterval(uinfo);
-      if (calculatedPollIntervalString != null) {
+      } else if (customPollInterval != null) {
+        pollIntervalStr = customPollInterval;
+      } else if (calculatedPollIntervalString != null) {
         pollIntervalStr = calculatedPollIntervalString;
       }
       log.info("Will start replication from leader with poll interval: {}", pollIntervalStr);
--- a/solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
@@ -75,16 +75,8 @@ public class ReplicateFromLeader {
         }
       }
 
-      SolrConfig.UpdateHandlerInfo uinfo = core.getSolrConfig().getUpdateHandlerInfo();
-      String pollIntervalStr = "00:00:03";
-      if (System.getProperty("jetty.testMode") != null) {
-        pollIntervalStr = "00:00:01";
-      }
-
-      String calculatedPollIntervalString = determinePollInterval(uinfo);
-      if (calculatedPollIntervalString != null) {
-        pollIntervalStr = calculatedPollIntervalString;
-      }
+      final SolrConfig.UpdateHandlerInfo uinfo = core.getSolrConfig().getUpdateHandlerInfo();
+      final String pollIntervalStr = determinePollInterval(uinfo);
       log.info("Will start replication from leader with poll interval: {}", pollIntervalStr);
 
       NamedList<Object> followerConfig = new NamedList<>();
@@ -134,20 +126,23 @@ public class ReplicateFromLeader {
   }
 
   /**
-   * Determine the poll interval for replicas based on the auto soft/hard commit schedule
+   * Determine the poll interval for replicas based on the auto soft/hard commit schedule or
+   * configured commit poll interval
    *
    * @param uinfo the update handler info containing soft/hard commit configuration
    * @return a poll interval string representing a cadence of polling frequency in the form of
-   *     hh:mm:ss
+   *     hh:mm:ss, never <code>null</code>
    */
   public static String determinePollInterval(SolrConfig.UpdateHandlerInfo uinfo) {
     int hardCommitMaxTime = uinfo.autoCommmitMaxTime;
     int softCommitMaxTime = uinfo.autoSoftCommmitMaxTime;
     boolean hardCommitNewSearcher = uinfo.openSearcher;
     String customCommitPollInterval = uinfo.commitPollInterval;
-    String pollIntervalStr = null;
+    String pollIntervalStr = "00:00:03";
 
-    if (customCommitPollInterval != null) {
+    if (System.getProperty("jetty.testMode") != null) {
+      pollIntervalStr = "00:00:01";
+    } else if (customCommitPollInterval != null) {
       pollIntervalStr = customCommitPollInterval;
     } else if (hardCommitMaxTime != -1) {
       // configured hardCommit places a ceiling on the interval at which new segments will be
@@ -172,6 +167,8 @@ public class ReplicateFromLeader {
       pollIntervalStr = toPollIntervalStr(softCommitMaxTime / 2);
     }
 
+    // validate poll interval and fail early
+    ReplicationHandler.readIntervalNs(pollIntervalStr);
     return pollIntervalStr;
   }
 
--- a/solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
@@ -76,16 +76,13 @@ public class ReplicateFromLeader {
       }
 
       SolrConfig.UpdateHandlerInfo uinfo = core.getSolrConfig().getUpdateHandlerInfo();
-      String customPollInterval =
-          core.getSolrConfig().get("updateHandler").get("commitPollInterval").txt();
       String pollIntervalStr = "00:00:03";
-      String calculatedPollIntervalString = determinePollInterval(uinfo);
-
       if (System.getProperty("jetty.testMode") != null) {
         pollIntervalStr = "00:00:01";
-      } else if (customPollInterval != null) {
-        pollIntervalStr = customPollInterval;
-      } else if (calculatedPollIntervalString != null) {
+      }
+
+      String calculatedPollIntervalString = determinePollInterval(uinfo);
+      if (calculatedPollIntervalString != null) {
         pollIntervalStr = calculatedPollIntervalString;
       }
       log.info("Will start replication from leader with poll interval: {}", pollIntervalStr);
@@ -147,8 +144,12 @@ public class ReplicateFromLeader {
     int hardCommitMaxTime = uinfo.autoCommmitMaxTime;
     int softCommitMaxTime = uinfo.autoSoftCommmitMaxTime;
     boolean hardCommitNewSearcher = uinfo.openSearcher;
+    String customCommitPollInterval = uinfo.commitPollInterval;
     String pollIntervalStr = null;
-    if (hardCommitMaxTime != -1) {
+
+    if (customCommitPollInterval != null) {
+      pollIntervalStr = customCommitPollInterval;
+    } else if (hardCommitMaxTime != -1) {
       // configured hardCommit places a ceiling on the interval at which new segments will be
       // available to replicate
       if (softCommitMaxTime != -1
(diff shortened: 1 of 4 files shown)


solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
```diff
@@ -74,14 +74,18 @@ public void startReplication(boolean switchTransactionLog) {
               "SolrCore not found:" + coreName + " in " + CloudUtil.getLoadedCoreNamesAsString(cc));
         }
       }
+
       SolrConfig.UpdateHandlerInfo uinfo = core.getSolrConfig().getUpdateHandlerInfo();
+      String customPollInterval =
+          core.getSolrConfig().get("updateHandler").get("commitPollInterval").txt();
```
cpoerschke, line 80: implementation detail: might a `UpdateHandlerInfo.commitPollInterval` member be an alternative?

solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
```diff
@@ -74,14 +74,18 @@ public void startReplication(boolean switchTransactionLog) {
...
-      }
-
-      String calculatedPollIntervalString = determinePollInterval(uinfo);
-      if (calculatedPollIntervalString != null) {
+      } else if (customPollInterval != null) {
+        pollIntervalStr = customPollInterval;
+      } else if (calculatedPollIntervalString != null) {
         pollIntervalStr = calculatedPollIntervalString;
       }
       log.info("Will start replication from leader with poll interval: {}", pollIntervalStr);
```
cpoerschke, line 91: logging: wondering about somewhere (not necessarily here) recording the custom interval and what it would have been otherwise.

solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
justinrsweeney: +1 on that, then I think in determinePollInterval you could just account for the configured commitPollInterval

epugh:
Looks like some feedback @tboeghk that could be addressed and then maybe this would be ready for commit!

epugh:
Does this change need a unit test?

tboeghk:
@cpoerschke is absolutely right, I added it "the hacky way". I'll extract the config setting the proper way 👍

With that, a unit test is easy to add @epugh 👍

I'll update the docs also as soon as I'm back at my desk (after Easter)

epugh:
@tboeghk  I'm still interested in this one getting in ;-).  And it's post 🐰 season!  ;-)

tboeghk:
Finally made those changes 🙏

solr/core/src/java/org/apache/solr/core/SolrConfig.java
```diff
@@ -810,6 +810,7 @@ public static class UpdateHandlerInfo implements MapSerializable {
     public final boolean openSearcher; // is opening a new searcher part of hard autocommit?
     public final boolean commitWithinSoftCommit;
     public final boolean aggregateNodeLevelMetricsEnabled;
+    public final String commitPollInterval;
```
cpoerschke, line 813: Wondering about string vs. something else here? E.g. something like this maybe:
```suggestion
    public final int commitPollIntervalSeconds;
```
or
```suggestion
    public final Integer commitPollIntervalSeconds;
```

cpoerschke: Never mind, it is a string where it is e.g. https://github.com/apache/solr/blob/releases/solr/9.7.0/solr/core/src/java/org/apache/solr/handler/ReplicationHandler.java#L251-L253

cpoerschke, line 813: minor/subjective, to retain (partial) match order with constructor args
```suggestion
    public final String commitPollInterval;
    public final boolean aggregateNodeLevelMetricsEnabled;
```

solr/core/src/java/org/apache/solr/core/SolrConfig.java
```diff
@@ -836,6 +838,8 @@ public UpdateHandlerInfo(
 
       this.commitWithinSoftCommit = commitWithinSoftCommit;
       this.aggregateNodeLevelMetricsEnabled = false;
+
+      this.commitPollInterval = commitPollInterval;
```
cpoerschke, line 842:
```suggestion
      this.commitPollInterval = commitPollInterval;
      this.aggregateNodeLevelMetricsEnabled = false;
```

solr/core/src/java/org/apache/solr/core/SolrConfig.java
```diff
@@ -852,6 +856,7 @@ public UpdateHandlerInfo(ConfigNode updateHandler) {
           updateHandler.get("commitWithin").get("softCommit").boolVal(true);
       this.aggregateNodeLevelMetricsEnabled =
           updateHandler.boolAttr("aggregateNodeLevelMetricsEnabled", false);
+      this.commitPollInterval = updateHandler.get("commitPollInterval").txt();
```
cpoerschke, line 859:
```suggestion
      this.commitPollInterval = updateHandler.get("commitPollInterval").txt();
      this.aggregateNodeLevelMetricsEnabled =
          updateHandler.boolAttr("aggregateNodeLevelMetricsEnabled", false);
```

solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
```diff
@@ -143,8 +144,12 @@ public static String determinePollInterval(SolrConfig.UpdateHandlerInfo uinfo) {
     int hardCommitMaxTime = uinfo.autoCommmitMaxTime;
     int softCommitMaxTime = uinfo.autoSoftCommmitMaxTime;
     boolean hardCommitNewSearcher = uinfo.openSearcher;
+    String customCommitPollInterval = uinfo.commitPollInterval;
```
cpoerschke, line 147: minor: javadoc update to reflect change in behaviour e.g.



```
- Determine the poll interval for replicas based on the auto soft/hard commit schedule

+ Determine the poll interval for replicas based on the auto soft/hard commit schedule or configured commit poll interval
```

solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
```diff
@@ -74,6 +74,7 @@ public void startReplication(boolean switchTransactionLog) {
               "SolrCore not found:" + coreName + " in " + CloudUtil.getLoadedCoreNamesAsString(cc));
         }
       }
+
       SolrConfig.UpdateHandlerInfo uinfo = core.getSolrConfig().getUpdateHandlerInfo();
       String pollIntervalStr = "00:00:03";
       if (System.getProperty("jetty.testMode") != null) {
```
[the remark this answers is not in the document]
tboeghk: Yes, makes sense 👍

solr/core/src/test/org/apache/solr/cloud/ReplicateFromLeaderTest.java
```diff
@@ -28,44 +28,64 @@ public class ReplicateFromLeaderTest {
...
+            -1,
+            60000,
+            -1,
+            true,
+            -1,
+            -1,
+            false,
+            "garbage-unfortunately");
+    pollInterval = ReplicateFromLeader.determinePollInterval(updateHandlerInfo);
+    assertEquals("garbage-unfortunately", pollInterval);
```
cpoerschke, line 89: Perhaps `ReplicationHander.readIntervalNs` could be made public and used to validate either in UpdateHandlerInfo or in determinePollInterval. With `expectThrows(...)` in the test logic here then.

epugh:
thanks @cpoerschke for your review on this!   Definitly a better patch!

solr/core/src/test/org/apache/solr/cloud/ReplicateFromLeaderTest.java
tboeghk: Fail early, fail fast it is

solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
```diff
@@ -133,18 +126,25 @@ public static String getCommitVersion(SolrCore solrCore) {
...
     int softCommitMaxTime = uinfo.autoSoftCommmitMaxTime;
     boolean hardCommitNewSearcher = uinfo.openSearcher;
-    String pollIntervalStr = null;
-    if (hardCommitMaxTime != -1) {
+    String customCommitPollInterval = uinfo.commitPollInterval;
+    String pollIntervalStr = "00:00:03";
+
+    if (System.getProperty("jetty.testMode") != null) {
+      pollIntervalStr = "00:00:01";
+    } else if (customCommitPollInterval != null) {
```
cpoerschke, line 145:
```suggestion
    }

    if (customCommitPollInterval != null) {
```

would preserved existing behaviour, though perhaps the alternative here is actually "more correct" than the existing behaviour?

solr/core/src/test/org/apache/solr/cloud/ReplicateFromLeaderTest.java
```diff
@@ -18,54 +18,94 @@
...
+    System.clearProperty("jetty.testMode");
+  }
+
+  @After
+  public void tearDown() {
+    if (jettyTestMode != null) {
+      System.setProperty("jetty.testMode", jettyTestMode);
+    }
+  }
+
```
cpoerschke, line 45: Would retaining existing behaviour -- https://github.com/apache/solr/pull/2316/files#r1755039181 -- mean that we don't have to make this test change here?

solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
```diff
@@ -133,18 +126,25 @@ public static String getCommitVersion(SolrCore solrCore) {
...
 
   /**
-   * Determine the poll interval for replicas based on the auto soft/hard commit schedule
+   * Determine the poll interval for replicas based on the auto soft/hard commit schedule or
+   * configured commit poll interval
    *
    * @param uinfo the update handler info containing soft/hard commit configuration
    * @return a poll interval string representing a cadence of polling frequency in the form of
-   *     hh:mm:ss
+   *     hh:mm:ss, never <code>null</code>
```
cpoerschke, line 134: Nice, didn't know that `never` is a special value. Perhaps we could add test coverage for it too whilst in the neighbourhood? Thanks!

solr/core/src/test/org/apache/solr/cloud/ReplicateFromLeaderTest.java
tboeghk: Yes, this is due to `jetty.testMode` being set during unit tests by default :-/

solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
tboeghk: Really? Sure!

solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
tboeghk: I'd suggest leaving it this way. I'll add a test in which we test the test behaviour: regardless of the arguments given, the `commitPollInterval` ist always `00:00:01` if the system property `jetty.testMode` is present.

solr/core/src/java/org/apache/solr/cloud/ReplicateFromLeader.java
cpoerschke: Ah, I mis-read the comment and thought that "never" was a string value that could be returned when actually the comment was about a `null` value meaning never.

Yes, if null is never returned then no need to comment about it in the javadocs (and then also no need for the caller to consider what to do when null is returned).

cpoerschke:
> thanks @cpoerschke for your review on this! Definitly a better patch!

So just listing out likely remaining "TODO" items in case that's helpful?

- solr ref guide documentation update
- merging of latest `origin/main`
- `solr/CHANGES.txt` entry in the 9.8 section

cpoerschke:
Opened #3144 as a continuation of this pull request here since pushing to this PR's branch is not possible (due to org vs. individual repo fork).

PR #2320, opened 2024-02-28
Move EnvToSyspropMappings.properties to solrj module

janhoy:
I discovered that the properties file `EnvToSyspropMappings.properties` required by the new `EnvUtils` (SOLR-15960) was located in `core`. But since `EnvUtils` class is in `solrj` and will fail static `init()` if the property file is not found (perhaps it should not fail?) then when solrj is used on client side without solr-core, this will cause an exception:

```java
java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:209)
	at org.apache.solr.common.util.EnvUtils.<clinit>(EnvUtils.java:51)
```

This PR simply moves the property file to the solrj module.


HoustonPutman (approved):
Looks good to me if the tests pass

janhoy:
I discovered this by chance. Obviously all our solrJ tests run with access to solr-core on classpath? 
We should likely have a test with only SolrJ and its dependencies on classpath?

janhoy:
Created SOLR-17188 to track creation of such as new test

janhoy:
Related JIRA just opened: https://issues.apache.org/jira/browse/SOLR-17199

PR #2327, opened 2024-03-01
SOLR-17189 Fix DockMakerTest.testRealisticUnicode

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17189

    -Dsolr.bench.seed=1392507964231541

----
WIP.  Didn't fix the problem yet but just tried to make the benchmark tests actually repeatable

### DIFF
--- a/solr/benchmark/src/java/org/apache/solr/bench/BaseBenchState.java
+++ b/solr/benchmark/src/java/org/apache/solr/bench/BaseBenchState.java
@@ -41,9 +41,29 @@ import org.slf4j.LoggerFactory;
 @State(Scope.Benchmark)
 public class BaseBenchState {
 
-  private static final long RANDOM_SEED = System.nanoTime();
+  public static final long RANDOM_SEED;
 
-  private static final SplittableRandom random = new SplittableRandom(getInitRandomSeed());
+  static {
+    Long seed = Long.getLong("solr.bench.seed");
+
+    if (seed == null) {
+      String prop = System.getProperty("tests.seed"); // RandomizedTesting framework
+      if (prop != null) {
+        // if there is a test failure we remain reproducible based on the test seed:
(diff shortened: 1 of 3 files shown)


dsmiley:
I wrote a tiny script that loops over the code points here and there are many whitespace chars, including a space char (ASCII digit 32).  This and many others are in the first block.  @markrmiller did you intend the "realistic unicode" to include whitespace?  What makes these characters "realistic" anyway?

markrmiller:
Realistic is not referring to the characters.

The random Unicode character code likely came from Lucene. If there is a regex check that fails in the test, then it’s likely the generator wasn’t intended to generate whitespace characters. I’d bet random string generation is meant to generate a sequence of none whitespace characters.

dsmiley:
Okay.  For simplicity, let's just remap each whitespace to the first non-whitespace in the chosen block.  Or maybe even simpler -- the letter 'X' (hey why not?).  Or maybe you might recommend something else.

The coding style / framework here is unusual to me and I think most people.  If I had to name it, it'd be "extreme-streaming" or "latent-generation" or I dunno.  I won't even bother giving it to ChatGPT as it doesn't know this unique framework.  Do you have advice or a tip on how to approach this little programming problem?  Feel free to send a commit to this branch :-)

Separately, note this PR includes a fix for the non-repeatability of the randomness.  It's not perfect -- the RandomizedContext seed isn't being passed in unless I set it explicitly via the standard `tests.seed`.

dsmiley (commented):
I fixed the issue; the test passed.
I used a wrapping generator to ensure all strings that come through do not have a whitespace char.
If I don't hear back, I'll merge in a couple days.

dsmiley:
After reading some QuickTheories docs, it seems using an `assume(Predicate)` would be an alternative; less code too.  I'll switch it.

PR #2329, opened 2024-03-01
SOLR-17197: Fix getting fieldType by its name in FileBasedSpellChecker

AndreyBozhko:
https://issues.apache.org/jira/browse/SOLR-17197

# Description

Fix logic in FileBasedSpellChecker that uses wrong method to retrieve fieldType by its name

# Tests

There are a few existing tests in FileBasedSpellCheckerTest - but those tests refer to `schema.xml` which defines the field `teststop` and fieldType by the same name `teststop` (so the bug stays hidden).

I changed the tests to refer to `schema-spellchecker.xml` instead (to avoid changes to other unrelated tests that may depend on `schema.xml`), and made sure that the `teststop` field and its type do not have exactly the same name.

# Checklist

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -28,8 +28,6 @@ Optimizations
 Bug Fixes
 ---------------------
 
-* SOLR-17197: Fix getting fieldType by its name in FileBasedSpellChecker (Andrey Bozhko)
-
 Deprecation Removals
 ----------------------
 
@@ -107,7 +105,7 @@ Improvements
 
 * SOLR-17145: The INSTALLSHARDDATA API now includes a 'requestid' field when run asynchronously (Jason Gerlowski)
 
-* SOLR-17159: bin/solr post now has proper unit testing.  Users can specify a --dry-run option to 
+* SOLR-17159: bin/solr post now has proper unit testing.  Users can specify a --dry-run option to
   simulate posting documents without sending them to Solr. (Eric Pugh)
 
 * SOLR-17058: Add 'distrib.statsCache' parameter to disable distributed stats requests at query time. (Wei Wang, Mikhail Khludnev)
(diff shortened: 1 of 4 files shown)


solr/CHANGES.txt
```diff
@@ -27,7 +27,8 @@ Optimizations
 
 Bug Fixes
 ---------------------
-(No changes)
+
+* PR#2329: Fix getting fieldType by its name in FileBasedSpellChecker (Andrey Bozhko)
```
epugh, line 31: I *believe* we normally create a JIRA for these user facing, so we can track which version of Solr it gets released.   The pattern has recently become a bit muddier...  non user facing changes sometimes skipping JIRA..   If you add a JIRA, then that makes it easier for me to merge this!

solr/core/src/test-files/solr/collection1/conf/schema-spellchecker.xml
```diff
@@ -49,9 +49,22 @@
     </analyzer>
   </fieldType>
 
+  <fieldType name="teststop_type" class="solr.TextField">
```
epugh, line 52: I haven't checked in more detail, but don't we normally camelCase?   testStopType ???

epugh (changes requested):
Looks like a nice fix!   Do we have this same issue in lots of other places do you know?

solr/CHANGES.txt
epugh: i confirmed that yes, please open a jira...

AndreyBozhko: I submitted a request for a JIRA account, will open the issue once the account is set up

solr/core/src/test-files/solr/collection1/conf/schema-spellchecker.xml
AndreyBozhko: I saw that in different test files, the fields and field types appear in either camel case or snake case (an sometimes even in kebab case).

I personally use snake case, and I believe the Ref Guide examples also suggest snake case (e.g., field types `text_general` or `text_ws`).

AndreyBozhko (commented):
> Do we have this same issue in lots of other places do you know?

I searched for the usage of methods like `getFieldTypeByName`, `getFieldTypeNoEx`, etc. - and at a quick glance, nothing else stood out that was obviously wrong.

solr/core/src/test-files/solr/collection1/conf/schema-spellchecker.xml
epugh: thanks for taking the time to check the Ref Guide!

epugh: we probably should `snake_case` it everywhere......   in all of our test-files..  to have a single pattern.

PR #2331, opened 2024-03-01
Tests: JettyConfig.Builder.setContext("/solr")

dsmiley:
Isn't needed because it's the default

Note this is on branch_9x.  Main already has this; was done some time ago I guess.


epugh (approved):
Much nicer!

PR #2332, opened 2024-03-04
fail distrib.statsCache=true when LocalStatsCache

mkhludnev:
https://issues.apache.org/jira/browse/SOLR-17173

also invert properties


mkhludnev:
@wwang30 what do you think of it?

weiwang19:
@mkhludnev Looking good to me

weiwang19:
@mkhludnev looking good to me. thanks for putting all together

PR #2347, opened 2024-03-14
SOLR-17203 fix calculation for CpuAllowedLimit

gus-asf:
https://issues.apache.org/jira/browse/SOLR-17203

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/CpuAllowedLimit.java
+++ b/solr/core/src/java/org/apache/solr/search/CpuAllowedLimit.java
@@ -34,7 +34,9 @@ import org.slf4j.LoggerFactory;
 public class CpuAllowedLimit implements QueryTimeout {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
-  private final long limitAtNs;
+  private final long
+      timeOutAtThreadTotalCpuNs; // this is a value based on total cpu since the inception of the
+  // thread
   private final ThreadCpuTimer threadCpuTimer;
 
   /**
@@ -57,17 +59,17 @@ public class CpuAllowedLimit implements QueryTimeout {
       throw new IllegalArgumentException(
           "Check for limit with hasCpuLimit(req) before creating a CpuAllowedLimit");
     }
-    // calculate when the time limit is reached, account for the time already spent
-    limitAtNs =
-        threadCpuTimer.getStartCpuTimeNs()
+    // calculate when the time when the limit is reached, e.g. account for the time already spent
+    timeOutAtThreadTotalCpuNs =
+        threadCpuTimer.getStartCpuTimeNs() // how much the thread already used since it began
             + TimeUnit.NANOSECONDS.convert(reqCpuLimit, TimeUnit.MILLISECONDS);
   }
 
   @VisibleForTesting
   CpuAllowedLimit(long limitMs) {
     this.threadCpuTimer = new ThreadCpuTimer();
-    limitAtNs =
-        threadCpuTimer.getCurrentCpuTimeNs()
+    timeOutAtThreadTotalCpuNs =
+        threadCpuTimer.getThreadTotalCpuNs()
             + TimeUnit.NANOSECONDS.convert(limitMs, TimeUnit.MILLISECONDS);
   }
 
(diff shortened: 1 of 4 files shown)


solr/core/src/java/org/apache/solr/util/ThreadCpuTimer.java
```diff
@@ -87,23 +87,27 @@ public long getStartCpuTimeNs() {
...
-          ? THREAD_MX_BEAN.getCurrentThreadCpuTime() - this.startCpuTimeNanos
+          ? getThreadTotalCpuNs() - this.startCpuTimeNanos
           : UNSUPPORTED;
     } else {
       return UNSUPPORTED;
     }
   }
 
+  public long getThreadTotalCpuNs() {
+    return THREAD_MX_BEAN.getCurrentThreadCpuTime();
```
sigram, line 101: Since it's a public method we need to check for null here, as in other methods.

Also, to be consistent with other names it should be named something like `getTotalCpuTimeNs()` ... although at this point the names become confusing again. Maybe we should use `elapsed` and `absolute` instead?

solr/core/src/java/org/apache/solr/util/ThreadCpuTimer.java
```diff
@@ -87,23 +87,27 @@ public long getStartCpuTimeNs() {
    *
    * @return current value, or {@link #UNSUPPORTED} if not supported.
    */
-  public long getCurrentCpuTimeNs() {
+  public long getElapsedTimerCpuNs() {
```
sigram, line 90: We're getting the elapsed CPU time, not a timer ... so a better name would be `getElapsedCpuTimeNs()`

solr/core/src/java/org/apache/solr/util/ThreadCpuTimer.java
```diff
@@ -74,7 +74,7 @@ public static boolean isSupported() {
   /**
    * Return the initial value of CPU time for this thread when this instance was first created.
    * NOTE: absolute value returned by this method has no meaning by itself, it should only be used
-   * when comparing elapsed time between this value and {@link #getCurrentCpuTimeNs()}.
+   * when comparing elapsed time between this value and {@link #getElapsedTimerCpuNs()}.
```
sigram, line 77: This comment is not true - elapsed time is already relative to the start of the timer, whereas the startCpuTime is an absolute value. The javadoc should say that the comparison should be between two absolute values, ie. startCpuTime and threadTotalCpu.

solr/core/src/java/org/apache/solr/search/CpuAllowedLimit.java
```diff
@@ -34,7 +34,9 @@
 public class CpuAllowedLimit implements QueryTimeout {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
-  private final long limitAtNs;
+  private final long
+      timeOutAtThreadTotalCpuNs; // this is a value based on total cpu since the inception of the
+  // thread
```
sigram, line 39: Weird wrapping?

solr/core/src/java/org/apache/solr/search/CpuAllowedLimit.java
```diff
@@ -57,17 +59,17 @@ public CpuAllowedLimit(SolrQueryRequest req) {
       throw new IllegalArgumentException(
           "Check for limit with hasCpuLimit(req) before creating a CpuAllowedLimit");
     }
-    // calculate when the time limit is reached, account for the time already spent
-    limitAtNs =
-        threadCpuTimer.getStartCpuTimeNs()
+    // calculate when the time when the limit is reached, e.g. account for the time already spent
```
sigram, line 62: "calculate the time"

solr/core/src/java/org/apache/solr/util/ThreadCpuTimer.java
gus-asf: Ah I spent a bunch of time trying to come up with names and didn't think to use the word absolute. That might work, Ill fiddle with that.

solr/core/src/java/org/apache/solr/util/ThreadCpuTimer.java
gus-asf: ah refactor tools...

PR #2348, opened 2024-03-15
SOLR-17018: add QueryLimits support to Learning To Rank rescoring

alessandrobenedetti:
https://issues.apache.org/jira/browse/SOLR-17018

# Description

Learning To Rank was ignoring any queryLimit parameter (such as timeAllowed) during the reranking phase.
This means that a particularly expensive feature to extract may cause extremely slow responses and failures out of control.

# Solution

The idea is to check the query limits after the rescore of each search result (Learning To Rank rescore topK results).
If the query limit is exhausted before the completion of the rescoring, the reranking is aborted and reverted (all or nothing approach).
The response is marked as a partialResult and the original score and ranking are returned.

# Tests

A test has been added with a simple Learning To Rank model that uses just one single slow feature (implemented with a sleep function query).
This feature simulates a slow feature extraction.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/IncompleteRerankingException.java
+++ b/solr/core/src/java/org/apache/solr/search/IncompleteRerankingException.java
@@ -18,7 +18,7 @@ package org.apache.solr.search;
 
 public class IncompleteRerankingException extends RuntimeException {
 
-  public IncompleteRerankingException(String msg) {
-    super(msg);
+  public IncompleteRerankingException() {
+    super();
   }
 }
--- a/solr/core/src/java/org/apache/solr/search/IncompleteRerankingException.java
+++ b/solr/core/src/java/org/apache/solr/search/IncompleteRerankingException.java
@@ -1,0 +1,24 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.search;
+
+public class IncompleteRerankingException extends RuntimeException {
+
+  public IncompleteRerankingException(String msg) {
+    super(msg);
+  }
+}
--- a/solr/core/src/java/org/apache/solr/search/ReRankCollector.java
+++ b/solr/core/src/java/org/apache/solr/search/ReRankCollector.java
@@ -128,8 +128,8 @@ public class ReRankCollector extends TopDocsCollector<ScoreDoc> {
       }
 
       ScoreDoc[] mainScoreDocs = mainDocs.scoreDocs;
-      ScoreDoc[] mainScoreDocsClone =
-          deepClone(mainScoreDocs, reRankScaler != null && reRankScaler.scaleScores());
+      boolean zeroOutScores = reRankScaler != null && reRankScaler.scaleScores();
+      ScoreDoc[] mainScoreDocsClone = deepClone(mainScoreDocs, zeroOutScores);
       ScoreDoc[] reRankScoreDocs = new ScoreDoc[Math.min(mainScoreDocs.length, reRankDocs)];
       System.arraycopy(mainScoreDocs, 0, reRankScoreDocs, 0, reRankScoreDocs.length);
 
--- a/solr/core/src/java/org/apache/solr/search/ReRankCollector.java
+++ b/solr/core/src/java/org/apache/solr/search/ReRankCollector.java
@@ -144,7 +144,7 @@ public class ReRankCollector extends TopDocsCollector<ScoreDoc> {
                     .getReplaceRescorer()
                     .rescore(searcher, mainDocs, mainDocs.scoreDocs.length)
                 : reRankQueryRescorer.rescore(searcher, mainDocs, mainDocs.scoreDocs.length);
-      } catch (QueryLimitsExceededException ex) {
+      } catch (IncompleteRerankingException ex) {
         mainDocs.scoreDocs = mainScoreDocsClone;
         rescoredDocs = mainDocs;
       }
--- a/solr/core/src/java/org/apache/solr/search/ReRankCollector.java
+++ b/solr/core/src/java/org/apache/solr/search/ReRankCollector.java
@@ -129,21 +129,25 @@ public class ReRankCollector extends TopDocsCollector<ScoreDoc> {
 
       ScoreDoc[] mainScoreDocs = mainDocs.scoreDocs;
       ScoreDoc[] mainScoreDocsClone =
-          (reRankScaler != null && reRankScaler.scaleScores())
-              ? deepCloneAndZeroOut(mainScoreDocs)
-              : null;
+          deepClone(mainScoreDocs, reRankScaler != null && reRankScaler.scaleScores());
       ScoreDoc[] reRankScoreDocs = new ScoreDoc[Math.min(mainScoreDocs.length, reRankDocs)];
       System.arraycopy(mainScoreDocs, 0, reRankScoreDocs, 0, reRankScoreDocs.length);
 
       mainDocs.scoreDocs = reRankScoreDocs;
 
       // If we're scaling scores use the replace rescorer because we just want the re-rank score.
-      TopDocs rescoredDocs =
-          reRankScaler != null && reRankScaler.scaleScores()
-              ? reRankScaler
-                  .getReplaceRescorer()
-                  .rescore(searcher, mainDocs, mainDocs.scoreDocs.length)
-              : reRankQueryRescorer.rescore(searcher, mainDocs, mainDocs.scoreDocs.length);
+      TopDocs rescoredDocs;
+      try {
+        rescoredDocs =
+            reRankScaler != null && reRankScaler.scaleScores()
+                ? reRankScaler
+                    .getReplaceRescorer()
+                    .rescore(searcher, mainDocs, mainDocs.scoreDocs.length)
+                : reRankQueryRescorer.rescore(searcher, mainDocs, mainDocs.scoreDocs.length);
+      } catch (QueryLimitsExceededException ex) {
+        mainDocs.scoreDocs = mainScoreDocsClone;
+        rescoredDocs = mainDocs;
+      }
 
       // Lower howMany to return if we've collected fewer documents.
       howMany = Math.min(howMany, mainScoreDocs.length);
@@ -208,13 +212,15 @@ public class ReRankCollector extends TopDocsCollector<ScoreDoc> {
     }
   }
 
-  private ScoreDoc[] deepCloneAndZeroOut(ScoreDoc[] scoreDocs) {
+  private ScoreDoc[] deepClone(ScoreDoc[] scoreDocs, boolean zeroOut) {
     ScoreDoc[] scoreDocs1 = new ScoreDoc[scoreDocs.length];
     for (int i = 0; i < scoreDocs.length; i++) {
       ScoreDoc scoreDoc = scoreDocs[i];
       if (scoreDoc != null) {
         scoreDocs1[i] = new ScoreDoc(scoreDoc.doc, scoreDoc.score);
-        scoreDoc.score = 0f;
+        if (zeroOut) {
+          scoreDoc.score = 0f;
+        }
       }
     }
     return scoreDocs1;
--- a/solr/core/src/java/org/apache/solr/search/ReRankCollector.java
+++ b/solr/core/src/java/org/apache/solr/search/ReRankCollector.java
@@ -139,7 +139,7 @@ public class ReRankCollector extends TopDocsCollector<ScoreDoc> {
       TopDocs rescoredDocs;
       try {
         rescoredDocs =
-            reRankScaler != null && reRankScaler.scaleScores()
+            zeroOutScores // previously zero-ed out scores are to be replaced
                 ? reRankScaler
                     .getReplaceRescorer()
                     .rescore(searcher, mainDocs, mainDocs.scoreDocs.length)
--- a/solr/modules/ltr/src/java/org/apache/solr/ltr/LTRRescorer.java
+++ b/solr/modules/ltr/src/java/org/apache/solr/ltr/LTRRescorer.java
@@ -31,8 +31,8 @@ import org.apache.lucene.search.TopDocs;
 import org.apache.lucene.search.TotalHits;
 import org.apache.lucene.search.Weight;
 import org.apache.solr.ltr.interleaving.OriginalRankingLTRScoringQuery;
+import org.apache.solr.search.IncompleteRerankingException;
 import org.apache.solr.search.QueryLimits;
-import org.apache.solr.search.QueryLimitsExceededException;
 import org.apache.solr.search.SolrIndexSearcher;
 
 /**
@@ -239,9 +239,9 @@ public class LTRRescorer extends Rescorer {
     if (QueryLimits.getCurrentLimits()
         .maybeExitWithPartialResults(
             "Learning To Rank rescoring -"
-                + " The full reranking didn't complete and got reverted."
-                + " All documents preserved their original score and ranking.")) {
-      throw new QueryLimitsExceededException("A query limit has been exceeded when rescoring");
+                + " The full reranking didn't complete."
+                + " If partial results are tolerated the reranking got reverted and all documents preserved their original score and ranking.")) {
+      throw new IncompleteRerankingException("A query limit has been exceeded when rescoring");
     }
     if (hitUpto < topN) {
       reranked[hitUpto] = hit;
--- a/solr/modules/ltr/src/java/org/apache/solr/ltr/LTRRescorer.java
+++ b/solr/modules/ltr/src/java/org/apache/solr/ltr/LTRRescorer.java
@@ -235,13 +235,13 @@ public class LTRRescorer extends Rescorer {
     boolean logHit = false;
 
     scorer.getDocInfo().setOriginalDocScore(hit.score);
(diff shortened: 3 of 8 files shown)


alessandrobenedetti:
I would like to thank @gus-asf and @dsmiley for the great work in query limits.
I scratched my head for a while on this matter (we had discussions with @cpoerschke on another thread), but after the https://github.com/apache/solr/pull/2237 contribution, I kinda feel this PR is elegant and useful enough.

let me know for your feedback and then I'll add the CHANGES.TXT and backport it.

cpoerschke:
> ... (we had discussions with @cpoerschke on another thread) ...

#8 was maybe the other thread?

solr/core/src/java/org/apache/solr/search/ReRankCollector.java
```diff
@@ -129,21 +129,25 @@ public TopDocs topDocs(int start, int howMany) {
 
       ScoreDoc[] mainScoreDocs = mainDocs.scoreDocs;
       ScoreDoc[] mainScoreDocsClone =
-          (reRankScaler != null && reRankScaler.scaleScores())
-              ? deepCloneAndZeroOut(mainScoreDocs)
-              : null;
+          deepClone(mainScoreDocs, reRankScaler != null && reRankScaler.scaleScores());
```
dsmiley, line 132: boolean args can be unclear at the call site.  To mitigate that, please declare a local var zeroOutScores initialized to this and pass in.

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRRescorer.java
```diff
@@ -233,7 +235,14 @@ protected static boolean scoreSingleHit(
     boolean logHit = false;
 
     scorer.getDocInfo().setOriginalDocScore(hit.score);
+    QueryLimits queryLimits = QueryLimits.getCurrentLimits();
```
dsmiley, line 238: IMO it's better do do such checks at the spot where there is a loop over something like docs.  This also allows fetching the QueryLimits outside the loop.  Thus I suggest modifying `scoreFeatures`.  You could do the `queryLimits.maybeExitWithPartialResults` check at the start of the `while` loop.

dsmiley: Ah; I see you do this here so that LTRInterleavingRescorer will incorporate this, which overrides `scoreFeatures` completely yet calls `scoreSingleHit`.  Okay fine.   But note `scoreFeatures` has lots of code duplication; it'd benefit from a refactor.

If you want to keep queryLimits check in `scoreSingleHit` then you needn't declare queryLimits to only then call a method on it 2 lines later (kinda weird; why score in-between?).  Just do `QueryLimits.getCurrentLimits().maybeExitWithPartialResults(...)`

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRRescorer.java
```diff
@@ -233,7 +235,14 @@ protected static boolean scoreSingleHit(
...
     hit.score = scorer.score();
+    if (queryLimits.maybeExitWithPartialResults(
```
sigram, line 240: This method doesn't throw exception and returns a boolean only when partial results are acceptable - meaning that since we already computed some expensive results we can still decide here to return them if we want to. We can also return with empty results if nothing was computed yet.
However, the method `maybeExitWithPartialResults` will throw an exception if the user decided he doesn't want partial results (by setting request param `partialResults=false`) - so we don't need to explicitly throw any exception here.

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRRescorer.java
```diff
@@ -233,7 +235,14 @@ protected static boolean scoreSingleHit(
...
+        "Learning To Rank rescoring -"
+            + " The full reranking didn't complete and got reverted."
+            + " All documents preserved their original score and ranking.")) {
+      throw new QueryLimitsExceededException("The Time Allowed has been exceeded when rescoring");
```
sigram, line 244: It's not necessarily `timeAllowed` now ... any other limit may have been exceeded. A better wording would be "A query limit has been exceeded".

solr/core/src/java/org/apache/solr/search/ReRankCollector.java
alessandrobenedetti: I agree, change is coming in the next commit

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRRescorer.java
alessandrobenedetti: Yes David you are correct about the reason and I agree scoreFeatures would deserve some care, but I prefer to dedicate a separate PR to that.

Your observation about the declare and then call is also welcome, the change is coming in my next commit.
Not sure why I did it in that way, maybe a copy and paste from a previous version of the code outside the loop,

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRRescorer.java
alessandrobenedetti: Thanks for the correction, it's coming in the next commit.
In regards to the exception discussion, I'll talk about it in the other comment

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRRescorer.java
alessandrobenedetti: Hi 
@sigram, I see your point now, I added a test and a couple of refinements with the latest commit, I guess it's ok now?

solr/modules/ltr/src/test/org/apache/solr/ltr/TestLTRQParserPlugin.java
```diff
@@ -137,4 +140,63 @@ public void ltrNoResultsTest() throws Exception {
...
+     * */
+
+    final String solrQuery = "_query_:{!edismax qf='id' v='8^=10 9^=5 7^=3 6^=1'}";
+    final SolrQuery query = new SolrQuery();
+    query.setQuery(solrQuery);
+    query.add("fl", "*, score");
+    query.add("rows", "4");
+    query.add("fv", "true");
+    query.add("rq", "{!ltr model=slowModel reRankDocs=3}");
+    query.add("timeAllowed", "300");
```
cpoerschke, line 158: minor/style: could use `setTimeAllowed` and `setRows` etc. here it seems e.g.
```suggestion
    query. setTimeAllowed(300);
```

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRRescorer.java
```diff
@@ -234,6 +236,13 @@ protected static boolean scoreSingleHit(
 
     scorer.getDocInfo().setOriginalDocScore(hit.score);
     hit.score = scorer.score();
+    if (QueryLimits.getCurrentLimits()
+        .maybeExitWithPartialResults(
+            "Learning To Rank rescoring -"
+                + " The full reranking didn't complete."
+                + " If partial results are tolerated the reranking got reverted and all documents preserved their original score and ranking.")) {
```
cpoerschke, line 243: subjective, code complexity/simplicity vs. user/operational friendliness: the _"If partial results are tolerated"_ probably is 'knowable' at this point in the code i.e. if the info/knowledge were to passed down to this API then the wording could be differentiated here i.e. no 'if' qualifier in the wording.

solr/core/src/java/org/apache/solr/search/ReRankCollector.java
```diff
@@ -128,22 +128,26 @@ public TopDocs topDocs(int start, int howMany) {
...
-      TopDocs rescoredDocs =
-          reRankScaler != null && reRankScaler.scaleScores()
-              ? reRankScaler
-                  .getReplaceRescorer()
-                  .rescore(searcher, mainDocs, mainDocs.scoreDocs.length)
-              : reRankQueryRescorer.rescore(searcher, mainDocs, mainDocs.scoreDocs.length);
+      TopDocs rescoredDocs;
+      try {
+        rescoredDocs =
+            reRankScaler != null && reRankScaler.scaleScores()
```
cpoerschke, line 142:
```suggestion
            zeroOutScores // previously zero-ed out scores are to be replaced
```

solr/core/src/java/org/apache/solr/search/IncompleteRerankingException.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.search;
+
+public class IncompleteRerankingException extends RuntimeException {
+
+  public IncompleteRerankingException(String msg) {
+    super(msg);
+  }
```
cpoerschke, line 23: minor: omit `msg` since it's never used currently
```suggestion
  public IncompleteRerankingException() {
    super("");
  }
```

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRRescorer.java
cpoerschke: > ... scoreFeatures would deserve some care, but I prefer to dedicate a separate PR to that. ...

+1 to not mixing code changes with code care.

And I vaguely recall something around this previously, possibly the #478 (now stale) draft.

cpoerschke (commented):
Thanks @alessandrobenedetti for working on this!

Wondering if https://github.com/apache/solr/blob/main/solr/solr-ref-guide/modules/query-guide/pages/common-query-parameters.adoc#timeallowed-parameter or https://github.com/apache/solr/blob/main/solr/solr-ref-guide/modules/query-guide/pages/common-query-parameters.adoc#partialresults-parameter or https://github.com/apache/solr/blob/main/solr/solr-ref-guide/modules/query-guide/pages/learning-to-rank.adoc or some combination might also mention about this?

solr/modules/ltr/src/test/org/apache/solr/ltr/TestLTRQParserPlugin.java
alessandrobenedetti: Coming in the next commit!

solr/core/src/java/org/apache/solr/search/ReRankCollector.java
alessandrobenedetti: coming in the next commit

solr/core/src/java/org/apache/solr/search/IncompleteRerankingException.java
alessandrobenedetti: coming in the next commit!

alessandrobenedetti:
Thanks @cpoerschke for your feedback!
I'll push a commit soon with changes accordingly to your suggestions.
In terms of documentation I am inclined to add a line in the Learning To Rank section only.
Not because I don't want to add it in the other two sections you mentioned (Partial Results and TImeAllowed), but because I believe those two sections should be enriched with all the components they support, explaining what happens for each of the components if the limit is reached.
This will include also what happens in Learning To Rank but I see this general improvement as a separate documentation Pull Request.
I'll keep iterating on this PR for a bit, so additional comments are welcome. When closing I'll also create the Jiras with all the children tasks identified (refactor of ScoreFeatures, Double label MaybeShouldExit and better documentation for timeAllowed and partial results)

solr/modules/ltr/src/test/org/apache/solr/ltr/TestLTRQParserPlugin.java
```diff
@@ -137,4 +140,63 @@ public void ltrNoResultsTest() throws Exception {
...
+
+    assertJQ(
```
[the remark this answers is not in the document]
alessandrobenedetti: Done!

solr/modules/ltr/src/test/org/apache/solr/ltr/TestLTRQParserPlugin.java
```diff
@@ -137,4 +140,92 @@ public void ltrNoResultsTest() throws Exception {
...
+  public void ltr_expensiveFeatureRescoring_shouldTimeOutAndReturnPartialResults()
+      throws Exception {
+    /* One SolrFeature is defined: {!func}sleep(1000,999)
+     * It simulates a slow feature extraction, sleeping for 1000ms and returning 999 as a score when finished
+     * */
+
+    final String solrQuery = "_query_:{!edismax qf='id' v='8^=10 9^=5 7^=3 6^=1'}";
+    final SolrQuery query = new SolrQuery();
+    query.setQuery(solrQuery);
+    query.setFields("*", "score");
```
cpoerschke, line 154:
```suggestion
    query.setFields("id", "score");
```

solr/modules/ltr/src/test/org/apache/solr/ltr/TestLTRQParserPlugin.java
```diff
@@ -137,4 +140,92 @@ public void ltrNoResultsTest() throws Exception {
...
+  @Test
+  public void ltr_expensiveFeatureRescoringAndPartialResultsNotTolerated_shouldRaiseException()
+      throws Exception {
+    /* One SolrFeature is defined: {!func}sleep(1000,999)
+     * It simulates a slow feature extraction, sleeping for 1000ms and returning 999 as a score when finished
+     * */
+    final String solrQuery = "_query_:{!edismax qf='id' v='8^=10 9^=5 7^=3 6^=1'}";
+    final SolrQuery query = new SolrQuery();
+    query.setQuery(solrQuery);
+    query.setFields("*", "score");
```
cpoerschke, line 188:
```suggestion
    query.setFields("id", "score");
```

solr/modules/ltr/src/test/org/apache/solr/ltr/TestLTRQParserPlugin.java
```diff
@@ -137,4 +140,92 @@ public void ltrNoResultsTest() throws Exception {
...
+  public void ltr_expensiveFeatureRescoringWithinTimeAllowed_shouldReturnRerankedResults()
+      throws Exception {
+    /* One SolrFeature is defined: {!func}sleep(1000,999)
+     * It simulates a slow feature extraction, sleeping for 1000ms and returning 999 as a score when finished
+     * */
+
+    final String solrQuery = "_query_:{!edismax qf='id' v='8^=10 9^=5 7^=3 6^=1'}";
+    final SolrQuery query = new SolrQuery();
+    query.setQuery(solrQuery);
+    query.setFields("*", "score");
```
cpoerschke, line 213:
```suggestion
    query.setFields("id", "score");
```

solr/modules/ltr/src/test/org/apache/solr/ltr/TestLTRQParserPlugin.java
```diff
@@ -137,4 +140,92 @@ public void ltrNoResultsTest() throws Exception {
...
+    query.setRows(4);
+    query.setTimeAllowed(5000);
+    query.add("fv", "true");
```
cpoerschke, line 216: Maybe we could also test that `fv` is correctly returned here (and not in the time-ran-out case)?

solr/modules/ltr/src/test/org/apache/solr/ltr/TestLTRQParserPlugin.java
```diff
@@ -137,4 +140,92 @@ public void ltrNoResultsTest() throws Exception {
...
+    assertJQ(
+        "/query" + query.toQueryString(),
+        "/response/numFound/==4",
+        "/response/docs/[0]/id=='7'",
+        "/response/docs/[0]/score==999.0",
+        "/response/docs/[1]/id=='8'",
+        "/response/docs/[1]/score==999.0",
+        "/response/docs/[2]/id=='9'",
+        "/response/docs/[2]/score==999.0",
+        "/response/docs/[3]/id=='6'",
```
cpoerschke, line 228:
```suggestion
        "/response/docs/[3]/id=='6'", // original score for the 4th document due to reRankDocs=3 limit
```

cpoerschke (approved):

solr/modules/ltr/src/test/org/apache/solr/ltr/TestLTRQParserPlugin.java
```diff
@@ -137,4 +140,92 @@ public void ltrNoResultsTest() throws Exception {
...
+    assertJQ(
+        "/query" + query.toQueryString(),
+        "/response/numFound/==4",
+        "/response/docs/[0]/id=='7'",
+        "/response/docs/[0]/score==999.0",
+        "/response/docs/[1]/id=='8'",
+        "/response/docs/[1]/score==999.0",
+        "/response/docs/[2]/id=='9'",
+        "/response/docs/[2]/score==999.0",
+        "/response/docs/[3]/id=='6'", // original score for the 4th document due to reRankDocs=3 limit
```
cpoerschke, line 228: to pass the `./gradlew :solr:modules:ltr:spotlessJavaCheck` part of `precommit`
```suggestion
        "/response/docs/[3]/id=='6'",
        // original score for the 4th document due to reRankDocs=3 limit
```

solr/modules/ltr/src/test/org/apache/solr/ltr/TestLTRQParserPlugin.java
alessandrobenedetti: I spent some time today on this, at the moment the reranking feature extraction is disconnected from the transformer feature extraction (org.apache.solr.ltr.response.transform.LTRFeatureLoggerTransformerFactory).
For the sake of this contribution that specifically focuses on the rescoring, I won't touch that part. But I will add as an additional task, in the output list of Jiras produced by this work, among the others.
Then we'll be able to tackle each of them independently.
We also have some work in the backlog partially done to re-use the feature extraction between the rescorer and transformer but I had to pause the work because of a lack of funding.

alessandrobenedetti:
All tests are green, I am finalising some checks and I'll merge them tomorrow!
Thanks to everyone involved with this contribution!

alessandrobenedetti:
Thanks to everyone involved, the contribution is now merged and backported to 9.x
I'll also backported the bugfix to 9.1.x as a client requested it. But the contribution there will be much different as there was no support for Query Limits back then.

As additional follow ups I created this Jira issues:

* Sleep function query to sleep on execution rather than parsing (https://issues.apache.org/jira/browse/SOLR-17225))
* Query limits should be checked also for feature extraction (https://issues.apache.org/jira/browse/SOLR-17226)
* Refactor of ScoreFeatures (https://issues.apache.org/jira/browse/SOLR-17227)
* Double label MaybeShouldExit (https://issues.apache.org/jira/browse/SOLR-17229)
* Better documentation for timeAllowed and partial results (https://issues.apache.org/jira/browse/SOLR-17229)

Anyone is welcome to contribute to them, I may do some of them myself (or my team) later on

PR #2349, opened 2024-03-15
SOLR-16866: Reorder nested directory deletions to avoid NoSuchFileException

magibney:
https://issues.apache.org/jira/browse/SOLR-16866

alternative to #2336

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
+++ b/solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
@@ -23,6 +23,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
```diff
@@ -70,13 +71,9 @@ public CacheValue(String path, Directory directory) {
     public boolean closeCacheValueCalled = false;
     public boolean doneWithDir = false;
     private boolean deleteAfterCoreClose = false;
-    public Set<CacheValue> removeEntries = new HashSet<>();
```
dsmiley, line 73: It's nice to see one less data structure to track

solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
```diff
@@ -302,6 +294,53 @@ private boolean closeCacheValue(CacheValue cacheValue) {
...
 
+  private static Iterable<CacheValue> sorted(Set<CacheValue> vals) {
+    if (vals.size() < 2) {
+      return vals;
+    }
+    // here we reverse-sort entries by path, in order to trivially ensure that
+    // subpaths are removed before parent paths.
+    CacheValue[] ret = vals.toArray(new CacheValue[0]);
+    Arrays.sort(ret, (a, b) -> b.path.compareTo(a.path));
+    return Arrays.asList(ret);
```
dsmiley, line 305: minor: this would be nicer as a single statement using Java streams

solr/core/src/test/org/apache/solr/core/CachingDirectoryFactoryTest.java
```diff
@@ -46,6 +57,90 @@ private static class Tracker {
     Directory dir;
   }
 
+  @Test
+  public void reorderingTest() throws Exception {
+    // failure: 58F920FAAE5904EF
```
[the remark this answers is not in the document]
magibney: ah, good catch. I think that crept in during development, and I neglected to remove it once the underlying issue was fixed. Definitely not failing for that seed anymore; removed that comment with dbc84cefa2b632487385258885bb269531c78506

magibney:
Assuming we want to wait until after 9.6 for this? It should be safe, but it's really only about reducing log noise, so is pretty low priority.

HoustonPutman:
I've seen this cause failures when moving replicas around, so I would love to see it fixed in 9.6

PR #2351, opened 2024-03-18
SOLR-16503: Introduce new default Http2SolrClient in UpdateShardHandler

iamsanjay:

Currently default HTTP (Apache Http 1) client in UpdateShardHandler is being used at many places in the codebase and consequently removing it at once and replacing with Jetty HTTP2 would requires lots of efforts and testing. Instead, the possibility of introducing another default Jetty HTTP2 client in UpdateShardHandler is proposed to facilitate incremental changes. 

# Checklist

### DIFF
--- a/solr/solr-ref-guide/modules/deployment-guide/pages/solr-in-docker.adoc
+++ b/solr/solr-ref-guide/modules/deployment-guide/pages/solr-in-docker.adoc
@@ -53,7 +53,7 @@ Next load some of the example data that is included in the container:
 
 [source,bash]
 ----
-docker exec -it my_solr post -c gettingstarted example/exampledocs/manufacturers.xml
+docker exec -it my_solr solr post -c gettingstarted example/exampledocs/manufacturers.xml
 ----
 
 In the UI, find the "Core selector" popup menu and select the "gettingstarted" core, then select the "Query" menu item.
@@ -245,7 +245,7 @@ docker run --name my_solr -p 8983:8983 solr solr-precreate books
 # get data
 mkdir mydata
 wget -O mydata/books.csv https://raw.githubusercontent.com/apache/solr/main/solr/example/exampledocs/books.csv
-docker run --rm -v "$PWD/mydata:/mydata" --network=host solr post -c books /mydata/books.csv
+docker run --rm -v "$PWD/mydata:/mydata" --network=host solr solr post -c books /mydata/books.csv
 ----
 
 The same works if you use the xref:docker-compose[example docker compose cluster], or you can just start your loading container in the same network:
@@ -253,7 +253,7 @@ The same works if you use the xref:docker-compose[example docker compose cluster
 [source,bash]
 ----
 docker run -e SOLR_HOST=solr1 --network=mycluster_solr solr solr create -c books -p 8983
-docker run --rm -v "$PWD/mydata:/mydata" --network=mycluster_solr solr post -c books /mydata/books.csv --host solr1
+docker run --rm -v "$PWD/mydata:/mydata" --network=mycluster_solr solr solr post -c books /mydata/books.csv --host solr1
 ----
 
 Alternatively, you can make the data available on a volume at Solr start time, and then load it from `docker exec` or a custom start script.


solr/core/src/java/org/apache/solr/update/UpdateShardHandler.java
```diff
@@ -236,11 +246,17 @@ public SolrMetricsContext getSolrMetricsContext() {
...
   }
 
-  // if you are looking for a client to use, it's probably this one.
+  // please use getDefaultHttpSolrClient instead.
+  @Deprecated(since = "9.6")
   public HttpClient getDefaultHttpClient() {
     return defaultClient;
   }
 
+  // if you are looking for a client to use, it's probably this one.
```
dsmiley, line 255: It's really weird that UpdateShardHandler holds the "default" httpClient (i.e. clients that have nothing to do with updates).  I don't want to scope creep much but I propose simply that CoreContainer have a getHttpSolrClient method that simply gets it from here, and nobody should call it here.  It's still initialized here for now.  I'm aware of SolrClientCache in CoreContainer but I could see transitioning away from that as well, especially with adding convenience base path / url request method to Http2SolrClient but even then it's not essential.

dsmiley (commented):
Lets get this PR up to date with master.

solr/core/src/java/org/apache/solr/update/UpdateShardHandler.java
iamsanjay: I think it's the name, `UpdateShardHandler`, which is not appropriate anymore. It started off with right intentions as there was only one client in the beginning, where now it encloses three HTTP clients. 

I can create one method in `CoreContainer#getDefaultHttpSolrClient` and delegate to SolrClientCache for the Default client?

iamsanjay: `solrClientCache` object in `CoreContainer.java` still relies on the old HTTP client. Before we delegate to solrClientCache for HttpClient, we have to change the HttpClient for solrClientCache.

dsmiley: Good point on UpdateShardHandler's name.  I think it's a terrible name!  In the project we colloquially refer to "handlers" as an abbreviation of RequestHandler which receives requests.  But this "ShardHandler" doesn't even receive.... what, shards?  To even be worthy of a Handler suffix in any shape or form, it must *receive* what it supposedly handles so it can then act on it.  But this class just holds a collection of things mostly (but not totally) related to updates. 

SolrClientCache is for when we need a separate SolrClient instance per base URL.  Or similarly a CloudSolrClient for each cluster.  Keep in mind it was written at a time when SolrClient classes were not immutable, so there was risk of sharing one.  It's nifty but I don't think it should be the go-to place for an internal (within the cluster) clients -- basically any user of UpdateShardHnadler.getDefaultHttpClient today.  Javadocs are needed!

iamsanjay: 1. Okay so no delegating to SolrClientCache.
2. Creating `CoreContainer#getDefaultHttpSolrClient` and delegating it to `UpdateShardHandler` can be one option(I will add code for this!)
3. If we are thinking of name change that would requires changes at around 40 places in the codebase as per quick search.
Some new names suggestions as per AI:
HttpOperationManager
HttpExecutorCollection
MultiHttpClientHandler
HttpTaskCoordinator
HttpOperationCoordinator
HttpClientManager
HttpTaskManager
4. We can also create Default HTTP client inside the CoreContainer. On the other hand, CoreContainer already have too much, Not sure If want to give them more.
5. Or may be it's time to divide UpdateShardHandler to few more classes.

dsmiley: Maybe we should subclass SolrClientCache, named "ServerSolrClientCache" (differentiating from the base one that's _mostly_ for streaming-expressions, from which it came) and provide these methods there.  That way we don't add ever more responsibilities to CoreContainer.  And it keeps these together with other SolrClients, so developers may be looking over in SolrClientCache.  Might as well consolidate.

Perhaps defer UpdateShardHandler going away to another work item that is "main" only.  This keeps compatibility with a number of plugins.  Don't want to distract us from the overall goal of getting Apache HttpClient out of Solr.

iamsanjay: How about we defer "the refactoring" for now? Because regardless we have to create a public API in UpdateShardHandler for default Http2SolrClient. So let's keep the scope of ticket there. Later on, when we are going to change the client in SolrClientCache, we can discuss then?

Or, Do you not recommend creating "default" HTTP2 in UpdateShardHandler?

dsmiley: As we are introducing an Http2SolrClient to use instead of an (Apache) HttpClient, this isn't a refactoring.  So now is an appropriate time to think about where best to place this new client so that we needn't go updating all the callers an additional time.  We agree UpdateShardHandler doesn't make sense because it's not for updates.

epugh:
@iamsanjay anythingbi can do to help get this in?  Some great work here!!!

PR #2355, opened 2024-03-19
SOLR-16982 Reference guide change for sharded request and circuit breaker

janhoy:
https://issues.apache.org/jira/browse/SOLR-16982

In the original PR we changed code but did not update reference guide.


epugh (approved):
LGTM

PR #2356, opened 2024-03-19
SOLR-12089: FileBasedSpellChecker docs have some missing params

AndreyBozhko:
https://issues.apache.org/jira/browse/SOLR-12089

# Description

Updated the Ref Guide about Spell Checking.

Also, I noticed some inconsistencies with how different spell checkers accept accuracy parameter:

- `FileBasedSpellChecker` and `IndexBasedSpellChecker` expect a string which is then parsed as float,
- `DirectSolrSpellChecker` can handle input as both string and float.

As for the existing examples in the Ref Guide - even though they were technically correct - it is still confusing that some examples suggest to provide accuracy as string, while other suggest to use a float.

# Checklist


solr/solr-ref-guide/modules/query-guide/pages/spell-checking.adoc
```diff
@@ -25,26 +25,34 @@ The basis for these suggestions can be terms in a field in Solr, externally crea
 === Define Spell Check in solrconfig.xml
 
 The first step is to specify the source of terms in `solrconfig.xml`.
-There are three approaches to spell checking in Solr, discussed below.
+There is a number of approaches to spell checking in Solr, discussed below.
```
epugh, line 28: Maybe "There are a number of"?

solr/solr-ref-guide/modules/query-guide/pages/spell-checking.adoc
```diff
@@ -25,26 +25,34 @@ The basis for these suggestions can be terms in a field in Solr, externally crea
...
+    <!-- optional parameters for all spellcheckers -->
+    <str name="spellcheckIndexDir">./spellcheckerDir</str>
+    <str name="name">default</str>
+    <str name="fieldType">content_ft</str>
+    <str name="queryAnalyzerFieldType">text_general</str>
     <str name="distanceMeasure">org.apache.lucene.search.spell.LevenshteinDistance</str>
-    <str name="accuracy">0.5</str>
-    -->
+    <str name="comparatorClass">score</str>
+    <float name="accuracy">0.5</float>
```
epugh, line 52: This in interesting...    I don't know that I have every used the `<float` attribute in my `solrconfig.xml` file...    I guess I assumed that everything was a `<str` and it was just how we passed in parameters.     This may be a larger question, but does having a type in the `solrconfig.xml` actually help us?     I wonder if everything should have just been `<param name="accuracy">0.5</param>` and let the Java code sort it out???     I am curious if we use `<float` and `<int` widely...?

solr/solr-ref-guide/modules/query-guide/pages/spell-checking.adoc
```diff
@@ -530,4 +597,4 @@ For example:
 http://localhost:8983/solr/techproducts/spell?spellcheck=true&spellcheck.build=true&spellcheck.q=toyata&shards.qt=/spell&shards=solr-shard1:8983/solr/techproducts,solr-shard2:8983/solr/techproducts
 
 In case of a distributed request to the SpellCheckComponent, the shards are requested for at least five suggestions even if the `spellcheck.count` parameter value is less than five.
-Once the suggestions are collected, they are ranked by the configured distance measure (Levenstein Distance by default) and then by aggregate frequency.
+Once the suggestions are collected, they are ranked by the configured distance measure (LevenshteinDistance by default) and then by aggregate frequency.
```
epugh, line 600: did you mean to remove the space?

epugh (approved):
Spell check is powerful but also very hard to understand, so this looks great...

epugh (changes requested):
Err, meant to call out a few small things...

epugh:
I'd love to review a PR that rationalizes the string/floats..    We would probably need to make the string version deprecated in 9x with a warning, and then require just the float in 10x...

solr/solr-ref-guide/modules/query-guide/pages/spell-checking.adoc
AndreyBozhko: Actually, I fixed the typo `Levenstein` -> `Levenshtein`, and then removed the space because this seems to refer to the actual class name?

solr/core/src/java/org/apache/solr/spelling/AbstractLuceneSpellChecker.java
```diff
@@ -81,7 +81,6 @@ public abstract class AbstractLuceneSpellChecker extends SolrSpellChecker {
   public String init(NamedList<?> config, SolrCore core) {
     super.init(config, core);
     indexDir = (String) config.get(INDEX_DIR);
-    String accuracy = (String) config.get(ACCURACY);
```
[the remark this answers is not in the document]
epugh: I think this is probably best for another JIRA..    I like consistency across things where possible because it just removes those small paper cut feelings that users get when trying out the various spell checkers...

solr/solr-ref-guide/modules/query-guide/pages/spell-checking.adoc
epugh: good question..   if we are referring to a class, shouldn't we link it to javadocs?

epugh:
Tests pass!, let me know if you are ready for this to be merged!

solr/core/src/java/org/apache/solr/spelling/AbstractLuceneSpellChecker.java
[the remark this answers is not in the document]
AndreyBozhko: Just to clarify - would you prefer to have this particular change about the `accuracy` type removed from this PR?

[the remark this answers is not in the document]
epugh: I think it's fine!  thanks...

AndreyBozhko:
Thanks for the review @epugh, please feel free to merge this

solr/core/src/java/org/apache/solr/spelling/WordBreakSolrSpellChecker.java
```diff
@@ -62,15 +62,15 @@ public class WordBreakSolrSpellChecker extends SolrSpellChecker {
   public static final String PARAM_MIN_BREAK_WORD_LENGTH = "minBreakLength";
 
   /** See {@link BreakSuggestionTieBreaker} for options. */
-  public static final String PARAM_BREAK_SUGGESTION_TIE_BREAKER = "breakSugestionTieBreaker";
+  public static final String PARAM_BREAK_SUGGESTION_TIE_BREAKER = "breakSuggestionTieBreaker";
```
cpoerschke, line 65: subjective: Wondering if for the 9.x line of releases we might want to maintain backwards compatibility for user convenience e.g.
```suggestion
  @Deprecated since 9.6 and will be removed from 10.0
  private static final String PARAM_BREAK_SUGESTION_TIE_BREAKER = "breakSugestionTieBreaker";

  public static final String PARAM_BREAK_SUGGESTION_TIE_BREAKER = "breakSuggestionTieBreaker";
```

here and reading both in the code. That way users upgrading (say) from 9.5 to 9.6 need not upgrade their configuration just as yet but they would receive a friendly WARNing logging or so to let them know about the rename.

epugh: Good catch...      In your suggested commit you don't have the warning right?  Okay, I guess what the plan is is to add the warning for the deprecation..   commit this and backport, and then in main we remove the mispelled variable.     and mention that "breaking change" in the upgrade notes to 10?

AndreyBozhko: Thanks for the inputs @cpoerschke and @epugh - I added the deprecation warning to the logs, and tweaked the changes entry to mention that the old parameter is deprecated.

epugh:
Thanks for this.   I'm actually a little surprised we don't have a well defined approach for sharing deprecation messages in logs..

epugh:
> Thanks for this. I'm actually a little surprised we don't have a well defined approach for sharing deprecation messages in logs..

I poked about and yeah, some things have helpers, but others it's just `log.warn` ;-).

epugh:
This is really great @AndreyBozhko.   Would you be interested in doing the follow up PR to remove the deprecated code from 10x?   And add an entry into the major changes for 10x release notes?

AndreyBozhko:
Sure @epugh - here's the PR: https://github.com/apache/solr/pull/2365

PR #2360, opened 2024-03-20
SOLR-17205 De-couple SolrJ Java version from server Java version

janhoy:
https://issues.apache.org/jira/browse/SOLR-17205

This PR only adds the plumbing, does not bump main

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,6 +30,10 @@ plugins {
   id 'com.github.node-gradle.node' version '7.0.1' apply false
 }
 
+// Declare default Java versions for the entire project and for SolrJ separately
+rootProject.ext.minJavaVersionDefault = JavaVersion.VERSION_11
+rootProject.ext.minJavaVersionSolrJ = JavaVersion.VERSION_11
+
 apply from: file('gradle/globals.gradle')
 
 // General metadata.
@@ -84,10 +88,6 @@ ext {
   buildDate = DateTimeFormatter.ofPattern("yyyy-MM-dd").format(tstamp)
   buildTime = DateTimeFormatter.ofPattern("HH:mm:ss").format(tstamp)
   buildYear = DateTimeFormatter.ofPattern("yyyy").format(tstamp)
-
-  minJavaVersion = JavaVersion.VERSION_11
-  // For client compatibility, build solrj artifacts for an older Java generation
-  minJavaVersionSolrJ = JavaVersion.VERSION_11
 
   // Allow definiting external tool locations using system props.
   externalTool = { name ->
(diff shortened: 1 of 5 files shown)


uschindler (changes requested):
I have not checked forbiddenapis and javadocs, but it won't be a deal-breaker there.

Maybe we should assign a per-project `ext.javaVersion` in the general settings file in the gradle root folder once and reuse it at all places where `rootProject.minJavaVersion` was used.

uschindler:
> Maybe we should assign a per-project `ext.javaVersion` in the general settings file in the gradle root folder once and reuse it at all places where `rootProject.minJavaVersion` was used.

Yes, I would do the following:
- Move the code which defines the Java version per project to `gradle/globals.gradle`. Maybe define it there and also use it in rootProject. Assign it to `ext.minJavaVersion`
- Remove definition of minJavaVersion in the root project (so it only exists in `globals.gradle`)
- grep through all grade files and replace `rootProject.manJavaVersion` by `project.minJavaVersion`

This is much cleaner.

janhoy:
I implemented your proposals @uschindler. I kept the definition of default and solrj java versions in main `build.gradle` for better visibility (needed to move it above the sourcing of `globals.gradle`) instead of burying it in other code. 

@HoustonPutman Which other modules are you thinking about, except the `solrj-*` ones?

IntelliJ is not good at sef-configuring language level, deleting `.idea`, it does not pick up the level to use for each module, any idea how to fix?

uschindler (approved):
Looks good now.

About the IDE settings it's complicated because in Solr's Eclipse build all is one big project with a single Java version. As the default version is generally >= SolrJ's version it works for Eclipse, the enforcement of compatibility with older versions is then only enforced by gradle (not the IDE). This is a known problem, not related to this issue. We have same problem with MemorySegment APIs in Lucene.

There is another thing: renderJavaDoc has hardcoded java version, see https://github.com/apache/lucene/pull/13132 for how to fix that. This PR should possibly be ported to Solr.

uschindler:
Will you port the renderJavadocs from Lucene, I think it might be a good idea.

janhoy:
I parked it here so we don't forget: SOLR-17223

HoustonPutman:
> @HoustonPutman Which other modules are you thinking about, except the `solrj-*` ones?

@janhoy Sorry I missed this, I'm thinking about the ones that would be used by SolrJ, like `hadoop-auth`, or `jwt-auth`. Or more security ones coming in the future.

Definitely not a blocker for this, but something to think about in the future, when we get these auth schemes working better with things like the CLI and SolrJ

janhoy:
I might be mis interpreting your statement. But as far as I know, all `modules` are server-side only. The two you mentioned both depend on `solr-core`.

HoustonPutman:
Currently yes, but there's a PR for one that might be SolrJ compatible: https://github.com/apache/solr/pull/1994. Just thinking about an eventual possibility.

PR #2362, opened 2024-03-20
HttpJdkSolrClient: Close the underlying http client, if it is supported (java 21+)

jdyer1:
By checking if the client implements AutoCloseable, we can properly close the underlying http client should the user have a Java 21+ jvm.


dsmiley (approved):
Thanks

risdenk (approved):

PR #2364, opened 2024-03-22
Update GC logging options for Zookeeper

radu-gheorghe:
So they work for Java 9+, given that we require Java 11+. Existing ones are for Java 8.

Manually tested these options (which are stolen from Solr's own GC logging options) on a 9.5 installation with Java 17.


dsmiley:
Thanks Radu!

radu-gheorghe:
My pleasure :)

PR #2375, opened 2024-03-25
[Experimental] Expose Prometheus metrics directly from Solr with prometheus client

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-10654

# Description

Experimental PR not meant for merging. Incomplete code.

This PR created a new response writer that directly transforms and bridges dropwizard to prometheus metrics 1 to 1.
`/solr/admin/metrics?wt=prometheus` 

Biggest issue is that 1 to 1 makes these metrics basically useless and an anti-pattern to prometheus naming that the `prometheus exporter` would take care of. This stems from the fact that Dropwizard doesn't have the support of tags/labels.

E.g
Dropwizard to Prometheus: 
`"QUERY./select.requests":0` -> `QUERY__select_requests 0.0`

Prometheus Exporter:
`solr_metrics_core_requests_total{category="QUERY",handler="/select",core="xyz_shard1_replica_t1",collection="xyz",shard="shard1",replica="replica_t1",base_url="localhost:8983/solr",}`

Would need to find some way to either parse the dropwizard registry and append tags to them with prometheus or add new way that Solr collects metrics and creates labels (prometheus_client registry in parallel or possibly over engineered effort?)


mlbiscoc:
Closing. New PR with correct implementation in https://github.com/apache/solr/pull/2405

PR #2378, opened 2024-03-29
SOLR-17194: CloudHttp2SolrClient doesn't properly use httpClient Builder on certain paths

epugh:
https://issues.apache.org/jira/browse/SOLR-17194

# Description

Discovered a bug in the use of Export Tool with Basic Auth parameters doesn't create a proper CloudHttp2SolrClient.  

Undercovered that we pass in a httpClient builder to CloudHttp2SolrClient, but in at least this code path we never use, so we have a null httpClient, which means we build a default one, which means if you passed one in then you don't get your Basic Auth params!

# Solution

A test and a small fix.   I did not try to reasoon about all the builder stuff..  I find the combination of httpClient, and the solrClient and their various builders VERY hard to understand...   I wrote at test, I fixed the code, the test passes...

# Tests

Added a test to the Export Tool

# Checklist

### DIFF
--- a/solr/core/src/test/org/apache/solr/cli/TestExportTool.java
+++ b/solr/core/src/test/org/apache/solr/cli/TestExportTool.java
@@ -263,7 +263,7 @@ public class TestExportTool extends SolrCloudTestCase {
 
     CloudSolrClient cloudSolrClient = cluster.getSolrClient();
 
-    CollectionAdminRequest.createCollection(COLLECTION_NAME, "conf", 8, 1)
+    CollectionAdminRequest.createCollection(COLLECTION_NAME, "conf", 2, 1)
         .setBasicAuthCredentials(USER, PASS)
         .process(cluster.getSolrClient());
     cluster.waitForActiveCollection(COLLECTION_NAME, 8, 8);
--- a/solr/core/src/test/org/apache/solr/cli/TestExportTool.java
+++ b/solr/core/src/test/org/apache/solr/cli/TestExportTool.java
@@ -17,6 +17,12 @@
 
 package org.apache.solr.cli;
 
+import static java.util.Collections.singletonList;
+import static java.util.Collections.singletonMap;
+import static org.apache.solr.cli.SolrCLI.findTool;
+import static org.apache.solr.cli.SolrCLI.parseCmdLine;
+import static org.apache.solr.security.Sha256AuthenticationProvider.getSaltedHashedValue;
+
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
@@ -27,10 +33,12 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Predicate;
+import org.apache.commons.cli.CommandLine;
 import org.apache.lucene.tests.util.TestUtil;
 import org.apache.solr.SolrTestCaseJ4;
 import org.apache.solr.client.solrj.SolrClient;
 import org.apache.solr.client.solrj.SolrQuery;
+import org.apache.solr.client.solrj.impl.CloudSolrClient;
 import org.apache.solr.client.solrj.impl.Http2SolrClient;
 import org.apache.solr.client.solrj.request.AbstractUpdateRequest;
 import org.apache.solr.client.solrj.request.CollectionAdminRequest;
@@ -43,10 +51,15 @@ import org.apache.solr.common.cloud.Replica;
 import org.apache.solr.common.cloud.Slice;
 import org.apache.solr.common.util.FastInputStream;
 import org.apache.solr.common.util.JsonRecordReader;
+import org.apache.solr.common.util.Utils;
+import org.apache.solr.security.BasicAuthPlugin;
+import org.apache.solr.security.RuleBasedAuthorizationPlugin;
+import org.junit.Test;
 
 @SolrTestCaseJ4.SuppressSSL
 public class TestExportTool extends SolrCloudTestCase {
 
+  @Test
   public void testBasic() throws Exception {
     String COLLECTION_NAME = "globalLoaderColl";
     configureCluster(4).addConfig("conf", configset("cloud-dynamic")).configure();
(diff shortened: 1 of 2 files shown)


epugh:
Pinged community for feedback, I'd like to merge this week...

solr/core/src/test/org/apache/solr/cli/TestExportTool.java
```diff
@@ -218,6 +231,66 @@ public void testVeryLargeCluster() throws Exception {
...
+                    singletonMap(USER, getSaltedHashedValue(PASS)))));
+
+    configureCluster(2)
+        .addConfig("conf", configset("cloud-minimal"))
+        .withSecurityJson(SECURITY_JSON)
+        .configure();
+
+    CloudSolrClient cloudSolrClient = cluster.getSolrClient();
+
+    CollectionAdminRequest.createCollection(COLLECTION_NAME, "conf", 8, 1)
```
dsmiley, line 266: 8 shards?!
To tame our tests, tests should have a comment justifying >1 nodes or >1 shards.  If there is a reason; justify it.

epugh: total copy paste ;-)

epugh: Oh, wait...  I found it..  The test is `testVeryLargeCluster` and it's a nightly...   so I think the is why....

epugh: do you want another specific comment on the test....

epugh:
@dsmiley do the rest of the changes look okay to you?

solr/core/src/test/org/apache/solr/cli/TestExportTool.java
epugh: I made the change to reduce the number...

solr/core/src/test/org/apache/solr/cli/TestExportTool.java
```diff
@@ -218,6 +231,66 @@ public void testVeryLargeCluster() throws Exception {
...
+                "authentication",
+                Map.of(
+                    "class",
+                    BasicAuthPlugin.class.getName(),
+                    "blockUnknown",
+                    true,
+                    "credentials",
+                    singletonMap(USER, getSaltedHashedValue(PASS)))));
+
+    configureCluster(2)
```
dsmiley, line 259: Can the behavior be tested with one node instead of two?  Authentication seems unrelated to the size of the cluster.

solr/core/src/test/org/apache/solr/cli/TestExportTool.java
```diff
@@ -218,6 +231,66 @@ public void testVeryLargeCluster() throws Exception {
...
+        .addConfig("conf", configset("cloud-minimal"))
+        .withSecurityJson(SECURITY_JSON)
+        .configure();
+
+    CloudSolrClient cloudSolrClient = cluster.getSolrClient();
+
+    CollectionAdminRequest.createCollection(COLLECTION_NAME, "conf", 2, 1)
+        .setBasicAuthCredentials(USER, PASS)
+        .process(cluster.getSolrClient());
+    cluster.waitForActiveCollection(COLLECTION_NAME, 8, 8);
```
dsmiley, line 269: You won't get 8 replicas anymore; you changed this.  The test probably fails now.  I hate `waitForActiveCollection`; the collection is already active when createCollection finishes.  The risk this method mitigates is that the SolrClient returned by `cluster.getSolrClient` doesn't know it yet, but this test doesn't use the client henceforth (unlike most tests).

solr/core/src/test/org/apache/solr/cli/TestExportTool.java
```diff
@@ -218,6 +231,66 @@ public void testVeryLargeCluster() throws Exception {
...
+                    singletonMap(USER, getSaltedHashedValue(PASS)))));
+
+    configureCluster(2)
+        .addConfig("conf", configset("cloud-minimal"))
+        .withSecurityJson(SECURITY_JSON)
+        .configure();
+
+    CloudSolrClient cloudSolrClient = cluster.getSolrClient();
+
+    CollectionAdminRequest.createCollection(COLLECTION_NAME, "conf", 2, 1)
```
dsmiley, line 266: Two shards doesn't seem pertinent to what is being tested.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudHttp2SolrClient.java
```diff
@@ -422,7 +422,10 @@ public CloudHttp2SolrClient build() {
           }
         } else if (!solrUrls.isEmpty()) {
           try {
-            stateProvider = new Http2ClusterStateProvider(solrUrls, httpClient);
+            if (this.internalClientBuilder != null) {
+              this.httpClient = internalClientBuilder.build();
```
stillalex, line 426: I think there might be an issue here with the `httpClient` instance created here is not being closed correctly. not 100% sure I will need to check more closely. before this change I agree there was a default client being created lacking the correct params but it also closed.

epugh: thanks for chiming in...  so, this change is making my TestExportTool blow up with unclosed httpclients..   I've banged my head on this for a LONG time, and not making headway.   What I can say is that the test with the BasicAuth demonstrates the problem, but I can't grok the soluiton...   @stillalex I would love your ideas if you wanted to take a stab at it..   it felt like such an easy fix!

laminelam: That's funny, I pushed 2 weeks ago an internal PR (at my work) for this exact same bug, as I faced it myself.

@stillalex  is correct, the resource won't be released if an exception is raised when instantiating the  _Http2ClusterStateProvider_

Plus we end up creating two _httpCient_ instances. CloudHttp2SolrClient and stateProvider would use 2 different _httClient_s.

In my PR, I pushed most of the code into the CloudHttp2SolrClient constructor. I added a test case to test if the same _httpClient_ is preserved in both _CloudHttp2SolrClient_ and _Http2ClusterStateProvider_.
All test cases passed, including _TestExportTool_.

Will see if I can push the PR here (upstream).

epugh: Thanks all for looking at this.  I have definitly hit the "end" of my rope on this one, just can't figure it out...   My Java-fu isn't up to this one.   

I am inclined to close this PR in favour of what you can whip up..  However, since I really really really really really really want basic auth to work on all the tools, I am hesitating to actually do that!

laminelam: Actually I have 2 PRs that I am pretty sure would help a lot on this.
One that fixes this bug.
And anther one that I talked about [here](https://github.com/apache/solr/pull/1994#discussion_r1535512663) where I refactored ZK related SolrCli tools to separate _SolrZkClient_ and _CloudHttp2SolrClient_  instantiation/usage.

Will see if I can speed up the internal review and push them upstream.

laminelam: @epugh 
Here's the [PR](https://github.com/apache/solr/pull/2412)
Tested (locally) your testWithBasicAuth() code and the test passed

epugh:
Closing in favour of #2412

PR #2379, opened 2024-03-31
SOLR-17158  Terminate distributed processing quickly when query limit is reached - Initial impl

gus-asf:
still needs better tests, and it is still incorrectly caching empty results. Also, a few TODOs remain.

https://issues.apache.org/jira/browse/SOLR-17158


solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
```diff
@@ -227,18 +243,27 @@ public ShardResponse takeCompletedOrError() {
 
   private ShardResponse take(boolean bailOnError) {
     try {
+      // although nothing in this class guarantees that pending has been incremented to the total
```
gus-asf, line 246: @CaoManhDat if you can double check my reading of this code and let me know if this comment is inaccurate in any way I'd appreciate it. (I see your name in the git blame a bunch here)

gus-asf:
One thing folks may want to review is the addition of synchronization around manipulations of HttpShardHandler#responseCancelableMap which were necessary to avoid CME such as:
```
{
  "responseHeader":{
    "zkConnected":true,
    "status":500,
    "QTime":18,
    "params":{
      "q":"document",
      "indent":"true",
      "q.op":"OR",
      "timeAllowed":"5",
      "allowPartialResults":"false",
      "useParams":"",
      "_":"1711320289606"    }  },
  "error":{
    "trace":"java.util.ConcurrentModificationException
at java.base/java.util.HashMap$HashIterator.nextNode(HashMap.java:1597)
at java.base/java.util.HashMap$ValueIterator.next(HashMap.java:1625)
at org.apache.solr.handler.component.HttpShardHandler.cancelAll(HttpShardHandler.java:257)
at org.apache.solr.handler.component.SearchHandler.handleRequestBody(SearchHandler.java:577)
at org.apache.solr.handler.RequestHandlerBase.handleRequest(RequestHandlerBase.java:238)
at org.apache.solr.core.SolrCore.execute(SolrCore.java:2886)
at org.apache.solr.servlet.HttpSolrCall.executeCoreRequest(HttpSolrCall.java:876)
at org.apache.solr.servlet.HttpSolrCall.call(HttpSolrCall.java:560)
at org.apache.solr.servlet.SolrDispatchFilter.dispatch(SolrDispatchFilter.java:254)
at org.apache.solr.servlet.SolrDispatchFilter.lambda$doFilter$0(SolrDispatchFilter.java:215)
at org.apache.solr.servlet.ServletUtils.traceHttpRequestExecution2(ServletUtils.java:247)
at org.apache.solr.servlet.ServletUtils.rateLimitRequest(ServletUtils.java:211)
at org.apache.solr.servlet.SolrDispatchFilter.doFilter(SolrDispatchFilter.java:209)
at org.apache.solr.servlet.SolrDispatchFilter.doFilter(SolrDispatchFilter.java:192)
at org.eclipse.jetty.servlet.FilterHolder.doFilter(FilterHolder.java:210)
at org.eclipse.jetty.servlet.ServletHandler$Chain.doFilter(ServletHandler.java:1635)
```

This happens because the code is iterating an unsynchronized hash map when the request thread signals a cancel operation. I chose to also enclose pending.incrementAndGet() or decrementAndGet() and responses.take() and even though they are by themselves probablly safe, there's a loop on the value of pending > 0 and these three operations are logically linked so it seemed a bit dangerous to leave them independent.

Also I noticed that ShardeResponse.code is being set, but the IDE flags it as "assigned but never read", and against the recommendation of the previously existing comments it's not volatile...

solr/core/src/java/org/apache/solr/search/Grouping.java
```diff
@@ -461,7 +463,7 @@ private void searchWithTimeLimiter(final Query filterQuery, Collector collector)
       log.info("Query: {}; ", query);
       // to make WARN logged exception content more visible
       log.warn("Query: {}; ", query.getClass().getName(), x);
-      qr.setPartialResults(true);
+      qr.setPartialResults(shouldKeepPartials());
```
sigram, line 466: Hmm, I'm not sure if this is correct - when these exceptions are thrown we know for sure that results are incomplete. Why would we set `partialResults=false` here? ? Unless there's an implicit behavior here that if we discard the results then we (always?) return `partialResults=false`.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -276,8 +278,8 @@ private Collector buildAndRunCollectorChain(
     } catch (TimeLimitingCollector.TimeExceededException
         | ExitableDirectoryReader.ExitingReaderException
         | CancellableCollector.QueryCancelledException x) {
-      log.warn("Query: [{}]; ", query, x);
-      qr.setPartialResults(true);
+      log.warn("Query: [{}]; was timed out or canceled.", query, x);
+      qr.setPartialResults(shouldKeepPartials());
```
sigram, line 282: See above.

solr/core/src/java/org/apache/solr/search/grouping/CommandHandler.java
```diff
@@ -224,7 +226,7 @@ public NamedList<NamedList<Object>> processResult(
     if (docSet != null) {
       queryResult.setDocSet(docSet);
     }
-    queryResult.setPartialResults(partialResults);
+    queryResult.setPartialResults(partialResults && shouldKeepPartials());
```
sigram, line 229: See above.

solr/core/src/test/org/apache/solr/search/TestCpuAllowedLimit.java
```diff
@@ -215,4 +225,116 @@ public void testDistribLimit2() throws Exception {
...
+                "id desc",
+                ExpensiveSearchComponent.CPU_LOAD_COUNT_PARAM,
+                "1",
+                "stages",
+                "prepare,process",
+                "cpuAllowed",
+                "1000"));
+    assertNull("should have full results", rsp.getHeader().get("partialResults"));
+
+    // cpuAllowed set, should return partial results
```
sigram, line 298: Ditto.

solr/core/src/test/org/apache/solr/search/TestCpuAllowedLimit.java
```diff
@@ -215,4 +225,116 @@ public void testDistribLimit2() throws Exception {
...
+                String.valueOf(sleepMs),
+                "stages",
+                "prepare,process"));
+    // System.err.println("rsp=" + rsp.jsonStr());
+    assertEquals(rsp.getHeader().get("status"), 0);
+    Number qtime = (Number) rsp.getHeader().get("QTime");
+    assertTrue("QTime expected " + qtime + " >> " + sleepMs, qtime.longValue() > sleepMs);
+    assertNull("should not have partial results", rsp.getHeader().get("partialResults"));
+
+    // timeAllowed set, should return partial results
```
sigram, line 257: This comment is incorrect.

solr/core/src/test/org/apache/solr/search/TestCpuAllowedLimit.java
```diff
@@ -215,4 +225,116 @@ public void testDistribLimit2() throws Exception {
...
+                ExpensiveSearchComponent.CPU_LOAD_COUNT_PARAM,
+                "10",
+                "stages",
+                "prepare,process",
+                "cpuAllowed",
+                "50"));
+    assertNull("should not have partial results", rsp.getHeader().get("partialResults"));
+    assertEquals(0, rsp.getResults().size());
+
+    // cpuAllowed set, should return partial results
```
sigram, line 319: Ditto.

solr/solrj/src/java/org/apache/solr/client/solrj/util/AsyncListener.java
```diff
@@ -19,10 +19,23 @@
 
 /** Listener for async requests */
 public interface AsyncListener<T> {
-  /** Callback method invoked before processing the request */
+  /**
+   * Setup method invoked in the original thread before processing the request on a different
```
sigram, line 23: 💯 very important clarification that explains why we're later getting the right `SolrRequestInfo`.

sigram (changes requested):

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -497,7 +495,12 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
           debug.add("explain", new NamedList<>());
           rb.rsp.add("debug", debug);
         }
-        rb.rsp.setPartialResults();
+        if (req.shouldDiscardPartials()) {
+          // TODO: this isn't actually showing up... possibly need to do tis at a different point.
+          rb.rsp.getResponseHeader().add("partialResultsOmitted", "true");
```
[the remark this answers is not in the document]
gus-asf: :+1:

solr/core/src/java/org/apache/solr/request/json/RequestUtil.java
```diff
@@ -216,24 +222,26 @@ public static void processParams(
...
           arr = true;
           isQuery = true;
         } else if ("fields".equals(key)) {
-          out = "fl";
+          out = FL;
           arr = true;
         } else if ("offset".equals(key)) {
-          out = "start";
+          out = START;
+        } else if ("allowPartialResults".equals(key)) {
```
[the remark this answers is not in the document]
gus-asf: Hmm I didn't perceive the mechanism by which it would be respected without this. I'll look again.

solr/core/src/java/org/apache/solr/search/Grouping.java
gus-asf: This seemed simplest and more readable. Code checks this value upstream in addition to it getting reported in the final output. Maybe if I track those down and ensure the up stream parsing handles it, this could be set to "omitted" that could remove the need for an additional response header value hmm.

solr/solrj/src/java/org/apache/solr/client/solrj/util/AsyncListener.java
gus-asf: yeah got very scared initially when I saw SolrRequestInfo referenced in an implementation of AsyncListner

solr/solrj/src/java/org/apache/solr/common/params/CommonParams.java
```diff
@@ -228,6 +228,7 @@ public interface CommonParams {
           METRICS_PATH);
   String APISPEC_LOCATION = "apispec/";
   String INTROSPECT = "/_introspect";
+  String ALLOW_PARTIAL_RESULTS = "allowPartialResults";
```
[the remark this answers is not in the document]
gus-asf: Yup, will check/fix

solr/solrj/src/java/org/apache/solr/common/params/ShardParams.java
```diff
@@ -118,12 +122,27 @@ public interface ShardParams {
...
+            "Use of "
+                + SHARDS_TOLERANT
+                + " requires that "
+                + ALLOW_PARTIAL_RESULTS
+                + " is true. If "
+                + ALLOW_PARTIAL_RESULTS
+                + " is defaulted to true for this handler or system wide, explicitly"
+                + "passing "
+                + ALLOW_PARTIAL_RESULTS
+                + "=true will allow shards.tolerant to work");
```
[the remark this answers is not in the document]
gus-asf: oops, yeah s/true/false there. That's a typo from rewriting it six times :), also you're right it probably is givign too much detail...

solr/core/src/java/org/apache/solr/request/json/RequestUtil.java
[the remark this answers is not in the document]
gus-asf: removed it and it still works. I still can't pick apart this code sufficiently to verify why... this area could use either comments or refactor I think.

solr/core/src/java/org/apache/solr/search/Grouping.java
gus-asf: Next commit will contain the "omitted" value I mention, but discussion elsewhere indicates that even though I can maintain back compat it creates a tricky situation where the attribute is only sometimes boolean (and unquoted in the json) and other times can be a string (with quotes).  I'm probably better off just deprecating the partialResults flag and making a new string based attribute, perhaps "partialResultStatus" that can be "complete results", "partial results" or "results omitted" - that will come in the following commit

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
```diff
@@ -140,58 +150,64 @@ public void submit(
...
-      SolrException exception =
-          new SolrException(
-              SolrException.ErrorCode.SERVICE_UNAVAILABLE, "no servers hosting shard: " + shard);
-      srsp.setException(exception);
-      srsp.setResponseCode(exception.code());
-      responses.add(srsp);
-      return;
-    }
+      // all variables that set inside this listener must be at least volatile
+      responseCancellableMap.put(
```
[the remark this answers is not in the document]
gus-asf: I didn't feel comfortable with that solution because this map is updated while looping on an Atomic integer `pending` and also is updated immediately after `responses.take()` all of which appear to need to happen as a unit. The prior code wasn't touching it from the child threads, but since I'm changing that I want to be sure that all of these operations happen as a unit to avoid potentially weird cases where the loop fails to exit but then the map is empty etc...

solr/solrj/src/java/org/apache/solr/client/solrj/request/RequestParamsSupplier.java
```diff
@@ -0,0 +1,70 @@
...
+  //  variations in their solrconfg.xml that prevent simple search/replace this would be
+  //  impractical. Thus a means of setting a global default seems appropriate. System property is
+  //  a quick and dirty way but In "user managed" mode  it could possibly be a solr.xml (user's
+  //  responsibility to manage, similar to a system property), and in cloud mode this maybe should
+  //  hinge on a  cluster property since there is no conceivable reason to have individual nodes
+  //  with differing defaults. For the cloud case we probably would want to ignore the setting in
+  //  solr.xml to guard against inconsistent shard behavior due to misconfiguration. A further
+  //  consideration is how things like a docker compose or kubernetes system ought to be able
+  //  to initialize cluster properties like this for consistent non-interactive deployment...
+  String SOLR_DISLIKE_PARTIAL_RESULTS = "solr.dislike.partial.results";
```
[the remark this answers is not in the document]
gus-asf: Solr.xml with property substitution it is

solr/core/src/java/org/apache/solr/core/SolrXmlConfig.java
```diff
@@ -384,6 +384,9 @@ private static NodeConfig fillSolrSection(NodeConfig.NodeConfigBuilder builder,
                 log.warn("solr.xml transientCacheSize -- transient cores is deprecated");
                 builder.setTransientCacheSize(it.intVal(-1));
                 break;
+              case "allowPartialResultsDefault":
+                builder.setAllowPartialResultsDefault(it.boolVal(true));
+                break;
```
dsmiley, line 389: I would prefer you not bother with all the machinery in solr.xml/NodeConfig and instead have a simple EnvUtils (env or sys prop) to enable.  It's just a boolean; doesn't have any interesting config to it.  Yes there are other booleans here but I think we Solr maintainers should consider how NodeConfig might systematically / dynamically understand primitive values (boolean, integer, ...) and apply to EnvUtils automatically without having to touch the NodeConfig class (which is kind of a pain; that builder!).  For example imagine "solr.search.partialResults" being settable as a system property, or also settable in solr.xml automatically via "searchPartialResults".

CC @janhoy 

Another take on this is that, shouldn't the user simply go to their solrconfig.xml and set a default partialResults like we all do for miscellaneous things for search?

sigram: I agree, IMHO having a per-collection setting instead of a global sysprop is more flexible (and you can still change it globally using property substitution).

solr/core/src/java/org/apache/solr/request/SolrQueryRequest.java
```diff
@@ -39,6 +40,39 @@
...
+   * @return true if partials should be discarded.
+   * @param params the request parameters
+   */
+  static boolean shouldDiscardPartials(SolrParams params) {
+    Boolean userParamAllowPartial = params.getBool(CommonParams.ALLOW_PARTIAL_RESULTS);
+    if (userParamAllowPartial != null) {
+      return !userParamAllowPartial;
+    } else {
+      return ALLOW_PARTIAL_RESULTS_DEFAULT;
+    }
```
dsmiley, line 73: can we use the overloaded version of getBool that provides a default?  We'd then have a one-liner here.

dsmiley (commented):
Wow, this PR touches 33 files, which is super surprising compared to https://github.com/apache/solr/pull/2493
Can this be simplified or split up?

gus-asf:
> Wow, this PR touches 33 files, which is super surprising compared to #2493 Can this be simplified or split up?

Ok finally back to this. Sorry about the delay, the merge of a multi-threading implementation that broke (or prevented se of) features I had just released co-opted most of my available Solr time...

gus-asf:
Reworked version compatible with current main, and touching fewer files here: https://github.com/apache/solr/pull/2666

PR #2387, opened 2024-04-03
SOLR-17223 Remove hardcoded java version in renderJavadoc task

janhoy:
Oneliner :)

### DIFF
--- a/gradle/documentation/render-javadoc.gradle
+++ b/gradle/documentation/render-javadoc.gradle
@@ -370,7 +370,7 @@ class RenderJavadocTask extends DefaultTask {
       opts << [ '-linkoffline', url, dir ]
     }
 
-    opts << [ '--release', 11 ]
+    opts << [ '--release', project.minJavaVersion.toString() ]
     opts << '-Xdoclint:all,-missing'
 
     // Increase Javadoc's heap.


gradle/documentation/render-javadoc.gradle
```diff
@@ -370,7 +370,7 @@ class RenderJavadocTask extends DefaultTask {
       opts << [ '-linkoffline', url, dir ]
     }
 
-    opts << [ '--release', 11 ]
+    opts << [ '--release', project.minJavaVersion.toString() ]
```
cpoerschke, line 373: https://github.com/apache/solr/blob/2eb9b21d434da748d26ec82cc2b76aa1fd41237b/gradle/globals.gradle#L175 approximately looks to be where `project.minJavaVersion` comes from.

uschindler:
The approach you used did not work in Lucene because inputs/outputs and "rootProject" was not known in the task instance. Here it is different, if it works, +1

cpoerschke (approved):
LGTM

uschindler:
P.S.: I missed this in Lucene, there is also a reference to the JDK Javadocs and its package file above. We may also make this dynamic.

gradle/documentation/render-javadoc.gradle
janhoy: That's correct, each project wil have this variable set, so javadoc for SolrJ will use 11 while core may use 17 eventually.

uschindler:
> Will take another look at the javadoc links. What gradle command do I run to build javadocs that should have substituted local javadocs? Do I have to unzip some jdk javadoc file into my git checkout and then run gradlew javadocs?

See those commits: https://github.com/apache/lucene/commit/dfce6ee8d29da75b67fa0679e192c4defa9f9f78 and https://github.com/apache/lucene/commit/5aaaeaee391934363b207718c6eea7853db3ca81

The coresponding file can be found in the ZIP download of JDK (it may be named 'package-list' or 'element-list'. You can also get it from the official Javadocs page by appending that filename instaed of index.html. Maybe we should generate an automated task to regenerate ist.

To generate the javadocs on Lucene, run `gradlew documentation`, this will build both local Javadocs as well as full site. If you want to quickly test a single project use: `gradlew :solr:xyz:javadoc`

epugh (approved):
Seems like this is needed with upcoming Java bumps!!!

epugh:
@janhoy is this ready for merging?

uschindler:
I just merged this.

PR #2388, opened 2024-04-03
Use hostContext property for RewriteHandler rules

wwelling:

# Description

# Solution

# Tests

# Checklist

### DIFF
--- a/solr/server/etc/jetty.xml
+++ b/solr/server/etc/jetty.xml
@@ -93,7 +93,7 @@
       <Call name="addRule">
         <Arg>
           <New class="org.eclipse.jetty.rewrite.handler.HeaderPatternRule">
-            <Set name="pattern">/solr/*</Set>
+            <Set name="pattern"><Property name="hostContext" default="/solr"/>/*</Set>
             <Set name="name">Content-Security-Policy</Set>
             <Set name="value">default-src 'none'; base-uri 'none'; connect-src 'self'; form-action 'self'; font-src 'self'; frame-ancestors 'none'; img-src 'self' data:; media-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self'; worker-src 'self';</Set>
           </New>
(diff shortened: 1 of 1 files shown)


HoustonPutman:
I believe we recently removed the ability to specify a new hostContext in `main`. @epugh @gerlowskija

wwelling:
@HoustonPutman thanks for the quick reply. I have been reading about that in the Jira and previous issues/PRs.

I noticed `hostContext` is still used in [solr-jetty-context.xml](https://github.com/apache/solr/blob/main/solr/server/contexts/solr-jetty-context.xml#L4).

To you by chance now why Solr is removing ability to configure its context path?

epugh:
IIRC it was actually not well supported in recent versions of Solr, it may even not have worked at all in the 8x line...   With the V2 api's starting to live under `/api`, we want to make `/solr` and `/api` fixed paths.   I don't recall why, but this did faciliate the v2 api stuff...   Maybe @gerlowskija  can shed osme light.    

@wwelling I'd love to hear the use cases that you would *want* this for..  make sure we haven't shut the door on some reasons..

wwelling:
Hi @epugh, our use case is through a constraint from load balancing requiring matching path with application context path. We have a solution with an additional application sidecar proxy but would like to be able to configure Solr's Jetty servlet container.

dsmiley:
@wwelling You can configure Solr's Jetty to map `/foo` to `/solr` if that would help.

I think the overall objective was to reduce internal code maintenance on an option that didn't seem worth supporting.

wwelling:
@dsmiley, thanks. I'll look into additional Jetty routing. Do think it is a reasonable request of the [Solr Docker](https://hub.docker.com/_/solr) image to afford environment driven context path configuration?

dsmiley:
ehhh, no I don't TBH.  At least it's somewhat easy to create a new server/etc/jetty-customAdditions.xml and it'll be loaded.

wwelling:
@dsmiley thanks. We will give that a try. Would be easy enough to mount a file for additional routing. Will have to see if it supports system variable or environment variable substitution. Otherwise, the context path configuration will still not be [12-factor](https://12factor.net/config).

dsmiley:
My concern is avoiding complexity 99+% of users don't need.

epugh:
@wwelling based on the conversation above, I am going to close this PR.  If you decide to move forward with the nesting of both `/solr` and `/api` as a new feature, please open a JIRA, and I'd be happy to review your PR!

PR #2391, opened 2024-04-05
SOLR-18094 Support running embedded-zk in "ensemble" mode with new node role

gerlowskija:
# Description

Prior to this commit, Solr only supported running embedded-ZK in "standalone" mode (which cannot take part in any larger ZK ensemble or quorum). But there are usecases that would benefit from being able to do this, both on the development and testing side, and even for adventurous users who might want the benefits of a small multi-node SolrCloud cluster without the headache of also deploying ZK.

See [SIP-14](https://cwiki.apache.org/confluence/display/SOLR/SIP-14+Embedded+Zookeeper), SOLR-15636 for context.

> **⚠️ Experimental / Alpha feature** — this feature is intentionally rough and should not be used in production. The API and configuration format may change without notice.

# Solution

This commit augments the embedded-ZK code to support running in "quorum" or ensemble mode. Multiple Solr nodes can now all have their embedded ZK join a multi-node quorum upon startup. Other than Solr and ZK sharing a process, the embedded-ZK ensemble behaves identically to one formed of independent processes.

Embedded-ensemble-ZK is enabled by assigning the new `zookeeper_quorum` node role (`on`/`off`) via the `solr.node.roles` system property, along with an explicit ZK host string containing all quorum members. Solr identifies which host in the ZK connection string it should be (based on host/port matching) and starts a `ZooKeeperServerEmbedded` instance in-process to join the ensemble. For example:

```
export LH="localhost"

# Run each below in a separate terminal, one for each host in the zk-conn string
bin/solr start -p 8983 -z $LH:9983,$LH:9984,$LH:9985 -Dsolr.node.roles=data:on,overseer:allowed,zookeeper_quorum:on
bin/solr start -p 8984 -z $LH:9983,$LH:9984,$LH:9985 -Dsolr.node.roles=data:on,overseer:allowed,zookeeper_quorum:on
bin/solr start -p 8985 -z $LH:9983,$LH:9984,$LH:9985 -Dsolr.node.roles=data:on,overseer:allowed,zookeeper_quorum:on
```

> Note: ZK client port is `solr_port + 1000`, quorum port is `zk_client_port + 1`, election port is `zk_client_port + 2`.

Known limitations (Phase 1 / experimental):
- Does not work with ZK's dynamic-ensemble feature — all ZK nodes must be specified in a static connection string at startup.
- Quorum loss and recovery is not yet reliable (see `testQuorumLossAndRecovery` marked `@AwaitsFix`).
- Requires that hostnames in the ZK connection string exactly match Solr's configured host (`solr.host`).
- Runs best with the security manager disabled.
- The configuration interface is rough and subject to change.

# Tests

* `TestEmbeddedZkQuorum` — basic quorum startup, collection creation, and node join/leave scenarios

# Checklist


gerlowskija:
FYI: https://cwiki.apache.org/confluence/display/SOLR/SIP-14+Embedded+Zookeeper for context on others' interest in moving this direction.

janhoy:
Great work. I think we should rip out the existing embedded ZK, which is a hack.

Do you want to proceed in a PR or perhaps in a central feature branch?

gerlowskija:
Spent a few minutes pulling in the latest 'main'.  Hoping to give it a bit more cleanup in the next few days if I can, but I've been hoping that for more than a year at this point (😬 ), so if anyone else is interested please feel free to move this forward!

epugh:
@gerlowskija I took a stab at getting in the use of the solr roles to determine if we hsould run embedded zk quorum mode, and it worked!  Here is my start script, where each runs with `-f` so you need three terminals.  Omit the `-f` and you don't need the seperate terminals.  

```
mkdir -p ./three_nodes/node1/solr
mkdir -p ./three_nodes/node2/solr
mkdir -p ./three_nodes/node3/solr

SOLR_SECURITY_MANAGER_ENABLED=false bin/solr start -f -p 8983 --solr-home "./three_nodes/node1/solr" -z localhost:9983,localhost:9984,localhost:9985 -Dsolr.node.roles=data:on,overseer:allowed,zookeeper_quorum:on
SOLR_SECURITY_MANAGER_ENABLED=false bin/solr start -f -p 8984 --solr-home "./three_nodes/node2/solr" -z localhost:9983,localhost:9984,localhost:9985 -Dsolr.node.roles=data:on,overseer:allowed,zookeeper_quorum:on
SOLR_SECURITY_MANAGER_ENABLED=false bin/solr start -f -p 8985 --solr-home "./three_nodes/node3/solr" -z localhost:9983,localhost:9984,localhost:9985 -Dsolr.node.roles=data:on,overseer:allowed,zookeeper_quorum:on
```

janhoy:
* Merged in main branch
* Properly clean up the ZK server resource in close()

solr/core/src/java/org/apache/solr/core/ZkContainer.java
```diff
@@ -192,9 +193,13 @@ public void initZooKeeper(final CoreContainer cc, CloudConfig config) {
       // we are ZooKeeper enabled
       try {
         // If this is an ensemble, allow for a long connect time for other servers to come up
-        if (zkServerEnabled && zkServer.getServers().size() > 1) {
+        if (zkServerEnabled && zkServer != null && zkServer.getServers().size() > 1) {
           zkClientConnectTimeout = 24 * 60 * 60 * 1000; // 1 day for embedded ensemble
           log.info("Zookeeper client={}  Waiting for a quorum.", zookeeperHost);
+        } else if (zkServerEnabled && runAsQuorum) {
+          // Quorum mode also needs long timeout for other nodes to start
+          zkClientConnectTimeout = 24 * 60 * 60 * 1000; // 1 day for embedded quorum
```
epugh, line 269: Lovely.   I think we were trying to put units in the variable names in other places ;-).

solr/core/src/java/org/apache/solr/core/ZkContainer.java
```diff
@@ -326,14 +331,17 @@ public SolrMetricsContext getSolrMetricsContext() {
...
-    p.load(new FileInputStream(zkHomeDir + "/zoo.cfg"));
+    try (FileInputStream fis = new FileInputStream(zkHomeDir + "/zoo.cfg")) {
+      p.load(fis);
+    }
     p.setProperty("clientPort", String.valueOf(port));
 
-    // TODO NOCOMMIT - hang onto the created ZooKeeperServerEmbedded to be able to close it
-    // gracefully the way we do today with zkServer
-    ZooKeeperServerEmbedded.builder().baseDir(Path.of(zkHomeDir)).configuration(p).build().start();
+    zkServerEmbedded =
```
epugh, line 396: Ah ha!  The magic!

epugh:
FYI, the `.bats` test worked a few times for me and then quit working ;-(

janhoy:
With the help of Claude Code I added a JUnit test `TestEmbeddedZkQuorum`, that actually spins up a quorum cluster with `zookeeper_quorum:on` node roles (after securing a port so we know beforehand what the zkHost string will be).

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
```diff
@@ -342,6 +344,185 @@ public MiniSolrCloudCluster(
...
+      throw new IllegalArgumentException(
+          "ZooKeeper quorum requires at least 3 nodes, got: " + numServers);
+    }
+
+    Objects.requireNonNull(securityJson);
+    this.baseDir = Objects.requireNonNull(baseDir);
+    this.jettyConfig = Objects.requireNonNull(jettyConfig);
+    this.solrXml = solrXml == null ? DEFAULT_CLOUD_SOLR_XML : solrXml;
+    this.trackJettyMetrics = trackJettyMetrics;
+    this.externalZkServer = true; // No ZkTestServer in quorum mode
```
epugh, line 384: something about this being `true` seems odd...   Should it be false since there isn't a externalZkServer?

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
```diff
@@ -342,6 +344,185 @@ public MiniSolrCloudCluster(
...
+    }
+    this.zkHost = zkHostBuilder.toString(); // Save for later use
+
+    if (log.isInfoEnabled()) {
+      log.info("Reserved ports for {} nodes: {}", numServers, java.util.Arrays.toString(ports));
+      log.info("ZK connection string: {}", this.zkHost);
+    }
+
+    // Set system properties for embedded ZK quorum mode
+    System.setProperty("solr.zookeeper.server.enabled", "true");
```
epugh: I am a bit curious if we need this system property AND `zookeeper_quorum:on`?   I wonder if the existence of `zookeeper_quorum:on` is enough?   if this system property is what turns on zookeeper, I wish maybe it was `solr.mode=clustered` or something to specificy that it's the clustered (cloud) mode...

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
```diff
@@ -342,6 +344,185 @@ public MiniSolrCloudCluster(
...
+    // Initialize ZK paths and security (if provided)
+    try (SolrZkClient zkClient =
+        new SolrZkClient.Builder()
+            .withUrl(this.zkHost)
+            .withTimeout(60000, TimeUnit.MILLISECONDS)
+            .build()) {
+      if (!zkClient.exists("/solr", true)) {
+        zkClient.makePath("/solr", false, true);
+      }
+      if (!zkClient.exists("/solr/initialized", true)) {
```
epugh: is this a concept in solr?  this path?   New to me!

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
```diff
@@ -1174,16 +1379,33 @@ public MiniSolrCloudCluster build() throws Exception {
...
-              baseDir,
-              solrXml,
-              jettyConfig,
-              null,
-              securityJson,
-              trackJettyMetrics,
-              formatZkServer);
+      MiniSolrCloudCluster cluster;
+
+      if (useEmbeddedZkQuorum) {
```
epugh, line 1557: this big if statement seems challenging...   Could a builder pattern make it easier?   Thir are a lot of params and overlap with subtle differences ;-).

solr/test-framework/src/java/org/apache/solr/cloud/SolrCloudWithEmbeddedZkQuorumTestCase.java
```diff
@@ -0,0 +1,76 @@
...
+ * limitations under the License.
+ */
+package org.apache.solr.cloud;
+
+import java.lang.invoke.MethodHandles;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Base class for SolrCloud tests that use embedded ZooKeeper running in quorum mode
```
epugh: I really like the javadocs!   as someone who doesn't live this java code everyday, they are helpful.

solr/core/src/test/org/apache/solr/cloud/TestEmbeddedZkQuorum.java
```diff
@@ -0,0 +1,152 @@
...
+    // Configure cluster with 3 nodes, each running embedded ZK
+    cluster =
+        configureClusterWithEmbeddedZkQuorum(NUM_NODES).addConfig("conf1", configPath).build();
+
+    log.info("Cluster configured with {} nodes", NUM_NODES);
+  }
+
+  @Test
+  public void testBasicQuorumFunctionality() throws Exception {
+    log.info("Starting testBasicQuorumFunctionality");
```
epugh: I wonder if claude tools could go back and revamp old unit tests to be clearer?

solr/core/src/test/org/apache/solr/cloud/TestEmbeddedZkQuorum.java
epugh, line 1: This test is verifying the happy path!  I think that we need a test that demonstrates that when you LOSE a node from a three node cluster, we still function, just degraded.   I don't know if we have a test that verifies this behavior elsewhere (a stress test??) that could be reused with embedded zk?

epugh (commented):
Three committers on the same PR!  Woot!

solr/core/src/java/org/apache/solr/core/ZkContainer.java
```diff
@@ -75,32 +81,40 @@ public class ZkContainer {
...
+    }
 
-    if (zkRun && config == null)
+    if (zkServerEnabled && config == null) {
       throw new SolrException(
           SolrException.ErrorCode.SERVER_ERROR,
           "Cannot start Solr in cloud mode - no cloud config provided");
+    }
+
+    if (config == null) {
```
epugh, line 102: How does this even get hit?  Sometimes I get confused by all the null checks!  If you are running in standalone, wouldn't this class never be run???!

janhoy: Who knows? Tons of tests failed due to NPE, and the existing code on main has this check, which was removed in this PR...

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
janhoy: Yea, what is even an "external" zk in the context of our test framework?

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
janhoy: Sounds like node role should be enough, it should turn things on, perhaps set that property automagically?

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
janhoy: Could be some hallucination going on

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
janhoy: Hmm, this code **is** inside the builder :) So the builder chose to instantiate the MSCC using one of two constructors.

solr/test-framework/src/java/org/apache/solr/cloud/SolrCloudWithEmbeddedZkQuorumTestCase.java
janhoy: True.

Regarding this base class, we might not need it, as one can extend `SolrCloudTestCase` with a configure statement like this

```java
    cluster =
        configureCluster(NUM_NODES)
            .addConfig("conf1", configPath)
            .withEmbeddedZkQuorum()
            .build();
```

So I vote for removing it...

solr/core/src/test/org/apache/solr/cloud/TestEmbeddedZkQuorum.java
janhoy: There used to be such a test in the first phase, but it kept failing, so it was removed 😂🤣

Good point though. With this foundation it should be possible to keep adding more exceptional tests. Hey, perahps even randomize running with and without qorum. But currently we requre >=3 nodes. We could evolve the quorum to use 1 zk node if <3 and use 3 zk nodes for >=3. Then if you kill a solr node, it should all work even if that node also has zk.

But I think there were some issues with stopping a `JettySolrRunner` in quorum mode and then restarting it...

solr/test-framework/src/java/org/apache/solr/cloud/SolrCloudWithEmbeddedZkQuorumTestCase.java
epugh: agreed.

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
epugh: If you dont have the node role, then you need it right?   OR!!!?????    here is an idea...    What if you are starting a single solr node and want zk...   The `bin/solr start` of SOlr 10, and was previoulsly `bin/solr start -c` in 9x...   What if, instead, if you do NOT supply a `-Dsolr.node.roles=data:on,overseer:allowed,zookeeper_quorum:on`  or a `-z` property then we do that FOR you in the start up script.   I believe that we ALREADY have logic that if you don't provide a `solr.node.roles` then you get `data:on,overseer:allowed` by default.   See https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/core/NodeRoles.java#L32

Would that simplify things both internally and from a user perspective?

You either:

Start Solr old school with `bin/solr start --user-managed`

Or you start single node with embedded zk with `bin/solr start`

Or you start three nodes with embedded zk quorum with `bin/solr start -z blahblah:blahblah:blahblah -Dsolr.node.roles=data:on,overseer:allowed,zookeeper_quorum:on`

Or you start N nodes and connect to an external ZK with `bin/solr start -z external_zk:2181`.

epugh:
I asked Claude to summarize the last seven commits and he said:

_Recent work on the `spike-zk-quorum` branch has focused on refining Solr's embedded ZooKeeper quorum mode implementation through targeted code improvements. The codebase has been streamlined by removing unnecessary variables and conditional logic in the `SolrZkServer` class, particularly eliminating redundant checks around the `zkServerEnabled` flag and renaming methods like `startZKSE` to more descriptive alternatives. Initialization has been optimized by removing the `/solr/initialized` ZooKeeper node, which was only needed for multi-threading protection that's unnecessary in the current implementation. The ZooKeeper initialization logic in `ZkContainer` has been restructured with clearer nesting and better separation between standalone and quorum modes, making the codebase more maintainable. These changes collectively enhance the embedded ZooKeeper quorum functionality while simplifying the code structure, resulting in a more robust and easier-to-understand implementation that properly manages ZooKeeper server resources throughout Solr's lifecycle._

I think my summary is:  **I discovered that we have TWO ways of starting ZK, not including our independent `ZkTestServer` that runs a `ZkTestServer.ZKServerMain` during tests.**   

We now have:

* For testing only, we have `org.apache.solr.cloud.ZkTestServer.ZKServerMain` that wraps `org.apache.zookeeper.server.ZooKeeperServer`.
* `org.apache.solr.cloud.SolrZkServer` that wraps either a `org.apache.zookeeper.server.ZooKeeperServer` OR if there is more then one, then does a `org.apache.zookeeper.server.quorum.QuorumPeerMain`
* `org.apache.solr.core.ZkContainer` that starts up `org.apache.zookeeper.server.embedded.ZooKeeperServerEmbedded`

I am now wondering could we convert SolrZkServer to using a `ZooKeeperServerEmbedded` instead of a `ZooKeeperServer`?

Secondly, now seeing the use of `QuorumPeerMain`, which currently kicks in if there are multiple ZooKeeper servers defined (presumably only in an external cluster set up), I wonder if we need to rename roles: `zookeeper_quorum` to `zookeeper`.   If you have `zookeeper` then you fire up `ZooKeeperServerEmbedded`.   And if not, then you default to `QuorumPeerMain`.   

Thirdly, why is `ZkContainer` in `o.a.s.core` and not `o.a.s.cloud`?   

I did take a stab at the `SolrZkServer` using the `ZooKeeperServerEmbedded`, updating the start/stop methods, but failed ;-(.   Going to need to put this down for a bit.

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
janhoy: I wonder if `-z` should be reserved for connecting to an external ZK quorum, and we invent a new option for telling SolrCloud which solr nodes will run the zk node role? That would make it clearer to read, and we make ZK an implementation detail by not refering to it as ZK, but as `quorum-node`, e.g. 

```bash
bin/solr start --quorum-nodes=solr-node-1:8983,solr-node-2:8983,solr-node-3:8983
```

That would mean we want Solr to manage our ZK, and translate into the correct `solr.zk.host` setting for each host, no matter if the host you are starting is one of the quorum nodes or not. And if the node you are starting is a quorum node, it will detect that it's own host name matches a quorum host, and add `-Dsolr.node.roles...` during startup, and the ZK bootstrap code will configure its internal ZK config using the `--quorum` / `-Dsolr.quorum.nodes` values.

janhoy:
> I did take a stab at the `SolrZkServer` using the `ZooKeeperServerEmbedded`, updating the start/stop methods, but failed ;-(. Going to need to put this down for a bit.

Thanks for chipping away on this. At some point we'll have something that can be released in 10.1 or 10.2 as an experimental feature for folks to try out, without dynamic reconfig. Then we can tackle the dynamic stuff...

solr/core/src/java/org/apache/solr/core/ZkContainer.java
janhoy: This was an existing timeout variable from 2013 so to keep the changes minimal and easy to review I stuck with it.

epugh:
Thanks for taking a look at this @janhoy I think I've stalled a bit on this...  It feels "so close' yet not sure how to get it over the hump.

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
janhoy: It's not hallucination, but a way for the test to make sure that `urlScheme` and `security.json` are initialized properly in the cluster on first startup. As we only need one node to do this job, a dummy znode is written to zookeeper to say we are initialized.

This is test-code only so not critical what znodes we use?

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
janhoy: Eric, what do you think about the `--quorum-hosts` / `SOLR_QUORUM_HOSTS` / `solr.quorum.hosts` setting? It will replace need for `-c` and `-z` and explicit node-role setting. When inspecting the "ps" command line of a solr node, you'll then clearly see that this cluster is started with the three given quorum nodes, and you immediately know it is a self-managed zk setup, as it lacks the explicit `-z` for an external cluster.

janhoy: I'm thinking ahead to solr-operator as well. When we have self managed zk, the operator should probably spin up two independent `StatefulSet`s, one "solr-quorum" with three low speced PODs with `zookeeper_quorum:on,data:off,overseer:preferred`. While the "solr-data" set of nodes will be today's set of data-nodes, automatically provisioned with a `ZK_HOST` pointing to the three quorum nodes. In k8s the host / service names will be stable.

janhoy:
> Thanks for taking a look at this @janhoy I think I've stalled a bit on this... It feels "so close' yet not sure how to get it over the hump.

Yep, I got it up to speed with main, hardened the qourum test, and try to add test for stopping a quorum node and resume operation.

I think we need to focus on the startup modes / params to have a clear story for the three modes as you mentioned
* `--user-managed`
* `-c` (implicit) with one embedded ZK
* `-z external:2181` for existing external zk cluster
* Managed embedded quorum

For the managed embedded quorum, I believe we have proved that setting the node role explicitly and passing `-z` to all nodes pointing to those quorum nodes will do the trick. That is probably good enough for a first release?

## Phase 1:
Ship a experimental version with a working zk node-role, that starts a quorum that is somewhat stable.

## Phase 2:
Make it easier to setup for a small test cluster with some convenience flags (see above). integrate into solr operator.
Enable ZK ACL stuff by default.

## Phase 3:
Good support for SSL between solr and embedded zk quorum. Either self-provided certs, or self-signed certs during startup. Note, we can have non-ssl Solr API while still having SSL between solr and zookeeper.

## Phase 4:
Support dynamic reconfiguration. I don't know how important that really is. In k8s service-names are stable and you don't need to reconfigure zk really, except when changing the size of your quorum. But in self managed infrastructure, you may want to retire some servers and move zk over to another set of nodes by reconfiguring in two phases.

solr/core/src/test/org/apache/solr/cloud/TestEmbeddedZkQuorum.java
janhoy: @epugh See last commit which adds two resilience tests to `TestEmbeddedZkQuorum`. 
* `testQuorumResilienceWithNodeFailure` tests that we are operational when losing 1/3 nodes
* `testQuorumLossAndRecovery` tests that we cannot create a collection when losing 2/3 nodes, and that it recovers back to quorum again
 
I have not beasted them but pretty happy to see them pass in the first place. And as you probably guessed I did not write those tests, but Claude did and he needed some iterations, so there may absolutely be rough edges.

epugh:
I wonder if the -c mode and the `Managed embedded quorum` are just the same?   Maybe it's "the first three nodes that get fired up are all zk nodes".

janhoy:
> I wonder if the -c mode and the `Managed embedded quorum` are just the same? Maybe it's "the first three nodes that get fired up are all zk nodes".

I don't like that kind of magic. I hope we can introduce this as an additional mode, then once it proves itself and is stable, we can make this mode the default when just starting a single node. But then we're at phase 4, where we have some degree of dynamic configuration and support for growing a quorum. I don't think that is important in the first few releases.

I come to value stability and some degree of static non-mutable config. So it is not a crisis if you need to specify the `ZK_HOST` with all your quorum nodes when starting a solr cluster. It gives you clarify and stability. You don't get the cool dynamic stuff but in large clusters you setup quorum nodes once, either 3 or 5 and then don't touch them for years. All your 50 solr data nodes will have that exact same `ZK_HOST` string. And if you change it, well, you'll have a fun rolling restart of the entire cluster.

epugh:
> > I wonder if the -c mode and the `Managed embedded quorum` are just the same? Maybe it's "the first three nodes that get fired up are all zk nodes".
> 
> I don't like that kind of magic. I hope we can introduce this as an additional mode, then once it proves itself and is stable, we can make this mode the default when just starting a single node. But then we're at phase 4, where we have some degree of dynamic configuration and support for growing a quorum. I don't think that is important in the first few releases.
> 
> I come to value stability and some degree of static non-mutable config. So it is not a crisis if you need to specify the `ZK_HOST` with all your quorum nodes when starting a solr cluster. It gives you clarify and stability. You don't get the cool dynamic stuff but in large clusters you setup quorum nodes once, either 3 or 5 and then don't touch them for years. All your 50 solr data nodes will have that exact same `ZK_HOST` string. And if you change it, well, you'll have a fun rolling restart of the entire cluster.

that makes sense...   I suspect you have a lot more experience in running clusters at scale than I do!

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
epugh: Can I point you to https://github.com/apache/solr/pull/2783 where I added a page to the Ref Guide on "Deployment Topologies".  I would LOVE your input, becasue I sort of guess how people would do things in a Kubernetes environment with solr-operator.  I assumed that you would have one big pool, with some less resources nodes for `zookeeper_quorum:on,data:off,overseer:preferred ` and others with more resources for `zookeeper_quorum:off,data:on`.   I think you are suggesting that isn't actually likely, they would be seperate. 

Here is a diagram of some nodes doing both data and zookeeper: 
image

and then what I have for Kubernetes ( but based on my very vague knowledge)

image.

I'd love some PR updates on how from your expereince people actually would scale up!

janhoy: Much simpler to split in two separate statefulsets, and let users configure resources individually, and delegate to k8s to keep them alive. 

I’m keen to get to a solid MVP in this branch, release it and iterate for each minor release based on feedback.

epugh: Makes sense.   Once this gets released, then I'll be ready to merge #2783 and update it to reflect the best practice of "seperate statefulsets"!

solr/core/src/test/org/apache/solr/cloud/TestEmbeddedZkQuorum.java
janhoy: I had to mark the `testQuorumLossAndRecovery` test case as AwaitsFix, as it never got stable. After killing 2 nodes an starting them again, we had data loss in the collection, even if it had three replicas. I did not check if the three replicas were distributed across the three nodes, but I assume that's the default.. But the test killing one node/zk seems to pass.

solr/core/src/java/org/apache/solr/core/ZkContainer.java
```diff
@@ -75,42 +85,119 @@ public class ZkContainer {
...
+        zkServer.parseConfig();
+        zkServer.start();
+
+        // set client from server config if not already set
+        if (zookeeperHost == null) {
+          zookeeperHost = zkServer.getClientString();
+        }
+        // TODO - should this code go in SolrZkServer to augment or replace its current
+        // capabilities?  Doing so
+        //  would definitely keep ZkContainer cleaner...
```
janhoy: @gerlowskija I removed NOCOMMIT in this comment to unblock other precommit tests. Making this review comment to make sure we don't forget about it :)

janhoy:
I created SOLR-18094 as jira for this issue (sub-task of the SIP JIRA). 

Let's focus on getting this PR in shape for releasing the building blocks of the Zk code and the node-role. Target 10.1 as an experimental feature, can document it as such and solicit feedback from users.

I made Precommit run, will spend some time some day to try to do a more thorough review. Let's make review comments for all known rough edges and things lacking.

solr/core/src/test/org/apache/solr/cloud/TestEmbeddedZkQuorum.java
epugh: Would writing these as bats style change up the stability dynamics?  And or help figure out why not working??

janhoy:
Merged in latest main, hoping to pay some attention to this again.
Also kicked off a Copilot review, to have some early feedback.

Are you on board with trying to polish this PR branch in a "Phase 1" stable state which is mergable for 10.1? It would be a fantastic thing to get some alpha testers trying out, and then move on to Phase 2 in 10.2?

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
```diff
@@ -389,6 +663,87 @@ public void waitForNode(JettySolrRunner jetty, int timeoutSeconds)
...
+      if (runningNodes == expectedCount) {
+        log.info("Verified {} live nodes", runningNodes);
+        return;
+      }
+      Thread.sleep(200);
+    }
+    // Final check after timeout
+    long actualCount = jettys.stream().filter(JettySolrRunner::isRunning).count();
+    throw new TimeoutException(
+        "Live node count mismatch: expected " + expectedCount + " but got " + actualCount);
```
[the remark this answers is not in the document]
janhoy: Added a TODO to code..

janhoy:
@gerlowskija FYI - I updated the PR description to match current code state

solr/core/src/java/org/apache/solr/cloud/SolrZkServer.java
```diff
@@ -253,10 +263,10 @@ public static Properties getProperties(Path configPath) throws ConfigException {
...
   // server.0=localhost:1112:1113
   // server.1=localhost:2223:2224
-  public static void injectServers(Properties props, boolean zkRun, String zkHost) {
+  public static void injectServers(Properties props, String zkHost) {
 
     // if clientPort not already set, use zkRun
-    if (zkRun && props.getProperty("clientPort") == null) {
+    if (props.getProperty("clientPort") == null) {
       // int portIdx = zkRun.lastIndexOf(':');
       int portIdx = "".lastIndexOf(':');
```
janhoy: @epugh This line was modified by you and causes the below if() to be dead code. Remember what's going on here?

solr/core/src/java/org/apache/solr/core/ZkContainer.java
```diff
@@ -60,7 +65,12 @@ public class ZkContainer {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   protected ZkController zkController;
+
+  // zkServer (and SolrZkServer) wrap a ZooKeeperServerMain if standalone mode, but in quorum we
+  // just use ZooKeeperServerEmbedded
+  // directly!  Why?  Can we use ZooKeeperServerEmbedded in one node directly instead?
   private SolrZkServer zkServer;
+  private ZooKeeperServerEmbedded zkServerEmbedded;
```
janhoy: Must be volatile

solr/core/src/java/org/apache/solr/core/ZkContainer.java
```diff
@@ -70,42 +80,109 @@ public class ZkContainer {
...
+        // matching, or matching by port alone (which is unique per node in a single-machine setup).
+        final String targetConnStringSection = config.getHost() + ":" + zkPort;
+        if (log.isInfoEnabled()) {
+          log.info(
+              "Trying to match {} against zkHostString {} to determine myid",
+              targetConnStringSection,
+              config.getZkHost());
+        }
+        for (int i = 0; i < zkHosts.length; i++) {
+          final String host = zkHosts[i];
```
janhoy: Risk of ArrayIndexOutOfBounds

solr/packaging/build.gradle
```diff
@@ -305,6 +305,7 @@ task integrationTests(type: BatsTask) {
   environment SOLR2_PORT: solrPort + 1
   environment SOLR3_PORT: solrPort + 2
   environment ZK_PORT: solrPort + 1000
+  environment SOLR_EXPORTER_PORT: solrPort + 100
```
janhoy: Unrelated change

janhoy:
I addressed Copilot's review points, then ran another review with Claude and addressed some of those concerns. 

Added ref-guide docs, marked as "experimental":

Release-notes:
Skjermbilde 2026-05-11 kl  10 29 33

Zookeeper config:
Skjermbilde 2026-05-11 kl  10 21 03

Node roles:
Skjermbilde 2026-05-11 kl  10 22 07

solr/core/src/java/org/apache/solr/core/ZkContainer.java
```diff
@@ -70,42 +81,119 @@ public class ZkContainer {
...
+      log.info("Solr is running in standalone mode");
+      return;
+    }
+
+    final boolean runAsQuorum = config.getZkHost() != null && zkQuorumNode;
 
     String zookeeperHost = config.getZkHost();
+    final var solrHome = cc.getSolrHome();
+    if (zkServerEnabled) {
+      if (!runAsQuorum) {
```
janhoy: It should be enough to specify the node role to start zk, no need for also `-dsolr.zookeeper.server.enabled`.

solr/core/src/java/org/apache/solr/core/ZkContainer.java
```diff
@@ -313,8 +413,37 @@ public void close() {
...
-          zkServer.stop();
+        try {
+          if (zkServer != null) {
+            zkServer.stop();
+          }
+        } finally {
+          if (zkServerEmbedded != null) {
+            try {
+              zkServerEmbedded.close();
+              log.info("Closed embedded ZooKeeper server in quorum mode");
```
janhoy: Close log should come after the wait loop

solr/core/src/java/org/apache/solr/core/ZkContainer.java
```diff
@@ -241,9 +333,17 @@ public SolrMetricsContext getSolrMetricsContext() {
...
   }
 
-  private String stripChroot(String zkRun) {
-    if (zkRun == null || zkRun.trim().isEmpty() || zkRun.lastIndexOf('/') < 0) return zkRun;
-    return zkRun.substring(0, zkRun.lastIndexOf('/'));
+  private void startZooKeeperServerEmbedded(int port, String zkHomeDir) throws Exception {
+    Properties p = new Properties();
+    try (FileReader fr = new FileReader(zkHomeDir + "/zoo.cfg", StandardCharsets.UTF_8)) {
+      p.load(fr);
+    }
```
janhoy: Redundant read of zoo.cfg, the file is already in memory in call site. Refactor this method

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
janhoy: So I modified things so that it is now enough to specify the `zookeeper_quorum` node role, and it will then enable embedded zk without the need for also `solr.zookeeper.server.enabled`.

solr/core/src/java/org/apache/solr/core/ZkContainer.java
janhoy: Fixed. AI assistant refactored relevant code fragments, see the change in this commit https://github.com/apache/solr/pull/2391/changes/47251b9c7cc7904975bb1be303552cbf18fd5fdc?diff=split&w=1 - it looks solid..

solr/solr-ref-guide/modules/deployment-guide/pages/zookeeper-ensemble.adoc
```diff
@@ -546,6 +546,68 @@ set SOLR_OPTS=%SOLR_OPTS% -Djute.maxbuffer=0x200000
...
+
+[source,bash]
+----
+export LH="localhost"
+ZK="-z $LH:21000,$LH:31000,$LH:41000"
+ROLES="-Dsolr.node.roles=data:on,overseer:allowed,zookeeper_quorum:on"
+
+bin/solr start -p 20000 $ZK $ROLES
+bin/solr start -p 30000 $ZK $ROLES
+bin/solr start -p 40000 $ZK $ROLES
```
janhoy: Running three nodes like this causes all three embedded zookeepers to use exact same `server/solr/zoo_home/data` folder, which is not good of course.

janhoy: I fixed it it commit ccf6ec3add1d0ee583d25f6cf518e4030d1f04be - appending `-<port>` to the `zoo_home` folder name.

solr/solr-ref-guide/modules/deployment-guide/pages/zookeeper-ensemble.adoc
```diff
@@ -546,6 +546,68 @@ set SOLR_OPTS=%SOLR_OPTS% -Djute.maxbuffer=0x200000
...
+|`solr_port + 1000`
+|9983
+
+|ZK quorum peer
+|`zk_client_port + 1`
+|9984
+
+|ZK leader election
+|`zk_client_port + 2`
+|9985
```
janhoy: I modified the default ports used for quorum peer and leader election ports. They were something like client-port minus 3000 and minus 4000. Now they are +1 and +2. I feel we spread our port numbers too wide. But then after the change we got another problem that we could not start three test solr nodes on 8983, 8984, 8985, since the first one would occupy port 9983, 9984, 9985 for zookeeper. 

I am willing to revert the +1, +2 to something else, but -3000 and -4000 seemed a bit arbitrary, why not +2000 and +3000 since client-port is +1000?

solr/core/src/java/org/apache/solr/cloud/SolrZkServer.java
epugh: not totally...   And I think it was me hacking around a bit trying to figure out all the various sys props that existed!

solr/solr-ref-guide/modules/deployment-guide/pages/zookeeper-ensemble.adoc
epugh: so, we can't do 8983, 8984,8985?   That is definitly a conceptual change for all of us.  I'm used to zk being 1000 more that my port...

solr/solr-ref-guide/modules/deployment-guide/pages/zookeeper-ensemble.adoc
epugh: Yeah, after commenting elsewhere, I like your +2000 and +3000 idea.

janhoy: I’d prefer if Solr occupied a more narrow port space, say within 100 ports total, so you could start Solr on 8000, 8100, 8200 and each instance would have ports close. Instead of node 1 using 8000, 9000, 10000, 11000, while node 2 using 8100, 9100, 10100, 11100….

solr/core/src/java/org/apache/solr/cloud/SolrZkServer.java
janhoy: Feel like figuring out what is the correct code here?

solr/solr-ref-guide/modules/deployment-guide/pages/zookeeper-ensemble.adoc
janhoy: And come to think of it, each and every of these ports should probably be configurable individually even if they default to some port offset...

solr/core/src/java/org/apache/solr/core/ZkContainer.java
```diff
@@ -192,13 +196,62 @@ public void initZooKeeper(final CoreContainer cc, CloudConfig config) {
...
       }
     } else if (zkServerEnabled) {
-      // Old school ZooKeeperServerMain being used under the covers (bin/solr -c without ZK_HOST).
-      zkServer =
-          SolrZkServer.createAndStart(config.getZkHost(), solrHome, config.getSolrHostPort());
+      try {
+        final Path zkDataHome =
+            Path.of(
+                EnvUtils.getProperty(
+                    "solr.zookeeper.server.datadir", solrHome.resolve("zoo_data").toString()));
```
janhoy: Do we normally keep constants like this inline or do we have somewhere we try to keep all sysprop-names for e.g. zookeeper together in one place?

epugh: There is NOT a strong pattern.     I think it's personal preference.   For me, I always find the text version easier to read than the `SOLR_ZOOKEEPER_SERVER_DETAILED` version ;-)

solr/solr-ref-guide/modules/deployment-guide/pages/zookeeper-ensemble.adoc
epugh: yeah...

solr/solr-ref-guide/modules/deployment-guide/pages/zookeeper-ensemble.adoc
```diff
@@ -589,16 +589,17 @@ The ZK connection string must list all nodes' ZK client ports (`solr_port + 1000
...
 [source,bash]
 ----
 export LH="localhost"
 ZK_HOST="$LH:21000,$LH:31000,$LH:41000"
 ROLES="data:on,overseer:allowed,zookeeper_quorum:on"
 
-bin/solr start -p 20000 -z "$ZK_HOST" -Dsolr.node.roles="$ROLES"
-bin/solr start -p 30000 -z "$ZK_HOST" -Dsolr.node.roles="$ROLES"
-bin/solr start -p 40000 -z "$ZK_HOST" -Dsolr.node.roles="$ROLES"
+bin/solr start -p 20000 -z "$ZK_HOST" -Dsolr.node.roles="$ROLES" -Dsolr.zookeeper.server.datadir=zoo_home_1
```
epugh, line 602: is this really better?   What if we did soemthing like `./blah/blah/zoo_home/localhost:21000` and `./blah/blah/zoo_home/localhost:31000` and ``./blah/blah/zoo_home/localhost:41000`, and generated it.   I don't think the vast majority of folks even want to think about zook_home!

janhoy: The multi solr on single host is mostly for dev/test, so nice to have but not front and center? At least, now the zoo-data folder is configurable if folks want to place it somewhere special. But maybe we should also allow auto unique data folder through property? Also the zoo-config folder (defaulting to SOLR_HOME) needs the same...

solr/core/src/java/org/apache/solr/core/ZkContainer.java
janhoy: Yea, can be extracted to a constant if we start using the same property multiple times..

janhoy:
I just made the offsets for zookeeper server port and quorum and election ports configurable with sysprops. Still defaults to +1000. And defaulting quorum port to +1010 and election port to +1020. This means you can run 10 instances on same host on 8983, 8984, 8985, 8986 without any conflicts. And it is easily configurable to +1, +2, +3

epugh:
> I just made the offsets for zookeeper server port and quorum and election ports configurable with sysprops. Still defaults to +1000. And defaulting quorum port to +1010 and election port to +1020. This means you can run 10 instances on same host on 8983, 8984, 8985, 8986 without any conflicts. And it is easily configurable to +1, +2, +3

I like it!

solr/solr-ref-guide/modules/deployment-guide/pages/zookeeper-ensemble.adoc
epugh: Yeah, I think the whole point of this embedded mode is that people don't need to think about it.   I think if "I want to do it somewhere special" then that leads you more towards the "I will set up my own zk ensemble"...   Just thinking from the perspective of a non Solr expert...  This feature is really going to be a hit among the novices to Solr who don't want to think about Cloud architectures ;-).

PR #2392, opened 2024-04-06
SOLR-17153: Resolve IndexOutOfBoundsException in HttpSolrCall.resolveDocCollection()

aparnasuresh85:
https://issues.apache.org/jira/browse/SOLR-17153

# Description

Resolve an IndexOutOfBoundsException in HttpSolrCall.resolveDocCollection(), at line: 
String collectionName = collectionsList.get(0);

# Solution

Check if collectionsList is empty, in that case return early. 

# Tests

Ensured existing tests pass

# Checklist


dsmiley (commented):
+1 to the change.
needs "tidy" clearly (you don't have a space to the left of the parenthesis)

Also something was eating at it me a bit... the method you added, `resolveDocCollection`, returns a DocCollection but in HttpSolrCall.init you added a call to it that does nothing with the response.  Feels weird.  Instead, shouldn't HttpSolrCall.getCoreByCollection use this method?

dsmiley:
This PR looks weird; looks like the old PR but one commit added.  Should be a new PR.

aparnasuresh85:
> +1 to the change. needs "tidy" clearly (you don't have a space to the left of the parenthesis)
> 
> Also something was eating at it me a bit... the method you added, `resolveDocCollection`, returns a DocCollection but in HttpSolrCall.init you added a call to it that does nothing with the response. Feels weird. Instead, shouldn't HttpSolrCall.getCoreByCollection use this method?

I agree. The reason I kept the call to resolveDocCollection() outside of getCoreByCollection() is because the doc around the latter (if getCoreByCollection returned null ) said "// this collection exists , but this node does not have a replica for that collection". 

Instead, should we throw a SolrException if inside getCoreByCollection, the call to resolveDocCollection() returned null?

aparnasuresh85:
> > +1 to the change. needs "tidy" clearly (you don't have a space to the left of the parenthesis)
> > Also something was eating at it me a bit... the method you added, `resolveDocCollection`, returns a DocCollection but in HttpSolrCall.init you added a call to it that does nothing with the response. Feels weird. Instead, shouldn't HttpSolrCall.getCoreByCollection use this method?
> 
> I agree. The reason I kept the call to resolveDocCollection() outside of getCoreByCollection() is because the doc around the latter (if getCoreByCollection returned null ) said "// this collection exists , but this node does not have a replica for that collection".
> 
> Instead, should we throw a SolrException (collection or alias not found) if inside getCoreByCollection, the call to resolveDocCollection() returned null?

Maybe not. I see a number of tests fail if an exception is thrown in case collection is null from resolveDocCollection(). The tests pass otherwise. I will send a new PR.

PR #2394, opened 2024-04-09
Renovatebot config fixes

janhoy:
Ref discussion in https://lists.apache.org/thread/ovwytr47zv2y9xs7fxbgl6kbpfmbqc5v

Try to be explicit with schedule on the packages we want to update less frequently.
Also fix changelog location for some gitbox repos, where renovate was complaining, e.g.

```
 INFO: Unknown platform, skipping changelog fetching. (repository=apache/solr, branch=renovate/org.apache.zookeeper)
       "sourceUrl": "https://gitbox.apache.org/repos/asf/zookeeper.git",
       "hostType": null
```

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -34,7 +34,7 @@
     },
     {
       "description": "Noisy, frequently updated dependencies checked less often",
-      "matchPackageNames": ["software.amazon.awssdk", "com.google.cloud"],
+      "matchPackagePrefixes": ["software.amazon.awssdk", "com.google.cloud"],
       "schedule": ["before 9am on the first day of the month"]
     },
     {
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -30,12 +30,12 @@
     {
       "description": "Test-dependencies are checked less often than the shipped deps",
       "matchDepTypes": ["test"],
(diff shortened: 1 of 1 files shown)


epugh (approved):
LGTM.

epugh:
This looks great.  Though, it doesn't change that we need to be better as a community on merging these SolrBot PR's right?

janhoy:
Currently the bot opens PRs all week. Earlier it did only on Sundays. Should we go back to the Sunday schedule for less churn during the week?

.github/renovate.json
```diff
@@ -46,6 +46,26 @@
       "description": "Lucene dependencies should be skipped",
       "matchPackageNames": ["org.apache.lucene"],
```
janhoy, line 47: Another issue is that this rule does not work. The bot opened a lucene upgrade PR on Feb 27 even if this rule has been in place since June 2023. https://github.com/apache/solr/pull/2309

janhoy: I tried to fix this by using `matchPackagePrefixes` instead of `matchPackageNames`, which matches package name exactly, which would be e.g. `org.apache.lucene:lucene-core`, i.e. including group and artifact.

PR #2395, opened 2024-04-09
SOLR-17192: Add "field-limiting" URP to catch ill-designed schemas

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-17192

# Description

Solr isn't infinitely scalable when it comes to the number of fields in each core/collection. Most deployments start to experience problems any time a core has upwards of a few hundred fields. Usually this doesn't exhibit itself right away. instead waiting until segment-merge or some other time to rear its head.

Despite this being a known limitation Solr doesn't have any (active) way of helping users avoid this, excepting one or two references buried in the Solr ref-gudie. 

# Solution

This commit adds a new UpdateRequestProcessor, `NumFieldsLimitingUpdateRequestProcessor`, that flags potentially-dangerous schema design for users by failing all updates once the relevant core has exceeded a configurable limit.  The proposed URP has two configuration properties:

- `maxFields` - (required) a positive integer value representing the maximum number of fields a core should be allowed to have.
- `warnOnly` - (optional, defaults to 'false' if not specified) a boolean flag indicating whether Solr should "really" fail requests once the threshold has been hit, or should merely log a warning instead.

**Note:** In order to be as "lightweight" as possible, this URP only updates its "view" of the number of fields in the core on each commit.  Of course, this means that it operates on somewhat of a delay, and isn't a guarantee that users won't exceed the threshold at all.  Rather, the real goal of the URP is less about serving as a hard limit, and more about forcing users to reconsider their schema design once the number of fields starts trending into "questionable" territory. 

# Tests

Unit tests in `NumFieldsLimitingUpdateRequestProcessorFactoryTest`.  Still needs some higher level testing prior to merge.

# Checklist


gerlowskija:
Still TBD - additional tests, ref-guide docs.

gerlowskija:
In a sample solrconfig.xml, configuration for the proposed URP would look like:

```
  <updateProcessor class="solr.NumFieldLimitingUpdateRequestProcessorFactory" name="max-fields">
    <int name="maxFields">123</int>
  </updateProcessor>
  ...
  <updateRequestProcessorChain name="add-unknown-fields-to-the-schema"
           processor="uuid,remove-blank,field-name-mutating,parse-boolean,parse-long,parse-double,parse-date,add-schema-fields,max-fields">
```

gerlowskija:
This should all be ready to go: I've added unit and integration tests, `./gradlew test` passes, there's ref-guide coverage and decent Javadocs, etc.  I'll aim to merge EOW, absent any feedback folks might have here.

One big question I have is: should we add this into the "_default" configset's default URP-chain?  At least for 10.x?

I'd argue we should.  Safety guardrails like this one aren't any good unless users enable them, and having it be enabled by default is the best way to do that.  We might get some complaints from folks who run into this during an inplace 9->10 upgrade, but:

1. We tell users to carefully reconsider their configsets when doing upgrades, so this shouldn't be a surprise.
2. We can highlight it clearly in the upgrade notes
3. Users who hit this on an in-place upgrade can always bump the limit, toggle the URP into 'warnOnly' mode, or disable the URP altogether if they really want.

I'd love a sanity-check/+1 from someone else though before I make this enabled by default on `main`/10.x, so please chime in!

solr/core/src/java/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessor.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  public void processAdd(AddUpdateCommand cmd) throws IOException {
+    if (!isCloudMode() || /* implicit isCloudMode==true */ isLeaderThatOwnsTheDoc(cmd)) {
+      if (coreExceedsFieldLimit()) {
+        throwExceptionOrLog(cmd.getPrintableId());
+      } else {
+        if (log.isDebugEnabled()) {
+          log.debug(
+              "Allowing document {}, since current core is under the max-field limit (numFields={}, maxThreshold={})",
```
epugh, line 62: max-field or maxField?

solr/core/src/java/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessorFactory.java
```diff
@@ -0,0 +1,110 @@
...
+
+    if (args.indexOf(MAXIMUM_FIELDS_PARAM, 0) < 0) {
+      throw new IllegalArgumentException(
+          "The "
+              + MAXIMUM_FIELDS_PARAM
+              + " parameter is required for "
+              + getClass().getName()
+              + ", but no value was provided.");
+    }
+    final Object rawMaxFields = args.get(MAXIMUM_FIELDS_PARAM);
```
epugh, line 85: somdeay we will have a utility method that does this kind of thing and is used everywhere!

solr/core/src/java/org/apache/solr/update/processor/NumFieldsMonitor.java
```diff
@@ -0,0 +1,57 @@
...
+    this.numFields = -1;
+  }
+
+  @Override
+  public void postCommit() {
+    RefCounted<SolrIndexSearcher> indexSearcher = null;
+    try {
+      indexSearcher = getCore().getSearcher();
+      // Get the number of fields directly from the IndexReader instead of the Schema object to also
+      // include the
```
epugh, line 39: formatting?

epugh: are there any other interesting places in Solr to use a NumFieldsMonitor?

solr/core/src/test/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessorIntegrationTest.java
```diff
@@ -0,0 +1,115 @@
...
+import org.apache.solr.client.solrj.request.CollectionAdminRequest;
+import org.apache.solr.cloud.SolrCloudTestCase;
+import org.apache.solr.common.SolrInputDocument;
+import org.hamcrest.Matchers;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class NumFieldLimitingUpdateRequestProcessorIntegrationTest extends SolrCloudTestCase {
+
+  private static long ADMIN_OPERATION_TIMEOUT = 15 * 1000;
```
epugh, line 32: use TimeUnit instead?  I'm guessing this is meant to be 15 seconds?

solr/server/solr/configsets/_default/conf/solrconfig.xml
```diff
@@ -939,7 +939,7 @@
   </updateProcessor>
 
   <!-- The update.autoCreateFields property can be turned to false to disable schemaless mode -->
-  <updateRequestProcessorChain name="add-unknown-fields-to-the-schema" default="${update.autoCreateFields:true}"
```
epugh, line 942: intentional?

solr/solrj/src/java/org/apache/solr/common/util/StrUtils.java
```diff
@@ -388,4 +388,19 @@ public static String stringFromReader(Reader inReader) throws IOException {
       return stringWriter.toString();
     }
   }
+
```
epugh, line 391: Really?   Sigh...   we don't have this already!???

epugh (commented):
Some questions/comments...

dsmiley (commented):
Seeing the non-trivial checks determine the doc will be indexed in this core looks painful to me.  If we have an URP, document that it needs to be placed to the right of DistributedUpdateProcessor and/or we enhance Solr to make sure that certain URPs cannot be misplaced relative to that pivotal URP.

I like the SolrEventListener for commits part.  I feel like this should be its own component someone should be able to configure if they want, like for monitoring purposes but don't necessarily want to block adds.  To block all indexing, it could set `SolrCore.readOnly = true` and be done with it; no URP.

I see more test configs here; these are a heavy burden for our project to maintain.  Please try to write the test to generate the configuration so that the test and its config is self-contained.  You can see an example of this in `org.apache.solr.search.TestThinCache#setupSolrHome`

dsmiley:
Sorry if I rained on your parade, attempting to get this in by EOW.

I like the idea of having checks like this by default but I'd like Solr to have such built-in plugins simply be registered automatically unless you set some env/sys-prop to disable, like "solr.core.eventListener.fieldLimit.enabled".  We already register a ton of query parsers even though nobody asked for them :-). I think that should be the model for built-in plugins so that we can continue slowly to reduce what's in a solrconfig.xml yet also allow someone to disable.

solr/core/src/java/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessor.java
gerlowskija: Done 👍

solr/core/src/java/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessorFactory.java
gerlowskija: Agreed - it's really unfortunate.  I looked at a number of URP's and NamedListInitializedPlugin's, but couldn't find anything I liked.

Maybe we could file a ticket for that separately; it'd be a really nice cleanup for all our NLIP's to have a set of utilities:

```
int getRequiredNamedListValueAsInteger(NamedList nl, String paramName);
Integer getNamedListValueAsInteger(NamedList nl, String paramName);
boolean getRequiredNamedListValueAsBoolean(NamedList nl, String paramName);
Boolean getNamedListValueAsBoolean(NamedList nl, String paramName);
```

Of course, even better would be to come up with a way for some XML parsing library to serialize this all into a strongly typed POJO for us, so we don't have to write any map/NL inspection code at all.  We already do essentially the same thing for v2 requests - why not lean on Jackson or something similar to do that when reading XML config?

But obviously either of those approaches is pretty beyond the scope of this PR 😦

solr/core/src/java/org/apache/solr/update/processor/NumFieldsMonitor.java
gerlowskija: > formatting?

Done 👍 

> are there any other interesting places in Solr to use a NumFieldsMonitor?

Good question.  It does seem like a kindof neat thing to have around, but I can't think of any other use-cases off the top of my head...

solr/core/src/test-files/solr/configsets/cloud-minimal-field-limiting/conf/solrconfig.xml
```diff
@@ -0,0 +1,73 @@
...
+  <updateRequestProcessorChain name="add-unknown-fields-to-the-schema" default="true"
+           processor="max-fields">
+    <processor class="solr.LogUpdateProcessorFactory"/>
+    <processor class="solr.DistributedUpdateProcessorFactory"/>
+    <processor class="solr.RunUpdateProcessorFactory"/>
+  </updateRequestProcessorChain>
+
+  <indexConfig>
+    <mergeScheduler class="${solr.mscheduler:org.apache.lucene.index.ConcurrentMergeScheduler}"/>
+:  </indexConfig>
```
gerlowskija, line 71: WTH?!

I frequently insert `:` or `:wq` into files by accident when I make small edits using vim, but shouldn't this make the file invalid XML?  Maybe the parser is treating the ':' as the `<indexConfig>` tag's CDATA?  Is it valid for an XML tag to have both children and CDATA?

EDIT: Nvm, [I guess it is](https://stackoverflow.com/questions/12130228/can-a-xml-element-contain-text-and-child-elements-at-the-same-time)

solr/core/src/test/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessorFactoryTest.java
```diff
@@ -0,0 +1,84 @@
...
+    assertThat(thrown.getMessage(), Matchers.containsString("no value was provided"));
+  }
+
+  @Test
+  public void testReportsErrorIfMaximumFieldsIsInvalid() {
+    final var initArgs = new NamedList<>();
+    initArgs.add("maxFields", "nonIntegerValue");
+    IllegalArgumentException thrown =
+        expectThrows(
+            IllegalArgumentException.class,
```
gerlowskija, line 53: [0] I really hate the formatting that 'tidy' applies here.  Why make this 6 lines?  What's wrong with:

```
var thrown = expectThrows(IllegalArgumentException.class, () -> {
    factory.init(initArgs);
});
```

solr/core/src/test/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessorIntegrationTest.java
gerlowskija: Even better, this was an unused constant.  Removed.

solr/server/solr/configsets/_default/conf/solrconfig.xml
gerlowskija: No, fixed.

solr/solrj/src/java/org/apache/solr/common/util/StrUtils.java
gerlowskija: We've long used Apache common-lang3's `StringUtils`, which offers this and all kinds of really nice String-utilities.

But recently we decided to put any use of `StringUtils` (or maybe commons-lang altogether?) in our forbidden-api checking, and so we've been adding to the Solr `StrUtils` class slowly since then.  I'm sure there's a good reason for the forbidden-api, but not knowing the "why" I'm always irked when I hit this and have to write yet-another function that we could be getting "for free" from any number of libraries.

And then precommit fails because we're doing a reference-equality check, so we have to put in a Suppression for that. Just so so so much fun...

old-man-cloud

gerlowskija:
> Sorry if I rained on your parade, attempting to get this in by EOW.

Oh no worries; there's no particular rush.  That was just how long I was planning to wait if no one commented in the interim.  I'm glad you both chimed in! 

Though tbh I can't quite tell where you are on the PR approach overall:

1. Are you '-1' this being an URP at all?
2. Are you OK with this being an URP, with the minor tweaks you mentioned (e.g. documenting that it goes _after_ DUP, etc.)?
3. Are you only OK with this being an URP, but only if we make some improvements to the URP framework (e.g. to ensure certain URPs always follow DUP)
4. ...something else?

Please let me know!

One last reply inline:

> To block all indexing, it could set SolrCore.readOnly = true and be done with it

IMO `SolrCore.readOnly` is too blunt a tool for the "field limiting" this PR attempts.  First because `readOnly` would block deletions and operations that we'd want to allow.  And second because a user blocked by 'readOnly' would have their `/update` requests fail with a message that's pretty inscrutable or at least unrelated to the root cause.

That's what I like about the URP approach - it gives us more granularity into the different types of "/update" operations and allows us to get a nice (or at least, nicer) error message back to users.

dsmiley:
I don't like the complexity in this URP relating to tolerance of where the URP is placed in the chain; I'd feel better if the URP were simplified from that concern and we expect the user to place it at an appropriate spot.   We don't have complexity like that elsewhere and/or I argue it's not the right approach.

I sympathize with why an URP feels right.  Okay.  On each `addDoc`, don't we just need to do the check on the current SolrIndexSearcher but remember who the searcher was so that the next `addDoc` can see it's the same searcher and if so don't do the check?  It'd need to cache the searcher reference for the instance equality; use a `WeakReference` so we allow the searcher to close & GC.  If we do this, we don't need a commit event listener, thus don't need an additional component / confusing interaction complexity.  Don't get the SolrIndexSearcher from the request (we don't want a per-request cache instance), get it from the SolrCore so we see a possibly evolving SolrIndexSearcher.

gerlowskija:
> I don't like the complexity in this URP relating to tolerance of where the URP is placed in the chain; I'd feel better if the URP were simplified from that concern and we expect the user to place it at an appropriate spot

Yeah I understand.  I still feel a little reluctant I guess, but I'm likely being paranoid.  It's harder to feel good about the "win" in making things safer for users if the change opens up another trappy state when misconfigured.  But I'll take out the leader-check for now.  We can always re-evaluate later if we see people getting bitten by this, and there are other ways to mitigate the risk of misconfig (e.g. putting into the "_default" config for 10.0 so users needn't tweak things).

gerlowskija:
One error case that removing the leader logic will introduce is that due to segment merging, etc. not all replicas will necessarily agree on the number of fields.  So there's a chance that when NRT leaders forward the doc along to replicas, that it might get a mix of success/failure responses back.

Just noting this here as a reminder to explain this case in the little blurb the URP gets in the ref guide.

solr/core/src/java/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessor.java
```diff
@@ -0,0 +1,140 @@
...
+      boolean warnOnly) {
+    super(next);
+    this.req = req;
+    this.fieldThreshold = fieldThreshold;
+    this.currentNumFields = currentNumFields;
+    this.warnOnly = warnOnly;
+  }
+
+  public void processAdd(AddUpdateCommand cmd) throws IOException {
+    if (!isCloudMode() || /* implicit isCloudMode==true */ isLeaderThatOwnsTheDoc(cmd)) {
```
dsmiley, line 56: isCloudMode -- let's not; the user can configure or not as they wish.

"isLeaderThatOwnsTheDoc" can also be removed.  Instead in the URP Factory, if we are not a leader, then skip this URP.  The "owns the doc concept" I think we are agreeing we don't need in this PR.

solr/core/src/java/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessorFactory.java
dsmiley: If SolrParams has some nice conveniences like this; shouldn't NamedList?  I'm guilty of converting an NL to a SolrParams purely for those conveniences; others have too.

gerlowskija:
Alright, I've removed a lot of the leadership and shard checking logic based on review feedback.  I think this should be ready to go, unless you have remaining concerns @dsmiley ?

I've also added the new URP to the URP-chain defined in the "_default" configset (`/solr/server/solr/configsets/_default`).  My intent here is that this part of the PR would be a "main" only change.  9.x would either remain without the URP in the _default configset, or have it enabled but in "warnOnly" mode (my personal preference).  I used a relatively high field limit, 1000, but we can raise it if folks think 1000 is too low?

solr/core/src/java/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+import java.io.IOException;
+import java.lang.invoke.MethodHandles;
+import java.util.Locale;
+import org.apache.solr.common.SolrException;
+import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.update.AddUpdateCommand;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class NumFieldLimitingUpdateRequestProcessor extends UpdateRequestProcessor {
```
dsmiley, line 28: javadoc needed, even if one line linking to a factory for further info.
Or don't make public :-)

solr/core/src/java/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessorFactory.java
```diff
@@ -0,0 +1,113 @@
...
+ * <ul>
+ *   <li><code>maxFields</code> - (required) The maximum number of fields before update requests
+ *       should be aborted. Once this limit has been exceeded, additional update requests will fail
+ *       until fields have been removed or the "maxFields" is increased.
+ *   <li><code>warnOnly</code> - (optional) If <code>true</code> then the URP logs verbose warnings
+ *       about the limit being exceeded but doesn't abort update requests. Defaults to <code>false
+ *       </code> if not specified
+ * </ul>
+ *
+ * @since 9.6.0
```
dsmiley, line 54: 9.7.0

dsmiley: You know what'd be cool?  If we had a magic string like VERSION_ON_RELEASE and the release wizard automated a find-replace on the relevant branches.  Hey @janhoy -- just sharing my wild ideas :-)

solr/core/src/java/org/apache/solr/update/processor/NumFieldsMonitor.java
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+package org.apache.solr.update.processor;
+
+import org.apache.solr.core.AbstractSolrEventListener;
+import org.apache.solr.core.SolrCore;
+import org.apache.solr.search.SolrIndexSearcher;
+import org.apache.solr.util.RefCounted;
+
+public class NumFieldsMonitor extends AbstractSolrEventListener {
```
dsmiley, line 24: public?  make inter class of the URP / factory?

solr/core/src/test/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessorIntegrationTest.java
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.solr.client.solrj.request.CollectionAdminRequest;
+import org.apache.solr.cloud.SolrCloudTestCase;
+import org.apache.solr.common.SolrInputDocument;
+import org.hamcrest.Matchers;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class NumFieldLimitingUpdateRequestProcessorIntegrationTest extends SolrCloudTestCase {
+
+  private static String SINGLE_SHARD_COLL_NAME = "singleShardColl";
```
dsmiley, line 32: this is the only collection you are using in your test, yet it has this curious name.  Why not just use "collection1" and call it COLL_NAME?

solr/core/src/java/org/apache/solr/update/processor/NumFieldsMonitor.java
dsmiley: I suppose you didn't like my proposal that would eliminate the need for this event listener; it'd keep the logic within NumFieldLimitingURP to grab the latest searcher, if unchanged, and basically just lookup the fieldInfos size at that point.  Maybe the need for a WeakReference made it be sort of complex, albeit this new component for an event listener perhaps could be said is also complex.

solr/solr-ref-guide/modules/configuration-guide/pages/update-request-processors.adoc
```diff
@@ -337,6 +337,12 @@ Documents processed prior to the offender are indexed by Solr; documents followi
 +
 Alternatively, the processor offers a "permissive" mode (`permissiveMode=true`) which skips the offending document and logs a warning, but doesn't abort the remainder of the batch or return an error to users.
 
+{solr-javadocs}/core/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessorFactory.html[NumFieldLimitingUpdateRequestProcessorFactory]:: Fails update requests once a core has exceeded a configurable "maximum" number of fields.
++
+Solr performance can degrade and even become unstable if cores accumulate too many (e.g. more than 500) fields.  The "NumFieldLimiting" URP is offered as a safeguard that helps users notice potentially-dangerous schema design and/or mis-use of dynamic fields, before these performance and stability problems would manifest.
+Note that the field count an index reports can be influenced by deleted (but not yet purged) documents, and may vary from replica to replica.
+In order to avoid these sort of discrepancies between replicas, use of this URP should almost always precede DistributedUpdateProcessor in when running in SolrCloud mode.
```
dsmiley, line 344: This advice surprised me.  I guess why not, as only the first node receiving the batch will evaluate this logic.  It may/may-not be the leader but whatever.

solr/core/src/test-files/solr/configsets/cloud-minimal-field-limiting/conf/solrconfig.xml
```diff
@@ -0,0 +1,73 @@
...
+    <lst name="defaults">
+      <str name="echoParams">explicit</str>
+      <str name="indent">true</str>
+      <str name="df">text</str>
+    </lst>
+  </requestHandler>
+
+  <updateProcessor class="solr.NumFieldLimitingUpdateRequestProcessorFactory" name="max-fields">
+    <int name="maxFields">${solr.test.maxFields:1234}</int>
+  </updateProcessor>
```
dsmiley, line 49: why declare this separately instead of in the one chain that uses it?  It's not obvious, when looking at the chain, that there is this additional processor.

solr/core/src/java/org/apache/solr/update/processor/NumFieldsMonitor.java
```diff
@@ -0,0 +1,56 @@
...
+  public NumFieldsMonitor(SolrCore core) {
+    super(core);
+    this.numFields = -1;
+  }
+
+  @Override
+  public void postCommit() {
+    RefCounted<SolrIndexSearcher> indexSearcher = null;
+    try {
+      indexSearcher = getCore().getSearcher();
```
dsmiley, line 37: see SolrCore.withSearcher to make this code less involved.  I added that long ago and getSearcher's javadocs recommend using it.

solr/core/src/java/org/apache/solr/update/processor/NumFieldsMonitor.java
```diff
@@ -0,0 +1,56 @@
...
+
+  private int numFields;
+
+  public NumFieldsMonitor(SolrCore core) {
+    super(core);
+    this.numFields = -1;
+  }
+
+  @Override
+  public void postCommit() {
```
dsmiley, line 34: do we need this at all?  Won't the callback for newSearcher address this?

solr/core/src/java/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessorFactory.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+
+  @Override
+  public UpdateRequestProcessor getInstance(
+      SolrQueryRequest req, SolrQueryResponse rsp, UpdateRequestProcessor next) {
+
+    // TODO Should we skip to the next URP if running in SolrCloud and *not* a leader?
+    return new NumFieldLimitingUpdateRequestProcessor(
+        req, next, maximumFields, numFieldsMonitor.getCurrentNumFields(), warnOnly);
```
dsmiley, line 103: Woah, okay, I wasn't expecting this.  The `numFieldsMonitor.getCurrentNumFields` is only read on load of the URP; it's not per-doc.  If that's adequate (I suppose; this is just a soft limit) then this can be much simpler -- just go read the SolrIndexSearcher right now to get the field count!  No NumFieldsMonitor needed.  Furthermore the URP could be renamed to simply "BlockNewDocsUrp" and is small enough can be an inner class really.

dsmiley (commented):
Lots of feedback... some is irrelevant depending on what path we take.  Especially, read my comment in the factory first.

solr/solr-ref-guide/modules/configuration-guide/pages/update-request-processors.adoc
gerlowskija: Well, the URP certainly can't go _after_ DistributedUpdateProcessor for the reason laid out in the docs - you don't want some subset of your replicas rejecting certain docs because their segments have them disagree on the current field count.

Putting it _before_ DUP also has downsides, as you pointed out.  The field count check is going to happen on that somewhat arbitrary "first node", instead of a the more authoritative "leader".  And that adds wonkiness - if replicas disagree on field-count, this means you could get into situations where a given update could either succeed or fail depending on whether it gets lucky or not with its "first node".  (Hopefully this isn't a big deal though, since the URP isn't trying to be an exact limit - but it's still ugly and potentially confusing.)

This was what earlier versions of this PR were trying to avoid by gating the field-count check so that it only happened for docs that were being processed by their shard leader.

solr/core/src/java/org/apache/solr/update/processor/NumFieldsMonitor.java
gerlowskija: Is there a reason that `withSearcher` throws an IOException?  Would it be worth having one of those methods that consumes a vanilla Function instead of a `IOFunction`?

solr/core/src/test-files/solr/configsets/cloud-minimal-field-limiting/conf/solrconfig.xml
dsmiley: Having no-config vs config doesn't conceptually change where the <updateProcessor> element can be placed.
ex:
```<updateProcessor blah lbah />```
vs
```
<updateProcessor blah blah>
  <int blah blah/>
</updateProcessor>
```
Can be in the same places.
I went to the ref guide's [page on URPs](https://solr.apache.org/guide/solr/latest/configuration-guide/update-request-processors.html#custom-update-request-processor-chain) and observe the first example shows a chain with several URPs, one of which has config.  I think it's natural to place URPs inline in the chain.  Separating it out is for composability if there are multiple chains and URPs and you want to avoid possible repetition.

solr/core/src/java/org/apache/solr/update/processor/NumFieldsMonitor.java
dsmiley: Maybe; Java exception handling is such a PITA.... I didn't want to defeat the lambda convenience if the lambda threw an IOException which is rather common.

solr/core/src/java/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessorFactory.java
dsmiley: > factory to be abstracted away from exactly how the currentNumFields is calculated

Unless that abstraction shows utility here, I don't think it's worth another class.

> it allows us to make as few (potentially expensive) calls to SIS.getFieldInfos()

That's not expensive!  `req.getSearcher().getFieldInfos()`  It gets a searcher, then returns a value in a field.  Furthermore you'll do it just once on the indexing stream request.

> as epugh pointed out in an earlier comment, NumFieldsMonitor seems like it could be applied in other places.

Sorry, what?

I didn't mean rename the factory (it's well named!).  I think the URP should be a small inner class of this factory named whatever makes sense.  Not public.

dsmiley: To be clear, I am proposing that the whole NumFieldsMonitor thing (and how we init it, etc.) be replaced by no more than *one line* of code.  :-)

gerlowskija: > replaced by no more than one line of code

NumFieldsMonitor is more LOC and more Java classes, but IMO both of those are red-herrings in terms of complexity and maintenance burden.  My subjective opinion is that an unfamiliar reader would have an easier time grokking NumFieldsMonitor than they would parsing through the reference-checking logic that (as I understand it) the alternative would entail.

That said I'm surprised to hear your confidence in "one line".  My understanding of what you proposed, based on your earlier comment [here](https://github.com/apache/solr/pull/2395#issuecomment-2050095834), sounded like a good bit more than that.  It'd need to: hold onto a weak-reference to the previous searcher, do some reference equality check on each invocation, fetch the updated number of fields, update the weak-ref if the current searcher is now different, etc.  If using `withSearcher()` there's IOException handling to do as well, etc.

If you've got one line to do this all, or I'm misunderstanding your suggestion - please share 😄 .  I'm happy to take a commit on this branch if it can do it all in 1 line.

solr/core/src/java/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessor.java
gerlowskija: Missed this earlier; done!

solr/core/src/java/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessorFactory.java
dsmiley: Multiple ideas have been discussed, so it's a bit confusing.  My epiphany from your recent changes is that we only need to know how many fields exist at the time that the URP is instantiated, unlike the first incarnation.  This ought to be a major simplifier and it nullifies my idea around wanting a WeakReference to a SolrIndexSearcher.  May I do the change to your PR here in one commit that you might revert if you don't like it?

dsmiley:
See https://github.com/gerlowskija/solr/pull/5 for a commit that makes a large simplification:
* No NumFieldsMonitor
* No named URP; anonymous class is fine (and less code)
* URP need not be inserted in the happy path!

The single line I refer to earlier is as follows in my PR to your PR:
`final int currentNumFields = req.getSearcher().getFieldInfos().size();`

solr/core/src/test-files/solr/configsets/cloud-minimal-field-limiting/conf/solrconfig.xml
gerlowskija: I've updated this to be in line with your request.

Though I'm still curious if there's a reason to "prefer" this going forward, or whether it's subjective.  I've not paid much attention historically to how these appear in our configsets, but naively I'd assume that more composability is better.

solr/core/src/java/org/apache/solr/update/processor/NumFieldLimitingUpdateRequestProcessorFactory.java
gerlowskija: Absolutely, that'd be great.  Thanks @dsmiley !

gerlowskija:
Thanks for the change David - apologies for the back and forth trying to understand what you were suggesting.  I've merged your PR into this branch.

With that wrapped up, I'm going to re-run tests/check/precommit and aim to merge this afternoon.

dsmiley:
Cool; please remember to mention me in CHANGES.txt.  And thanks for adding this protection.

PR #2400, opened 2024-04-11
SOLR-17232: PropertiesOutputStream overrides write(byte[], int, int).

bruno-roustant:
https://issues.apache.org/jira/browse/SOLR-17232


bruno-roustant:
I think PropertiesOutputStream could be renamed IndexOutputOutputStream because it is used to wrap an IndexOutput to expose it as an OutputStream.
I can add some JavaDoc. Do you agree to rename it to IndexOutputOutputStream?

dsmiley:
Lets rename this in main; separate no-JIRA PR.

epugh:
The name "IndexOutputOutputStream", while slightly funny, is fine, with just a bit of docs to highlight what you said "wrap an IndexOutput to expose it as an OutputStream" would have helped me grok it!

bruno-roustant:
Renamed. It was used only in two internal locations.

solr/core/src/java/org/apache/solr/util/IndexOutputOutputStream.java
```diff
@@ -20,11 +20,12 @@
 import java.io.OutputStream;
 import org.apache.lucene.store.IndexOutput;
 
-public class PropertiesOutputStream extends OutputStream {
+/** Wraps an {@link IndexOutput} to expose it as an {@link OutputStream}. */
```
epugh, line 23: thank you!   Helps me understand what it's for (and why the double "Output" in the name!)...

epugh (approved):
lgtm

PR #2402, opened 2024-04-11
SOLR-14763 SolrJ HTTP/2 Async API using CompletableFuture (update for 2024)

jdyer1:
This adds a standard `CompletableFuture`-based API to `Http2SolrClient`, `HttpJdkSolrClient` and `LBHttp2SolrClient`.  The existing methods `asyncReq` are marked `@deprecated` and defer to the new API methods.  Also deprecated are interfaces `AsyncListener` and `Cancellable`.  Uses within Solr of these deprecated methods/interfaces are migrated to the new API.  Also test coverage for these async methods is improved.

This is based loosely on [PR 1770](https://github.com/apache/lucene-solr/pull/1770).  This PR does not add async methods to `CloudHttp2SolrClient`, neither does it attempt to unify the async API between `LBHttp2SolrClient` and  `Http2SolrClient`.  

There are several TODOs around `LBHttpSolrClient` & `LBHttp2SolrClient` to investigate and fix possible bugs and improve test coverage.  These TODOs are beyond the scope of this PR.

Once merged, we should follow-up with a separate PR to remove the deprecated methods and interfaces in main/Solr10 only.

cc: [rishisankar](https://github.com/rishisankar)


solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -417,67 +419,101 @@ public void send(OutStream outStream, SolrRequest<?> req, String collection) thr
...
+                        mdcCopyHelper.onBegin(null);
+                        log.debug("response processing success");
+                        future.complete(body);
+                      } catch (RemoteSolrException | SolrServerException e) {
                         mdcCopyHelper.onBegin(null);
                         log.debug("response processing failed", e);
-                        asyncListener.onFailure(e);
+                        future.completeExceptionally(e);
+                      } finally {
+                        log.debug("response processing completed");
```
dsmiley, line 490: for consistency with 2 cases right above, these 2 lines should be reversed.

solr/solrj/src/java/org/apache/solr/client/solrj/util/Cancellable.java
```diff
@@ -18,9 +18,11 @@
 package org.apache.solr.client.solrj.util;
 
 /**
- * The return type for solrJ asynchronous requests, providing a mechanism whereby callers may
- * request cancellation.
+ * @deprecated Use the async variants that return CompletableFuture.
```
dsmiley, line 21: This line is supposed to be at the end of the javadoc; isn't supposed to encompass current description.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/HttpSolrClientTestBase.java
```diff
@@ -585,17 +597,21 @@ protected void testUpdateAsync() throws Exception {
...
     DebugServlet.addResponseHeader("Content-Type", "application/xml; charset=UTF-8");
     String url = getBaseUrl() + DEBUG_SERVLET_PATH;
     HttpSolrClientBuilderBase<?, ?> b =
         builder(url, DEFAULT_CONNECTION_TIMEOUT, DEFAULT_CONNECTION_TIMEOUT).withResponseParser(rp);
     int limit = 10;
-    CountDownLatch cdl = new CountDownLatch(limit);
-    DebugAsyncListener[] listeners = new DebugAsyncListener[limit];
-    Cancellable[] cancellables = new Cancellable[limit];
+
+    CountDownLatch cdl = new CountDownLatch(limit); // Deprecated API use
```
dsmiley, line 609: nitpick: just call this "latch".  Same goes elsewhere.  Abbreviations like this are hard to decipher later.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/LBHttp2SolrClientTest.java
```diff
@@ -51,4 +67,249 @@ public void testLBHttp2SolrClientWithTheseParamNamesInTheUrl() {
...
+            fail("Response " + index + " ended in failure: " + ee);
+          }
+          lastResponse = lastRsp.getResponse();
+        }
+
+        // The Mock will return {"response": index}.
+        assertEquals("" + index, lastResponse.get("response"));
+      }
+
+      // TODO: LBHttp2SolrClient creates a new "endpoint iterator" per request, thus
```
dsmiley, line 216: Huh.  I think they should behave the same, first of all, because it's not pertinent to the HTTP Client.  But I think the first is supposed to be chosen because the order can be chosen intentionally (say prefer the replica that is a leader, or prefer pull replicas, etc.) and passed in sorted by such rules.

dsmiley (commented):
Great stuff here James!

solr/solrj/src/test/org/apache/solr/client/solrj/impl/LBHttp2SolrClientTest.java
jdyer1: My concern is that unless the first node in the passed-in list is in the Zombies, it always gets all the requests.  Maybe this is the intention, that the caller is supposed to shuffle the list themselves so as to give ultimate control.  But this is not what I expected from a class named "Load Balancing Client".  There seemed to be little by way of doc comments and unit tests to confirm this. Probably, we need a follow-up issue to add more test coverage to these LB clients and then better document or fix the behavior.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/HttpSolrClientTestBase.java
jdyer1: I think I hunted these all down and changed them.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
jdyer1: so true...

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -417,67 +419,101 @@ public void send(OutStream outStream, SolrRequest<?> req, String collection) thr
...
+            new InputStreamResponseListener() {
+              @Override
+              public void onHeaders(Response response) {
+                super.onHeaders(response);
+                InputStreamResponseListener listener = this;
+                executor.execute(
+                    () -> {
+                      InputStream is = listener.getInputStream();
+
+                      // TODO: Original PR had this, but we do not close the stream.
```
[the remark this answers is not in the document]
jdyer1: yes, removed this comment.  I just wanted to be sure someone agreed with me as the original PR seemed to have received a lot of scrutiny and perhaps this was intentional.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -247,7 +247,9 @@ private HttpClient createHttpClient(Builder builder) {
     this.authenticationStore = new AuthenticationStoreHolder();
     httpClient.setAuthenticationStore(this.authenticationStore);
 
-    httpClient.setConnectTimeout(builder.connectionTimeoutMillis);
+    if (builder.connectionTimeoutMillis != null) {
```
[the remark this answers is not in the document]
jdyer1: yes true.  I've removed it and maybe will do a separate PR to fix the issue when the user does not pass in a connection timeout.

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
```diff
@@ -155,43 +154,33 @@ public void submit(
...
+    CompletableFuture<LBSolrClient.Rsp> future = this.lbClient.requestAsync(lbReq);
+    future.whenComplete(
+        (rsp, throwable) -> {
+          if (!future.isCompletedExceptionally()) {
+            ssr.nl = rsp.getResponse();
+            srsp.setShardAddress(rsp.getServer());
+            ssr.elapsedTime =
+                TimeUnit.MILLISECONDS.convert(System.nanoTime() - startTime, TimeUnit.NANOSECONDS);
+            responses.add(srsp);
+          } else if (!future.isCancelled()) {
```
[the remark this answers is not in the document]
jdyer1: It makes me a little nervous to be modifying this class as part of this change.  I cannot figure out what test might cover it and I do not know how getting this wrong could cause regressions.  I originally had taken nearly verbatim what the original PR from 3 years ago had.  With your suggestion, I am just checking for a not-null exception to replace "onFailure" and a  not-null response to replace "onSuccess".  The existing code does not seem to worry about cancellation here.

jdyer1:
@rishisankar I would like to include your name in the `CHANGES.txt` entry as "Rishi Sankar", as you did the initial work nearly 4 years ago!  If you would rather opt out or use an alternate spelling feel free to let us know.

rishisankar:
@jdyer1 thanks, cool to see this being implemented!

solr/solrj/src/test/org/apache/solr/client/solrj/impl/LBHttp2SolrClientTest.java
dsmiley: Understood.
BTW to be more specific, the sorting happens here: https://github.com/apache/solr/blob/c5709bd4135538fffba3b2c4947dd91ff28d5750/solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java#L1018 with ReplicaListTransformers that sort & shuffle.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java
```diff
@@ -299,7 +299,9 @@ public void testAsyncException() throws Exception {
   public void testAsyncAndCancel() throws Exception {
     String url = getBaseUrl() + DEBUG_SERVLET_PATH;
     Http2SolrClient.Builder b =
-        new Http2SolrClient.Builder(url).withResponseParser(new XMLResponseParser());
+        new Http2SolrClient.Builder(url)
+            .withConnectionTimeout(DEFAULT_CONNECTION_TIMEOUT, TimeUnit.MILLISECONDS)
```
dsmiley, line 303: Not sure if this is out of scope; tied to the other thing you agreed should be separate.  A minor matter anyway.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/LBHttp2SolrClientTest.java
jdyer1: Ok, I can try and improve the doc comment to let the user know they need to sort the list themselves and also take the TODO's out of the unit test.  But as someone who isn't expert on the codebase, I was very surprised when I found it does not actually Load Balance.  It seems there is too tight of coupling between Cloud Solr Client and LB Solr Client here.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java
jdyer1: The reason I had added the null check to the connection timeout was because if the test does not explicitly set it, the user gets a NPE.  I removed it because I agree it is out of scope, but that means I need to explicitly set it in the test, as all prior tests had done.

dsmiley (approved):
Generally +1 from me, subjects to tests passing.

jdyer1:
I would like to merge this tomorrow, and also include with 9.6.0, unless someone says it is too late.

PR #2403, opened 2024-04-15
SOLR-17151 - stop processing components once we have exceeded a query limit

gus-asf:
https://issues.apache.org/jira/browse/SOLR-17151

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
@@ -20,8 +20,16 @@ import static org.apache.solr.common.params.CommonParams.DISTRIB;
 import static org.apache.solr.common.params.CommonParams.FAILURE;
 import static org.apache.solr.common.params.CommonParams.PATH;
 import static org.apache.solr.common.params.CommonParams.STATUS;
+import static org.apache.solr.handler.component.ResponseBuilder.STAGE_DONE;
+import static org.apache.solr.handler.component.ResponseBuilder.STAGE_EXECUTE_QUERY;
+import static org.apache.solr.handler.component.ResponseBuilder.STAGE_GET_FIELDS;
+import static org.apache.solr.handler.component.ResponseBuilder.STAGE_PARSE_QUERY;
+import static org.apache.solr.handler.component.ResponseBuilder.STAGE_START;
+import static org.apache.solr.handler.component.ResponseBuilder.STAGE_TOP_GROUPS;
+import static org.apache.solr.request.SolrRequestInfo.getQueryLimits;
 
 import com.codahale.metrics.Counter;
+import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.lang.invoke.MethodHandles;
@@ -33,6 +41,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
 import org.apache.lucene.index.ExitableDirectoryReader;
 import org.apache.lucene.search.TotalHits;
 import org.apache.solr.client.solrj.SolrRequest.SolrRequestType;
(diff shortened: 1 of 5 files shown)


gus-asf:
Still needs unit test...

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -621,46 +684,72 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
+    if (rb.isDebug()) {
+      NamedList<Object> debug = new NamedList<>();
+      debug.add("explain", new NamedList<>());
+      rb.rsp.add("debug", debug);
     }
   }
 
+  private static boolean checkLimitsBefore(
+      SearchComponent c,
+      String when,
```
sigram, line 739: Hmm, this name looks confusing ... maybe `stageName` or `stageLabel` ?

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -621,46 +684,72 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
+      SolrQueryRequest req,
+      SolrQueryResponse resp,
+      List<SearchComponent> components) {
+
+    return getQueryLimits(req, resp)
+        .maybeExitWithPartialResults(
+            () -> {
```
sigram, line 746: This lambda is evaluated in every call to `checkLimitsBefore`, which seems excessive. Maybe we could cache the list of component names somewhere?

sigram (changes requested):

solr/core/src/java/org/apache/solr/search/QueryLimits.java
```diff
@@ -108,12 +110,21 @@ public String formatExceptionMessage(String label) {
...
+    return maybeExitWithPartialResults(label.get());
+  }
+
   public boolean maybeExitWithPartialResults(String label) throws QueryLimitsExceededException {
     if (isLimitsEnabled() && shouldExit()) {
       if (allowPartialResults) {
         if (rsp != null) {
           rsp.setPartialResults();
-          rsp.addPartialResponseDetail(formatExceptionMessage(label));
+          if (rsp.getResponseHeader().get(RESPONSE_HEADER_PARTIAL_RESULTS_DETAILS_KEY) == null) {
```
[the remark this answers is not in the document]
gus-asf: Well typically before I did this we would get the initial limit violation, and then at least one component violation which isn't really interesting because the first one is where we really stopped. Also there are tests that check the contents of partialResultsDetail using `JSONTestUtil.match()` which doesn't seem to have a syntax for handling duplicate keys (or if it does I can't figure out what it is), and also if someone out there has been using a Jackson Object Mapper or similar to turn the header into an object (or JSON.parse() in javascript where they'll loose one anyway) things could go sideways if they suddenly encounter either a duplicate key or an array there... I can provide a rant about the fact we allow/use duplicate keys in the first place, how we should be referring to our APIs as noggit APIs not JSON APIs because we accept illegal JSON... etc... If you're interested :)

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
gus-asf: oops, meant to do it the other way around so that the lambda is only evaluated if we're busting the limit. Will fix.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
gus-asf: stage is what we use for the finishing steps... maybe phase?

solr/core/src/java/org/apache/solr/search/QueryLimits.java
```diff
@@ -108,22 +110,31 @@ public String formatExceptionMessage(String label) {
...
       } else {
-        throw new QueryLimitsExceededException(formatExceptionMessage(label));
+        throw new QueryLimitsExceededException(formatExceptionMessage(label.get()));
       }
     } else {
       return false;
     }
   }
 
+  public boolean maybeExitWithPartialResults(String label) throws QueryLimitsExceededException {
```
sigram, line 134: I think we should add some javadoc here that explains why we have two different methods for doing essentially the same work, and when to prefer one over the other.

solr/core/src/java/org/apache/solr/search/QueryLimits.java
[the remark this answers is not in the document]
sigram: Hmm, ok... then maybe we should add some processing similar to `computeShardCpuTime` to aggregate multiple details from shard responses into a single value?

PR #2404, opened 2024-04-16
SOLR-12813: subqueries should respect basic auth

rseitz:
https://issues.apache.org/jira/browse/SOLR-12813

# Description

This PR fixes an issue where subqueries don't work when basic auth is enabled. The problem surfaces when 2 or more shards are involved, and when the solr node(s) are not started with the -Dbasicauth system property.

The root cause is that the SubQueryAugmenter discards any basic auth credentials that have been sent with the original query request. There are two separate places where basic auth credentials are lost. First, the SubQueryAugmenter's transform() method issues a subquery by calling EmbeddedSolrServer.query() without ever setting a user principal on the generated QueryRequest. Second, if we look at how EmbeddedSolrServer actually processes a QueryRequest, we see that various transformations are applied, resulting in a SolrQueryRequestBase that fails to return the user principal via getUserPrincipal() even if it had been properly set on the original QueryRequest.

# Solution

SubQueryAugment.transform() now generates a QueryRequest explicitly, so that the user principal can be set on this QueryRequest before it is processed.

EmbeddedSolrServer now attempts to preserve the user principal on a QueryRequest when generating a SolrQueryRequestBase. To do this, EmbeddedSolrServer relies on an updated buildRequestFrom() utility method in SolrRequestParsers that allows for a user principal to be provided explicitly. 

# Tests

TestSubQueryTransformerDistrib has been updated to use basic auth. I have confirmed that the updated test fails without the fix, but passes with the fix. I have also manually tested the change in a 2 node cluster with two shards, where I enabled basic auth and issued a subquery successfully.

# Checklist


solr/core/src/test/org/apache/solr/response/transform/TestSubQueryTransformerDistrib.java
```diff
@@ -61,22 +75,51 @@ public static void setupCluster() throws Exception {
 
     String configName = "solrCloudCollectionConfig";
     int nodeCount = 5;
-    configureCluster(nodeCount).addConfig(configName, configDir).configure();
+
+    final String SECURITY_JSON =
```
epugh, line 79: Starting to feel like `SECURITY_JSON` needs to be moved into some sort of cross cutting concern/utility method as we see this basic code proliferate more and more ;-).   Don't get me wrong, I think it's great we are testing our code with Basic Auth more!!!

solr/core/src/java/org/apache/solr/client/solrj/embedded/EmbeddedSolrServer.java
```diff
@@ -215,8 +215,9 @@ public NamedList<Object> request(SolrRequest<?> request, String coreName)
       if (handler == null) {
         throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "unknown handler: " + path);
       }
-
-      req = _parser.buildRequestFrom(core, params, getContentStreams(request));
+      req =
+          _parser.buildRequestFrom(
```
epugh, line 219: I wonder, should we have still have a _parser.buildRequestFrom() that doesn't require the user principal?   These days, do we need to make sure that every method has a user principal?   Maybe what I am asking is, are there lots of other places where, if you enable basic auth, then you get errors because the oroginal code didn't plan for that?

epugh (commented):
I am curious if this specific situation is symptomatic of other places where Solr doesn't work if you have Basic Auth turned on?   Or do you think this is a one off kind of bug?  Wondering if the patch needs to be more global in nature?

solr/core/src/test/org/apache/solr/response/transform/TestSubQueryTransformerDistrib.java
rseitz: I'm seeing identical SECURITY_JSON constants declared in CreateToolTest, DeleteToolTest, PackageToolTest, PostToolTest, DistribDocExpirationUpdateProcessorTest, TestPullReplicaWithAuth, and BasicAuthIntegrationTest. I'd be happy to move all those identical ones to a common place. Also happy to be told where the common place should be :)

There are some other SECURITY_JSON declarations that differ from the above ones: CloudAuthStreamingTest, SQLWithAuthzEnabledTest, and TestAuthorizationTest.

The one I've declared here in TestSuQueryTransformerDistrib is basically the same as the first group, but I added forwardCredentials=true which tells the basic auth plugin to propagate creds across internode requests. I thought it would be _necessary_ for me to add this here, because forwardCredentials=true was needed in my manual testing. That's to say, in manual testing with forwardCredentials=false, subqueries did not work -- PKI authentication did not kick in seamlessly as one might have hoped. So that PKI case is an unresolved detail re: this fix -- should we _expect_ subqueries to work with basic auth enabled but forwardCredentials=false and if so, how do we make this work? The other unresolved detail is that here in TestSuQueryTransformerDistrib, I'm finding the test passes _regardless_ of whether forwardCredentials is true, false, or not provided, and I'm not yet sure why this differs from what's observed in manual testing.

solr/core/src/java/org/apache/solr/client/solrj/embedded/EmbeddedSolrServer.java
rseitz: We could remove the buildRequestFrom() that doesn't take a Principal; outside of tests, it's used in only two places. Where it's used in `EmbeddedSolrServer.request(SolrRequest<?> request, String coreName)`, we could get the Principal from the SolrRequest. Where it's used in `DirectSolrConnection.request(SolrRequestHandler handler, SolrParams params, String body)` I think we'd need to pass a null Principal as I'm not seeing where to get one from in that context.

About the larger question -- are there lots of other cases where enabling basic auth causes errors -- I don't think I know enough to give a good answer. I do think that the subquery use case is the most obvious/glaring place for such an error to arise because it involves creating and executing a brand new query.

solr/core/src/java/org/apache/solr/servlet/SolrRequestParsers.java
```diff
@@ -252,7 +260,11 @@ private SolrQueryRequest buildRequestFrom(
         new SolrQueryRequestBase(core, params, requestTimer) {
           @Override
           public Principal getUserPrincipal() {
-            return req == null ? null : req.getUserPrincipal();
+            if (principal != null) {
+              return principal;
+            } else {
+              return req == null ? null : req.getUserPrincipal();
+            }
           }
```
dsmiley, line 268: An idea; not totally sure if it'll work until you try:  Instead of passing in a Principal to buildRequestFrom, I think the line you are changing here could instead see that *if* HttpServletRequest is null, then see if we can use `org.apache.solr.request.SolrRequestInfo#getUserPrincipal` which works off a ThreadLocal internally.  There may be no SolrRequestInfo either; so ultimately may have null.

solr/core/src/java/org/apache/solr/response/transform/SubQueryAugmenterFactory.java
```diff
@@ -340,7 +346,9 @@ public void transform(SolrDocument doc, int docid) {
     final SolrParams docWithDeprefixed =
         SolrParams.wrapDefaults(new DocRowParams(doc, prefix, separator), baseSubParams);
     try {
-      QueryResponse rsp = server.query(coreName, docWithDeprefixed);
+      QueryRequest req = new QueryRequest(docWithDeprefixed);
+      req.setUserPrincipal(principal);
+      QueryResponse rsp = req.process(server, coreName);
```
dsmiley, line 351: I'm kind of concerned this is merely one spot of potentially many other cases.  Could/should we depend on SolrRequestInfo to make this automatic?

solr/core/src/java/org/apache/solr/servlet/SolrRequestParsers.java
rseitz: I tried it and got a stack overflow. Here is what happens:

`EmbeddedSolrServer#request ` uses `SolrRequestParsers#buildRequestFrom` to obtain a`SolrQueryRequest req`. It then calls `SolrRequestInfo.setRequestInfo(new SolrRequestInfo(req, rsp))` to designate that request as the current one (see [here](https://github.com/apache/solr/blob/9ad70743a218665254f8aff15952ab4859e65b75/solr/core/src/java/org/apache/solr/client/solrj/embedded/EmbeddedSolrServer.java#L224)). 

The `getUserPrincipal` method that I'm changing here belongs to an anonymous inner class that extends `SolrQueryRequestBase`. At the time this `getUserPrincipal` method is actually called, `SolrRequestInfo.getRequestInfo` would return a `SolrRequestInfo` containing the same `SolrQueryRequest` instance that `getUserPrincipal` is being called on in the first place.

solr/core/src/java/org/apache/solr/response/transform/SubQueryAugmenterFactory.java
rseitz: I commented re: SolrRequestInfo separately but will add detail there. The question is: what can we assume about the contents of SolrRequestInfo? In the context of subquery execution, what we find at the top of the stack inside SolrRequestInfo is the _new_ subquery request that we've constructed, not the enclosing/original request that contained the authentication info. The following excerpt from EmbeddedSolrServer shows how SolrRequestInfo gets set before a subquery is actually executed:

```
      req = _parser.buildRequestFrom(core, params, getContentStreams(request));
      SolrRequestInfo.setRequestInfo(new SolrRequestInfo(req, rsp));
      core.execute(handler, req, rsp);
```

My basic intuition is that if we're creating a brand new QueryRequest, we really _should_ be copying any authentication info onto it that needs to be used in its execution. I don't see any notion in the codebase of a parent/child relationship between queries, where a child query could somehow know who its parent is and derive auth info from the parent. Therefore, it really seems to me like copying the auth info is the logical option.

dsmiley (approved):
+1 works for me; thanks for your detailed investigation on this @rseitz !

epugh:
Thanks @rseitz  for the contribution.   Tag me on the `security.json` refactor please!

epugh:
I think open a seperate PR, just to keep the one that was merged and committed seperate from a new one.  I'd love to see the suggested PR...   Do we also need a test there?

rseitz:
I opened a separate PR here: https://github.com/apache/solr/pull/2429
Basically, it's a one-line change that updates another `_parser.buildRequestFrom()` call inside `EmbeddedSolrServer.request()`.  This other call doesn't seem to be involved in the subquery execution codepath, so it didn't need to be updated for fixing SOLR-12813. In earlier discussion we had here, folks asked if there were other cases we might be missing where a Principal is lost, so I did some more looking around. This is the other such case that I've come across so far. I added some more description on the new PR.

PR #2405, opened 2024-04-17
SOLR-10654: Introduce output of Prometheus metrics directly from Solr

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-10654

# Description

Introduce a new Prometheus response writer `/solr/admin/metrics?wt=prometheus` to output Dropwizards Solr Core registry metrics directly from Solr. Output looks similar to the prometheus exporter. This exports `solr.core.*` `solr.node` `solr.jvm`  `solr.jetty` metric registries 

# Solution

Currently, Solr metrics are collected by Dropwizard and differentiated by metric names. This is an anti-pattern to Prometheus. Manually export Dropwizard Solr core registry metrics to a new `SolrPrometheusExporter` base class to parse out the labels from the metric name and output in a Prometheus readable format for the Prometheus Response writer. The exporter is just temporarily created to export a snapshot of what the current Dropwizard registry is at. These are not held in memory.

The exporter will categorize these metrics and export to prometheus format with labels depending on the category and metric names.

In standalone, all labels for the `solr.core` registry includes`core`. In cloud mode, all metrics include `core`, `collection`, `shard`, `replica`

**NOTE: Not all metrics are possible to convert to Prometheus such as Dropwizard string metrics. This is because Prometheus does not support non-numeric types**

# Tests

Added `MetricsHandlerTest`, `SolrPrometheusExporterTest`, `TestPrometheusReponseWriter`

# Checklist


solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreSearcherMetric.java
```diff
@@ -0,0 +1,59 @@
...
+  public SolrCoreSearcherMetric(
+      Metric dropwizardMetric, String coreName, String metricName, boolean cloudMode) {
+    super(dropwizardMetric, coreName, metricName, cloudMode);
+  }
+
+  @Override
+  public SolrCoreMetric parseLabels() {
+    String[] parsedMetric = metricName.split("\\.");
+    if (dropwizardMetric instanceof Gauge) {
+      String type = parsedMetric[2];
```
richardmlu, line 39: Is it possible for `metricName` to be size 1 at this point? If so, we should surround this in a try/throw or a check.

richardmlu (approved):

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreSearcherMetric.java
mlbiscoc: I have not seen a metric with delimited with a size of 1 yet but in cases that does, I did surround with try/catch at a higher level [here](https://github.com/apache/solr/blob/d594f50761ec835fedc632ba8d19521918079e05/solr/core/src/java/org/apache/solr/util/stats/MetricUtils.java#L211)

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreMetric.java
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+package org.apache.solr.metrics.prometheus.core;
+
+import com.codahale.metrics.Metric;
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.solr.metrics.prometheus.SolrPrometheusCoreRegistry;
+
+public abstract class SolrCoreMetric {
```
dsmiley, line 24: I know this is a draft but even a one sentence Javadoc would be helpful.  At a glance, this is Prometheus specific.

solr/core/src/java/org/apache/solr/metrics/prometheus/core/package-info.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * The {@link org.apache.solr.metrics.prometheus.core.SolrCoreMetric} is a wrapper to export {@link
```
dsmiley, line 19: Perfect; say this on SolrCoreMetric itself

solr/core/src/java/org/apache/solr/metrics/prometheus/SolrPrometheusRegistry.java
```diff
@@ -0,0 +1,144 @@
...
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.Timer;
+import io.prometheus.metrics.core.metrics.Counter;
+import io.prometheus.metrics.core.metrics.Gauge;
+import io.prometheus.metrics.model.registry.PrometheusRegistry;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+
+public abstract class SolrPrometheusRegistry {
```
dsmiley, line 28: What is the lifecycle of these?  In other words, are they created temporarily to serve a metrics request, or are they lived longer?  I'm a little concerned about synchronization if it lives longer (i.e. when a core goes away, etc.)

dsmiley (commented):
There are enough additional dependencies that I feel strongly this should be another module.

It's a shame to see so much code to do the mapping but... shrug, ah well.

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreMetric.java
mlbiscoc: Will add Javadocs across the new classes

solr/core/src/java/org/apache/solr/metrics/prometheus/SolrPrometheusRegistry.java
mlbiscoc: They are just created temporarily to serve metrics requests so keeping anything in sync will not be an issue. Dropwizard holds all the metrics. We just export the registry to Prometheus when called.

mlbiscoc: Changed name from PrometheusRegistry to Exporter to better understand we are just exporting/transforming dropwizard to prometheus instead of keeping separate registries.

mlbiscoc:
Moving this out of draft for review

dsmiley:
> There are enough additional dependencies that I feel strongly this should be another module.

WDYT?

dsmiley:
```
io.prometheus:prometheus-metrics-config:1.1.0 (1 constraints: 3e164ed8)
io.prometheus:prometheus-metrics-exposition-formats:1.1.0 (1 constraints: 0405f335)
io.prometheus:prometheus-metrics-model:1.1.0 (2 constraints: 411b133b)
io.prometheus:prometheus-metrics-shaded-protobuf:1.1.0 (1 constraints: 
```

No, there are 4 dependencies.  Only the last one concerns me.  Why is protobuf needed at all?  Can it be excluded?  Can any of the others be excluded as well, while you're at it?

mlbiscoc:
> ```
> io.prometheus:prometheus-metrics-config:1.1.0 (1 constraints: 3e164ed8)
> io.prometheus:prometheus-metrics-exposition-formats:1.1.0 (1 constraints: 0405f335)
> io.prometheus:prometheus-metrics-model:1.1.0 (2 constraints: 411b133b)
> io.prometheus:prometheus-metrics-shaded-protobuf:1.1.0 (1 constraints: 
> ```
> 
> No, there are 4 dependencies. Only the last one concerns me. Why is protobuf needed at all? Can it be excluded? Can any of the others be excluded as well, while you're at it?

Ah ok you were correct. I went ahead and removed the transitive dependencies `prometheus-metrics-shaded-protobuf` and `prometheus-metrics-config` so there should now only be 2

mlbiscoc:
@dsmiley or anyone else interested in reviewing. Was curious for more feedback on approach. Is this going in the correct direction or do some disagree if this is correct implementation path towards prometheus in Solr.

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreSearcherMetric.java
```diff
@@ -0,0 +1,70 @@
...
+    String[] parsedMetric = metricName.split("\\.");
+    if (!(dropwizardMetric instanceof Counter)) {
+      String type = parsedMetric[2];
+      labels.put("type", type);
+    }
+    return this;
+  }
+
+  @Override
+  public void toPrometheus(SolrPrometheusCoreExporter solrPrometheusCoreRegistry) {
```
dsmiley, line 50: these var names should be changed everywhere so that it's not a "registry"

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreMetric.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public SolrCoreMetric(
+      Metric dropwizardMetric, String coreName, String metricName, boolean cloudMode) {
+    this.dropwizardMetric = dropwizardMetric;
+    this.coreName = coreName;
+    this.metricName = metricName;
+    labels.put("core", coreName);
+    if (cloudMode) {
+      String[] coreNameParsed = coreName.split("_");
+      labels.put("collection", coreNameParsed[1]);
```
dsmiley, line 45: I could see adding some resiliency to this such that underscores in the collection name still work

solr/core/src/java/org/apache/solr/util/stats/MetricUtils.java
```diff
@@ -165,6 +169,71 @@ public static void toSolrInputDocuments(
...
+   *     one</em> of the filters from this list to be included in the output
+   * @param mustMatchFilter a {@link MetricFilter}. A metric <em>must</em> match this filter to be
+   *     included in the output.
+   * @param propertyFilter limit what properties of a metric are returned
+   * @param skipHistograms discard any {@link Histogram}-s and histogram parts of {@link Timer}-s.
+   * @param skipAggregateValues discard internal values of {@link AggregateMetric}-s.
+   * @param compact use compact representation for counters and gauges.
+   * @param consumer consumer that accepts produced {@link SolrPrometheusCoreExporter}-s
+   */
+  public static void toPrometheusRegistry(
```
dsmiley, line 188: It's debatable that this belongs here; it doesn't seem utility-like.  It's called by one place, MetricsHandler, so that would be a decent spot.  Alternatively, given its specific-ness to Prometheus, perhaps the PrometheusResponseWriter is a better home.

solr/core/src/java/org/apache/solr/metrics/prometheus/SolrPrometheusCoreExporter.java
```diff
@@ -0,0 +1,104 @@
...
+   * @param metricName Dropwizard metric name
+   */
+  @Override
+  public void exportDropwizardMetric(Metric dropwizardMetric, String metricName) {
+    SolrCoreMetric solrCoreMetric = categorizeCoreMetric(dropwizardMetric, metricName);
+    solrCoreMetric.parseLabels().toPrometheus(this);
+  }
+
+  private SolrCoreMetric categorizeCoreMetric(Metric dropwizardMetric, String metricName) {
+    String metricCategory = metricName.split("\\.")[0];
```
dsmiley, line 68: I'd rather we don't split merely to get the first component. 
An alternative approach here is to have many startsWith checks so that we're not even parsing/extracting.  I leave it to you to pick a path; either are fine I think.

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
```diff
@@ -0,0 +1,46 @@
...
+import io.prometheus.metrics.expositionformats.PrometheusTextFormatWriter;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Map;
+import org.apache.solr.common.util.NamedList;
+import org.apache.solr.metrics.prometheus.SolrPrometheusCoreExporter;
+import org.apache.solr.request.SolrQueryRequest;
+
+@SuppressWarnings(value = "unchecked")
+public class PrometheusResponseWriter extends RawResponseWriter {
```
dsmiley, line 28: Needs javadocs.  Like say that this is *ONLY* used by the MetricsHandler

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -3032,6 +3034,7 @@ public PluginBag<QueryResponseWriter> getResponseWriters() {
     m.put("csv", new CSVResponseWriter());
     m.put("schema.xml", new SchemaXmlResponseWriter());
     m.put("smile", new SmileResponseWriter());
+    m.put(PROMETHEUS_METRICS_WT, new PrometheusResponseWriter());
```
dsmiley, line 3037: This is so special-purpose that I think it doesn't belong here.  Can we get away with not registering it?

solr/core/src/java/org/apache/solr/util/stats/MetricUtils.java
```diff
@@ -165,6 +169,71 @@ public static void toSolrInputDocuments(
...
+        shouldMatchFilters,
+        mustMatchFilter,
+        propertyFilter,
+        skipHistograms,
+        skipAggregateValues,
+        compact,
+        false,
+        (metricName, metric) -> {
+          try {
+            Metric dropwizardMetric = dropwizardMetrics.get(metricName);
```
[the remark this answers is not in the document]
mlbiscoc: A bit confused where you would prefer this. The metric is retrieved within the scope of the `BiConsumer` from `toMap`. I can't move it where `SolrPrometheusCoreExporter` is declared because it is out of scope of this block and the metrics are looped through the registry from `toMaps`.

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreSearcherMetric.java
```diff
@@ -0,0 +1,70 @@
...
+    if (dropwizardMetric instanceof Gauge) {
+      if (metricName.endsWith("liveDocsCache")) {
+        solrPrometheusCoreRegistry.exportGauge(
+            CORE_CACHE_SEARCHER_METRICS,
+            (Gauge<?>) dropwizardMetric,
+            Labels.of(new ArrayList<>(labels.keySet()), new ArrayList<>(labels.values())));
```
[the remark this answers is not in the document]
mlbiscoc: Addressed with a `getLabels()` and I agree, it is much cleaner.

solr/core/src/java/org/apache/solr/metrics/prometheus/SolrPrometheusCoreExporter.java
```diff
@@ -0,0 +1,104 @@
...
+  public static final String QUERY = "QUERY";
+  public static final String UPDATE = "UPDATE";
+  public static final String REPLICATION = "REPLICATION";
+  public static final String TLOG = "TLOG";
+  public static final String CACHE = "CACHE";
+  public static final String SEARCHER = "SEARCHER";
+  public static final String HIGHLIGHTER = "HIGHLIGHTER";
+  public static final String INDEX = "INDEX";
+  public static final String CORE = "CORE";
+
```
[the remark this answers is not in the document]
mlbiscoc: I created `PrometheusExporterInfo` that holds all these constant strings. Let me know what you think.

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreMetric.java
```diff
@@ -0,0 +1,54 @@
...
+ * io.prometheus.metrics.model.snapshots.DataPointSnapshot} and register to a {@link
+ * SolrPrometheusCoreExporter}. {@link com.codahale.metrics.MetricRegistry} does not support tags
+ * unlike prometheus. Metrics registered to the registry need to be parsed out from the metric name
+ * to be exported to {@link io.prometheus.metrics.model.snapshots.DataPointSnapshot}
+ */
+public abstract class SolrCoreMetric {
+  public Metric dropwizardMetric;
+  public String coreName;
+  public String metricName;
+  public Map<String, String> labels = new HashMap<>();
```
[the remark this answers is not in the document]
mlbiscoc: The `PrometheusTextFormatWriter` will take the MetricSnapshots exported and just output in lexicographical order. So a `HashMap` to `LinkedHashMap` wouldn't really change anything.

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
```diff
@@ -0,0 +1,46 @@
...
+  @Override
+  public void write(OutputStream out, SolrQueryRequest request, SolrQueryResponse response)
+      throws IOException {
+    NamedList<Object> prometheusRegistries =
+        (NamedList<Object>) response.getValues().get("metrics");
+    Map<String, Object> registryMap = prometheusRegistries.asShallowMap();
```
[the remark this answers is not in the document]
mlbiscoc: Got it, moved to an `Iterator` instead.

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreSearcherMetric.java
mlbiscoc: Thanks for catching this. Mistakenly did not change the var names after changing the classes name. Should all be changed now.

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreMetric.java
mlbiscoc: Good catch. I created a new function with this regex pattern `"^core_(.*)_(shard[0-9]+)_(replica_n[0-9]+)"` for core names looking like `core_test_collection_shard1_replica_n1`. Also created a test under `SolrCoreMetricTest` to avoid regressions if for some reason the pattern for core name changes.

solr/core/src/java/org/apache/solr/util/stats/MetricUtils.java
mlbiscoc: That's a fair point. Moved it to the PrometheusResponseWriter so it's known it's only used for this.

solr/core/src/java/org/apache/solr/metrics/prometheus/SolrPrometheusCoreExporter.java
mlbiscoc: Personally more a fan of split instead of many `startsWith`. I'll keep as if unless someone feels strongly otherwise.

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
mlbiscoc: Done!

solr/core/src/java/org/apache/solr/core/SolrCore.java
mlbiscoc: Could not figure out a way to do with without registering it. Is there some example somewhere where a response writer exists without registering the class?

solr/core/src/java/org/apache/solr/metrics/prometheus/SolrPrometheusCoreExporter.java
dsmiley: split takes a "limit" second arg.

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreSearcherMetric.java
```diff
@@ -47,24 +45,18 @@ public SolrCoreMetric parseLabels() {
   }
 
   @Override
-  public void toPrometheus(SolrPrometheusCoreExporter solrPrometheusCoreRegistry) {
+  public void toPrometheus(SolrPrometheusCoreExporter solrPrometheusCoreExporter) {
```
dsmiley, line 48: nice.  You also could have simply called it "exporter" or "prometheus".  It might then be so short enough  that the usages become one-liners.  Any way it's fine.

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
```diff
@@ -16,31 +16,115 @@
...
   @Override
   public void write(OutputStream out, SolrQueryRequest request, SolrQueryResponse response)
       throws IOException {
     NamedList<Object> prometheusRegistries =
         (NamedList<Object>) response.getValues().get("metrics");
-    Map<String, Object> registryMap = prometheusRegistries.asShallowMap();
-    PrometheusTextFormatWriter prometheusTextFormatWriter = new PrometheusTextFormatWriter(false);
-    registryMap.forEach(
-        (name, registry) -> {
+    Iterator<Map.Entry<String, Object>> it = prometheusRegistries.iterator();
```
dsmiley, line 57: Using an Iterator manually reminds me of the Java 1.1 days.... in a bad way.  It's needless here, can use Java 7 for loop syntax on any Iterable., saving you 2 lines.

solr/core/src/java/org/apache/solr/metrics/prometheus/PrometheusExporterInfo.java
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.metrics.prometheus;
+
+public interface PrometheusExporterInfo {
+  /** Category of prefix Solr Core dropwizard handler metric names */
+  enum CoreCategory {
```
dsmiley, line 21: Should this go into SolrCoreMetric as it's CORE specific?

solr/core/src/java/org/apache/solr/core/SolrCore.java
dsmiley: Nevermind.  We even have schema xml format, another special purpose one.  Maybe some day this could be improved.

solr/core/src/java/org/apache/solr/metrics/prometheus/SolrPrometheusCoreExporter.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  private SolrCoreMetric categorizeCoreMetric(Metric dropwizardMetric, String metricName) {
+    String metricCategory = metricName.split("\\.")[0];
+    switch (CoreCategory.valueOf(metricCategory)) {
+      case ADMIN:
+      case QUERY:
+      case UPDATE:
+      case REPLICATION:
+        {
```
dsmiley, line 64: can these redundant brackets in these cases be removed?

solr/core/src/java/org/apache/solr/metrics/prometheus/SolrPrometheusExporter.java
```diff
@@ -0,0 +1,203 @@
...
+      metricGauges.put(metricName, new ArrayList<>());
+    }
+    metricGauges.get(metricName).add(dataPoint);
+  }
+
+  /**
+   * Returns an immutable {@link MetricSnapshots} from the {@link
+   * io.prometheus.metrics.model.snapshots.DataPointSnapshot}s collected from the registry
+   */
+  public MetricSnapshots collect() {
```
dsmiley, line 191: Does the order matter, like will we return all counters then all gauges in the result?  If so, I could see that being kind of annoying but not a deal breaker of course.

mlbiscoc: Order doesn't matter. The constructor of `MetricSnapshots` gets sorted by the Metric name. See this [constructor](https://github.com/prometheus/client_java/blob/c0a38270c021e4d63998a98a68bacf2b6f32ace5/prometheus-metrics-model/src/main/java/io/prometheus/metrics/model/snapshots/MetricSnapshots.java#L40). But with metrics, it shouldn't matter anyways although I agree, it can be annoying. All the metrics would be aggregated on a Grafana level with prometheus queries.

solr/core/src/java/org/apache/solr/metrics/prometheus/SolrPrometheusCoreExporter.java
mlbiscoc: Set limit to 2 and only grab the `[0]` index.

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreHandlerMetric.java
```diff
@@ -43,47 +42,35 @@ public SolrCoreMetric parseLabels() {
     String category = parsedMetric[0];
     String handler = parsedMetric[1];
     String type = parsedMetric[2];
-    labels.put("category", category);
-    labels.put("type", type);
-    labels.put("handler", handler);
+    labels.putAll(Map.of("category", category, "type", type, "handler", handler));
```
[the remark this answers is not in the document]
mlbiscoc: haha duh... 🤦 Changed back accordingly and inline'd them!

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreSearcherMetric.java
mlbiscoc: Changed to just exporter. It still makes sense, I was always just big on verbosity but shorter names for one-liners is better.

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
mlbiscoc: Changed to for loop syntax

solr/core/src/java/org/apache/solr/metrics/prometheus/PrometheusExporterInfo.java
mlbiscoc: Implemented it on `SolrPrometheusCoreExporter` instead. It is `CORE` specific so you are correct. Doesn't make sense implementing on the base class.

solr/core/src/java/org/apache/solr/metrics/prometheus/PrometheusCoreExporterInfo.java
```diff
@@ -30,4 +32,7 @@ enum CoreCategory {
     INDEX,
     CORE
   }
+
+  String PROMETHEUS_CLOUD_CORE_REGEX = "^core_(.*)_(shard[0-9]+)_(replica_n[0-9]+)$";
```
dsmiley, line 36: Why separate this into a constant?

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
```diff
@@ -54,11 +54,11 @@ public void write(OutputStream out, SolrQueryRequest request, SolrQueryResponse
       throws IOException {
     NamedList<Object> prometheusRegistries =
         (NamedList<Object>) response.getValues().get("metrics");
-    Iterator<Map.Entry<String, Object>> it = prometheusRegistries.iterator();
     var prometheusTextFormatWriter = new PrometheusTextFormatWriter(false);
-    while (it.hasNext()) {
-      Map.Entry<String, Object> entry = it.next();
-      SolrPrometheusCoreExporter prometheusExporter = (SolrPrometheusCoreExporter) entry.getValue();
+    for (Iterator<Map.Entry<String, Object>> iter = prometheusRegistries.iterator();
```
dsmiley, line 58: This isn't what I meant.  If you use IntelliJ, it should give you a hint -- "'for' loop can be replaced with enhanced 'for' ".  Do that and I think you'll appreciate the results.

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
```diff
@@ -54,11 +54,11 @@ public void write(OutputStream out, SolrQueryRequest request, SolrQueryResponse
...
+        iter.hasNext(); ) {
+      SolrPrometheusCoreExporter prometheusExporter =
```
dsmiley, line 60: Another use of "var" here is plenty clear enough and will make this a one-liner

solr/core/src/java/org/apache/solr/metrics/prometheus/PrometheusCoreExporterInfo.java
```diff
@@ -30,4 +32,7 @@ enum CoreCategory {
...
+  Pattern PROMETHEUS_CLOUD_CORE_PATTERN = Pattern.compile(PROMETHEUS_CLOUD_CORE_REGEX);
```
dsmiley, line 37: But isn't the word "PROMETHEUS" in here actually inaccurate?  This is parsing existing internal names, it's not an external prometheus exported pattern.  I think you could just drop this part.

dsmiley (commented):
Much better.
I know I've been a bit picky/detailed but I'm trying to achieve a level of clarity & conciseness here that will end up being repeated for the other metrics.

What do you think the next step should be?  Totally completing the core metrics so that you have parity with the default Prometheus Exporter config?

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
mlbiscoc: Sorry was never they best at Java. The enhanced for from the hint is a lot better

solr/core/src/java/org/apache/solr/metrics/prometheus/PrometheusCoreExporterInfo.java
mlbiscoc: You're right dropped `PROMETHEUS` in the prefix and also removed the `PROMETHEUS_CLOUD_CORE_REGEX` constant

mlbiscoc:
> Much better. I know I've been a bit picky/detailed but I'm trying to achieve a level of clarity & conciseness here that will end up being repeated for the other metrics.
> 
> What do you think the next step should be? Totally completing the core metrics so that you have parity with the default Prometheus Exporter config?

No worries, I'd prefer you be picky than someone in the future seeing this and it makes no sense to them. I would say next steps are to export the other 3 registries `solr.jetty`, `solr.jvm`, `solr.node` and yes getting as close to the prometheus exporters config as possible. Luckily, in my opinion this was the hardest registry (hopefully I was right) and the ground work for this is already in place to do the rest.

dsmiley:
What's here looks good to me.  Just need to add more; lots more metrics.  Maybe it'll evolve a little once you add another category.  My interest on this at-work is not the core metrics but all the other ones (Solr, JVM, Jetty).  I'd then be inclined to take this to our Solr fork to kick the tires for real world usage.

mlbiscoc:
> What's here looks good to me. Just need to add more; lots more metrics. Maybe it'll evolve a little once you add another category. My interest on this at-work is not the core metrics but all the other ones (Solr, JVM, Jetty). I'd then be inclined to take this to our Solr fork to kick the tires for real world usage.

Thanks. I've been caught up in some other things recently but plan on continuing on the other registries somewhat soon and it shouldn't be as bad to export. Would it be better for me to continue and commit them to this branch, or was this planned to be merged and export the other 3 registries in a separate PR?

dsmiley:
It doesn't feel right to merge this PR right now.  Needs docs.  Also the choice of what metrics are done so far is inverted from the value that would make most sense for this whole approach -- OS/JVM/node issues are highest value.

mlbiscoc:
> It doesn't feel right to merge this PR right now. Needs docs. Also the choice of what metrics are done so far is inverted from the value that would make most sense for this whole approach -- OS/JVM/node issues are highest value.

@dsmiley Just finished up exporting `solr.node` `solr.jvm` and `solr.jetty` metrics to prometheus and is ready for another round of review whenever you are. I updated the description above and attached files if you want to see what the prometheus output looks like. Last thing I did not finish writing up is docs. I'm assuming the 2 places to update would be the [metrics-api](https://solr.apache.org/guide/solr/latest/deployment-guide/metrics-reporting.html#metrics-api) section and possibly [Monitoring with Prometheus and Grafana](https://solr.apache.org/guide/solr/latest/deployment-guide/monitoring-with-prometheus-and-grafana.html)?

mlbiscoc:
Added ref-guide documentation on Prometheus endpoint

solr/core/src/java/org/apache/solr/metrics/prometheus/jvm/SolrJvmBuffersMetric.java
```diff
@@ -0,0 +1,52 @@
...
+  public static final String JVM_BUFFERS = "solr_metrics_jvm_buffers";
+  public static final String JVM_BUFFERS_BYTES = "solr_metrics_jvm_buffers_bytes";
+
+  public SolrJvmBuffersMetric(Metric dropwizardMetric, String metricName) {
+    super(dropwizardMetric, metricName);
+  }
+
+  @Override
+  public SolrJvmMetric parseLabels() {
+    String[] parsedMetric = metricName.split("\\.");
```
dsmiley, line 34: When parsing stuff, it really helps to have a one-line comment of a sample value

solr/core/src/java/org/apache/solr/metrics/prometheus/node/SolrNodeContainerMetric.java
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+package org.apache.solr.metrics.prometheus.node;
+
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.Metric;
+import org.apache.solr.metrics.prometheus.SolrMetric;
+import org.apache.solr.metrics.prometheus.SolrPrometheusExporter;
+
+/* Dropwizard metrics of name CONTAINER.* */
```
dsmiley, line 24: Great; please add similar javadocs for other classes where you forgot

solr/core/src/java/org/apache/solr/metrics/prometheus/node/SolrNodeHandlerMetric.java
```diff
@@ -0,0 +1,64 @@
...
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.Metric;
+import org.apache.solr.metrics.prometheus.SolrMetric;
+import org.apache.solr.metrics.prometheus.SolrPrometheusExporter;
+
+/* Dropwizard metrics of name ADMIN.* and UPDATE.* */
+public class SolrNodeHandlerMetric extends SolrNodeMetric {
+  public static final String NODE_REQUESTS = "solr_metrics_node_requests";
+  public static final String NODE_SECONDS_TOTAL = "solr_metrics_node_requests_time";
+  public static final String NODE_CONNECTIONS = "solr_metrics_node_connections";
```
dsmiley, line 30: Okay but FWIW *IMO* in situations like this class (parsing/exporting) constants like this only diffuse readability (add a needless indirection).  Take it or leave it; your choice.

solr/core/src/java/org/apache/solr/metrics/prometheus/node/SolrPrometheusNodeExporter.java
```diff
@@ -0,0 +1,99 @@
...
+
+  @Override
+  public void exportDropwizardMetric(Metric dropwizardMetric, String metricName) {
+    if (metricName.contains(".threadPool.")) {
+      exportThreadPoolMetric(dropwizardMetric, metricName);
+      return;
+    }
+
+    SolrMetric solrNodeMetric = categorizeMetric(dropwizardMetric, metricName);
+    solrNodeMetric.parseLabels().toPrometheus(this);
```
dsmiley, line 48: if we forget to call parseLabels, will we recognize a mistake?  not a big deal but just a thought

solr/core/src/java/org/apache/solr/metrics/prometheus/node/SolrPrometheusNodeExporter.java
```diff
@@ -0,0 +1,99 @@
...
+      case CONTAINER:
+        return new SolrNodeContainerMetric(dropwizardMetric, metricName);
+      default:
+        return new SolrNoOpMetric();
+    }
+  }
+
+  private void exportThreadPoolMetric(Metric dropwizardMetric, String metricName) {
+    Labels labels;
+    String[] parsedMetric = metricName.split("\\.");
```
dsmiley, line 71: again and in many places, a sample value would be awesome

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
```diff
@@ -118,13 +108,41 @@ public static void toPrometheus(
...
+  }
+
+  public static SolrPrometheusExporter getExporterType(String registryName) {
+    String coreName;
+    boolean cloudMode = false;
+    String[] rawParsedRegistry = registryName.split("\\.");
+    List<String> parsedRegistry = new ArrayList<>(Arrays.asList(rawParsedRegistry));
+
+    switch (parsedRegistry.get(1)) {
+      case ("core"):
```
dsmiley, line 128: These parenthesis in your case statements are pointless; not sure I've ever seen that

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
```diff
@@ -118,13 +108,41 @@ public static void toPrometheus(
...
 
-    consumer.accept(solrPrometheusCoreExporter);
+    consumer.accept(exporter);
+  }
+
+  public static SolrPrometheusExporter getExporterType(String registryName) {
+    String coreName;
+    boolean cloudMode = false;
+    String[] rawParsedRegistry = registryName.split("\\.");
+    List<String> parsedRegistry = new ArrayList<>(Arrays.asList(rawParsedRegistry));
```
dsmiley, line 125: why clone?

solr/solr-ref-guide/modules/deployment-guide/pages/monitoring-with-prometheus-and-grafana.adoc
```diff
@@ -587,3 +594,27 @@ This screenshot shows what it might look like:
 
 .Grafana Dashboard
 image::monitoring-with-prometheus-and-grafana/grafana-solr-dashboard.png[image,width=800]
+
+== Prometheus Metrics API
```
dsmiley, line 598: Not sure we want to name it this way... perhaps "Metrics API with Prometheus format"

dsmiley (commented):
I'd like to suggest an integration test showing all the output (say to a file) stripped of numbers (to remove brittle/changing stuff) to show that we export what we mean to export.  I could imagine someone renaming some metric and then forgetting to update the code here.

mlbiscoc:
> I'd like to suggest an integration test showing all the output (say to a file) stripped of numbers (to remove brittle/changing stuff) to show that we export what we mean to export. I could imagine someone renaming some metric and then forgetting to update the code here.

Thanks for reviewing again. Added in an integration test with a file stripped of values. Also the unit tests I wrote originally should catch if any changes that happen to the naming of the prometheus metric or labels.

Also thanks for catching missing javadocs. Made sure to add them in for each class and added sample values for what is being exported to prometheus.

solr/core/src/java/org/apache/solr/metrics/prometheus/node/SolrNodeHandlerMetric.java
mlbiscoc: I personally prefer constants so I'd leave it unless someone feels strongly about it.

solr/core/src/java/org/apache/solr/metrics/prometheus/node/SolrPrometheusNodeExporter.java
mlbiscoc: I wouldn't say its a mistake. I split these into 2 abstract functions to split out the logic for getting the labels, then converting the correct dropwizard type to the correct prometheus type instead of 1 big function.

I think it's fine if not called. If down the line additional metrics want to be exported from dropwizard, it's possible to just have it exported without any labels and just it's value to which parselabels isn't needed.

solr/solr-ref-guide/modules/deployment-guide/pages/monitoring-with-prometheus-and-grafana.adoc
```diff
@@ -16,13 +16,20 @@
...
+
+* Prometheus Exporter
+* Metrics API with Prometheus format
 
 The Prometheus exporter is included with the full Solr distribution, and is located under `prometheus-exporter/`.
 It is not included in the `slim` Solr distribution.
 
 A Prometheus exporter (`solr-exporter`) allows users to monitor not only Solr metrics which come from the xref:metrics-reporting.adoc#metrics-api[Metrics API], but also facet counts which come from xref:query-guide:faceting.adoc[] and responses to xref:configuration-guide:collections-api.adoc[] commands and xref:ping.adoc[] requests.
 
+The Metrics API provides a Prometheus Response Writer to output Solr metrics natively to be scraped. It is more efficient and drops the need of running the Prometheus Exporter but at the cost of a fixed output and not as flexiable in terms of configurability.
```
cpoerschke, line 29:
```suggestion
The Metrics API provides a Prometheus Response Writer to output Solr metrics natively to be scraped. It is more efficient and drops the need of running the Prometheus Exporter but at the cost of a fixed output and not as flexible in terms of configurability.  
```

dsmiley (commented):
I tried this PR locally using the "techproducts" example.  I thing I noticed is that the metrics filtering appears to not work, which is a big problem.  A test should be added for this, and it should work.  

http://localhost:8983/solr/admin/metrics?category=solr.jetty&wt=prometheus

solr/core/src/java/org/apache/solr/core/SolrCore.java
dsmiley: I filed https://issues.apache.org/jira/browse/SOLR-17354 for improving this matter.

mlbiscoc:
> I tried this PR locally using the "techproducts" example. I thing I noticed is that the metrics filtering appears to not work, which is a big problem. A test should be added for this, and it should work.
> 
> http://localhost:8983/solr/admin/metrics?category=solr.jetty&wt=prometheus

Filtering does work I think the parameter you are looking for is `group`. Try `localhost:8983/solr/admin/metrics?group=jvm&wt=prometheus` and you can change it with `jetty` `node` `core` for the corresponding registries

But still you are correct, I should have a test for this. Will add one in. 

The only other thing I remember now about this is that `regex`, `key`, `expr` parameters for filtering will not work with this prometheus `wt`. I did that on purpose because those 3 parameters are for filtering based on the name of the metric while the others are kind of more generic such as registry or type filtering. I didn't think it made sense because prometheus naming is different from the dropwizard naming after exporting. For example: `solr\.core\..*:QUERY\..*\.requestTimes:max_ms` is not an output of prometheus but to allow `expr` filtering around that felt odd.

Even then if you need extensive filtering or trying to get some specific metric, maybe the Prometheus Exporter process with a custom config might be the better solution.

dsmiley:
Oops; maybe I was hurried.  I'm glad the category works at least!  The other filtering options could be deferred to a future JIRA by matching the prometheus name.  It's a very useful capability; some metrics implemented as gauges are expensive to compute, and others are just excessive output that aren't needed for O(N) for N cores if N is for some users quite high.

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreHandlerMetric.java
```diff
@@ -0,0 +1,77 @@
...
+  @Override
+  public SolrCoreMetric parseLabels() {
+    String[] parsedMetric = metricName.split("\\.");
+    labels.put("category", parsedMetric[0]);
+    labels.put("handler", parsedMetric[1]);
+    labels.put("type", parsedMetric[2]);
+    return this;
+  }
+
+  /*
```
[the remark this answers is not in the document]
mlbiscoc: Sorry for this. Went with your other suggestion of putting it above `parseLabels` and add an empty line below it.

If you still prefer just normal inline comments, I can make the appropriate changes.

solr/core/src/java/org/apache/solr/metrics/prometheus/node/SolrNodeHandlerMetric.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  /*
+   * Metric examples being exported
+   * ADMIN./admin/collections.requests
+   * UPDATE.updateShardHandler.maxConnections
+   */
+  @Override
+  public void toPrometheus(SolrPrometheusExporter exporter) {
+    if (dropwizardMetric instanceof Meter) {
```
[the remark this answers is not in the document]
mlbiscoc: I went ahead and removed most of the `instanceof` checks but kept in some places. In the places I kept, it seemed cleaner to export that way as not all my checks are just straight names checks and we don't export every metric. So we can just export a large amount by just checking its type. I can try refactoring further of removing `instanceof` checks if you still disagree.

solr/core/src/test-files/solr/solr-prometheus-output.txt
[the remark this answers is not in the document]
mlbiscoc: Created it's own prometheus directory to place the file in a place we can place future test files relating to prometheus

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
```diff
@@ -0,0 +1,103 @@
...
+    req.setResponseParser(new NoOpResponseParser("prometheus"));
+    try (SolrClient adminClient = getHttpSolrClient(jetty.getBaseUrl().toString()); ) {
+      NamedList<Object> res = adminClient.request(req);
+      assertNotNull("null response from server", res);
+      String actual = (String) res.get("response");
+      String expectedOutput =
+          Files.readString(
+              new File(TEST_PATH().toString(), "solr-prometheus-output.txt").toPath(),
+              StandardCharsets.UTF_8);
+      assertEquals(expectedOutput, actual.replaceAll("(?<=}).*", ""));
```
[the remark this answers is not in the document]
mlbiscoc: Parsed out the JVM item tag from the test with additional regex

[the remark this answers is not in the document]
mlbiscoc: Added in a comment with explanation.

mlbiscoc:
> Oops; maybe I was hurried. I'm glad the category works at least! The other filtering options could be deferred to a future JIRA by matching the prometheus name. It's a very useful capability; some metrics implemented as gauges are expensive to compute, and others are just excessive output that aren't needed for O(N) for N cores if N is for some users quite high.

Added in a test for the filtering. I'd be happy expand on this and add further filtering feature specific to prometheus after this is ready to be merged.

solr/core/src/test/org/apache/solr/handler/admin/MetricsHandlerTest.java
```diff
@@ -677,6 +706,340 @@ public void testExprMetrics() throws Exception {
...
+
+    actualSnapshot = getMetricSnapshot(actualSnapshots, "solr_metrics_jvm_buffers_bytes");
+    actualGaugeDataPoint =
+        getGaugeDatapointSnapshot(
+            actualSnapshot, Labels.of("item", "MemoryUsed", "pool", "direct"));
+    assertNotNull(actualGaugeDataPoint);
+
+    actualSnapshot = getMetricSnapshot(actualSnapshots, "solr_metrics_jvm_gc_seconds");
+    actualGaugeDataPoint =
+        getGaugeDatapointSnapshot(actualSnapshot, Labels.of("item", "G1-Old-Generation"));
```
dsmiley, line 954: This assertion will fail for varying JVMs; skip this one.  Some others too above... it's okay to make the test looser / less specific.

solr/solr-ref-guide/modules/deployment-guide/pages/metrics-reporting.adoc
```diff
@@ -879,7 +879,7 @@ The response will look like this:
   ]}
 ----
 
-Like other request handlers, the Metrics API can also take the `wt` parameter to define the output format.
+Like other request handlers, the Metrics API can also take the `wt` parameter to define the output format. The Metrics API also provides an additional special `wt` parameter `prometheus` that will output metrics in Prometheus format. This can be used for xref:monitoring-with-prometheus-and-grafana.adoc#metrics-api-with-prometheus-format[Monitoring with Prometheus and Grafana]
```
dsmiley, line 882: After each sentence, put a newline (doesn't change the final render but it to keep editing & diffs more manageable).

Should document what options *do not work* with prometheus.  The regex one doesn't, you said.

dsmiley (commented):
This is super close now.
Can you propose a brief CHANGES.txt under 9.7 "New Features" (I think it's closer to this than an improvement).

solr/core/src/test/org/apache/solr/handler/admin/MetricsHandlerTest.java
mlbiscoc: Removed a few more tests from the JVM registry

solr/solr-ref-guide/modules/deployment-guide/pages/metrics-reporting.adoc
mlbiscoc: Added in a small note about the filter query parameters. Let me know if you prefer a different way to write it on the ref-guide

mlbiscoc:
> This is super close now. Can you propose a brief CHANGES.txt under 9.7 "New Features" (I think it's closer to this than an improvement).

Added entry into CHANGES.txt

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
```diff
@@ -0,0 +1,105 @@
...
+
+  @ClassRule public static SolrJettyTestRule solrClientTestRule = new SolrJettyTestRule();
+  public static JettySolrRunner jetty;
+
+  @BeforeClass
+  public static void beforeClass() throws Exception {
+    solrClientTestRule.startSolr(LuceneTestCase.createTempDir());
+    jetty = solrClientTestRule.getJetty();
+    solrClientTestRule.newCollection().withConfigSet(ExternalPaths.DEFAULT_CONFIGSET).create();
+    jetty.getCoreContainer().waitForLoadingCoresToFinish(30000);
```
dsmiley, line 48: @mlbiscoc Why did you add this line to wait 30secs?  The [default](https://github.com/apache/solr/blob/2c55000296bb04e12e2bfae83eca7e5f11caba0e/solr/test-framework/src/java/org/apache/solr/embedded/JettyConfig.java#L85) is 5min which is plenty.  Creating a core should be synchronous.

@epugh probably copied you in a recent PR and I had the same question.

epugh: I DID copy it from this method!   I was looking for a recent test case that used the rules, and landed on this one...

epugh: I have fixed this in #3988 after auditing places were .waitForLoadingCoresToFinish is called...

mlbiscoc: Hmm I don't remember as this was almost 2 years ago... But I know I would steal templates from other tests as well and it looks like `MetricsHandlerTest` does the same which might be why I did it.

PR #2410, opened 2024-04-18
SOLR-17240: Fix Http2SolrClient maxConnectionsPerDestination usage

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17240

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -139,6 +139,9 @@ Improvements
 * SOLR-14763: Add a CompletableFuture based asynchronous API to Http2SolrClient, HttpJdkSolrClient and LBHttp2SolrClient.
   The previous asynchronous API is deprecated. (Rishi Sankar, James Dyer)
 
+* SOLR-17240: Use the same logic for "maxConnectionsPerDestination" between http1 and http2 in the Http2SolrClient.
+  Http2 is no longer hardcoded to use 4 connections. (Houston Putman, Mark Miller)
+
 Optimizations
 ---------------------
 * SOLR-17144: Close searcherExecutor thread per core after 1 minute (Pierre Salagnac, Christine Poerschke)
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
@@ -206,6 +206,8 @@ public class Http2SolrClient extends HttpSolrClientBase {
     clientConnector.setSelectors(2);
 
     HttpClientTransport transport;
+    int maxConnectionsPerHost =
+        builder.maxConnectionsPerHost != null ? builder.maxConnectionsPerHost : 0;
     if (builder.useHttp1_1) {
       if (log.isDebugEnabled()) {
         log.debug("Create Http2SolrClient with HTTP/1.1 transport");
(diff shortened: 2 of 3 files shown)


jdyer1 (approved):

markrmiller (approved):
Oh wow, I thought it was just defaulting to 4 and you couldn't necessarily change that everywhere in internal use because I saw the setMaxConnectionsPerHost in the builder. I didn't realize that that is a complete noop for http2 mode. That seems like a silly choice in any scenario.

magibney:
javadoc should be updated on [HttpSolrClientBuilderBase.withMaxConnectionsPerHost(int)](https://github.com/apache/solr/blob/188ca8872cdbe61ea229bc40fadee20e7ba50053/solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java#L108-L111), explaining the new difference between how this is handled for http1 vs http2.

I wonder, to approximate similar high-level behavior for http1 vs http2, could the http2 client, under the hood, interpret this setting as `((maxConnectionsPerDest - 1) / 300) + 1`, or something? I don't like having that 300 constant hardcoded there, as presumably the multiplexing factor in jetty could change ... but idk, something along those lines might address [Sanjay's concern](https://issues.apache.org/jira/browse/SOLR-17240?focusedCommentId=17838863#comment-17838863).

HoustonPutman:
Ok I've updated the docs, and used the `maxConcurrentPushedStreams` to get the multiplexing factor.

This should probably be good to go?

markrmiller:
> under the hood, interpret this setting as `((maxConnectionsPerDest - 1) / 300) + 1`, or something? I don't like having that 300 constant hardcoded there, as presumably the multiplexing factor in jetty could change ... but idk, something along those lines might address [Sanjay's concern](https://issues.apache.org/jira/browse/SOLR-17240?focusedCommentId=17838863#comment-17838863).

I don't think you want to do that, this should do what it says and set the max connections and not relate to multiplexing. Multiplexing should be it's own setter if you want it be configurable but trying to tie the two together really wouldn't do the user any favors.

HoustonPutman:
> > under the hood, interpret this setting as `((maxConnectionsPerDest - 1) / 300) + 1`, or something? I don't like having that 300 constant hardcoded there, as presumably the multiplexing factor in jetty could change ... but idk, something along those lines might address [Sanjay's concern](https://issues.apache.org/jira/browse/SOLR-17240?focusedCommentId=17838863#comment-17838863).
> 
> I don't think you want to do that, this should do what it says and set the max connections and not relate to multiplexing. Multiplexing should be it's own setter if you want it be configurable but trying to tie the two together really wouldn't do the user any favors.

Fair point mark. I'll make these things independent, add documentation, and make the defaults across solr make sense for both http1 and http2.

epugh:
Anything I can do to help this move forward?  (Testing wise!).

iamsanjay:
I came across this article where they tested with various different combinations of concurrent Streams and number of http connections.

https://blog.vespa.ai/http2/

> To compare the throughput of HTTP/1.1 vs HTTP/2 in the Vespa container, we measured the rate of HTTP requests (QPS) we could obtain on a single node with 48 CPU cores and 256GB RAM, using [h2load](https://nghttp2.org/documentation/h2load-howto.html) as a benchmarking tool. As expected, single connection throughput increased significantly with HTTP/2. HTTP/2 with 256 concurrent streams gave a throughput of 115 000 requests per seconds compared to 6 500 for HTTP/1.1. Increasing the number of connections from 1 to 128 increased throughput to 115 000 for HTTP/1.1, while HTTP/2 gave 125 000 with the same setup (without request multiplexing). HTTP/2 was also more efficient, with lower CPU utilization—a consequence of its compact protocol representation and header compression. **The highest throughput for HTTP/2 was observed with 4 clients and 256 concurrent streams.** This configuration resulted in 225 000 requests per seconds—roughly double the best case for HTTP/1 (128 connections). The h2load tool was CPU constrained for the single connection benchmark as it could only utilize a single CPU core per connection. Having 4 connections removed the CPU bottleneck. Increasing beyond 4 connections resulted in gradually more overhead and degraded throughput.

For them 4 clients and 256 concurrent streams gives highest throughput.

We should try to test it with different configuration and try to find the right configuration that works for us. Also, I believe, most of the Recovery code in main branch is now using HTTP2 and we can start with testing that part. 

NOTE :- maxConcurrentStreams is a server level configuration.

epugh:
@iamsanjay  did you get a chance to look at this one from a testing perspective?  Seems like such a great ticket...

dsmiley:
Curious if this change has been tested/deployed in any way?

janhoy (approved):

risdenk (approved):

HoustonPutman:
I want to update this PR since a lot of people came and approved it along the way. I did testing a while back, and didn't really see any improvement increasing this number. We are also hitting very high QPS numbers per Solr node (10k QPS) and not really seeing this as a bottleneck.

We can still make this configurable, maybe especially when the http1 method goes away. But I don't see it necessarily as a reason for bad performance...

PR #2411, opened 2024-04-19
SOLR-17234: LBHttp2SolrClient does not skip "zombie" endpoints

jdyer1:
This is to fix a regression bug in `main` only, in that LBHttp2SolrClient does not skip "zombie" Endpoints.  See JIRA for more details.

### DIFF
--- a/solr/solrj/src/test/org/apache/solr/client/solrj/impl/LBHttp2SolrClientTest.java
+++ b/solr/solrj/src/test/org/apache/solr/client/solrj/impl/LBHttp2SolrClientTest.java
@@ -108,8 +108,13 @@ public class LBHttp2SolrClientTest extends SolrTestCase {
           } catch (TimeoutException | ExecutionException e) {
             fail(iterMessage + " Response ended in failure: " + e);
           }
-          if (j == 0) {
-            // The first endpoint gives an exception, so it retries.
+          if (i == 0) {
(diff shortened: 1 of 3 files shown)


jdyer1:
@gerlowskija I'd appreciate your review on this, as here I am undoing behavior you were protecting in unit test `LBSolrClientTest#testServerIteratorWithZombieServers`.  My change restores the behavior of `branch_9x` on `main`.  I do not want to do the wrong thing here.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/LBHttp2SolrClientTest.java
```diff
@@ -108,8 +108,13 @@ public void testAsyncWithFailures() {
           } catch (TimeoutException | ExecutionException e) {
             fail(iterMessage + " Response ended in failure: " + e);
           }
-          if (j == 0) {
-            // The first endpoint gives an exception, so it retries.
+          if (i == 0) {
```
gerlowskija, line 111: [+1] Appreciate the explanatory comment (though it looks like 'tidy' messed up your formatting)!

gerlowskija (approved):
Thanks for catching and fixing this @jdyer1 - it was definitely unintentional on my part.

The fix looks great!

PR #2414, opened 2024-04-21
SOLR-17243: CloudSolrClient support for req.getBasePath

dsmiley:
Now CloudSolrClient can be used to route requests that specify a specific URL. LBSolrClient has never supported this and shouldn't.

https://issues.apache.org/jira/browse/SOLR-17243

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/SolrRequest.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/SolrRequest.java
@@ -139,8 +139,11 @@ public abstract class SolrRequest<T extends SolrResponse> implements Serializabl
   }
 
   /**
-   * The path of this request following the core or collection (if applicable). Sometimes called the
-   * "handler path". Example: {@code /select}
+   * The HTTP path suffix used when sending this request; sometimes called the handler path. The
+   * full path is basically {@link #getBasePath()} + {@link #getCollection()} + {@link #getPath()},
+   * albeit sometimes there is no core/collection component. A typical handler path for a
+   * core/collection is "/select" or "/update". Other handlers are "node handlers"; an example is
+   * "/admin/info/health".
    */
   public void setPath(String path) {
     this.path = path;
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/SolrRequest.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/SolrRequest.java
@@ -31,9 +31,15 @@ import org.apache.solr.common.params.SolrParams;
 import org.apache.solr.common.util.ContentStream;
 
 /**
- * @since solr 1.3
+ * Base class for a SolrJ request to Solr. A common way to send it is via {@link
+ * #process(SolrClient)}.
+ *
+ * <p>Not thread-safe; in fact {@link org.apache.solr.client.solrj.impl.LBSolrClient} will
+ * temporarily modify {@link #setBasePath(String)} during request processing.
  */
 public abstract class SolrRequest<T extends SolrResponse> implements Serializable {
+  // TODO remove usage of setBasePath during processing
+
   // This user principal is typically used by Auth plugins during distributed/sharded search
   private Principal userPrincipal;
 
@@ -132,6 +138,10 @@ public abstract class SolrRequest<T extends SolrResponse> implements Serializabl
     return path;
   }
 
+  /**
+   * The path of this request following the core or collection (if applicable). Sometimes called the
+   * "handler path". Example: {@code /select}
+   */
   public void setPath(String path) {
     this.path = path;
   }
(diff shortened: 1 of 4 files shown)


dsmiley:
FYI @iamsanjay this will be useful for the transition to Jetty based clients which also involves less one-off client creations.  I have WIP that will use this.

solr/solrj/src/java/org/apache/solr/client/solrj/SolrRequest.java
```diff
@@ -31,9 +31,15 @@
 import org.apache.solr.common.util.ContentStream;
 
 /**
- * @since solr 1.3
+ * Base class for a SolrJ request to Solr. A common way to send it is via {@link
+ * #process(SolrClient)}.
+ *
+ * <p>Not thread-safe; in fact {@link org.apache.solr.client.solrj.impl.LBSolrClient} will
```
gerlowskija, line 37: 😬 

Not an issue for this PR, but man, this is kindof scary to see written out.

solr/solrj/src/java/org/apache/solr/client/solrj/SolrRequest.java
```diff
@@ -132,6 +138,10 @@ public String getPath() {
     return path;
   }
 
+  /**
+   * The path of this request following the core or collection (if applicable). Sometimes called the
```
gerlowskija, line 142: [0] Maybe tweak the wording here to make it a little clearer what value this method expects for requests that **don't** target a particular core/collection.

Perhaps something like:

> The HTTP path suffix used when sending this request.
>
> Commonly used with {@link #getBasePath()} and {@link #getCollection()} to assemble the full HTTP path for the request.
> 
> For requests targeting a specific core or collection, callers need only specify those path components that follow the collection name (e.g. "/select" for the request "/solr/coreName/select"). This is sometimes called the "handler" path.
> 
> For requests that don't target a particular core or collection, callers should specify any path components that follow "/solr" or "/api" (e.g. "/admin/info/health" in "/solr/admin/info/health").

Feel free to ignore this advice though - any javadocs here are an improvement.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java
```diff
@@ -86,6 +87,16 @@
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
+/**
+ * SolrCloud-aware implementation of {@link SolrClient}. It's able to route requests to the ideal
+ * node, knowing where the collection and it's replicas are hosted. It will even break up an update
+ * request with a batch of documents knowing where each document goes (assuming there is more than
+ * one shard). It will do some retries and fail-over to alternative replicas in an attempt to proces
```
gerlowskija, line 94: [0] Typo: "proces" -> "process"

gerlowskija (commented):
Left a few minor comments if you do choose to go forward with this - a suggested javadoc changes, caught a typo, etc.  All very minor.

My main concern here is conceptual: CloudSolrClient is the client that folks use if they want all the load-balancing and auto-routing and "intelligence" build in.  If a user wanted to send a request to a particular end-point, why not create their own Http[2]SolrClient.  Or better yet, just pull an existing client out of CSC through the getters that CSC makes available i.e. `cloudClient.getLBClient().getClient(...)`

(Note that this isn't a -1, just some concerns.)

solr/solrj/src/java/org/apache/solr/client/solrj/SolrRequest.java
dsmiley: I'm airing dirty laundry :-)

dsmiley:
>  If a user wanted to send a request to a particular end-point, why not create their own Http[2]SolrClient. Or better yet, just pull an existing client out of CSC through the getters that CSC makes available i.e. cloudClient.getLBClient().getClient(...)

Oh right, I forgot to share I have the same concern.  But I'd like a convenience method like `CloudHttpSolrClient.getHttpSolrClient(String url)`.  How does that sound?

dsmiley:
I think as long as QueryRequest continues to have a getBasePath, there is probably a fundamental conceptual issue with some clients supporting it and some not; it doesn't feel well placed on this abstraction.  Furthermore it sucks that LBSolrClient mutates this path during processing.  I would prefer to see it removed, and instead only HttpSolrClient should support a special request method overloaded that accepts a URL to the node to where the request should be sent.  LBSolrClient would use this method instead of mutating setBasePath, and furthermore all the other callers of setBasePath out there would instead use this pattern.  I just scanned over all of them outside tests and nearly every one submits the request with a client immediately following calling setBasePath.  Again, this suggests that Http2SolrClient should have such an overload.

gerlowskija:
> I forgot to share I have the same concern. But I'd like a convenience method like CloudHttpSolrClient.getHttpSolrClient(String url). How does that sound?

Ah, glad it's not just me.  I think a convenience method is a solid approach here - if you're not familiar with the LB client, `cloudClient.getLBClient().getClient(...)` is pretty unintuitive.

I could quibble about the method name - I'd opt for something that focuses on conveying the usecase instead of just telling me what the method returns.  e.g. `getClientForUrl(...)`. But that's just my 2c, and I don't have strong opinions there.

dsmiley:
As I said in my last comment, I'd prefer to abandon QueryRequest.getBasePath and setBasePath.  https://issues.apache.org/jira/browse/SOLR-17256
It'd still be nice to expose an HttpSolrClient out of CloudSolrClient but my motivation for doing so may be handled by SOLR-17256 so this PR here may not be needed anyway.  I'll take the javadocs though :-)

PR #2419, opened 2024-04-24
Build: gradlew avoid downloader

dsmiley:
* Avoid running GradleWrapper.java if we already have the file
* Don't pass `--source 11` to either script; it's unnecessary and I found there to be an issue in a JDK config where this is (supposedly) incompatible with `--release`, which I don't see specified here (shrug).

### DIFF
--- a/gradlew.bat
+++ b/gradlew.bat
@@ -75,9 +75,11 @@ goto fail
 
 @rem LUCENE-9266: verify and download the gradle wrapper jar if we don't have one.
 set GRADLE_WRAPPER_JAR=%APP_HOME%\gradle\wrapper\gradle-wrapper.jar
-"%JAVA_EXE%" %JAVA_OPTS% --source 11 "%APP_HOME%/buildSrc/src/main/java/org/apache/lucene/gradle/WrapperDownloader.java" "%GRADLE_WRAPPER_JAR%"
-IF %ERRORLEVEL% EQU 1 goto failWithJvmMessage
-IF %ERRORLEVEL% NEQ 0 goto fail
+IF NOT EXIST "%GRADLE_WRAPPER_JAR%" (
+    "%JAVA_EXE%" %JAVA_OPTS% "%APP_HOME%/buildSrc/src/main/java/org/apache/lucene/gradle/WrapperDownloader.java" "%GRADLE_WRAPPER_JAR%"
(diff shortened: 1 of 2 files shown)


dsmiley:
I'll merge this in a couple days if I don't hear back.

uschindler:
Could we have that for Windows, too.

I don't know what your exact problem with '--source 11' was. Basically it would be better to use `--release 11` then it works correctly. The main problem was that `--source` alone is not really a useful combination (as it only enforces source code syntax with Java 11, but not the Java class file format.

We can remove all that source/target/release completely, as we compile the code as-is. It it won't compile anymore if something changes incompatibly. In Lucene's APIJAR generator we also do not pass any version to "java".

uschindler (approved):
Can we have that in Lucene, too?

dsmiley:
I forgot the Windows one; I'm glad I tagged you for review :-)  Maybe you could review that?

Admittedly the --source issue was specific to a custom JVM variant where I work.  Maybe --release alone would work; I didn't check.  Any way, I'd rather remove needless specificity.

I'll do a Lucene PR after this.

No CHANGES.txt for this minor build improvement stuff IMO.

gradlew.bat
```diff
@@ -87,7 +89,7 @@ set CLASSPATH=%GRADLE_WRAPPER_JAR%
 IF NOT EXIST "%APP_HOME%\gradle.properties" (
   @rem local expansion is needed to check ERRORLEVEL inside control blocks.
   setlocal enableDelayedExpansion
-  "%JAVA_EXE%" %JAVA_OPTS% --source 11 "%APP_HOME%/buildSrc/src/main/java/org/apache/lucene/gradle/GradlePropertiesGenerator.java" "%APP_HOME%\gradle\template.gradle.properties" "%APP_HOME%\gradle.properties"
+  "%JAVA_EXE%" %JAVA_OPTS% "%APP_HOME%/buildSrc/src/main/java/org/apache/lucene/gradle/GradlePropertiesGenerator.java" "%APP_HOME%\gradle\template.gradle.properties" "%APP_HOME%\gradle.properties"
```
uschindler, line 92: I have the feeling this should be backslashes, but it's ok (was the same before).

uschindler (approved):
Have not tested this, but looks fine. The "IF NOT EXIST" code is identical to the few lines below, so I assume it works correctly.

PR #2424, opened 2024-04-26
Parses CHANGES.txt to identify contributors

dsmiley:
This script can be used to list contributors as indicated in CHANGES.txt so that we can easily thank them in our release notes.

I tested it on 9.6 and I'm happy with the results!  Added them to: https://cwiki.apache.org/confluence/display/SOLR/ReleaseNote9_6_0

Ideally we incorporate this into the Release Wizard.

### DIFF
--- a/dev-tools/scripts/parseContributorsFromChanges.py
+++ b/dev-tools/scripts/parseContributorsFromChanges.py
@@ -1,0 +1,63 @@
+#  Licensed to the Apache Software Foundation (ASF) under one or more
+#  contributor license agreements.  See the NOTICE file distributed with
+#  this work for additional information regarding copyright ownership.
+#  The ASF licenses this file to You under the Apache License, Version 2.0
+#  (the "License"); you may not use this file except in compliance with
+#  the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import sys
+import re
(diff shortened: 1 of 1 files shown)


dsmiley:
Unlike a Git log, CHANGES.txt has cleaner data and can be edited if we see a formatting or spelling glitch.  For better or worse (?), it doesn't have more trivial changes that were not worthy of CHANGES.txt.

dev-tools/scripts/parseContributorsFromChanges.py
```diff
@@ -0,0 +1,63 @@
...
+    # might have a "via" committer; we only want the author here
+    attribution = attribution.split(" via ")[0] # keep left side
+    # Split the contributors by comma and strip whitespace
+    for contributor in attribution.split(','):
+      contributor = contributor.strip()
+      contributors[contributor] += 1
+
+del contributors['solrbot']
+
+sorted_contributors = sorted(contributors.items(), key=lambda item: item[1], reverse=True)
```
janhoy, line 56: It's not obvious that the list is sorted by most contributions, so perhaps better to do a lexical sort?

dsmiley: Wether it's obvious or not, is this a problem?  It's just a list of people to thank.

gus-asf: Here's what it looks like: LMK if it should change https://solr.staged.apache.org/news.html

dsmiley: A smaller font would be nice... kind of like credits so as to not distract from the primary message of what the release is.  Maybe centered.

janhoy:
If you make an effort to get it into wizard, I can try help if you get stuck..

dsmiley:
I don't actually know Python (producing this script with ChatGPT).  I looked at the release wizard script, which is almost 2k lines and I'm simply out of my depth to be honest.

janhoy (approved):
Let’s merge the script first then integration in wizard can come later

PR #2425, opened 2024-04-27
SOLR-17254 - Major Changes in 9.6

gus-asf:
https://issues.apache.org/jira/browse/SOLR-17254

This section doesn't seem to have been maintained by feature authors very well (myself included), so I've added most of the items from the release announcement. @epugh and I chatted and he agreed that the post tool note however is too small for this list. 

Feedback/suggestions very welcome. I left in the references to the Jira issues, but that's different from prior releases, so I wonder how people feel about it. Personally when I have read this section for prior releases I would always have been interested to have an easy way to find more info these sorts of things. It is a departure from prior style, though, so I could be talked into either removing them or enhancing them with hyperlinks...

### DIFF
--- a/solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-9.adoc
+++ b/solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-9.adoc
@@ -73,15 +73,15 @@ Lucene is upgraded to 9.10.0 and a variety of library dependencies have been upg
 
 === SolrJ
 * The `Http2SolrClient` classes powered by the Jetty Http library are no longer marked as experimental.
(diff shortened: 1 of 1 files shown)


janhoy (commented):
We normally don’t include Jira numbers in upgrade notes.

The auto prune of inactive cores could be under an “Other” header? “Admin” makes you think of Admin UI.

gus-asf:
> We normally don’t include Jira numbers in upgrade notes.
> 
> The auto prune of inactive cores could be under an “Other” header? “Admin” makes you think of Admin UI.

Ok other... as for Jira's any idea why not? I want to get on with it so I'll take them out, but I wonder why...

PR #2430, opened 2024-04-30
SOLR-17260 ReleaseWizard should remove old minor versions from Docker Hub

janhoy:
https://issues.apache.org/jira/browse/SOLR-17260

This PR implements the suggested policy of deleting version `N.M-2` from hub. This means that when we release 9.7, 9.5 will be marked as "unsupported" in docker hub. That is, it will still be available for pull but not advertised as "supported".

### DIFF
--- a/dev-tools/scripts/releaseWizard.py
+++ b/dev-tools/scripts/releaseWizard.py
@@ -110,7 +110,7 @@ def expand_jinja(text, vars=None):
         'get_next_version': state.get_next_version(),
         'current_git_rev': state.get_current_git_rev(),
         'keys_downloaded': keys_downloaded(),
-        'get_docker_version_to_remove': state.get_docker_version_to_remove(),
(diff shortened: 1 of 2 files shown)


madrob:
Should we remove 9.x.0 when we release 9.x.1?

janhoy:
> Should we remove 9.x.0 when we release 9.x.1?

That already happens automatically today since the Dockerfile is `x.y/Dockerfile` in the repo.

So you could argue that someone using tag 9.4.0 will not get new JRE bugfix updates since the only supported 9.4 version is 9.4.1. But if they use tag 9.4 then they will get the latest. However there’s a best practice of pinning exact versions to avoid unintended upgrades in prod during e.g a normal restart; even bugfix releases can have regressions.

HoustonPutman (approved):
Looks good to me!

PR #2431, opened 2024-04-30
SOLR-17261: Allow cores to load without a timeout

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17261

ExecutorUtil once again has the ability to wait for an ExecutorService to terminate without a timeout.

I looked through the other uses of `ExecutorUtil.shutdownAndAwaitTermination(coreLoadExecutor)`, however almost all the other usages are in a `shutdown()` or `close()` method, which should be time-boxed. I think these two usages are the only two currently that really shouldn't be timeboxed. However, I've added java docs to help developers chose the correct one in the future.

I will probably include this in a 9.6.1 release as well.

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/common/util/ExecutorUtil.java
+++ b/solr/solrj/src/java/org/apache/solr/common/util/ExecutorUtil.java
@@ -181,26 +181,16 @@ public class ExecutorUtil {
    */
   public static void awaitTerminationForever(ExecutorService pool) {
     boolean shutdown = false;
-    while (!shutdown) {
-      try {
-        try {
-          // Wait a while for existing tasks to terminate
-          shutdown = pool.awaitTermination(60, TimeUnit.SECONDS);
-        } catch (InterruptedException ie) {
-          // Force cancel if current thread also interrupted
-          pool.shutdownNow();
-          // Preserve interrupt status
-          Thread.currentThread().interrupt();
-          if (!pool.awaitTermination(60, TimeUnit.SECONDS)) {
-            // there was an issue shutting down the threads, so throw a RuntimeException,
-            // using the original InterruptedException as a cause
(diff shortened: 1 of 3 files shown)


anshumg (approved):
Thanks @HoustonPutman

solr/solrj/src/java/org/apache/solr/common/util/ExecutorUtil.java
```diff
@@ -143,6 +173,37 @@ static void awaitTermination(ExecutorService pool, long timeout, TimeUnit unit)
...
+    while (!shutdown) {
+      try {
+        try {
+          // Wait a while for existing tasks to terminate
+          shutdown = pool.awaitTermination(60, TimeUnit.SECONDS);
+        } catch (InterruptedException ie) {
+          // Force cancel if current thread also interrupted
+          pool.shutdownNow();
+          // Preserve interrupt status
+          Thread.currentThread().interrupt();
```
dsmiley, line 193: Arguably, if we've given up waiting for awaiting termination, then the interruption has done its job and we don't need to keep the flag.

solr/solrj/src/java/org/apache/solr/common/util/ExecutorUtil.java
```diff
@@ -143,6 +173,37 @@ static void awaitTermination(ExecutorService pool, long timeout, TimeUnit unit)
...
+          if (!pool.awaitTermination(60, TimeUnit.SECONDS)) {
```
dsmiley, line 194: Wouldn't this always fail if you "preserve interrupt status" the line before?

HoustonPutman:
I've simplified the logic, definitely went overboard on the first attempt.

dsmiley (approved):

PR #2435, opened 2024-05-01
 SOLR-17263: HttpJdkSolrClient doesn't encode curly braces etc (follow-up PR)

jdyer1:
This is a follow up to [PR ](https://github.com/apache/solr/pull/2433).  

Using @andywebb1975  fix for `HttpJdkSolrClient` this adds a unit test, adding `fq="{!terms f=myfield}value1,value2` to both GET, POST and PUT requests both for `HttpJdkSolrClient` and `Http2SolrClient`.  Prior to the fix, the test addition fails for GET requests on `HttpJdkSolrClient`.

I found the modifications to POST and PUT unnecessary, but the GET change is needed.

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpJdkSolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpJdkSolrClient.java
@@ -238,7 +238,7 @@ public class HttpJdkSolrClient extends HttpSolrClientBase {
     validateGetRequest(solrRequest);
     reqb.GET();
     decorateRequest(reqb, solrRequest);
-    reqb.uri(new URI(url + "?" + queryParams));
+    reqb.uri(new URI(url + queryParams.toQueryString()));
     return new PreparedRequest(reqb, null);
   }
 
--- a/solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java
+++ b/solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java
@@ -131,6 +131,7 @@ public class Http2SolrClientTest extends HttpSolrClientTestBase {
     String url = getBaseUrl() + DEBUG_SERVLET_PATH;
     SolrQuery q = new SolrQuery("foo");
     q.setParam("a", "\u1234");
+    q.setParam("fq", "{!terms f=myfield}value1,value2");
     Http2SolrClient.Builder b =
         new Http2SolrClient.Builder(url).withDefaultCollection(DEFAULT_CORE);
     if (rp != null) {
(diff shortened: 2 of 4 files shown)


solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpJdkSolrClient.java
```diff
@@ -238,7 +238,7 @@ private PreparedRequest prepareGet(
     validateGetRequest(solrRequest);
     reqb.GET();
     decorateRequest(reqb, solrRequest);
-    reqb.uri(new URI(url + "?" + queryParams));
+    reqb.uri(new URI(url + queryParams.toQueryString()));
```
andywebb1975, line 241: I've had a closer look. I should say I'm never looked "under the hood" of the clients before and also I'm not familiar with all their use-cases.

I hadn't clocked that `toQueryString()` adds its own `?` - so yes, this line is better than my version.

I do think line 316 should change this way too, as `toQueryString()` does a more comprehensive job of encoding strings than `toString()` and it explicitly uses UTF-8 encoding, while the description of `toString()` indicates it's just intended for logging purposes.

My PR changed the method used at line 305 too, which would have injected an unwanted `?` - but I'm not sure that using `toString()` is right either for the same reasons as above. I think it'd be worth revisiting this block in general - I'm not clear on the intent of the code to be able to say how it should change, but my understanding is that line 302 makes a second reference to the same object, which doesn't seem right to me. I may be wrong!

dsmiley (commented):
Can we embrace randomized testing by indexing a random chunk of text and then query for it using the raw QP?  Over time, this will show that escaping is handled properly.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpJdkSolrClient.java
jdyer1: I had removed the change from lines 305 and 316 as adding the `fq` parameter you supplied the unit test only seemed to show that `GET` was broken.  Likes 305 & 316 are building `POST` or `PUT` requests.  On the other hand, so long as the extra ? is removed, adding `toQueryString()` there doesn't seem to break anything either.  Yet again, I do not like "fixing" things if I cannot first demonstrate it is broken.

andywebb1975: hi James - I've explored this some more. I'm even more convinced we should use `toQueryString()` now.

https://github.com/apache/solr/pull/2433 now has a series of commits where I've added test cases and other checks to demonstrate the value of the changes.

The current tests don't exercise the URI encoding enough to show the difference in behaviour - but adding in another `U+1234` and some curly quotes triggers failures for the PUT/POST URL generation when using `toString()` - I saw the same exception as we had previously for GETs.

I've also seen that the request body content length becomes inconsistent with that of the provided query string unless it's pre-encoded by `toQueryString()` - when we use `toString()` it's relying on `HttpRequest.BodyPublisher` re-encoding it. As I've noted in the code comment I wouldn't suggest doing that check for real, but I'll see if I can convert it to a test.

Hope this helps!
Andy

dsmiley:
I merged the first PR to main, branch_9x, and branch_9_6.  Shall this PR be merged too?  There are conflicts.  Are we all "Watching" the parent JIRA issue?  I commented there yesterday.

andywebb1975:
> I merged the first PR to main, branch_9x, and branch_9_6. Shall this PR be merged too? There are conflicts. Are we all "Watching" the parent JIRA issue? I commented there yesterday.

Thanks @dsmiley - no, this one's redundant now.

PR #2436, opened 2024-05-01
SOLR-17265: Fix randomized PRS testing

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17265

This fixes the test errors introduced by https://github.com/apache/solr/pull/2230

I'm not really certain what the previous logic was supposed to be doing, but `isPRS()` practically returned a random true or false every time it was called, since the `"use.per-replica"` sysProp was never set.

I've simplified the logic considerably. Before a class is run `"solr.prs.default"` is set to either `true` or `false` (either randomly, or determined by the Class annotation). Then ever call to `isPRS()` will read that sysProp to determine if the default is PRS or not.

### DIFF
--- a/solr/test-framework/src/java/org/apache/solr/cloud/AbstractBasicDistributedZk2TestBase.java
+++ b/solr/test-framework/src/java/org/apache/solr/cloud/AbstractBasicDistributedZk2TestBase.java
@@ -17,14 +17,12 @@
 package org.apache.solr.cloud;
 
 import static org.apache.solr.cloud.SolrCloudTestCase.configurePrsDefault;
-import static org.apache.solr.cloud.api.collections.CreateCollectionCmd.PRS_DEFAULT_PROP;
 
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.concurrent.TimeUnit;
 import org.apache.lucene.tests.mockfile.FilterPath;
 import org.apache.lucene.tests.util.LuceneTestCase;
-import org.apache.lucene.tests.util.TestRuleRestoreSystemProperties;
 import org.apache.solr.client.solrj.SolrClient;
 import org.apache.solr.client.solrj.SolrQuery;
 import org.apache.solr.client.solrj.SolrServerException;
@@ -42,9 +40,7 @@ import org.apache.solr.embedded.JettySolrRunner;
 import org.apache.solr.handler.BackupStatusChecker;
 import org.apache.solr.handler.ReplicationHandler;
 import org.junit.BeforeClass;
-import org.junit.ClassRule;
 import org.junit.Test;
-import org.junit.rules.TestRule;
 
 /**
  * This test simply does a bunch of basic things in solrcloud mode and asserts things work as
@@ -57,10 +53,6 @@ public abstract class AbstractBasicDistributedZk2TestBase extends AbstractFullDi
   private static final String SHARD1 = "shard1";
   private static final String ONE_NODE_COLLECTION = "onenodecollection";
   private final boolean onlyLeaderIndexes = random().nextBoolean();
-
-  @ClassRule
-  public static TestRule syspropRestore =
-      new TestRuleRestoreSystemProperties(NUMERIC_DOCVALUES_SYSPROP, PRS_DEFAULT_PROP);
 
   public AbstractBasicDistributedZk2TestBase() {
     super();
--- a/solr/test-framework/src/java/org/apache/solr/cloud/AbstractBasicDistributedZk2TestBase.java
+++ b/solr/test-framework/src/java/org/apache/solr/cloud/AbstractBasicDistributedZk2TestBase.java
@@ -16,14 +16,15 @@
  */
 package org.apache.solr.cloud;
 
-import static org.apache.solr.cloud.SolrCloudTestCase.setPrsDefault;
-import static org.apache.solr.cloud.SolrCloudTestCase.unsetPrsDefault;
+import static org.apache.solr.cloud.SolrCloudTestCase.configurePrsDefault;
+import static org.apache.solr.cloud.api.collections.CreateCollectionCmd.PRS_DEFAULT_PROP;
 
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.concurrent.TimeUnit;
 import org.apache.lucene.tests.mockfile.FilterPath;
 import org.apache.lucene.tests.util.LuceneTestCase;
+import org.apache.lucene.tests.util.TestRuleRestoreSystemProperties;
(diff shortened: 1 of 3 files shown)


solr/test-framework/src/java/org/apache/solr/cloud/SolrCloudTestCase.java
```diff
@@ -83,7 +83,7 @@ public class SolrCloudTestCase extends SolrTestCaseJ4 {
 
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
-  public static final String PRS_DEFAULT_PROP = System.getProperty("use.per-replica", null);
+  public static final String PRS_DEFAULT_PROP = "solr.prs.default";
```
dsmiley, line 86: Out of scope for now but but I'd like such a setting to exist Solr side officially

solr/test-framework/src/java/org/apache/solr/cloud/SolrCloudTestCase.java
```diff
@@ -97,11 +97,9 @@ protected static SolrZkClient zkClient() {
     return cluster.getZkStateReader().getZkClient();
   }
 
-  /** if the system property is not specified, use a random value */
+  /** if the system property is not specified, default to false. The SystemProperty will be set in a beforeClass method. */
   public static boolean isPRS() {
-    return PRS_DEFAULT_PROP == null
-        ? random().nextBoolean()
```
dsmiley, line 103: oh wow I see how this was the problem, leading to too-little control in consistency over wether it's enabled or not in practice

solr/test-framework/src/java/org/apache/solr/cloud/SolrCloudTestCase.java
```diff
@@ -97,11 +97,9 @@ protected static SolrZkClient zkClient() {
...
-        : Boolean.parseBoolean(PRS_DEFAULT_PROP);
+    return Boolean.parseBoolean(System.getProperty(PRS_DEFAULT_PROP, "false"));
```
dsmiley, line 102: Can EnvUtils be used here?

solr/test-framework/src/java/org/apache/solr/cloud/SolrCloudTestCase.java
```diff
@@ -140,29 +138,18 @@ public static void shutdownCluster() throws Exception {
...
   public static void setPrsDefault() {
     Class<?> target = RandomizedTest.getContext().getTargetClass();
-    if (target != null && target.isAnnotationPresent(NoPrs.class)) return;
-    if (isPRS()) {
-      System.setProperty("solr.prs.default", "true");
+    boolean prsDefault;
+    if (target != null && target.isAnnotationPresent(NoPrs.class)) {
+      prsDefault = false;
+    } else {
+      prsDefault = random().nextBoolean();
```
dsmiley, line 145: shouldn't this **read** a sys prop so we can choose it for the build?

gus-asf (approved):
+1 to this and David's comments

solr/test-framework/src/java/org/apache/solr/cloud/SolrCloudTestCase.java
HoustonPutman: Put it in Solr where it is used.

solr/test-framework/src/java/org/apache/solr/cloud/SolrCloudTestCase.java
HoustonPutman: Used it, in all places, but really in the tests, it will only be cared about in the `setPrsDefault()` since we overwrite the sysProp always.

solr/test-framework/src/java/org/apache/solr/cloud/SolrCloudTestCase.java
HoustonPutman: Done, and added another sysProp to save the user-provided value that we might be overwriting because of Class annotations

solr/test-framework/src/java/org/apache/solr/cloud/SolrCloudTestCase.java
```diff
@@ -140,28 +143,29 @@ public static void shutdownCluster() throws Exception {
...
+    if (target != null && target.isAnnotationPresent(NoPrs.class)) {
+      if (existingPrsDefault != null) {
+        System.setProperty(SAVED_PRS_DEFAULT_PROP, existingPrsDefault);
+      }
+      newPrsDefault = "false";
+    } else if (existingPrsDefault != null) {
+      newPrsDefault = existingPrsDefault;
+    } else {
+      newPrsDefault = Boolean.toString(random().nextBoolean());
     }
```
dsmiley, line 157: Can't we instead rely on TestRuleRestoreSystemProperties, thus don't need to restore whatever the pre-existing sys prop (set at CLI?) is?

dsmiley (commented):
I'd like us to approach this in a general/patterned way; this isn't the only setting to do this for.  Another is the distributed cluster processing booleans, which is already handled.  See org.apache.solr.cloud.SolrCloudTestCase#configureCluster and MiniSolrCloudCluster.build

I feel both should be configured similarly (at the same places mostly).

HoustonPutman:
Thanks for the info on those. I've standardized both the reseting of sysProps as well as where the option is set from.

solr/test-framework/src/java/org/apache/solr/cloud/AbstractBasicDistributedZk2TestBase.java
```diff
@@ -56,6 +58,10 @@ public abstract class AbstractBasicDistributedZk2TestBase extends AbstractFullDi
   private static final String ONE_NODE_COLLECTION = "onenodecollection";
   private final boolean onlyLeaderIndexes = random().nextBoolean();
 
+  @ClassRule
+  public static TestRule syspropRestore =
+      new TestRuleRestoreSystemProperties(NUMERIC_DOCVALUES_SYSPROP, PRS_DEFAULT_PROP);
```
dsmiley, line 63: AFAIK, this shouldn't be necessary because SolrTestCase (root of all Solr tests) already declares SystemPropertiesRestoreRule as a ClassRule and also STCJ4 has a test rule for this as well for test/method level restoration.  Even LuceneTestCase adds a ClassRule for TestRuleRestoreSystemProperties, thus STC's is maybe redundant.

HoustonPutman: Ahhh that makes sense. I'll remove the custom ones.

dsmiley (approved):
Looks good, assuming tests pass.

PR #2437, opened 2024-05-02
SOLR-17088: Fix custom solr.xml usage in tests with extra jettys

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17088

### DIFF
--- a/solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
+++ b/solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
@@ -162,6 +162,7 @@ public class MiniSolrCloudCluster {
   private final Path baseDir;
   private CloudSolrClient solrClient;
   private final JettyConfig jettyConfig;
+  private final String solrXml;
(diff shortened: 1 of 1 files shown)


solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
```diff
@@ -276,6 +277,7 @@ public MiniSolrCloudCluster(
     Objects.requireNonNull(securityJson);
     this.baseDir = Objects.requireNonNull(baseDir);
     this.jettyConfig = Objects.requireNonNull(jettyConfig);
+    this.solrXml = solrXml == null ? DEFAULT_CLOUD_SOLR_XML : solrXml;
```
cpoerschke, line 280: Wondering if the similar logic at lines 491-493/492-494 can be removed now then?

HoustonPutman: Yeah, I didn't see the other usages of this method with null. So I moved the defaulting here instead. Thanks for the review!

PR #2438, opened 2024-05-02
SOLR-17269: Do not publish synthetic solr core (of Coordinator node) to ZK

patsonluk:
https://issues.apache.org/jira/browse/SOLR-17269

# Description

The synthetic collection/replica introduced in Solr 9 registers such collection to zookeeper. This might cause issues with other tools which monitor solr collections via zk and they might be mistaken the synthetic collection as a real collection.

# Solution
We are introducing a new class `SyntheticSolrCore` which extends `SolrCore`, `SyntheticSolrCore` is only used in Coordinator node to support a subset of `SolrCore` functionalities required by Coordinator operations such as aggregating and writing out response and providing configset info. There will be one instance of `SyntheticSolrCore` per config set.

Such synthetic core is only registered to the `CoreContainer` but is no longer registering itself to zookeeper. 

Take note that to minimize code changes, there will still be snapshot/index data directory written to the Coordinator node. We could have overridden `initSnapshotMetaDataManager` and `initIndex` to be no-op to avoid creating them, however, there is currently other code (and possibly future code too) that might rely on the existence of those fields. It is safer to just keep them as is. 

The data folder for the synthetic core would NOT be loaded on Coordinator node start-up as there's no `core.properties` in such directory. This is because instead of going through `CoreContainer#create` for such synthetic core, it's using `SyntheticSolrCore#createAndRegisterCore`, which bypass `core.properties` creation and zk registration

# Tests

The existing `TestCoordinatorRole` has already provided good coverage of various scenarios. Take note that we have removed the test code that verifies the existence of cores from the synthetic collection since zk no longer keeps track of synthetic collections. Instead we added a check in `TestCoordinatorRole#testSimple` to ensure that the synthetic collection no longer exists in cluster state.

# Checklist


solr/core/src/java/org/apache/solr/core/ConfigSetService.java
```diff
@@ -391,6 +402,21 @@ protected NamedList<Object> loadConfigSetFlags(SolrResourceLoader loader) throws
...
+   *
+   * <p>By default, this will just call {@link
+   * ConfigSetService#createConfigSetService(CoreContainer)}. Child implementation might override
+   * this to make use of the configSetName directly
+   *
+   * @param cd the core's CoreDescriptor
+   * @param configSetName an optional config set name
+   * @return a SolrResourceLoader
+   */
+  protected SolrResourceLoader createCoreResourceLoader(CoreDescriptor cd, String configSetName) {
```
dsmiley, line 416: Since the configSetName is also on the CoreDescriptor, isn't this API needless?  Same elsewhere (loadConfigSet)

solr/core/src/java/org/apache/solr/servlet/CoordinatorHttpSolrCall.java
```diff
@@ -314,6 +182,19 @@ public CloudDescriptor getCloudDescriptor() {
     };
   }
 
+  /**
+   * Overrides the MDC context as the core set was synthetic core, which does not reflect the
+   * collection being operated on
+   */
+  private static void setMDCLoggingContext(String collectionName) {
```
dsmiley, line 189: We use Google Java code standards, thus use "Mdc" capitalization.  Plenty of old code did a mis-mash of things, MDCLoggingContext being one.

dsmiley (commented):
Nice to see a reduction in code in some places

dsmiley:
I noticed you tagged nobody for review; it's good to think of one two pertinent people.

solr/core/src/test/org/apache/solr/search/TestCoordinatorRole.java
```diff
@@ -105,14 +102,11 @@ public void testSimple() throws Exception {
...
       DocCollection collection =
           cluster.getSolrClient().getClusterStateProvider().getCollection(SYNTHETIC_COLLECTION);
-      assertNotNull(collection);
-
-      Set<String> expectedNodes = new HashSet<>();
-      expectedNodes.add(coordinatorJetty.getNodeName());
-      collection.forEachReplica((s, replica) -> expectedNodes.remove(replica.getNodeName()));
-      assertTrue(expectedNodes.isEmpty());
+      // this should be empty as synthetic collection does not register with ZK
+      assertNull(collection);
```
noblepaul, line 109: I think we should invoke the core API command to confirm a synthetic core is created

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
patsonluk: Yes the choice I took was problematic! 😅 

The ultimate goal is to minimize code change and yet avoid calling [this part of the code](https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/cloud/ZkConfigSetService.java#L77-L100) for the synthetic core, so we wouldn't be registering against ZK. 

Since I saw the quoted code above "reads the config name from cluster state and set it against the CoreDescriptor", so I thought maybe a new "explicit" configset can indicate that we can "skip" those logic for `ZkConfigSetService`.

But it didn't turn out well if I confused other devs 😞 ...

Adding any flags/override method in `ConfigSetService` is likely not great since other non ZK implementation shouldn't really care. So im proposing to add a new class `SythethticCoreDescriptor` which extends `CoreDescriptor`, and that `ZkConfigSetService#createCoreResourceLoader` could do a special case (`instanceof  SythethticCoreDescriptor`, not great...but at least it's all contained in Zk related class) to bypass the zk logic.

Going to commit and push the change, please do lemme know if there are better alternatives!

solr/core/src/test/org/apache/solr/search/TestCoordinatorRole.java
patsonluk: Good call! And also we should confirm that there's no node created for the synthetic core.

Thank you for the suggestion!

patsonluk:
@dsmiley @noblepaul Many thanks for the helpful code review!

I have pushed some changes based on the suggestions, would you mind to review again? 🙏🏼

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
dsmiley: > The ultimate goal is to minimize code change and yet avoid calling [this part of the code](https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/cloud/ZkConfigSetService.java#L77-L100) for the synthetic core, so we wouldn't be registering against ZK.

Why?  It appears it'd have no ultimate effect -- it ought to resolve the same config for the same collection.  Oh, is it the same collection for this feature?

I don't love subclassing CoreDescriptor to be honest; you could just add a generic property in the core descriptor if you needed to flag one like "synthetic".  Nonetheless overall I'm unsure why loading the ConfigSet needs to be different here.

Also if for some reason we felt it useful, we could change ZkConfigSetServer.loadConfigSet such that an existing configSet name in the CD will be used, and thus avoid resolving that.

patsonluk: @dsmiley thank you for the review again!

>Why? It appears it'd have no ultimate effect -- it ought to resolve the same config for the same collection. Oh, is it the same collection for this feature?

We want to avoid registering the synthetic collection with ZK, since it creates various issues for our various tooling which watches the ZK for collection changes. We have to do various checks (and it's very easy to forget for new code) to skip the synthetic collection, which is not ideal. 

>I don't love subclassing CoreDescriptor to be honest; you could just add a generic property in the core descriptor if you needed to flag one like "synthetic". Nonetheless overall I'm unsure why loading the ConfigSet needs to be different here.

>Also if for some reason we felt it useful, we could change ZkConfigSetServer.loadConfigSet such that an existing configSet name in the CD will be used, and thus avoid resolving that.

Got it! If it's more preferable to add property to `CoreDescriptor` instead of subclassing, we could totally do that! I thought we want to minimize changes to existing class hence didn't touch `CoreDescriptor` 😓 

Currently `ZkConfigSetService#createCoreResourceLoader` does 3 things as far as I can tell:
1. Auto create collection in ZK if it's not already here
2. Set the configset read from ZK (off the collection from cluster state) to CD. Ignore the pre-existing on in CD
3. Create a `ZkSolrResourceLoader` with access to the resources under the configset in ZK
 
Honestly, I couldn't quite tell the purposes of 1 and 2 (probably not the use cases I'm aware of, I am sure you will be more knowledgeable!) . For synthetic core, we want to avoid both since the for 1., we do not want to register the synthetic collection to ZK and for 2., since the collection is not in ZK it would fail loading such collection from cluster stateu. 

We do want 3 tho as the synthetic core is there to serve resources from the ZK configset.


Please let me know if it all makes sense? If so, I will proceed to add property to `CoreDescriptor` 😊

dsmiley: We should remove "1. Auto create collection in ZK if it's not already here"; this refers to legacy behavior.  Or gate this with "bootstrap_conf" system property being specified or something like that.  Heck, I wonder if this chunk of code actually works.... like if you did a little experiment locally, could you get the collection to appear merely by having a core descriptor for a non-existent collection?  I wonder if it's even tested (easy to check by trial and error).

Hopefully there will be no need to designate this CoreDescriptor as special.

There's legacy stuff you are seeing in Solr.  We don't need to support it forever and ever!  Certainly not if it's getting in our way and causing us to contemplate adding otherwise needless complexity.

patsonluk: So i tried removing 
```
String configSetName =
          zkController.getClusterState().getCollection(cd.getCollectionName()).getConfigName();
      cd.setConfigSet(configSetName);
```

Then the core loading starts failing. With some investigation:
1. `CoreDescriptor` instantiation does not set the `configSet` in the `coreProperties` via both admin create path and startup core discovery path. Those paths would set `collection.configName` instead. 
2. Therefore without setting it as a part of `ZkConfigSetService#createCoreResourceLoader`, `cd.getConfigSet` would give null, and could trigger error when calling `new ZkSolrResourceLoader(
        cd.getInstanceDir(), cd.getConfigSet(), parentLoader.getClassLoader(), zkController)` (or anywhere else relies on `CoreDescriptor#getgetConfigSet`

I'm not 100% sure if calling `cd.setConfigSet(configSetName)` is a good thing to do within `createCoreResourceLoader` since it appears to be an unexpected side effect. However, to focus on synthetic core alone. I pushed a commit to do a null check with a side note
```
    // Currently, cd.getConfigSet() is always null. Except that it's explicitly set by
    // {@link org.apache.solr.core.SyntheticSolrCore}.
    // Should we consider setting it for all cores as a part of CoreDescriptor creation/loading
    // process?
    if (cd.getConfigSet() == null) {
      String configSetName =
          zkController.getClusterState().getCollection(cd.getCollectionName()).getConfigName();
      cd.setConfigSet(configSetName);
    }
```


Any thoughts? @dsmiley  😊

solr/core/src/java/org/apache/solr/cloud/ZkConfigSetService.java
```diff
@@ -71,36 +68,21 @@ public ZkConfigSetService(SolrZkClient zkClient) {
...
-    } catch (InterruptedException e) {
-      Thread.currentThread().interrupt();
-      throw new ZooKeeperException(
-          SolrException.ErrorCode.SERVER_ERROR, "Interrupted auto-creating collection", e);
-    } catch (KeeperException e) {
-      throw new ZooKeeperException(
-          SolrException.ErrorCode.SERVER_ERROR, "Failure auto-creating collection", e);
-    }
-
     // The configSet is read from ZK and populated.  Ignore CD's pre-existing configSet; only
```
dsmiley, line 71: > Ignore CD's pre-existing configSet

Wrong, now.

solr/core/src/java/org/apache/solr/core/SyntheticSolrCore.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+public class SyntheticSolrCore extends SolrCore {
+  public SyntheticSolrCore(CoreContainer coreContainer, CoreDescriptor cd, ConfigSet configSet) {
+    super(coreContainer, cd, configSet);
+  }
+
+  public static SyntheticSolrCore createAndRegisterCore(
+      CoreContainer coreContainer, String syntheticCoreName, String configSetName) {
+    Map<String, String> coreProps = new HashMap<>();
+    coreProps.put(CoreAdminParams.CORE_NODE_NAME, coreContainer.getHostName());
```
dsmiley, line 44: is this necessary; why is the host name in the core descriptor as it's not really related to the core (it's at a different level)

solr/core/src/java/org/apache/solr/core/SyntheticSolrCore.java
```diff
@@ -0,0 +1,83 @@
...
+ * <p>There should only be one instance of SyntheticSolrCore per configset
+ */
+public class SyntheticSolrCore extends SolrCore {
+  public SyntheticSolrCore(CoreContainer coreContainer, CoreDescriptor cd, ConfigSet configSet) {
+    super(coreContainer, cd, configSet);
+  }
+
+  public static SyntheticSolrCore createAndRegisterCore(
+      CoreContainer coreContainer, String syntheticCoreName, String configSetName) {
+    Map<String, String> coreProps = new HashMap<>();
```
dsmiley, line 43: Would `Map.of` work?

solr/core/src/java/org/apache/solr/core/SyntheticSolrCore.java
```diff
@@ -0,0 +1,83 @@
...
+    coreProps.put(CoreAdminParams.CORE_NODE_NAME, coreContainer.getHostName());
+    coreProps.put(CoreAdminParams.COLLECTION, syntheticCoreName);
+
+    CoreDescriptor syntheticCoreDescriptor =
+        new CoreDescriptor(
+            syntheticCoreName,
+            Paths.get(coreContainer.getSolrHome() + "/" + syntheticCoreName),
```
dsmiley, line 50: Path & String slash concatenation together is a sad thing; misses a value proposition of the Path API.   Someday in main branch, ought to finally switch getSolrHome to return a Path but nonetheless you can do Paths.of(..).resolve(syntheticCoreName) for example.

solr/core/src/java/org/apache/solr/core/SyntheticSolrCore.java
```diff
@@ -0,0 +1,83 @@
...
+  protected void bufferUpdatesIfConstructing(CoreDescriptor coreDescriptor) {
+    // no updates to SyntheticSolrCore
+  }
+
+  @Override
+  protected RestManager initRestManager() throws SolrException {
+    // returns an initialized RestManager. As init routines requires reading configname of the
+    // core's collection from ZK
+    // which synthetic core is not registered in ZK.
+    // We do not expect RestManager ops on Coordinator Nodes
```
dsmiley, line 75: Rewrap.

Glad to see this kind of comment.  Hmm; I wonder if it's actually prevented?  Like how does this overall feature (which I confess knowing nothing about) actually prevent these cores from being used for _anything_ except the one thing it needs to exist for -- distributed-search?  e.g. maybe. no handlers should be loaded except those extending SearchHandler?

solr/core/src/java/org/apache/solr/core/SyntheticSolrCore.java
```diff
@@ -0,0 +1,83 @@
...
+    return new RestManager();
+  }
+
+  @Override
+  public void close() {
```
dsmiley, line 80: remove; right?

solr/core/src/java/org/apache/solr/servlet/CoordinatorHttpSolrCall.java
```diff
@@ -88,198 +81,70 @@ public static SolrCore getCore(
...
-                + " collection: "
-                + Utils.toJSONString(rsp.getValues()));
-      }
-    } catch (SolrException e) {
-      throw e;
-
-    } catch (Exception e) {
-      throw new SolrException(SolrException.ErrorCode.SERVER_ERROR, e);
-    }
+  public static String getSyntheticCoreName(String configName) {
```
dsmiley, line 146: From a pure type signature perspective (thus erasure of "configName" name), this signature could be confusing.  The string I'd expect to give such a method would be the collection name but no it's not that.  Maybe "FromConfig"  suffix?

dsmiley (commented):
Looking good now; I like your change.

Can a SyntheticSolrCore be prevented from registering itself with CoreContainer?  I would imagine so since this functionality is inserted at the HttpSolrCall level thus the SolrCore could come from wherever.  It might be less hacky to prevent weird/unusual SolrCores from presenting themselves in the CoreContainer to other functionality (who knows) that is expecting normal cores.  I don't have a strong opinion on this; just sharing.

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
dsmiley: I could understand that it's questionable for the ConfigSetService to be responsible for resolving what the proper named config even ought to be.  But it works and there's not an obvious other location, and it empowers this ConfigSetService abstraction to do interesting things if needed, so I suggest let it be.

dsmiley: Thanks so much for doing that manual check and being able to remove old legacy code :-)

solr/core/src/java/org/apache/solr/core/SyntheticSolrCore.java
dsmiley: `Path.of(coreContainer.getSolrHome(), "solrconfig.xml")`

patsonluk:
> Can a SyntheticSolrCore be prevented from registering itself with CoreContainer? I would imagine so since this functionality is inserted at the HttpSolrCall level thus the SolrCore could come from wherever. It might be less hacky to prevent weird/unusual SolrCores from presenting themselves in the CoreContainer to other functionality (who knows) that is expecting normal cores. I don't have a strong opinion on this; just sharing.

I think that's a very good question! AFAIK, we probably really only need `SyntheticSolrCore` be present when `CoordinatorHttpSolrCall` is doing its special logic, which means in theory, it does not have to live within the`CoreContainer`/`SolrCores`. And it will be much cleaner w/o that!

I will investigate a bit and see if there are any dependencies that prevent us from doing that 😊

solr/core/src/java/org/apache/solr/core/SyntheticSolrCore.java
patsonluk: AFAIK, this special core is restricted by the `getCoreByCollection` of `CoordinatorHttpSolrCall`/`CoordinatorV2HttpSolrCall` which only accept suffix [`/select`](https://github.com/apache/solr/blob/a42c605fb916439222a086356f368f02cf80304a/solr/core/src/java/org/apache/solr/servlet/CoordinatorHttpSolrCall.java#L82).

solr/core/src/java/org/apache/solr/core/SyntheticSolrCore.java
patsonluk: sure! removing since it only calls super.close()

solr/core/src/java/org/apache/solr/servlet/CoordinatorHttpSolrCall.java
patsonluk: Sure! Using `getSyntheticCollectionNameFromConfig` and `getSyntheticCoreNameFromConfig` instead :)

patsonluk:
@dsmiley I have committed changes based on your feedback last week! And have replied to several of your comments.

Would you mind to go over them again and let me know your thoughts? Many many thanks! 🙏🏼

dsmiley:
Looks great.  Might you suggest a succinct CHANGES.txt entry for 9.7?  CHANGES.txt is written for users to read/understand.

Also can you propose a commit message here, which often adds detail.   Like removal of that legacy logic.

patsonluk:
> Looks great. Might you suggest a succinct CHANGES.txt entry for 9.7? CHANGES.txt is written for users to read/understand.

Many thanks for all the reviews! I will add an entry to CHANGES.txt under 9.7!

 
> Also can you propose a commit message here, which often adds detail. Like removal of that legacy logic.
Do you mean the commit message for the CHANGE.txt change? Or is it something else? :)

dsmiley:
Yes to both.  It's helpful if you provide this info as it affords peer review and it makes it easier for me :-)

solr/CHANGES.txt
```diff
@@ -132,6 +132,8 @@ Optimizations
 
 * SOLR-17099: Do not return spurious tags when fetching metrics from a Solr node to another. (Pierre Salagnac)
 
+* SOLR-17269: Simplify cores of Coordinator node to avoid unnecessary registrations on Zookeeper (Patson Luk)
```
dsmiley, line 135: Maybe:  Prevent the "Coordinator node" feature from registering synthetic cores in ZooKeeper.

I worded it this way as a non-user of this trying to understand it.

dsmiley (commented):
I plan to merge this soon... perhaps this weekend.  Albeit I'm on vacation so we'll see.

PR #2443, opened 2024-05-06
SOLR-17271: PerReplicaState: Shard leader elections still impact state.json

noblepaul:
pls review


dsmiley (commented):
It's nice to see a test here!

Are there any watchers of a collection that want to know about a change that is in PRS (e.g. a change in leadership) that will now not get notified?

dsmiley:
When creating a PR, instead of deleting absolutely everything in the template, please leave the first tiny bit with a template JIRA URL then edit the URL.  It's very convenient to click both ways between JIRA and PR.

PR #2447, opened 2024-05-08
SOLR-17284: Remove deprecated BlobRepository

epugh:
https://issues.apache.org/jira/browse/SOLR-17284

# Description

This approach to managing shared jar files was deprecated in the 8.x line and tagged for removal in Solr 9.   Removing deprecated code.

# Solution

Removed the code, and tested the Schema designer, which relies on .system existing, still works.

# Tests

# Checklist


epugh:
This started with me spelunking around the code base and chasing some small typos, and then realized that this code was deprecated and we should remove it!

dsmiley:
I see BlobHandler and supporting infrastructure still.  It's not explicitly deprecated.

Should the Schema Designer API use the "File Store" instead?

Any way, I suppose this PR, scoped to what it is, is low hanging fruit to remove.

CC @noblepaul

gus-asf:
I'm not so keen on the loss of https://issues.apache.org/jira/browse/SOLR-8349 along with this. It was only tied to the blob repository because that seemed to be how folks wanted me to load resources at the time.

epugh:
> I see BlobHandler and supporting infrastructure still. It's not explicitly deprecated.
> 
> Should the Schema Designer API use the "File Store" instead?
> 
> Any way, I suppose this PR, scoped to what it is, is low hanging fruit to remove.
> 
> CC @noblepaul

I was actually sort of wondering how SchemaDesigner was working after i made my changes, I tested it manually and it all worked.  Somehow I never stumbled over `BlobHandler` while spelunking in IntelliJ!   I will poke around some more, and maybe it's own bit of work...   (I also note that `.system` doesn't show up in admin ui or have much docs..  likewise blobhandler doesn't appear in ref guide as far as I can tell).

epugh:
> I'm not so keen on the loss of https://issues.apache.org/jira/browse/SOLR-8349 along with this. It was only tied to the blob repository because that seemed to be how folks wanted me to load resources at the time.

Can you share more about how this is being used?  I took a quick looksee at SOLR-8349 and it points at SOLR-3443 which appears to never had been completed.   

I do think thinking about sharing memory might be something we need to have a good handle on for our future ML powered world..  As we add more models to our query/ingest steps, then maybe we would want to share memory?  (Dunno how that would all work!).    

If we move forward with this commit, we can always restore the code from Github right?   And use our FilestoreAPI or other newer APIs?

gus-asf:
> > I'm not so keen on the loss of https://issues.apache.org/jira/browse/SOLR-8349 along with this. It was only tied to the blob repository because that seemed to be how folks wanted me to load resources at the time.
> 
> Can you share more about how this is being used? I took a quick looksee at [SOLR-8349](https://issues.apache.org/jira/browse/SOLR-8349) and it points at [SOLR-3443](https://issues.apache.org/jira/browse/SOLR-3443) which appears to never had been completed.
> 
> I do think thinking about sharing memory might be something we need to have a good handle on for our future ML powered world.. As we add more models to our query/ingest steps, then maybe we would want to share memory? (Dunno how that would all work!).
> 
> If we move forward with this commit, we can always restore the code from Github right? And use our FilestoreAPI or other newer APIs?

Yeah I just noticed that I never circled back to 3443, but when I made the contribution the original patch (not using blobs) was used by a client because they had 40 core server that wanted to load a 1 GB post-code to lat/long dictionary (US & UK) in a custom component that was tagging documents with lat/long based on post codes. Loading it once rather than 40 times was pretty important. At a later date I think they converted this to the blob version to upgrade, but I was elsewhere by that time.

gus-asf:
So basically it's a facility for folks writing custom components... which should have been also dogfooded back into 3443

gus-asf:
> The "File Store" is the answer to saving & loading big resources. It's better than the Blob thing in that you have direct file access, which can be a requirement for some tools/plugins (that I see at work too).

Yeah I've never cared what the location for the bytes to be fetched was, the only part I cared about wast that it was possible to fetch the bytes just once (from wherever). But I'm seeing the test class and tests for this removed (ResourceSharingTestComponent etc).

epugh:
Okay, I dug around and start to see the FileStoreAPI.  It also looks like we never really documented it in the Ref Guide..   Another "magic API"....    So, the value of FileStoreAPI is you distribute things around the cluster, cool...   I *think* the value of BlobStore is you have only 1 copy of the item in the cluster, reachable from anywhere..   So the SchemaDesigner uses it to store a single copy of some data to be analyzed, and it needs that single copy to be reachable from whereever, but doesn't need replciation or robustness.....   Which is kind of cool.

epugh:
I'd like to get this clean up merged soon.  If, as it is currently structured with strenghts and weaknesses isn't good, then please do speak up with what we should do.  Otherwise I'd like to get this code that we agreed to deprecate back in 8.x line completed!

dsmiley (approved):

gus-asf:
I don't see any restoration of the resource sharing tests with file store?

solr/core/src/test/org/apache/solr/handler/component/ResourceSharingTestComponent.java
```diff
@@ -1,142 +0,0 @@
...
-import java.util.stream.Stream;
-import org.apache.solr.common.params.ModifiableSolrParams;
-import org.apache.solr.common.params.SolrParams;
-import org.apache.solr.core.BlobRepository;
-import org.apache.solr.core.SolrCore;
-import org.apache.solr.util.plugin.SolrCoreAware;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-public class ResourceSharingTestComponent extends SearchComponent implements SolrCoreAware {
```
gus-asf, line 38: This file should not be going away. It should be using whatever the appropriate file store is (and probably should have been less closely tied to blob store at the start, but back then I was just happy to have a committer working with me on it).

solr/core/src/test/org/apache/solr/core/BlobRepositoryCloudTest.java
```diff
@@ -1,127 +0,0 @@
...
-    document = new SolrInputDocument();
-    document.addField("id", "1");
-    document.addField("text", "col2");
-    solrClient.add("col2", document);
-    solrClient.commit("col2");
-    Thread.sleep(2000);
-  }
-
-  @Test
-  public void test() throws Exception {
```
gus-asf, line 78: This test demonstrates that the object shared remains visible across cores. We need to not loose this. It's really about the memory sharing more than the blobstore.

gus-asf (changes requested):

epugh:
I'm not familiar with Resource sharing aspect.  Not to put too fine a point on it, but would you be willing to port it over to the right api?   I suspect you can migrate it much faster than I can!   And then slap some docs into ref guide maybe so that others know this functionality exists?   Right now it just feels like orphaned code that doesn't have any love.   Heck, if you update the code, I'd be happy to take on the Ref Guide work...

gus-asf:
I'm willing in principal, but I don't really have a ton of time for it right now. I might be faster in coding cycles, but probably not in wall clock.

gus-asf:
The ideal thing would be to use it to solve https://issues.apache.org/jira/browse/SOLR-3443 with it at the same time since that's essentially the same problem my client faced. Not sure where in the ref guide it would go, since it's not an end-user feature directly but rather a component writer's aid, but if the fix for 3443 added a means to opt in to memory sharing... that would want to be documented.

epugh:
If this update passes all the tests, then I'm planning on merging it tomorrow, May 23rd....

epugh:
I am going to merge this for 10 only..  That way no one get's a nasty surprise in 9.7 ;-)

PR #2451, opened 2024-05-09
SOLR-17274: allow JSON atomic updates to use multiple modifiers or a modifier like 'set' as a field name if child docs not enabled 

eukaryote:
Fixes problems with the JsonLoader's parsing of atomic update values that use multiple modifiers in a single update or reference a field name that matches a modifier (like `set` or `add`).

https://issues.apache.org/jira/browse/SOLR-17274

# Description

Fixes two problems with JSON atomic updates where an atomic update value like `{"set": {"set": "foo"}}` (set a field named `set` to value `foo`) or `{"foo": {"add": "bar", "remove": "baz"}}` (add `bar` value to and remove `baz` values from field `foo`) would throw an exception due to `isChildDoc` thinking a nested doc was being updated. The exact exception message was:

`Unable to index docs with children: the schema must include definitions for both a uniqueKey field and the '_root_' field, using the exact same fieldType`

# Solution

I modified the `isChildDoc` method of `JsonLoader` to check if the schema is usable for child docs, and to return `false` if not. I also had to revert a part of the logic in `parseObjectFieldValue` so that it would return all the modifier values when there are multiple. The `isChildDoc` behavior is unchanged if the schema does support child docs, so the same error with continue to be thrown in that case.

# Tests

There weren't any atomic update tests that used the JSON loader previously, so I added a new schema file with the minimal fields needed to test this functionality and some tests for existing functionality via Json and for what is changed by this PR.

# Checklist

The only tests that failed were the S3 tests, which I believe is common and not necessarily a problem if my changes shouldn't have affected the s3 functionality.

### DIFF
--- a/solr/core/src/test/org/apache/solr/update/processor/AtomicUpdateJsonTest.java
+++ b/solr/core/src/test/org/apache/solr/update/processor/AtomicUpdateJsonTest.java
@@ -26,7 +26,7 @@ import org.junit.Test;
 
 // Tests atomic updates using JSON loader, since the existing
 // tests all use XML format, and there have been some atomic update
-// issues that were specific to the JSONformat.
+// issues that were specific to the JSON format.
 public class AtomicUpdateJsonTest extends SolrTestCaseJ4 {
 
   @BeforeClass
--- a/solr/core/src/test/org/apache/solr/update/processor/AtomicUpdateJsonTest.java
+++ b/solr/core/src/test/org/apache/solr/update/processor/AtomicUpdateJsonTest.java
@@ -30,7 +30,7 @@ import org.junit.rules.TestRule;
 // Tests atomic updates using JSON loader, since the existing
 // tests all use XML format, and there have been some atomic update
 // issues that were specific to the JSONformat.
-public class AtomicUpdatesJsonTest extends SolrTestCaseJ4 {
+public class AtomicUpdateJsonTest extends SolrTestCaseJ4 {
 
   @ClassRule
   public static final TestRule noReverseMerge = RandomNoReverseMergePolicyFactory.createRule();
@@ -52,7 +52,7 @@ public class AtomicUpdatesJsonTest extends SolrTestCaseJ4 {
     // the schema we loaded shouldn't be usable for child docs,
     // since we're testing JSON loader functionality that only
     // works in that case and is ambiguous if nested docs are supported.
-    assert !h.getCore().getLatestSchema().isUsableForChildDocs();
+    assertFalse(h.getCore().getLatestSchema().isUsableForChildDocs());
   }
 
   @Test
--- a/solr/core/src/test/org/apache/solr/update/processor/AtomicUpdateJsonTest.java
+++ b/solr/core/src/test/org/apache/solr/update/processor/AtomicUpdateJsonTest.java
@@ -149,7 +149,7 @@ public class AtomicUpdateJsonTest extends SolrTestCaseJ4 {
 
     doc = new SolrInputDocument();
     doc.setField("id", "1");
-    Map<String, String> removeSetField = new HashMap<String, String>();
+    Map<String, String> removeSetField = new HashMap<>();
     removeSetField.put("set", null);
     doc.setField("set", removeSetField);
     updateJ(jsonAdd(doc), null);
--- a/solr/core/src/test/org/apache/solr/update/processor/AtomicUpdateJsonTest.java
+++ b/solr/core/src/test/org/apache/solr/update/processor/AtomicUpdateJsonTest.java
@@ -37,7 +37,7 @@ public class AtomicUpdateJsonTest extends SolrTestCaseJ4 {
 
   @Before
   public void before() {
-    h.update("<delete><query>*:*</query></delete>");
+    assertU(delQ("*:*"));
     assertU(commit());
   }
 
--- a/solr/core/src/test/org/apache/solr/update/processor/AtomicUpdateJsonTest.java
+++ b/solr/core/src/test/org/apache/solr/update/processor/AtomicUpdateJsonTest.java
@@ -62,14 +62,14 @@ public class AtomicUpdateJsonTest extends SolrTestCaseJ4 {
     doc.setField("name", new String[] {"aaa"});
     updateJ(jsonAdd(doc), null);
     assertU(commit());
-    assertQ(req("q", "name:bbb", "indent", "true"), "//result[@numFound = '0']");
+    assertQ(req("q", "name:bbb"), "//result[@numFound = '0']");
 
     doc = new SolrInputDocument();
     doc.setField("id", "1");
     doc.setField("name", Map.of("add", "bbb"));
     updateJ(jsonAdd(doc), null);
     assertU(commit());
-    assertQ(req("q", "name:bbb", "indent", "true"), "//result[@numFound = '1']");
+    assertQ(req("q", "name:bbb"), "//result[@numFound = '1']");
   }
 
   @Test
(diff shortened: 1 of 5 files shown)


solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
```diff
@@ -0,0 +1,206 @@
...
+    h.update("<delete><query>*:*</query></delete>");
+    assertU(commit());
+  }
+
+  @Test
+  public void testSchemaIsNotUsableForChildDocs() throws Exception {
+    // the schema we loaded shouldn't be usable for child docs,
+    // since we're testing JSON loader functionality that only
+    // works in that case and is ambiguous if nested docs are supported.
+    assert !h.getCore().getLatestSchema().isUsableForChildDocs();
```
dsmiley, line 55: use assertFalse in tests, not assert keyword

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
```diff
@@ -0,0 +1,206 @@
...
+    doc = new SolrInputDocument();
+    doc.setField("id", "1");
+    doc.setField(
+        "name",
+        new HashMap<String, Object>() {
+          {
+            put("add", new String[] {"ddd", "eee"});
+            put("remove", new String[] {"aaa", "ccc"});
+          }
+        });
```
dsmiley, line 111: Please do Map.of instead

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
```diff
@@ -0,0 +1,206 @@
...
+    updateJ(jsonAdd(doc), null);
+    assertU(commit());
+    assertQ(req("q", "set:modval", "indent", "true"), "//result[@numFound = '1']");
+    assertQ(req("q", "set:setval", "indent", "true"), "//result[@numFound = '0']");
+
+    doc = new SolrInputDocument();
+    doc.setField("id", "1");
+    doc.setField(
+        "set",
+        new HashMap<String, String>() {
```
dsmiley, line 166: Map.of

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
```diff
@@ -0,0 +1,206 @@
...
+          {
+            put("set", null);
+          }
+        });
+    updateJ(jsonAdd(doc), null);
+    assertU(commit());
+    assertQ(req("q", "set:modval", "indent", "true"), "//result[@numFound = '0']");
```
dsmiley, line 173: Don't need indent:true in any of this file.  I recall Solr does the indenting anyway in a test failure

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
```diff
@@ -0,0 +1,206 @@
...
+    doc.setField("add", new String[] {"aaa", "bbb", "ccc"});
+    updateJ(jsonAdd(doc), null);
+    assertU(commit());
+    assertQ(req("q", "add:bbb", "indent", "true"), "//result[@numFound = '1']");
+
+    doc = new SolrInputDocument();
+    doc.setField("id", "1");
+    doc.setField(
+        "add",
+        new HashMap<String, Object>() {
```
dsmiley, line 192: Map.of

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
```diff
@@ -0,0 +1,206 @@
...
+import org.junit.Test;
+import org.junit.rules.TestRule;
+
+// Tests atomic updates using JSON loader, since the existing
+// tests all use XML format, and there have been some atomic update
+// issues that were specific to the JSONformat.
+public class AtomicUpdatesJsonTest extends SolrTestCaseJ4 {
+
+  @ClassRule
+  public static final TestRule noReverseMerge = RandomNoReverseMergePolicyFactory.createRule();
```
dsmiley, line 36: there are no nested docs in this schema so this thing isn't needed

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
```diff
@@ -0,0 +1,206 @@
...
+
+  @BeforeClass
+  public static void beforeTests() throws Exception {
+    System.setProperty("enable.update.log", "true");
+    initCore("solrconfig.xml", "atomic-update-json-test.xml");
+  }
+
+  @Before
+  public void before() {
+    h.update("<delete><query>*:*</query></delete>");
```
dsmiley, line 46: see `delQ`

dsmiley (commented):
Nice test

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
```diff
@@ -0,0 +1,206 @@
...
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.rules.TestRule;
+
+// Tests atomic updates using JSON loader, since the existing
+// tests all use XML format, and there have been some atomic update
+// issues that were specific to the JSONformat.
+public class AtomicUpdatesJsonTest extends SolrTestCaseJ4 {
```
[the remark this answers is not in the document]
eukaryote: Updated. Should I rename the `AtomicUpdatesTest.java` as well, which I didn't otherwise change?

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
eukaryote: Updated all.

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
eukaryote: `Map.of` doesn't seem to allow `null` values, so I can't do `Map.of("set", null)` here. What do you suggest instead of `HashMap`?

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
eukaryote: Done.

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
eukaryote: Removed.

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
eukaryote: Updated.

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
eukaryote: Updated. Does the reviewer generally mark the conversation as resolved after reviewing the change? I haven't used GitHub in ages.

dsmiley: Typically I think but we have no standards in this regard.

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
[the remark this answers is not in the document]
dsmiley: no

dsmiley:
Thinking of wording for CHANGES.txt:

> Improved disambiguation of atomic updates from child docs for JSON based updates (not an issue for other formats). 
> Multiple atomic updates to a field in one update is supported (again).  Schema fields can be named "set", "add", etc. (again).  These improvements only apply when the schema doesn't have the root field (no child docs).

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesJsonTest.java
eukaryote: Oh, it looks like one of the checks is failing because of the instance initializer block, so I'll change it to use `HashMap` but avoid that.

eukaryote: Updated.

solr/core/src/test/org/apache/solr/update/processor/AtomicUpdateJsonTest.java
```diff
@@ -149,14 +149,9 @@ public void testAtomicUpdateModifierNameSingleValued() throws Exception {
...
     doc.setField("id", "1");
-    //    doc.setField("set", Map.of("set", null));
-    doc.setField(
-        "set",
-        new HashMap<String, String>() {
-          {
-            put("set", null);
-          }
-        });
+    Map<String, String> removeSetField = new HashMap<String, String>();
```
dsmiley, line 152:
```suggestion
    Map<String, String> removeSetField = new HashMap<>();
```

dsmiley:
I'll merge this tonight unless I hear to the contrary.

pasupathi-rajamanickam-odp:
When this will be published for public use?

dsmiley:
Whenever 9.7 is released.  Should be within a month.

PR #2452, opened 2024-05-09
User Behavior Insights implementation for Apache Solr

epugh:
# Description

I am working with other folks, especially Stavros Macrakis (macrakis@gmail.com),  to come up with a solution for understanding what users are doing in response to search results.  We have great visibility and understanding of an incoming query, what we do with it, and then what docs are sent back.   We do NOT have a way of tying that search to then what does the user do next, and if the following query is connected to the original one.

Many teams lean on GA or Snowplow or custom code for tracking click through, add to cart, etc as signals, but nothing that is drop dead simple to use and open source.

# Solution

_User Behavior Insights_ is a shared schema for tracking search related activities.   There is a basic implementation for [OpenSearch](https://github.com/opensearch-project/OpenSearch/issues/12084) and this is a version for Apache Solr.

Tasks to be done:

- [x] Demonstrate providing a `.expr` file and using it to write to another Solr collection.
- [ ] Look at performance implications of the every query generates a streaming expressoin.
- [ ] Check we only record on the main node, not the replicas when sharding.
- [ ] How can I load test this?
- [ ] Write up Ref Guide Docs
- [ ] Can we add it to `techproducts` as an example?
- [x] Add UBI to Admin UI as flag
- [ ] Add UBI to SolrJ basic client
- [ ] Add UBI to SolrJ JSON Query client

# Tests

Bats test to demonstrate the end to end use of UBI.

# Checklist


HoustonPutman:
I'd love to review here, but I think I need some more starting information either in a ref guide page or a JIRA, I'm kind of lost right now...

chatman:
Usually, features like these are discussed in the dev@ list, or in JIRA or a SIP.
Most important question I have in mind is whether this needs to be in the core search engine? If not, can this not be a plugin/package, shipped outside of solr-core?

epugh:
This is definitely draft mode code...   I opened it as a PR just to be able to track the work, and once it gets a bit furthur, I plan on opening a proper discussion about it.   Module?  Solr Sandbox? A Component?  A full blown package?   So many fun options...

epugh:
I figured out how to parse and run a streaming expression that is used to write the query analytics data to, well, anywhere we want ;-).   The next area is to look at is actually integrating the streaming expression INTO the component as more than just a one off..  I gotta figure out how to take the data and pass it into the streaming expression...  `input()` maybe???   Also think about how to not rebuild/destroy/rebuild the streaming expression for every query.   

Then more Ref Guide docs, a BATS integration test maybe, and then a discussion about who wants to use it first!  Plus of course the ever critical, "where does the code live" conversation.

epugh:
Oh, and of course, we now have a machine readable schema via Json Schema available here https://github.com/o19s/ubi

epugh:
I am having some second thoughts about the idea of logging ubi queries to disk...   Why?  Any real use case you want them to go somewhere.  Plus log4j is a pain to touch...   So...   May just rip that part out.  You want to log to disk?  just write a streaming expression ;-)

epugh:
Making progress....    I've added some tasks to do, and then I think I'm going to flip it from Draft to Ready for Review and email the community.  I'd like to demo it at a upcoming community meeting.

BTW, UBI now has an actual website!   https://www.ubisearch.dev/

dsmiley:
Something like this would be great to discuss at a Solr meetup/conf/whatever.  Feel free to take half of a community meetup to show & tell when you're ready.

epugh:
A question for the smarter folks that me.  Should the classes `UBIQuery` and `UBIQueryStream` be added to the `UBIComponent.java`?   UBIQuery is just a pojo...    And UBIQueryStream wires the use of the component up to a streaming expression.  I don't see either ever being used elsewhere....

epugh:
Just stubbed my toe on the "Distributed processing is harder than single core processing"!   With a two node set up, I discovered that I am logging to a SINGLE `userfiles/ubi_queries.jsonl` file, and I log once for each shard..  instead of just logging on the collector step..

```
{"query_id":"c4e40af6-67b7-4824-8b63-5aae70a485f6","timestamp":"2024-11-27T13:42:19.121Z"}
{"query_id":"5dfedf02-fd89-4e40-b3aa-7700c162800b","timestamp":"2024-11-27T13:42:19.121Z"}
```

Sigh.

epugh:
Argh, a bit stuck.  I can't figure out how to have the UBIComponent during a distributed query, look up the final doc id's and record them before sending them back to the user.  With a single node single shard, it works great, but not in a distributed fashion.

I keep getting:

```
2024-11-28 12:36:31.368 ERROR (qtp428039780-40-localhost-11) [c:twoshard s:shard1 r:core_node4 x:twoshard_shard1_replica_n2 t:localhost-11] o.a.s.s.HttpSolrCall 500 Exception => java.lang.NullPointerException: Cannot read field "docList" because the return value of "org.apache.solr.handler.component.ResponseBuilder.getResults()" is null
	at org.apache.solr.handler.component.UBIComponent.doStuff(UBIComponent.java:315)
java.lang.NullPointerException: Cannot read field "docList" because the return value of "org.apache.solr.handler.component.ResponseBuilder.getResults()" is null
	at org.apache.solr.handler.component.UBIComponent.doStuff(UBIComponent.java:315) ~[?:?]
	at org.apache.solr.handler.component.UBIComponent.distributedProcess(UBIComponent.java:252) ~[?:?]
	at org.apache.solr.handler.component.SearchHandler.processComponents(SearchHandler.java:552) ~[?:?]
	at org.apache.solr.handler.component.SearchHandler.handleRequestBody(SearchHandler.java:429) ~[?:?]
	at org.apache.solr.handler.RequestHandlerBase.handleRequest(RequestHandlerBase.java:238) ~[?:?]
	at org.apache.solr.core.SolrCore.execute(SolrCore.java:2875) ~[?:?]

```

solr/core/src/java/org/apache/solr/handler/component/UBIComponent.java
```diff
@@ -0,0 +1,424 @@
...
+        Map paramsProperties = (Map) jsonProperties.get("params");
+        if (paramsProperties.containsKey(QUERY_ATTRIBUTES)) {
+          @SuppressWarnings("rawtypes")
+          Map queryAttributesAsMap = (Map) paramsProperties.get(QUERY_ATTRIBUTES);
+          ubiQuery.setQueryAttributes(queryAttributesAsMap);
+        }
+      }
+    }
+
+    ResultContext rc = (ResultContext) rb.rsp.getResponse();
```
mkhludnev, line 296: To make it distributed, we need to split this method here, and execute parts per certain stage.
It seems, UBIComp needs to submit docids found. Right? 
In distributed context, we need to record before-merge per-shards ids or resulting merged&cropped result ids? I bet the later, please confirm.

solr/core/src/java/org/apache/solr/handler/component/UBIComponent.java
```diff
@@ -0,0 +1,424 @@
...
+    recordQuery(ubiQuery);
+  }
+
+  private void recordQuery(UBIQuery ubiQuery) throws IOException {
+    TupleStream stream;
+
+    stream = constructStream(streamFactory, streamExpression);
+
+    streamContext.put("ubi-query", ubiQuery);
+    stream.setStreamContext(streamContext);
```
mkhludnev, line 313: I know nothing about streams, but is it possible to submit UbiQuery here, which isn't a subclass of any serializable framework? You know passing a ref might work locally, but not it remote/distributed env.

mkhludnev (commented):
Hi @epugh !

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -126,19 +125,24 @@ public class SearchHandler extends RequestHandlerBase
...
-    names.add(FacetModule.COMPONENT_NAME);
-    names.add(MoreLikeThisComponent.COMPONENT_NAME);
-    names.add(HighlightComponent.COMPONENT_NAME);
-    names.add(StatsComponent.COMPONENT_NAME);
-    names.add(DebugComponent.COMPONENT_NAME);
-    names.add(ExpandComponent.COMPONENT_NAME);
-    names.add(TermsComponent.COMPONENT_NAME);
-
-    return names;
+    List<String> l = new ArrayList<String>(SearchComponent.STANDARD_COMPONENTS.keySet());
```
mkhludnev, line 136: As a result RTG occurs as a default component that causes a problem in an essential cloud test. 😭

solr/core/src/java/org/apache/solr/handler/component/UBIComponent.java
epugh: Thanks @mkhludnev  for looking at this, gives me renewed energy to know someone else is looking at it!  Yes, it is the merged&cropped result ids.   We want to record in the plugin what the candidate result ids that the user MIGHT have seen, which later is used to compare against clicks and impressions to identify which docs are NOT attractive to the user.

solr/core/src/java/org/apache/solr/handler/component/UBIComponent.java
epugh: GOod question.   I *believe* that the `getTuple` method will be immediately run, which means this code doesn't actually get run in a distributed sense..  I.e, the `ubiQuery`  object that we put into the `streamContext` is immediately read back out in `getTuple()` method..    That is the job of the `UBIQueryStream` class, to convert the `UBIQuery` found in the context into a `Tuple` that is used by streaming expressons.

I am going to try actually making `UBIQuery` and `UBIQueryStream` inner classes of `UBIComponent` to see how that looks...

solr/core/src/java/org/apache/solr/handler/component/UBIComponent.java
```diff
@@ -0,0 +1,424 @@
...
+
+    if (schema.getUniqueKeyField() == null) {
+      log.error("Can't track documents for query without unique field.");
+      return "";
+    }
+    StringBuilder sb = new StringBuilder();
+
+    Set<String> fields = Collections.singleton(schema.getUniqueKeyField().getName());
+    for (DocIterator iter = dl.iterator(); iter.hasNext(); ) {
+      sb.append(schema.printableUniqueKey(searcher.getDocFetcher().doc(iter.nextDoc(), fields)))
```
mkhludnev, line 336: what about commas in id values? Isn't it safer to use json array as a convention for this field?

solr/core/src/java/org/apache/solr/handler/component/UBIComponent.java
mkhludnev: ok. got it. found `toMap()`. Shouldn't it yield docIds as well? I can't see refs to this field there https://github.com/epugh/solr/blob/99d6b7a7eb7b28a92f4cb36d4a525f8b901ba93c/solr/core/src/java/org/apache/solr/handler/component/UBIQuery.java#L103

solr/core/src/java/org/apache/solr/handler/component/UBIComponent.java
epugh: humm...    maybe?   Is that at all ever done?   So, most folks are going to take this data, and load it into a pandas dataframe or soemthing like that...    Now, since the doc is already JSON..   maybe it's okay to make this a array of JSON as well?    Since you already have to parse some JSON to do anything with it......

solr/core/src/java/org/apache/solr/handler/component/UBIComponent.java
epugh: That's a fair question...   OMG...     I *can't* believe that I forgot to add the damn doc ids...   I suck.   

Thanks for spotting that!   How embarrassing!   yeah...

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
epugh: humm...    Maybe I just back out this optimization...

epugh: check out the changes I made to back this out, but still keep the name `STANDARD_COMPONENTS`...

solr/core/src/java/org/apache/solr/handler/component/UBIComponent.java
epugh: And fixed.

gerlowskija:
> I think I'm going to flip it from Draft to Ready for Review and email the community. I'd like to demo it at a upcoming community meeting.

@epugh did you ever demo this? I don't remember it, and think it'd be an awesome thing to highlight (and a great way to garner interest and get feedback)

epugh:
Hey @ozlerhakan here si the draft PR...  it just needs a bit of love to run in solr cloud mode...

solr/core/src/java/org/apache/solr/handler/component/SearchComponent.java
```diff
@@ -110,16 +111,20 @@ public void initializeMetrics(SolrMetricsContext parentContext, Attributes attri
...
-          QueryComponent.COMPONENT_NAME, QueryComponent.class,
-          FacetComponent.COMPONENT_NAME, FacetComponent.class,
-          FacetModule.COMPONENT_NAME, FacetModule.class,
-          MoreLikeThisComponent.COMPONENT_NAME, MoreLikeThisComponent.class,
-          StatsComponent.COMPONENT_NAME, StatsComponent.class,
-          DebugComponent.COMPONENT_NAME, DebugComponent.class,
-          RealTimeGetComponent.COMPONENT_NAME, RealTimeGetComponent.class,
-          ExpandComponent.COMPONENT_NAME, ExpandComponent.class,
-          TermsComponent.COMPONENT_NAME, TermsComponent.class);
+  public static final Map<String, Class<? extends SearchComponent>> STANDARD_COMPONENTS;
```
epugh, line 114: just passed the limit of `Map.of`!

dsmiley: That doesn't at all mean you need to go back to HashMap!  Tell an LLM this and it'll get this better.

PR #2454, opened 2024-05-10
Refactor preparePutOrPost in HttpJdkSolrClient

andywebb1975:
# Description

This refactors the `preparePutOrPost` method in HttpJdkSolrClient to make the code clearer, and remove the previous unreachable `else` block.

# Tests

All tests still pass.

# Checklist

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpJdkSolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpJdkSolrClient.java
@@ -298,15 +298,17 @@ public class HttpJdkSolrClient extends HttpSolrClientBase {
 
       InputStream is = streams.iterator().next().getStream();
       bodyPublisher = HttpRequest.BodyPublishers.ofInputStream(() -> is);
-    } else if (queryParams != null && urlParamNames != null) {
-      ModifiableSolrParams requestParams = queryParams;
-      queryParams = calculateQueryParams(urlParamNames, requestParams);
-      queryParams.add(calculateQueryParams(solrRequest.getQueryParams(), requestParams));
-      // note the toQueryString() method adds a leading question mark which needs to be removed here
-      bodyPublisher =
-          HttpRequest.BodyPublishers.ofString(requestParams.toQueryString().substring(1));
     } else {
-      bodyPublisher = HttpRequest.BodyPublishers.noBody();
+      // move any params specified in urlParamNames or solrRequest from queryParams into urlParams
+      ModifiableSolrParams urlParams = calculateQueryParams(urlParamNames, queryParams);
+      urlParams.add(calculateQueryParams(solrRequest.getQueryParams(), queryParams));
+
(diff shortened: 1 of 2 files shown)


solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpJdkSolrClient.java
```diff
@@ -299,10 +299,19 @@ private PreparedRequest preparePutOrPost(
...
     } else if (queryParams != null && urlParamNames != null) {
-      ModifiableSolrParams requestParams = queryParams;
-      queryParams = calculateQueryParams(urlParamNames, requestParams);
-      queryParams.add(calculateQueryParams(solrRequest.getQueryParams(), requestParams));
-      bodyPublisher = HttpRequest.BodyPublishers.ofString(requestParams.toString());
+      // move params specified in urlParamNames from queryParams into urlParams
+      ModifiableSolrParams urlParams = calculateQueryParams(urlParamNames, queryParams);
+
+      // note this is only triggered if urlParamNames is set too - this this correct?
+      urlParams.add(calculateQueryParams(solrRequest.getQueryParams(), queryParams));
```
andywebb1975, line 306: I'm unclear when this would kick in - it can't trigger if `urlParamNames` is null.

jdyer1: Indeed `urlParamNames` is never null (see `HttpSolrClientBase` constructor line 107ff), so the null check is spurious.  Should you comment-out line 306, you will see `testQueryString` fail in the case where there are query parameters but nothing was set for `urlParamNames` (using the `withTheseParamNamesInTheUrl` option).

andywebb1975: `queryParams` can't be null either, can it? I've made this the `else` block - I added an `UnsupportedOperationException` in the original `noBody()` one and saw it was never triggered by tests. Does this change look good to you?

dsmiley:
Shall this be merged?

andywebb1975:
@jdyer1 does this look OK to you? The tests pass but I'm not totally familiar with the use case to know that this is definitely correct.

(It's not a bug so I'd be happy to target 9.7.0 with this if we merge it.)

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpJdkSolrClient.java
jdyer1: Well, you are right, it does not seem urlParams or queryParams can ever be null.  So you're correct, the "noBody" case never gets touched.  I guess I had thought there would be a case where we would could post with an empty body but clearly this cannot happen.

jdyer1:
> @jdyer1 does this look OK to you? The tests pass...

I think your changes are valid and if all existing tests still pass then its a win.  I am all in favor of no dead code branches!

PR #2457, opened 2024-05-11
SOLR-17066 Fix ShardSplitTest & StressHdfsTest

dsmiley:
SOLR-17066 broke some tests marked Nightly since the beginning of February; they didn't fail previously then they fail always.  It's so sad we don't have mechanisms to bring our attention to something that is so glaring.  I just happened to be looking at build failures, and I saw so many and I started looking at them.

http://fucit.org/solr-jenkins-reports/history-trend-of-recent-failures.html#series/org.apache.solr.cloud.api.collections.ShardSplitTest.test
https://ge.apache.org/scans/tests?search.rootProjectNames=*solr*&search.timeZoneId=America%2FNew_York&tests.container=org.apache.solr.cloud.api.collections.ShardSplitTest&tests.help=executedTest&tests.test=test

Even though SOLR-17066 shipped, I'm nonetheless doing a PR against the same issue because I _basically_ just fixed tests.

### DIFF
--- a/solr/core/src/test/org/apache/solr/cloud/api/collections/ShardSplitTest.java
+++ b/solr/core/src/test/org/apache/solr/cloud/api/collections/ShardSplitTest.java
@@ -1277,9 +1277,7 @@ public class ShardSplitTest extends BasicDistributedZkTest {
     QueryRequest request = new QueryRequest(params);
     request.setPath("/admin/collections");
 
-    String baseUrl =
-        ((HttpSolrClient) shardToJetty.get(SHARD1).get(0).client.getSolrClient()).getBaseURL();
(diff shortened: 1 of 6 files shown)


gerlowskija (approved):
Ah, apologies for not catching these!

This LGTM - I especially appreciate the javadoc tweaks on the getUrl methods to indicate which style of URL is returned.

dsmiley:
I'm going to merge this even though not all tests are fixed in ShardSplitTest.  At least the ones introduced by SOLR-17066 are.  

Sadly there are still more failures that seemed to have been introduced earlier, like org.apache.solr.cloud.api.collections.ShardSplitTest.testSplitStaticIndexReplication starting 2023-11 which I'm looking in to.

PR #2460, opened 2024-05-14
SOLR-17290: Update SyncStrategy and PeerSyncWithLeader to use the recovery Http2SolrClient 

iamsanjay:

As per the discussion on this thread [SOLR-16505](https://issues.apache.org/jira/browse/SOLR-16505), It has been identified that `PeerSyncWithLeader` and `SyncStrategy` both were using the default HttpClient. Because both of these classes are meant for "recovery" operations, then why not switch their client to use recovery Http Client.

Also, now that SOLR-16505 has been merged, the above change will enable `PeerSyncWithLeader` and `SyncStrategy` to use Jetty HTTP2 client.

# Tests

No Test Cases Provided.

# Checklist


dsmiley:
I don't think this PR should be for SOLR-16503.  It should be SOLR-16505 (it's okay it's closed, can be reopened; is unreleased so there's time) as it's focused on "recovery".  Or perhaps a new JIRA.

dsmiley (commented):
The code changes look good.

iamsanjay:
> I don't think this PR should be for [SOLR-16503](https://issues.apache.org/jira/browse/SOLR-16503). It should be [SOLR-16505](https://issues.apache.org/jira/browse/SOLR-16505) (it's okay it's closed, can be reopened; is unreleased so there's time) as it's focused on "recovery". Or perhaps a new JIRA.

New JIRA, Done!

iamsanjay:
For SyncStrategy we have re-created Http2SolrClient due to custom timeouts and never called `withListenerFactory` meaning the auth listener is not available. The solr client is being used for REQUESTRECOVERY and did not seem to need the auth for that.

dsmiley:
+1; will merge tomorrow.

> The solr client is being used for REQUESTRECOVERY and did not seem to need the auth for that.

Maybe you know something I don't here; what technical mechanism doesn't happen for recovery for auth?  AFAIK all Solr-to-Solr communication needs to support AuthenticationPlugin.  And I believe it is == see `org.apache.solr.update.UpdateShardHandler#setSecurityBuilder` invoked by the auth plugin initialization and that which configures the update & recovery clients.

dsmiley (approved):

iamsanjay:
> +1; will merge tomorrow.
> 
> > The solr client is being used for REQUESTRECOVERY and did not seem to need the auth for that.
> 
> Maybe you know something I don't here; what technical mechanism doesn't happen for recovery for auth? AFAIK all Solr-to-Solr communication needs to support AuthenticationPlugin. And I believe it is == see `org.apache.solr.update.UpdateShardHandler#setSecurityBuilder` invoked by the auth plugin initialization and that which configures the update & recovery clients.

I found a bug in SOLR-16505! 
RecoveryStrategy is using a client without auth listener. It did not failed because, i believe, there is no specific test for Recovery where SolrCloud is configured along with Security.json. Going to submit PR today for a fix.
@dsmiley Please open again SOLR-16505.

epugh:
Would love some thoughts on how we could handle the auth scenarios better without having to write an explicit test for it on every piece of code, once with auth and once with out ;-)

iamsanjay:
> Would love some thoughts on how we could handle the auth scenarios better without having to write an explicit test for it on every piece of code, once with auth and once with out ;-)

IMO, There are not many test cases that have written along with Auth enabled in SolrCloud environment, and apparently it's none for this specific scenario #2462 where a RecoveryStrategy is begin tested in Auth enabled environment. I am not sure how many cases this new test class #2462 can cover, however the idea would be to keep on extending this one to cover more cases for Cloud related scenario.

For instance, after merging #2462, same test case would be extends to test SyncStrategy and PeerSyncWithLeader Http2SolrClient calls to make sure that they are executing without any errors in auth enabled SolrCloud environment.

dsmiley:
Should this be merged or should it be pending some authentication/authorization fix?

solr/core/src/java/org/apache/solr/cloud/SyncStrategy.java
```diff
@@ -361,12 +360,11 @@ private void requestRecovery(
               RequestRecovery recoverRequestCmd = new RequestRecovery();
               recoverRequestCmd.setAction(CoreAdminAction.REQUESTRECOVERY);
               recoverRequestCmd.setCoreName(coreName);
-
               try (SolrClient client =
-                  new HttpSolrClient.Builder(baseUrl)
-                      .withHttpClient(SyncStrategy.this.client)
+                  new Http2SolrClient.Builder(baseUrl)
```
iamsanjay, line 364: > Should this be merged or should it be pending some authentication/authorization fix?

Yes wherever we are re-creating the Http2SolrClient and in SyncStrategy we are -- because of custom timeouts. This code for listener factory seems a broken window to me as there is no way to enforce it and can be easily missed in future. @dsmiley We had discussion regarding the timeouts handling at request level then we don't need to re-create the client.

Other option that I am also in favour of is to copy variable via calling `withHttpClient(solrClient)` as we are already doing with other configuration.

https://github.com/apache/solr/blob/6af52f33ac78155523ec6fc610c563127b81c139/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java#L1033-L1058

iamsanjay:
I updated the SyncStrategy to call `withListenerFactory`. However, this one can wait, let's merge this one first #2460 to fix RecoveryStrategy.

solr/core/src/java/org/apache/solr/cloud/SyncStrategy.java
dsmiley: Shouldn't withHttpClient copy *everything* over, including the ListenerFactory?  It's worthy of its own JIRA/PR.

iamsanjay: @dsmiley #2467 is for that one.

solr/CHANGES.txt
```diff
@@ -150,10 +150,13 @@ Other Changes
 
 * GITHUB#2454: Refactor preparePutOrPost method in HttpJdkSolrClient (Andy Webb)
 
+* SOLR-16503: Use Jetty HTTP2 for SyncStrategy and PeerSyncWithLeader for "recovery" operations (Sanjay Dutt, David Smiley)
+
 ==================  9.6.1 ==================
 Bug Fixes
 ---------------------
 * SOLR-17296: Remove (broken) singlePass attempt when reRankScale + debug is used, and fix underlying NPE. (hossman)
+>>>>>>> main
```
dsmiley, line 159: a glitch

dsmiley (commented):
Just one little glitch here but otherwise I approve.

PR #2461, opened 2024-05-15
SOLR-17294: The stall detection in the ConcurrentUpdateSolrClients easily detects false positives.

markrmiller:
The current stall detection mechanism in the ConcurrentUpdateSolrClients is prone to generating false positives, especially under load. As pointed out by Jason, the existing approach simply intermittently monitors the queue size over time to detect stalls. However, this method is insufficient because the queue can report being full for extended periods when the system is under load, even if no actual stall has occurred.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -265,6 +265,8 @@ Bug Fixes
 
 * SOLR-17740: When the V2 API is receiving raw files, it could sometimes skip the first byte. (David Smiley)
 
+* SOLR-17294: ConcurrentUpdateSolrClient no longer detects "false-positive" stalls. (Mark Miller, Jason Gerlowski)
+
 Dependency Upgrades
 ---------------------
 * SOLR-17471: Upgrade Lucene to 9.12.1. (Pierre Salagnac, Christine Poerschke)
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/ConcurrentUpdateSolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/ConcurrentUpdateSolrClient.java
@@ -83,7 +83,6 @@ public class ConcurrentUpdateSolrClient extends SolrClient {
   final int threadCount;
   boolean shutdownExecutor = false;
   int pollQueueTimeMillis = 250;
-  int stallTimeMillis;
   private final boolean streamDeletes;
   private boolean internalHttpClient;
   private final int connectionTimeout;
@@ -94,6 +93,8 @@ public class ConcurrentUpdateSolrClient extends SolrClient {
   AtomicInteger pollExits;
   AtomicInteger blockLoops;
   AtomicInteger emptyQueueLoops;
+
+  protected StallDetection stallDetection;
 
   /**
    * Use builder to construct this class. Uses the supplied HttpClient to send documents to the Solr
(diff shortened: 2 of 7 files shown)


markrmiller:
There is no real reason you need to check the size of the queue as well, but I just kept it anyway.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/ConcurrentUpdateSolrClient.java
```diff
@@ -566,24 +576,28 @@ public NamedList<Object> request(final SolrRequest<?> request, String collection
...
           } else {
-            if (lastStallTime == -1) {
-              // mark a stall but keep trying
-              lastStallTime = System.nanoTime();
-            } else {
-              long currentStallTime =
-                  TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - lastStallTime);
-              if (currentStallTime > stallTimeMillis) {
+            long timeElapsed = TimeUnit.NANOSECONDS.toMillis(currentTime - lastCheckTime);
+            if (timeElapsed > stallTimeMillis) {
```
sigram, line 589: Please note that the same logic is present in 3 places - request(), blockUntilFinished() and waitForEmptyQueue(). Also, we're supposed to use `ConcurrentUpdateHttp2SolrClient` now ;)

solr/solrj/src/java/org/apache/solr/client/solrj/impl/ConcurrentUpdateSolrClient.java
```diff
@@ -566,24 +576,28 @@ public NamedList<Object> request(final SolrRequest<?> request, String collection
         if (!success) {
           // stall prevention
           int currentQueueSize = queue.size();
+          long currentTime = System.nanoTime();
+          long processed = processedCount.sum();
```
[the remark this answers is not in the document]
markrmiller: > [Q] Should we still be comparing queue-sizes in  L581 below?  As you've highlighted in the JIRA ticket and this PR description - queue size isn't a pretty ambivalent indicator of whether the queue processing is stalled or not.

Answered in the first comment: "There is no real reason you need to check the size of the queue as well, but I just kept it anyway."

I don't believe it adds anything, but I just kept this first draft patch to bare minimum changes basically.

markrmiller:
Leaving Draft state on still on this one, but last commit brings it closer to what I had in mind. Still have to review it and run tests.

markrmiller:
Was just reminded about this recently when I tried to do some bulk indexing. To me, this Band-Aid ended up worse than what it was trying to cover. When it works, it just hides the underlying issue that caused it from developers, making it unlikely to ever be fixed, but it has never really worked as long as you are indexing fast enough to keep the queue full for stall timeout seconds. And at least with http2, when it kicks in, it totally botches the indexing job from what I've seen. Stall exceptions lead to cancelled stream errors which lead to the whole indexing job grinding down until your using all the cpu you had been using when properly indexing but with no progress.

Regardless, I've removed the Draft label and run the tests.

solr/CHANGES.txt
```diff
@@ -243,6 +243,8 @@ Bug Fixes
 * SOLR-17692: Core unload/deletion now preempts all forms of ongoing "recovery", rather than inadvertently waiting for
   completion in some cases. (Jason Gerlowski)
 
+* SOLR-17294: The stall detection in the ConcurrentUpdateSolrClients easily detects false positives. (Mark Miller, Jason Gerlowski)
```
gerlowskija, line 246: [-0] The user-facing benefit might be clearer if this is phrased in terms of the new/improved behavior. e.g.

> ConcurrentUpdateSolrClient no longer detects "false-positive" stalls, improving throughput and decreasing logging "noise" (Mark Miller)

gerlowskija (approved):
LGTM, other than a small suggestion on the CHANGES.txt entry

gerlowskija:
I guess the one gap here is tests - is there any way to refactor the stall-detection logic into its own class in a way that makes it easier to validate _when_ stalls get detected, when the stall-timer gets reset, etc?

solr/CHANGES.txt
markrmiller: > improving throughput and decreasing logging "noise"

I'd typically be fine with that, but here it requires saying more than I'm comfortable with. In the indexing I was recently doing, it completely hosed my indexing job and I had to work around it to even finish indexing. Which I guess you could argue is still a loss of throughput (infinite loss?), but I'm not confident saying that it just increases throughput. Similar with logging "noise". Noise to me is unnecessary logging, but this logging is coming from a false claim that there is a stall, so reduced logging noise is also accurate I suppose, but feels misleading.

The positive is much harder to define exactly, whereas the problem, it was easy to detect false stalls, that I'm sure of.

gerlowskija (approved):
+1 - love the new tests.

LGTM whenever you want to merge.

HoustonPutman:
@markrmiller this is creating some flaky failures. Such as: https://jenkins.thetaphi.de/view/Solr/job/Solr-main-Linux/24834/

markrmiller:
I rewrote that test method. Will put up tomorrow.

PR #2464, opened 2024-05-16
SOLR-17307: Use file separator instead of '/' in CachingDirectoryFactory

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17307


magibney (approved):
LGTM, thanks for fixing.

I'm not able to test on Windows, but this fix would definitely stand to reason.

epugh:
@HoustonPutman  would you like me to merge this one?

HoustonPutman:
I need to make a separate JIRA so we can target 9.6.1 with it. I'll do it though, thanks for the offer 🙂

PR #2465, opened 2024-05-17
SOLR-17296: fix "explain" bugs with rerank scaling + multi-shard

hossman:
https://issues.apache.org/jira/browse/SOLR-17296

Implements the fix proposed in jira...

> I recommend we immediately:
> 
> * revert the logic added by [SOLR-16931](https://issues.apache.org/jira/browse/SOLR-16931)
> * Open a new SOLR-XXX jira for long term consideration of how to redesign debug component to get the multistage info needed for explaining reranked results 
> * harden ReRankScaler.explain so that if the dat it expects isn't available, it wraps the data it DOES have with a "0.0 == scaled reranking not available (consider using distrib.singlePass - see SOLR-XXX)"

...branch currently has one `nocommit` pertaining to what (new) jira we want to cite and link to (where we can have a longer term discussion & implementation of a fix that doesn't require using `distrib.singlePass`)

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -147,6 +147,11 @@ Other Changes
 
 * GITHUB#2454: Refactor preparePutOrPost method in HttpJdkSolrClient (Andy Webb)
 
+==================  9.6.1 ==================
+Bug Fixes
+---------------------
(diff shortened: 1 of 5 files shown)


hossman:
manually squash merged to main & backported to 9x & 9.6.1

PR #2470, opened 2024-05-20
SOLR-17302: Convert /cluster filestore APIs to JAX-RS

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-17302

# Description

Solr has slowly been modifying its v2 APIs to align around a more REST-ful model, implemented using the JAX-RS framework.  Many APIs have already been through this re-alignment, but many more remain including Solr's filestore APIs (used internally by package store).

# Solution

This commit migrates several of the filestore APIs, namely those under the `/cluster/files` path to use the JAX-RS framework.  The APIs themselves are unchanged from an end-user perspective.

Other endpoints, particularly the "fetch filestore entry" (`GET /api/node/files/somepath.txt`) and "delete local filestore entry" (`DELETE /api/node/files/somepath.txt`) are not attempted in this PR, but left for another.

# Tests

Existing filestore and package-store tests continue to pass.

# Checklist


solr/api/src/java/org/apache/solr/client/api/endpoint/ClusterFileStoreApis.java
```diff
@@ -0,0 +1,56 @@
...
+import jakarta.ws.rs.PathParam;
+import jakarta.ws.rs.QueryParam;
+import java.io.InputStream;
+import java.util.List;
+import org.apache.solr.client.api.model.SolrJerseyResponse;
+import org.apache.solr.client.api.model.UploadToFileStoreResponse;
+
+@Path("/cluster")
+public interface ClusterFileStoreApis {
+  // TODO Better understand the purpose of the 'sig' parameter and improve docs here.
```
epugh, line 34: Yeah...    There is some signing of stuff, and it probably should be a more global capability for managing uploading things.  Maybe you "sign" a configset for example.   Or treat a file the same ways we treat a configset?

solr/api/src/java/org/apache/solr/client/api/endpoint/ClusterFileStoreApis.java
```diff
@@ -0,0 +1,56 @@
...
+  @PUT
+  @Operation(
+      summary = "Upload a file to the filestore.",
+      tags = {"file-store"})
+  @Path("/files{filePath:.+}")
+  UploadToFileStoreResponse uploadFile(
+      @Parameter(description = "File store path") @PathParam("filePath") String filePath,
+      @Parameter(description = "Signature(s) for the file being uploaded") @QueryParam("sig")
```
epugh, line 42: Should `sig` be `signature` ?

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -217,7 +217,7 @@ private boolean fetchFileFromNodeAndPersist(String fromNode) {
     }
 
     boolean fetchFromAnyNode() {
-      ArrayList<String> l = coreContainer.getFileStoreAPI().shuffledNodes();
+      ArrayList<String> l = FileStoreUtils.shuffledNodes(coreContainer);
```
epugh, line 220: i know you are worried about scope, but maybe not `l`?  `liveNodes`?

epugh: or `nodes`?

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -634,7 +635,7 @@ public Map<String, byte[]> getKeys() throws IOException {
   // reads local keys file
   private static Map<String, byte[]> _getKeys(Path solrHome) throws IOException {
     Map<String, byte[]> result = new HashMap<>();
-    Path keysDir = _getRealPath(FileStoreAPI.KEYS_DIR, solrHome);
+    Path keysDir = _getRealPath(ClusterFileStore.KEYS_DIR, solrHome);
```
epugh, line 638: Sigh I hate you `_` methods and objects!

solr/core/src/java/org/apache/solr/filestore/FileStoreAPI.java
epugh, line 1: I sort of thought maybe `FileStoreAPI.java` would be totally removed in favour of the new classes?

solr/core/src/java/org/apache/solr/filestore/FileStoreUtils.java
```diff
@@ -0,0 +1,145 @@
...
+import org.apache.solr.core.CoreContainer;
+import org.apache.solr.util.CryptoKeys;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Common utilities used by filestore-related code. */
+public class FileStoreUtils {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  /** get a list of nodes randomly shuffled * @lucene.internal */
```
epugh, line 40: why `lucene.internal`?

solr/core/src/java/org/apache/solr/filestore/FileStoreUtils.java
```diff
@@ -0,0 +1,145 @@
...
+        log.error("Error reading file ", e);
+        errs.accept("Error reading file " + path + " " + e.getMessage());
+      }
+    }
+  }
+
+  /**
+   * Validate a file for signature
+   *
+   * @param sigs the signatures. atleast one should succeed
```
epugh, line 89: `@param signatures The Signatures, at least one should succeed` ?

solr/core/src/java/org/apache/solr/packagemanager/PackageUtils.java
```diff
@@ -302,7 +303,7 @@ public static String getCollectionParamsPath(String collection) {
   }
 
   public static void uploadKey(byte[] bytes, String path, Path home) throws IOException {
-    FileStoreAPI.MetaData meta = FileStoreAPI._createJsonMetaData(bytes, null);
+    FileStoreAPI.MetaData meta = ClusterFileStore._createJsonMetaData(bytes, null);
```
epugh, line 306: guessing minimizing change, ubt maybe we can remove `_` ?

solr/core/src/java/org/apache/solr/pkg/PackageAPI.java
```diff
@@ -254,7 +250,7 @@ public void refresh(PayloadObj<String> payload) {
       }
       // first refresh my own
       packageLoader.notifyListeners(p);
-      for (String s : coreContainer.getFileStoreAPI().shuffledNodes()) {
+      for (String s : FileStoreUtils.shuffledNodes(coreContainer)) {
```
epugh, line 253: better name on what is returned...?

solr/core/src/java/org/apache/solr/pkg/PackageAPI.java
```diff
@@ -426,7 +422,7 @@ private void syncToVersion(int expectedVersion) {
   }
 
   void notifyAllNodesToSync(int expected) {
-    for (String s : coreContainer.getFileStoreAPI().shuffledNodes()) {
+    for (String s : FileStoreUtils.shuffledNodes(coreContainer)) {
```
epugh, line 425: LIke wise?   what is the string...

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
epugh, line 1: Do we need DistribFileStore when we have ClusterFileStore?

epugh (commented):
Looks like some great work here!   I tagged some places where it would be nice ot have more descriptive variable names, however I suspect you are trying to minimize change.   Do we still need DistribFileStore in a world with ClusterFileStore?

gerlowskija:
Thanks for the quick review Eric!

> I tagged some places where it would be nice ot have more descriptive variable names, however I suspect you are trying to minimize change

Your suspicion is correct.  This PR aims to convert the API(s) to JAX-RS without delving too deeply into cleaning up the underlying implementation.  I think some of the stuff you commented on is innocuous enough that I can be tempted into making a few cleanups though, particularly some of the naming and javadoc/comment issues. 

> Do we still need DistribFileStore in a world with ClusterFileStore?

Yes, we do.  There's some naming-related confusion going on here that I've definitely caused:

1. `FileStore` is an interface that defines the operations that Solr's file-store has to support.
2. `DistribFileStore` is the lone implementation of the `FileStore` interface.
3. `ClusterFileStore` is the entrypoint for the relevant v2 APIs.

`ClusterFileStore` is, on its surface, a pretty bad naming choice given the pre-existing `FileStore` interface.  It 100% appears to be an implementation of that interface, but it's not.

I do have an excuse, though it may be a bad one 😛 I chose `ClusterFileStore` for the v2 API class name to be consistent with a naming convention we've had since creating the "api" gradle module earlier this year.  (i.e. interfaces in "api" module are `FooApi`, with classes in solr-core being `Foo`).

The filestore (for reasons I still don't really understand) makes APIs available under two paths: `/api/cluster/files` and `/api/node/files`.  So I chose `ClusterFileStoreApi`/`ClusterFileStore` partially in an attempt to distinguish between these two sets of filestore APIs, and partially to maintain consistency with the existing naming convention.

The end result is clearly confusing; very open if anyone has other suggestions that'd avoid the ambiguity and confusion!?

epugh:
If `DistribFileStore` is the lone implementation of `FileStore`, then do we need `FileStore`?    I never liked the whole "Let me have an interface for everything" cause it doubles the files to navigate through.   In the future, if we need a `FileStore` implementation because some wrote `NonDistribFileStore` then we can make it then...  We aren't C code, we don't need a seperate `.h` file for every `.cpp` file!

solr/api/src/java/org/apache/solr/client/api/endpoint/ClusterFileStoreApis.java
gerlowskija: Probably, or `signatures` - but I'm going to avoid cosmetic changes to the API (in this PR) as making those would drag in changes to other places the API is called.  I do aim to update those spots (to remove GenericSolrRequest usage, as we've discussed), just not trying to do it in this PR.

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
gerlowskija: I went with `nodesToAttemptFetchFrom`

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
gerlowskija: Agreed, I don't love the convention either.  There's several of these in DistribFileStore though - enough that I'm worried renaming will bloat this PR too much 😦

solr/core/src/java/org/apache/solr/filestore/FileStoreAPI.java
gerlowskija: Ultimately it will, but for scope reasons this PR only converts some of FileStoreAPI API's to JAX-RS.

I initially tried to do all the APIs at once, but the change got a bit unmanageable as time went on so I've split it into two PRs (of which this is the first).

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
epugh: Was it "errorprone" that we used to help us clean up our code and enforce some conventions?  For another JIRA, see if errorprone would flag any "_" methods for example.   We *do* need to make all our conventions enforced via automated check, there are too many people over too many years to otherwise keep code standards intact...

solr/core/src/java/org/apache/solr/filestore/FileStoreUtils.java
gerlowskija: Idk - it was in the Javadoc  when I copied this method out of FileStoreAPI.

We're not in SolrJ so the original author probably wasn't worried about signaling backcompat - my guess was that 'lucene.internal' was more about signaling that this was an implementation detail and that folks should be leery of calling the method?

I'll remove it.

solr/core/src/java/org/apache/solr/filestore/FileStoreUtils.java
gerlowskija: Done

solr/core/src/java/org/apache/solr/packagemanager/PackageUtils.java
gerlowskija: I've opted out of renaming this '_' methods, as there are several of them and I'm afraid of bloating this PR's diff.

gerlowskija:
> If DistribFileStore is the lone implementation of FileStore, then do we need FileStore?

I suspect the intention was that having an interface leaves the door open to adding a Solr-standalone filestore implementation at some point if someone is motivated enough.

I don't feel strongly on whether the interface remains or not.  Maybe that's worth cleaning up at some point, but IMO it's not necessary here since this PR isn't changing anything about how the FileStore is implemented, just some APIs around it.

gerlowskija:
> is there an argument that ClusterFileStore should do everything DistribFileStore does?

I guess?  Idk about "do everything", but ClusterFileStore/ClusterFileStoreApi will eventually _expose everything_ that DistribFileStore supports.

And I don't think that's a bad thing - it's an intentional "separation of concerns" thing.  "ClusterFileStore"/"ClusterFileStoreApi" define the HTTP interface an end-user sees, and some of the API concerns around that.  "DistribFileStore" is concerned with how we actually implement the "Filestore" idea in a SolrCloud context.  In theory, we could rewrite the filestore entirely without having to touch the API code, and that's a good thing!

The names are a mess and that definitely needs straightened out, but even in the long term IMO each of these classes has a distinct purpose that make them worth keeping around.  Totally agree about the other TODOs you mentioned in your comment above though - it'd be great if we could migrate all the BlobHandler usage onto the filestore.  (Though as you mentioned - those are different tickets for sure.)

gerlowskija:
I've since noticed two problems with this PR:

1. The generated SolrRequest class for uploading files doesn't handle the request body correctly, making it largely unusable.  We can fix this by modifying the JAX-RS annotations to be similar to those used in ZooKeeperReadAPI.
2. This breaks a test that @AndreyBozhko had in-flight on PR #2471.

I'm going to revert until I've had a chance to fix (1) and so that Andrey isn't blocked.

PR #2473, opened 2024-05-21
SOLR-16659: Properly construct V2 base urls instead of replacing substring "/solr" with "/api"

AndreyBozhko:
https://issues.apache.org/jira/browse/SOLR-16659

# Description

Update codebase to use methods that directly construct V2 urls instead of doing substring replacement on a V1 url.

I basically searched for occurrences of `replace("/solr"`, `replaceFirst("/solr"` or `replaceAll("/solr"`, and adjusted those.

# Checklist

### DIFF
--- a/solr/core/src/test/org/apache/solr/cloud/TestConfigSetsAPI.java
+++ b/solr/core/src/test/org/apache/solr/cloud/TestConfigSetsAPI.java
@@ -1581,7 +1581,7 @@ public class TestConfigSetsAPI extends SolrCloudTestCase {
       final ByteBuffer fileBytes =
           TestSolrConfigHandler.getFileContent(file.getAbsolutePath(), false);
       final String uriEnding =
-          "/api/cluster/configs/"
+          "/cluster/configs/"
               + configSetName
               + suffix
               + (!overwrite ? "?overwrite=false" : "")
@@ -1590,8 +1590,7 @@ public class TestConfigSetsAPI extends SolrCloudTestCase {
       Map<?, ?> map =
           postDataAndGetResponse(
               cluster.getSolrClient(),
-              cluster.getJettySolrRunners().get(0).getBaseUrl().toString().replace("/solr", "")
-                  + uriEnding,
+              cluster.getJettySolrRunners().get(0).getBaseURLV2().toString() + uriEnding,
               fileBytes,
               username,
               usePut);
@@ -1634,7 +1633,7 @@ public class TestConfigSetsAPI extends SolrCloudTestCase {
       final ByteBuffer sampleConfigFile =
           TestSolrConfigHandler.getFileContent(file.getAbsolutePath(), false);
       final String uriEnding =
-          "/api/cluster/configs/"
+          "/cluster/configs/"
               + configSetName
               + suffix
               + "/"
@@ -1646,8 +1645,7 @@ public class TestConfigSetsAPI extends SolrCloudTestCase {
       Map<?, ?> map =
           postDataAndGetResponse(
               cluster.getSolrClient(),
-              cluster.getJettySolrRunners().get(0).getBaseUrl().toString().replace("/solr", "")
-                  + uriEnding,
+              cluster.getJettySolrRunners().get(0).getBaseURLV2().toString() + uriEnding,
               sampleConfigFile,
               username,
               usePut);
--- a/solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
+++ b/solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
@@ -572,7 +572,7 @@ public class TestContainerPlugin extends SolrCloudTestCase {
 
   public void waitForAllNodesToSync(String path, Map<String, Object> expected) throws Exception {
     for (JettySolrRunner jettySolrRunner : cluster.getJettySolrRunners()) {
-      String baseUrl = jettySolrRunner.getBaseUrl().toString().replace("/solr", "/api");
+      String baseUrl = jettySolrRunner.getBaseURLV2().toString();
       String url = baseUrl + path + "?wt=javabin";
       TestDistribFileStore.assertResponseValues(1, new Fetcher(url, jettySolrRunner), expected);
     }
--- a/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
+++ b/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
@@ -119,7 +119,7 @@ public class ZkStateReader implements SolrCloseable {
 
   /**
    * This ZooKeeper file is no longer used starting with Solr 9 but keeping the name around to check
-   * if it is still present and non empty (in case of upgrade from previous Solr version). It used
+   * if it is still present and non-empty (in case of upgrade from previous Solr version). It used
    * to contain collection state for all collections in the cluster.
    */
   public static final String UNSUPPORTED_CLUSTER_STATE = "/clusterstate.json";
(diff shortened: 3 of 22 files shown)


solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClient.java
```diff
@@ -374,7 +374,7 @@ protected HttpRequestBase createMethod(SolrRequest<?> request, String collection
 
     if (request instanceof V2Request) {
       if (System.getProperty("solr.v2RealPath") == null || ((V2Request) request).isForceV2()) {
-        basePath = baseUrl.replace("/solr", "/api");
+        basePath = changeV2RequestEndpoint(baseUrl);
```
AndreyBozhko, line 377: Not sure if this needs to be updated as HttpSolrClient class is deprecated

solr/core/src/test/org/apache/solr/handler/TestContainerPlugin.java
```diff
@@ -572,7 +572,7 @@ private V2Request postPlugin(Object payload) {
 
   public void waitForAllNodesToSync(String path, Map<String, Object> expected) throws Exception {
     for (JettySolrRunner jettySolrRunner : cluster.getJettySolrRunners()) {
-      String baseUrl = jettySolrRunner.getBaseUrl().toString().replace("/solr", "/api");
+      String baseUrl = jettySolrRunner.getBaseURLV2().toString();
```
epugh, line 575: so many `.toString()` ;-).   I wonder...    Should we be working more with URL objects and less with String objects when we build our urls???

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -1331,12 +1331,24 @@ public ConfigData getSecurityProps(boolean getFresh) {
    * Returns the baseURL corresponding to a given node's nodeName -- NOTE: does not (currently)
    * imply that the nodeName (or resulting baseURL) exists in the cluster.
    *
-   * @lucene.experimental
```
epugh, line 1334: nice clean up!

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -1331,12 +1331,24 @@ public ConfigData getSecurityProps(boolean getFresh) {
...
 
+  /**
+   * Returns the V2 baseURL corresponding to a given node's nodeName -- NOTE: does not (currently)
+   * imply that the nodeName (or resulting baseURL) exists in the cluster.
+   *
+   * @param nodeName name of the node
+   * @return url that looks like {@code https://localhost:8983/api}
+   */
+  public String getBaseUrlV2ForNodeName(final String nodeName) {
+    return Utils.getBaseUrlForNodeName(nodeName, getClusterProperty(URL_SCHEME, "http"), true);
```
epugh, line 1349: is it just me, or does this method look exactly like the one in `getBaseUrlForNodeName`???   Where do we do the `/api` conversion?

epugh: Looking up to line 1338

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClient.java
epugh: unfortunately, until we get `HttpSolrClient` gone, we probably should because it is used in so many places.  We really DO need to get the migration done in Solr 10 so we can cut down on the surface of changes every time we touch an implementation.

epugh (changes requested):
Great refactoring!  @gerlowskija you may want to look as well since you are in the space.

One thing?  I flagged that the `getBaseUrlV2ForNodeName` appears to be the same as `getBaseUrlForNodeName` method???

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
AndreyBozhko: It's actually a different overload of the method which accepts `boolean isV2` parameter. But I agree - perhaps it's more clear to also use the same overload in line 1338, and pass `isV2 = false` explicitly.

https://github.com/apache/solr/blob/3837eebcd73e103c083d7b02bdf6904d08fc34b5/solr/solrj/src/java/org/apache/solr/common/util/Utils.java#L719-L738

epugh: yeah...  Or maybe some javadocs to explain it?

epugh: Let me know when you decide what you want to do here, and then i look forward to merging it!

AndreyBozhko: I added javadoc to the methods in the Utils class, and refactored the code a little in 50c959f

epugh: thanks!

solr/core/src/test/org/apache/solr/cloud/TestConfigSetsAPI.java
```diff
@@ -1590,8 +1590,7 @@ private long uploadGivenConfigSet(
       Map<?, ?> map =
           postDataAndGetResponse(
               cluster.getSolrClient(),
-              cluster.getJettySolrRunners().get(0).getBaseUrl().toString().replace("/solr", "")
-                  + uriEnding,
+              cluster.getJettySolrRunners().get(0).getBaseURLV2().toString() + uriEnding,
```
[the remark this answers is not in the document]
gerlowskija: > Just wondering if it does it implicitly.

It doesn't 😦 

@epugh I agree with what I think is your implicit point (and with a point you made explicitly on one of my recent PRs haha) -  that we should add a variant of these "getUrl" methods that returns a string instead of a URL, since this is such a common idiom.

But let's handle that as a separate PR.

solr/solrj/src/java/org/apache/solr/common/util/Utils.java
```diff
@@ -716,10 +716,30 @@ public static String applyUrlScheme(final String url, final String urlScheme) {
     return (at == -1) ? (urlScheme + "://" + url) : urlScheme + url.substring(at);
   }
 
+  /**
```
gerlowskija, line 719: [+1] Awesome Javadocs, thanks for adding these.

[0] Definitely not a job for this PR, but I think we could make SolrJ easier to understand if we were a bit tighter in our URL/path terminology.    "Base" means a lot of different things to a lot of different people.

You could imagine defining terms like "root URL" = `scheme://host:port/$contextRoot`, "API base URL" = `$root/api` (v2), `$root/solr` (v1), "collection URL" = ...

Just thinking aloud.  This is all very minor, but I've spent a lot of time lately looking at how SolrJ builds these APIs so it's on my mind.

gerlowskija (approved):
This is awesome; thanks @AndreyBozhko !

dsmiley:
It feel weird to me that we go to ZkStateReader to build node URLs.  Conceptually, this seem like something that ought to exist at a higher conceptual layer.  On the other hand, ZkStateReader has become a kitchen-sink of anything relating to SolrCloud, and increasingly more and more things assume SolrCloud.  Yet its name has a low-level sounding name.  Hmm.

PR #2478, opened 2024-05-24
fix Lucene[95-->99] typo in SchemaCodecFactory javadocs

cpoerschke:
Okay, I could have just pushed this typo fix directly but wanted to take opportunity to briefly wonder about the (lack of a?) connection between the Lucene `Codec` returned here and the `luceneMatchVersion` in `solrconfig.xml` configuration.


solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java
```diff
@@ -151,7 +151,7 @@ public Codec getCodec() {
   }
 
   /**
-   * This class exists because Lucene95HnswVectorsFormat's getMaxDimensions method is final and we
+   * This class exists because Lucene99HnswVectorsFormat's getMaxDimensions method is final and we
```
dsmiley, line 154: why not directly reference it using `{@link ...}`.  It's take less chars overall!  (no need to refer to "method")

solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java
```diff
@@ -151,8 +151,9 @@ public Codec getCodec() {
   }
 
   /**
-   * This class exists because Lucene95HnswVectorsFormat's getMaxDimensions method is final and we
-   * need to workaround that constraint to allow more than the default number of dimensions
+   * This class exists because {@link Lucene99HnswVectorsFormat#getMaxDimensions(String)} method is
+   * final and we need to workaround that constraint to allow more than the default number of
```
cpoerschke, line 155: Actually the method is not (no longer?) final: https://github.com/apache/lucene/blob/releases/lucene/9.10.0/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java#L220

dsmiley (approved):

PR #2479, opened 2024-05-25
SOLR-14673: Add bin/solr stream CLI 

epugh:
https://issues.apache.org/jira/browse/SOLR-14673

# Description

Bring in code that @joel-bernstein wrote, but using the SolrCLI infrastructure.

# Solution

Another CLI client ;-)

# Tests

Copied over the basic tests from the patch.  I still need to write an integration style test and ideally one that exercies the basic auth.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cli/StreamTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/StreamTool.java
@@ -1,0 +1,435 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.cli;
+
+import java.io.BufferedReader;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.LineNumberReader;
+import java.io.PrintStream;
+import java.nio.charset.Charset;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.Option;
+import org.apache.solr.client.solrj.io.Lang;
+import org.apache.solr.client.solrj.io.SolrClientCache;
+import org.apache.solr.client.solrj.io.Tuple;
+import org.apache.solr.client.solrj.io.comp.StreamComparator;
+import org.apache.solr.client.solrj.io.stream.LetStream;
+import org.apache.solr.client.solrj.io.stream.PushBackStream;
+import org.apache.solr.client.solrj.io.stream.StreamContext;
+import org.apache.solr.client.solrj.io.stream.TupleStream;
+import org.apache.solr.client.solrj.io.stream.expr.Explanation;
+import org.apache.solr.client.solrj.io.stream.expr.Expressible;
+import org.apache.solr.client.solrj.io.stream.expr.StreamExpression;
+import org.apache.solr.client.solrj.io.stream.expr.StreamExpressionParser;
+import org.apache.solr.client.solrj.io.stream.expr.StreamFactory;
+
+/** Supports stream command in the bin/solr script. */
+public class StreamTool extends ToolBase {
+
+    // not sure this is right!!!
+    public static final String OUTPUT_FIELDS = "fields";
+
+    public StreamTool() {
+        this(CLIO.getOutStream());
+    }
+
+    public StreamTool(PrintStream stdout) {
+        super(stdout);
+    }
+
+    @Override
+    public String getName() {
+        return "stream";
+    }
+
+    @Override
+    public List<Option> getOptions() {
+        return List.of(
+                SolrCLI.OPTION_ZKHOST,
+                SolrCLI.OPTION_SOLRURL,
+                Option.builder("f")
+                        .longOpt("fields")
+                        .argName("FIELDS")
+                        .hasArg()
+                        .required(false)
+                        .desc("The fields in the tuples to output. (defaults to fields in the first tuple of result set).")
+                        .build(),
+                Option.builder("h")
+                        .longOpt("header")
+                        .required(false)
+                        .desc("Whether or not to include a header line. (default=false)")
+                        .build(),
+                Option.builder("d")
+                        .longOpt("delimiter")
+                        .argName("CHARACTER")
+                        .hasArg()
+                        .required(false)
+                        .desc("The output delimiter. (default=tab).")
+                        .build(),
+                Option.builder("a")
+                        .longOpt("array-delimiter")
+                        .argName("CHARACTER")
+                        .hasArg()
+                        .required(false)
+                        .desc("The delimiter multi-valued fields. (default=|)")
+                        .build(),
+
+                SolrCLI.OPTION_CREDENTIALS,
+                SolrCLI.OPTION_VERBOSE);
+    }
+
+    @Override
+    @SuppressWarnings({"rawtypes"})
+    public void runImpl(CommandLine cli) throws Exception {
+        SolrCLI.raiseLogLevelUnlessVerbose(cli);
+
+        String zkHost = SolrCLI.getZkHost(cli);
+        String[] outputHeaders;
+        String arrayDelim = cli.getOptionValue("array-delimiter",  "|");
+        String delimiter = cli.getOptionValue("delimiter",  "   ");
+        boolean includeHeaders = cli.hasOption("header");
+        // not sure I need verbose however.
+        verbose = cli.hasOption(SolrCLI.OPTION_VERBOSE.getOpt());
+
+
+
+        String exprFile = cli.getArgs()[0];
+        SolrClientCache solrClientCache = null;
+        TupleStream stream;
+        PushBackStream pushBackStream = null;
+        LineNumberReader bufferedReader = null;
+
+
+
+        // likewise not sure...
+        PrintStream filterOut = new PrintStream(System.err) {
+            public void println(String l) {
+                if (! l.startsWith("SLF4J")) {
+                    super.println(l);
+                }
+            }
+        };
+
+        System.setErr(filterOut);
+
+        try {
+
+            bufferedReader = new LineNumberReader(new InputStreamReader(new FileInputStream(exprFile), Charset.defaultCharset()));
+            String expr = readExpression(bufferedReader, cli.getArgs());
+            StreamExpression streamExpression = StreamExpressionParser.parse(expr);
+            StreamFactory streamFactory = new StreamFactory();
+
+
+
+            // do i need this?
+            streamFactory.withDefaultZkHost("localhost:8983");
+
+            Lang.register(streamFactory);
+
+            streamFactory.
+                    withFunctionName("stdin", StandardInStream.class);
+            stream = constructStream(streamFactory, streamExpression);
+
+
+            Map params = validateLetAndGetParams(stream, expr);
+
+            streamFactory.withDefaultZkHost(zkHost);
+            outputHeaders = getOutputFields(params);
+
+
+            pushBackStream = new PushBackStream(stream);
+
+            //Now we can run the stream and return the results.
+            solrClientCache = new SolrClientCache();
+
+            if(zkHost != null) {
+                echoIfVerbose("Connecting to ZooKeeper at " + zkHost, cli);
+                long start = System.currentTimeMillis();
+                solrClientCache.getCloudSolrClient(zkHost);
+                long end = System.currentTimeMillis();
+                echoIfVerbose("Connected in " + (end - start), cli);
+            }
+
+            StreamContext streamContext = new StreamContext();
+            streamContext.setSolrClientCache(solrClientCache);
+            System.setProperty("COMMAND_LINE_EXPRESSION", "true");
+            //Output the headers
+
+
+            if(verbose) {
+                CLIO.err("");
+                CLIO.err("CLI -- Running Expression: " +  expr +  " ...");
+                CLIO.err("");
+            }
+
+            pushBackStream.setStreamContext(streamContext);
+
+            pushBackStream.open();
+
+            if(outputHeaders == null) {
+
+                Tuple tuple = pushBackStream.read();
+
+                if(!tuple.EOF) {
+                    outputHeaders = getHeadersFromFirstTuple(tuple);
+                }
+
+                pushBackStream.pushBack(tuple);
+            }
+
+            if(includeHeaders) {
+                StringBuilder headersOut = new StringBuilder();
+                for (int i = 0; i < outputHeaders.length; i++) {
+                    if (i > 0) {
+                        headersOut.append(delimiter);
+                    }
+                    headersOut.append(outputHeaders[i]);
+                }
+                CLIO.out(headersOut.toString());
+            }
+
+            while (true) {
+                Tuple tuple = pushBackStream.read();
+                if (tuple.EOF) {
+                    break;
+                } else {
+                    StringBuilder outLine = new StringBuilder();
+                    for (int i = 0; i < outputHeaders.length; i++) {
+                        if (i > 0) {
+                            outLine.append(delimiter);
+                        }
+
+                        Object o = tuple.get(outputHeaders[i]);
+                        if (o != null) {
+                            if(o instanceof List) {
+                                List outfields = (List)o;
+                                outLine.append(listToString(outfields, arrayDelim));
+                            } else {
+                                outLine.append(o);
+                            }
+                        }
+                    }
+                    CLIO.out(outLine.toString());
+                }
+            }
+        } finally {
+
+            if(verbose) {
+                CLIO.err("");
+                CLIO.err("CLI -- Expression complete");
+                CLIO.err("CLI -- Cleaning up resources ...");
+            }
+
+            if(pushBackStream != null) {
+                pushBackStream.close();
+            }
+
+            if(bufferedReader != null) {
+                bufferedReader.close();
+            }
+
+            if(solrClientCache != null) {
+                solrClientCache.close();
+            }
+
+            if(verbose) {
+                CLIO.err("CLI -- Done.");
+                CLIO.err("");
+            }
+        }
+
+
+    }
+
+    public static class StandardInStream extends TupleStream implements Expressible {
+
+        private BufferedReader reader;
+        private InputStream inputStream = System.in;
+        private boolean doClose = false;
+
+        public StandardInStream() {
+
+        }
+
+        public StandardInStream(StreamExpression expression, StreamFactory factory) throws IOException {
+
+        }
+
+        public List<TupleStream> children() {
+            return null;
+
+        }
+
+        public void setInputStream(InputStream inputStream) {
+            this.inputStream = inputStream;
+            this.doClose = true;
+        }
+
+        public void open() {
+            reader = new BufferedReader(new InputStreamReader(inputStream));
+        }
+
+        public void close() throws IOException {
+            if(doClose) {
+                inputStream.close();
+            }
+        }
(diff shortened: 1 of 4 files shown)


gerlowskija:
A few high-level questions/concerns:

1. `bin/solr` already has an "api" tool, which can be used to invoke streaming expressions e.g. `bin/solr api -get "$SOLR_URL/techproducts/stream?expr=search(techproducts)"`. I'm all for syntactic-sugar, but I wonder whether this is worth the maintenance cost if the main thing that it "buys" us is saving people from having to provide the full API path as the "api" tool requires?
2. If I'm reading the PR correctly, it looks like one other capability of the proposed `bin/solr stream` tool is that it can evaluate streams "locally" in some cases i.e. without a full running Solr.  Which is pretty cool - you could imagine a real super-user doing some pretty involved ETL that builds off of an expression like: `update(techproducts, unique(cat(...)))`.  

    But I'd worry about some of the documentation challenges surrounding this.  For instance, how would a user know which expressions can be run locally, and which require a Solr to execute on?  For expressions that have a mix of both locally and remotely-executed clauses, is there any way for a user to know which clauses are executed where?
    
    To clarify - I think the upside here is pretty cool, I'm just worried about the documentation end and what we might need to make it usable by folks in practice.

epugh:
Thanks for sharing the feedback @gerlowskija !   I think the value of the tool is only there if your second comment about being able to run a streaming expression locally is valid, and then having it do what yoru first comment highlights falls out easy, otherwise it really is a thin wrapper/duplication of the `bin/solr api` call.   Especially without any special value add in formatting tuples or error handling etc.

I do believe the second part is the really cool thing, that I can run a streaming expression locally and use it to process some data.   

We clearly need some way of specifying where the processing is happening, in the cluster or locally.   I was trying to think if we have any other places in Solr where we define "Where am I doing work" that might provide a name for a parameter.   `bin/solr stream --environment cluster BLAH` ?   The `search()` expression has a `qt` parameter..   `bin/solr stream -qt=/stream BLAH` ?    

Reading through docs more, we have the `parralel()` and it refers to workers.    Maybe the command should be something like `bin/solr stream --workers=local BLAH` which would run on your laptop, and if you don't specify --workers then it runs on the cluster via `/stream`?   

I have found that lots of streaming expressions don't require a Solr connection, especially during development.  I'm just iterating on the logic, and I'm starting and ending iwth tuples..   it's only later when I get the mappings etc working that I then move to adding in my `search()` or `update()` clauses.

Also, as far as docs go, we have a LONG way to go in Streaming expressions.   It's both the best docuemnted code, with all the howtos and guides, but also, I find a million expressions that exist but don't show up in our reference docs ;-).

epugh:
Okay, I think this is ready for review!  I've added some docs..   I especially liked being able to cat some local data right into a Solr collection!   

```
cat example/exampledocs/books.csv | bin/solr stream -e local 'update(gettingstarted,parseCSV(stdin()))'
```

epugh:
@gerlowskija since you provided some early review, do you think the docs I've added etc are enough that I can merge this in?

gerlowskija (commented):
Thanks for those added docs @epugh .  They're a huge help, and I suggested a few tweaks that might help even more.

One remaining question I have - wdyt about marking the tool syntax as "experimental" in some way?  Seeing all the hard work you've put into improving syntax on the other tools, and considering that we might not notice some rough edges to the syntax of this tool until it's out in the wild a bit...might be prudent to give this script equivalent of "@lucene.experimental" so that we wouldn't need to worry about backcompat if we want to make any future tweaks?

solr/solr-ref-guide/modules/query-guide/pages/stream-tool.adoc
```diff
@@ -0,0 +1,167 @@
...
+ -c,--name <NAME>                   Name of the collection to execute on if workers are 'solr'.  Required for 'solr'
+                                    worker.
+    --delimiter <CHARACTER>         The output delimiter. (default=tab).
+ -f,--fields <FIELDS>               The fields in the tuples to output. (defaults to fields in the first tuple of result
+                                    set).
+    --header                        Whether or not to include a header line. (default=false)
+ -u,--credentials <credentials>     Credentials in the format username:password. Example: --credentials solr:SolrRocks
+ -url,--solr-url <HOST>             Base Solr URL, which can be used to determine the zk-host if that's not known;
+                                    defaults to: http://localhost:8983.
+ -e,--execution <CONTEXT>           Execution context is either 'local' or 'solr'. Default is 'solr'
```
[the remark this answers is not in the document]
epugh: Thoughts on: 

```
Execution context is either 'local' (i.e CLI process) or 'solr'.
```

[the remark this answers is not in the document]
gerlowskija: I noodled on this a bit but still couldn't come up with anything I liked better, so maybe you picked the best option.  +1 for the current enum values with the "(i.e. CLI process)" clarification you suggested.

solr/solr-ref-guide/modules/query-guide/pages/stream-tool.adoc
```diff
@@ -71,13 +72,15 @@ CORSAIR ValueSelect 1GB 184-Pin DDR SDRAM Unbuffered DDR 400 (PC 3200) System Me
 A-DATA V-Series 1GB 184-Pin DDR SDRAM Unbuffered DDR 400 (PC 3200) System Memory - OEM
 ----
 
-The `--help` (or simply `-h`) option will output information on its usage (i.e., `bin/solr stream --help)`.
-
 == Using the bin/solr stream Tool
 
+NOTE: The Stream Tool is classified as "experimental".
```
gerlowskija, line 77: Might be worth putting this at the top of the page.  Seems a little "buried" as-is.

gerlowskija (approved):
LGTM.

Left one last comment about the "experimental" designation, but feel free to ignore.

solr/solr-ref-guide/modules/query-guide/pages/stream-tool.adoc
epugh: love it

solr/solr-ref-guide/modules/query-guide/pages/stream-tool.adoc
```diff
@@ -0,0 +1,167 @@
...
+
+update(
+  gettingstarted,
+  parseCSV(
+    cat(./example/exampledocs/books.csv, maxLines=2)
+  )
+)
+----
+
+Running this expression will read in the local file and send the first two lines to the collection `gettingstarted`.
```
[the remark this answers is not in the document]
epugh: okay, I took a stab at it...   If I didn't quite nail it, please edit it but with myabe the suggestion thing?

solr/core/src/java/org/apache/solr/cli/StreamTool.java
```diff
@@ -0,0 +1,507 @@
...
+            .hasArg()
+            .required(false)
+            .desc(
+                "The fields in the tuples to output. (defaults to fields in the first tuple of result set).")
+            .build();
+
+  private static final Option HEADER_OPTION =
+          Option.builder()
+                  .longOpt("header")
+                  .required(false)
```
malliaridis: Any `.required(false)` is obsolete and could be removed.

solr/core/src/java/org/apache/solr/cli/StreamTool.java
```diff
@@ -0,0 +1,507 @@
...
+            .argName("CHARACTER")
+            .hasArg()
+            .required(false)
+            .desc("The delimiter multi-valued fields. (default=|)")
+            .build();
+
+  @Override
+  public Options getOptions() {
+    Options opts =
+            super.getOptions()
```
malliaridis: I think you could simplify it by directly returning the value of `opts` here. Prevents any "surprises" further down this method.

solr/core/src/java/org/apache/solr/cli/StreamTool.java
```diff
@@ -0,0 +1,507 @@
...
+
+
+  @Override
+  @SuppressWarnings({"rawtypes"})
+  public void runImpl(CommandLine cli) throws Exception {
+
+    String expressionArgument = cli.getArgs()[0];
+    String execution = cli.getOptionValue(EXECUTION_OPTION,"solr");
+    String arrayDelimiter = cli.getOptionValue(ARRAY_DELIMITER_OPTION,"|");
+    String delimiter = cli.getOptionValue(DELIMITER_OPTION, "   ");
```
malliaridis, line 152: Are multiple spaces as delimeter fine / expected?

solr/core/src/java/org/apache/solr/cli/StreamTool.java
```diff
@@ -0,0 +1,507 @@
...
+    boolean includeHeaders = cli.hasOption(HEADER_OPTION);
+    String[] outputHeaders = getOutputFields(cli);
+
+    LineNumberReader bufferedReader = null;
+    String expr;
+    try {
+      Reader inputStream =
+          expressionArgument.toLowerCase(Locale.ROOT).endsWith("expr")
+              ? new InputStreamReader(
+                  new FileInputStream(expressionArgument), Charset.defaultCharset())
```
malliaridis, line 162: If this loads an exprssion from a file, would it not be better to use ".expr"  instead of "expr"?

solr/core/src/java/org/apache/solr/cli/StreamTool.java
```diff
@@ -0,0 +1,507 @@
...
+              : new StringReader(expressionArgument);
+
+      bufferedReader = new LineNumberReader(inputStream);
+      expr = StreamTool.readExpression(bufferedReader, cli.getArgs());
+      echoIfVerbose("Running Expression: " + expr);
+    } finally {
+      if (bufferedReader != null) {
+        bufferedReader.close();
+      }
+    }
```
malliaridis, line 172: There is a lot happening in the `runImpl`, maybe extracting the two blocks into separate private methods would keep a better overview? (this is block no 1)

solr/core/src/java/org/apache/solr/cli/StreamTool.java
```diff
@@ -0,0 +1,507 @@
...
+        }
+      }
+    } finally {
+
+      if (pushBackStream != null) {
+        pushBackStream.close();
+      }
+
+      solrClientCache.close();
+    }
```
malliaridis, line 241: This is block no 2

solr/core/src/java/org/apache/solr/cli/StreamTool.java
```diff
@@ -0,0 +1,507 @@
...
+
+    echoIfVerbose("StreamTool -- Done.");
+  }
+
+  public PushBackStream doLocalMode(CommandLine cli, String expr) throws Exception {
```
malliaridis, line 246: I believe these methods could all be private if the tests do not directly access them.

solr/core/src/java/org/apache/solr/cli/StreamTool.java
```diff
@@ -0,0 +1,507 @@
...
+    // Output the headers
+    pushBackStream.setStreamContext(streamContext);
+
+    return pushBackStream;
+  }
+
+  public PushBackStream doRemoteMode(CommandLine cli, String expr) throws Exception {
+
+    String solrUrl = SolrCLI.normalizeSolrUrl(cli);
+    if (!cli.hasOption("name")) {
```
malliaridis, line 289: Use option object instead for consistency.

solr/core/src/java/org/apache/solr/cli/StreamTool.java
```diff
@@ -0,0 +1,507 @@
...
+      throw new IllegalStateException(
+          "You must provide --name COLLECTION with --worker solr parameter.");
+    }
+    String collection = cli.getOptionValue("name");
```
malliaridis, line 293: Use option object instead for consistency.

solr/core/src/java/org/apache/solr/cli/StreamTool.java
```diff
@@ -0,0 +1,507 @@
...
+
+    if (expr.toLowerCase(Locale.ROOT).contains("stdin(")) {
+      throw new IllegalStateException(
+          "The stdin() expression is only usable with --worker local set up.");
+    }
+
+    final SolrStream solrStream =
+        new SolrStream(solrUrl + "/solr/" + collection, params("qt", "/stream", "expr", expr));
```
malliaridis, line 301: Don't like string concats for URLs, because we get into "v1 / v2" dilemma, but I am not sure if there is another way for this?

solr/core/src/java/org/apache/solr/cli/StreamTool.java
```diff
@@ -0,0 +1,507 @@
...
+    }
+
+    @Override
+    public StreamComparator getStreamSort() {
+      return null;
+    }
+  }
+
+  public static String[] getOutputFields(CommandLine cli) {
+    if (cli.hasOption("fields")) {
```
malliaridis, line 389: Use option object instead for consistency.

solr/core/src/java/org/apache/solr/cli/StreamTool.java
```diff
@@ -0,0 +1,507 @@
...
+
+      String fl = cli.getOptionValue("fields");
```
malliaridis, line 391: Use option object instead for consistency.

solr/core/src/java/org/apache/solr/cli/StreamTool.java
```diff
@@ -0,0 +1,507 @@
...
+
+    echoIfVerbose("StreamTool -- Done.");
+  }
+
+  public PushBackStream doLocalMode(CommandLine cli, String expr) throws Exception {
+    String zkHost = SolrCLI.getZkHost(cli);
+
+    echoIfVerbose("Connecting to ZooKeeper at " + zkHost);
+    solrClientCache.getCloudSolrClient(zkHost);
+    solrClientCache.setBasicAuthCredentials(cli.getOptionValue(CommonCLIOptions.CREDENTIALS_OPTION));
```
malliaridis: What about user-managed mode? Is it still supported, or is only cloud mode supported yet? If I am not mistaken `getConnectionOptions` include a solr-url as well.

solr/benchmark/.factorypath
```diff
@@ -0,0 +1,6 @@
+<factorypath>
+    <factorypathentry kind="EXTJAR" id="/Users/epugh/.gradle/caches/modules-2/files-2.1/org.apache.commons/commons-math3/3.6.1/e4ba98f1d4b3c80ec46392f25e094a6a2e58fcbf/commons-math3-3.6.1.jar" enabled="true" runInBatchMode="false"/>
+    <factorypathentry kind="EXTJAR" id="/Users/epugh/.gradle/caches/modules-2/files-2.1/org.openjdk.jmh/jmh-generator-annprocess/1.37/da93888682df163144edf9b13d2b78e54166063a/jmh-generator-annprocess-1.37.jar" enabled="true" runInBatchMode="false"/>
+    <factorypathentry kind="EXTJAR" id="/Users/epugh/.gradle/caches/modules-2/files-2.1/net.sf.jopt-simple/jopt-simple/5.0.4/4fdac2fbe92dfad86aa6e9301736f6b4342a3f5c/jopt-simple-5.0.4.jar" enabled="true" runInBatchMode="false"/>
+    <factorypathentry kind="EXTJAR" id="/Users/epugh/.gradle/caches/modules-2/files-2.1/org.openjdk.jmh/jmh-core/1.37/896f27e49105b35ea1964319c83d12082e7a79ef/jmh-core-1.37.jar" enabled="true" runInBatchMode="false"/>
+</factorypath>
```
malliaridis, line 6: I believe you didn't mean to push this file?

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/stream/LetStream.java
```diff
@@ -224,4 +224,9 @@ public StreamComparator getStreamSort() {
   public int getCost() {
     return 0;
   }
+
+  @SuppressWarnings({"rawtypes"})
+  public Map getLetParams() {
+    return this.letParams;
+  }
```
malliaridis, line 231: This method has no usage, is it needed? And if so, should we not simply add the types to `Map<String, Object>` rather than suppressing them?

malliaridis (changes requested):
I'll try the stream command next, but here are a few points to look into while I'm trying stream expressions out. :)

malliaridis:
A few more notes:
- [ ] `bin/solr stream --help` does not include any reference to the stream expression argument required at position 0 (or any other position that is valid)
- [ ] When providing an invalid stream expression, error message is not helpful
   ```sh
    bin\solr.cmd stream "search(q=*:*)" -e local -s http://localhost:8983 -c techproducts
    ```
  (missing collection in expression) returns "Unable to construct instance of org.apache.solr.client.solrj.io.stream.SearchFacadeStream"
    - I would like a message that at least says something like "double check your stream expression" if it is one of the causes
- [ ] For my understanding, is it a technical limitation that `-e local` does not require a collection (`-c`), but `-e solr` requires one?
- [ ] I also noticed that providing `-s` requires a URI scheme, but `-z` does not allow one. Is this inconsistency for all commands?

solr/core/src/java/org/apache/solr/cli/StreamTool.java
malliaridis: Apparently, this works for both cloud and user-managed. Strange, but it is what it is.

solr/core/src/java/org/apache/solr/cli/StreamTool.java
epugh: Thanks!

solr/core/src/java/org/apache/solr/cli/StreamTool.java
epugh: we could.. yet we have this pattern everywehre!

solr/core/src/java/org/apache/solr/cli/StreamTool.java
epugh: I think we said tab!   So trying it with a tab.  Yet, it could be just two spaces...

solr/core/src/java/org/apache/solr/cli/StreamTool.java
epugh: sure!  Might make it more readable.

epugh:
@gerlowskija @malliaridis  What do you think about using the word "environment" versus "context" for the option you pass for execution?   I.e, your environment is either local or solr...   Context to me means more like "this is athing that you set up that has lots of variables"....  "enviroment" is where you run it...?

solr/core/src/java/org/apache/solr/cli/StreamTool.java
malliaridis: These are space characters (3 spaces to be precise), I am not sure if this is intended if you wanted a tab? IDEs usually convert tabs to spaces by default.

epugh:
Leaning towards "execution" and either "local" or "remote".

epugh:
@malliaridis in reference to your question, yeah -z never has a scheme, but -s always does.  Why exactly, not totally sure..  I guess that ZK communciates over same port regardless of ssl or not?

solr/core/src/java/org/apache/solr/cli/StreamTool.java
epugh: only cloud mode for stream...

solr/core/src/java/org/apache/solr/cli/StreamTool.java
epugh: i think for expressions they historically have had `.expr`!

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/SolrClientCache.java
```diff
@@ -55,6 +50,8 @@ public class SolrClientCache implements Closeable {
   private static final int minSocketTimeout =
       Math.max(Integer.getInteger(HttpClientUtil.PROP_SO_TIMEOUT, MIN_TIMEOUT), MIN_TIMEOUT);
 
+  private String basicAuthCredentials = null; // Only support with the http2SolrClient
```
dsmiley, line 53: @epugh  I don't agree with this because it's basically a one-off setting when there are many other potential settings.  I don't think basic auth is so special or so extremely widely used that it deserves its very own setter.  It sets an unfortunate precedent from someone interesting in setting something else.  The SolrClientCache _already_ provides a constructor that accepts an existing HTTP client , and that which can be initialized with basic auth.

epugh: I could get behind that.  It's not always obvious which settings are done on the http client level and which are at the solr level ;-).

dsmiley: >  It's not always obvious which settings are done on the http client level and which are at the solr level ;-).

Hmm; I'm not sure how that's relevant to this situation.  If you create an Http2SolrClient (with basic auth and whatever else) and pass it to SolrClientCache, it'll be used completely not partially.

I'm marking this setter deprecated in an upcoming PR loosely related to at least signal we don't want this.

PR #2482, opened 2024-05-27
Adding Powershell snippets to Five Minutes Tutorial

mkhludnev:
# Description

Adding Powershell snippets. FWIW PowerShell is a kind of bash in MS Windows. 
I've checked these commands on `main` devSlim. 
They are little bit raw now. Also, I don't know how to represent tabs in *.adoc.


epugh:
you know, if we got the tabs working, this actually would be nice to add..  At first I thought having Powershell implied a lot of change and maintenance burden...   But this looks pretty straightforward.   And it's not like these commands and therefore the work to keep updated in ref guide are changing often....

epugh:
BTW, I think this was a nice idea, and maybe in the future we'll add it if we see more folks using powershell.   I'm also thinking that in the future as our CLI maybe gets richer, we may find that it's possible to use the CLI for all the thigns int he tutorial, and not need language specific examples...

PR #2484, opened 2024-05-29
SOLR-16093: Tests: don't require IPv6

dsmiley:
The Java VM/host, doesn't always support IPv6, our tests shouldn't require it. security.policy: Removed 3 lines that were for tests.

https://issues.apache.org/jira/browse/SOLR-16093

In a fork of Solr that I use at work, these changes are what was needed to get the build passing finally.
I wasn't thinking of adding to CHANGES.txt over build/test matters like this.


dsmiley:
One failing test; [reported it](https://issues.apache.org/jira/browse/SOLR-16654?focusedCommentId=17850547&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17850547) as it's a recurring failure.

I'll merge this in a couple days if I get no further feedback.

PR #2485, opened 2024-05-29
SOLR-17313: Remove deprecated class SolrLogPostTool

epugh:
https://issues.apache.org/jira/browse/SOLR-17313

# Description

Class is no longer called by any shell scripts and is just lingering.

# Solution

Removed

# Tests

Run


solr/CHANGES.txt
```diff
@@ -67,6 +67,8 @@ Deprecation Removals
 
 * SOLR-17284: Remove deprecated BlobRepository in favour of FileStore. (Eric Pugh)
 
+* SOLR-17313: Remove deprecated class SolrLogPostTool.  PostLogsTool and the bin/solr post command has replaced this. (Eric Pugh)
```
dsmiley, line 70: I argue we needn't cluster CHANGES.txt with a change of this nature.  Nobody will care about this.  No feature was removed; it was an internal change of where the functionality lives.

PR #2486, opened 2024-05-29
Move setShardAttributesToParams

dsmiley:
From ModifiableSolrParams to SearchHandler.  Too special-purpose to warrant any back-compat concern.  Was here since 2023-06


epugh (approved):
Lgtm

PR #2488, opened 2024-05-31
Add `alwaysStopwords` option to `edismax` so its "all stopwords" behaviour can be controlled

andywebb1975:
_work in progress!_ If anyone sees this and has thoughts on it, please comment below :-)

# Description

We were surprised by `edismax`'s behaviour for pure-stopword queries, having expected that these would return zero results. Its _'If a query consists of all stopwords, such as "to be or not to be", then all words are required.'_ behaviour is the opposite to what we want as we're using query-time stopwords to prevent particular query terms matching, but there's no way to disable the behaviour other than using `dismax` instead, which may have other impacts.

# Solution

This PR adds an `alwaysStopwords` option that disables the default behaviour. Its name is TBD.

I've noticed that the query plan becomes `+()` rather than `MatchNoDocsQuery("")` for pure-stopword queries, also when the query contains only tokenising characters (e.g. punctuation). Does this make any difference?

# Tests

_TO DO!_

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/ExtendedDismaxQParser.java
+++ b/solr/core/src/java/org/apache/solr/search/ExtendedDismaxQParser.java
@@ -98,6 +98,9 @@ public class ExtendedDismaxQParser extends QParser {
 
     /** If set to true, stopwords are removed from the query. */
     public static String STOPWORDS = "stopwords";
+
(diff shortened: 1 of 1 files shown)


andywebb1975:
superseded by https://github.com/apache/solr/pull/3802

PR #2489, opened 2024-05-31
Introduce support for Reciprocal Rank Fusion (combining queries)

alessandrobenedetti:
https://issues.apache.org/jira/browse/SOLR-17319

# Description

Reciprocal Rank Fusion (RRF) is an algorithm that takes in input multiple ranked lists to produce a unified result set.
Examples of use cases where RRF can be used include hybrid search and multiple Knn vector queries executed concurrently.
RRF is based on the concept of reciprocal rank, which is the inverse of the rank of a document in a ranked list of search results.
The combination of search results happens taking into account the position of
the items in the original rankings, and giving higher score to items that are ranked higher in multiple lists. RRF was introduced the first time by Cormack et al. in [1].
The syntax proposed:
JSON Request
```
{
    "queries": {
        "lexical1": {
            "lucene": {
                "query": "id:(10^=2 OR 2^=1 OR 4^=0.5)"
            }
        },
        "lexical2": {
            "lucene": {
                "query": "id:(2^=2 OR 4^=1 OR 3^=0.5)"
            }
        }
    },
    "limit": 10,
    "fields": "[id,score]",
    "params": {
        "combiner": true,
        "combiner.upTo": 5,
        "facet": true,
        "facet.field": "id",
        "facet.mincount": 1
    }
}
```

[1] Cormack, Gordon V. et al. “Reciprocal rank fusion outperforms condorcet and individual rank learning methods.” Proceedings of the 32nd international ACM SIGIR conference on Research and development in information retrieval (2009)
# Solution

The support has been introduced leveraging the "queries" support in the JSON request syntax.
The combination of results happen at QueryComponent, processing level.
The debug component has the responsibility of managing the explainability part.

Limitations: 

- grouping/field collapsing won't be supported in a first version
-  the distributed support is present but not super advanced: query combination happens at node level and then results are combined at shard level.

# Tests

Tests have been added to cover the main use cases

# Checklist


alessandrobenedetti:
As of today the solution is clean enough to be reviewed and tests are green.
The usage is intuitive and it should cover many use cases.
Pending activities:
- add a new documentation page
- add tests for the distributed scenario

solr/core/src/java/org/apache/solr/search/combining/ReciprocalRankFusion.java
```diff
@@ -0,0 +1,208 @@
...
+    combinedRankedList.setDocSet(docSet);
+
+    return docIdToRanks;
+  }
+
+  private Map<Integer, Integer[]> getRanks(List<DocList> docLists, int[] combinedResultsDocIDs) {
+    Map<Integer, Integer[]> docIdToRanks;
+    docIdToRanks = new HashMap<>();
+    for (int docID : combinedResultsDocIDs) {
+      docIdToRanks.put(docID, new Integer[docLists.size()]);
```
renatoh, line 135: Any particular reason for not using Integer.valueOf()?

alessandrobenedetti: To do what? Isn't Integer.valueOf(int a) used to wrap an integer?
Can you elaborate?

renatoh: oh dear, I completely missed the[] and thought you are create a new Integer by doing: new Integer() and therefore preventing them from being cached. sorry, my bad!

solr/core/src/test/org/apache/solr/handler/component/DebugComponentTest.java
```diff
@@ -52,7 +64,7 @@ public void testBasicInterface() {
         "//str[@name='querystring']='*:*'",
         "//str[@name='parsedquery']='MatchAllDocsQuery(*:*)'",
         "//str[@name='parsedquery_toString']='*:*'",
-        "count(//lst[@name='explain']/*)=3",
+        "count(//lst[@name='explain']/*)=6",
```
dsmiley, line 67: How does this PR affect this assertion of an existing test?

solr/core/src/java/org/apache/solr/request/json/RequestUtil.java
```diff
@@ -241,13 +243,16 @@ public static void processParams(
...
+            ArrayList<String> queryKeys = new ArrayList<>();
             for (Map.Entry<String, Object> queryJsonProperty : queriesAsMap.entrySet()) {
               out = queryJsonProperty.getKey();
+              queryKeys.add(out);
               arr = true;
               isQuery = true;
               convertJsonPropertyToLocalParams(
                   newMap, jsonQueryConverter, queryJsonProperty, out, isQuery, arr);
             }
+            newMap.put(COMBINER_KEYS, queryKeys.toArray(new String[queryKeys.size()]));
```
dsmiley, line 255:
```suggestion
            newMap.put(COMBINER_KEYS, queryKeys.toArray(new String[0]));
```
We standardized on this recently; surprisingly a 0 helps perf

dsmiley (commented):
I haven't looked at this in detail but I'm wondering could we instead have a single 'q' query parser, that is itself a "query combiner" of its children clauses?  Ultimately this would mean not needing all this special code in QueryComponent.  I'm trying to make recommendations that "fit in" well with Solr's architecture; less bolted / hacked on.

CC @hossman I think you'd have good inputs on this one too

alessandrobenedetti:
> I haven't looked at this in detail but I'm wondering could we instead have a single 'q' query parser, that is itself a "query combiner" of its children clauses? Ultimately this would mean not needing all this special code in QueryComponent. I'm trying to make recommendations that "fit in" well with Solr's architecture; less bolted / hacked on.
> 
> CC @hossman I think you'd have good inputs on this one too

In principle I agree, correct me if I'm wrong, your proposed solution means:
1) a Lucene contribution to create a new query
2) a Solr query parser to combine N queries into the new Lucene one?

I would need to think and design this possible solution, but I have no idea when this will happen as I am afraid it would take a significant amount of time. (of course anyone is happy to help).

Do you think it would make sense to start refining this contribution, bake it in so that users have this new possibility decently quickly and then iteratively replace it?

It's not because I have a strong opinion on this, I am just realistically think what could bring value from my side, in accordance with my future availabilty.
Happy to discuss!

dsmiley (approved):
I should have spent just another 10 minutes reading this to understand what's going on; I understand it better now.  I had overlooked the query loop.  I don't think a new Query would be that helpul for this use-case after-all.  Even if one were to exist, *somewhere* there would be code to detect it and loop it so _where_ would that go... either QueryComponent or SolrIndexSearcher which is another complexity minefield of greater magnitude.  It's tempting to think of a QueryComponent subclass for this use-case (and for that matter grouping).

BTW I like the query syntax, especially the JSON variant.

Sorry for any distress; +1 overall.

alessandrobenedetti:
> I should have spent just another 10 minutes reading this to understand what's going on; I understand it better now. I had overlooked the query loop. I don't think a new Query would be that helpul for this use-case after-all. Even if one were to exist, _somewhere_ there would be code to detect it and loop it so _where_ would that go... either QueryComponent or SolrIndexSearcher which is another complexity minefield of greater magnitude. It's tempting to think of a QueryComponent subclass for this use-case (and for that matter grouping).
> 
> BTW I like the query syntax, especially the JSON variant.
> 
> Sorry for any distress; +1 overall.

No distress David! I'm not the kind of developer that get attached to its own solutions.
If there are better ways I'm always open to suggestions.

I also agree in principle with proposing ideas first, before spending days inplementing them, the only reason that I don't do that often is because I am not full time on Solr contributions right now, actually I end up at the moment with limited time to do them, so I prefer to produce something that could be valuable and iterate.
Having to wait for other volunteers time to discuss the idea could end up in exhausting the time I had allocated for the contrib and have to wait for the next availability window between clients. 
Worst case scenario we end up with a discussion and we may ignore the Pull Request.

I remember at the beginning I tried query parsing and other approaches but the cleanest I got was the loop in the query component.

And I absolutely agree that component is messy, so after Berlin I'll spend a bit more time polishing the changes, and if in the meantime we come up with better ideas, the better!

No rush to merge, but at the same time if we get a decent consensus I would love to have this in as this feature will open many doors!

Thanks for the time you already spent on this, it's always very much appreciated!

solr/core/src/test/org/apache/solr/handler/component/DebugComponentTest.java
alessandrobenedetti: I added more docs in the setup of the test class, for the new debug tests I added.
I thought that changing the few lines of debug that were counting the total docs was not much of a big deal.
I can obviously change that if you think it is worth.

solr/solrj/src/java/org/apache/solr/common/params/CombinerParams.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.apache.solr.common.params;
+
+/** Defines the request parameters used by all combiners. */
+public interface CombinerParams {
+  String COMBINER = "combiner";
+
+  String COMBINER_ALGORITHM = COMBINER + ".algorithm";
+
+  String COMBINER_KEYS = COMBINER + ".keys";
```
dsmiley, line 25: WDYT of instead do `combine.q` multi-valued, so just supply each query as another value?  One less indirection and thing to name.

dsmiley (commented):
I could still imagine a query parser being used to express the inputs, and might map to users understanding of things.  Users know query parsers, know they can be rich and form a tree, know it matches documents.  To the user, maybe this is a special query even though from an implementation perspective, it wouldn't really be implemented as such because the clauses needed to each be evaluated independently.  The approach would get more logic out of QueryComponent (the feature parsing aspect).  It would add more net code, however, albeit not much.  It might better encapsulate the feature, even though the QueryComponent is an integral piece of what implements it.  Heck, we could think of this as giving the QueryComponent to a special Query if we want, thus enabling interesting use cases of queries that can be in charge of producing their document list and other stuff.

solr/core/src/java/org/apache/solr/search/combining/QueriesCombiner.java
```diff
@@ -0,0 +1,92 @@
...
+  protected int upTo;
+
+  public QueriesCombiner(SolrParams requestParams) {
+    this.upTo =
+        requestParams.getInt(CombinerParams.COMBINER_UP_TO, CombinerParams.COMBINER_UP_TO_DEFAULT);
+  }
+
+  public abstract QueryResult combine(QueryResult[] rankedLists);
+
+  QueryResult initCombinedResult(QueryResult[] rankedLists) {
```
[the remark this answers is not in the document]
alessandrobenedetti: done, coming in next commit!

solr/core/src/java/org/apache/solr/search/combining/QueriesCombiner.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Combining considers two or more query rankedLists: resultA, resultB ...<br>
+ * For a given query, each query result is a ranked list of documents La = (a1,a2,...), Lb = (b1,
+ * b2, ...)...<br>
+ * A combining algorithm creates a unique ranked list I = (i1, i2, ...).<br>
+ * This list is created by combining elements from the lists la and lb as described by the
+ * implementation algorithm.<br>
+ */
+public abstract class QueriesCombiner {
```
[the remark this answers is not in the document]
alessandrobenedetti: done, coming in next commit!

solr/core/src/java/org/apache/solr/handler/component/QueryComponent.java
```diff
@@ -173,22 +175,51 @@ public void prepare(ResponseBuilder rb) throws IOException {
...
+    rb.isCombined = params.getBool(CombinerParams.COMBINER, false);
+    if (rb.isCombinedSearch()) {
+      String[] queriesToCombineKeys = params.getParams(CombinerParams.COMBINER_KEYS);
+      unparsedQueries = new ArrayList<>(queriesToCombineKeys.length);
+      rb.queriesCombiningStrategy = QueriesCombiner.getImplementation(params);
+      for (String queryKey : queriesToCombineKeys) {
+        unparsedQueries.add(params.get(queryKey));
+      }
+    } else {
+      unparsedQueries = new ArrayList<>(1);
```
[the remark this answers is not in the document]
alessandrobenedetti: Done, coming in next commit

solr/solrj/src/java/org/apache/solr/common/params/CombinerParams.java
alessandrobenedetti: mmmmm, I do believe it can be useful to have a "query key", especially because the debug component explains both the query parsing and combination:

0.032522473 = 
1/(60+1) + 1/(60+2) because its ranks were: 1 for query(lexical), 2 for query(lexical2)

So the keys help in my opinion.
If you were referring to this specific parameter, this is not intended for the user, it gets used automatically by Solr.

Happy to elaborate!

dsmiley:
I did the work of migrating to a QParser but don't have permissions to push to your PR branch.  

Beware, I'm reverting most/all of the files that you modified of existing Solr source files.  If you would prefer I submit a separate PR, I'll do that.  Instead of `combiner=true`, you'd do `q={!combine}`.

Wouldn't the docSet (assuming it's needed, like for faceting) be computed on each subQuery, and if so how is that rolled up to the final QueryResult?  I see you tested that faceting works but I suspect there may be a very sad performance bug here.

alessandrobenedetti:
> I did the work of migrating to a QParser but don't have permissions to push to your PR branch.
> 
> Beware, I'm reverting most/all of the files that you modified of existing Solr source files. If you would prefer I submit a separate PR, I'll do that. Instead of `combiner=true`, you'd do `q={!combine}`.
I don't remember if this is the best way to invite you as a collaborator but I just added you on our repo, hopefully that works!

I wasn't entirely sure a Query Parser is the right place for this, given we are not really building a "new query" but then I thought : "hey we have the boolean query parser, that more or less does the same", so I have nothing against it!

You can go ahead and add your commit, I don't care much if you overwrote many of my changes, I just want a nice and clean solution, however we come up to it! Worst case scenario we go back in commits and re-use what's necessary!
> 
> Wouldn't the docSet (assuming it's needed, like for faceting) be computed on each subQuery, and if so how is that rolled up to the final QueryResult? I see you tested that faceting works but I suspect there may be a very sad performance bug here.



In regards to the docSet It was rolled up in:
`SortedIntDocSet docSet = new SortedIntDocSet(combinedResultsDocIds, combinedResultsLength);
    combinedRankedList.setDocSet(docSet);`
    
combinedResultsDocIds is the merged array of doc Ids.
I admit I haven't spent the required time on this part and it was quickly done as a draft, so feel free to recommend a different way/correct me!

solr/core/src/java/org/apache/solr/handler/component/QueryComponent.java
```diff
@@ -1689,10 +1689,11 @@ private void doProcessUngroupedSearch(ResponseBuilder rb, QueryCommand cmd, Quer
 
     SolrIndexSearcher searcher = req.getSearcher();
 
-    try {
+    if (cmd.getQuery() instanceof SelfExecutingQuery) {
+      // TODO QueryResult ought not to be created and passed in methods of QueryComponent
+      result = ((SelfExecutingQuery) cmd.getQuery()).search(searcher, cmd);
+    } else {
```
dsmiley, line 1695: Here's where the minimal essential bit to alter QueryComponent by adding a new abstraction (SelfExecutingQuery) that might be useful for other features; not just "combining".

solr/core/src/java/org/apache/solr/request/json/RequestUtil.java
```diff
@@ -241,13 +243,16 @@ public static void processParams(
...
+            ArrayList<String> queryKeys = new ArrayList<>();
             for (Map.Entry<String, Object> queryJsonProperty : queriesAsMap.entrySet()) {
               out = queryJsonProperty.getKey();
+              queryKeys.add(out);
               arr = true;
               isQuery = true;
               convertJsonPropertyToLocalParams(
                   newMap, jsonQueryConverter, queryJsonProperty, out, isQuery, arr);
             }
+            newMap.put(COMBINER_KEYS, queryKeys.toArray(new String[0])); // nocommit a hack
```
dsmiley, line 255: Making RequestUtils add a param specific to a niche feature is a hack.  If we really need this here (?), I think we should use a generic name like "json.queries".

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -706,9 +707,12 @@ public <K, V> boolean regenerateItem(
     }
   }
 
-  public QueryResult search(QueryResult qr, QueryCommand cmd) throws IOException {
-    getDocListC(qr, cmd);
-    return qr;
+  public void search(QueryResult qr, QueryCommand cmd) throws IOException {
+    try {
+      getDocListC(qr, cmd);
+    } catch (FuzzyTermsEnum.FuzzyTermsException e) { // unsure where best to catch this; shrug
```
dsmiley, line 713: I moved the FuzzyTermsException catch block from QueryComponent to here.  I wanted it in a deeper place that would be caught by multiple inward code paths instead of just QC.  It still feels it ought to be deeper but I didn't look for a better spot.  CombineQuery calls into this.

solr/core/src/java/org/apache/solr/search/combining/CombineQParserPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+import org.apache.solr.search.ExtendedQueryBase;
+import org.apache.solr.search.QParser;
+import org.apache.solr.search.QParserPlugin;
+import org.apache.solr.search.QueryCommand;
+import org.apache.solr.search.QueryParsing;
+import org.apache.solr.search.QueryResult;
+import org.apache.solr.search.SolrIndexSearcher;
+import org.apache.solr.search.SyntaxError;
+
+public class CombineQParserPlugin extends QParserPlugin {
```
dsmiley, line 43: This source file captures the essential aspect of my proposal.  Needs another iteration of refinement at least (e.g. javadocs, and param naming / defaulting).  Maybe more.

solr/core/src/java/org/apache/solr/search/combining/CombineQParserPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+        singleQueryDebug.add("queryparser", queryParsers.get(i).getClass().getSimpleName());
+        singleQueryDebug.add("parsedquery", QueryParsing.toString(queries.get(i), req.getSchema()));
+        singleQueryDebug.add("parsedquery_toString", queries.get(i).toString());
+        queriesDebug.add(queryKeys[i], singleQueryDebug);
+      }
+      dbg.add("queriesToCombine", queriesDebug);
+    }
+  }
+
+  static final class CombineQuery extends ExtendedQueryBase
```
dsmiley, line 115: A "SelfExecutingQuery" isn't normal; Lucene isn't going to work with this thing since it doesn't implement getWeight nor can it.  But users don't need to know/care.

solr/core/src/java/org/apache/solr/search/combining/CombineQParserPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+      QueryResult[] results = new QueryResult[queries.size()];
+      // TODO do in multiple threads?
+      for (int i = 0; i < queries.size(); i++) {
+        cmd.setQuery(queries.get(i));
+        QueryResult qr = new QueryResult();
+        searcher.search(qr, cmd);
+        results[i] = qr;
+      }
+
+      // nocommit but how is the docSet (e.g. for faceting) or maybe other things supported?
```
dsmiley, line 162: This seems like a fundamental limitation in the whole feature honestly.  What Hossman said about re-rank query is interesting... not sure if it addresses this.

solr/core/src/test/org/apache/solr/handler/component/DebugComponentTest.java
dsmiley: I suppose I'd need to see a before & after to truly "see" the change.

dsmiley (commented):
My commit here is itself a draft but significantly reduces the impact to existing Solr source files.  Combiner tests pass.

I didn't tend to the debug-ability aspect very much.  Just a draft :-)

solr/core/src/java/org/apache/solr/handler/component/QueryComponent.java
```diff
@@ -403,8 +402,9 @@ public void process(ResponseBuilder rb) throws IOException {
 
     req.getContext().put(SolrIndexSearcher.STATS_SOURCE, statsCache.get(req));
 
+    // TODO QueryResult ought not to be created and passed in methods of QueryComponent.
+    //  Should be created ~exclusively in SolrIndexSearcher and returned by it.
```
alessandrobenedetti, line 406: I tend to not be a huge fan of comments in the code, what does this mean? it's  just for your convenience in the draft?
If that's the case, ignore my comment

solr/core/src/java/org/apache/solr/request/json/RequestUtil.java
alessandrobenedetti: +1 for a generic name
json.queries looks all right to me

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
alessandrobenedetti: It seems all right to me, but I don't have a strong opinion

solr/core/src/java/org/apache/solr/search/combining/CombineQParserPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+
+    public CombineQParser(
+        String qstr, SolrParams localParams, SolrParams params, SolrQueryRequest req) {
+      super(qstr, localParams, params, req);
+    }
+
+    @Override
+    public Query parse() throws SyntaxError {
+
+      var queriesToCombineKeys = localParams.getParams("keys");
```
alessandrobenedetti, line 67: I am not a huge fan of var as in my opinion it reduces readability, but if it's ok as stadnard in Solr, no strong opinion

solr/core/src/java/org/apache/solr/search/combining/CombineQParserPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+        unparsedQueries.add(unparsedQuery);
+        queryParsers.add(parser);
+        queries.add(query);
+      }
+
+      return new CombineQuery(queriesCombiningStrategy, queries);
+    }
+
+    @Override
+    public void addDebugInfo(NamedList<Object> dbg) {
```
alessandrobenedetti, line 98: maybe 'addQueryDebugInfo' ?

solr/core/src/java/org/apache/solr/search/combining/CombineQParserPlugin.java
alessandrobenedetti: mmm probably I have lost in the comments why the way I created the docSet was incorrect. Can you elaborate? (a nice occasion to learn how that docSet part works, as I didn't have time yet)

solr/core/src/test/org/apache/solr/handler/component/DebugComponentTest.java
```diff
@@ -40,6 +40,18 @@ public static void beforeClass() throws Exception {
...
+    assertU(
+        adoc(
+            "id",
+            "5",
+            "title",
+            "Kakaroth aka Son Goku is a legendary saiyan warrior",
+            "inStock_b1",
+            "false"));
+    assertU(
+        adoc("id", "6", "title", "Gohan is the first saiyan-human hybrid.", "inStock_b1", "false"));
```
alessandrobenedetti, line 54: as you can see @dsmiley here I added more docs, that's the reason they are now 6

alessandrobenedetti (commented):
I like your changes David, they are less intrusive than my original version!
let me know once you do an additional refinement and I'll review them again

solr/solrj/src/java/org/apache/solr/common/params/CombinerParams.java
cpoerschke: > ... `combine.q` multi-valued ...

In the case of multi-valued parameters, is ordering 'guaranteed' or could `&combine.q=Foo&combine.q=Bar` turn into `&combine.q=Bar&combine.q=Foo` and if it did would it matter?

solr/solrj/src/java/org/apache/solr/common/params/CombinerParams.java
```diff
@@ -0,0 +1,36 @@
...
+
+  String RECIPROCAl_RANK_FUSION = "rrf";
```
cpoerschke, line 27:
```suggestion
  String RECIPROCAL_RANK_FUSION = "rrf";
```

solr/core/src/java/org/apache/solr/search/combining/CombineQParserPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+
+      unparsedQueries = new ArrayList<>(queriesToCombineKeys.length);
+      queryParsers = new ArrayList<>(queriesToCombineKeys.length);
+      queries = new ArrayList<>(queriesToCombineKeys.length);
+
+      // nocommit blend localParams and params without needless suffix in local
+      var blendParams = SolrParams.wrapDefaults(localParams, params);
+      queriesCombiningStrategy = QueriesCombiner.getImplementation(blendParams);
+
+      String defType = blendParams.get(QueryParsing.DEFTYPE, DEFAULT_QTYPE);
```
cpoerschke, line 80: Maybe could support different `defType` for different keys, though bit of a niche thing that perhaps.

solr/core/src/java/org/apache/solr/search/combining/QueriesCombiner.java
```diff
@@ -0,0 +1,93 @@
...
+
+    boolean segmentTerminatedEarly = false;
+    for (QueryResult result : rankedLists) {
+      if (result.getSegmentTerminatedEarly() != null) {
+        segmentTerminatedEarly |= result.getSegmentTerminatedEarly();
+      }
+    }
+    combinedRankedList.setSegmentTerminatedEarly(segmentTerminatedEarly);
+
+    combinedRankedList.setNextCursorMark(rankedLists[0].getNextCursorMark());
```
cpoerschke, line 71: Not sure about picking the first list here and/or what does use of a cursor mark mean when combining queries. Alternative might be to just disallow cursor marks when combining queries.

solr/solrj/src/java/org/apache/solr/common/params/CombinerParams.java
dsmiley: AFAIK it's guaranteed; we use a List not a Set intentionally

dsmiley:
@cpoerschke see the parent JIRA issue for the real state of this controvertial PR

cpoerschke:
> @cpoerschke see the parent JIRA issue for the real state of this controvertial PR

Thanks for the pointer! It inspired me to do some #2597 scribbling.

PR #2496, opened 2024-06-04
SOLR-17317: Correct 'wt' interpretation for v2 APIs

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-17317

# Description

A mailing list user recently pointed out that certain JAX-RS v2 APIs return XML content when "wt=json" is specified.  The cause is an embarrassing little coding error: a missing switch-case clause for "wt=json" in `V2ApiUtils` 

# Solution

This PR trivially corrects the error.

# Tests

Tests added for various 'wt' cases in `V2ApiUtilsTest`.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/api/V2ApiUtils.java
+++ b/solr/core/src/java/org/apache/solr/handler/api/V2ApiUtils.java
@@ -26,9 +26,10 @@ import java.util.List;
 import java.util.Map;
 import org.apache.solr.client.api.model.SolrJerseyResponse;
 import org.apache.solr.common.MapWriter.EntryWriter;
+import org.apache.solr.common.params.SolrParams;
 import org.apache.solr.common.util.NamedList;
+import org.apache.solr.common.util.StrUtils;
 import org.apache.solr.common.util.Utils;
-import org.apache.solr.request.SolrQueryRequest;
 import org.apache.solr.response.SolrQueryResponse;
 
 /** Utilities helpful for common V2 API declaration tasks. */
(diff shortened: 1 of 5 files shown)


HoustonPutman (approved):

epugh (approved):
Lgtm

solr/core/src/java/org/apache/solr/handler/api/V2ApiUtils.java
```diff
@@ -87,13 +88,14 @@ public static void squashIntoNamedListWithoutHeader(
...
-      SolrQueryRequest solrQueryRequest, String defaultMediaType) {
-    final String wtParam = solrQueryRequest.getParams().get(WT);
-    if (wtParam == null) return "application/json";
+  public static String getMediaTypeFromWtParam(SolrParams params, String defaultMediaType) {
+    final String wtParam = params.get(WT);
+    if (StrUtils.isBlank(wtParam)) return defaultMediaType;
 
     // The only currently-supported response-formats for JAX-RS v2 endpoints.
     switch (wtParam) {
+      case "json":
```
madrob, line 97: This feels like something that should be pluggable and stored in a Map instead of a switch. Thinking about what if somebody wants cbor or whatever.

solr/core/src/test/org/apache/solr/handler/api/V2ApiUtilsTest.java
```diff
@@ -36,4 +40,21 @@ public void testReadsDisableV2ApiSysprop() {
...
+    var params = new ModifiableSolrParams();
+    params.add("wt", "json");
+    assertEquals(MediaType.APPLICATION_JSON, V2ApiUtils.getMediaTypeFromWtParam(params, null));
+
+    params.set("wt", "xml");
+    assertEquals(MediaType.APPLICATION_XML, V2ApiUtils.getMediaTypeFromWtParam(params, null));
+
+    params.set("wt", "javabin");
+    assertEquals(BINARY_CONTENT_TYPE_V2, V2ApiUtils.getMediaTypeFromWtParam(params, null));
+  }
```
madrob, line 59: params.set(wt, foo)

solr/core/src/java/org/apache/solr/handler/api/V2ApiUtils.java
gerlowskija: Yeah, I had this thought as I was adding the case.  A map would be a bit cleaner, especially as we add new content-types.

I'll leave it as-is for now, and switch it over the next time we add a content-type.

solr/core/src/test/org/apache/solr/handler/api/V2ApiUtilsTest.java
gerlowskija: Can you clarify? Not sure what you're trying to convey here...

madrob: You're missing test coverage for an unknown type

PR #2499, opened 2024-06-05
SOLR-17195: Add 'minPrefixLength' soft limit

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-17195

# Description

Prefix-based queries consume memory in proportion to the number of terms in the index that start with the prefix.  Short prefixes tend to match many more indexed terms, and consume more memory as a result, often causing instability issues on the node.

Yet Solr (prior to this PR) offers no way to restrict the prefixes used in queries.

# Solution

This PR adds a solrconfig.xml property, `minPrefixLength`, which operates similarly to the existing `maxBooleanClauses`.  Users who submit a query with a prefix shorter than the minimal acceptable length will receive an error of the form:

```
Query <snip> does not meet the minimum prefix length [2] (actual=[1]).  Please try with a larger prefix, or adjust minPrefixLength in your solrconfig.xml
```
Some notes on the implementation:

- currently only enforced for 'string' fields, where this cardinality problem occurs most frequently
- defaults to '2' in the default configset, prohibiting only single-character prefixes.
- can be overridden in the default configset with the `solr.min.prefixLength` sysprop/env-var
  
# Tests

Tests for solrconfig.xml in `SolrCoreTest`. Additional tests on actual functionality still needed.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/SolrConfig.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrConfig.java
@@ -103,6 +103,8 @@ public class SolrConfig implements MapSerializable {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   public static final String DEFAULT_CONF_FILE = "solrconfig.xml";
+  public static final String MIN_PREFIX_LENGTH = "minPrefixLength";
+  public static final int DEFAULT_MIN_PREFIX_LENGTH = 0;
   private final String resourceName;
 
   private int znodeVersion;
@@ -291,6 +293,8 @@ public class SolrConfig implements MapSerializable {
             BooleanQuery.getMaxClauseCount(),
             "set 'maxBooleanClauses' in solr.xml to increase global limit");
       }
+      prefixQueryMinPrefixLength =
+          get("query").get(MIN_PREFIX_LENGTH).intVal(DEFAULT_MIN_PREFIX_LENGTH);
 
       // Warn about deprecated / discontinued parameters
       // boolToFilterOptimizer has had no effect since 3.1
@@ -667,6 +671,7 @@ public class SolrConfig implements MapSerializable {
 
   /* The set of materialized parameters: */
   public final int booleanQueryMaxClauseCount;
+  public final int prefixQueryMinPrefixLength;
   // SolrIndexSearcher - nutch optimizer -- Disabled since 3.1
   //  public final boolean filtOptEnabled;
   //  public final int filtOptCacheSize;
@@ -1018,6 +1023,7 @@ public class SolrConfig implements MapSerializable {
     m.put("queryResultMaxDocsCached", queryResultMaxDocsCached);
     m.put("enableLazyFieldLoading", enableLazyFieldLoading);
     m.put("maxBooleanClauses", booleanQueryMaxClauseCount);
+    m.put(MIN_PREFIX_LENGTH, prefixQueryMinPrefixLength);
 
     for (SolrPluginInfo plugin : plugins) {
       List<PluginInfo> infos = getPluginInfos(plugin.clazz.getName());
--- a/solr/core/src/java/org/apache/solr/schema/StrField.java
+++ b/solr/core/src/java/org/apache/solr/schema/StrField.java
@@ -20,17 +20,21 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import org.apache.lucene.document.SortedDocValuesField;
 import org.apache.lucene.document.SortedSetDocValuesField;
 import org.apache.lucene.index.IndexableField;
 import org.apache.lucene.queries.function.ValueSource;
 import org.apache.lucene.queries.function.valuesource.SortedSetFieldSource;
+import org.apache.lucene.search.PrefixQuery;
+import org.apache.lucene.search.Query;
 import org.apache.lucene.search.SortField;
 import org.apache.lucene.search.SortedSetSelector;
 import org.apache.lucene.util.BytesRef;
 import org.apache.solr.common.SolrException;
 import org.apache.solr.common.util.ByteArrayUtf8CharSequence;
+import org.apache.solr.core.SolrConfig;
 import org.apache.solr.response.TextResponseWriter;
 import org.apache.solr.search.QParser;
 import org.apache.solr.uninverting.UninvertingReader.Type;
@@ -66,6 +70,34 @@ public class StrField extends PrimitiveFieldType {
       fval = docval;
     }
     return Collections.singletonList(fval);
+  }
+
+  @Override
+  public Query getPrefixQuery(QParser parser, SchemaField sf, String termStr) {
+    final var query = super.getPrefixQuery(parser, sf, termStr);
+
+    // Some internal usage (e.g. faceting) creates PrefixQueries without a surrounding QParser, so
+    // check for null here before using QParser to access the limit value
+    if (query instanceof PrefixQuery && parser != null) {
+      final var minPrefixLength =
+          parser.getReq().getCore().getSolrConfig().prefixQueryMinPrefixLength;
+      // TODO Should we provide a query-param to disable the limit on a request-by-request basis?  I
+      // can imagine scenarios where advanced users may want to enforce the limit on most fields,
+      // but ignore it for a few fields that they know to be low-cardinality and therefore "less
+      // risky"
+      if (termStr.length() < minPrefixLength) {
+        final var message =
+            String.format(
+                Locale.ROOT,
+                "Query [%s] does not meet the minimum prefix length [%d] (actual=[%d]).  Please try with a larger prefix, or adjust %s in your solrconfig.xml",
+                query,
+                minPrefixLength,
+                termStr.length(),
+                SolrConfig.MIN_PREFIX_LENGTH);
+        throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, message);
+      }
+    }
+    return query;
   }
 
   public static BytesRef getBytesRef(Object value) {
--- a/solr/core/src/test-files/solr/collection1/conf/solrconfig.xml
+++ b/solr/core/src/test-files/solr/collection1/conf/solrconfig.xml
@@ -89,6 +89,18 @@
     -->
     <maxBooleanClauses>${solr.max.booleanClauses:1024}</maxBooleanClauses>
 
+    <!-- Minimum acceptable prefix-size for prefix-based queries on string fields.
+
+         Prefix-based queries consume memory in proportion to the number of terms in the index
+         that start with that prefix.  Short prefixes tend to match many many more indexed-terms
+         and consume more memory as a result, sometimes causing stability issues on the node.
+
+         This setting allows administrators to require that prefixes meet or exceed a specified
+         minimum length requirement.  Prefix queries that don't meet this requirement return an
+         error to users.
+    -->
+    <minPrefixLength>${solr.min.prefixLength:2}</minPrefixLength>
+
     <!-- Cache specification for Filters or DocSets - unordered set of *all* documents
          that match a particular query.
       -->
--- a/solr/core/src/java/org/apache/solr/search/facet/FacetFieldProcessor.java
+++ b/solr/core/src/java/org/apache/solr/search/facet/FacetFieldProcessor.java
@@ -544,6 +544,10 @@ abstract class FacetFieldProcessor extends FacetProcessor<FacetField> {
   private void calculateNumBuckets(SimpleOrderedMap<Object> target) throws IOException {
     DocSet domain = fcontext.base;
     if (freq.prefix != null) {
+      // TODO - Should we enforce minPrefixLength here in the case of 'string' fields, or omit
+      //  since this is an "internal" request? If we want to enforce the limit in this case,
+      //  we should have StrField read the configured limit and cache it in 'init' so that it can
+      //  be read at 'getPrefixQuery' call-time without a QParser.
       Query prefixFilter = sf.getType().getPrefixQuery(null, sf, freq.prefix);
       domain = fcontext.searcher.getDocSet(prefixFilter, domain);
     }
--- a/solr/core/src/test/org/apache/solr/ConvertedLegacyTest.java
+++ b/solr/core/src/test/org/apache/solr/ConvertedLegacyTest.java
@@ -637,8 +637,8 @@ public class ConvertedLegacyTest extends SolrTestCaseJ4 {
         "<add><doc><field name=\"id\">107</field><field name=\"val_s\">port</field></doc></add>");
     assertU("<commit/>");
 
-    assertQ(req("val_s:a*"), "//*[@numFound='3']");
-    assertQ(req("val_s:p*"), "//*[@numFound='4']");
+    assertQ(req("val_s:ap*"), "//*[@numFound='3']");
+    assertQ(req("val_s:pe*"), "//*[@numFound='3']");
     // val_s:* %//*[@numFound="8"]
 
     // test wildcard query
--- a/solr/core/src/test/org/apache/solr/core/SolrCoreTest.java
+++ b/solr/core/src/test/org/apache/solr/core/SolrCoreTest.java
@@ -298,6 +298,7 @@ public class SolrCoreTest extends SolrTestCaseJ4 {
     assertEquals(
         "wrong config for slowQueryThresholdMillis", 2000, solrConfig.slowQueryThresholdMillis);
     assertEquals("wrong config for maxBooleanClauses", 1024, solrConfig.booleanQueryMaxClauseCount);
+    assertEquals("wrong config for minPrefixLength", 2, solrConfig.prefixQueryMinPrefixLength);
     assertTrue("wrong config for enableLazyFieldLoading", solrConfig.enableLazyFieldLoading);
     assertEquals("wrong config for queryResultWindowSize", 10, solrConfig.queryResultWindowSize);
   }
--- a/solr/core/src/test/org/apache/solr/search/TestFiltering.java
+++ b/solr/core/src/test/org/apache/solr/search/TestFiltering.java
@@ -64,7 +64,7 @@ public class TestFiltering extends SolrTestCaseJ4 {
 
     String[] queries = {
       "foo_s:foo",
-      "foo_s:f*",
+      "foo_s:fo*",
       "*:*",
       "id:[* TO *]",
       "id:[0 TO 99]",
--- a/solr/core/src/test/org/apache/solr/search/join/BJQParserTest.java
+++ b/solr/core/src/test/org/apache/solr/search/join/BJQParserTest.java
@@ -55,6 +55,7 @@ public class BJQParserTest extends SolrTestCaseJ4 {
 
   @BeforeClass
   public static void beforeClass() throws Exception {
+    System.setProperty("solr.min.prefixLength", String.valueOf(1));
     initCore("solrconfig.xml", "schema15.xml");
     createIndex();
   }
--- a/solr/server/solr/configsets/_default/conf/solrconfig.xml
+++ b/solr/server/solr/configsets/_default/conf/solrconfig.xml
@@ -360,6 +360,19 @@
       -->
     <maxBooleanClauses>${solr.max.booleanClauses:1024}</maxBooleanClauses>
 
+
+    <!-- Minimum acceptable prefix-size for prefix-based queries on string fields.
+
+         Prefix-based queries consume memory in proportion to the number of terms in the index
+         that start with that prefix.  Short prefixes tend to match many many more indexed-terms
+         and consume more memory as a result, sometimes causing stability issues on the node.
+
+         This setting allows administrators to require that prefixes meet or exceed a specified
+         minimum length requirement.  Prefix queries that don't meet this requirement return an
+         error to users.
+    -->
+    <minPrefixLength>${solr.min.prefixLength:2}</minPrefixLength>
+
     <!-- Solr Internal Query Caches
          Starting with Solr 9.0 the default cache implementation used is CaffeineCache.
     -->


gerlowskija:
Three questions I'd like to answer before this is merged:

1. I can imagine scenarios where an advanced user might want to enable this feature, but still be able to exempt certain queries.  (Perhaps they know that certain fields are low-cardinality enough to be "safe", etc.)  Is it worth adding a query-param feature flag, that can be used to disable the limit on a query-by-query basis?
2. Some faceting codepaths run prefix queries in the course of processing certain types of facets.  Should we enforce this limit on these codepaths or skip it?
3. Is there any value in expanding the limit to cover field types other than 'string'?

Other TODOs:
  - ref-guide docs
  - additional testing
  - tidy/check.

solr/core/src/java/org/apache/solr/core/SolrConfig.java
```diff
@@ -103,6 +103,8 @@ public class SolrConfig implements MapSerializable {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   public static final String DEFAULT_CONF_FILE = "solrconfig.xml";
+  public static final String MIN_PREFIX_LENGTH = "minPrefixLength";
```
dsmiley, line 106: from a naming standpoint, this is very ambiguous.  Prefix of what?  And should this be here at all vs. a request parameter, interpreted by some (not all) query parsers?

dsmiley (commented):
I think this should be a query-parser based limit, not a field type enforced one.

dsmiley:
I wish we brainstormed this a bit on the dev list or something prior to you showing up with a PR where there is maybe inertia to continue with it the way it is

gerlowskija:
I don't really think that's a fair characterization.  The JIRA ticket was open and advertised to all of us on `issues@` 3 months ago, and no one chimed in there.  Sometimes the brainstorming and feedback just doesn't appear until there's a 'patch' to stimulate it, so I wrote one 🤷 

In any case, I'm happy to discuss/brainstorm the overall approach now.  I think it stands on its own without considering "inertia".

---

In short, my thought process was that the more broadly applicable the limit is, the more effective it will be at protecting stability.  It's hard to achieve stability if the safeguards you're relying on are opt-in, "off" by default, or only apply to a subset of relevant queries.  So I wanted to make the "minPrefix" limit be opt-out, apply broadly, etc.

Solr already has a query-time limit that fits this "broadly applicable" model pretty well, also in service of stability: `maxBooleanClauses`.  mBC has caused its share of issues, but the underlying model is sound and has been a part of Solr for at least a decade.  Users are already familiar with this sort of limit coming in under solrconfig.xml, etc.

So that's mostly what I've tried to do in this PR: (1) make it "broad" by design, and (2) follow mBC as a model in terms of how to configure, document, and implement.

@dsmiley - You noted above your preference for something parser-specific.  Can you elaborate on that preference a bit?

dsmiley:
I'm sorry for challenging your due diligence on soliciting input.

> In short, my thought process was that the more broadly applicable the limit is, the more effective it will be at protecting stability. It's hard to achieve stability if the safeguards you're relying on are opt-in, "off" by default, or only apply to a subset of relevant queries. So I wanted to make the "minPrefix" limit be opt-out, apply broadly, etc.

This runs the real risk of an obscure limit showing itself in production that would hot have been caught in a test setup for a use-case that should absolutely not have such a limit.  Generally "q" is provided by a user, uses a more user-facing query parser (namely edismax), and would be a nice place for such a limit.  But otherwise, queries show up in tons of other places for internal use-cases written by systems programmers (i.e. us devs) using the default "lucene" query parser where I don't think such a limit at all makes sense (at least not by default).  Random examples -- delete-by-query, facet.query, join query.

The same is so for maxBooleanClauses!  I was hoping you wouldn't bring that up but of course you did; it's the precedent for your decision.  That unfortunate precedent caused a huge ugly debate with only one person holding veto power that insisted on maxBooleanClauses as it exists today.  That might have made Yonik leave to do Heliosearch over, if there was a huge straw break that back.  I hope we don't repeat that mistake.  I've only ever increased maxBooleanClauses to something insane, not because there isn't a risk of users doing expensive queries that we want to limit, but instead because the correct (IMO) place to enforce query constraints is at the query parser.  And we do this where I work for various things that in-effect constraint the query complexity.

@markrmiller you and most others were involved in that ugly debate... what do you think of Jason's approach here?

gerlowskija:
> Generally "q" is provided by a user, uses a more user-facing query parser (namely edismax), and would be a nice place for such a limit. But otherwise, queries show up in tons of other places for internal use-cases written by systems programmers (i.e. us devs) using the default "lucene" query parser

I'm not sure the edismax=user-queries, lucene=system-queries generalization is a safe one to make.

And even if it is, I think you're making an additional assumption here that "system" queries wouldn't also benefit from this check.  I'm not sure I agree.  System-devs are more likely than the average user to know about Solr's performance pitfalls, but is that safe to assume across the board?  I think our many years with the project put us in danger of overestimating the knowledge and context other users might have, and underestimating the value of safeguards.

That's the key issue for me IMO - if I was more comfortable with those generalizations I'd be happy with an edismax fix.  But as-is they feel like pretty shaky ground to base cluster-stability on.


> I've only ever increased maxBooleanClauses to something insane, not because there isn't a risk of users doing expensive queries that we want to limit

`maxBooleanClauses` is, admittedly, a pain.  For one, it's terribly calibrated, or at least terribly out of date.  (Is 1024 clauses even a lot on today's hardware?  It seems like nothing!). I'm not here to defend it.  And I can't speak to the old debate without reading up on it.  (Do you have a pointer handy @dsmiley ?)

But conceptually I think `maxBooleanClauses` takes the right response to a "DANGER" stimulus.  If something has the potential to take down a cluster a global block is the safest thing to do.  Solr should be stable by default, even if that means admins have a little extra work to do to "unlock" certain query patterns.

gerlowskija:
> Generally "q" is provided by a user, uses a more user-facing query parser (namely edismax), and would be a nice place for such a limit. But otherwise, queries show up in tons of other places for internal use-cases written by systems programmers (i.e. us devs) using the default "lucene" query parser

Thinking about this a bit more, I wonder if the real issue here is that there's no way to tell Solr unequivocally, on a request-by-request basis: "Don't second-guess me, I know this is safe".  If users had a way to communicate that, there wouldn't be any need for us as developers to try to generalize about which queries came from "end" users, what users may or may not know about Solr performance, etc.

Would adding something like that  make you feel any better about the parser-agnostic approach @dsmiley ?  It'd give savvy "system" users a way to sidestep the "hitting a limit for the first time in prod" false-positive you expressed concern about above.

markrmiller:
“what do you think of Jason's approach here?”

It is a kind thorny practice that evades a good definitive judgment.

I suppose the idea is, at some, perhaps fairly arbitrary limit, you fail things to force the user to reckon with what they are doing more carefully and actively raise limits of they think that's the correct action. I say perhaps fairly arbitrary because it's difficult to do any better, given varying hardware and data.

The cost is that is that things that worked fine in testing may start blowing up randomly in production, perhaps needlessly, until you address the situation.
Thorny. Benifit / reward, what lines are drawn where, it's all pretty subjective. So I don't know.

As a user, I certainly would want to be able to opt in, though yes, that makes the effort likely of little value, particularly if it's opt-in per limit. A universal opt-in would be better but would likely still end up as little value then. As a user, I'll still be pissed when production starts failing things the first time, and I judge it, should not be doing that for my case. And at that point, I would be very unhappy if there was not a universal opt-out. Thorny.

I do see the value, but in most cases, that value would not likely be high enough to convince me as a user that this was a good default system practice. Other than for very particular things, it doesn't seem like a common practice because of how thorny it is.

I think some limits have a bit more clear value, though. Things that can build up over time and then are very difficult to address. Something along the lines of no limit to the number of fields on a document, where things can work fine for the first 6 months, but over time, you are using more and more fields and eventually the system just kind of seizes under the weight, and you can't do much but start over. Maybe hitting something much earlier that said you are using a strangely large number of fields, this is not a very scable approach, you have to change a configuration to continue down this path, would make a lot more sense.

markrmiller:
One use case where opt-in still has a lot of value is when one team runs Solr as a service or for other teams. In that case, it's much more of an available feature than automatic user self protection.

gerlowskija:
@markrmiller mentioned to me offline that it was clear that my preference was for cluster-stability over convenience/admin-friendliness, but that I never made the rationale behind that preference explicit.  He suggested I take one last crack at making the case for across-the-board limiting, as explicitly as possible.  So here it is:

Ultimately my rationale is pretty simple: aggressive prefix queries have the potential to block _all_ traffic if they trigger an OOM.  That's a worst-case-scenario for an administrator. 

There's pain for administrators either way.  Realizing that some setting has errantly blocked some "known safe" queries definitely brings a serious "annoyance-factor".  But to put myself in an administrator's shoes, I'd rather deal with that on 10, 20, even 100 clusters than deal with an "all traffic" outage at 3am because of a runaway prefix query.

----

So that's my last pitch for across-the-board limiting.  I'll give this a few days to see if it sways anyone; if no one else is convinced enough to chime in or reverse course though I'll go forward with something like the more limited approach that David championed.

dsmiley:
I've been on vacation lately.
I suggested a limit based on the query parser.  This suggestion isn't limited to edismax; it'd include some others that users are likely to use.  Another angle is to employ this limit only on a TextField, which generally _should_ be used for keyword search and thus user queries.  I think I can be comfortable with that.  But not StrField in particular (in terms of opt-in vs opt-out).  That's where my concern is most likely to be used in practice.

gerlowskija:
> I suggested a limit based on the query parser. This suggestion isn't limited to edismax; it'd include some others that users are likely to use.

Did you have particular parsers in mind @dsmiley ?  Or is there somewhere else that the project relied on this "end-user" vs "admin" parser distinction, that I could get a list from?

I'm on board with expanding the set of parsers that will enforce this limit, but I don't have a great sense what you or others would consider "end-user" parsers.  If you have a concrete subset in mind, then I'm happy to run with that.

> Another angle is to employ this limit only on a TextField, which generally should be used for keyword search and thus user queries.

I think I agree that "field-type" is a better predictor of "user query or not" than the query-parser would be.

Maybe my experience is idiosyncratic, but prefix queries on TextField aren't the common case from what I've seen.  Anecdotally, prefix queries are much much more common on StrField.

We could apply the limit only on TextField, but if that only covers (e.g.) 20% of the prefix queries that'll hit the index - is it providing stability?  I guess it still could, if the terms dictionary was large enough.  But still, leaving the majority prefix-query use-case uncovered doesn't feel great...  I'm just thinking-aloud/rambling at this point I guess.

----

Anyway, let me think through which of these two approaches (field-based vs parser-based) I like better, and I'll push up a modification shortly.

Thanks for sticking with this through your time off any everything @dsmiley ; hope you had a good trip!

gerlowskija:
The more I think through this today, the more I start to wonder whether I was wrong to oppose an "opt-in" solution originally.

It doesn't benefit all users the way an opt-out solution would, but maybe that's for the best given the high level of concern around annoyance and false-positives. Shrewd users, and folks who especially need it (e.g. the "service" use case Mark mentioned) would get a really comprehensive limit at their disposal once they enable it.  And of course, it'd keep us out of the business of trying to guess which queries are likely to be "user queries".

Is that something you could get behind @dsmiley ?

epugh:
One idea is that maybe we would want to have a page in the ref guide that was "The Defensive Solr Setup" and you could reference this type of feature and other features that enable this?   Maybe someday we ship a configset that represents a configuration that is very protective?   To make it easier for non experts to adopt this type of feature?

dsmiley:
"user" query parsers: complexphrase, dismax, edismax.  All others seem internal to me.  We could have an abstract method on QParserPlugin (or QParser) that compels plugin authors to indicate what the _default_ prefix query limit should be (if any).

Field types:  (I accept this approach too, albeit I like it less)

> Is that something you could get behind @dsmiley ?

I'm unsure exactly what you are asking me.  I'm definitely behind "opt-in" :-)

gerlowskija:
> I'm unsure exactly what you are asking me. I'm definitely behind "opt-in" :-)

Sorry - I should've been more concrete.  I'm asking if you could live with a parser-agnostic (i.e. global) limit like what the PR currently implements, if it was made "opt-in" rather than "opt-out".

AFAICT that combination would address your concerns around false-positives, making work for admins, etc.  And it'd be my preference over a parser-aware approach, as I'm not totally comfortable with the idea of inferring query origin from the parser implementation.

If you're -1 on any sort of a "global" limit, even as an opt-in feature, I'll implement parser-aware.  But I didn't want to do that without checking where you stood on the global+optIn approach.

solr/core/src/java/org/apache/solr/schema/StrField.java
```diff
@@ -68,6 +71,20 @@ public List<IndexableField> createFields(SchemaField field, Object value) {
...
+  @Override
+  public Query getPrefixQuery(QParser parser, SchemaField sf, String termStr) {
+    final var query = super.getPrefixQuery(parser, sf, termStr);
+
+    // Some internal usage (e.g. faceting) creates PrefixQueries without a surrounding QParser, so
+    // check for null here before using QParser to access the limit value
+    if (query instanceof PrefixQuery && parser != null) {
+      final var minPrefixLength =
+          parser.getReq().getCore().getSolrConfig().prefixQueryMinPrefixLength;
+      QueryUtils.ensurePrefixQueryObeysMinimumPrefixLength(query, termStr, minPrefixLength);
```
dsmiley, line 83: At this juncture, I think you should call parser.getPrefixQueryMinPrefixLength so that a QParser has the opportunity to choose a default.  The default implementation of this method can be as you indicate.

solr/core/src/java/org/apache/solr/schema/TextField.java
```diff
@@ -165,6 +167,20 @@ public Query getFieldTermQuery(QParser parser, SchemaField field, String externa
     return new TermQuery(new Term(field.getName(), br));
   }
 
+  @Override
+  public Query getPrefixQuery(QParser parser, SchemaField sf, String termStr) {
+    final var query = super.getPrefixQuery(parser, sf, termStr);
```
dsmiley, line 172: Why override these methods in TextField & StrField with duplicative code when you could modify `FieldType.getPrefixQuery` instead?  Doing it there also allows the cool existence query thing.

solr/core/src/java/org/apache/solr/search/ComplexPhraseQParserPlugin.java
```diff
@@ -134,6 +135,19 @@ protected Query newWildcardQuery(org.apache.lucene.index.Term t) {
               }
             }
 
+            @Override
+            protected Query getPrefixQuery(String field, String termStr) throws ParseException {
+              // TODO check the field type and call QueryUtils.ensureBlah here
```
dsmiley, line 140: It's not clear why this parser needs modifications like this

solr/core/src/java/org/apache/solr/search/QueryUtils.java
```diff
@@ -83,6 +85,25 @@ public static boolean isConstantScoreQuery(Query q) {
     }
   }
 
+  public static void ensurePrefixQueryObeysMinimumPrefixLength(
+      Query query, String prefix, int minPrefixLength) {
+    // TODO Should we provide a query-param to disable the limit on a request-by-request basis?  I
+    // can imagine scenarios where advanced users may want to enforce the limit on most fields,
+    // but ignore it for a few fields that they know to be low-cardinality and therefore "less
```
dsmiley, line 92: needs javadocs

solr/core/src/java/org/apache/solr/search/QueryUtils.java
```diff
@@ -83,6 +85,25 @@ public static boolean isConstantScoreQuery(Query q) {
     }
   }
 
+  public static void ensurePrefixQueryObeysMinimumPrefixLength(
+      Query query, String prefix, int minPrefixLength) {
+    // TODO Should we provide a query-param to disable the limit on a request-by-request basis?  I
```
dsmiley, line 90: yes indeed; in fact the QParser default algorithm could lookup a local-param (not request level) like prefixQueryMinimumLength.  This will be straightforward once you create the QParser default method as I suggested.

solr/core/src/test-files/solr/collection1/conf/solrconfig.xml
```diff
@@ -89,6 +89,18 @@
...
+
+         Prefix-based queries consume memory in proportion to the number of terms in the index
+         that start with that prefix.  Short prefixes tend to match many many more indexed-terms
+         and consume more memory as a result, sometimes causing stability issues on the node.
+
+         This setting allows administrators to require that prefixes meet or exceed a specified
+         minimum length requirement.  Prefix queries that don't meet this requirement return an
+         error to users.
+    -->
+    <minPrefixLength>${solr.min.prefixLength:2}</minPrefixLength>
```
dsmiley, line 102: again; name is confusing and I disagree with the choice of `solr.min.prefixLength`.  The second component should be a category like "query", so could be "solr.query.minPrefixLength" for example.

solr/solr-ref-guide/modules/configuration-guide/pages/caches-warming.adoc
```diff
@@ -235,6 +235,18 @@ This is the same system property used in the xref:configuring-solr-xml.adoc#glob
 <maxBooleanClauses>${solr.max.booleanClauses:1024}</maxBooleanClauses>
 ----
 
+=== <minPrefixLength> Element
```
dsmiley, line 238: hmm; weird that a documentation file named "caches-warming" would document this query specific limit

dsmiley (commented):
I think we can have it both ways.  Opt-in (thus no limit by default).  And a relatively minor change in this PR to consult the QParser which *really* is the right place to have a hook for this.  The FieldType can continue to be involved, but I don't see that this is a FieldType specific option.

solr/solr-ref-guide/modules/configuration-guide/pages/caches-warming.adoc
gerlowskija: Agreed.  Looking at the file it, it covers everything in the `<query>` section of solrconfig.xml.  A bunch of that _is_ cache related, but not all of it obviously.

Maybe the scope of the file changed over time and no one ever noticed as the name became more and more out of date?  Or perhaps someone noticed, but the name was intentionally retained for SEO or other reasons since it appears in the ref-guide URL?

gerlowskija:
Alright @dsmiley - I took in most of your suggestions:

1. The logic for determining the limit now lives in QParser
2. QParser looks at the solrconfig.xml setting, but uses a local-param override if specified
3. The setting is now "opt-in", with the flag value "-1" used in the default configset.

Lmk what you think; hopefully this is close.

solr/core/src/java/org/apache/solr/core/SolrConfig.java
dsmiley: Does the "Term" in that add anything?  I suggest removing that component.  Not a strong opinion though.

solr/core/src/java/org/apache/solr/schema/TextField.java
dsmiley: I see that the limit doesn't make much sense for those two yet it's also harmless (I think).  I prefer a consistent approach with no duplication.  Thanks for doing this.

dsmiley (commented):
Definitely in great shape.

solr/solr-ref-guide/modules/configuration-guide/pages/caches-warming.adoc
dsmiley: I don't think we should keep this named as such for legacy SEO purposes.  Any way; it'd be a follow-up PR, not here.

solr/core/src/java/org/apache/solr/search/ComplexPhraseQParserPlugin.java
dsmiley: Hopefully a simple fix to this parser to consult the field type will work.  That's the right thing to do.  Perhaps whoever wrote this didn't know better or it didn't exist at the time.

solr/core/src/java/org/apache/solr/schema/TextField.java
gerlowskija: Well there is _some_ harm, albeit small.  A prefix query on a "known safe" field type like Enum might trigger the limit, and cause the "Admin aggravation/annoyance" we've discussed and tried to minimize throughout this PR.

That'd require a user to submit a query like `my_enum_field:i*`, which is presumably a pretty rare case.  But you can imagine a user maybe doing that as a shorthand way to get docs matching a few target enum values

solr/solr-ref-guide/modules/configuration-guide/pages/caches-warming.adoc
gerlowskija: Yeah, maybe not SEO purposes.  But I know there have been discussions about tracking changes to page names (I think for the purposes of the version-switcher our ref-guide has now?)

Anyway, something for another day...

solr/core/src/java/org/apache/solr/search/ComplexPhraseQParserPlugin.java
gerlowskija: I did find a way to make this change, but ended up backing off of it.  (I pushed that change to this PR but then reverted it, if you're curious to what was involved.)

A few reasons for that back off:

1. It ended up causing some complications on the Lucene-side.  ComplexPhraseQueryParser (the Lucene class) makes certain type assumptions about the Query instances it handles.  Injecting our own FieldType/schema aware query-creation logic into that class breaks those expectations in a pretty big way and causes a lot of test failures 😦 
2. It's orthogonal to my main goal in this PR, and I'm leery of scope creep.
3. If we do want to address this, it's probably something that merits its own ticket to investigate and fix.  After all, this probably isn't the only place in Solr that calls directly into a Lucene QueryParser without somehow creating sub-queries in a FieldType-aware way.

solr/core/src/java/org/apache/solr/search/ComplexPhraseQParserPlugin.java
```diff
@@ -134,6 +135,14 @@ protected Query newWildcardQuery(org.apache.lucene.index.Term t) {
               }
             }
 
+            @Override
+            protected Query getPrefixQuery(String field, String termStr) throws ParseException {
+              final var query = super.getPrefixQuery(field, termStr);
+              QueryUtils.ensurePrefixQueryObeysMinimumPrefixLength(
+                  qParserReference, query, termStr);
+              return query;
```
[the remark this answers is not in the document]
gerlowskija: See [my response on a similar topic above](https://github.com/apache/solr/pull/2499#discussion_r1668958020)

gerlowskija:
AFAIK - the last outstanding question here is whether to bundle in a separate change to make ComplexPhraseQParser be schema-aware in how it creates its queries.  There's a few technical hurdles there unfortunately - so my plan (unless I hear any vetos) is to punt on that for now.  [Full discussion of the idea and subsequent issues here.](https://github.com/apache/solr/pull/2499#discussion_r1668958020).

I'll give a little more time for feedback, but I'll look to merge sometime this week otherwise

dsmiley (approved):
+1 looks good.

PR #2500, opened 2024-06-06
Remove 'IndexUpgraderTool' page from ref-guide

gerlowskija:
The IndexUpgraderTool doesn't have a clear use case for Solr users, and after some discussion on the mailing list we decided documenting it isn't worth the confusion it causes for users.

# Description

The ref-guide has an entire page to document Lucene's IndexUpgraderTool.  This made sense at one point, but the tool isn't of much practical use for Solr users today, and the callout in the docs ends up confusing more users than it helps.

# Solution

This PR removes the page from the ref-guide in future releases.

# Tests

N/A - refguide build and 'check'/'precommit' continue to succeed

# Checklist

- [x] I have REMOVED documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)


epugh (commented):
Isn't this a "major change" that should be documented for 10x??

gerlowskija:
As I see this PR, it's a pretty minor change - it's "just" removing some stale, unnecessary docs.

But if you'd prefer the removal be documented, I'm happy to add a notice somewhere (CHANGES.txt?).  What did you have in mind?

epugh:
So the code change, i.e removing docs, no big deal.  However, IF you are a long time user of Solr, you may have used the tool in the past, and then go look for it, thinking "Oh, it's time to upgrade, let me go check the state of IndexUpgraderTool" and then not find it anymore.   So not a major code change, but still, soemthing I can imagine folks would WANT to know is no longer a "thing to use".   Especially since I do see that https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/IndexUpgrader.java remains, and there isn't anything saying "don't use this".    So adding a mention in https://solr.apache.org/guide/solr/latest/upgrade-notes/major-changes-in-solr-9.html would save a lot of us I suspect!

gerlowskija:
I think you're right that we should give confused users who might remember the docs _some_ signpost, but I opted to not to do that in "Major Changes"

Largely because nothing has changed 😛 The tool is still available.  It's as safe as it's ever been.  And it's utility in the Solr context hasn't changed either.

That, and "Major Changes" is dense with really crucial information.  A signpost comment like we're describing seemed a bit out of place there when I wrote it up.

Anyway, I ended up adding the note in the "upgrading-a-solr-cluster.adoc" page instead.  That's the parent page that "indexupgrader-tool.adoc" was nested under previously, and topically it's obviously very relevant.

If you disagree, lmk!

gus-asf:
This seems fine, though maybe adding something explaining why we removed it like: 

"The references to this tool were removed because in the vast majority of cases it is better to re-index your data with the new version instead. Also note that, by design, the tool can only be used once per index. Using the tool to upgrade 7 to 8 and then later upgrade to 9 is impossible so relying on it is not a good long term plan." 

after the first sentence would be good?

epugh (approved):
LGTM

epugh:
@gerlowskija want me to shepherd merging?  Looks like it is ready!

gerlowskija:
Sorry, I lost track of this, but I can get it merged.  Thanks for following up!

PR #2497, opened 2024-06-04
SOLR-17321: Bump minimum required Java version to 21

iamsanjay:
https://issues.apache.org/jira/browse/SOLR-17321

We are upgrading the minimum Java version for Solr main branch to 21. However, at the same, It has been suggested to be not so aggressive with SolrJ (and thus solr-api, a dependency) Java version – setting it to 17.

**Depreciated API**
1.  AccessController [Deprecated since 17 and marked for removal]
2. AccessControlContext [Deprecated since 17 and marked for removal]
3. java.net.URL ctors [Deprecated since 20]
4. SecurityManager [Deprecated since 17 and marked for removal]
5. Thread#getId() [Deprecated in Java 19]

To not make any extra code changes below annotation were used:

```
 @SuppressForbidden(reason = "<custom message>") //Still need to decide what the message should be for all. Thinking of adding "Deprecated in Java <version>"

@SuppressWarnings("removal")
```

However, at few places, for demonstration purpose, URL has been created via new suggested way instead of using the ctors:

```
url = new URL(urlStr);
url = URI.create(urlStr).toURL();
```
Or,
 ```
URL solrURL = new URI(solrUrl).toURL();
```

Also, Bump up the ecj version to 3.37 and ca.cutterslade.analyze to 1.9.2.

# Checklist

Please review the following and check all that apply:

- [x] I have reviewed the guidelines for [How to Contribute](https://github.com/apache/solr/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.
- [x] I have created a Jira issue and added the issue ID to my pull request title.
- [x] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [x] I have developed this patch against the `main` branch.
- [x] I have run `./gradlew check`.
- [] I have added tests for my changes.
- [ ] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

PR #2501, opened 2024-06-06
SOLR-17321:  Remove Deprecated URL API in Preparation for Java 21

iamsanjay:
[SOLR-17321](https://issues.apache.org/jira/browse/SOLR-17321)

As we started preparing (https://github.com/apache/solr/pull/2497 ) the move to change the Java version for the main branch, It has been decided to merge the deprecated API changes first, and thus reducing the change burden for https://github.com/apache/solr/pull/2497.

Here we are replacing the deprecated URL with new Syntax using URI.
```
//Deprecated
URL url = new URL(urlStr);
```
to 
```
URL url = URI.create(urlStr).toURL();
```

# Checklist


iamsanjay:
Though `new Locale` is deprecated, however the newer syntax `Locale.of` is not available in Java 11. Therefore It can only be done after migrating to Java21 or along with it.

Should we use forbidden APIs to Suppress the API which is forbidden in Java 21? `Thread#getID` and some other stuff related to `SecurityManager`.

epugh (approved):
This does make sense as the right route!  Lgtm!!!

epugh:
Seems a shame to add a to. Of suppress for locale and then remove it...  might be cleaner to do that one as part of the upgrade.

iamsanjay:
> As far as security manager goes, don't we need to just tackle that and solve the issue? Versus suppress it and ignore the problem?

Removing a SecurityManager will not be a simple task as some of these settings, I believe, were put in place to fix Security issues. For now, using forbidden would clear the path to the upgradation, Lucene did the same thing. However, now that the JDK has marked it for removal, future JDK upgradation would only be possible after removing it So we should start looking into it right after upgrading to Java 21. 

And that's why I suggested marking SecurityManager, and related stuff, forbidden which is a viable solution for now.

epugh:
> > As far as security manager goes, don't we need to just tackle that and solve the issue? Versus suppress it and ignore the problem?
> 
> Removing a SecurityManager will not be a simple task as some of these settings, I believe, were put in place to fix Security issues. For now, using forbidden would clear the path to the upgradation, Lucene did the same thing. However, now that the JDK has marked it for removal, future JDK upgradation would only be possible after removing it So we should start looking into it right after upgrading to Java 21.
> 
> And that's why I suggested marking SecurityManager, and related stuff, forbidden which is a viable solution for now.

yeah..   It sort of feels like that is the opposite of "get the code ready for java 21 first" like you are doing here with Url.   However I also understand that it may spiral into a bigger thing.    The important thing is to keep moving forward!   Thanks for this.

iamsanjay:
Actually I started looking into it, the first issue that I found was "XXE attack in SolrConfig.xml" and a proposed fix (https://issues.apache.org/jira/browse/SOLR-12316). Afterwards the more improved version of fix involved SecurityManager as in running Solr in sandbox via SecurityManager (https://issues.apache.org/jira/browse/SOLR-13984).

And this was the comment!

> The correct way to fix all issues we have seen the last time is very simple: LET'S RUN SOLR INSIDE A SECURITY MANAGER IN PRODUCTION (like in tests). Elasticsearch is doing this, so please please let's do this instead. But this requires to finally get rid of the webapplication and start.jar and add our own bootstrapping (like in tests) that configure Jetty and Security Manager from our own org.apache.solr.bootstrap.Main.java (or similar).  -- Uwe

But I will keep on investigating, How and where Solr using SecurityManager?

epugh:
I like the forbidden api addition!

iamsanjay:
All the usages of `new URL` (except one) has been replaced with `URI.create(url).toURL()`. The exception was one specific case which is testing the presence of caret in the URL as It was an old issues (https://issues.apache.org/jira/browse/SOLR-5971) that failed due to the caret. Now replacing it with new Syntax fails because caret is in the disallowed list for URI and thus failed. I do not want to encode the character, because the whole point of test is to test caret without encoding it, instead I used `@SuppressForbidden`.

Because #2497 is still in progress, So, meanwhile,  I added "new URL" to the forbidden api to make sure no one would use it.

solr/CHANGES.txt
```diff
@@ -127,6 +127,8 @@ Improvements
 * SOLR-17300: Http2SolrClient.Builder.withHttpClient now copies HttpListenerFactory (e.g. for auth, metrics, traces, etc.)
   (Sanjay Dutt, David Smiley)
 
+* SOLR-17321: Remove Deprecated URL and replace it with URI in Preparation for Java 21 (Sanjay Dutt, David Smiley)
```
dsmiley, line 130: I would put this in "Other"; it's code maintenance in nature.  "Improvement" should be something a user might perceive in some way.

iamsanjay:
I added the changes under 9.7, I did not discussed this. So Should we consider this change for 10 only or release it in 9.7 as well?

HoustonPutman:
Backport it to 9.7 IMO

epugh:
Agreed.   When I added basic auth to the solr cli I did NOT backport it, and now it's a constant source of pain.

iamsanjay:
Going to merge this and back port to 9x on Friday, around 7PM IST.

dsmiley (commented):
Why use ForbiddenAPI for a deprecated thing?  The JDK (with IDE support too) and javac already handle it in its own way. ForbiddenAPI principally exists for non-deprecated things we insist on not using. 
cc @uschindler

iamsanjay:
> Why use ForbiddenAPI for a deprecated thing? 

As we are still in migration phase, It will not be deprecated until we move onto Java 21. Meanwhile, If someone tries to use it , they will get the warning message with an alternative API recommendation.

uschindler:
> Why use ForbiddenAPI for a deprecated thing? The JDK (with IDE support too) and javac already handle it in its own way. ForbiddenAPI principally exists for non-deprecated things we insist on not using. cc @uschindler

In Java 11 it is not yet deprecated. So this allows to forbid it now.

Actually once we are on Java 21, the forbiddenapis signatures will complain, too (as deprecated APIs are reported, too). If you ask why forbiddenapis reports deprecated stuff: It's because normally during compilation, JDK only reports a warning for all deprecations (not only in the JDK also in your own code or dependencies). Forbiddenapis allows to restrict only JDK-forbidden stuff and you can keep other deprecations supressed. This is especially important for API developers, because you deprecate your own stuff but still have tests for it.

solr/solrj/src/test/org/apache/solr/common/util/ContentStreamTest.java
```diff
@@ -161,7 +161,7 @@ public void testURLStreamCSVGZIPExtention() throws IOException {
     }
 
     ContentStreamBase stream =
-        new ContentStreamBase.URLStream(new URL(file.toURI().toASCIIString()));
+        new ContentStreamBase.URLStream(URI.create(file.toURI().toASCIIString()).toURL());
```
uschindler, line 164: this code makes no sense (also at other places above). Just call `file.toURI().toURL()`

uschindler: see other occurrences of this in this and other files

solr/core/src/test/org/apache/solr/cloud/TestRequestForwarding.java
```diff
@@ -69,6 +71,12 @@ public void testMultiCollectionQuery() throws Exception {
     }
   }
 
+  // Restricting the Scope of Forbidden API
+  @SuppressForbidden(reason = "java.net.URL#<init> deprecated since Java 20")
+  private URL createURL(String url) throws MalformedURLException {
```
uschindler, line 76: why not use `URI.create().toURL()`?

uschindler (changes requested):
Actually this is partly a bit strange. In general the move should be to go away from URL completely. Theres code changed here which at some point creates an URL but later passes it to Jetty as URI. So It would be better to replace all URL by URI from beginning.

There are also some roundtrip conversions, please remove them.

solr/core/src/test/org/apache/solr/cloud/TestRequestForwarding.java
uschindler: Ah you mentioned that above.

dsmiley:
What I love about our community is that we have so much expertise to call upon; thanks Uwe :-)

> Forbiddenapis allows to restrict only JDK-forbidden stuff

Then lets do that and not single out a specific thing (URL).

solr/core/src/java/org/apache/solr/cli/PostTool.java
```diff
@@ -389,7 +389,7 @@ private void doWebMode() {
       numPagesPosted = postWebPages(args, 0, out);
       info(numPagesPosted + " web pages indexed.");
 
-    } catch (MalformedURLException e) {
+    } catch (URISyntaxException e) {
```
uschindler, line 392: URISyntaxException is only thrown by `new URI()`, the create function does not. Not sure how we should handle this.

See documentation of `URI#create(String)`: "This convenience factory method works as if by invoking the [URI(String)](https://docs.oracle.com/javase%2F8%2Fdocs%2Fapi%2F%2F/java/net/URI.html#URI-java.lang.String-) constructor; any [URISyntaxException](https://docs.oracle.com/javase%2F8%2Fdocs%2Fapi%2F%2F/java/net/URISyntaxException.html) thrown by the constructor is caught and wrapped in a new [IllegalArgumentException](https://docs.oracle.com/javase%2F8%2Fdocs%2Fapi%2F%2F/java/lang/IllegalArgumentException.html) object, which is then thrown."

uschindler: I would make this consistent to other code and prefer a clear exception.

uschindler (approved):
Small change regarding exception handling in the POST tool, otherwise looks fine to me.

solr/core/src/java/org/apache/solr/cli/PostTool.java
iamsanjay: I started using `URI.create` for the same reason that you just explained as it throws unchecked exception which allows us to avoid explicitly handling the `URISyntaxException`.

Should we switch to `new URI` wherever we can without changing too much? I am open to that.

uschindler: I think both is fine. I just noticed that the "catch" here looks obsolete (I wonder why compiler did not complain).

iamsanjay: `doWebMode()` is calling `appendUrlPath` that, as per new changes, is throwing `URISyntaxException`.

uschindler: OK, as we use new URI here, it might be better to do this also in the main code path. This would bring consistent exceptions.

iamsanjay: I have replaced the `URI.create` with `new URI` in `PostTool.java`.  Also replaced in few test classes as well.

epugh: Looks great!

uschindler: > I have replaced the `URI.create` with `new URI` in `PostTool.java`. Also replaced in few test classes as well.

Actually you did this everywhere, which I am not sure if its a good idea. In PostTool i recommend to change this to get a consistent error message in both cases: for original URI and for the appendUrlPath() method.

I am not a big fan of checked Exceptions, so I would have used new URI only to make it consistent with the code around it.

iamsanjay: I've reverted the changes and updated it only for Post.

iamsanjay:
```
git diff f2841ac..0007bb082 --name-only
solr/CHANGES.txt
solr/core/src/java/org/apache/solr/cli/PostTool.java
```

iamsanjay:
Ran the failed test multiple times and it's passing.  Flaky!
```
  - org.apache.solr.client.solrj.TestLBHttp2SolrClient.testTwoServers (:solr:solrj)
    Test history: https://ge.apache.org/scans/tests?search.rootProjectNames=solr-root&tests.container=org.apache.solr.client.solrj.TestLBHttp2SolrClient&tests.test=testTwoServers http://fucit.org/solr-jenkins-reports/history-trend-of-recent-failures.html#series/org.apache.solr.client.solrj.TestLBHttp2SolrClient.testTwoServers
    Test output: /tmp/src/solr/solr/solrj/build/test-results/test/outputs/OUTPUT-org.apache.solr.client.solrj.TestLBHttp2SolrClient.txt
    Reproduce with: ./gradlew :solr:solrj:test --tests "org.apache.solr.client.solrj.TestLBHttp2SolrClient.testTwoServers" -Ptests.jvms=96 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=BC6F7BFC1228CAE8 -Ptests.timeoutSuite=600000! -Ptests.file.encoding=ISO-8859-1
```

It recently started failing (I hope I am looking at it in a right way!)

Screenshot 2024-06-17 at 9 48 33 PM

PS :- Thank You @dsmiley. Test history URL in Action!

iamsanjay:
If there are no more comments/issues, I would like to merge this by Friday. Thanks!

uschindler (approved):

epugh (approved):

solr/test-framework/src/java/org/apache/solr/handler/BackupRestoreUtils.java
```diff
@@ -109,7 +110,7 @@ public static void runReplicationHandlerCommand(
   }
 
   static void executeHttpRequest(String requestUrl) throws IOException {
-    URL url = new URL(requestUrl);
+    URL url = URI.create(requestUrl).toURL();
```
HoustonPutman, line 113: This is causing `TestSolrCoreSnapshots.testBackupRestore` to fail on Windows machines.

```
java.lang.IllegalArgumentException: Illegal character in query at index 117: http://127.0.0.1:64027/solr/admin/cores?action=BACKUPCORE&core=SolrCoreSnapshots_shard1_replica_n1&name=u&location=C:\Users\jenkins\workspace\Solr-main-Windows\solr\core\build\tmp\tests-tmp\org.apache.solr.core.snapshots.TestSolrCoreSnapshots_9D4706448466E215-001\tempDir-002&incremental=false&commitName=ydfbp
	at __randomizedtesting.SeedInfo.seed([9D4706448466E215:2BFC06FAB6D52C5B]:0)
	at java.base/java.net.URI.create(URI.java:883)
	at org.apache.solr.handler.BackupRestoreUtils.executeHttpRequest(BackupRestoreUtils.java:113)
```

PR #2502, opened 2024-06-06
Clean openApiGenerate output dir before generating

gerlowskija:
# Description

Before compiling SolrJ, the gradle build creates an OpenAPI Spec ("OAS") describing our JAX-RS v2 APIs and then uses that spec to generate SolrRequest ".java" files.  The responsible task (`solr:solrj:openApiGenerate`) adds newly generated code to `solrj/build/generated`, but it doesn't (by default) wipe the directory at the start of each invocation.

This can cause generated files to "leak" across branches - sticking around in `solrj/build/generated` even when the developer has changed to a branch that shouldn't contain the file.

These leaks can cause compilation issues, as pointed out by @dsmiley recently [on the dev list](https://lists.apache.org/thread/rth0hbo8y0gtcvrntyvg9khvfl8xh4x8).

# Solution

The solution is to have the `solr:solrj:openApiGenerate` task clear its output directory each time it runs.  Luckily the OpenAPI tooling supports this, and it's a one-line change in build.gradle.
 
# Tests

Manual build testing.

# Checklist

### DIFF
--- a/solr/solrj/build.gradle
+++ b/solr/solrj/build.gradle
@@ -132,6 +132,7 @@ openApiGenerate {
   generateModelTests = false
   generateApiDocumentation = false
   generateModelDocumentation = false
+  cleanupOutput = true
   additionalProperties = ["modelPackage": "org.apache.solr.client.api.model"]
 }
 


epugh (approved):
LGTM!

dsmiley (approved):
It's "safe" and will reduce user annoyance (as I've experienced / felt) so I approve.

I don't know if there is a notable negative performance impact, as there would be if this was a java compiler, which knows not to re-generate unchanged things.  Update: I did a little trivial experimentation just now with this and I don't think it's a big concern.

HoustonPutman:
This looks good to me.

It will be even better when https://github.com/OpenAPITools/openapi-generator/issues/17380 is fixed, which will stop the task from running as often as it is (the task will be skipped since the input & output will be unchanged).

PR #2503, opened 2024-06-06
SOLR-18327: Balance replicas API does not balance duplicate replicas on same node

janhoy:
https://issues.apache.org/jira/browse/SOLR-18327

Spinoff from Slack https://the-asf.slack.com/archives/CEKUCUNE9/p1717603676039659

# Problem

I had a cluster with many collections, where several of them had multiple replicas of the same shard on the same node. Configured affinity plugin and called the balance replicas API, and some of my collections were healed, but a few were left with replicas on same node, while it could easily have moved them to other nodes.

Note that this won't happen if placement plugin is configured at collection create time, then it is an absolute requirement to have enough different eligible nodes to place the replicas.

# Root cause

`OrderedNodePlacementPlugin.computeBalancing` only moves replicas when nodes' weights differ, and no weight function accounts for two replicas of the same shard sitting on the same node. In a symmetric cluster (e.g. every node hosting the same number of cores) the balancer exits immediately with zero movements. Adding a duplicate penalty to the weight function would not help: every node with a duplicate carries the same penalty, so weights stay equal and the balancing loop never engages. The "one replica per shard per node" rule exists only as a hard constraint on *adding* replicas (`WeightedNode.canAddReplica`), so placement never creates duplicates — but balancing could never repair them either.

# Fix

Added a pre-pass `moveDuplicateShardReplicas()` in `OrderedNodePlacementPlugin.computeBalancing`, benefiting all placement plugins that extend it (Affinity, MinimizeCores, Simple, Random). Before weight-based balancing it finds shards with multiple replicas on one node and moves the extras to the lowest weighted node that passes `canAddReplica` (which enforces the no-duplicate rule plus each plugin's disk/type/collocation constraints), respecting `canRemoveReplicas` on the source.

# Tests

New unit test with a sample 4 node cluster, with two collections, each with 1 shard and four replicas:

| Coll | N0 | N1 | N2 | N3 |
|--------|--------|--------|--------|--------|
| A | A1, A2 |  | A3, A4 |  |
| B |  | B1, B2 |  | B3, B4 |

Balance now moves replicas to unique nodes:

| Coll | N0 | N1 | N2 | N3 |
|--------|--------|--------|--------|--------|
| A | A1 | A2 | A3 | A4 |
| B | B2 | B1 | B4 | B3 |

Two existing balancing tests per plugin (`testBalancingBareMetrics`, `testBalancingExistingMetrics` in Affinity and MinimizeCores variants) had expectations encoding the old behavior — their layouts include an NRT+TLOG duplicate on node 0 that was left in place because node 0 was the lightest node. The new expected plans reach the same core distribution as before, but without duplicates.


janhoy:
Follow-up after a QA pass of the fix (4c8006706c7):

- **Target selection now uses projected weight**: the duplicate-replica pre-pass picked the target node by its *current* weight, while `computePlacements()` uses `calcRelevantWeightWithReplica()` — the projected weight with the replica added. In multi-AZ clusters the old selection could move a duplicate into an already-over-represented AZ. The pre-pass now mirrors placement semantics.
- **New test** `testBalancingAvoidMultiReplicaOnNodeAcrossAZs` verifies a duplicate moves to the under-represented AZ rather than an empty node in the same AZ.
- Verified no explicit weight recomputation is needed after the pre-pass: node weights are computed on demand, and the `TreeSet` for the weight-based balancing loop is built after the pre-pass completes.
- All 54 placement unit tests pass (beasted with `-Dtests.iters=5`), plus the end-to-end `BalanceReplicasTest`.

Noted for possible follow-up (pre-existing, untouched): the balancing loop's move-acceptance condition can accept a move that makes the combined weight strictly worse when non-conserved weight terms (e.g. AZ skew) are involved.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/OrderedNodePlacementPlugin.java
```diff
@@ -305,6 +312,69 @@ public BalancePlan computeBalancing(
         .createBalancePlan(balanceRequest, replicaMovements);
   }
 
+  /**
+   * Move replicas that share a node with another replica of the same shard to other nodes, since
+   * multiple replicas of the same shard on one node give neither availability nor capacity
+   * benefits. Placement will never create such a state, per the default {@link
```
HoustonPutman, line 318: So this is true for most of the placement plugins, but the affinity actually does allow it if necessary (if I recall correctly). But I'm saying this for additional information rather than a request to change the comment.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/OrderedNodePlacementPlugin.java
```diff
@@ -305,6 +312,69 @@ public BalancePlan computeBalancing(
...
+    sourceNodes.sort(Comparator.comparing(node -> node.getNode().getName()));
+    for (WeightedNode sourceNode : sourceNodes) {
+      Map<String, List<Replica>> replicasPerShard =
+          sourceNode.getAllReplicasOnNode().stream()
+              .collect(
+                  Collectors.groupingBy(
+                      replica ->
+                          replica.getShard().getCollection().getName()
+                              + "%"
+                              + replica.getShard().getShardName()));
```
HoustonPutman: We do this all the time. Would be useful to have a `getShard().getUniqueShardId()` utility or something equivalent.

HoustonPutman (approved):
I like it. Do the dirty work of greedily move the replicas off of the over-packed nodes. Then the balancing logic will do the less-greedy optimization.

solr/core/src/java/org/apache/solr/cluster/placement/plugins/OrderedNodePlacementPlugin.java
janhoy: Done. Now using it only in this location, but I'll make another PR after this one to use the new method also to simplify `AffinityPlacementFactory` with the same method.

PR #2506, opened 2024-06-07
SOLR-17322: Make RankQuery.getTopDocsCollector use covariant generic types

stephen-woods:
https://issues.apache.org/jira/browse/SOLR-17322 

# Description

Updates `RankQuery` so that the `TopDocsCollector` it returns is properly covariant instead of invariant. This will once again allow rank queries to use custom `TopDocsCollectors` that operate on types that _extend_ `ScoreDocs`, and not just `ScoreDoc`. Currently,  any attempt to use a class other than `ScoreDoc` will result in a compilation error.

This is a trivial update: A one line change that corrects the method signature, so that it matches the call site in `SolrIndexSearcher`.

# Tests

After this change, custom concrete `RankQuery` classes that return `TopDocsCollectors` that use subclasses of `ScoreDocs` now compile. The existing `RankQueryTest` now has additional static classes that are intended to ensure that support of covariant `ScoreDocs` is supported in rank queries; this test class will not compile if support of this feature is removed.

# Checklist

- [x] I have run `./gradlew check`. Though it fails on unrelated licensing header issues for `:solr:modules:rat`


stephen-woods:
If this gets accepted, I'd like to back port it to 9.X branches

cpoerschke (approved):
LGTM.

Wondering if it might be worthwhile to also add simple test coverage in `RankQueryTest` or so, so that if that coverage had been there previously, then the compile error mentioned on the JIRA would have happened when the raw types removal happened.

```
static class MyScoreDoc extends ScoreDoc {
static class MyTopDocsCollector ... {
static class MyRankQuery extends RankQuery {
```

cpoerschke:
> If this gets accepted, I'd like to back port it to 9.X branches

Just FYI, whoever merges this PR would cherry-pick to `branch_9x` typically also i.e. no need for a separate pull request for back porting.

stephen-woods:
I added those static classes as you requested.

solr/core/src/test/org/apache/solr/search/RankQueryTest.java
```diff
@@ -106,4 +119,65 @@ public void testPluggableCollector() {
...
   }
+
+  // The following static classes are intended to ensure that support of covariant
+  // ScoreDocs is supported in rank queries. MyRankQuery will fail to compile
+  // if covariant ScoreDocs are not supported because it returns a TopDocsCollector
+  // for MyScoreDoc (a subtype of ScoreDoc).
+  static class MyScoreDoc extends ScoreDoc {
+    public int someOtherField = 0;
+    public MyScoreDoc(int doc, float score, int someOtherField) {
+      super(doc, score, -1);
```
cpoerschke, line 130: minor/subjective

```suggestion
    public MyScoreDoc(int doc, float score, int shardIndex, int someOtherField) {
      super(doc, score, shardIndex);
```

or

```suggestion
    public MyScoreDoc(int doc, float score, int someOtherField) {
      super(doc, score);
```

solr/core/src/test/org/apache/solr/search/RankQueryTest.java
```diff
@@ -106,4 +119,65 @@ public void testPluggableCollector() {
         "//result/doc[2]/str[@name='id'][.='6']",
         "//result/doc[1]/str[@name='id'][.='5']");
   }
+
+  // The following static classes are intended to ensure that support of covariant
+  // ScoreDocs is supported in rank queries. MyRankQuery will fail to compile
+  // if covariant ScoreDocs are not supported because it returns a TopDocsCollector
+  // for MyScoreDoc (a subtype of ScoreDoc).
+  static class MyScoreDoc extends ScoreDoc {
+    public int someOtherField = 0;
```
cpoerschke, line 128:
```suggestion
    final public int someOtherField = 0;
```

cpoerschke (approved):

solr/core/src/test/org/apache/solr/search/RankQueryTest.java
cpoerschke: My bad, I meant `final public int someOtherField` here i.e. forgot to remove the ` = 0` when making the suggestion. https://github.com/apache/solr/pull/2506/commits/4bd9f94210e66ad1c0bd1d3abf5226fbc608307f to add the `final`.

stephen-woods: Oh, I thought you wanted it final because I didn't expose it originally in the constructor. It's fine in this test example, but in practice, it would need to remain mutable so that the `ScoreDoc` can be reused and changed within a `PriorityQueue` - none of the other fields in `ScoreDoc` are final.

cpoerschke: Ah, good point about it needing to be mutable in practice. I'd noticed it being public and there being no accessors and presumed that modification would/should be via an accessor and marking `final` would allow access without modification (presuming that no modification is needed after construction).

Good point about none of the other `ScoreDoc` fields being `final` -- so yes, let's keep things consistent here then and for `someOtherField` to be mutable too. Good catch!

cpoerschke: https://github.com/apache/solr/pull/2506/commits/c078d39a5cf127a1891f16ca0d3cf668b6a4596e reverts my earlier commit.

cpoerschke (approved):
LGTM. Planning to wait a few days before merging and back porting, in case of further suggestions etc.

Thanks @stephen-woods for diagnosing/reporting/fixing this issue!

madrob:
I don't think this is the right approach. If you can get your hands on a copy of **Effective Java, 3rd Edition**, this is **Item 31: Use bounded wildcards to increase API flexibility**. Later in that section it comes with a warning:

> Note that the return type is still Set<E>. Do not use bounded wildcard types as return types. Rather than providing additional flexibility for your users, it would force them to use wildcard types in client code.

I suspect that the SolrIndexSearcher method needs to be changed to `<T extends ScoreDoc> TopDocsCollector<T> buildTopDocsCollector` but that doesn't compile and I'm not sure what it would take.

stephen-woods:
Are you saying that we should add a generic type to RankQuery?
ie.
```java
public abstract class RankQuery<T extends ScoreDoc> extends ExtendedQueryBase {
  public abstract TopDocsCollector<T> getTopDocsCollector(
      int len, QueryCommand cmd, IndexSearcher searcher) throws IOException;

```

Though that approach would probably be more ideal given the reasons Bloch indicates in his book, it is by far more disruptive. All existing `RankQuery` subclasses would need to be modified including any custom classes that may exist in third party plugin libraries - it's a breaking change... something I was trying to avoid as much as possible.

All that effort to "make API flexibility" at the one and only call site: `SolrIndexSearcher`, a class that uses `instanceof` gratuitously all over the place.

stephen-woods:
Assuming I understood what @madrob was getting at when mentioning Bloch's flexibility argument,  I attempted to refactor this PR in order to to introduce a type parameter to the `RankQuery` class... 

Now, the `getTopDocsCollector` method returns an invariant type. **Note** that this actually makes things _less_ flexible, since the rank query can no longer return `TopDocCollectors` of different types (though, honestly... locking it down to a single type might actually be a good thing).

So as not to undo the work that has already been done on this PR, I'm including a link to a branch on my fork that includes this refactoring. Again, this is a breaking change that will cause compilation issues for third party plugin developers. I'm not convinced that this is appropriate for a minor version change on the 9.x branch, but it might be alright on main (10.X) if enough people like it.

If you want, I can make a separate PR for this new approach to make discussing a little easier.

https://github.com/stephen-woods/solr/tree/SOLR-17322generic

madrob:
The problem with introducing a bounded wildcard is that it also is a breaking change. A type parameter for RankQuery might be a good improvement, I'll take a look tomorrow. A lot of this code is very old and carries a lot of history from before people understood best (or just better) practices, so we might not be able to fix everything. Or it might be very disruptive as you noted

stephen-woods:
Fair enough... I recognize that introducing a bounded wild card would normally be a breaking change, but in this particular case, the boundary is consistent with the one that `TopDocsCollector` already has in place (and has had since the stone age). All `TopDocsCollectors` are upper bounded to `ScoreDoc`. 
```java
public abstract class TopDocsCollector<T extends ScoreDoc> implements Collector {
```

Given that.... it really isn't a breaking change. All previously RankQuery classes that worked in Solr 8.x work again after this PR change. It's actually the same thing at runtime as what the previous raw type had (because of the bounds in the `TopDocsCollector` class). Making it invariant was the change that was breaking.

But yeah... take a look at the other commit I made that introduces the type variable into `RankQuery` and see if that addresses your concerns.

solr/CHANGES.txt
```diff
@@ -220,6 +220,8 @@ Other Changes
 * SOLR-17347: EnvUtils: Removed "env" methods, as they were problematic.  Code should almost never
   refer to env vars.  (David Smiley)
 
+* SOLR-17322: Once again allow rank queries to use custom TopDocsCollectors that operate on types that extend ScoreDocs. (Stephen Woods via Christine Poerschke)
```
dsmiley, line 223: "Once again" -- can we be more specific ... since what Solr version?

dsmiley (approved):
FWIW I think we can make breaking changes to individual plugins (like this one) on a minor release.  I might be less inclined to say so on one of our most popular plugin abstractions, but RankQuery is more on the obscure end.

solr/CHANGES.txt
stephen-woods: I updated the CHANGES to include that Solr 9.0 broke it.

dsmiley:
Cool; maybe (@cpoerschke reviewer and listed in CHANGES.txt) can merge this in a day or so to make 9.7.  @madrob now's your chance to object if you have an issue with this.

PR #2522, opened 2024-06-17
SOLR-17333: Various rate limiting fixes

magibney:
1. fix live-update of configuration
2. fix slot borrowing bug
3. cleaner state tracking via try-with-resources
4. fix refguide documentation to mention limitations and request header requirement

See [SOLR-17333](https://issues.apache.org/jira/browse/SOLR-17333)


noblepaul:
The following are not supported today or they don't have any effect whatsoever.I think we should just get rid of them and make the feauture simpler. 
```
 "guaranteedSlots":5,
"slotBorrowingEnabled":true,
```

solr/core/src/test/org/apache/solr/servlet/TestRequestRateLimiter.java
```diff
@@ -282,4 +289,130 @@ public RateLimitManager build() {
...
+            true,
+            guaranteed,
+            20,
+            allowed /* allowedRequests */,
+            true /* isSlotBorrowing */);
+    RequestRateLimiter limiter = new RequestRateLimiter(config);
+    ExecutorService exec = ExecutorUtil.newMDCAwareCachedThreadPool("tests");
+    try (Closeable c = () -> ExecutorUtil.shutdownAndAwaitTermination(exec)) {
+      for (int j = 0; j < 5; j++) {
+        System.err.println("for " + allowed + "/" + guaranteed);
```
noblepaul, line 314: is this a debug statement? Shouldn't we use logger ?

magibney:
hmm re: getting rid of slot borrowing entirely -- it would surely simplify things. It's true that the current implementation is such that this feature has never been relevant. But it's been documented for some time, and I can see why it would be desirable. Ordinarily I'd say that removing a relatively long-standing public-facing feature would not be done lightly. This surely an unusual case though, in the sense that it has never worked as advertised.

We should definitely consider as a factor in the decision whether the feature is expected to be useful. I may not be the best advocate for the feature -- perhaps @atris or @stillalex might have opinions?

The main though I have is actually _against_ the feature -- or at least, against it as currently implemented. Basically, in most cases I suspect one would want a hard limit per type (since different request types have different performance impacts and functional impacts). So allowing slot borrowing to be only configured on the type being borrowed _from_ would allow all other requests to borrow any slot, limited only by the overall max number of borrowable requests, which is a global limit. This doesn't seem that useful to me 🤷.

If we want to enable more complex configurations, it would be more flexible (and in some ways, simpler) to support multiple `Solr-Request-Type` buckets per request, all of which must be acquired for the request to proceed. Though this could lead to some pathological configurations where a request could acquire a slot in one bucket (blocking other requests) only to be blocked themselves upon trying to acquire a slot in another bucket.

Another option would be to configure a "borrowing pool", limiting the scope of which request types may borrow from each other. This approach would avoid the pathological configuration of the "multiple `Solr-Request-Type`" approach, and also avoid the "global borrowing pool" problem that would make the current implementation (IMO) unuseful.

atris:
I am aligned with adding a per bucket based borrowing pool - but did not quite understand the comment around current implementation becoming not useful (or are you referring just to the current global borrowing pool implementation?)

magibney:
I was referring to the fact that the current implementation only supports `Solr-Request-Type: QUERY` -- and as long as only one request type is supported, the concept of borrowing across request types is not really applicable.

atris:
This was added more from a perspective of more types coming in - since the interfaces are designed that way. I agree that in the current form, it inherently leads to an implicit limit within the resource.

I am curious though - what is the main issue that you see with sustaining slow borrowing in its current form - it essentially enforces a per bucket slot borrowing policy. We can add a per request type limit if it makes more sense to explicitly limit the borrowing within a request type

magibney:
>This was added more from a perspective of more types coming in - since the interfaces are designed that way. I agree that in the current form, it inherently leads to an implicit limit within the resource.

Right, totally understood.

> what is the main issue that you see with sustaining slow borrowing in its current form

@noblepaul raised this as a possibility, iiuc just for the sake of simplicity. In this PR I set out to (among other things) improve the implementation of slot borrowing (the extant implementation could yield sub-optimal behavior under some circumstances). My default approach was to make the feature work, but given that it can't really be used practically in its current state, I thought it was fair to consider Noble's suggestion as if from a clean slate (making the case for slot-borrowing).

atris:
I am still confused - the parameters, in their current form, are functional, albeit not useful?

magibney:
I think we're on the same page? The implementation is built out to support the "borrowable slot" concept, but in a single-request-type context, it's not useful (yet).

This PR proposes changes to the current implementation because the current implementation handles the distinction between borrowable/guaranteed by using 2 semaphores, and [trying serially to acquire each](https://github.com/apache/solr/blob/709a1ee27df23b419d09fe8f67c3276409131a4a/solr/core/src/java/org/apache/solr/servlet/RequestRateLimiter.java#L61-L78). No matter what order you do this in, with this approach it's possible (indeed, likely under heavy load) for the following things to happen, for a request of the owning request type:
1. if all slots are taken, the request will block twice as long as it should (once for `guaranteedSlotsPool`, once for `borrowableSlotsPool`
2. if all guaranteed slots are taken, but none of the borrowable slots are taken, the request will still block for `waitForSlotAcquisition` millis before falling back to `borrowableSlotsPool` (so can block although plenty of slots are available)

noblepaul:
I suggested removing those in favor of something simpler and easier to understand as the current impl is not working because of lack of multiple types

magibney:
On further reflection, I think the question of whether to remove or fix the "slot borrowing" impl can remain out of scope for this PR.

The goal of this PR is to fix demonstrable bugs in the existing code. Fixing the issues with slot borrowing has no impact at all on whether we're free to remove it, so if we want to remove it, we can do so in a follow-up PR. The only change that would make it more awkward to remove slot borrowing would be enabling different request types (since as long as only one request type is possible, slot borrowing has no practical effect. This PR has no effect on whether slot borrowing _works_, it just affects performance if/when it _does_ work).

So unless there are concerns with the substance of this PR (according to the scope of "fix what's there and has demonstrable bugs") I'll go ahead and merge this soon.

magibney:
Proposed CHANGES.txt entry:
```
* SOLR-17333: Rate-limiting feature: fix live-update of config (Michael Gibney)
```
^ focuses on the only practically public-facing feature that's changed -- other changes will not be functionally apparent to users.

PR #2524, opened 2024-06-18
Refactor: QueryResult: don't provide to SolrIndexSearcher

dsmiley:
IMO it's wrong for SolrIndexSearcher.search(QueryCommand) to also take a QueryResult param and not return it.  It's an "out" parameter, but there is no need for this!  This PR deprecates that method and replaces it with one that does the right thing.  In one spot, I see QueryResult early segment termination being set up in QueryComponent but that should be done in SolrIndexSearcher.

I also deprecated many low-value SolrIndexSearcher methods involving QueryCommand; a QueryCommand is a superior API to methods taking lots of parameters.  Some weren't even called.  Some were in a place or two.  I inlined some of those albeit  not all.  Some were called by SolrPluginUtils methods that themselves felt like questionable utility methods so I deprecated them too; they are used by DebugComponent and MLTHandler but IMO if they need to exist, DebugComponent would be the best home as it's the principal place for debut output to be.


solr/core/src/test/org/apache/solr/search/TestSearchPerf.java
```diff
@@ -173,7 +172,14 @@ private long doListGen(Query q, List<Query> filt) throws Exception {
     long ret = 0;
     for (int i = 0; i < ITERATIONS; i++) {
       DocList l =
-          searcher.getDocList(q, filt, (Sort) null, 0, 10, SolrIndexSearcher.NO_CHECK_QCACHE);
+          new QueryCommand()
+              .setQuery(q)
+              .setFilterList(filt)
+              .setOffset(0)
```
cpoerschke, line 178:
```suggestion
```

solr/core/src/java/org/apache/solr/handler/MoreLikeThisHandler.java
```diff
@@ -165,8 +166,12 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
           // Find the base match
           DocList match =
-              searcher.getDocList(
-                  query, null, null, matchOffset, 1, flags); // only get the first one...
+              new QueryCommand()
+                  .setQuery(query)
+                  .setOffset(matchOffset)
+                  .setLen(1) // only get the first one...
+                  .setFlags(flags)
+                  .search(searcher).getDocList();
```
cpoerschke, line 174: precommit
```suggestion
                  .search(searcher)
                  .getDocList();
```

solr/core/src/java/org/apache/solr/handler/component/QueryComponent.java
```diff
@@ -403,14 +403,9 @@ public void process(ResponseBuilder rb) throws IOException {
...
     req.getContext().put(SolrIndexSearcher.STATS_SOURCE, statsCache.get(req));
 
-    QueryResult result = new QueryResult();
-
     cmd.setSegmentTerminateEarly(
         params.getBool(
             CommonParams.SEGMENT_TERMINATE_EARLY, CommonParams.SEGMENT_TERMINATE_EARLY_DEFAULT));
-    if (cmd.getSegmentTerminateEarly()) {
-      result.setSegmentTerminatedEarly(Boolean.FALSE);
-    }
```
cpoerschke, line 413: observation: deferring this bit also helps with the `// not supported, silently ignore any segmentTerminateEarly flag` bit at lines 420/421 415/416 below i.e. if the results have `Boolean.FALSE` in future then it would not be despite the silent ignoring.

solr/core/src/java/org/apache/solr/handler/component/QueryComponent.java
```diff
@@ -421,22 +416,22 @@ public void process(ResponseBuilder rb) throws IOException {
...
+          doProcessGroupedDistributedSearchFirstPhase(rb, cmd);
           return;
         } else if (params.getBool(GroupParams.GROUP_DISTRIBUTED_SECOND, false)) {
-          doProcessGroupedDistributedSearchSecondPhase(rb, cmd, result);
+          doProcessGroupedDistributedSearchSecondPhase(rb, cmd);
           return;
         }
 
-        doProcessGroupedSearch(rb, cmd, result);
+        doProcessGroupedSearch(rb, cmd);
```
cpoerschke, line 426: subjective/style: if/since we're changing this code area, how about adding a `else { }` wrap and then removing the two `return;` in the first and second phases?
```suggestion
        else {
          doProcessGroupedSearch(rb, cmd);
        }
```

solr/core/src/java/org/apache/solr/handler/component/QueryComponent.java
dsmiley: Do you mean for grouping, which apparently doesn't support early termination?

solr/core/src/test/org/apache/solr/search/TestSearchPerf.java
cpoerschke: Hmm, I think the default is zero, no?

https://github.com/apache/solr/blob/releases/solr/9.6.1/solr/core/src/java/org/apache/solr/search/QueryCommand.java#L36

solr/core/src/java/org/apache/solr/handler/MoreLikeThisHandler.java
```diff
@@ -161,13 +162,16 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
           mltDocs = mlt.getMoreLikeThis(reader, start, rows, filters, flags);
         } else if (q != null) {
           // Matching options
```
cpoerschke, line 164:
```suggestion
```

dsmiley:
(no CHANGES.txt; just deprecating low level stuff & refactoring stuff that shouldn't have an effect)

WDYT @cpoerschke

solr/core/src/test/org/apache/solr/search/TestSearchPerf.java
dsmiley: (embarrassed) right of course

solr/core/src/java/org/apache/solr/handler/component/QueryComponent.java
cpoerschke: Yes. Currently with grouping the response would (via the presence of 'false') suggest that early termination was used but did not terminate early. With this change the 'segment terminated early' flag will be not present in the response, reflecting that early termination was not used (despite the user having asked for it).

cpoerschke (approved):

dsmiley:
> Therefore it could be nice to "Other Changes" mention about the deprecation before it goes away in Solr 10 or so. The @deprecated annotation serves the same purpose but only slightly so i.e. they might only get noticed when users try to upgrade whereas a mention in the CHANGES.txt might trigger an earlier realisation of "hey, we need to do some prep work or keep this in mind when upgrading our custom query component in future" etc.

I think deprecation is enough of an aid to users along with git commit message too so users see the explanation.  IMO the value to effort ratio small things to CHANGES.txt is low.  But since you ask for it, I'll do so.

dsmiley:
Planned Git commit message:

> QueryResult refactoring so that it's only returned from SolrIndexSearcher instead of
  being provided to it.  Deprecated APIs in 9x; should be removed later.
  Deprecated old methods in SolrIndexSearcher that are less clear than using QueryCommand.
  Deprecated related old methods in QueryPluginUtils; some should move to DebugComponent

dsmiley:
oh the irony; a CHANGES.txt conflict; sigh

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -716,6 +716,12 @@ public <K, V> boolean regenerateItem(
     }
   }
 
+  /** Primary entrypoint for searching, using a {@link QueryCommand}. */
+  public QueryResult search(QueryCommand cmd) throws IOException {
+    return search(new QueryResult(), cmd);
+  }
+
+  @Deprecated
```
cpoerschke, line 724: #2662 to remove use of this API variant

PR #2532, opened 2024-06-24
SOLR-17302: Convert remaining filestore APIs to JAX-RS

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-17302

# Description

PR #2470 converted a few file store APIs (those available under the `/api/cluster` path) to JAX-RS.  But the PR didn't touch several other APIs, which still needed JAX-RS conversion, including:

- Fetch filestore file
- Fetch filestore metadata
- Delete "local" filestore entry (internal API)

# Solution

This PR migrates the remaining filestore APIs to JAX-RS.  This migration is largely done "as-is", avoiding any cosmetic changes to the APIs.  IMO cosmetic improvements are much needed here (see SOLR-17302 for more details), but I've skipped those here because despite being "experimental" v2 APIs, these APIs are the only way to access Solr's filestore, so we might need to be a little more careful about changes.  I've started a discussion around this in SOLR-17302 for a hopefully wider audience to chime in.

One exception to this is the internal "local-delete" API, which is used internally by Solr.  It's not documented, nor does it have any use from an end-user perspective, so changing this API seemed a little more clear-cut.  This functionality can now be triggered via a `localDelete` parameter on the existing filestore delete API (e.g. `DELETE /cluster/files/somePath.txt?localDelete=true`)

We utilize "RawResponseWriter" for serializing the raw file data when fetching filestore files, to match the approach taken by the existing ZooKeeperReadAPI code.
  
# Tests

Existing filestore tests continue to pass, as this is mostly a refactor.  Additional unit tests in V2ApiUtils and elsewhere.

# Checklist


solr/api/src/java/org/apache/solr/client/api/endpoint/NodeFileStoreApis.java
```diff
@@ -0,0 +1,55 @@
...
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.PathParam;
+import jakarta.ws.rs.QueryParam;
+import org.apache.solr.client.api.model.MergeIndexesRequestBody;
+import org.apache.solr.client.api.model.SolrJerseyResponse;
+
+/**
+ * V2 APIs for fetching filestore files, syncing them across nodes, or fetching related metadata.
+ *
+ * @see MergeIndexesRequestBody
```
epugh, line 31: Does this do the magic linking?   the @see?

solr/core/src/java/org/apache/solr/filestore/FileStore.java
```diff
@@ -109,7 +108,7 @@ enum FileType {
     METADATA
   }
 
-  interface FileDetails extends MapWriter {
```
epugh, line 112: I like this!

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
```diff
@@ -0,0 +1,211 @@
...
+      DistribFileStore fileStore,
+      SolrQueryRequest req,
+      SolrQueryResponse rsp) {
+    this.coreContainer = coreContainer;
+    this.req = req;
+    this.rsp = rsp;
+    this.fileStore = fileStore;
+  }
+
+  // TODO - this single "get" operation actually supports several different chunks of functionality:
```
epugh, line 70: I wish we had JIRA's that referenced the "Todo" work...

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
```diff
@@ -0,0 +1,211 @@
...
+                }
+                log.info("downloaded file: {}", pathCopy);
+              });
+      return response;
+    }
+
+    if (path == null) {
+      path = "";
+    }
+    FileStore.FileType typ = fileStore.getType(path, false);
```
epugh, line 108: "type"?

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
```diff
@@ -0,0 +1,211 @@
...
+        final var fileMetaResponse =
+            instantiateJerseyResponse(FileStoreDirectoryListingResponse.class);
+        fileMetaResponse.files =
+            Collections.singletonMap(path, l.isEmpty() ? null : convertToResponse(l.get(0)));
+        return fileMetaResponse;
+      }
+    } else { // User wants to get the "raw" file
+      // TODO Should we be trying to json-ify otherwise "raw" files in this way?  It seems like a
+      // pretty sketchy idea, esp. for code with very little test coverage.  Consider removing
+      final var pathCopy2 = path;
```
epugh, line 141: pathCopy2???

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
```diff
@@ -0,0 +1,211 @@
...
+          throw new SolrException(
+              SolrException.ErrorCode.SERVER_ERROR, "Error getting file from path " + path);
+        }
+      } else {
+        ModifiableSolrParams solrParams = new ModifiableSolrParams();
+        solrParams.add(CommonParams.WT, FILE_STREAM);
+        req.setParams(SolrParams.wrapDefaults(solrParams, req.getParams()));
+        rsp.add(
+            CONTENT,
+            (SolrCore.RawWriter)
```
epugh, line 170: wow!

solr/core/src/java/org/apache/solr/response/RawResponseWriter.java
```diff
@@ -117,6 +119,10 @@ public void write(OutputStream out, SolrQueryRequest request, SolrQueryResponse
       try (InputStream in = content.getStream()) {
         in.transferTo(out);
       }
+    } else if (obj != null && (obj instanceof SolrCore.RawWriter)) {
+      final var rawWriter = (SolrCore.RawWriter) obj;
+      rawWriter.write(out);
+      if (rawWriter instanceof Closeable) ((Closeable) rawWriter).close();
```
epugh, line 125: maybe a `{` and `}`?

epugh (commented):
Great work here moving things forward...    Out of curisoity, do the existing tests then cover this all with no change?  amazing?

gerlowskija:
Looks like `PackageToolTest` is still failing with this change, will need to figure that out prior to merging...

solr/api/src/java/org/apache/solr/client/api/endpoint/NodeFileStoreApis.java
gerlowskija: It's not doing anything "magic" for our code-generation, etc. - it's just something the Javadoc syntax supports.

And in this case, it's there by copy/paste error 😛 Thanks for pointing it out haha!

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
gerlowskija: It would be nice.  I think the lack of a JIRA reference in this case reflects my own uncertainty about how to proceed.

In theory SOLR-17302 is already sortof meant to cover this type of "cosmetic improvement" change. That's how many of our v2 tickets are setup - to cover both JAX-RS conversion and "cosmetic changes" all in one.

But in practice I'm not sure whether it's safe/responsible to modify the appearance of these particular v2 endpoints.  They're technically "experimental", but they're the only way to access the filestore - there's no v1 equivalent that we can recommend to folks.

I'm going to start a discussion on SOLR-17302 about this, which will hopefully draw some more eyes and get us some clarity on how to proceed.

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
gerlowskija: This is something that carried over from the existing code, and I'm leery of getting sucked into larger improvements to that code.  There's a lot that could be done there.
But I can change it since it's a small modification.

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
gerlowskija: This is something that carried over from the existing code, and I'm leery of getting sucked into larger improvements to that code. There's a lot that could be done there.
But I can change it since it's a small modification.

solr/core/src/java/org/apache/solr/response/RawResponseWriter.java
gerlowskija: This is something that carried over from existing code.

I've agreed to a few variable renames above in code that preexists this PR, but I'll probably skip this change.  IMO this code could use lots of reformatting/restructuring attention, but we need to draw the line somewhere.

gerlowskija:
The PackageToolTest failure actually surfaced an interesting little v2 API bug - the v2 authz code wasn't properly detecting when a request had been authenticated as the special PKI principal '$'.  This PR now fixes that bug as well, using code similar to what we use in the v1 case.  With this fixed I'm marking the PR as "ready for review"!

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
gerlowskija: Actually - small correction - apparently when I wrote up SOLR-17302 I worded it in such a way that it only covers migration to JAX-RS.

So there's no ticket (yet) for breaking up this API into more manageable chunks.  I'll file a separate ticket for this and reference it here (and in the code) when its ready.

epugh:
> The PackageToolTest failure actually surfaced an interesting little v2 API bug - the v2 authz code wasn't properly detecting when a request had been authenticated as the special PKI principal '$'. This PR now fixes that bug as well, using code similar to what we use in the v1 case. With this fixed I'm marking the PR as "ready for review"!

Gnarly little fix there!

epugh (approved):
Some great work here!

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
gerlowskija: FYI: https://issues.apache.org/jira/browse/SOLR-17351

PR #2535, opened 2024-06-26
SOLR-17349: SolrDocumentFetcher should always skip lazy field loading overhead if documentCache==null

magibney:
https://issues.apache.org/jira/browse/SOLR-17349


magibney:
Opened https://github.com/apache/solr/pull/2537 for the failing test (it's unrelated to this PR).

dsmiley (approved):
+1 Makes sense to me.

Should have a CHANGES.txt entry in Optimizations.

magibney:
[This CI test failure](https://github.com/apache/solr/actions/runs/9688093653/job/26733983471?pr=2535) looks definitely unrelated, should not block merging. (I suspect that because the default timerSupplier uses an ExponentiallyDecayingReservoir, the mean may be weighted slightly toward more recent values, hence the higher-than-expected reported mean).

PR #2536, opened 2024-06-26
SOLR-10255 Add support for docValues to solr.BinaryField

serba:
# Description

Add support for docValues to solr.BinaryField

# Solution

Lucene has support for BinaryDocValuesField since forever. This is pretty straightforward PR with exposing that Lucene field in Solr.

# Tests

Added additional asserts to test solr.BinaryField with `docValues="true"` in existing `TestBinaryField` test.

# Checklist


solr/core/src/java/org/apache/solr/schema/BinaryField.java
```diff
@@ -94,6 +99,10 @@ public IndexableField createField(SchemaField field, Object val) {
       log.trace("Ignoring unstored binary field: {}", field);
       return null;
     }
+    return new org.apache.lucene.document.StoredField(field.getName(), getBytesRef(val));
+  }
+
+  private static BytesRef getBytesRef(Object val) {
```
serba, line 105: Moved the code for converting `Object val` to `BytesRef` for code reuse (and readability)

solr/core/src/java/org/apache/solr/schema/BinaryField.java
```diff
@@ -112,7 +121,36 @@ public IndexableField createField(SchemaField field, Object val) {
       len = buf.length;
     }
 
-    return new org.apache.lucene.document.StoredField(field.getName(), buf, offset, len);
+    return new BytesRef(buf, offset, len);
+  }
+
+  @Override
+  public List<IndexableField> createFields(SchemaField field, Object val) {
```
serba, line 128: This is pretty much borrowed from `StrField` type that supports creating `stored` and/or `docValued` Lucene fields for a single Solr field.

solr/core/src/java/org/apache/solr/schema/BinaryField.java
```diff
@@ -112,7 +121,36 @@ public IndexableField createField(SchemaField field, Object val) {
...
+    IndexableField fval = createField(field, val);
+
+    if (field.hasDocValues()) {
+      IndexableField docval;
+      if (field.multiValued()) {
+        docval = new SortedSetDocValuesField(field.getName(), getBytesRef(val));
+      } else {
+        docval = new BinaryDocValuesField(field.getName(), getBytesRef(val));
```
serba, line 136: This is the crux of the changes - we want to use Lucene `BinaryDocValuesField` if user specified `docValues="true"` in schema.

solr/core/src/java/org/apache/solr/search/SolrDocumentFetcher.java
```diff
@@ -609,7 +609,7 @@ private Object decodeDVField(
       case BINARY:
         BinaryDocValues bdv = e.getBinaryDocValues(localId, leafReader, readerOrd);
         if (bdv != null) {
-          return BytesRef.deepCopyOf(bdv.binaryValue());
+          return BytesRef.deepCopyOf(bdv.binaryValue()).bytes;
```
serba, line 612: Previously this code was returning `BytesRef` objects that were serialized up the stack with a simple `.toString` into `BytesRef@<hash-code>` string. You can revert this change and see it yourself in a test failure.

Now we return a `byte[]` which is correctly supported by SolrJ client and serialized as base64.

Note that there's no any additional performance overhead as we were already doing `deepCopy` into a new byte array and .bytes simply return a reference to a byte array.

solr/core/src/test/org/apache/solr/schema/TestBinaryField.java
```diff
@@ -82,73 +82,78 @@ public void testSimple() throws Exception {
...
-            byte b = data[i];
-            assertEquals((byte) (i + 4), b);
-          }
-
-        } else if (id == 3) {
-          assertEquals(10, data.length);
-          for (int i = 0; i < data.length; i++) {
-            byte b = data[i];
-            assertEquals((byte) i, b);
+        for (String field : new String[] {"data", "data_dv"}) {
```
serba, line 109: Please review these code changes with "Hide Whitespaces" toggle on.

Screenshot 2024-06-25 at 11 11 10 PM

solr/core/src/java/org/apache/solr/schema/BinaryField.java
```diff
@@ -112,7 +125,31 @@ public IndexableField createField(SchemaField field, Object val) {
...
-    return new org.apache.lucene.document.StoredField(field.getName(), buf, offset, len);
+    return new BytesRef(buf, offset, len);
+  }
+
+  @Override
+  public List<IndexableField> createFields(SchemaField field, Object val) {
+    IndexableField fval = createField(field, val);
+
+    if (field.hasDocValues() && !field.multiValued()) {
+      IndexableField docval = new BinaryDocValuesField(field.getName(), getBytesRef(val));
```
serba, line 136: This is the crux of the changes - we want to use Lucene `BinaryDocValuesField` if user specified `docValues="true"` in the schema.

solr/core/src/java/org/apache/solr/schema/BinaryField.java
```diff
@@ -112,7 +125,31 @@ public IndexableField createField(SchemaField field, Object val) {
...
+      if (fval != null) {
+        List<IndexableField> fields = new ArrayList<>(2);
+        fields.add(fval);
+        fields.add(docval);
+        return fields;
+      }
+
+      fval = docval;
+    }
+    return Collections.singletonList(fval);
```
mkhludnev, line 148: May it be simplified as 
```
Stream.of(createField(field, val), 
(field.hasDocValues() && !field.multiValued()) ? new BinaryDocValuesField(field.getName(), getBytesRef(val)):null)
                 .filter(obj -> obj != null)
```
?

solr/core/src/test-files/solr/collection1/conf/bad-schema-unsupported-docValues.xml
mkhludnev, line 1: I'm not sure why to bother with random field. Shouldn't we just remove this file and test, since now we have binary DV?

mkhludnev (approved):

solr/core/src/java/org/apache/solr/schema/BinaryField.java
serba: Yeah, my understanding is that this code is getting executed every time the field of this type is being indexed.

I borrowed the code from `StrField` type that supports creating `stored` and/or `docValued` Lucene fields for a single Solr field and there were such optimizations to create an `ArrayList` only in case of when both `stored` and `docValues` options are enabled. 

I guess it makes sense to keep this code consistent across different field types until there's some appetite in rewriting this in `streams` or `List.of` style across the board.

solr/core/src/test-files/solr/collection1/conf/bad-schema-unsupported-docValues.xml
serba: I think this test is not testing binary doc values per say, but exception handling for **any** field type that does not support docValues. I was going through different field types to find the best candidate now and figured `RandomSortField` is the best one as it will never get DocValues support.

dsmiley (approved):
+1 nice straight-forward change!

PR #2538, opened 2024-06-27
SOLR-17340: add cache on top of BeanInfo instrospector

psalagnac:
https://issues.apache.org/jira/browse/SOLR-17340

# Description

Solr endpoint `/solr/admin/info/system` is slow because of time spent in Java beans introspection.

# Solution

Add a simple cache on top of bean info introspection. For each bean class, the introspection will be run only once for the JVM lifespan.

# Tests

Naive performance benchmark. See results in [Jira comment](https://issues.apache.org/jira/browse/SOLR-17340?focusedCommentId=17860407&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17860407)

# Checklist

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -152,7 +152,7 @@ Optimizations
 
 * SOLR-17349: SolrDocumentFetcher should always skip lazy field loading overhead if documentCache==null (Michael Gibney)
 
-* JIRA-17340: Add cache on top of system metrics BeanInfos to make call to /admin/info/system faster (Pierre Salagnac)
+* SOLR-17340: Add cache on top of system metrics BeanInfos to make calls to /admin/info/system faster (Pierre Salagnac)
 
 Bug Fixes
 ---------------------
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -152,6 +152,8 @@ Optimizations
 
 * SOLR-17349: SolrDocumentFetcher should always skip lazy field loading overhead if documentCache==null (Michael Gibney)
 
+* JIRA-17340: Add cache on top of system metrics BeanInfos to make call to /admin/info/system faster (Pierre Salagnac)
+
 Bug Fixes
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/util/stats/MetricUtils.java
```diff
@@ -827,7 +845,7 @@ public static <T extends PlatformManagedObject> void addMXBeanMetrics(
       try {
         final Class<? extends PlatformManagedObject> intf =
             Class.forName(clazz).asSubclass(PlatformManagedObject.class);
-        MetricUtils.addMXBeanMetrics(obj, intf, null, consumer);
+        MetricUtils.addMXBeanMetrics(obj, intf, prefix, consumer);
```
dsmiley, line 848: What's this about here; seems like metrics would be published in a different spot, right?

psalagnac: I forgot to comment on this when creating the pull request.

I think this was a typo when this code was initially added. The `prefix` parameter of the method is not passed through to the other method. In practice, all callers to `addMXBeanMetrics()` in Solr code set null for this parameter, so this will not change anything.

Another option could be to just remove the parameter, but I think it's better to keep it if we have later more categories for metrics.

solr/CHANGES.txt
```diff
@@ -152,6 +152,8 @@ Optimizations
 
 * SOLR-17349: SolrDocumentFetcher should always skip lazy field loading overhead if documentCache==null (Michael Gibney)
 
+* JIRA-17340: Add cache on top of system metrics BeanInfos to make call to /admin/info/system faster (Pierre Salagnac)
```
dsmiley, line 155:
```suggestion
* SOLR-17340: Add cache on top of system metrics BeanInfos to make call to /admin/info/system faster (Pierre Salagnac)
```

PR #2539, opened 2024-06-27
SOLR-17321: Fix TestSolrCoreSnapshots.testBackupRestore

iamsanjay:
TestSolrCoreSnapshots.testBackupRestore test started failing on Windows after merging #2501.

Thank you @HoustonPutman for reporting this. This one specifically failed due to the backslash in the windows location path that passed to the URI ctor. 

The solution for this issue is to encode all the passed params before passing it to the URI ctor.

# Checklist


iamsanjay:
Is there anyway to test this on Windows?

iamsanjay:
Replaced the `StringBuilder` with `URIBuilder` and also there is no need to explicitly encode the params as `URIBuilder` takes care of it. 

Is there any way to test `TestHdfsBackupRestoreCore`?

epugh (approved):
LGTM.   Is the URIBuilder something that we should be using in more places?

HoustonPutman:
> LGTM. Is the URIBuilder something that we should be using in more places?

I certainly think so

epugh:
> > LGTM. Is the URIBuilder something that we should be using in more places?
> 
> 
> 
> I certainly think so

How do we get these tasks actually tracked and moved forward?  Worth opening a jira?  We have a patten of "we should do" but then it's just in our heads!

dsmiley:
`URIBuilder` is in the Apache HttpClient dependency, using it goes in the opposite direction we're trying to go in!  Granted in tests, this isn't a priority, so this PR is acceptable.  But be mindful please.

> We have a patten of "we should do"

That could be quite a list; I'm not sure what tracking it accomplishes.  If something is important enough, create a JIRA.  Ideally when a "we should do" is identified, it can be done everywhere instead of just one spot.

PR #2540, opened 2024-06-27
SOLR-16824: BACKPORT Adopt Linux Command line tool pattern of -- for long option commands (branch_9x)

epugh:
https://issues.apache.org/jira/browse/SOLR-16824

# Description

Backport to branch_9x the code in https://github.com/apache/solr/commit/0df3e63ffbdcde0eb953876a3fb47994d334d8f2.   The changes are backwards compatible, however we need to remove the use of OPTION_CREDENTIALS from all the commands.

# Solution

# Tests

# Checklist


epugh:
Humm..  I expected Crave and all the script tests to run on github.....

HoustonPutman:
I think you used the wrong JIRA in the title and description. Could you add some content to the title in addition to the backport?

epugh:
> I think you used the wrong JIRA in the title and description. Could you add some content to the title in addition to the backport?

Argh, that is what I get for not copy-pasting the JIRA ticket number!

epugh:
Okay, this is starting to turn into a bear...    I'm working through dealing with the fact that in 9x most of our tools don't have basicauth, and that there are lots of bats tests on main that aren't on branch_9x...    We'll see....

janhoy:
Perhaps more of the actions should run for PRs against the stable branch, e.g. add stable branch to https://github.com/apache/solr/blob/main/.github/workflows/tests-via-crave.yml#L6

solr/CHANGES.txt
```diff
@@ -39,6 +39,10 @@ Improvements
 
 * SOLR-15591: Make using debugger in Solr easier by avoiding NPE in ExternalPaths.determineSourceHome.  (@charlygrappa via Eric Pugh)
 
+* SOLR-16824: Adopt Linux standard pattern of -- for long option commands, and make all commands "kebab" formatting.  I.e -zkHost is now -zk-host.  The old parameters
+  such as -zkHost continue to be supported in the 9.x line of Solr.  -u is now used to specify user credentials everywhere, this only impacts the bin/solr assert 
+  commands "same user" check which has -u as the short form of --same-user. (Eric Pugh, janhoy, Jason Gerlowski)
```
janhoy, line 44:
```suggestion
  commands "same user" check which has -u as the short form of --same-user. (Eric Pugh, Jan Høydahl, Jason Gerlowski)
```
```suggestion
  commands "same user" check which has -u as the short form of --same-user. (Eric Pugh, janhoy, Jason Gerlowski)
```

janhoy:
I merged the changes to github workflows and triggered a new build. The precommit workflow now runs, but I cannot see the other workflows kicking off...

janhoy:
I'll take a stab at some of the failures...

janhoy:
Pushed some commits that at least makes Solr compile, and Fixes HealthCheckToolTest.

I propose that we revert most .bats files to testing original (deprecated) options, so we know the old cli API works.

I can see some confusion with `/solr` url suffix.

janhoy:
> Okay, this is starting to turn into a bear... I'm working through dealing with the fact that in 9x most of our tools don't have basicauth, and that there are lots of bats tests on main that aren't on branch_9x... We'll see....

Wrt basicAuth support with `--credentials`, I just kept what 9x already has, i.e. no basic auth support for all the tools, so removed `OPTION_CREDENTIALS` from those. Kept it for `AuthTool` only, of course, with added support for `-credentials`.

solr/packaging/test/bats_helper.bash
```diff
@@ -59,7 +59,7 @@ save_home_on_failure() {
 }
 
 shutdown_all() {
-  solr stop -all >/dev/null 2>&1
+  solr stop --all >/dev/null 2>&1
```
janhoy, line 62:
```suggestion
  solr stop -all >/dev/null 2>&1
```

I think the bats tests should stay as they on this branch, with main focus on regression testing existing opts. Bonus if we add tests that exercise new long-opts as well, but imo not that critical.

epugh:
Your changes look great, I especially like the helper for deprecated and current option look up.  That type of logic is spread about in the work I did and I guess easy to goof!

janhoy (commented):
I see test failures related to the `/solr` suffix part of baseURL.

There are also some failures where `***Tool.java` is updated, but `bin/solr` complains about an option. Guess we can pick these one by one.

janhoy:
Wrt the `create` tool vs `create_collection` and `create_core`, I tried to keep all three, may still be a bit un-polished.

janhoy:
I believe I fixed the docker tests (`gradlew testDocker`), which failed at `create_core_exec` :) 

But when building locally with `gradlew dev`, and then running `bin/solr create_core -c foo`, I get a scary error
```java.lang.NoClassDefFoundError: org/apache/solr/logging/MDCSnapshot```

I'm not able to proceed with the hunt more for now, but with PR tests running it should be easier for any committer to pick a test failure and fix it!

File: solr/core/src/test/org/apache/solr/cli/HealthcheckToolTest.java [hunk 3]
```diff
@@ -61,7 +59,22 @@ public void testHealthcheckWithSolrUrlParameter() throws Exception {
...
+  @Test
+  public void testHealthcheckWithSolrRootUrlParameter() throws Exception {
+
+    Set<String> liveNodes = cluster.getSolrClient().getClusterState().getLiveNodes();
+    String firstLiveNode = liveNodes.iterator().next();
+    String solrUrl =
+        ZkStateReader.from(cluster.getSolrClient()).getBaseUrlForNodeName(firstLiveNode);
+
+    // NOCOMMIT: epugh, why was this added?
+    // solrUrl = solrUrl.substring(0, solrUrl.indexOf("/solr"));
```
janhoy, line 75: @epugh I had to remove this for it to work, perhaps related to the recent #2519 ?

janhoy: Sorry, probably not related as that one only touches `bats` tests. Jason reverted the change in https://github.com/apache/solr/commit/b326f971f95c93216919cbfc2bf25c5d0853aa90

janhoy:
Fixed the `test_auth.bats` test in commit https://github.com/apache/solr/pull/2540/commits/f187b3e6fd762a9ba047bf6828193b56e83f6f74.

Reason: A mix of missing back-compat options parsing and missing /solr suffix on baseUrl.

solr/bin/solr
```diff
@@ -1697,7 +1702,7 @@ if [[ "$SCRIPT_CMD" == "auth" ]]; then
         fi
       done
   fi
-  run_tool auth "${AUTH_PARAMS[@]}" -solrUrl "$SOLR_URL_SCHEME://$SOLR_TOOL_HOST:${AUTH_PORT:-8983}/solr" -authConfDir "$SOLR_HOME" $VERBOSE
+  run_tool auth "${AUTH_PARAMS[@]}" --solr-url "$SOLR_URL_SCHEME://$SOLR_TOOL_HOST:${AUTH_PORT:-8983}/solr" --auth-conf-dir "$SOLR_HOME" $VERBOSE
```
janhoy, line 1705: A general comment/question is that in main, most of the arg parsing is completely delegated to SolrCLI, while even in this PR we keep the `bin/solr` parsing and construction of args to pass to auth tool.

Question is whether we should do as in main. Well, main branch does not delegate `"auth"` tool parsing to java, but it does for many other tools. It would then look like
```bash
if [[ "$SCRIPT_CMD" == "auth" ]]; then
  run_tool auth "$@"
  exit $?
fi
```

Benefit would be that we don't need to do all the arg parsing changes to `bin/solr.cmd` too.

I did a test, but there seems to be quite some effort needed for this too. The AuthTool needs `--solr-include-file` to edit it, so bin/solr passes this arg today, but user is also supposed to be able to override this param. Thus for AuthTool to handle all parsing, we probably need to let bin/solr pass include-file path in a different param `--solr-include-file-default` so AuthTool could arbitrate between them?

So not sure what is most work - fix windows cmd script or make delegation work. The delegation path would be an effort that could be forward-ported to main...

solr/core/src/test/org/apache/solr/cli/HealthcheckToolTest.java
janhoy: I'm getting really confused about the `/solr` suffix. I add it some place to fix some tests, but then other tests break that assume the suffix is not supposed to be there...

janhoy:
Pushed several changes
* Finally fixed the `/solr` url suffix normalization issue I hope
* `CreateTool` now used for `create`, `create_core` and `create_collection`.
* `solr create -h` usage options printed by SolrCLI (with 120 c width instead of 74)
* Docker tests once again pass
* BATS tests for `auth` and `create` now pass.

Still many BATS tests fail due to missing cmdline definitions...

solr/core/src/test/org/apache/solr/cli/HealthcheckToolTest.java
janhoy: Finally nailed this!

solr/bin/solr
epugh: Really appreciate the thought that you are bringing to this effort.   I hope to see significantly more of the arg parsing logic in `bin/solr` and `bin/solr.cmd` move to Java code happen before 10x.  So maybe the delegation effort, while being more work, would help move that overall goal forward?

epugh:
`PackageManagerCLITest.testPackageManager` was just a case of `solrUrl` and `solr-url`..    The other one may be about dealing with the `/solr` in the path or not...

epugh:
I looked a bit more at `SolrCloudExampleTest.testLoadDocsIntoGettingStartedCollection` to see if I could figure it out...   I thought I had some ideas, but no luck..   I *believe* we are struggling to detect if we are in cloud mode or not, and maybe we are not crafting the right url..   I checked, and the context route being only on main may be part of the problem:

```
* SOLR-16911: Establish /solr as the only host context supported by Solr, removing legacy ability to change this.  Solr paths will only be either /solr or /api now.  (Eric Pugh)
```

solr/bin/solr
```diff
@@ -982,31 +982,10 @@ else
...
-              else
-                break # out-of-args, stop looping
-              fi
-          ;;
-      esac
-    done
-  fi
   get_status
-  exit
+  exit $?
```
janhoy, line 988: I like these very much! Makes me almost believe we can get `solr.cmd` right as well 😄

janhoy: Only complaint I have is that usage help print in SolrCLI is not very nice. So I did #2545 in main to Improve on it..

janhoy:
Just fixed the `test_delete_collection.bats` test (Linux only).

janhoy:
Getting there... All BATS tests now run, except these five:

```
not ok 32 solr help flag prints help in 1521ms
# (from function `assert_output' in file /home/runner/work/solr/solr/.gradle/node/packaging/node_modules/bats-assert/src/assert.bash, line 247,
#  in test file test/test_help.bats, line 26)
#   `assert_output --partial 'Usage: solr COMMAND OPTIONS'' failed

not ok 33 solr with no flags prints help in 133ms
# (in test file test/test_help.bats, line 30)
#   `run -1 solr' failed, expected exit code 1, got 0

not ok 37 status help flag prints help in 183ms
# (from function `assert_output' in file /home/runner/work/solr/solr/.gradle/node/packaging/node_modules/bats-assert/src/assert.bash, line 247,
#  in test file test/test_help.bats, line 55)
#   `assert_output --partial 'usage: status'' failed

not ok 38 healthcheck help flag prints help in 196ms
# (from function `assert_output' in file /home/runner/work/solr/solr/.gradle/node/packaging/node_modules/bats-assert/src/assert.bash, line 247,
#  in test file test/test_help.bats, line 63)
#   `assert_output --partial 'usage: healthcheck'' failed
#
# -- output does not contain substring --
# substring (1 lines):
#   usage: healthcheck
# output (14 lines):
#
#   ERROR: Unrecognized or misplaced argument: --help!

not ok 78 status does not expose cli parameters to end user in 278ms
# (from function `assert_output' in file /home/runner/work/solr/solr/.gradle/node/packaging/node_modules/bats-assert/src/assert.bash, line 247,
#  in test file test/test_status.bats, line 45)
#   `assert_output --partial "ERROR: Unrecognized or misplaced argument: -solr!"' failed
```

janhoy:
All BATS are green 🦇

Please do a review of my latest couple commits to see if you agree with the decisions.

janhoy:
> I looked a bit more at `SolrCloudExampleTest.testLoadDocsIntoGettingStartedCollection` to see if I could figure it out... I thought I had some ideas, but no luck.. I _believe_ we are struggling to detect if we are in cloud mode or not, and maybe we are not crafting the right url.. I checked, and the context route being only on main may be part of the problem:
> 
> ```
> * SOLR-16911: Establish /solr as the only host context supported by Solr, removing legacy ability to change this.  Solr paths will only be either /solr or /api now.  (Eric Pugh)
> ```

Yea, I gave it a shot as well, but the test (`AbstractFullDistribZkTestBase`) insists in using some dynamic host-context instead of `/solr`. I tried hardcoding context as in main, but to no avail. This is a tricky one.

janhoy:
Managed to fix `SolrCloudExampleTest` by adapt `SolrCLI.getSolrClient` to use sysProp `hostContext` from tests if exists. This made the test pass, but now the BATS test for PostTool started failing. Reason was that it used `SolrCli.getSolrClient(solrUrl)` to get its client for doing **commit**, with the full `http://localhost:8983/solr/mycollection` url as base URL. Which no longer works (should it?). So I modified PostTool to instead extract collection name from URL, get a clinent with baseURL and call `client.commit(myCollection)`.

janhoy:
`BUILD SUCCESSFUL in 5m 10s` (crave from laptop)

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
```diff
@@ -435,11 +503,34 @@ public static boolean exceptionIsAuthRelated(Exception exc) {
...
+   */
   public static SolrClient getSolrClient(String solrUrl) {
-    return new Http2SolrClient.Builder(solrUrl)
-        .withKeyStoreReloadInterval(-1, TimeUnit.SECONDS)
-        .withMaxConnectionsPerHost(32)
-        .build();
+    // SOLR-16824: Workaround for making SolrCloudExampleTest pass with custom host contexts
+    String hostContext = System.getProperty("hostContext", "/solr");
+
+    solrUrl = normalizeSolrUrl(solrUrl, true) + hostContext;
```
janhoy, line 521: This breaks some prior practice, when you could assume that you could pass in `http://localhost:8983/solr/myColl` and you'd be given a client for which a `client.commit()` would actually issue a commit for the collection in that URL. 

With my latest changes, the client would be initialized on `http://localhost:8983/solr`, and the collection part would be thrown away. I.e. the `PostTool` then needs to provide collection name as argument to `commit()`.

Since I cannot see any other Tool that breaks due to this, I think this change is fine.

On option would be to stick closer to what's in main, i.e. accept the full URL and "patch" the `hostContext` by replacing `/solr/` with `$hostContext` and let rest of URL be as is.

solr/bin/solr.cmd
```diff
@@ -246,11 +246,12 @@ set FIRST_ARG=%1
...
-IF "%1"=="/?" goto usage
+IF "%1"=="-help" goto run_solrcli
+IF "%1"=="-usage" goto run_solrcli
+IF "%1"=="-h" goto run_solrcli
+IF "%1"=="--help" goto run_solrcli
+IF "%1"=="-help" goto run_solrcli
+IF "%1"=="/?" goto run_solrcli
 IF "%1"=="status" goto get_status
 IF "%1"=="version" goto run_solrcli
 IF "%1"=="-v" goto run_solrcli
```
janhoy, line 257: Regarding Windows script, we either need a volunteer who can use som cmd-fu to bring it up to speed.

Or, perhaps we can prune it from 2100 lines to a fraction by using it as a test bench for delegation to `SolrCLI`.
* Remove most arg parsing and just pass raw args to `run_solrcli`.
* Remove all or most usage help and let SolrCLI handle it
* Remove most ENV translation to sysprop since `EnvUtils` will handle it.

Doing so will not be 100% back compat wrt what the script prints, but it would be a test bed for improving the Java side of help formatting and arg parsing, ensuring 100% equal CLI API between Linux and Windows.

I'd rather slim down solr.cmd to something we can manage and risk some back-compat issues than ship a defunct script we cannot maintain.

epugh: I think your proposal makes sense...   how can we capture this as a decision point?  Maybe open a JIRA that specifies this work to be done?   Basically, how do we make sure the idea doesn't get dropped if no one does it immediately and I am not back for three weeks ;-)

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
janhoy: Turns out that the `SolrCloudExampleTest` still fails with custom hostContext such as `/x/y`, since it will only try to strip `/solr` in normalize step, so in this case the url will be `http://localhost:8983/x/y/x/y`, i.e. double. Have a fix.

solr/bin/solr.cmd
janhoy: The other PR with better help formatting is an enabler..

Maybe add a handful new sub-tasks to your umbrella JIRA - not sure if one JIRA per area (migrate usage) or one JIRA per tool (healthcheck etc) makes the most sense. Maybe one per tool, and we do one PR as an example and invite dev-list to pick the others? 

Removing usesless ENV translation can have one JIRA and multiple PRs.

janhoy: There is already a JIRA SOLR-15831 umbrella for much of this

epugh:
@janhoy, just looking again at this one, thanks for all the efforts!!!!    This looks like it's ready to commit?  However before I click "Squash and merge", I wanted to confirm with you!!

janhoy:
It’s the best I could do with my available time. I have NOT done a thorough review of all options in all tools, just fixed the ones causing test failures. 

The windows script lacks tests so who knows what works or not? Perhaps we are able to do follow up work on that front?

If no one else are willing to do a more thorough manual review, I believe merging this as is, is better than not merging.

And very important: some of my commits may include fixes that should be forward ported to main. We should make a placeholder JIRA to research what that might be.

epugh:
Good thoughts here @janhoy...   I will leave this open for another day or so, and then look to commit it.  I will put out an appeal to the dev mailing list when 9.7 release process starts asking for additional testing of `bin/solr.cmd`...

janhoy:
I filed SOLR-17371 for forward-porting.
Also synced PR branch with branch_9x.

epugh:
This fails:

```
./gradlew :solr:core:test --tests "org.apache.solr.cloud.SolrCloudExampleTest.testLoadDocsIntoGettingStartedCollection" -Ptests.jvms=96 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=44FFA474CD768986 -Ptests.timeoutSuite=600000! -Ptests.file.encoding=UTF-8
```

however this passes:

```
./gradlew :solr:core:test --tests "org.apache.solr.cloud.SolrCloudExampleTest.testLoadDocsIntoGettingStartedCollection"
```

epugh:
okay, it fails on that `tests.seed` value.   How can I step through the test in intellij with that seed value?

epugh:
looks like logic to identify a nodename assumes one `_` and we have two `_` chars!

janhoy:
Found a string comparison bug. For `hostContext=/box/`, in some code that was supposed to strip the context from URL, but failed due to training slash confusion. Pushed a fix, will see how it jives..

epugh:
> Found a string comparison bug. For `hostContext=/box/`, in some code that was supposed to strip the context from URL, but failed due to training slash confusion. Pushed a fix, will see how it jives..

Test now passes locally for me!  I think we are ready to merge!

PR #2542, opened 2024-06-28
SOLR-17353: Fix for golang version 1.18

khandnb:
https://issues.apache.org/jira/browse/SOLR-17353

# Description

# Solution

# Tests

# Checklist

### DIFF
--- a/solr/docker/templates/Dockerfile.body.template
+++ b/solr/docker/templates/Dockerfile.body.template
@@ -72,6 +72,8 @@ RUN set -ex; \
 RUN set -ex; \
     apt-get update; \
     apt-get -y --no-install-recommends install acl lsof procps wget netcat gosu tini jattach; \
+    wget -O /usr/local/bin/gosu "https://github.com/tianon/gosu/releases/download/1.17/gosu-amd64"; \ 
+    chmod +x /usr/local/bin/gosu; \
     rm -rf /var/lib/apt/lists/*;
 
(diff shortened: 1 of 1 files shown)


solr/docker/templates/Dockerfile.body.template
```diff
@@ -72,6 +72,8 @@ RUN set -ex; \
 RUN set -ex; \
     apt-get update; \
     apt-get -y --no-install-recommends install acl lsof procps wget netcat gosu tini jattach; \
+    wget -O /usr/local/bin/gosu "https://github.com/tianon/gosu/releases/download/1.17/gosu-amd64"; \ 
+    chmod +x /usr/local/bin/gosu; \
```
janhoy, line 76: It is a known fact that our version of ubuntu offers a slightly older `gosu` binary based on an older golang. However, there are no known attack vectors known to us how an external attacker might abuse this.

The fix presented here is not acceptable, as the image must work both for amd and arm architectures.

A better solution is to wait until we bump our default Ubuntu version in the docker image. Which I suppose could happen soon?

janhoy (changes requested):
Please instead look into upgrading the ubuntu platform-version and indirectly fix it that way.

solr/docker/templates/Dockerfile.body.template
HoustonPutman: We should definitely bump it for 10x

PR #2543, opened 2024-06-28
Run more workflows for branch_9x

janhoy:
Spinoff from #2540 

We should run all tests on PRs against `branch_9x`, to catch errors in larger backport efforts.

This should work fine at least until we bump main to some newer java version or make the branches incompatible, but then we can introduce separate workflows for stable and main branches..


dsmiley (approved):

janhoy:
@dsmiley The precommit ran on branch_9x now but not the other tests, any idea why?

PR #2545, opened 2024-07-01
SOLR-17357: Improve SolrCLI tool --help printout

janhoy:
https://issues.apache.org/jira/browse/SOLR-17357

Each tool now has a way to provide optional header, footer and usage string. See JIRA. Default will be to auto-generate usage string, and use some static text as header and footer. Example for assert tool:

```
$ bin/solr assert -h
usage: bin/solr assert [-c <url>] [-C <url>] [-e] [-m <message>] [-R] [-r] [-S <url>] [-s <url>] [--same-user
       <directory>] [-t <ms>] [-u <credentials>] [-x <directory>] [-X <directory>]

List of options:
 -c,--cloud <url>                 Asserts that Solr is running in cloud mode.  Also fails if Solr not running.  URL
                                  should be for root Solr path.
 -C,--not-cloud <url>             Asserts that Solr is not running in cloud mode.  Also fails if Solr not running.  URL
                                  should be for root Solr path.
 -e,--exitcode                    Return an exit code instead of printing error message on assert fail.
 -m,--message <message>           Exception message to be used in place of the default error message.
 -R,--not-root                    Asserts that we are NOT the root user.
 -r,--root                        Asserts that we are the root user.
 -S,--not-started <url>           Asserts that Solr is NOT running on a certain URL. Default timeout is 1000ms.
 -s,--started <url>               Asserts that Solr is running on a certain URL. Default timeout is 1000ms.
    --same-user <directory>       Asserts that we run as same user that owns <directory>.
 -t,--timeout <ms>                Timeout in ms for commands supporting a timeout.
 -u,--credentials <credentials>   Credentials in the format username:password. Example: --credentials solr:SolrRocks
 -x,--exists <directory>          Asserts that directory <directory> exists.
 -X,--not-exists <directory>      Asserts that directory <directory> does NOT exist.

Please see the Reference Guide for more tools documentation:
https://solr.apache.org/guide/solr/latest/deployment-guide/solr-control-script-reference.html
```

The reason for the many changed files is an API change to pass the `Tool` object to `SolrCLI.processCommandLineArgs()` instead of passing the name and `List<Option>` separately. Then SolrCLI can ask the tool for its options, name, header etc.


janhoy:
Other change is some utility method for `printToolHelp` that creates a `HelpFormatter` with width 120 instead of 74, and that handles the custom usage, header and footer.

The commons-cli lib generates usage string. Alternatively, we could have a static usage string:

```
Usage: bin/solr <tool> [options]
```

But for tools that have mandatory args or options, it would be nice to print them in usage, perhaps we could write our own custom usage generator based on some introspection of the tool options?

janhoy:
This is how the create tool will look like:

```
usage: bin/solr create -c <NAME> [-d <DIR>] [-n <NAME>] [-rf <#>] [-s <#>] [-u <credentials>] [-url <HOST>] [-z <HOST>]

List of options:
 -c,--name <NAME>                 Name of collection or core to create.
 -d,--conf-dir <DIR>              Configuration directory to copy when creating the new collection; default is _default.
 -n,--conf-name <NAME>            Configuration name; default is the collection name.
 -rf,--replication-factor <#>     Number of copies of each document across the collection (replicas per shard); default
                                  is 1.
 -s,--shards <#>                  Number of shards; default is 1.
 -u,--credentials <credentials>   Credentials in the format username:password. Example: --credentials solr:SolrRocks
 -url,--solr-url <HOST>           Base Solr URL, which can be used to determine the zk-host if that's not known;
                                  defaults to: http://localhost:8983.
 -z,--zk-host <HOST>              Zookeeper connection string; unnecessary if ZK_HOST is defined in solr.in.sh;
                                  otherwise, defaults to localhost:9983.

Please see the Reference Guide for more tools documentation:
https://solr.apache.org/guide/solr/latest/deployment-guide/solr-control-script-reference.html
```

epugh:
This is awesome.   I honestly had been willing to go backwards and lose the per tool customizations of help in the interest of simplicity...  can't wait to eview this pr.

solr/core/src/java/org/apache/solr/cli/ToolBase.java
```diff
@@ -66,5 +66,10 @@ public int runTool(CommandLine cli) throws Exception {
     return toolExitStatus;
   }
 
+  @Override
```
epugh, line 69: If you are using the default, then do you need to override just to return null?

epugh: Never mind, this is the base class!!

solr/core/src/test/org/apache/solr/cli/AuthToolTest.java
```diff
@@ -81,7 +81,7 @@ public void testEnableAuth() throws Exception {
   private int runTool(String[] args) throws Exception {
     Tool tool = findTool(args);
     assertTrue(tool instanceof AuthTool);
-    CommandLine cli = parseCmdLine(tool.getName(), args, tool.getOptions());
+    CommandLine cli = parseCmdLine(tool, args);
```
epugh, line 84: This new signature is much nicer!  I never liked the tool name call...

epugh (approved):
The refactoring looks good, nice to see that we removed lines of code wrapping!   I will look again but I do think see this used for the help in PackageTool or assert...?

janhoy:
> The refactoring looks good, nice to see that we removed lines of code wrapping! I will look again but I don think I see in this PR the enhancements to printing out help with custom header or footer etc ...???

Custom header/footer is in the `Tool` interface, with default methods. I did not add any custom header/footer to any tool yet, so the defaults are always used. The help print is done tin `SolrCLI#printToolHelp`.

PR #2546, opened 2024-07-01
add build badges to README.md

cpoerschke:
Inspired by @stefanvodita's https://github.com/apache/lucene/pull/13476 change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,8 @@
 Solr is the popular, blazing fast open source search platform for all your
 enterprise, e-commerce, and analytics needs, built on [Apache Lucene](https://lucene.apache.org/).
 
-[![Build Status](https://ci-builds.apache.org/job/Solr/job/Solr-Artifacts-main/badge/icon?subject=Solr)](https://ci-builds.apache.org/job/Solr/job/Solr-Artifacts-main/)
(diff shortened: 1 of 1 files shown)


janhoy (approved):

HoustonPutman:
Thank you! Curious why you chose the `artifact` build over the `check` build though.

cpoerschke:
> ... Curious why you chose the `artifact` build over the `check` build though.

Errm, not a choice as such, just did the same as what the Lucene badge does. I guess technically we could have badges for both or even more than two?

As an aside, the Gradle dashboard link doesn't yet work for me, could be local issue, not yet looked into. Does it work for anyone else?

dsmiley:
Check/Artifact:   If the intent is to offer a very latest download for Solr, then should be "Artifact".  I doubt this is useful in a world of Docker and Maven JAR artifact publication, however (ideally we conveniently link to both).  If the intent is to communicate the test status, then should link to a "check" build.  Or maybe we should link to the whole category in Jenkins.

cpoerschke:
> As an aside, the Gradle dashboard link doesn't yet work for me, could be local issue, not yet looked into. Does it work for anyone else?

Just removed it for now.

epugh:
Love to see this get added!

PR #2548, opened 2024-07-02
SOLR-17102: Replaced VersionBucket array with locks on-demand

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17102

DistributedUpdateProcessor some refactoring to balance locks clearly.
VersionInfo: moved locks out to UpdateLocks.
Refactor: OrderedExecutor uses generics.  Allows use of the ID now (a BytesRef) instead of a hash.


solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -652,35 +631,35 @@ private long doWaitForDependentUpdates(
...
+          condition.await(timeLeftInNanos, TimeUnit.NANOSECONDS);
+        } catch (InterruptedException e) {
+          Thread.currentThread().interrupt();
+          throw new RuntimeException(e);
         }
-        lookedUpVersion = vinfo.lookupVersion(cmd.getIndexedId());
-        lastFoundVersion = lookedUpVersion == null ? 0L : lookedUpVersion;
       }
-    } finally {
-      bucket.unlock();
```
dsmiley, line 683: again; moved this elsewhere for better lock balancing; allowed removing an indentation level

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -1089,94 +1057,89 @@ private boolean doVersionDelete(
       long signedVersionOnUpdate,
       boolean isReplayOrPeersync,
       boolean leaderLogic,
-      boolean forwardedFromCollection,
-      VersionBucket bucket)
+      boolean forwardedFromCollection)
```
dsmiley, line 1060: again, was able to balance locks and remove an indentation level.  Not clear in GitHub.

solr/core/src/java/org/apache/solr/update/UpdateLocks.java
```diff
@@ -0,0 +1,149 @@
...
+    final var startTimeNanos = System.nanoTime();
+
+    lockForUpdate();
+    try {
+
+      LockAndCondition lock;
+      while (true) {
+        lock = idToLock.computeIfAbsent(id, (k) -> new LockAndCondition());
+        synchronized (lock) {
+          if (lock.refCount < 0) {
```
bruno-roustant, line 65: In which conditions could it be < 0?

bruno-roustant (commented):
Surprisingly clean and concise refactoring!
Did you run the `./jmh.sh CloudIndexing` benchmark on it?

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -652,35 +631,35 @@ private long doWaitForDependentUpdates(
...
-        long timeLeftInNanos = waitTimeout.timeLeft(TimeUnit.NANOSECONDS);
-        if (timeLeftInNanos > 0) { // 0 means: wait forever until notified, but we don't want that.
-          bucket.awaitNanos(timeLeftInNanos);
+    while (Math.abs(lastFoundVersion) < cmd.prevVersion && !waitTimeout.hasTimedOut()) {
+      long timeLeftInNanos = waitTimeout.timeLeft(TimeUnit.NANOSECONDS);
+      if (timeLeftInNanos > 0) { // 0 means: wait forever until notified, but we don't want that.
+        try {
+          condition.await(timeLeftInNanos, TimeUnit.NANOSECONDS);
+        } catch (InterruptedException e) {
+          Thread.currentThread().interrupt();
```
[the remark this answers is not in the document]
dsmiley: Good catch.

solr/core/src/java/org/apache/solr/update/UpdateLocks.java
```diff
@@ -0,0 +1,149 @@
...
+      throws IOException {
+    // Acquire the lock
+    try {
+      if (docLockTimeoutMs == 0) {
+        lock.lock.lockInterruptibly();
+      } else {
+        long remainingNs =
+            TimeUnit.MILLISECONDS.toNanos(docLockTimeoutMs) - (System.nanoTime() - startTimeNanos);
+        boolean timedOut =
+            !lock.lock.tryLock(remainingNs - System.nanoTime(), TimeUnit.NANOSECONDS);
```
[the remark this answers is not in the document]
dsmiley: Doh!  Good eye!

solr/core/src/test/org/apache/solr/update/processor/DistributedUpdateProcessorTest.java
```diff
@@ -220,4 +185,45 @@ protected boolean tryLock(int lockTimeoutMs) {
...
     }
   }
+
+  private static DistributedUpdateProcessor newDurp(SolrQueryRequest req, long lockTimeoutMs) {
+    if (lockTimeoutMs <= 0) {
+      // default
+      return new DistributedUpdateProcessor(req, null, null);
+    }
+    // customize UpdateLocks with the provided timeout.  And simulate docs taking longer to index
+    final var sleepMs = 5000;
```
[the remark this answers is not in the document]
dsmiley: Yeah I was thinking that (this would be an improvement over the test as it ws written).  I'll do that; maybe still giving it more time in case the test is running slow.

solr/core/src/java/org/apache/solr/update/UpdateLocks.java
dsmiley: The comment below it says "// is being removed; try again".

Basically when this races with the lock getting removed by another thread that's done with it.  It was in the map but the next instant we see the refCount is too low.

dsmiley:
> Surprisingly clean and concise refactoring!

Thanks :-). There were a few rounds of attempts before getting at this.

> Did you run the ./jmh.sh CloudIndexing benchmark on it?

oh, great suggestion.  I'll do that as the next step; I remember you used that when you did the earlier VersionBucket change.

FYI I removed the documentation setting for the former timeout without replacement of the new.  I think it's even more unlikely that someone would set it.  When it was a "bucket" there was a reason and the docs elaborated on that reason.  It's reasonable to by default block indefinitely for a concurrent update _for the same doc_.  I'll get another opinion on this once I see the benchmarks look good.

solr/core/src/java/org/apache/solr/update/UpdateLocks.java
bruno-roustant: Shouldn't it be == 0 ?

dsmiley:
Before

```
Benchmark               (nodeCount)  (numReplicas)  (numShards)  (preGenerate)  (useSmallDocs)   Mode  Cnt     Score      Error  Units
CloudIndexing.indexDoc            4              1            4          50000            true  thrpt    4  9714.465 ± 2035.592  ops/s
CloudIndexing.indexDoc            4              1            4          50000           false  thrpt    4   961.674 ±   64.489  ops/s
CloudIndexing.indexDoc            4              3            4          50000            true  thrpt    4  3766.626 ± 2087.101  ops/s
CloudIndexing.indexDoc            4              3            4          50000           false  thrpt    4   270.518 ±  273.984  ops/s
```
After
```
Benchmark               (nodeCount)  (numReplicas)  (numShards)  (preGenerate)  (useSmallDocs)   Mode  Cnt     Score      Error  Units
CloudIndexing.indexDoc            4              1            4          50000            true  thrpt    4  8876.582 ± 6487.282  ops/s
CloudIndexing.indexDoc            4              1            4          50000           false  thrpt    4   939.311 ±  140.602  ops/s
CloudIndexing.indexDoc            4              3            4          50000            true  thrpt    4  3467.660 ±  625.605  ops/s
CloudIndexing.indexDoc            4              3            4          50000           false  thrpt    4   263.458 ±  306.638  ops/s
```
Perhaps an 8% regression.  Hmmm.

bruno-roustant:
Could you rerun the benchmark 'After'? The error range is high on the first line.
Does the benchmark outputs if there is more GC? I'm thinking of the new transient small lock objects. Not sure it would be visible though.

dsmiley (commented):
3 more commits show some iterations:
(1) "CHM, fix" fixes a bug due to complexities in having dual things happening automatically, the CHM computeIfAbsent and secondly the explicit synchronized -- things can happen in the gap between the two.
(2) CHM.compute -- an improvement to switch to using ConcurrentHashMap.compute which I didn't know can result in removing an entry if that's what you want (yes), which thus allowed removing of the synchronized sections -- overall very readable and an improvement.
(3) Switch to HPPC IntObjectHashMap & synchronize on it & also add a small pool of locks to reuse.  The key is the hashcode of the ID, thus only need to compute the hashcode once per doc processing instead of some number of more times.  Could also use some advanced methods on HPPC's map involving an "index".

It's nice that the iterations only involve changing one method.

dsmiley:
I simplified it a bit more when removing from the map.  And added more code comments / javadocs.

I benchmarked another time, this time with HTTP1.1 which didn't have the connection issue, thus could get a more complete run.  The findings are the same:

Baseline
```
Benchmark               (nodeCount)  (numReplicas)  (numShards)  (preGenerate)  (useSmallDocs)   Mode  Cnt      Score     Error  Units
CloudIndexing.indexDoc            1              1            1          50000            true  thrpt    8  28078.538 ± 841.866  ops/s
  MAX: 28546
```
Latest HPPC, simplify remove
```
Benchmark               (nodeCount)  (numReplicas)  (numShards)  (preGenerate)  (useSmallDocs)   Mode  Cnt      Score     Error  Units
CloudIndexing.indexDoc            1              1            1          50000            true  thrpt    8  28580.605 ± 857.046  ops/s
  MAX: 29310
```

bruno-roustant (approved):
Thanks David!

solr/core/src/java/org/apache/solr/update/UpdateLocks.java
```diff
@@ -0,0 +1,192 @@
...
+  private <R> R runWithLockInternal(
+      BytesRef id, IOFunction<Condition, R> function, LockAndCondition lock, long startTimeNanos)
+      throws IOException {
+    // Acquire the lock
+    try {
+      if (docLockTimeoutMs == 0) {
+        lock.lock.lockInterruptibly();
+      } else {
+        long remainingNs =
+            TimeUnit.MILLISECONDS.toNanos(docLockTimeoutMs) - (System.nanoTime() - startTimeNanos);
```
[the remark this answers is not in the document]
dsmiley: Shrug.  The conversion time is trivial.  I prefer to save the metric as it was provided, which commonly is Milliseconds, but sometimes isn't, admittedly.   Maybe I'll change it.

dsmiley:
* updated the upgrade notes to point out the legacy config settings.
* removed settings from any config
* CHANGES.txt:

> * SOLR-17102: The VersionBucket indexing lock mechanism was replaced with something just as fast yet
  that which consumes almost no memory, saving 1MB of memory per SolrCore.  (David Smiley)

Will merge tomorrow.

PR #2549, opened 2024-07-03
Stalebot: increase to 100 operations per run

janhoy:
Also re-word the stale-message text (with help from GPT).

Increasing operations budget will allow for more PRs to be tagged as stale per run - there is currently a backlog..

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -29,10 +29,9 @@ jobs:
           stale-pr-label: "stale"       # label to use when marking as stale
 
           stale-pr-message: >
-            This PR had no visible activity in the past 60 days, labeling it as stale.
-            Any new activity will remove the stale label. To attract more reviewers, please tag
-            someone or notify the dev@solr.apache.org mailing list.
+            This PR has had no activity for 60 days and is now labeled as stale. 
+            Any new activity or setting it as a draft will remove the stale label. 
+            To attract more reviewers, tag someone or notify the dev@solr.apache.org mailing list. 
             Thank you for your contribution!
 
-          # TODO: Increase budget after initial testing
-          operations-per-run: 30        # operations budget
+          operations-per-run: 100       # operations budget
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -29,10 +29,9 @@ jobs:
           stale-pr-label: "stale"       # label to use when marking as stale
 
           stale-pr-message: >
-            This PR had no visible activity in the past 60 days, labeling it as stale.
-            Any new activity will remove the stale label. To attract more reviewers, please tag
-            someone or notify the dev@solr.apache.org mailing list.
+            This PR has had no activity for 60 days and is now labeled as stale. 
+            Any new activity or setting it as a draft will remove the stale label.
```
cpoerschke, line 33: I like the mention of 'draft' as a choice. How about rewording to make it closer to the 'Convert to draft' UI element?
```suggestion
            Any new activity or converting it to draft will remove the stale label. 
```

.github/workflows/stale.yml
```diff
@@ -29,10 +29,9 @@ jobs:
           stale-pr-label: "stale"       # label to use when marking as stale
 
           stale-pr-message: >
-            This PR had no visible activity in the past 60 days, labeling it as stale.
-            Any new activity will remove the stale label. To attract more reviewers, please tag
-            someone or notify the dev@solr.apache.org mailing list.
+            This PR has had no activity for 60 days and is now labeled as stale. 
+            Any new activity or setting it as a draft will remove the stale label. 
+            To attract more reviewers, tag someone or notify the dev@solr.apache.org mailing list.
```
cpoerschke, line 34: Maybe we could broaden from '(some)one' to 'several' (which might be less or more daunting depending on point of view) and say a little more about how to identify people to tag?
```suggestion
            To attract more reviewers, tag people who might be familiar with the code area and/or notify the dev@solr.apache.org mailing list. 
```

PR #2566, opened 2024-07-17
SOLR-17357 Improve default --help formatting for CLI Tools (branch_9x)

janhoy:
This will be merged AFTER #2540, as it is backported on top of those changes. It will then replace some of the new code in that PR with what was committed to main.


solr/core/src/java/org/apache/solr/cli/SolrCLI.java
```diff
@@ -189,10 +185,6 @@ public class SolrCLI implements CLIO {
               "Credentials in the format username:password. Example: --credentials solr:SolrRocks")
           .build();
 
-  static {
```
epugh, line 192: nice to get rid of this odd code!

epugh (approved):
LGTM!

Would be nice to get this into 9.7, and then hopefully that will set us up for smoother backports from `main` in future releases...

epugh:
I want to click the Squash and Merge button, but don't totally understand the target branch, so will hold off!

.github/workflows/bin-solr-test.yml
```diff
@@ -9,6 +9,7 @@ on:
       - '.github/workflows/bin-solr-test.yml'
       - 'solr/bin/**'
       - 'solr/packaging/**'
+      - 'solr/core/src/java/org/apache/solr/cli/**'
```
janhoy, line 12: @epugh I added the change I proposed in https://github.com/apache/solr/pull/2573#pullrequestreview-2188208038 to trigger bats tests for all changes in cli code.

NOTE this is not in main, so perhaps add it in another PR?

janhoy: Never mind, reverted.

janhoy:
> I want to click the Squash and Merge button, but don't totally understand the target branch, so will hold off!

Moved it to correct branch, thik it is ready. Was expecting bats tests to run after merging #2575 but they did not. Guess we'll need to do some (but not all) bats test adjustments as well for the back port. Not all, since main uses more help output from SolrCLI than branch_9x...

epugh:
Fingers crossed a good build!   Changes look good.

PR #2571, opened 2024-07-18
SOLR-14985 Address perf issues to switch CloudSolrClient to HttpCluste…

aparnasuresh85:
…rStateProvider

https://issues.apache.org/jira/browse/SOLR-14985

# Description
Outlined below is the perf impact of using `CloudSolrClient` with `HttpClusterStateProvider` _as-is_ for `/select `request. Similar issues exist on the /update path

The test revealed 1 call being made to fetch entire cluster state and 5 other calls to fetch cluster status given a collection parameter, suggesting there is no caching mechanism in HttpClientClusterStateProvider

**1. Call to Fetch Entire Cluster State**
[
entire cluster state
](url)

**2. Several Calls to Fetch Cluster State for a Collection**

3 calls originating from CloudSolrClient.getDocCollection()
image2

**3. 2 calls originating from CloudSolrClient.resolveAliases()**

image3

Similar problems exist in the processing of a /update request

# Solution

**Issue 1 Resolution:** Modified the implementation of the default method `default <T> T getClusterProperty(String key, T defaultValue)` in `ClusterStateProvider` to call `getClusterProperty(key)` directly. This change avoids using `getClusterProperties().get(key)`, which in `BaseHttpClusterStateProvider` would fetch the entire cluster state.
**Issue 2 Resolution:** Addressed the lack of a caching mechanism in HTTP-based `ClusterStateProvider` (CSP). The `CloudSolrClient` only caches DocCollection instances where `ref.isLazilyLoaded()` returns `true`. However, for `CollectionRef` instances handled by HTTP-based CSP, the default method `CollectionRef.isLazilyLoaded()` returns `false`. This PR disables the check in `CloudSolrClient.getDocCollection()` to ensure caching works regardless of the `ref.isLazilyLoaded `return value.
**Issue 3 Resolution:** Incorporated code changes from prior [PR](https://github.com/apache/lucene-solr/pull/2086) for CSC to call `getDocCollection()` instead of `CSP.getState()` to lookup inputCollectionName, whereby it would be checked against collectionStateCache/Cluster state. 

# Tests

A new test, `testHttpCSpPerf()`, was added to initially identify performance issues. After implementing changes to address issues 1 and 2, this test confirms that these performance issues are resolved. The test also reveals that the remaining two calls to fetch the `CLUSTERSTATUS`, given a collection parameter, stem from issue 3, which has not been addressed in this PR.

# Checklist

- [] I have run `./gradlew check`.


solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -22,12 +22,7 @@
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.time.Instant;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
epugh, line 25: I think we normally list every import!!!

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -124,7 +119,10 @@ private ClusterState fetchClusterState(
       throws SolrServerException, IOException, NotACollectionException {
     ModifiableSolrParams params = new ModifiableSolrParams();
     if (collection != null) {
+      log.debug("Making a call to Solr to fetch cluster state for collection: ", collection);
```
epugh, line 122: Needs the if debug check...

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java
```diff
@@ -1213,10 +1193,6 @@ protected DocCollection getDocCollection(String collection, Integer expectedVers
       // no such collection exists
       return null;
     }
-    if (!ref.isLazilyLoaded()) {
```
epugh, line 1216: Any chance this 
Ethos is now no longer used anywhere??

solr/solrj/src/java/org/apache/solr/client/solrj/impl/ClusterStateProvider.java
```diff
@@ -31,7 +29,7 @@
 public interface ClusterStateProvider extends SolrCloseable {
 
   static ClusterStateProvider newZkClusterStateProvider(
-      Collection<String> zkHosts, String zkChroot, boolean canUseZkACLs) {
+          Collection<String> zkHosts, String zkChroot, boolean canUseZkACLs) {
```
epugh, line 32: Run the Gradle tidy task to fix code formatting

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
aparnasuresh85: IntellJ did this - I ran ./gradle tidy and noticed it complained too, fixed in the recent commit

solr/solrj/src/java/org/apache/solr/client/solrj/impl/ClusterStateProvider.java
aparnasuresh85: Done

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
aparnasuresh85: Done

dsmiley: Why do you say that Eric?  If it was needed, the precommit validation would tell her.  Part of the value of the PR validation is that we don't need to waste any time telling contributors something that the automation will inform them to do.  (similar for code formatting with tidy)  In this case, the if-check is *not* needed because this log call is simple -- no method call inside the arguments.  If the parameter there was foobar.callingAMethod().toString() or some-such, *then* the precommit check would say to add a condition.  Even though the verbosity of adding two lines for the isDebugEnabled check is minor, it annoys me -- I don't want to see it added if the validation doesn't ask us to.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -124,7 +124,14 @@ private ClusterState fetchClusterState(
       throws SolrServerException, IOException, NotACollectionException {
     ModifiableSolrParams params = new ModifiableSolrParams();
     if (collection != null) {
+      if (log.isDebugEnabled()) {
+        log.debug("Making a call to Solr to fetch cluster state for collection: {}", collection);
+      }
       params.set("collection", collection);
+    } else {
+      if (log.isDebugEnabled()) {
+        log.debug("Making a call to Solr to fetch entire cluster state");
```
dsmiley, line 133: From a performance perspective, I think requesting the entire state is unacceptable.  If we throw an UnsupportedOperationException here, what breaks?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java
dsmiley: I don't know what that means, but any way, I find this aspect of the PR/change most mysterious.  It appears this check is an optimization.  What effect does removing it have?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
dsmiley: I recommend using the following IntelliJ code style: https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml
"Tidy" only manipulates whitespace; it does not adjust imports, unlike that IntelliJ code style config.
Ideally this suggestion would be listed in `dev-docs/IDEs.adoc`

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java
dsmiley: > The CloudSolrClient only caches DocCollection instances where ref.isLazilyLoaded() returns true. **However, for CollectionRef instances handled by HTTP-based CSP, the default method CollectionRef.isLazilyLoaded() returns false.** This PR disables the check in CloudSolrClient.getDocCollection() to ensure caching works regardless of the ref.isLazilyLoaded return value.

(I added emphasis).  This is surprising.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -246,6 +249,45 @@ public void testAliasHandling() throws Exception {
         2, client.query(null, paramsWithMixedCollectionAndAlias).getResults().getNumFound());
   }
 
+  @Test
+  @LogLevel("org.apache.solr.client.solrj.impl.BaseHttpClusterStateProvider=DEBUG")
+  public void testHttpCSPPerf() throws Exception {
```
dsmiley, line 254: testHttpCspPerf is the capitalization approach to prefer

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java
```diff
@@ -789,7 +788,8 @@ protected NamedList<Object> requestWithRetryOnStaleState(
     if (!inputCollections.isEmpty()
         && !isAdmin
         && !isCollectionRequestOfV2) { // don't do _stateVer_ checking for admin, v2 api requests
-      Set<String> requestedCollectionNames = resolveAliases(inputCollections);
+      Set<String> requestedCollectionNames =
+          getClusterStateProvider().resolveAliases(inputCollections);
```
dsmiley, line 792: I like that you are moving more alias resolution logic to the CSP, but nonetheless CloudSolrClient is still explicitly involved.  I would prefer if the matter was invisible to CloudSolrClient (make no reference here to aliases; let it be a detail in CSP).  Admittedly that might be too much for this issue/PR.

dsmiley (commented):
I absolutely love seeing this test get added.

dsmiley:
I love the sequence diagrams!  What tool did you use to draw it?

RE aliases:  I was anticipating that this PR might regress the alias functionality.  But CloudHttp2SolrClientTest.testAliasHandling passes (as do all tests), and it uses the HTTP based CSP, so I suppose there's no down-side to any current user with this PR?  BTW, CloudHttp2SolrClientTest.getRandomClient is not doing what it should be doing.  Based on the commented code, I suspect the last person touching it (Dat) accidentally commented out the randomization.  I ran this test locally using both clients and it passes both ways.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudSolrClientCacheTest.java
```diff
@@ -156,6 +156,15 @@ public List<String> resolveAlias(String collection) {
...
         return def;
       }
+
+      @Override
+      public Set<String> resolveAliases(List<String> inputCollections) {
+        Set<String> res = new HashSet<>();
+        for (String c : inputCollections) {
+          res.add(c);
+        }
+        return res;
```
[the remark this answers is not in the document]
aparnasuresh85: Done the next commit reflects this

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -246,6 +249,45 @@ public void testAliasHandling() throws Exception {
...
+    SolrInputDocument doc = new SolrInputDocument("id", "1", "title_s", "my doc");
+    httpBasedCloudSolrClient.add(COLLECTION, doc);
+    httpBasedCloudSolrClient.commit(COLLECTION);
+
+    LogListener entireClusterStateLogs =
+        LogListener.debug(BaseHttpClusterStateProvider.class)
+            .substring("Making a call to Solr to fetch entire cluster state");
+    LogListener collectionClusterStateLogs =
+        LogListener.debug(BaseHttpClusterStateProvider.class)
+            .substring("Making a call to Solr to fetch cluster state for collection");
```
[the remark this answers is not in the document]
aparnasuresh85: I agree - I modified the test

aparnasuresh85:
> I love the sequence diagrams! What tool did you use to draw it?
> 
Plant UML - It can be installed as a plugin in IntelliJ

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -262,12 +263,18 @@ public void testHttpCSPPerf() throws Exception {
     httpBasedCloudSolrClient.add(COLLECTION, doc);
     httpBasedCloudSolrClient.commit(COLLECTION);
 
+    Pattern patternWithCollection =
+        Pattern.compile(
+            "path=/admin/collections.*params=\\{[^}]*action=CLUSTERSTATUS[^}]*collection=[^&}]+[^}]*\\}");
+
+    Pattern patternWithoutCollection =
+        Pattern.compile(
+            "path=/admin/collections.*params=\\{[^}]*action=CLUSTERSTATUS(?![^}]*collection=)[^}]*\\}");
```
dsmiley, line 272: Okay.   Albeit, these regex's are obtuse.  It'd be nice if LogListener had a `predicate` method taking `Predicate<String>` then you'd have something more readable like contains this and doesn't contain that.  Ah well.  Feel free to ignore this comment.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java
dsmiley: In summary, by bailing early here, we never even populated the cache.  The cache may not be important for a the ZK CSP because that has its own cache via ZkStateReader holding and always updating the ClusterState.  But this isn't true of HTTP CSP and I don't think layering another cache there is the ideal approach either.  The "stateVer" protocol in place makes sense; embraces an eventually-consistent mindset instead of pretending everything everywhere knows all.

By removing this check we'll get a _cached_ DocCollection, even for the ZK CSP which has access to the latest state (albeit eventually consistent).  This is kind of sad, removing a benefit of the ZK CSP in that it's routing is extremely closely up to date.

Perhaps instead, put this back and modify HttpClusterStateProvider to return a ClusterState.CollectionRef subclass that returns "true" for isLazilyLoaded, and that which actually fetches the state in get() (thus lazily not eagerly)?

aparnasuresh85: > In summary, by bailing early here, we never even populated the cache. The cache may not be important for a the ZK CSP because that has its own cache via ZkStateReader holding and always updating the ClusterState. But this isn't true of HTTP CSP and I don't think layering another cache there is the ideal approach either. The "stateVer" protocol in place makes sense; embraces an eventually-consistent mindset instead of pretending everything everywhere knows all.
That is correct.

> By removing this check we'll get a _cached_ DocCollection, even for the ZK CSP which has access to the latest state (albeit eventually consistent). This is kind of sad, removing a benefit of the ZK CSP in that it's routing is extremely closely up to date.

I believe ZK CSP always has its DocCollections cached in CSC since its works with _Lazy_ refs. However, the `collectionStateCache` in CSC has a `TTL` of 60 seconds, I believe that is when this cache is updated to have the "latest" state (get() on the lazy ref is invoked).

> Perhaps instead, put this back and modify HttpClusterStateProvider to return a ClusterState.CollectionRef subclass that returns "true" for isLazilyLoaded, and that which actually fetches the state in get() (thus lazily not eagerly)?

I tried this approach initially, but found that this cannot be done without adding another layer of cache in Http CSP.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -246,6 +250,51 @@ public void testAliasHandling() throws Exception {
...
+
+    LogListener entireClusterStateLogs =
+        LogListener.info(HttpSolrCall.class).regex(patternWithoutCollection);
+    LogListener collectionClusterStateLogs =
+        LogListener.info(HttpSolrCall.class).regex(patternWithCollection);
+
```
dsmiley, line 278: We also want to ensure that _no other HTTP calls are sent_ (except /select).  That way if we missed something or if somebody adds something later, that we have this test to ensure that the requests are reasonable.

Similarly do for indexing.

Similarly do for creating the HttpClient itself... albeit maybe defer that to another JIRA because I think it's fetching the entire cluster status there, which isn't something we should tackle in this PR.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
aparnasuresh85: Fetching entire cluster state could be prevented from the server side - maybe as part of another PR/jira issue

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java
dsmiley: BTW, I uncovered that this lazy check was introduced in SOLR-6521 as part of a performance improvement relating to ensuring that concurrent cache lookups of an expired entry don't redundantly fetch it in parallel. 

> I believe ZK CSP always has its DocCollections cached in CSP since its works with Lazy refs

I confirmed this as well in a debugger.  It's because a ZkStateReader created *for* ZK CSP doesn't watch the state of any single collection (albeit it does watch the parent /collections path), whereas a ZkStateReader created in ZkController (i.e. on the server) does watch individual collections for local replicas.

Two questions upon us:  do we
* (A) remove !isLazilyLoaded short-circuit.  Basically all users use ZK CSP and this doesn't trigger for that.  We fixed the HTTP CSP performance interaction here so it can be removed.
* (B) keep the short-circuit, and make HTTP CSP return a CollectionRef that is lazy

I actually think BOTH :-)
Just (A) is tempting; it "works" and addresses the obvious perf regression.  However, note the comment "we MUST try to get a new version" and "this is a call to ZK" some lines below.  This code is written around the idea that the potentially expensive part is down there where we hold a lock (SOLR-6521).  Code archeology here was helpful to uncover the intent.  If we don't make it lazy, we miss the point and then SOLR-6521 accomplishes nothing for the HTTP CSP scenario.  We should go as far as asserting that the ref.isLazilyLoaded and document getCollectionRef that it _must_ return "lazy" references.

Thinking of that lock and SOLR-6521, we could disregard the lazy/non-lazy issue and move `ClusterState.CollectionRef ref = getCollectionRef(collection);` from before the lock to right down into the lock before calling `get()` on it.  Then it doesn't matter if it's lazy or not; the real lookup (talking to either ZK or HTTP Solr) will be in there, which is the intent of SOLR-6521.  I played with this locally and tests pass.  Lets do this?

aparnasuresh85: I agree - the lazy/non-lazy check becomes irrelevant if collectionRef lookup happens inside the lock.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
aparnasuresh85: I am relying on logs in HttpSolrCall and removed the logging in BaseHttpCSP.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
aparnasuresh85: I moved the monitoring higher up  - so now it expands the check end to end. 

To ensure no other calls are sent,  I will need to add a debug log to HttpSolrCall as one doesnt exist now. Only admin requests are logged.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -93,6 +94,22 @@ public class CloudHttp2SolrClientTest extends SolrCloudTestCase {
   private static CloudSolrClient httpBasedCloudSolrClient = null;
   private static CloudSolrClient zkBasedCloudSolrClient = null;
 
+  private static LogListener entireClusterStateLogs;
+  private static LogListener collectionClusterStateLogs;
+  private static LogListener commandLogs;
```
dsmiley, line 99: LogListener is Closeable; it's intended to be used with try-with-resources pattern.  By having a static field of them, you need to take care to close them.  I'm not sure if LogListener is used this way across our tests... not sure I like it.  

Any way by listening globally for the whole suite, isn't asserting it kind of difficult to reason about as different tests will do stuff?

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -246,6 +267,36 @@ public void testAliasHandling() throws Exception {
...
+              .query(COLLECTION, params("q", "*:*"))
+              .getResults()
+              .getNumFound());
+    }
+
+    // still left with a call to fetch entire cluster state - to be addressed in a separate PR
+    assertTrue(entireClusterStateLogs.getQueue().size() == 1);
+
+    // Originates from CSP.resolveAliases() - 2 calls per request: 1 add, 1 commit, and 3 queries.
+    // This issue is not being addressed in this PR.
```
dsmiley, line 296: Don't refer to "this PR" (or commit or similar) in code.  If it were to be merged, it'd be a weird thing to stumble upon, reading code.  Instead, if we aren't addressing a known problem, reference a JIRA and/or say "TODO ..."

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
aparnasuresh85: I can look to create a "separate" `httpBasedCloudSolrClient`, different than the one created in the Test class. This way the LogListener can be used within this test with the "try-with-reources" pattern.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java
```diff
@@ -1228,9 +1199,15 @@ protected DocCollection getDocCollection(String collection, Integer expectedVers
...
+      // Moving the lookup inside the lock makes the lazy/non-lazy check irrelevant
+      ClusterState.CollectionRef ref = getCollectionRef(collection);
+      if (ref == null) {
+        // no such collection exists
+        return null;
+      }
       // We are going to fetch a new version
       // we MUST try to get a new version
-      DocCollection fetchedCol = ref.get(); // this is a call to ZK
+      DocCollection fetchedCol = ref.get(); // this is a call to ZK if CSP instance is a Zk CSP
```
dsmiley, line 1210: The ZK-ness isn't the point; the point is that a remote call is going to be made (expensive)

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java
```diff
@@ -1228,9 +1199,15 @@ protected DocCollection getDocCollection(String collection, Integer expectedVers
       if (col != null) {
         if (expectedVersion <= col.getZNodeVersion() && !cacheEntry.shouldRetry()) return col;
       }
+      // Moving the lookup inside the lock makes the lazy/non-lazy check irrelevant
```
dsmiley, line 1202: BTW a comment like "moving the ..." is only pertinent as a code review comment but not as a comment in the code.  It's a verb pertaining to a change that a commit/PR makes, which generally makes a poor/weird comment (similar to what I said in another feedback here today).  I'm not sure I'd bother with this comment.   If we are concerned that someone might add a call to getCollectionRef above the lock, then we could add a comment up there to warn against making remote calls without holding the lock.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -246,6 +257,81 @@ public void testAliasHandling() throws Exception {
...
 
+  @Test
+  @LogLevel("org.apache.solr.servlet.HttpSolrCall=DEBUG")
+  public void testHttpCspPerf() throws Exception {
+
+    performTest(false, "HTTPCSPTEST", 1, 10, 15, 15);
+
+  }
+  @Test
+  @LogLevel("org.apache.solr.servlet.HttpSolrCall=DEBUG")
```
aparnasuresh85, line 268: Not a great design for the tests I admit - I wanted to reuse code.

I also added a test using Zk based CSP - to compare metrics against the http one

After making the updates - I notice http based CSC also making a call to Solr to LISTALIASES (made obvious by the updated test)

The log listening can be consolidated somewhat.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -268,35 +258,90 @@ public void testAliasHandling() throws Exception {
...
+    performTest(true, "ZKCSPTEST",
+            0,
+            0,
+            //1 create collection, 2 /admin/core
+            3,
+            15);
+
+  }
+
+  private CloudSolrClient createHttpCSPBasedCloudSolrClient() {
```
dsmiley, line 285: "Csp" capitalization.
Why put "based" in it... The corresponding ZK one does not have that.  I suggest naming them both in harmony like createHttpCspCloudSolrClient and createZkCspCloudSolrClient and which are both plenty long names as it is

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -268,35 +258,90 @@ public void testAliasHandling() throws Exception {
...
   public void testHttpCspPerf() throws Exception {
 
-    String COLLECTION = "TEST_COLLECTION";
-    CollectionAdminRequest.createCollection(COLLECTION, "conf", 2, 1)
-        .process(cluster.getSolrClient());
-    cluster.waitForActiveCollection(COLLECTION, 2, 2);
+    performTest(false, "HTTPCSPTEST",
+            1,
+            10,
+            //1 create collection, 2 /admin/core , 1 LISTALIASES, 1 CLUSTERSTATUS, 10 CLUSTERSTATUS with coll param
```
dsmiley, line 267: I propose that collection creation be external to the logging observation.  Create it beforehand.  The reason why is that it'll generate a number of calls including internal distributed ones that are not what we're trying to observe (we're only trying to observe requests *from* our client, not from the Overseer or whatever Solr node.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -268,35 +258,90 @@ public void testAliasHandling() throws Exception {
...
+            15,
+            15);
 
-    SolrInputDocument doc = new SolrInputDocument("id", "1", "title_s", "my doc");
-    httpBasedCloudSolrClient.add(COLLECTION, doc);
-    httpBasedCloudSolrClient.commit(COLLECTION);
+  }
+  @Test
+  @LogLevel("org.apache.solr.servlet.HttpSolrCall=DEBUG")
+  public void testZkCspPerf() throws Exception {
```
dsmiley, line 274: Why test with ZkCsp using an HTTP logging approach?  ZkCsp does not make HTTP calls.  If we could monitor TCP ZK connections somehow, *that* would be interesting.  Such an approach would be doable with distributed tracing tech or SolrZkClient metrics (even easier) but I don't want to suggest expanding the scope/effort of this PR too much.

aparnasuresh85: Ok I guess it makes sense to remove this test

solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
```diff
@@ -325,6 +325,11 @@ protected void init() throws Exception {
         addCollectionParamIfNeeded(getCollectionsList());
 
         action = PROCESS;
+
+        //this request would be processed by Solr
+        if(invalidStates ==null){
```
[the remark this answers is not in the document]
aparnasuresh85: Removed listening for non-admin calls and related logging

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
aparnasuresh85: Removed the zk test

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
aparnasuresh85: Done

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -246,6 +257,52 @@ public void testAliasHandling() throws Exception {
...
+      SolrInputDocument doc = new SolrInputDocument("id", "1", "title_s", "my doc");
+      solrClient.add(collectionName, doc);
+      solrClient.commit(collectionName);
+      for (int i = 0; i < 3; i++) {
+        assertEquals(1, solrClient.query(collectionName, params("q", "*:*")).getResults().getNumFound());
+      }
+
+      assertLogCount(adminRequestLogs, 12);
+      assertLogCount(collectionClusterStateLogs, 10);
+      assertLogCount(entireClusterStateLogs, 1);
```
dsmiley, line 282: Could you add some comments explaining why these numbers are what they are?
I know the entireClusterStateLogs is to fetch live nodes, which is a TODO that we can reference another JIRA on.
Why do we get collection cluster state 10 times?  Shouldn't it be once?
I think admin should be the sum of the others but there's one more, suggesting there's another request we didn't anticipate?

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -246,6 +257,52 @@ public void testAliasHandling() throws Exception {
...
+  @Test
+  @LogLevel("org.apache.solr.servlet.HttpSolrCall=DEBUG")
+  public void testHttpCspPerf() throws Exception {
+
+    String collectionName = "HTTPCSPTEST";
+    CollectionAdminRequest.createCollection(collectionName, "conf", 2, 1).process(cluster.getSolrClient());
+    cluster.waitForActiveCollection(collectionName, 2, 2);
+
+    CloudSolrClient solrClient = null;
+    try (LogListener entireClusterStateLogs = LogListener.info(HttpSolrCall.class).regex(PATTERN_WITHOUT_COLLECTION); LogListener collectionClusterStateLogs = LogListener.info(HttpSolrCall.class).regex(PATTERN_WITH_COLLECTION); LogListener adminRequestLogs = LogListener.info(HttpSolrCall.class).substring("[admin]")) {
```
dsmiley, line 269: try-with-resources opportunity here for this CloudSolrClient

dsmiley (commented):
Validations show you need tidy

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
aparnasuresh85: Updated the PR

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
aparnasuresh85: I had the explanation in a prior commit I believe and forgot to add it back upon a subsequent refactor.

Out of the 12 /admin requests asserted by the CloudHttp2SolrClientTest, 10 originate from the call to fetch the CLUSTERSTATUS for a given collection, 1 from a call to fetch the entire CLUSTERSTATUS, and there is also a call to LISTALIASES that gets invoked from resolveAliases()  via the call to resolveAlias()

aparnasuresh85:
> The previous PR to solve this [apache/lucene-solr#2086](https://github.com/apache/lucene-solr/pull/2086) has a test suggesting the alias matter only results in one request?

That is only true for `add()` - `resolveAliases()` is only invoked once in that code path. Whereas for `commit()`, `/select` etc `resolveAliases()` would be called twice, once in `CSC.retryRequestWithStaleState()` and then again in `CSC.sendRequest()`

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java
```diff
@@ -1149,7 +1149,7 @@ private Set<String> resolveAliases(List<String> inputCollections) {
     }
     LinkedHashSet<String> uniqueNames = new LinkedHashSet<>(); // consistent ordering
     for (String collectionName : inputCollections) {
-      if (getClusterStateProvider().getState(collectionName) == null) {
+      if (getDocCollection(collectionName, -1) == null) {
```
aparnasuresh85, line 1152: Incorporated this change from https://github.com/apache/lucene-solr/pull/2086.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -246,6 +260,58 @@ public void testAliasHandling() throws Exception {
...
+
+      // 1 call to fetch entire cluster state via BaseHttpCSP.fetchLiveNodes()
+      // 1 call to LISTALIASES via getAliases() from CSC.sendRequest() -> (CSC.directUpdate ->
+      // 1 call to fetch CLUSTERSTATUS for collection via getDocCollection() (first collection
+      // lookup)
+      assertLogCount(adminRequestLogs, 3);
+      // 1 call to fetch CLUSTERSTATUS for collection via getDocCollection() (first collection
+      // lookup)
+      assertLogCount(collectionClusterStateLogs, 1);
+      // 1 call to fetch entire cluster state from HttpCSP.fetchLiveNodes()
```
aparnasuresh85, line 294: This issue to be addressed by https://issues.apache.org/jira/browse/SOLR-17381

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -262,26 +265,34 @@ public void testAliasHandling() throws Exception {
...
       for (int i = 0; i < 3; i++) {
-        assertEquals(1, solrClient.query(collectionName, params("q", "*:*")).getResults().getNumFound());
+        assertEquals(
+            1, solrClient.query(collectionName, params("q", "*:*")).getResults().getNumFound());
       }
 
-      assertLogCount(adminRequestLogs, 12);
-      assertLogCount(collectionClusterStateLogs, 10);
+      // 1 call to fetch entire cluster state via BaseHttpCSP.fetchLiveNodes()
+      // 1 call to LISTALIASES via getAliases() from CSC.sendRequest() -> (CSC.directUpdate ->
```
dsmiley, line 287: Shouldn't directUpdate use resolveAliases() just like requestWithRetryOnStaleState does?

aparnasuresh85: Yes, updated. There was an additional call to `LISTALIASES` from `directUpdate()` for the `isRoutedAlias()` check. The updated PR includes a change to make that check only applicable for multi-collection aliases.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java
```diff
@@ -362,8 +362,8 @@ private NamedList<Object> directUpdate(AbstractUpdateRequest request, String col
 
     // Check to see if the collection is an alias. Updates to multi-collection aliases are ok as
     // long as they are routed aliases
-    List<String> aliasedCollections = getClusterStateProvider().resolveAlias(collection);
-    if (getClusterStateProvider().isRoutedAlias(collection) || aliasedCollections.size() == 1) {
+    List<String> aliasedCollections = new ArrayList<>(resolveAliases(Collections.singletonList(collection)));
```
aparnasuresh85, line 365: Incorporated from: https://github.com/apache/lucene-solr/pull/2086

dsmiley:
I reported one of those test failures to Houston as it's a fresh issue.  The other is a long-time flaky failure; shouldn't be affected by this change. 

I think this is ready!

CHANGES.txt: 9.8 not 9.7 because too risky for 9.7 feature freeze any day now.  Optimizations section:

> SOLR-14985: Solrj CloudSolrClient with Solr URLs had serious performance regressions (since the beginning?) in which its collection state cache was not being used, resulting in many extra requests to Solr for cluster information.  (Aparna Suresh, shalin, David Smiley)

I plan to add this entry & merge this in about a day.

luiseterc:
Will that be backported to 8.11? We recently started using the CloudSolrClient client in our Java applications and the performance went south. We are experiencing high latencies, millions of requests to /admin/collections endpoint and Zookeeper is queuing requests.

An upgrade to 9.4 won't help in this case since this "optimization" will be included for 9.8, right?

luiseterc:
Thanks @janhoy We'll do that

PR #2574, opened 2024-07-19
SOLR-17367 Unbreak the build caused by bad backport

janhoy:
PostTool did not compile after https://github.com/apache/solr/commit/a405b3b2245350c2dae7e6979f33827e4b979511 due to the backport not being tailored to branch_9x which does not care about testing with basic auth.

Patched PostTool so that the new test creates collection and does the query without auth credentials, also using `url` variable, which actually exists in the class, as opposed to `uri`.

### DIFF
--- a/solr/core/src/java/org/apache/solr/cli/PostTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/PostTool.java
@@ -929,8 +929,8 @@ public class PostTool extends ToolBase {
 
     if (params.length() > 0) {
       try {
-        uri = new URI(appendParam(uri.toString(), params));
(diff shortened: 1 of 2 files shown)


epugh:
Thanks for sorting this out.

dsmiley:
Non-compile is very preventable.  FWIW I've increasingly been a fan of our `dev-tools/scripts/cherrypick.sh` which would not have let this happen.  Please try it @epugh

janhoy:
Yep, the script deals with most trivial back-ports. A separate 9x PR (like this) is safest route for non-trivial backports.

PR #2576, opened 2024-07-19
SOLR-15689: PeerSync to consistently handle 'delete' highest versions

cpoerschke:
(whether or not missed updates are requested should not depend on whether or not the latest version happens to (not) be a 'delete')

https://issues.apache.org/jira/browse/SOLR-15689


cpoerschke:
https://github.com/apache/solr/pull/2576/commits/e188720ca690d95555b752ae159594f9531be6c8 is the actual (small) change here if #343 first adds test coverage to model the current behaviour.

PR #2577, opened 2024-07-20
SOLR-16824: Replace deprecated single-dash arguments

malliaridis:
https://issues.apache.org/jira/browse/SOLR-16824

# Description

This PR replaces additional deprecated single-dash arguments that were depricated in previous PRs.

# Checklist

# Additional Notes / TODOs:
 _Updated notes were moved to Jira (see [comment](https://issues.apache.org/jira/browse/SOLR-16824?focusedCommentId=17868970&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17868970))_
 
@epugh I went through the main branch first and looked up all the arguments I thought may need replacement.


epugh:
This is some great analysis...    Not sure the best way to go through this...   Maybe a topic for next week community meetup?   Or a working sesion?   I think I'll comment in the code, and lets see how that goes.   One thing to keep in mind is that we are keeping in branch_9x the deprecated options, however we will be able to get rid of them in 10x.

solr/bin/solr.cmd
```diff
@@ -292,7 +292,6 @@ IF "%FIRST_ARG%"=="-help" goto run_solrcli
 IF "%FIRST_ARG%"=="-usage" goto run_solrcli
 IF "%FIRST_ARG%"=="-h" goto run_solrcli
 IF "%FIRST_ARG%"=="--help" goto run_solrcli
-IF "%FIRST_ARG%"=="-help" goto run_solrcli
```
epugh, line 295: `-help` is a special case (and we shoild probably comment that it is so) in the idea that folks might do -h, --help, and -help equally when trying to get to the help!

solr/bin/solr.cmd
```diff
@@ -314,7 +313,7 @@ goto done
 @echo   -f            Start Solr in foreground; default starts Solr in the background
 @echo                   and sends stdout / stderr to solr-PORT-console.log
 @echo.
-@echo   -c or -cloud  Start Solr in SolrCloud mode; if -z not supplied and ZK_HOST not defined in
+@echo   -c or --cloud Start Solr in SolrCloud mode; if -z not supplied and ZK_HOST not defined in
```
epugh, line 316: correct!   `--cloud` is what the docs should say.  We need to preserve the use of `-cloud` in 9x however...

solr/bin/solr.cmd
```diff
@@ -1525,10 +1524,10 @@ IF "!ZK_OP!"=="" (
 set CONNECTION_PARAMS=""
 
 IF "!ZK_OP!"=="" (
-  set CONNECTION_PARAMS="-solrUrl !ZK_SOLR_URL!"
+  set CONNECTION_PARAMS="--solr-url !ZK_SOLR_URL!"
```
epugh, line 1527: good catch!   I suspect because we still support `-solrUrl` that is why this works...

solr/bin/solr.cmd
```diff
@@ -1525,10 +1524,10 @@ IF "!ZK_OP!"=="" (
...
 )
 ELSE (
-  set CONNECTION_PARAMS="-zkHost ZK_HOST!"
+  set CONNECTION_PARAMS="--zk-host ZK_HOST!"
```
epugh, line 1530: yep!  similar to above issue with `-solrUrl`

solr/core/src/test/org/apache/solr/cli/AuthToolTest.java
```diff
@@ -70,7 +70,7 @@ public void testEnableAuth() throws Exception {
       dir.toAbsolutePath().toString(),
       "--solr-include-file",
       solrIncludeFile.toAbsolutePath().toString(),
-      "-credentials",
```
epugh, line 73: so, @gerlowskija  and I had a conversation that we ideally should be randomizing some of these, like `-credentials` and `--credentials` to make sure things continue to work with deprecated and non deprecated options.   I took the short cut of just randomly picking between the two, however maybe we should introduce a method here to handle this?  Alternatively, maybe a comment on why `-credentials` just to remind the next person?

solr/packaging/test/test_adminconsole_urls.bats
```diff
@@ -25,7 +25,7 @@ teardown() {
   # save a snapshot of SOLR_HOME for failed tests
   save_home_on_failure
 
-  solr stop -all >/dev/null 2>&1
+  solr stop --all >/dev/null 2>&1
```
epugh, line 28: I honestly stayed away from touching `bin/solr start` and `bin/solr stop` since I suspect lots of folks have scripts that call those and otherwise it seemed complicated.   I am up for changing `-all` to `--all`, however, can we augment the bats tests to try both?  Just to verify that `-all` and `-all` both work?

epugh (changes requested):
Some great work here!   A couple of comments!

epugh:
There are updates in `main` to `test_post.bats` and `test_help.bats` that you probably want ;-)

solr/core/src/test/org/apache/solr/cli/AuthToolTest.java
epugh: Here was the ticket for tracking the randomization idea: https://issues.apache.org/jira/browse/SOLR-17344

epugh: after dinking around with the code, I realize that the `-` versus `--` is dealt with by commons-cli, and we don't need to test that logic!   I will look for a place to introduce testing of deprecated Option versus current Option...

solr/bin/solr.cmd
malliaridis: Line 295 is a duplication of 291, which is why I removed it. Should probably not have an effect?

solr/bin/solr.cmd
malliaridis: For deprecation reasons I assume this should be updated to the new format, correct?

solr/packaging/test/test_adminconsole_urls.bats
malliaridis: I think adding tests in addition that checks that both still work is a good idea.

As far as I understand, the function "teardown" is responsible for test cleanup, so I guess it can be updated without problems? Or is there a chance that the test will be executed with a version that does not support `--all`?

PR #2585, opened 2024-07-23
Delegate Collection Property Management to CollectionPropertiesZkStateReader

aparnasuresh85:
# Description

**Summary**
This PR introduces `CollectionPropertiesZkStateReader` to manage collection property methods, shifting these responsibilities from ZkStateReader.

**Purpose**
`ZkStateReader` is currently overloaded with tasks, including refreshing and maintaining state for live nodes, collections, cluster state, collection properties, replica properties, aliases, and more. By distributing these responsibilities into specialized classes, we can ensure that each class has a single, well-defined responsibility, thereby reducing complexity.

This PR aims to transfer collection property management to `CollectionPropertiesZkStateReader`.

- Separation of Concerns: Enhances modularity by isolating logic specific to collection properties.
- Improved Maintainability: Simplifies `ZkStateReader`, making the codebase easier to manage.
- Backward Compatibility: `ZkStateReader`'s API remains unchanged, ensuring backward compatibility.

# Solution

# Tests

Ensure all existing tests pass

# Checklist

### DIFF
--- a/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
+++ b/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
@@ -1,3 +1,19 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.apache.solr.common.cloud;
 
 import static java.util.Collections.emptyMap;
--- a/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
+++ b/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
@@ -1,0 +1,403 @@
+package org.apache.solr.common.cloud;
+
+import static java.util.Collections.emptyMap;
+
+import java.lang.invoke.MethodHandles;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.solr.common.SolrCloseable;
+import org.apache.solr.common.SolrException;
+import org.apache.solr.common.cloud.ZkStateReader.CollectionWatch;
+import org.apache.solr.common.util.ExecutorUtil;
+import org.apache.solr.common.util.ObjectReleaseTracker;
+import org.apache.solr.common.util.SolrNamedThreadFactory;
+import org.apache.solr.common.util.Utils;
+import org.apache.zookeeper.KeeperException;
+import org.apache.zookeeper.WatchedEvent;
+import org.apache.zookeeper.Watcher;
+import org.apache.zookeeper.data.Stat;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CollectionPropertiesZkStateReader implements SolrCloseable {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  private volatile boolean closed = false;
+
+  private final SolrZkClient zkClient;
+
+  /** Collection properties being actively watched */
+  private final ConcurrentHashMap<String, VersionedCollectionProps> watchedCollectionProps =
+      new ConcurrentHashMap<>();
+
+  /**
+   * Manages ZooKeeper watchers for each collection. These watchers monitor changes to the
+   * properties of the collection in ZooKeeper. When a change is detected in ZooKeeper, the watcher
+   * triggers an update, which then notifies the relevant "collectionPropsObserver".
+   */
+  private final ConcurrentHashMap<String, PropsWatcher> collectionPropsWatchers =
+      new ConcurrentHashMap<>();
+
+  /**
+   * Manages a list of observers (listeners) for each collection. These observers need to be
+   * notified when the properties of the collection change. When a collection's properties change,
+   * all registered observers for that collection are notified by a "collectionPropWatcher".
+   */
+  private ConcurrentHashMap<String, CollectionWatch<CollectionPropsWatcher>>
+      collectionPropsObservers = new ConcurrentHashMap<>();
+
+  /** Used to submit notifications to Collection Properties watchers in order */
+  private final ExecutorService collectionPropsNotifications =
+      ExecutorUtil.newMDCAwareSingleThreadExecutor(
+          new SolrNamedThreadFactory("collectionPropsNotifications"));
+
+  private final ExecutorService notifications =
+      ExecutorUtil.newMDCAwareCachedThreadPool("cachecleaner");
+
+  // only kept to identify if the cleaner has already been started.
+  private Future<?> collectionPropsCacheCleaner;
+
+  public CollectionPropertiesZkStateReader(SolrZkClient zkClient) {
+    this.zkClient = zkClient;
+    assert ObjectReleaseTracker.track(this);
+  }
+
+  /**
+   * Get and cache collection properties for a given collection. If the collection is watched, or
+   * still cached simply return it from the cache, otherwise fetch it directly from zookeeper and
+   * retain the value for at least cacheForMillis milliseconds. Cached properties are watched in
+   * zookeeper and updated automatically. This version of {@code getCollectionProperties} should be
+   * used when properties need to be consulted frequently in the absence of an active {@link
+   * CollectionPropsWatcher}.
+   *
+   * @param collection The collection for which properties are desired
+   * @param cacheForMillis The minimum number of milliseconds to maintain a cache for the specified
+   *     collection's properties. Setting a {@code CollectionPropsWatcher} will override this value
+   *     and retain the cache for the life of the watcher. A lack of changes in zookeeper may allow
+   *     the caching to remain for a greater duration up to the cycle time of {@code CacheCleaner}.
+   *     Passing zero for this value will explicitly remove the cached copy if and only if it is due
+   *     to expire and no watch exists. Any positive value will extend the expiration time if
+   *     required.
+   * @return a map representing the key/value properties for the collection.
+   */
+  public Map<String, String> getCollectionProperties(final String collection, long cacheForMillis) {
+    synchronized (watchedCollectionProps) { // synchronized on the specific collection
+      Watcher watcher = null;
+      if (cacheForMillis > 0) {
+        watcher =
+            collectionPropsWatchers.compute(
+                collection,
+                (c, w) ->
+                    w == null ? new PropsWatcher(c, cacheForMillis) : w.renew(cacheForMillis));
+      }
+      VersionedCollectionProps vprops = watchedCollectionProps.get(collection);
+      boolean haveUnexpiredProps = vprops != null && vprops.cacheUntilNs > System.nanoTime();
+      long untilNs =
+          System.nanoTime() + TimeUnit.NANOSECONDS.convert(cacheForMillis, TimeUnit.MILLISECONDS);
+      Map<String, String> properties;
+      if (haveUnexpiredProps) {
+        properties = vprops.props;
+        vprops.cacheUntilNs = Math.max(vprops.cacheUntilNs, untilNs);
+      } else {
+        try {
+          VersionedCollectionProps vcp = fetchCollectionProperties(collection, watcher);
+          properties = vcp.props;
+          if (cacheForMillis > 0) {
+            vcp.cacheUntilNs = untilNs;
+            watchedCollectionProps.put(collection, vcp);
+          } else {
+            // we're synchronized on watchedCollectionProps and we can only get here if we have
+            // found an expired vprops above, so it is safe to remove the cached value and let the
+            // GC free up some mem a bit sooner.
+            if (!collectionPropsObservers.containsKey(collection)) {
+              watchedCollectionProps.remove(collection);
+            }
+          }
+        } catch (Exception e) {
+          throw new SolrException(
+              SolrException.ErrorCode.SERVER_ERROR,
+              "Error reading collection properties",
+              SolrZkClient.checkInterrupted(e));
+        }
+      }
+      return properties;
+    }
+  }
+
+  @Override
+  public void close() {
+    this.closed = true;
+    notifications.shutdownNow();
+    ExecutorUtil.shutdownAndAwaitTermination(notifications);
+    ExecutorUtil.shutdownAndAwaitTermination(collectionPropsNotifications);
+
+    assert ObjectReleaseTracker.release(this);
+  }
+
+  @Override
+  public boolean isClosed() {
+    return closed;
+  }
+
+  private static class VersionedCollectionProps {
+    int zkVersion;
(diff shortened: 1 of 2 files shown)


dsmiley:
(not ready for review; needs to be re-done)

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
```diff
@@ -0,0 +1,403 @@
...
+import org.apache.solr.common.util.SolrNamedThreadFactory;
+import org.apache.solr.common.util.Utils;
+import org.apache.zookeeper.KeeperException;
+import org.apache.zookeeper.WatchedEvent;
+import org.apache.zookeeper.Watcher;
+import org.apache.zookeeper.data.Stat;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CollectionPropertiesZkStateReader implements SolrCloseable {
```
dsmiley, line 29: Always add at least one public javadoc comment to a new class.

I noticed you make this public.  Do we consider this class to be the preferred API endpoint instead of ZkStateReader for this functionality? If not, it shouldn't be public, and the detailed javadocs should go back to where they were.  If so, arguably the ZkStateReader methods that delegate should perhaps be deprecated and we update clients?  @gus-asf / @tflobbe WDYT?

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -909,8 +884,8 @@ public void close() {
             });
 
     ExecutorUtil.shutdownAndAwaitTermination(notifications);
-    ExecutorUtil.shutdownAndAwaitTermination(collectionPropsNotifications);
-    if (closeClient) {
+    collectionPropertiesZkStateReader.close();
+    if (closeClient && zkClient != null) {
```
dsmiley, line 888: why the zkClient null check?  It wasn't there before.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
```diff
@@ -0,0 +1,403 @@
...
+  }
+
+  @Override
+  public void close() {
+    this.closed = true;
+    notifications.shutdownNow();
+    ExecutorUtil.shutdownAndAwaitTermination(notifications);
+    ExecutorUtil.shutdownAndAwaitTermination(collectionPropsNotifications);
+
+    assert ObjectReleaseTracker.release(this);
```
dsmiley, line 140: I don't recommend bothering with this; it's only closely associated with ZkStateReader so the accidental risk is extremely low.

dsmiley:
ZkStateReader has been getting too large; I like this change!

@gus-asf @tflobbe , I'm requesting your feedback on this pure refactoring PR (I wish the title used the word "refactoring" to emphasize this) by my colleague.  The previous attempts were wrong but this time it's a clean-room change here of just moving code around.  It's not intended to have any observed effect.

I will merge this in 48 hours if I get no +1 but I may merge sooner with a +1.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
```diff
@@ -0,0 +1,403 @@
...
+        + '/'
+        + ZkStateReader.COLLECTION_PROPS_ZKNODE;
+  }
+
+  private VersionedCollectionProps fetchCollectionProperties(String collection, Watcher watcher)
+      throws KeeperException, InterruptedException {
+    final String znodePath = getCollectionPropsPath(collection);
+    // lazy init cache cleaner once we know someone is using collection properties.
+    if (collectionPropsCacheCleaner == null) {
+      synchronized (this) { // Double-checked locking
```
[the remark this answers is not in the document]
aparnasuresh85: indeed this "this" is not the same as it was. Will make a change

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
aparnasuresh85: Added a javadoc for the new class

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -1191,130 +1166,27 @@ private void loadClusterProperties() {
   }
 
   /**
-   * Get collection properties for a given collection. If the collection is watched, simply return
```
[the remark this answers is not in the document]
aparnasuresh85: Reduced to one-line comment

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
aparnasuresh85: Removed the check

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
aparnasuresh85: Removed

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
```diff
@@ -26,11 +25,17 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * Fetches and manages collection properties from a ZooKeeper ensemble using {@link ZkStateReader}
+ * for ZooKeeper client access. Ensures thread safety by synchronizing on the update lock provided
+ * by {@link ZkStateReader}
```
dsmiley, line 31: IMO this is a detail (that may change) that isn't warranted in class javadocs.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -868,7 +869,11 @@ public ClusterState getClusterState() {
   }
 
   public Object getUpdateLock() {
-    return this;
+    return updateLock;
```
dsmiley, line 872: Scope is becoming slightly more than a refactor; now getUpdateLock has a dedicated object.  Was that necessary to achieve the refactoring aims?  For example it creates new problems... note that `createClusterStateWatchersAndUpdate()` is marked synchronized, thus it synchronizes on "this" but now it's different than what getUpdateLock returns.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
```diff
@@ -0,0 +1,406 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Fetches and manages collection properties from a ZooKeeper ensemble using {@link ZkStateReader}
+ * for ZooKeeper client access. Ensures thread safety by synchronizing on the update lock provided
+ * by {@link ZkStateReader}
+ */
+public class CollectionPropertiesZkStateReader implements SolrCloseable {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  private volatile boolean closed = false;
```
dsmiley, line 35: It's not clear we need a new "closed" boolean independent from ZkStateReader.  Since you've added ZkStateReader as a field, it's easy to call ZkStateReader.isClosed.  I understand my suggestion tightly couples this class to ZkStateReader but... what's wrong with that?  That's the reality here; heck look at its name.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
```diff
@@ -0,0 +1,406 @@
...
+import org.apache.zookeeper.data.Stat;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Fetches and manages collection properties from a ZooKeeper ensemble using {@link ZkStateReader}
+ * for ZooKeeper client access. Ensures thread safety by synchronizing on the update lock provided
+ * by {@link ZkStateReader}
+ */
+public class CollectionPropertiesZkStateReader implements SolrCloseable {
```
dsmiley, line 33: Implementing Closeable (not SolrCloseable) is adequate since we don't need the isClosed distinction (I think).

dsmiley (changes requested):

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
aparnasuresh85: Feels awkward to rely on ZkStateReader.isClosed(), particularly in `notifyPropsWatchers`, which relies on `closed`

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
aparnasuresh85: You are right. Fixed this

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
```diff
@@ -0,0 +1,397 @@
...
+  private ConcurrentHashMap<String, CollectionWatch<CollectionPropsWatcher>>
+      collectionPropsObservers = new ConcurrentHashMap<>();
+
+  /** Used to submit notifications to Collection Properties watchers in order */
+  private final ExecutorService collectionPropsNotifications =
+      ExecutorUtil.newMDCAwareSingleThreadExecutor(
+          new SolrNamedThreadFactory("collectionPropsNotifications"));
+
+  private final ExecutorService notifications =
+      ExecutorUtil.newMDCAwareCachedThreadPool("cachecleaner");
```
dsmiley, line 62: I was looking at some of these fields carefully, and I observed that "notifications" was something you duplicated (like you duplicated "closed" too).  BTW next time you do a refactor, it's good practice to self code-review in GitHub to comment on anything non-obvious.  Why was this duplicated?  In ZkStateReader, this thread pool is named "watcher" and here you name it "cachecleaner".  I see it being used as more than just the cache cleaner.  In the spirit of refactoring, can we provide access to the notifications executor to CollectionPropertiesZkStateReader as a kind of general purpose watching executor?  This way no need to shut it down (manage its lifecycle) in this class.

aparnasuresh85: > I was looking at some of these fields carefully, and I observed that "notifications" was something you duplicated (like you duplicated "closed" too). 
BTW next time you do a refactor, it's good practice to self code-review in GitHub to comment on anything non-obvious. 
I agree
> Why was this duplicated? In ZkStateReader, this thread pool is named "watcher" and here you name it "cachecleaner". I see it being used as more than just the cache cleaner. 
It is only being used for cache cleaning purposes within CPZSR
Felt awkward exposing the threadpool from ZSR initially

>In the spirit of refactoring, can we provide access to the notifications executor to CollectionPropertiesZkStateReader as a kind of general purpose watching executor? This way no need to shut it down (manage its lifecycle) in this class.
Done.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
```diff
@@ -0,0 +1,397 @@
...
+   */
+  private final ConcurrentHashMap<String, PropsWatcher> collectionPropsWatchers =
+      new ConcurrentHashMap<>();
+
+  /**
+   * Manages a list of observers (listeners) for each collection. These observers need to be
+   * notified when the properties of the collection change. When a collection's properties change,
+   * all registered observers for that collection are notified by a "collectionPropWatcher".
+   */
+  private ConcurrentHashMap<String, CollectionWatch<CollectionPropsWatcher>>
```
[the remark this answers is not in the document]
aparnasuresh85: Done

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -207,6 +207,10 @@ public static ZkStateReader from(CloudSolrClient solrClient) {
     }
   }
 
+  protected ExecutorService getNotifications(){
```
dsmiley, line 210: It doesn't return notifications, it returns an Executor to deliver asynchronous notifications on.  I suggest the name getWatcherExecutor or getNotificationExecutor.  Maybe the "Watcher" name because the thread pool is named this, thus that's how this executor is visible.  Anyway, no big deal.

BTW I can see you are forgetting to run "tidy".

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
```diff
@@ -50,17 +50,14 @@ public class CollectionPropertiesZkStateReader implements Closeable {
...
-  private ConcurrentHashMap<String, CollectionWatch<CollectionPropsWatcher>>
+  private final ConcurrentHashMap<String, CollectionWatch<CollectionPropsWatcher>>
       collectionPropsObservers = new ConcurrentHashMap<>();
 
   /** Used to submit notifications to Collection Properties watchers in order */
   private final ExecutorService collectionPropsNotifications =
       ExecutorUtil.newMDCAwareSingleThreadExecutor(
           new SolrNamedThreadFactory("collectionPropsNotifications"));
 
-  private final ExecutorService notifications =
```
dsmiley: I realize I was mistaken that it's used for multiple things, so your choice of separating it to a "cache cleaner" one is reasonable.  I like the way you had it in retrospect... but don't care enough either way so do as you like.

dsmiley (commented):
Remember to look at the validations!  I just did and see that you are missing the right source header.

Since the remaining matters are trivial or mechanical (tidy & headers), I'm re-iterating that I plan to merge this soon; I'd like to tomorrow.  After thoroughly reviewing some details, I have confidence this change isn't going to disturb anything.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/CollectionPropertiesZkStateReader.java
aparnasuresh85: Reverted to previous commit

gus-asf:
The concept looks good not able to fully detail review the whole thing so I'll give it a +0. If it does successfully achieve a "pure" refactor, shrinking a 2500 line class is clearly good.

dsmiley (approved):
It's ready to merge.  Thanks for your input Gus; I looked carefully for bugs / regressions.

There was a question I raised on if this helper class should be very internal/private to ZkStateReader or if it's the new face of CollectionProperties on the backend (instead of ZkStateReader.  At present it's in-between/ambiguous; ZkStateReader's methods remain, are not deprecated, and delegate to this new public class.  I could expect someone writing new code to be unsure which to call.

dsmiley:
Because ZkStateReader has no getter for CollectionPropertiesZkStateReader, in fact callers must continue to use ZkStateReader.  There isn't anything necessarily wrong with that but then CollectionPropertiesZkStateReader should probably not be public and it's helpful long javadocs on the methods should return to ZkStateReader since code will continue to call those methods.  WDYT @gus-asf ?

gus-asf:
Ah interesting question. Reminds me of something I said years ago. We are quite tightly bound to ZK for our coordination needs. This might be an opportunity to introduce an interface that is supplier agnostic (CollectionPropertiesManager) with an implementation as above. Then apply that interface to ZkStateReader to ease transition, At a later date when more things have appropriate interfaces, it might be possible to have implementations back-ended by:

- External ZK
- Embedded ZK
- etcd
- nextNewHotnessCoordinationd

We would want to think about the way one acquires implementations of course. And maybe that bit is a later PR...

Perhaps some such interfaces could live in org.apache.solr.cluster.coordination in solrj if they have user-useful aspects?

But overall I guess the main answer to your actual question is that I think implementations should be hidden and if possible interfaces exposed.

dsmiley:
I really don't think we'll be switching coordination layer backends, but I don't think that matters much because concealing details of the backend is good any way.  The name of this new class (blahBlahZkStateReader) here suggests it's more of a helper to ZkStateReader to split up the extensive responsibilities.  Fine.  Can be changed later of course; nobody will be talking to it except ZkStateReader anyway.

PR #2586, opened 2024-07-23
[SOLR-17368] TestPrometheusResponseWriter redesign

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-17368

# Description

The `TestPrometheusResponseWriter` was failing 100% of Jenkins builds due to leaking metrics from other tests running on the same JVM. This PR will redesigns the `TestPrometheusResponseWriter` to only test the structure of the output from the PrometheusResponseWriter and not the output of every metric that exists in Solr.

# Solution

Removed the `solr-prometheus-output.txt` output entirely. Filter out comments from the output (i.e `#` lines) and make an assertion that all metric names must start with `solr_metrics_*` and values must be a valid Prometheus accepted value `floating point number, NaN, +Inf, or -Inf` per Prometheus exposition format documentation. 

Also created `testPrometheusDummyOutput` that registers a dummy metrics for each registry and asserts against it's expected value. Because these are non-existent dummy values, any leaking metrics from other metrics should not override these values.

# Tests

`testPrometheusStructureOutput` and `testPrometheusDummyOutput`

# Checklist


solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
```diff
@@ -34,57 +38,104 @@
 import org.junit.ClassRule;
 import org.junit.Test;
 
-/** Tests the {@link PrometheusResponseWriter} behavior */
-@LuceneTestCase.AwaitsFix(bugUrl = "https://issues.apache.org/jira/browse/SOLR-17368")
 public class TestPrometheusResponseWriter extends SolrTestCaseJ4 {
-
   @ClassRule public static SolrJettyTestRule solrClientTestRule = new SolrJettyTestRule();
   public static JettySolrRunner jetty;
```
dsmiley, line 43: I think this field serves no purpose; we can get a jetty if needed from the rule on the fly.  Static fields in tests can have a maintenance burden to ensure they get null'ed.

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
```diff
@@ -34,57 +38,104 @@
...
+  public static final List<String> VALID_PROMETHEUS_VALUES = Arrays.asList("NaN", "+Inf", "-Inf");
 
   @BeforeClass
   public static void beforeClass() throws Exception {
     solrClientTestRule.startSolr(LuceneTestCase.createTempDir());
     jetty = solrClientTestRule.getJetty();
     solrClientTestRule.newCollection().withConfigSet(ExternalPaths.DEFAULT_CONFIGSET).create();
     jetty.getCoreContainer().waitForLoadingCoresToFinish(30000);
```
dsmiley, line 51: can call solrClientTestRule.getCoreContainer() instead and same below; maybe put in a var "cc" (a popular abbreviation in Solr for this super-common class)

dsmiley (commented):
In some sense this is an improvement in less flakiness (yay) but it doesn't address the intent of why solr-prometheus-output.txt was being written in the first place.  Alas; it's a contentious topic and Hossman is opinionated so I accept the PR.

dsmiley:
>  any leaking metrics from other metrics should not override these values.

That's dodging an issue that should be investigated as it might interfere with another metrics based test I suppose.  Any way; progress not perfection.

mlbiscoc:
> > any leaking metrics from other metrics should not override these values.
> 
> That's dodging an issue that should be investigated as it might interfere with another metrics based test I suppose. Any way; progress not perfection.

If I get some time, I can take a look into the lifecycle issue then we can revisit `solr-prometheus-output.txt`

PR #2526, opened 2024-06-19
SOLR-17337: Show proper distributed stage id

tboeghk:
https://issues.apache.org/jira/browse/SOLR-17337

# Description

When showing track debug for custom stages in distributed requests, the stage id is not translated into a meaningful string. It is displayed as an empty string (see attachment). Furthermore, if you have multiple custom stages, the details for the most recent stage is displayed. The details for other stages is omitted.

# Solution

If a given stage has no translation in the `stage` map, we format it's name in the format `STAGE_<STAGE_ID>`. By doing so, no stage details are omitted, even if you have multiple custom stages.

# Checklist

Please review the following and check all that apply:

- [x] I have reviewed the guidelines for [How to Contribute](https://github.com/apache/solr/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.
- [x] I have created a Jira issue and added the issue ID to my pull request title.
- [ ] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [x] I have developed this patch against the `main` branch.
- [x] I have run `./gradlew check`.
- [ ] I have added tests for my changes.
- [ ] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

PR #2594, opened 2024-07-25
SOLR-17337: Show proper distributed stage id

cpoerschke:
builds on @tboeghk's #2526 and makes the new method protected to allow overriding

https://issues.apache.org/jira/browse/SOLR-17337


epugh (approved):
LGTM.  Thanks for picking up @tboeghk  work and moving it forward!

dsmiley (approved):
Nice.

A test just to show, essentially, that something is override able seems sad to me to be honest.  It's a fair amount of LOC just for that.  I've done it before too; I'm just being circumspect about it right now.  Like; couldn't a one-liner comment at the method be reasonably sufficient?  "// KEEP THIS so can be overridden easily" to document the intentionality

PR #2597, opened 2024-07-26
SOLR-17319: CompoundQueryComponent scribbles

cpoerschke:
https://issues.apache.org/jira/browse/SOLR-17319

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/component/CompoundQueryComponent.java
+++ b/solr/core/src/java/org/apache/solr/handler/component/CompoundQueryComponent.java
@@ -1,0 +1,81 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


solr/core/src/java/org/apache/solr/handler/component/CompoundSearchHandler.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package org.apache.solr.handler.component;
+
+import java.util.List;
+import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.response.SolrQueryResponse;
+
+public class CompoundSearchHandler extends SearchHandler {
+
+  protected ResponseBuilder newResponseBuilder(
```
cpoerschke, line 25: https://issues.apache.org/jira/browse/SOLR-13511 added support to allow customisation like this.

solr/core/src/java/org/apache/solr/handler/component/CompoundQueryComponent.java
```diff
@@ -0,0 +1,81 @@
...
+    if (rb instanceof CompoundResponseBuilder) {
+      CompoundResponseBuilder crb = (CompoundResponseBuilder) rb;
+
+      ResponseBuilder rb_1 = new ResponseBuilder(rb.req, new SolrQueryResponse(), rb.components);
+      rb_1.setQueryString(rb_1.req.getParams().get("rrf.q.1"));
+      crb.responseBuilders.add(rb_1);
+
+      ResponseBuilder rb_2 = new ResponseBuilder(rb.req, new SolrQueryResponse(), rb.components);
+      rb_2.setQueryString(rb_2.req.getParams().get("rrf.q.2"));
+      crb.responseBuilders.add(rb_2);
```
cpoerschke, line 35: `QueryComponent.prepare` already has support for using other than `CommonParams.Q` i.e. `q` parameter for the query: https://github.com/apache/solr/blob/releases/solr/9.6.1/solr/core/src/java/org/apache/solr/handler/component/QueryComponent.java#L174-L181

cpoerschke:
#3648 is a continuation of these scribbles here.

PR #2598, opened 2024-07-26
Default to MockDirectoryFactory in test configs

gerlowskija:
# Description

IDE's like IntelliJ and Eclipse have internal test-runners that can often run tests faster than using the project's gradle build.  But many tests cannot use these IDE-internal runners because they rely on values set by the gradle build (particularly system properties set in 'randomization.gradle').

The primary offender here is the directory factory setting, which is set as RAMDirectoryFactory in test configs, but is then  overridden by randomization.gradle to be 'MockDirectoryFactory' instead.   
 
# Solution

This commit pushes the randomization.gradle override into individual test-config files, so that tests will see the same "effective default" regardless of whether they're run by an IDE's runner or using the gradle build.

# Tests

N/A - existing tests continue to pass

# Checklist


gerlowskija:
See [this dev list thread](https://lists.apache.org/thread/hz6ovvj611c91wwk6y0123zoff4c2v0w) for more context

dsmiley:
I did a search for RAMDirectoryFactory in source code to look for other possible issues, and I see what may be a bigger challenge in `org.apache.solr.SolrTestCaseJ4#initCore()`. 

```
factoryProp = System.getProperty("solr.directoryFactory");
    if (factoryProp == null) {
      System.setProperty("solr.directoryFactory", "solr.RAMDirectoryFactory");
    }
```

Your change means that factoryProp is going to be null when previously it would not have been.  I suppose we can simply remove this?  I like that... but admittedly it means that tests that do *not* use the test files you refer to would previously have used MockDirectoryFactory but now will get RAMDirectoryFactory.  I suppose you'll just have to triage any failures with this in mind.

dsmiley:
Since the issue I refer to is "initCore()"; it's old tests that extend STCJ4, most new ones don't (or shouldn't) use initCore().

gerlowskija:
As I read things the `factoryProp == null` branch is dead code on main/branch_9x, since `randomization.gradle` sets that prop on all builds?

I think you're right that with the PR as it stands, any tests that don't use the configsets would now be getting RAMDirectoryFactory, when they didn't previously.  In that case maybe the thing to do here isn't to delete the conditional outright but to tweak it so that it sets the same default (i.e. MockDirectoryFactory) as the configs.

dsmiley:
Agreed Jason.

dsmiley:
>  In that case maybe the thing to do here isn't to delete the conditional outright but to tweak it so that it sets the same default (i.e. MockDirectoryFactory) as the configs

Actually, let's simply delete this; one less potential moving part / layer for where this might be set.  It _should_ be enough to ensure all our test configs refer to MockDirectoryFactory.

I suppose there's a case for RAMDirectoryFactory, which is performance (faster running tests; less disk thrashing).  Albeit some tests need persistence, obviously.  But I'm unsure how to balance that with MockDirectoryFactory.  The safe bet is certainly MockDirectoryFactory.

gerlowskija:
Actually it looks like removing the check in `SolrTestCaseJ4.initCore()` causes a number of tests to fail, mostly with SecurityManager-related exceptions.

Taking `SolrCoreTest.testCoreInitDeadlockMetrics` as an example, it uses `core/src/test-files/configsets/minimal/` to create a core.  The relevant directoryFactory config there is:

```
  <directoryFactory name="DirectoryFactory"
                    class="${solr.directoryFactory:solr.NRTCachingDirectoryFactory}"/>
```

As things stand on 'main'/'branch_9x' currently, the gradle randomization ensures that `solr.directoryFactory` is always set to be MockDF, so the "default" value in this snippet doesn't ever come into play.  But in the most recent iteration of this PR neither `randomization.gradle` nor `SolrTestCaseJ4.initCore()` set a default dirFactory, so the test will start using NRTCachingDirectoryFactory and fail as a result.

I'm gonna try to go through and update the dirFactory default to MockDF in 'minimal' and some of these other lingering configsets, in the knowledge that these solrconfig.xml defaults were all being overridden by the randomization.gradle logic anyways.  But if I run into any issues I might rollback and put the `SolrTestCaseJ4.initCore()` back in place...

dsmiley (commented):
Then when do we actually test NRTCachingDirectoryFactory then -- the typical / common real-world DirectoryFactory?  I'd hope our more integration level tests, would use NRTCachingDirectoryFactory.  IMO cloud-minimal should use NRTCachingDirectoryFactory @gerlowskija

dsmiley:
Nevermind; all is well.

I looked closer at MockDirectoryFactory, which calls LuceneTestCase.newDirectory which usually uses a memory based directory (ByteBuffersDirectory) and sometimes wraps with NRTCachingDirectory, albeit "rarely".  And we have lots of SolrCloud tests using NRTCachingDirectoryFactory so even the factory is getting tested.

dsmiley:
After chasing some wild gooses on 9x, I realized the cause was that this wasn't backported.  I'm gonna do that tonight.

PR #2599, opened 2024-07-26
SOLR-17381: Make CLUSTERSTATUS configurable

aparnasuresh85:
https://issues.apache.org/jira/browse/SOLR-17381

# Description

Modify the requests to retrieve only the necessary information, such as the cluster status for a specific collection, live nodes, or cluster properties. Ensure these changes maintain backward compatibility. Additionally, update the HTTP CSP to reflect these optimizations.

# Solution

Introduce additional params to fetch CLUSTERSTATUS. Also added a INCLUDE_ALL param which would return all info by default to maintain backwards compatibility.
Additionally, made changes in BaseHttpClusterStateProvider to request minimal info for every use case.

# Tests

No new tests added. Ensured all existing tests pass

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
@@ -34,9 +34,9 @@ import org.apache.solr.common.cloud.DocRouter;
 import org.apache.solr.common.cloud.PerReplicaStates;
 import org.apache.solr.common.cloud.Replica;
 import org.apache.solr.common.cloud.Slice;
-import org.apache.solr.common.cloud.ZkNodeProps;
 import org.apache.solr.common.cloud.ZkStateReader;
 import org.apache.solr.common.params.ShardParams;
+import org.apache.solr.common.params.SolrParams;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.common.util.SimpleOrderedMap;
 import org.apache.solr.common.util.Utils;
@@ -44,8 +44,12 @@ import org.apache.zookeeper.KeeperException;
 
 public class ClusterStatus {
   private final ZkStateReader zkStateReader;
-  private final ZkNodeProps message;
+  private SolrParams solrParams;
   private final String collection; // maybe null
+
+  public static final String INCLUDE_ALL = "includeAll";
+  public static final String LIVENODES_PROP = "liveNodes";
+  public static final String CLUSTER_PROP = "clusterProperties";
 
   /** Shard / collection health state. */
   public enum Health {
@@ -89,10 +93,10 @@ public class ClusterStatus {
     }
   }
 
-  public ClusterStatus(ZkStateReader zkStateReader, ZkNodeProps props) {
+  public ClusterStatus(ZkStateReader zkStateReader, SolrParams params) {
     this.zkStateReader = zkStateReader;
-    this.message = props;
-    collection = props.getStr(ZkStateReader.COLLECTION_PROP);
+    this.solrParams = params;
+    collection = params.get(ZkStateReader.COLLECTION_PROP);
   }
 
   public void getClusterStatus(NamedList<Object> results)
@@ -100,11 +104,11 @@ public class ClusterStatus {
     List<String> liveNodes = null;
     NamedList<Object> clusterStatus = new SimpleOrderedMap<>();
 
-    boolean includeAll = this.message.getBool(ZkStateReader.INCLUDE_ALL, true);
+    boolean includeAll = this.solrParams.getBool(INCLUDE_ALL, true);
     boolean withLiveNodes =
-        this.message.getBool(ZkStateReader.LIVENODES_PROP, includeAll) || (collection != null);
-    boolean withClusterProperties = this.message.getBool(ZkStateReader.CLUSTER_PROP, includeAll);
-    boolean withRoles = this.message.getBool(ZkStateReader.ROLES_PROP, includeAll);
+        this.solrParams.getBool(LIVENODES_PROP, includeAll) || (collection != null);
+    boolean withClusterProperties = this.solrParams.getBool(CLUSTER_PROP, includeAll);
+    boolean withRoles = this.solrParams.getBool(ZkStateReader.ROLES_PROP, includeAll);
     boolean withCollection = includeAll || (collection != null);
 
     if (withLiveNodes) {
@@ -128,18 +132,16 @@ public class ClusterStatus {
     clusterStatus.add("properties", clusterProps);
 
     // add the roles map
+    Map<?, ?> roles = Collections.emptyMap();
     if (withRoles) {
-      Map<?, ?> roles = null;
       if (zkStateReader.getZkClient().exists(ZkStateReader.ROLES, true)) {
         roles =
             (Map<?, ?>)
                 Utils.fromJSON(
                     zkStateReader.getZkClient().getData(ZkStateReader.ROLES, null, null, true));
       }
-      if (roles != null) {
-        clusterStatus.add("roles", roles);
-      }
     }
+    clusterStatus.add("roles", roles);
     results.add("cluster", clusterStatus);
   }
 
@@ -163,8 +165,8 @@ public class ClusterStatus {
 
     ClusterState clusterState = zkStateReader.getClusterState();
 
-    String routeKey = message.getStr(ShardParams._ROUTE_);
-    String shard = message.getStr(ZkStateReader.SHARD_ID_PROP);
+    String routeKey = this.solrParams.get(ShardParams._ROUTE_);
+    String shard = this.solrParams.get(ZkStateReader.SHARD_ID_PROP);
 
     Map<String, DocCollection> collectionsMap = null;
     if (collection == null) {
@@ -234,7 +236,7 @@ public class ClusterStatus {
       }
       String configName = clusterStateCollection.getConfigName();
       collectionStatus.put("configName", configName);
-      if (message.getBool("prs", false) && clusterStateCollection.isPerReplicaState()) {
+      if (this.solrParams.getBool("prs", false) && clusterStateCollection.isPerReplicaState()) {
         PerReplicaStates prs = clusterStateCollection.getPerReplicaStates();
         collectionStatus.put("PRS", prs);
       }
--- a/solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
@@ -122,10 +122,8 @@ public class ClusterStatus {
     }
 
     if (withClusterProperties) {
-      Map<String, Object> clusterProps = Collections.emptyMap();
-      // read cluster properties
-      clusterProps = zkStateReader.getClusterProperties();
-      if (clusterProps == null || clusterProps.isEmpty()) {
+      Map<String, Object> clusterProps = zkStateReader.getClusterProperties();
+      if (clusterProps == null) {
         clusterProps = Collections.emptyMap();
       }
       clusterStatus.add("properties", clusterProps);
--- a/solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
@@ -50,6 +50,7 @@ public class ClusterStatus {
   public static final String INCLUDE_ALL = "includeAll";
   public static final String LIVENODES_PROP = "liveNodes";
   public static final String CLUSTER_PROP = "clusterProperties";
+  public static final String ALIASES_PROP = "aliases";
 
   /** Shard / collection health state. */
   public enum Health {
@@ -108,6 +109,7 @@ public class ClusterStatus {
     boolean withClusterProperties = solrParams.getBool(CLUSTER_PROP, includeAll);
     boolean withRoles = solrParams.getBool(ZkStateReader.ROLES_PROP, includeAll);
     boolean withCollection = includeAll || (collection != null);
+    boolean withAliases = solrParams.getBool(ALIASES_PROP, includeAll);
 
     List<String> liveNodes = null;
     if (withLiveNodes || collection != null) {
@@ -116,9 +118,19 @@ public class ClusterStatus {
       // add live_nodes
       if (withLiveNodes) clusterStatus.add("live_nodes", liveNodes);
     }
+
+    Aliases aliases = null;
+    if (withCollection || withAliases) {
+      aliases = zkStateReader.getAliases();
+    }
+
     if (withCollection) {
       assert liveNodes != null;
-      fetchClusterStatusForCollOrAlias(clusterStatus, liveNodes);
+      fetchClusterStatusForCollOrAlias(clusterStatus, liveNodes, aliases);
+    }
+
+    if (withAliases) {
+      addAliasMap(aliases, clusterStatus);
     }
 
     if (withClusterProperties) {
@@ -145,10 +157,9 @@ public class ClusterStatus {
   }
 
   private void fetchClusterStatusForCollOrAlias(
-      NamedList<Object> clusterStatus, List<String> liveNodes) {
+      NamedList<Object> clusterStatus, List<String> liveNodes, Aliases aliases) {
 
     // read aliases
-    Aliases aliases = zkStateReader.getAliases();
     Map<String, List<String>> collectionVsAliases = new HashMap<>();
     Map<String, List<String>> aliasVsCollections = aliases.getCollectionAliasListMap();
     for (Map.Entry<String, List<String>> entry : aliasVsCollections.entrySet()) {
@@ -241,16 +252,18 @@ public class ClusterStatus {
       collectionProps.add(name, collectionStatus);
     }
 
+    // now we need to walk the collectionProps tree to cross-check replica state with live nodes
+    crossCheckReplicaStateWithLiveNodes(liveNodes, collectionProps);
+
+    clusterStatus.add("collections", collectionProps);
+  }
+
+  private void addAliasMap(Aliases aliases, NamedList<Object> clusterStatus) {
     // add the alias map too
     Map<String, String> collectionAliasMap = aliases.getCollectionAliasMap(); // comma delim
     if (!collectionAliasMap.isEmpty()) {
       clusterStatus.add("aliases", collectionAliasMap);
     }
-
-    // now we need to walk the collectionProps tree to cross-check replica state with live nodes
-    crossCheckReplicaStateWithLiveNodes(liveNodes, collectionProps);
-
-    clusterStatus.add("collections", collectionProps);
   }
 
   /**
--- a/solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
@@ -109,7 +109,7 @@ public class ClusterStatus {
     boolean withClusterProperties = solrParams.getBool(CLUSTER_PROP, includeAll);
     boolean withRoles = solrParams.getBool(ZkStateReader.ROLES_PROP, includeAll);
     boolean withCollection = includeAll || (collection != null);
-    boolean withAliases = solrParams.getBool(ALIASES_PROP, includeAll) || withCollection;
+    boolean withAliases = solrParams.getBool(ALIASES_PROP, includeAll);
 
     List<String> liveNodes = null;
     if (withLiveNodes || collection != null) {
--- a/solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
@@ -109,7 +109,7 @@ public class ClusterStatus {
     boolean withClusterProperties = solrParams.getBool(CLUSTER_PROP, includeAll);
     boolean withRoles = solrParams.getBool(ZkStateReader.ROLES_PROP, includeAll);
     boolean withCollection = includeAll || (collection != null);
-    boolean withAliases = solrParams.getBool(ALIASES_PROP, includeAll);
+    boolean withAliases = solrParams.getBool(ALIASES_PROP, includeAll) || withCollection;
 
     List<String> liveNodes = null;
     if (withLiveNodes || collection != null) {
--- a/solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
@@ -45,7 +45,11 @@ import org.apache.zookeeper.KeeperException;
 public class ClusterStatus {
   private final ZkStateReader zkStateReader;
   private final ZkNodeProps message;
-  private final String collection; // maybe null
+  private final String collectionParam; // maybe null
+  private final boolean liveNodesParam;
+  private final boolean clusterPropertiesParam;
+  private final boolean rolesParam;
+  private final boolean includeAll;
 
   /** Shard / collection health state. */
   public enum Health {
@@ -92,11 +96,57 @@ public class ClusterStatus {
   public ClusterStatus(ZkStateReader zkStateReader, ZkNodeProps props) {
     this.zkStateReader = zkStateReader;
     this.message = props;
-    collection = props.getStr(ZkStateReader.COLLECTION_PROP);
+    collectionParam = props.getStr(ZkStateReader.COLLECTION_PROP);
+    liveNodesParam = props.getBool(ZkStateReader.LIVENODES_PROP, false);
+    clusterPropertiesParam = props.getBool(ZkStateReader.CLUSTER_PROP, false);
+    rolesParam = props.getBool(ZkStateReader.ROLES_PROP, false);
+    includeAll = props.getBool(ZkStateReader.INCLUDE_ALL, true);
   }
 
   public void getClusterStatus(NamedList<Object> results)
       throws KeeperException, InterruptedException {
+
+    List<String> liveNodes = null;
+    NamedList<Object> clusterStatus = new SimpleOrderedMap<>();
+    if (includeAll || collectionParam != null || liveNodesParam) {
+      liveNodes =
+          zkStateReader.getZkClient().getChildren(ZkStateReader.LIVE_NODES_ZKNODE, null, true);
+      // add live_nodes
+      clusterStatus.add("live_nodes", liveNodes);
+    }
+
+    if (includeAll || collectionParam != null)
+      fetchClusterStatusForCollOrAlias(clusterStatus, liveNodes);
+
+    if (includeAll || clusterPropertiesParam) {
+      // read cluster properties
+      Map<String, Object> clusterProps = zkStateReader.getClusterProperties();
+      if (clusterProps != null && !clusterProps.isEmpty()) {
+        clusterStatus.add("properties", clusterProps);
+      }
+    }
+
+    // add the roles map
+    if (includeAll || rolesParam) {
+
+      Map<?, ?> roles = null;
+      if (zkStateReader.getZkClient().exists(ZkStateReader.ROLES, true)) {
+        roles =
+            (Map<?, ?>)
+                Utils.fromJSON(
+                    zkStateReader.getZkClient().getData(ZkStateReader.ROLES, null, null, true));
+      }
+      if (roles != null) {
+        clusterStatus.add("roles", roles);
+      }
+    }
+
+    results.add("cluster", clusterStatus);
+  }
+
+  private void fetchClusterStatusForCollOrAlias(
+      NamedList<Object> clusterStatus, List<String> liveNodes) {
+
     // read aliases
     Aliases aliases = zkStateReader.getAliases();
     Map<String, List<String>> collectionVsAliases = new HashMap<>();
@@ -105,19 +155,11 @@ public class ClusterStatus {
       String alias = entry.getKey();
       List<String> colls = entry.getValue();
       for (String coll : colls) {
-        if (collection == null || collection.equals(coll)) {
+        if (collectionParam == null || collectionParam.equals(coll)) {
           List<String> list = collectionVsAliases.computeIfAbsent(coll, k -> new ArrayList<>());
           list.add(alias);
         }
       }
-    }
-
-    Map<?, ?> roles = null;
-    if (zkStateReader.getZkClient().exists(ZkStateReader.ROLES, true)) {
-      roles =
-          (Map<?, ?>)
-              Utils.fromJSON(
-                  zkStateReader.getZkClient().getData(ZkStateReader.ROLES, null, null, true));
     }
 
     ClusterState clusterState = zkStateReader.getClusterState();
@@ -126,33 +168,36 @@ public class ClusterStatus {
     String shard = message.getStr(ZkStateReader.SHARD_ID_PROP);
 
     Map<String, DocCollection> collectionsMap = null;
-    if (collection == null) {
+    if (collectionParam == null) {
       collectionsMap = clusterState.getCollectionsMap();
     } else {
       collectionsMap =
-          Collections.singletonMap(collection, clusterState.getCollectionOrNull(collection));
+          Collections.singletonMap(
+              collectionParam, clusterState.getCollectionOrNull(collectionParam));
     }
 
-    boolean isAlias = aliasVsCollections.containsKey(collection);
-    boolean didNotFindCollection = collectionsMap.get(collection) == null;
+    boolean isAlias = aliasVsCollections.containsKey(collectionParam);
+    boolean didNotFindCollection = collectionsMap.get(collectionParam) == null;
 
     if (didNotFindCollection && isAlias) {
       // In this case this.collection is an alias name not a collection
       // get all collections and filter out collections not in the alias
+      // clusterState.getCollectionsMap() should be replaced with an inexpensive call
       collectionsMap =
           clusterState.getCollectionsMap().entrySet().stream()
-              .filter((entry) -> aliasVsCollections.get(collection).contains(entry.getKey()))
+              .filter((entry) -> aliasVsCollections.get(collectionParam).contains(entry.getKey()))
               .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
     }
 
     NamedList<Object> collectionProps = new SimpleOrderedMap<>();
 
+    // loop will run only once
     for (Map.Entry<String, DocCollection> entry : collectionsMap.entrySet()) {
       Map<String, Object> collectionStatus;
       String name = entry.getKey();
       DocCollection clusterStateCollection = entry.getValue();
       if (clusterStateCollection == null) {
-        if (collection != null) {
+        if (collectionParam != null) {
           SolrException solrException =
               new SolrException(
                   SolrException.ErrorCode.BAD_REQUEST, "Collection: " + name + " not found");
@@ -198,36 +243,16 @@ public class ClusterStatus {
       collectionProps.add(name, collectionStatus);
     }
 
-    List<String> liveNodes =
-        zkStateReader.getZkClient().getChildren(ZkStateReader.LIVE_NODES_ZKNODE, null, true);
-
-    // now we need to walk the collectionProps tree to cross-check replica state with live nodes
-    crossCheckReplicaStateWithLiveNodes(liveNodes, collectionProps);
-
-    NamedList<Object> clusterStatus = new SimpleOrderedMap<>();
-    clusterStatus.add("collections", collectionProps);
-
-    // read cluster properties
-    Map<String, Object> clusterProps = zkStateReader.getClusterProperties();
-    if (clusterProps != null && !clusterProps.isEmpty()) {
-      clusterStatus.add("properties", clusterProps);
-    }
-
     // add the alias map too
     Map<String, String> collectionAliasMap = aliases.getCollectionAliasMap(); // comma delim
     if (!collectionAliasMap.isEmpty()) {
       clusterStatus.add("aliases", collectionAliasMap);
     }
 
-    // add the roles map
-    if (roles != null) {
-      clusterStatus.add("roles", roles);
-    }
+    // now we need to walk the collectionProps tree to cross-check replica state with live nodes
+    crossCheckReplicaStateWithLiveNodes(liveNodes, collectionProps);
 
-    // add live_nodes
-    clusterStatus.add("live_nodes", liveNodes);
-
-    results.add("cluster", clusterStatus);
+    clusterStatus.add("collections", collectionProps);
   }
 
   /**
--- a/solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
@@ -44,7 +44,7 @@ import org.apache.zookeeper.KeeperException;
 
 public class ClusterStatus {
   private final ZkStateReader zkStateReader;
-  private SolrParams solrParams;
+  private final SolrParams solrParams;
   private final String collection; // maybe null
 
   public static final String INCLUDE_ALL = "includeAll";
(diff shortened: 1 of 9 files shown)


solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -92,11 +96,57 @@ public static Health combine(Collection<Health> states) {
   public ClusterStatus(ZkStateReader zkStateReader, ZkNodeProps props) {
     this.zkStateReader = zkStateReader;
     this.message = props;
-    collection = props.getStr(ZkStateReader.COLLECTION_PROP);
+    collectionParam = props.getStr(ZkStateReader.COLLECTION_PROP);
+    liveNodesParam = props.getBool(ZkStateReader.LIVENODES_PROP, false);
+    clusterPropertiesParam = props.getBool(ZkStateReader.CLUSTER_PROP, false);
+    rolesParam = props.getBool(ZkStateReader.ROLES_PROP, false);
+    includeAll = props.getBool(ZkStateReader.INCLUDE_ALL, true);
```
aparnasuresh85, line 103: Add a new property to CLUSTERSTATUS request to fetch all info - live nodes, collections, aliases, roles, cluster properties. Set to true by default.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -92,11 +96,57 @@ public static Health combine(Collection<Health> states) {
   public ClusterStatus(ZkStateReader zkStateReader, ZkNodeProps props) {
     this.zkStateReader = zkStateReader;
     this.message = props;
-    collection = props.getStr(ZkStateReader.COLLECTION_PROP);
+    collectionParam = props.getStr(ZkStateReader.COLLECTION_PROP);
```
aparnasuresh85, line 99: I am open to renaming this back to "collection"

solr/solrj/src/java/org/apache/solr/client/solrj/impl/ClusterStateProvider.java
```diff
@@ -107,9 +111,15 @@ default DocCollection getCollection(String name) throws IOException {
 
   /** Obtain a cluster property, or the default value if it doesn't exist. */
   default <T> T getClusterProperty(String key, T defaultValue) {
+    Map<String, Object> properties = getClusterProperties();
```
aparnasuresh85, line 114: The configurable request could now return null, hence the change

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -88,6 +88,9 @@ public class ZkStateReader implements SolrCloseable {
   public static final String CORE_NAME_PROP = "core";
 
   public static final String COLLECTION_PROP = "collection";
+  public static final String INCLUDE_ALL = "includeAll";
+  public static final String LIVENODES_PROP = "liveNodes";
+  public static final String CLUSTER_PROP = "clusterProperties";
```
dsmiley, line 93: No; please don't add constants to ZkStateReader that are really HTTP parameter level constants.  Instead, see CollectionAdminParams if you must... but for this PR, these constants are really specific to a particular request, so I suggest just leaving them as constants in ClusterStatus.java

I'm aware some older constants here are used for multiple purposes but that's a bad practice.  The constants here _should_ only be used for objects in ZK.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -45,7 +45,11 @@
 public class ClusterStatus {
   private final ZkStateReader zkStateReader;
   private final ZkNodeProps message;
-  private final String collection; // maybe null
+  private final String collectionParam; // maybe null
+  private final boolean liveNodesParam;
+  private final boolean clusterPropertiesParam;
+  private final boolean rolesParam;
+  private final boolean includeAll;
```
dsmiley, line 52: Instead see `org.apache.solr.handler.admin.ColStatus#getColStatus` for better naming, like "withSegments" etc.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -92,11 +96,57 @@ public static Health combine(Collection<Health> states) {
...
+    if (includeAll || clusterPropertiesParam) {
+      // read cluster properties
+      Map<String, Object> clusterProps = zkStateReader.getClusterProperties();
+      if (clusterProps != null && !clusterProps.isEmpty()) {
+        clusterStatus.add("properties", clusterProps);
+      }
+    }
+
+    // add the roles map
+    if (includeAll || rolesParam) {
```
dsmiley, line 130: instead of constantly checking includeAll _as well as_ the one param for the category, consider defaulting the category boolean to be based on includeAll.  The way you coded it, it's impossible to include all and exclude one category.  I'm looking at the beginning of ColStatus.getColStatus for inspiration and seeing what I suggest would be really trivial; just change a bunch of "false" references to "includeAll"

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -173,6 +167,33 @@ private ClusterState fetchClusterState(
     return cs;
   }
 
+  private SimpleOrderedMap<?> submitClusterStateRequest(
+      SolrClient client,
+      String collection,
+      boolean fetchLiveNodes,
+      boolean fetchClusterProp,
+      boolean fetchNodeRoles)
```
dsmiley, line 175: Methods with tons of booleans is hard to read at the caller.  Would it be sufficiently concise to heave each caller request what they want?  I could imagine doing so in only a 2-3 lines each.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -228,12 +249,9 @@ > getCacheTimeout()) {
...
   @SuppressWarnings({"rawtypes", "unchecked"})
-  private static Set<String> fetchLiveNodes(SolrClient client) throws Exception {
-    ModifiableSolrParams params = new ModifiableSolrParams();
-    params.set("action", "CLUSTERSTATUS");
-    QueryRequest request = new QueryRequest(params);
-    request.setPath("/admin/collections");
-    NamedList cluster = (SimpleOrderedMap) client.request(request).get("cluster");
+  private Set<String> fetchLiveNodes(SolrClient client) throws Exception {
+
+    SimpleOrderedMap<?> cluster = submitClusterStateRequest(client, null, true, false, false);
```
dsmiley, line 254: oh yes, null, true false false.   Uh..... what does that refer to? ;-).    See my other comment.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/ClusterStateProvider.java
dsmiley: The change could introduce a regression to a caller, albeit rather minor.  Why not have it return Collections.empty() so the callers can skip the null check?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/ClusterStateProvider.java
```diff
@@ -107,9 +111,15 @@ default DocCollection getCollection(String name) throws IOException {
...
+    if (properties == null) {
+      return defaultValue;
+    }
     @SuppressWarnings({"unchecked"})
-    T value = (T) getClusterProperties().get(key);
-    if (value == null) return defaultValue;
+    T value = (T) properties.get(key);
```
dsmiley, line 119: call getOrDefault

dsmiley (commented):
@gerlowskija This PR is making the CLUSTERSTATUS request more configurable.  How do you recommend this be incorporated by V2 API?  Maybe V2 should support this and we don't worry about V1?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
aparnasuresh85: Addressed in latest commit

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
aparnasuresh85: Addressed in latest commit

solr/solrj/src/java/org/apache/solr/client/solrj/impl/ClusterStateProvider.java
aparnasuresh85: Done

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
aparnasuresh85: Addressed in latest commit

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
aparnasuresh85: Addressed in latest commit

solr/core/src/java/org/apache/solr/handler/admin/CollectionsHandler.java
```diff
@@ -976,7 +980,17 @@ public Map<String, Object> execute(
         CLUSTERSTATUS,
         (req, rsp, h) -> {
           Map<String, Object> all =
-              copy(req.getParams(), null, COLLECTION_PROP, SHARD_ID_PROP, _ROUTE_, "prs");
+              copy(
```
dsmiley, line 983: I have a small request; feel free to dismiss as scope creep. 

The "copy" here is pointless and furthermore it means we have to go add a whole bunch of parameter references within CollectionsHandler.  The conversion to ZkNodeProps is also pointless.  For other commands involving the Overseer, there's a point to it but that doesn't apply to this one and some others for that matter.  Couldn't you just change ClusterStatus's constructor to accept the request , and use params in there instead of the ZkNodeProps?

If that's too much scope creep for you, a smaller subset of this is simply removing the copy, and thus remove all the parameter references here.  Done; fewer lines of code.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/ClusterStateProvider.java
```diff
@@ -108,8 +112,7 @@ default DocCollection getCollection(String name) throws IOException {
   /** Obtain a cluster property, or the default value if it doesn't exist. */
   default <T> T getClusterProperty(String key, T defaultValue) {
     @SuppressWarnings({"unchecked"})
-    T value = (T) getClusterProperties().get(key);
-    if (value == null) return defaultValue;
+    T value = (T) getClusterProperties().getOrDefault(key, defaultValue);
     return value;
```
dsmiley, line 116: one line; no need to declare

dsmiley (commented):
Can you please update the ref guide, `cluster-node-management.adoc`, as well?

solr/core/src/java/org/apache/solr/handler/admin/CollectionsHandler.java
aparnasuresh85: Removed the copy, reading params from SolrRequest.getParams in ClusterStatus instead.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -97,6 +101,55 @@ public ClusterStatus(ZkStateReader zkStateReader, ZkNodeProps props) {
...
+    // add the roles map
+    if (withRoles) {
+      Map<?, ?> roles = null;
+      if (zkStateReader.getZkClient().exists(ZkStateReader.ROLES, true)) {
+        roles =
+            (Map<?, ?>)
+                Utils.fromJSON(
+                    zkStateReader.getZkClient().getData(ZkStateReader.ROLES, null, null, true));
+      }
+      if (roles != null) {
```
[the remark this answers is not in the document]
aparnasuresh85: Updated.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -413,4 +412,11 @@ public String getQuorumHosts() {
     }
     return String.join(",", this.liveNodes);
   }
+
+  public enum ClusterStateRequestType {
```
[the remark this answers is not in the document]
aparnasuresh85: Done

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -96,39 +96,43 @@ public static Health combine(Collection<Health> states) {
...
       // read cluster properties
-      Map<String, Object> clusterProps = zkStateReader.getClusterProperties();
-      if (clusterProps != null && !clusterProps.isEmpty()) {
-        clusterStatus.add("properties", clusterProps);
+      clusterProps = zkStateReader.getClusterProperties();
+      if (clusterProps == null || clusterProps.isEmpty()) {
+        clusterProps = Collections.emptyMap();
       }
     }
+    clusterStatus.add("properties", clusterProps);
```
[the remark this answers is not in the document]
aparnasuresh85: Yes. Updated

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -34,19 +34,23 @@
...
+import org.apache.solr.common.params.SolrParams;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.common.util.SimpleOrderedMap;
 import org.apache.solr.common.util.Utils;
 import org.apache.zookeeper.KeeperException;
 
 public class ClusterStatus {
   private final ZkStateReader zkStateReader;
-  private final ZkNodeProps message;
+  private SolrParams solrParams;
```
dsmiley, line 47: not final?

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -89,14 +93,61 @@ public static Health combine(Collection<Health> states) {
...
   }
 
   public void getClusterStatus(NamedList<Object> results)
       throws KeeperException, InterruptedException {
+    List<String> liveNodes = null;
+    NamedList<Object> clusterStatus = new SimpleOrderedMap<>();
+
+    boolean includeAll = this.solrParams.getBool(INCLUDE_ALL, true);
+    boolean withLiveNodes =
+        this.solrParams.getBool(LIVENODES_PROP, includeAll) || (collection != null);
```
dsmiley, line 109: why compare with collection here?

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -89,14 +93,61 @@ public static Health combine(Collection<Health> states) {
...
     this.zkStateReader = zkStateReader;
-    this.message = props;
-    collection = props.getStr(ZkStateReader.COLLECTION_PROP);
+    this.solrParams = params;
+    collection = params.get(ZkStateReader.COLLECTION_PROP);
   }
 
   public void getClusterStatus(NamedList<Object> results)
       throws KeeperException, InterruptedException {
+    List<String> liveNodes = null;
```
dsmiley, line 104: why define this up top?

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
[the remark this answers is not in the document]
dsmiley: (I don't see it)

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -89,14 +93,61 @@ public static Health combine(Collection<Health> states) {
...
+    Map<?, ?> roles = Collections.emptyMap();
+    if (withRoles) {
+      if (zkStateReader.getZkClient().exists(ZkStateReader.ROLES, true)) {
+        roles =
+            (Map<?, ?>)
+                Utils.fromJSON(
+                    zkStateReader.getZkClient().getData(ZkStateReader.ROLES, null, null, true));
+      }
+    }
+    clusterStatus.add("roles", roles);
```
dsmiley, line 144: don't add unless `withRoles`

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -139,6 +184,7 @@ public void getClusterStatus(NamedList<Object> results)
     if (didNotFindCollection && isAlias) {
       // In this case this.collection is an alias name not a collection
       // get all collections and filter out collections not in the alias
+      // clusterState.getCollectionsMap() should be replaced with an inexpensive call
```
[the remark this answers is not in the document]
dsmiley: (waiting)

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -191,43 +236,23 @@ public void getClusterStatus(NamedList<Object> results)
       }
       String configName = clusterStateCollection.getConfigName();
       collectionStatus.put("configName", configName);
-      if (message.getBool("prs", false) && clusterStateCollection.isPerReplicaState()) {
+      if (this.solrParams.getBool("prs", false) && clusterStateCollection.isPerReplicaState()) {
```
dsmiley, line 239: nitpick: I'd remove "this."

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -147,6 +191,7 @@ public void getClusterStatus(NamedList<Object> results)
 
     NamedList<Object> collectionProps = new SimpleOrderedMap<>();
 
+    // loop will run only once
```
dsmiley, line 194: Why is that?

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
dsmiley: they are still here.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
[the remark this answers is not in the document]
dsmiley: (waiting)

solr/solrj/src/java/org/apache/solr/client/solrj/impl/ClusterStateProvider.java
dsmiley: (waiting)

solr/solrj/src/test/org/apache/solr/client/solrj/impl/HttpClusterStateSSLTest.java
```diff
@@ -97,20 +97,20 @@ public void testHttpClusterStateWithSSL() throws Exception {
...
-    verifyUrlSchemeInClusterState(csp.getClusterState(), collectionId, expectedReplicas);
+    verifyUrlSchemeInClusterState(csp.getCollection(collectionId), collectionId, expectedReplicas);
   }
 
   private void verifyUrlSchemeInClusterState(
-      final ClusterState cs, final String collectionId, final int expectedReplicas) {
-    DocCollection dc = cs.getCollection(collectionId);
-    assertNotNull(dc);
-    List<Replica> replicas = dc.getReplicas();
+      final DocCollection collection, final String collectionId, final int expectedReplicas) {
```
dsmiley, line 111: why pass both DocCollection and it's name (collectionId)?

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
aparnasuresh85: Let me know what you think: When cluster status for a collection is fetched, there is a check at some point to `crossCheckReplicaStateWithLiveNodes()`. `liveNodes` would need to be assigned an empty List in the beginning so `crossCheckReplicaStateWithLiveNodes()` does not throw a `NPE`.  

So if `collection!=null` and `liveNodes=false` what would be returned is the list of replicas, but no indication of whether it is on a live node and therefore usable.

dsmiley: Okay I understand that crossCheckReplicaStateWithLiveNodes (used by collection processing) depends on liveNodes.  But the client didn't request liveNodes, therefore we shouldn't return it.  I think "withLiveNodes" boolean should be simple/obvious here, as the others.  Instead down below, you could modify the condition that decides to gather liveNodes by also checking if collection processing is needed.  But still; don't set on the response unless "withLiveNodes".

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
aparnasuresh85: Should be gone now.

aparnasuresh85:
> Can you please update the ref guide, `cluster-node-management.adoc`, as well?

Done, let me know if it needs more changes

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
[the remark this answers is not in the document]
aparnasuresh85: Should be visible now

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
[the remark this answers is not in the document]
aparnasuresh85: Done

solr/solrj/src/java/org/apache/solr/client/solrj/impl/ClusterStateProvider.java
aparnasuresh85: I am unable to remove the declare else the build fails with 

Task :solr:solrj:compileJava
/Users/aparna.suresh/blt/upstream/configurable-clusterstatus/solr/solrj/src/java/org/apache/solr/client/solrj/impl/ClusterStateProvider.java:114: warning: [unchecked] unchecked cast
  return (T) getClusterProperties().getOrDefault(key, defaultValue);
                         ^
 required: T
 found:  Object
 where T is a type-variable:
  T extends Object declared in method <T>getClusterProperty(String,T)
error: warnings found and -Werror specified
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
1 error
1 warning

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
aparnasuresh85: Done

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
aparnasuresh85: Updated

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
aparnasuresh85: Updated

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
aparnasuresh85: Was based on an earlier revision, removed it now

solr/solr-ref-guide/modules/deployment-guide/pages/cluster-node-management.adoc
```diff
@@ -99,6 +99,41 @@ Multiple shard names can be specified as a comma-separated list.
...
+|Optional |Default: will default to the default value of includeAll parameter specified below
+|===
++
+If set to true, returns the roles within the cluster.
+
+`includeAll`::
++
+[%autowidth,frame=none]
+|===
+|Optional |Default: false
```
dsmiley, line 133: includeAll must default to "true" to be backwards compatible, and this is what you used in the code.  And you didn't document it.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -89,14 +93,61 @@ public static Health combine(Collection<Health> states) {
...
+      liveNodes =
+          zkStateReader.getZkClient().getChildren(ZkStateReader.LIVE_NODES_ZKNODE, null, true);
+      // add live_nodes
+      if (withLiveNodes) clusterStatus.add("live_nodes", liveNodes);
+    }
+    if (withCollection) {
+      fetchClusterStatusForCollOrAlias(clusterStatus, liveNodes);
+    }
+
+    Map<String, Object> clusterProps = Collections.emptyMap();
```
dsmiley, line 123: Why define outside the "if"?

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -89,14 +93,61 @@ public static Health combine(Collection<Health> states) {
...
+      // read cluster properties
+      clusterProps = zkStateReader.getClusterProperties();
+      if (clusterProps == null || clusterProps.isEmpty()) {
+        clusterProps = Collections.emptyMap();
+      }
+      clusterStatus.add("properties", clusterProps);
+    }
+
+    // add the roles map
+    Map<?, ?> roles = Collections.emptyMap();
```
dsmiley, line 134: Why define outside the "if"?

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -89,14 +93,61 @@ public static Health combine(Collection<Health> states) {
...
       throws KeeperException, InterruptedException {
+    NamedList<Object> clusterStatus = new SimpleOrderedMap<>();
+
+    boolean includeAll = solrParams.getBool(INCLUDE_ALL, true);
+    boolean withLiveNodes = solrParams.getBool(LIVENODES_PROP, includeAll);
+    boolean withClusterProperties = solrParams.getBool(CLUSTER_PROP, includeAll);
+    boolean withRoles = solrParams.getBool(ZkStateReader.ROLES_PROP, includeAll);
+    boolean withCollection = includeAll || (collection != null);
+
+    List<String> liveNodes = Collections.emptyList();
```
dsmiley, line 112: shouldn't the default be null since it won't necessarily be fetched?

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -89,14 +93,61 @@ public static Health combine(Collection<Health> states) {
...
+    if (withLiveNodes || collection != null) {
+      liveNodes =
+          zkStateReader.getZkClient().getChildren(ZkStateReader.LIVE_NODES_ZKNODE, null, true);
+      // add live_nodes
+      if (withLiveNodes) clusterStatus.add("live_nodes", liveNodes);
+    }
+    if (withCollection) {
+      fetchClusterStatusForCollOrAlias(clusterStatus, liveNodes);
```
dsmiley, line 120: I would add a line above `assert liveNodes != null`

solr/solrj/src/java/org/apache/solr/client/solrj/impl/ClusterStateProvider.java
dsmiley: Wow, you're right; even the code for getOrDefault suggests it's necessary.  Okay.

dsmiley (commented):
Looking good; close to mergeable.

solr/solr-ref-guide/modules/deployment-guide/pages/cluster-node-management.adoc
aparnasuresh85: Done

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
aparnasuresh85: Done

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
aparnasuresh85: Updated. I had it as an empty list earlier to avoid a NPE further on, but the assert achieves the same purpose

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
aparnasuresh85: Updated

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
aparnasuresh85: Updated

dsmiley:
@gerlowskija in V2, I'm thinking these are separate endpoints

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -291,7 +297,7 @@ public void testHttpCspPerf() throws Exception {
       // lookup)
       assertLogCount(collectionClusterStateLogs, 1);
       // 1 call to fetch entire cluster state from HttpCSP.fetchLiveNodes()
-      assertLogCount(entireClusterStateLogs, 1);
+      assertLogCount(entireClusterStateLogs, 0);
```
aparnasuresh85, line 300: No more calls to fetch entire cluster state!

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
```diff
@@ -100,7 +100,13 @@ public class CloudHttp2SolrClientTest extends SolrCloudTestCase {
   private static final Pattern PATTERN_WITHOUT_COLLECTION =
       Pattern.compile(
           "path=/admin/collections.*params=\\{[^}]*action=CLUSTERSTATUS"
-              + "(?![^}]*collection=)[^}]*\\}");
+              + "(?=[^}]*includeAll=true)"
```
aparnasuresh85, line 103: Regex pattern (sad) updated to reflect the new params to CLUSTERSTATUS call

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -84,6 +84,11 @@ public void init(List<String> solrUrls) throws Exception {
   /** Create a SolrClient implementation that uses the specified Solr node URL */
   protected abstract SolrClient getSolrClient(String baseUrl);
 
+  @Override
```
aparnasuresh85, line 87: Super class ClusterStateProvider's implementation calls getClusterState().getCollectionOrNull(collectionName) - in that case BaseHttpCSP could make a call to fetch the entire cluster state! This change is to prevent that from happening.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -89,14 +93,61 @@ public static Health combine(Collection<Health> states) {
     }
   }
 
-  public ClusterStatus(ZkStateReader zkStateReader, ZkNodeProps props) {
+  public ClusterStatus(ZkStateReader zkStateReader, SolrParams params) {
     this.zkStateReader = zkStateReader;
-    this.message = props;
-    collection = props.getStr(ZkStateReader.COLLECTION_PROP);
+    this.solrParams = params;
+    collection = params.get(ZkStateReader.COLLECTION_PROP);
```
dsmiley, line 99: isn't it a little awkward that this param is initialized here despite lots of params being taken in getClusterStatus?   (change or not as you wish)

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -89,14 +93,61 @@ public static Health combine(Collection<Health> states) {
...
+    if (withCollection) {
+      assert liveNodes != null;
+      fetchClusterStatusForCollOrAlias(clusterStatus, liveNodes);
+    }
+
+    if (withClusterProperties) {
+      Map<String, Object> clusterProps = Collections.emptyMap();
+      // read cluster properties
+      clusterProps = zkStateReader.getClusterProperties();
+      if (clusterProps == null || clusterProps.isEmpty()) {
```
dsmiley, line 128: isEmpty check isn't accomplishing anything here

solr/solr-ref-guide/modules/deployment-guide/pages/cluster-node-management.adoc
```diff
@@ -99,6 +99,41 @@ Multiple shard names can be specified as a comma-separated list.
...
+|===
++
+If set to true, returns the roles within the cluster.
+
+`includeAll`::
++
+[%autowidth,frame=none]
+|===
+|Optional |Default: true
+|===
```
dsmiley, line 134: just a simple sentence to list the sections

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
aparnasuresh85: other params except `collection` are used only in `getClusterStatus()`, but collection is also used in `fetchClusterStatusForCollOrAlias()`

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -191,43 +234,23 @@ public void getClusterStatus(NamedList<Object> results)
...
-    // read cluster properties
-    Map<String, Object> clusterProps = zkStateReader.getClusterProperties();
-    if (clusterProps != null && !clusterProps.isEmpty()) {
-      clusterStatus.add("properties", clusterProps);
-    }
-
     // add the alias map too
     Map<String, String> collectionAliasMap = aliases.getCollectionAliasMap(); // comma delim
     if (!collectionAliasMap.isEmpty()) {
       clusterStatus.add("aliases", collectionAliasMap);
```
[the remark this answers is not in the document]
aparnasuresh85: Not sure. It is a simple fetch from another map (does list to comma delimited conversion). Not sure the caller would want to know about aliases too in this case.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
aparnasuresh85: You are right. Updated.

solr/solr-ref-guide/modules/deployment-guide/pages/cluster-node-management.adoc
aparnasuresh85: Done

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
dsmiley: Awesome.

At this juncture, I recommend changes to this test because the way it works now is awkward:  Only have one LogListener for the "adminRequestLogs" (simple; 3 is more confusing).  Consume the log messages in the test (don't merely count) to assert that the log message is what it should be, ideally with a substring check or a simple regex if you must.  Consume & assert this after each step that we know will emit a request:  after client construction and once more after the first "add" (doc).  Your code comments here are then redundant as the assertions will be more specific at the right times to be self explanatory.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
dsmiley: let's do away with this regex; it's gotten out of control and should cause us to re-think what we're doing (see my other comment)

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
aparnasuresh85: I agree. Improved the test

solr/solrj/src/test/org/apache/solr/client/solrj/impl/CloudHttp2SolrClientTest.java
aparnasuresh85: Updated test doesnt include regexes anymore.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
aparnasuresh85: Done

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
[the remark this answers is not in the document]
aparnasuresh85: added another parameter to fetch all aliases, updated the adoc for CLUSTERSTATUS

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -116,9 +118,19 @@ public void getClusterStatus(NamedList<Object> results)
       // add live_nodes
       if (withLiveNodes) clusterStatus.add("live_nodes", liveNodes);
     }
+
+    Aliases aliases = null;
+    if (withCollection || withAliases) {
```
dsmiley, line 123: why withCollection?

dsmiley (approved):
Nice!
Nearly ready to commit IMO; needs a CHANGES.txt (which I could add if needed).

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
aparnasuresh85: Whether a collection parameter is passed (where the input could be either a collection or an alias) or if includeAll is set to true, aliases are retrieved from ZkStateReader to create a mapping of collections to the list of aliases associated with them and returned as part of the response.

dsmiley: Since they occupy the same namespace, I could see returning aliases if you want collections.

aparnasuresh85:
> Nice! Nearly ready to commit IMO; needs a CHANGES.txt (which I could add if needed).

Done.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -109,7 +109,7 @@ public void getClusterStatus(NamedList<Object> results)
     boolean withClusterProperties = solrParams.getBool(CLUSTER_PROP, includeAll);
     boolean withRoles = solrParams.getBool(ZkStateReader.ROLES_PROP, includeAll);
     boolean withCollection = includeAll || (collection != null);
-    boolean withAliases = solrParams.getBool(ALIASES_PROP, includeAll);
+    boolean withAliases = solrParams.getBool(ALIASES_PROP, includeAll) || withCollection;
```
dsmiley, line 112: but if the user specifies "aliases=false" why overrule their decision?

aparnasuresh85: I agree - this conflicts with your comment here https://github.com/apache/solr/pull/2599/files#r1707539294

The `withAliases` refers to the property which when enabled would fetch the entire aliasMap stored in memory. 

When `collection` parameter is non-null, aliases are fetched from ZkStateReader to build the mapping from collections to aliases or vice versa when the input is an alias.

It then makes sense to keep this property separate as:
`boolean withAliases = solrParams.getBool(ALIASES_PROP, includeAll);`

I am not sure what your other comment meant.

aparnasuresh85: Updated the PR to request for aliases explicitly

dsmiley (approved):

dsmiley:
+1 Excellent; I'll merge soon

dsmiley (commented):
instead of getting the queue size, which keeps increasing, I think it'd be clearer to assertNull(adminLogs.pollMessage()) at the spots right after you read the last message.
I can see other tests that maybe you got inspiration from that assert the queue size but in the spots I looked at, it was the degenerate cases of 0 and 1, thus didn't seem so awkward as an increasing number.

iamsanjay:
git bisect points to this commit!
```
  - org.apache.solr.client.solrj.impl.ClusterStateProviderTest.testGetClusterState (:solr:solrj)
    Test history: https://ge.apache.org/scans/tests?search.rootProjectNames=solr-root&tests.container=org.apache.solr.client.solrj.impl.ClusterStateProviderTest&tests.test=testGetClusterState http://fucit.org/solr-jenkins-reports/history-trend-of-recent-failures.html#series/org.apache.solr.client.solrj.impl.ClusterStateProviderTest.testGetClusterState
    Test output: /tmp/src/solr/solr/solrj/build/test-results/test/outputs/OUTPUT-org.apache.solr.client.solrj.impl.ClusterStateProviderTest.txt
    Reproduce with: ./gradlew :solr:solrj:test --tests "org.apache.solr.client.solrj.impl.ClusterStateProviderTest.testGetClusterState" -Ptests.jvms=96 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=33F6364CB7D4507A -Ptests.timeoutSuite=600000! -Ptests.file.encoding=ISO-8859-1
```

PR #2017, opened 2023-10-17
SOLR-10808: SchemaVersion 1.7 - DocValues default to true

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-10808

# Checklist

Please review the following and check all that apply:

- [ ] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/solr/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [ ] I have created a Jira issue and added the issue ID to my pull request title.
- [ ] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [ ] I have developed this patch against the `main` branch.
- [ ] I have run `./gradlew check`.
- [ ] I have added tests for my changes.
- [ ] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

Need to:

- [x] Go through and update all of the test configSets
- [x] Make some tests for the schemaVersion logic
- [x] Confirm that we can do this in a minor release since the schemaVersion makes this backwards-compatible

PR #2601, opened 2024-07-26
SOLR-12963: Default uninvertible=false in schema version 1.7

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-12963

Follow up from https://github.com/apache/solr/pull/2017, make `uninvertible=false` default starting in schema version 1.7, along with `docValues=true`.


dsmiley (commented):
In most of the test schemas, shouldn't we use docValues instead of uninvertible?

HoustonPutman:
> In most of the test schemas, shouldn't we use docValues instead of uninvertible?

We do in most of the test schemas. Unfortunately, the randomization of doc values (`docValues="${solr.tests.numeric.dv}"`) in a large number of the test schemas makes it such that we have to have `uninvertible=true` for a lot of them. Other schemas are actually testing the non-docValues functionality, so those have to be uninvertible too.

Ultimately out of the ~230 schemas that I upgraded to 1.7, only 40 of them required `uninvertible=true`, so a relatively small percentage.

dsmiley (approved):
+1 cool; thanks Houston

PR #2613, opened 2024-08-01
[SOLR-17368] Change prometheus dummy metric to a more unique label

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-17368

# Description

`TestPrometheusResponseWriter` was failing ~25% of tests due to dummy metric `dummyMetrics.2xx-responses` not unique enough for exporting compared to `org.eclipse.jetty.server.handler.DefaultHandler.2xx-responses` which could leak between tests creating a duplicate metric.

# Solution

Change the dummy jetty metric with `999xx` that exports a unique enough label that `2xx` cannot clash with it, even with real leaking jetty metrics.

# Checklist

### DIFF
--- a/solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
+++ b/solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
@@ -20,6 +20,7 @@ import com.codahale.metrics.Counter;
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.SettableGauge;
+import com.codahale.metrics.SharedMetricRegistries;
 import java.lang.invoke.MethodHandles;
 import java.util.Arrays;
 import java.util.List;
@@ -34,6 +35,7 @@ import org.apache.solr.common.util.NamedList;
 import org.apache.solr.metrics.SolrMetricManager;
 import org.apache.solr.util.ExternalPaths;
 import org.apache.solr.util.SolrJettyTestRule;
+import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


mlbiscoc:
@dsmiley This should make the dummy metric unique enough to not clash with real leaking jetty metrics from other tests. I ran tests locally multiple times. Any ideas what we can do to make sure that is resilient enough? I want to avoid breaking CI and generating noise for a 3rd time...

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
```diff
@@ -78,7 +78,12 @@ public void testPrometheusStructureOutput() throws Exception {
           output.lines().filter(line -> !line.startsWith("#")).collect(Collectors.toList());
       filteredResponse.forEach(
           (actualMetric) -> {
-            String actualValue = actualMetric.substring(actualMetric.lastIndexOf("} ") + 1);
+            String actualValue;
+            if (actualMetric.contains("}")) {
+              actualValue = actualMetric.substring(actualMetric.lastIndexOf("} ") + 1);
+            } else {
+              actualValue = actualMetric.split(" ")[1];
+            }
```
mlbiscoc, line 86: Also adding this in to better parse numeric value out of prometheus metrics based on whether or not tags exist

dsmiley:
Were you able to reproduce the problem?  If I understand the cause, you'd need to run a number of tests, not just this one test.
The root fix would be to stop metric leaks between tests!  Albeit not today -- okay.

mlbiscoc:
Yes, on `main` branch it fails the same way when running Jetty based tests next to it. From Chris's suggestion, `./gradlew -p solr/core test -Ptests.jvms=1 -Ptests.verbose=true --tests TestPrometheusResponseWriter --tests \*Jetty\*` Ran is multiple times and this looks to clean up the issue on my branch.

Current test failure looks to be from `LocalFSCloudIncrementalBackupTest`? Don't think this is related to this though.

PR #2615, opened 2024-08-04
TestExportWriter fix, loosen order constraint

dsmiley:
This test was asserting the docs from /export came in the same order as an equivalent from /select.  But that's too strong an expectation; only the sorted values should be equivalent.

It's remarkable to me how much code was written to test it as it was before... I wonder if I'm missing something by replacing it with something fairly trivial?

----

The following test was failing with this seed:
> ./gradlew :solr:core:test --tests "org.apache.solr.handler.export.TestExportWriter.testMultipleSorts"   -Ptests.seed=D7C851B5C905B099

I debugged it for a while; didn't fail due to any recent changes; isn't concurrency related.  Even optimizing to one segment still had the same test failure.

### DIFF
--- a/solr/core/src/test/org/apache/solr/handler/export/TestExportWriter.java
+++ b/solr/core/src/test/org/apache/solr/handler/export/TestExportWriter.java
@@ -23,7 +23,6 @@ import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


dsmiley:
I'll merge this in a couple days.

joel-bernstein (commented):
This change looks fine to me.

PR #2619, opened 2024-08-06
SOLR-17391: Cached thread pools actually have more than one thread.

psalagnac:
https://issues.apache.org/jira/browse/SOLR-17391

# Description

Cached thread pools never start more than one thread. Consequently, all the async tasks are executed sequentially.
This is mostly a performance issue of backup/restore with large collections.

# Solution

Fix `newMDCAwareCachedThreadPool` so it behaves as expected.

# Tests

Added a test to assert the pool is multithreaded.

# Checklist


solr/solrj/src/test/org/apache/solr/common/util/ExecutorUtilTest.java
```diff
@@ -108,6 +111,48 @@ public void testExecutorUtilAwaitsTerminationWhenTaskRespectsInterupt() throws E
...
+
+      MDCLoggingContext.setCoreName(core);
+      executor.submit(task);
+    }
+
+    ExecutorUtil.shutdownAndAwaitTermination(executor);
+
+    // assert the pool was actually multithreaded. Since we submitted many tasks,
+    // all the threads should have been started
+    assertEquals(5, maxConcurrentTasks.get());
```
madrob, line 153: It's possible that taskCount will be < 5 and the test will fail here.

psalagnac: Theoretically yes. But that's very unlikely as we submit 1000 tasks.
I consider a decent implementation of the thread pool should always start the 5 requested threads.

psalagnac: I added a latch in the test, so we don't depend on the scheduler.

madrob: Your previous code submitted uniformly random 0-100 tasks, it would have failed 5% of the time. Thank you for adding the `5 + ...` and the latch.

dsmiley (approved):
Nice!  I didn't know about `executor.allowCoreThreadTimeOut` until today :-)
I see there are two callers,`org.apache.solr.handler.admin.CoreAdminHandler.CoreAdminAsyncTracker#expensiveExecutor` and `org.apache.solr.core.CoreContainer#replayUpdatesExecutor`.  Both seem suited for this change.

dsmiley:
After reading ThreadPoolExecutor's docs a bit more, I was surprised to see that a new thread will be created for a new task when the thread count is less than the core size _even if there are threads idle_.  I checked out this PR and did a trivial experiment where I executed two tasks in series (one completing before the other), even with a sleep in-between, and both reported different threads.  That's sad and makes no sense to me why it would be designed that way (or maybe it's more like a documented deficiency?).  Any way, It's not that important; not too many idle threads.  I suppose if the idle time were really short (a second?) then there would be no wasted idle threads as they would barely exist.  But it would mean more thread creation over time and I'm not sure what to think of that.

Any way, I think this is a nice change.

dsmiley:
Are we good with the CHANGES.txt and other changes?  (Pierre or anyone else?)

psalagnac:
> Are we good with the CHANGES.txt and other changes? (Pierre or anyone else?)

I would also mention shard Split, as it was also impacted. A single thread per node was processing all backup/restore/split command.

All other changes are good to me.
Thanks for handling this.

psalagnac:
I also updated Javadoc of the other variant (unbounded pool) to reduce confusion.

PR #2621, opened 2024-08-06
SOLR-17394: Check response status in IndexFetcher

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-17394

# Description

Prior to this commit, IndexFetcher would mis-read 404 responses from other nodes.  It eventually catches this mistake and retries, but not before it wastefully allocates a massive (> 1GB) byte array.

These allocations were relatively infrequent, as they require out of date leader information to occur.  But they are big enough to impact performance in production, and were observed to cause OOM's when running tests.

# Solution

This commit avoids this by checking the HTTP status code of all IndexFetcher requests/responses, even those that utilize InputStreamResponseParser.

# Tests

I don't have unit tests for this as it's a bit hard to exercise in isolation.  But I am able to see the newly-introduced IOE message in the logs, and the massive `byte[]` is no longer allocated.

The other way I've tested this, such as it is, is through 'beasting' a particular test which trigger a OOM and heap dump on about every 1/10 iterations.  Setting `tests.heapsize` to any moderate value (e.g. `1GB`) in `gradle.properties` and running the command below reproduces the OOM reliably for me on `main` without this PR.  (And reproduces the OOM never with the PR in place.)

```
./gradlew -p solr/solrj beast -Ptests.seed=F024B5067D7FCE8F -Ptests.neverUpToDate=true -Ptests.dups=50 --tests CloudHttp2SolrClientTest
```

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
+++ b/solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
@@ -71,6 +71,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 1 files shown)


solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
```diff
@@ -1988,16 +1989,38 @@ private FastInputStream getStream() throws IOException {
...
         }
         return new FastInputStream(is);
       } catch (Exception e) {
-        // close stream on error
-        IOUtils.closeQuietly(is);
-        throw new IOException("Could not download file '" + fileName + "'", e);
+        closeStreamAndThrowIOE(is, "Could not download file '" + fileName + "'", Optional.of(e));
+      }
+
+      // Unreachable b/c closeStreamAndThrowIOE will always throw
```
dsmiley, line 2013: You could have returned the IOException and let the caller throw it.

solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
```diff
@@ -1988,16 +1989,38 @@ private FastInputStream getStream() throws IOException {
...
+      return null;
+    }
+
+    private void closeStreamAndThrowIOE(
+        InputStream is, String exceptionMessage, Optional<Exception> e) throws IOException {
```
dsmiley, line 2018: A matter of taste but I would prefer we/programmers use Optionals conservatively -- when it's clearly a help.  They are verbose; should have been built into the language like Kotlin.  In particular, I suggest we not use Optional as a parameter; only return them if we use them.

See https://www.baeldung.com/java-optional #15 (and I've seen this advice elsewhere too)

dsmiley (commented):
This was merged so fast that it got not review; I wanted to review it

solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
gerlowskija: Yeah, that would've been better in hindsight, good point.

gerlowskija:
> This was merged so fast that it got not review; I wanted to review it

I thought 3 days was a reasonable length of time to wait on a pretty clear and minor bugfix; I hope i'm not ruffling any feathers by merging on that timeframe.

But, that said, I'm more than happy to take the review comments now, since you didn't get a chance to review beforehand.  I've replied to several comments "in-line", and will push fixes to address.

(FYI @dsmiley )

dsmiley:
I see less than 2 days: August 6th 3:52pm -> 8th 8:32am.  That's too short.  Also we generally give advanced merge notice if there was no feedback whatsoever.

solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
epugh: We really need more coding style guides and ideally lint checking tools for things like Optional.   Or in 12 months, we'll have this same conversation, but with someone new ;-(.

PR #2622, opened 2024-08-06
SOLR-17253 Release wizard suggests wrong minor version for jira

janhoy:
This should do the trick for suggesting correct jira version


fsparv (approved):

gus-asf (approved):
whoops wrong browser (and therefore account) on last approval

PR #2625, opened 2024-08-07
SOLR-17298 verify that graph, join and rerank queries are unaffected by the use of multithreaded search and re-enable them.

gus-asf:
https://issues.apache.org/jira/browse/SOLR-17298

### DIFF
--- a/solr/test-framework/src/java/org/apache/solr/SolrTestCaseJ4.java
+++ b/solr/test-framework/src/java/org/apache/solr/SolrTestCaseJ4.java
@@ -1299,17 +1299,6 @@ public abstract class SolrTestCaseJ4 extends SolrTestCase {
   }
 
   /**
-   * Generates a SolrQueryRequest using the LocalRequestFactory. Note that tests that want to test
-   * multi-threading should ensure that multiple commits are used and that the merge policy will
-   * leave a suitable number of segments.
-   *
-   * @see #lrf
-   */
-  public static SolrQueryRequest reqRandMulti(String... q) {
-    return lrf.makeRequest(true, q);
-  }
-
-  /**
    * Generates a SolrQueryRequest using the LocalRequestFactory
    *
    * @see #lrf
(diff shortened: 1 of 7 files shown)


solr/test-framework/src/java/org/apache/solr/SolrTestCaseJ4.java
```diff
@@ -1324,6 +1335,16 @@ public static SolrQueryRequest req(SolrParams params, String... moreParams) {
     return new LocalSolrQueryRequest(h.getCore(), mp);
   }
 
+  /** Generates a SolrQueryRequest randomly assigning multi or single threaded processing. */
+  public static SolrQueryRequest reqRandMulti(SolrParams params, String... moreParams) {
```
dsmiley, line 1339: I don't like this approach.  Imagine that we embrace this -- would this then mean all/most calls to `req` would switch to `reqAndMulti`?  Not good.  Assuming we imagine that `multiThreaded=true` should just work (even if may not actually use multiple threads for various reasons), I'd prefer we change a system property to have true/false and randomly chosen.

solr/test-framework/src/java/org/apache/solr/SolrTestCaseJ4.java
```diff
@@ -1324,6 +1335,16 @@ public static SolrQueryRequest req(SolrParams params, String... moreParams) {
...
+    ModifiableSolrParams mp = new ModifiableSolrParams(params);
+    for (int i = 0; i < moreParams.length; i += 2) {
+      mp.add(moreParams[i], moreParams[i + 1]);
+    }
+    mp.add("multiThreaded", String.valueOf(random().nextBoolean()));
```
dsmiley, line 1344: If we need to augment parameters, consider SolrParams.wrapDefaults.  It'd be nice if we also had a SolrParams.of(name, value) to make it easy/idiomatic to combine.

solr/test-framework/src/java/org/apache/solr/util/TestHarness.java
```diff
@@ -453,6 +455,22 @@ public LocalRequestFactory() {}
...
+     * leave a suitable number of segments.
+     *
+     * @param testMultiThreadedRandom if true 50% of requests will include the multiThreaded=true
+     *     parameter.
+     * @param q the parameters for the request.
+     * @see #makeRequest(String...)
+     * @return a LocalSolrQueryRequest based on the supplied parameters
+     */
+    @SuppressWarnings("unchecked")
+    public LocalSolrQueryRequest makeRequest(boolean testMultiThreadedRandom, String... q) {
```
dsmiley, line 473: TestHarness might not yet be deprecated but I'd prefer we treat it as such.  I should deprecate it.
https://issues.apache.org/jira/browse/SOLR-11872
Even if SOLR-11872 fails to go anywhere, I also object to this new method here.

PR #2602, opened 2024-07-29
SOLR-16996: Update Solr Exporter for Prometheus cli to use commons-cli

malliaridis:
https://issues.apache.org/jira/browse/SOLR-16996

# Description

This PR replaces the usage of `argparse4j` with `commons-cli` in `SolrExporter.java` of the `prometheus-exporter` module.

Note that the changes mark `baseUrl` as deprecated in favor to [SOLR-16824](https://issues.apache.org/jira/browse/SOLR-16824).

# Tests

I have executed the `prometheus-exporter` tests and `test_prometheus.bats` with Java 17 successfully. Running `test_prometheus.bats` with Java 21 seems to be problematic (even in current `main`).

# Checklist

- [X] I have reviewed the guidelines for [How to Contribute](https://github.com/apache/solr/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.
- [ ] I have created a Jira issue and added the issue ID to my pull request title.
- [X] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [X] I have developed this patch against the `main` branch.
- [X] I have run `./gradlew check`.
- [ ] I have added tests for my changes.
- [ ] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

PR #2627, opened 2024-08-07
SOLR-16996: remove no-longer-applicable solr/licenses/argparse4j-* files

cpoerschke:
#2602 follow-up

https://issues.apache.org/jira/browse/SOLR-16996


malliaridis (approved):
LGTM and thanks for cleaning up what I forgot to remove. 😄

epugh:
I wonder if this is important enough to bring to branch_9_7?

cpoerschke:
> I wonder if this is important enough to bring to branch_9_7?

I think so, yes, i.e. back porting to both branch_9x and branch_9_7 IMO. Otherwise we'll unnecessarily include/ship license files that are no longer needed in the 9.7 release.

PR #2629, opened 2024-08-08
Use EnvUtils to override cache timeout in HTTP ClusterStateProvider

aparnasuresh85:
# Description

The cache timeout to refresh aliases and live nodes in Http ClusterStateProvider is currently hardcoded to 5 seconds. This is pretty frequent.

# Solution

Introducing an override to set the timeout  via EnvUtils that is backwards compatible

# Tests

Ensured existing tests pass
# Checklist

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
@@ -39,6 +39,7 @@ import org.apache.solr.common.cloud.ClusterState;
 import org.apache.solr.common.cloud.DocCollection;
 import org.apache.solr.common.cloud.PerReplicaStates;
 import org.apache.solr.common.params.ModifiableSolrParams;
+import org.apache.solr.common.util.EnvUtils;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.common.util.SimpleOrderedMap;
 import org.apache.solr.common.util.Utils;
@@ -55,7 +56,10 @@ public abstract class BaseHttpClusterStateProvider implements ClusterStateProvid
   volatile Map<String, Map<String, String>> aliasProperties;
   long aliasesTimestamp = 0;
 
-  private int cacheTimeout = 5; // the liveNodes and aliases cache will be invalidated after 5 secs
+  private int cacheTimeout =
+      EnvUtils.getPropertyAsInteger("solr.httpcsp.cache.timeout", 5); // the liveNodes and
+
+  // aliases cache will be invalidated after 60 secs
 
   public void init(List<String> solrUrls) throws Exception {
     for (String solrUrl : solrUrls) {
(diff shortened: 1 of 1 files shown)


aparnasuresh85:
cc @dsmiley

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -55,7 +56,8 @@ public abstract class BaseHttpClusterStateProvider implements ClusterStateProvid
   volatile Map<String, Map<String, String>> aliasProperties;
   long aliasesTimestamp = 0;
 
-  private int cacheTimeout = 5; // the liveNodes and aliases cache will be invalidated after 5 secs
+  // the liveNodes and aliases cache will be invalidated after 5 secs
+  private int cacheTimeout = EnvUtils.getPropertyAsInteger("solr.httpcsp.cache.timeout", 5);
```
janhoy, line 60: Please consider including a category, ref [SIP-21](https://cwiki.apache.org/confluence/display/SOLR/SIP-21%3A+Standardize+system+property+naming). Candiate could be `solr.cloud.httpcsp.cache.timeout`.

dsmiley: This is specific to SolrJ and people don't know what "httpcsp" is.  Only in the last week did you and I start doing that abbreviation.
`solr.solrj.cache.timeout` and assume milliseconds thus 5000 value (that's the default assumption when units don't clarify)

FYI CloudSolrClient has a cache of DocCollection which has a separate timeout and there's no EnvUtils for that.  So I suppose this is okay to introduce in this PR a very general sounding thing so long as it doesn't mean there might be something more specific with its own timeout.  It's tempting to suggest adding "livenodes" as a name component above but it's both live nodes and aliases, which is awkward.  I'd be okay with that as well, if we think aliases might disappear as an explicit cacheable concern here.

aparnasuresh85: The timeout was originally specified in seconds, and the elapsed time—used for comparison with the timeout value for livenodes and aliases—is also interpreted in seconds. Therefore, I set the default value to 5 seconds. Let me know if the interpretation and the cachetimeout value needs to be changed to milliseconds.

dsmiley: Solr has been sloppy in time units but there have been recent improvements on standardization.  So I have no doubts you've seen seconds here & there.  Elsewhere I see a time suffix so it's not ambiguous -- which is ideal.  I recommend adding a ".sec".

aparnasuresh85: I updated the env variable name in the latest PR

dsmiley (approved):
Looks good to me.  Doesn't need a JIRA as it's too simple.  I'll merge this Monday if I don't get more feedback.

PR #2632, opened 2024-08-09
remove deprecated (DocValues,Norms)FieldExistsQuery use

cpoerschke:
for `main` and `branch_9x` only, no JIRA or solr/CHANGES.txt entry needed in my opinion.


solr/core/src/java/org/apache/solr/search/facet/MissingAgg.java
```diff
@@ -45,7 +45,7 @@ public SlotAcc createSlotAcc(FacetContext fcontext, long numDocs, int numSlots)
       if (sf.multiValued() || sf.getType().multiValuedFieldCache()) {
         Query query = null;
         if (sf.hasDocValues()) {
-          query = new DocValuesFieldExistsQuery(sf.getName());
+          query = new FieldExistsQuery(sf.getName());
         } else {
           query = sf.getType().getRangeQuery(null, sf, null, null, false, false);
         }
```
dsmiley, line 51: I suspect this code pre-dated FieldType.getExistenceQuery -- just call that.

dsmiley (changes requested):
As there is now FieldExistsQuery covering a range of cases (not just even docValues & norms), this probably obsoletes complexity inside FieldType.getExistenceQuery.  Can we just call that and remove getSpecializedExistenceQuery as needless in lieu of subtypes overriding getExistenceQuery?

CC @HoustonPutman as you worked on this method

solr/core/src/java/org/apache/solr/search/facet/MissingAgg.java
```diff
@@ -45,7 +45,7 @@ public SlotAcc createSlotAcc(FacetContext fcontext, long numDocs, int numSlots)
       if (sf.multiValued() || sf.getType().multiValuedFieldCache()) {
         Query query = null;
         if (sf.hasDocValues()) {
-          query = new DocValuesFieldExistsQuery(sf.getName());
+          query = new FieldExistsQuery(sf.getName());
```
cpoerschke, line 48: > I suspect this code pre-dated FieldType.getExistenceQuery -- just call that.

```suggestion
          query = sf.getType().getExistenceQuery(null, sf);
```

cpoerschke:
> As there is now FieldExistsQuery covering a range of cases (not just even docValues & norms), this probably obsoletes complexity inside FieldType.getExistenceQuery. Can we just call that and remove getSpecializedExistenceQuery as needless in lieu of subtypes overriding getExistenceQuery?
> 
> CC @HoustonPutman as you worked on this method

Personally I don't have bandwidth to pursue this, but perhaps others do, "allow edits by maintainers" is enabled for the PR i.e. feel free to push commits to the branch.

And/Or could `getSpecializedExistenceQuery` removal be done separately after this pull request?

cpoerschke:
Added @chatman and @noblepaul w.r.t. overlap with the #3053 upgrade to Lucene 10.

HoustonPutman:
> As there is now FieldExistsQuery covering a range of cases (not just even docValues & norms), this probably obsoletes complexity inside FieldType.getExistenceQuery. Can we just call that and remove getSpecializedExistenceQuery as needless in lieu of subtypes overriding getExistenceQuery?
> 
> CC @HoustonPutman as you worked on this method

It looks like the range of cases are just docValues, norms and vectors. Not sure we can get rid of much of the complexity. `getSpecializedExistenceQuery` is still necessary for double/float fields without docValues or norms. and Spatial fields. I think this just moves the three query classes into one class.

solr/core/src/java/org/apache/solr/search/facet/MissingAgg.java
HoustonPutman: Did this whenever I could

solr/core/src/java/org/apache/solr/schema/DenseVectorField.java
```diff
@@ -394,7 +394,7 @@ public Query getFieldQuery(QParser parser, SchemaField field, String externalVal
 
   /** Not Supported */
```
cpoerschke, line 395: Wondering if we might wanna
```
@Override
protected boolean treatUnboundedRangeAsExistence(SchemaField field) {
    return false;
}
```
to retain the behaviour of `getRangeQuery` throwing an exception.

cpoerschke: Answering own question: it already returns `false` via this implementation in the class hierarchy: https://github.com/apache/solr/blob/releases/solr/9.7.0/solr/core/src/java/org/apache/solr/schema/NumericFieldType.java#L360

HoustonPutman:
Ok I've undone the pointField stuff, as the only clean way is to throw an error to users. So let's do https://issues.apache.org/jira/browse/SOLR-14199 completely separately.

solr/core/src/java/org/apache/solr/schema/CurrencyFieldType.java
```diff
@@ -387,7 +386,7 @@ private Query getRangeQueryInternal(
         (p1 != null) ? p1.getCurrencyCode() : (p2 != null) ? p2.getCurrencyCode() : defaultCurrency;
 
     // ValueSourceRangeFilter doesn't check exists(), so we have to
-    final Query docsWithValues = new DocValuesFieldExistsQuery(getAmountField(field).getName());
+    final Query docsWithValues = getAmountField(field).getType().getExistenceQuery(parser, field);
```
cpoerschke, line 389: Able to reproduce the `./gradlew test --tests CurrencyFieldTypeTest` test failure locally, shall we go back to `FieldExistsQuery` here maybe, or perhaps the test needs revising (haven't looked).
```suggestion
    final Query docsWithValues = new FieldExistsQuery(getAmountField(field).getName());
```

HoustonPutman: Just debugged this, stupid error. Should be passing the amountField into the method instead of the currencyField.

cpoerschke:
Just noting retrospectively that my intention here had been to (slightly) reduce the scope of the #3053 changes (and for `branch_9x` to get the changes in this PR too) but then it sort of dropped off the radar somehow, sorry. Now no-longer-applicable since PR 3053 is merged.

PR #2634, opened 2024-08-10
Mark AccessControlContext, AccessController, and SecurityManager as forbidden

iamsanjay:

Updated code to mark `SecurityManager`, `AccessControlContext`, and `SecurityManager` as forbidden to align with the upcoming Java version.

Although I added a redundant @SuppressWarnings("removal"), which will be necessary when we upgrade the Java version, it is currently not having any effect. 

# Checklist


iamsanjay:
For some reason, forbiddenApis didn't flag DocumentObjectBinder from solrj, even though it uses AccessController.

dsmiley:
@uschindler I'm a little unclear on how a project should juggle Java warnings & forbiddenapis.  I might argue that if javac is going to give a warning, then wouldn't updating forbiddenapis config be rather redundant?  Maybe the latter compells/blocks you -- must at least add an annotation, whereas a javac warning isn't blocking?  But I recall we can configure it to be.

iamsanjay:
I'll share my thought process on how I used forbidden APIs. Initially, I added entries to the config files while we were still using Java 11. Once we move to Java 21, there will be no need to include those in the forbiddenapis config file, as it will automatically detects them. So why did I add them now? To ensure that if someone uses these "deprecated" APIs, they receive a warning, and the plugin forces them to either handle it differently or apply annotations.

Additionally, the Java compiler only fails the build if an API is marked for "removal," which is why I used two different annotations: one to suppress Java warnings and another for forbidden APIs. Some might question the need for both. I think it conveys the "risk factor." If an API is only deprecated, the forbidden annotation will do the job, but if it's marked for removal, another annotation is necessary.

uschindler:
Hi,
in my opinion the whole thing is not needed at all. On the day when you switch to Java 17 or 21 as minimum version you get the forbiddenwarnings automatically (by both forbiddenapis and javac). Just fix them at that time. Due to syntax changes you may still need to

Come on, we are talking here about 3 source files with SecurityManager. I think one can easily fix the issues at that time. But anyways, you can commit this PR as it works in the Apache Lucene style (separating the accesscontroller into a separate method) and add all annotations. But addings the forbidden signatures can be avoided as they get obsolete anyways.

> For some reason, forbiddenApis didn't flag DocumentObjectBinder from solrj, even though it uses AccessController.

The method is already annotated with @SuppressForbidden, so the refactoring to a separate method is not helping anything. You can revert that file: https://github.com/apache/solr/blob/21359aa8dac184525630fc4177a71060e7b3b4e3/solr/solrj/src/java/org/apache/solr/client/solrj/beans/DocumentObjectBinder.java#L139

uschindler:
> I'll share my thought process on how I used forbidden APIs. Initially, I added entries to the config files while we were still using Java 11. Once we move to Java 21, there will be no need to include those in the forbiddenapis config file, as it will automatically detects them. So why did I add them now? To ensure that if someone uses these "deprecated" APIs, they receive a warning, and the plugin forces them to either handle it differently or apply annotations.
> 
> Additionally, the Java compiler only fails the build if an API is marked for "removal," which is why I used two different annotations: one to suppress Java warnings and another for forbidden APIs. Some might question the need for both. I think it conveys the "risk factor." If an API is only deprecated, the forbidden annotation will do the job, but if it's marked for removal, another annotation is necessary.

Forbiddenapis automatically fails in Java 17 on *all* deprecated methods inside the JDK. You can make this a fatal error in javac, too, but this would prevent you from consuming your own deprecated APIs outside of the JDK. Therefore, forbiddenapis has "jdk-deprecated" bundled signatures, which fails the build.

If you want to check today for deprecated methods in java 17, you could temporarily change the forbiddenapis configuration to use "jdk-deprecated-17" instead of "jdk-deprecated" (which gets version of javac automatically): https://github.com/apache/solr/blob/e5051e8bed85646f268cb93af9305c0842e02c71/gradle/validation/forbidden-apis.gradle#L63

uschindler (changes requested):
see my previous comments. I would not do this at all.

iamsanjay:
We’ll drop this one for now and revisit it when we're about to upgrade. Thanks for your help @uschindler, and also for clearing up the confusion about the `DocumentObjectBinder`. Appreciate it!

PR #2645, opened 2024-08-14
SOLR-17408 - COLSTATUS is not optimized

mariemat:
https://issues.apache.org/jira/browse/SOLR-17408

# Checklist


solr/core/src/java/org/apache/solr/handler/admin/ColStatus.java
```diff
@@ -93,8 +93,14 @@ public void getColStatus(NamedList<Object> results) {
     if (withRawSizeSummary || withRawSizeDetails) {
       withRawSizeInfo = true;
     }
-    if (withFieldInfo || withSizeInfo) {
-      withSegments = true;
+    boolean getSegments = false;
```
epugh, line 96: maybe kind of bikeshedding, but what about `requireSegments`?   Or `segmentsNeeded`?   `getSegments` kind of looks like a method to me!

epugh (commented):
Does this change require a a new unit test?   Or do the existing tests cover this?

Seems like a reasonable optimization!

mariemat:
> Does this change require a a new unit test? Or do the existing tests cover this?
Previous unit tests cover it

> Seems like a reasonable optimization!
👍

solr/core/src/java/org/apache/solr/handler/admin/ColStatus.java
dsmiley: FWIW I think `getSegments` is named fine, but so are your suggestions.  I don't think I've ever confused a method for a variable/field because the syntax involved would be nonsensical if reversed.  The line commented on here would unfathomably refer to a method (`type varName = value`).  Lines of code referencing the var won't have calling parenthesis so wouldn't be a method call.  And would be used in boolean ways (`if (getSegments)`) not a more interesting structure.

solr/core/src/java/org/apache/solr/handler/admin/ColStatus.java
```diff
@@ -176,48 +182,50 @@ public void getColStatus(NamedList<Object> results) {
...
-              if (fields != null) {
-                for (Map.Entry<String, Object> fEntry : fields) {
-                  Object nc = ((NamedList<Object>) fEntry.getValue()).get("nonCompliant");
-                  if (nc != null) {
-                    nonCompliant.add(fEntry.getKey());
+        if (getSegments) {
+          try (SolrClient client = solrClientCache.getHttpSolrClient(url)) {
+            ModifiableSolrParams params = new ModifiableSolrParams();
+            params.add(CommonParams.QT, "/admin/segments");
+            params.add(FIELD_INFO_PROP, "true");
```
dsmiley, line 189: curious to see "true" here instead of withFieldInfo.  I know this isn't a change you brought about but it's in scope to either fix or comment justifying.

dsmiley:
There's no special V2 things needed in this PR. COLSTATUS is a case where the V2 API (original V2, not JAX-RS) is calling the V1 API, even though our plans call for the reverse: https://github.com/apache/solr/blob/c99af207c761ec34812ef1cc3054eb2804b7448b/solr/core/src/java/org/apache/solr/handler/admin/api/CollectionStatusAPI.java

solr/core/src/java/org/apache/solr/handler/admin/ColStatus.java
mariemat: I am not confident enough to make that change, given I am unaware of how people use it. I prefer to leave it as-is in that PR which is not related to your point.

mariemat:
The failed test is a flapper and is not related to my changes.

solr/core/src/java/org/apache/solr/handler/admin/ColStatus.java
dsmiley: Okay @mariemat ; I'll leave this be, despite it looking wrong.

@sigram you wrote it this way so if you can offer any input, now's a good time.  I looked at SOLR-13292 saw no review on this topic either.

PR #2652, opened 2024-08-16
SOLR-17412: Skip nested atomic updates for non-existent child docs

timatbw:
https://issues.apache.org/jira/browse/SOLR-17412

# Description

Allow updates that are atomic updates containing nested atomic updates to quietly skip any child docs that don't exist, rather than throw an exception for the whole update

# Solution

Now uses a request parameter (the same one used for very similar logic in `SkipExistingDocumentsUpdateProcessor`) as a flag, changing the behaviour to ignore the atomic update for a child doc when that doc is not found.

# Tests

Added a test that confirms the same original doc and update doc that throw an exception, when called using the new parameter set, do not throw an exception or create a new child doc.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
+++ b/solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
@@ -69,10 +69,12 @@ public class AtomicUpdateDocumentMerger {
 
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
+  protected final SolrQueryRequest request;
   protected final IndexSchema schema;
   protected final SchemaField idField;
 
   public AtomicUpdateDocumentMerger(SolrQueryRequest queryReq) {
+    request = queryReq;
     schema = queryReq.getSchema();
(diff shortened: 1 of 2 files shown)


timatbw:
Possibly, I will have a look through the docs and see what is said about atomic updates and child docs, especially the nested atomic updates which are relatively new.

solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
```diff
@@ -69,10 +69,12 @@ public class AtomicUpdateDocumentMerger {
 
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
+  protected final SolrQueryRequest request;
   protected final IndexSchema schema;
   protected final SchemaField idField;
 
   public AtomicUpdateDocumentMerger(SolrQueryRequest queryReq) {
+    request = queryReq;
```
cpoerschke, line 77: Maybe no need to retain the full request?
```suggestion
    skipUpdateIfMissing = queryReq.getParams().getBool("skipUpdateIfMissing", false);
```

PR #2661, opened 2024-08-23
HttpShardHandler: Fix extension hook transformResponse

dsmiley:
HttpShardHandler.transformResponse is an extension hook for subclasses to transform the shard response.  Ever since 9.0 when the HTTP2 Async stuff landed, this hook became inoperable/dormant (method existed but hasn't been called anymore).  This brings it back.

It's sufficiently minor that I'm thinking no JIRA/CHANGES.txt

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
@@ -168,7 +168,7 @@ public class HttpShardHandler extends ShardHandler {
             srsp.setShardAddress(rsp.getServer());
             ssr.elapsedTime =
                 TimeUnit.MILLISECONDS.convert(System.nanoTime() - startTime, TimeUnit.NANOSECONDS);
-            responses.add(srsp);
+            responses.add(transfomResponse(sreq, srsp, shard));
           } else if (throwable != null) {
             ssr.elapsedTime =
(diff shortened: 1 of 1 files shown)


mkhludnev (approved):

gus-asf:
@gerlowskija This seems to have landed a short time before your Parallel handler, but is not reflected in that implementation. Was that just an oversight due to near simultaneous development or a conscious omission due to some concern? The callback for both handlers is otherwise the same and maybe could be unified for re-use...

gus-asf:
Also the spelling error in the method should be fixed... transfomResponse -> transformResponse. I suppose that has to be a 10x only change since it's an extension point.

gus-asf:
> Also the spelling error in the method should be fixed... transfomResponse -> transformResponse. I suppose that has to be a 10x only change since it's an extension point.

(I can do both of these this in a separate ticket after I push my changes for SOLR-17158 if noone objects ) The rename should have an associated Jira I think.

dsmiley:
Renaming a protected method doesn't deserve a JIRA and wasting readers time to read about it in CHANGES.txt.

fsparv:
If users have implemented a custom handler (as is clearly implied via the doc on the method, and it's no-op default implementation) their code would need to be updated and rebuilt. Seems like something someone assessing an upgrade effort would want to know about...?

dsmiley:
it's about a trivial change as changes go.  Users (and I count myself as impacted at work BTW) would find out upon upgrading.  Tons of other changes happen without CHANGES.txt prescribing every little detail as the name of a method.

PR #2662, opened 2024-08-23
remove use of deprecated SolrIndexSearcher.search variant

cpoerschke:
#2524 deprecated the variant

### DIFF
--- a/solr/core/src/test/org/apache/solr/search/SolrIndexSearcherTest.java
+++ b/solr/core/src/test/org/apache/solr/search/SolrIndexSearcherTest.java
@@ -120,16 +120,14 @@ public class SolrIndexSearcherTest extends SolrTestCaseJ4 {
 
   private void assertMatchesEqual(int expectedCount, SolrIndexSearcher searcher, QueryCommand cmd)
       throws IOException {
-    QueryResult qr = new QueryResult();
-    searcher.search(qr, cmd);
+    QueryResult qr = searcher.search(cmd);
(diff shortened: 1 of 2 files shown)


cpoerschke:
https://github.com/apache/solr/actions/runs/10529517119/job/29177462369?pr=2662 CI failure seems unrelated

```
./gradlew :solr:core:test --tests "org.apache.solr.cloud.api.collections.LocalFSCloudIncrementalBackupTest.testSkipConfigset" -Ptests.jvms=96 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=92C1489090FDEECD -Ptests.timeoutSuite=600000! -Ptests.file.encoding=US-ASCII
```

epugh:
Will we be able to remove the deprecated `QueryResult search(QueryResult qr, QueryCommand cmd)` method in `main`?   Do we every label something Deprecated but don't mean to remove it for a long period of time?

dsmiley (approved):
> Will we be able to remove the deprecated QueryResult search(QueryResult qr, QueryCommand cmd) method in main?

Yes.

> Do we ever label something Deprecated but don't mean to remove it for a long period of time?

Yes but that's atypical.  I've done it hehe; not always intentional.  You add some new mechanism and deprecate the old but the old way is entangled (or is used a ton), so we have both.  That's not here though!

PR #2467, opened 2024-05-18
SOLR-17300: Copy existing listeners on re-creation of Http2SolrClient

iamsanjay:
https://issues.apache.org/jira/browse/SOLR-17300

Instead of #2462, this PR would copy the existing listeners to the new Http2SolrClient. This way re-creating client not requires to call explicitly `Builder#withListenerFactory` to copy existing listeners as `withHttpClient` method automatically  takes care of that. 

Added #2462 test case here.

# Checklist

Please review the following and check all that apply:

- [x] I have reviewed the guidelines for [How to Contribute](https://github.com/apache/solr/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.
- [x] I have created a Jira issue and added the issue ID to my pull request title.
- [x] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [x] I have developed this patch against the `main` branch.
- [ ] I have run `./gradlew check`.
- [x] I have added tests for my changes.
- [ ] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

PR #2665, opened 2024-08-25
Enabling test-retry-gradle-plugin

iamsanjay:
The test-retry mechanism allows failed test cases to be automatically re-run. It also includes functionality to mark the build as "successful" if the tests pass on re-run, though we’re not focusing on that feature at this time. Our primary goal in enabling this mechanism for Develocity is to classify tests as deterministic or non-deterministic. This classification will provide developers with better insights into failed test cases and help them make more informed decisions.

Additionally, it saves time by automating the re-running of failed test cases, so developers don’t have to manually retry them.

```
 retry {
        maxRetries = 3
        maxFailures = 20
        failOnPassedAfterRetry = false
      }
```

I believe the retry-test functionality is already integrated into the Gradle Enterprise plugin, which our project is currently using. Therefore, we don't need to add an explicit plugin to enable this feature.

https://github.com/apache/solr/pull/2467#issuecomment-2122297638

### DIFF
--- a/gradle/testing/defaults-tests.gradle
+++ b/gradle/testing/defaults-tests.gradle
@@ -181,9 +181,9 @@ allprojects {
         // Print some diagnostics about locations used.
         logger.info("Test folders for {}: cwd={}, tmp={}", project.path, testsCwd, testsTmpDir)
       }
-      if(isCIBuild) {
+      if (isCIBuild) {
         retry {
-          maxRetries = 20
+          maxRetries = 3
           maxFailures = 20
           failOnPassedAfterRetry = false
         }
--- a/gradle/testing/defaults-tests.gradle
+++ b/gradle/testing/defaults-tests.gradle
@@ -181,14 +181,6 @@ allprojects {
         // Print some diagnostics about locations used.
         logger.info("Test folders for {}: cwd={}, tmp={}", project.path, testsCwd, testsTmpDir)
       }
-      if (isCIBuild) {
-        // For info on params: https://github.com/gradle/test-retry-gradle-plugin
-        retry {
-          maxRetries = 3
-          maxFailures = 10
-          failOnPassedAfterRetry = true
-        }
-      }
     }
   }
 }
--- a/gradle/testing/defaults-tests.gradle
+++ b/gradle/testing/defaults-tests.gradle
@@ -181,13 +181,13 @@ allprojects {
         // Print some diagnostics about locations used.
         logger.info("Test folders for {}: cwd={}, tmp={}", project.path, testsCwd, testsTmpDir)
       }
-
-      retry {
-        maxRetries = 3
-        maxFailures = 20
-        failOnPassedAfterRetry = false
+      if(isCIBuild) {
+        retry {
+          maxRetries = 20
+          maxFailures = 20
+          failOnPassedAfterRetry = false
+        }
       }
-
     }
   }
 }
--- a/gradle/testing/defaults-tests.gradle
+++ b/gradle/testing/defaults-tests.gradle
@@ -181,6 +181,13 @@ allprojects {
         // Print some diagnostics about locations used.
         logger.info("Test folders for {}: cwd={}, tmp={}", project.path, testsCwd, testsTmpDir)
       }
+
+      retry {
+        maxRetries = 3
+        maxFailures = 20
+        failOnPassedAfterRetry = false
+      }
+
     }
   }
 }
--- a/gradle/testing/defaults-tests.gradle
+++ b/gradle/testing/defaults-tests.gradle
@@ -182,10 +182,11 @@ allprojects {
         logger.info("Test folders for {}: cwd={}, tmp={}", project.path, testsCwd, testsTmpDir)
       }
       if (isCIBuild) {
+        // For info on params: https://github.com/gradle/test-retry-gradle-plugin
         retry {
           maxRetries = 3
-          maxFailures = 20
-          failOnPassedAfterRetry = false
+          maxFailures = 10
+          failOnPassedAfterRetry = true
         }
       }
     }
--- a/build.gradle
+++ b/build.gradle
@@ -132,6 +132,7 @@ apply from: file('gradle/testing/fail-on-no-tests.gradle')
 apply from: file('gradle/testing/fail-on-unsupported-jdk.gradle')
 apply from: file('gradle/testing/alternative-jdk-support.gradle')
 apply from: file('gradle/java/jar-manifest.gradle')
+apply from: file('gradle/testing/retry-test.gradle')
 
 // Publishing and releasing
 apply from: file('gradle/maven/defaults-maven.gradle')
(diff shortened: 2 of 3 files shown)


dweiss:
I'm not sure how this will affect random test execution (whether it's the same main seed the tests are executed with). You'd have to look into that.

iamsanjay:
Thanks @dweiss for pointing that out. To test that, I added below code to one of the test case.

```
public void testFailure() {
    if(random().nextBoolean()) {
      assertFalse(true);
    }
  }
```


```
retry {
        maxRetries = 20
        maxFailures = 20
        failOnPassedAfterRetry = false
      }
```

If the same seed wasn't used, or if there was no seed at all, any retry could set nextBoolean to false, causing the test to pass. However, I observed that once the test failed, it kept failing 20 times more and stop there because it couldn't go beyond that limit, as maxFailures was configured. I've also attached the test output where you can see the same seed repeating!

[TEST-org.apache.solr.util.TestSolrVersion.txt](https://github.com/user-attachments/files/16748711/TEST-org.apache.solr.util.TestSolrVersion.txt)

dweiss:
Ok, so it restarts with the same master seed. Good to know.

gradle/testing/defaults-tests.gradle
```diff
@@ -181,6 +181,13 @@ allprojects {
         // Print some diagnostics about locations used.
         logger.info("Test folders for {}: cwd={}, tmp={}", project.path, testsCwd, testsTmpDir)
       }
+      if(isCIBuild) {
```
dsmiley, line 184: We have yet to apply the Google Java Style format to build files (we should do this for Solr 10!) but please try to adhere to it.
```suggestion
      if (isCIBuild) {
```

gradle/testing/defaults-tests.gradle
```diff
@@ -181,6 +181,13 @@ allprojects {
...
+        retry {
+          maxRetries = 20
+          maxFailures = 20
```
dsmiley, line 187: I read the docs on these two.
maxFailures -- maybe 10 instead?  If there are 11 failures or more, I'd think something is broadly wrong and we shouldn't do retries.  
maxRetries -- 20 seems like a lot; do we really want to rerun a test as many as 20 times?  I'd say 5 is plenty.

dsmiley (commented):
Thanks for doing this!

I expected to see the need to include a build plugin.  I guess this is now built-in?

iamsanjay:
> I expected to see the need to include a build plugin. I guess this is now built-in?

Yes this one does confused me in the start, as I removed the plugin declaration and retry was still working. Then I start looking into all the plugin in `build.gradle` one by one, hoping that it is coming with one of them. Ultimately, I discovered below dependencies in `setting.gradle` and explore it further to find out that the gradle enterprise does includes the 'test-retry' functionality as well. 

```
plugins {
    id 'com.gradle.enterprise' version '3.15.1'
    id 'com.gradle.common-custom-user-data-gradle-plugin' version '1.12'
}
```

I asked the same question at the Github and got this answer.

> @iamsanjay , as you figured out, the Develocity/Gradle Enterprise plugin bundles the retry plugin. If you are using it, then I'd recommend to remove the explicit dependency on the retry plugin. The extensions configured and their functionality are equivalent

https://github.com/gradle/gradle/blob/6586eb029f216bf8acfbea7bcfcb7cde8518a689/build-logic-commons/gradle-plugin/build.gradle.kts#L29

gradle/testing/defaults-tests.gradle
iamsanjay: Ohh did I commit `maxRetries = 20`. No It should not be that high, I believe I have committed the setting that I was testing for the randomization functionality. I think `maxRetries = 3` is alright.

iamsanjay: What I initially thought was this?

The maxFailure setting acts as a cumulative failure limit. If it's set to 10, and you have 5 test cases  fail, in the worst case scenario where retries also fail, it might retry the first 3 test cases 3 times, the 4th one only once, and skip the 5th one since the maxFailure limit would be reached. And I have seen around 4-5 test cases failing the recent build. I thought It would be great to give all the test cases a chance. With 20, we could re-run, in worst case, 7 tests. 

But I am also fine with 10 as well.

iamsanjay:
One test case failed! 

```
  - org.apache.solr.search.TestStressReorder.testStressReorderVersions (:solr:core)
    Test history: https://ge.apache.org/scans/tests?search.rootProjectNames=solr-root&tests.container=org.apache.solr.search.TestStressReorder&tests.test=testStressReorderVersions http://fucit.org/solr-jenkins-reports/history-trend-of-recent-failures.html#series/org.apache.solr.search.TestStressReorder.testStressReorderVersions
    Test output: /tmp/src/solr/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.search.TestStressReorder.txt
    Reproduce with: ./gradlew :solr:core:test --tests "org.apache.solr.search.TestStressReorder.testStressReorderVersions" -Ptests.jvms=96 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=576E4BEF513F4B02 -Ptests.timeoutSuite=600000! -Ptests.file.encoding=ISO-8859-1
```
It failed once before as well!

Screenshot 2024-08-28 at 9 59 48 AM

dsmiley (approved):
There's a known issue in JIRA assigned to Ishan to fix java.util.concurrent.RejectedExecutionException being thrown from some tests -- TestStressReorder tends to be one to do it.

Can you add a commented link to documentation for this configuration block so anyone can understand the params?  I thought I read a different definition of maxRetries.

I'm excited to see how this will affect our builds!  Merge at your convenience.

gradle/testing/defaults-tests.gradle
iamsanjay: You're correct. The maxRetries setting applies to each round, and if the total failures in that round exceed the limit, no retries are attempted. I retested with multiple tests, and this confirmed it, clearing up any doubts. I think 10 is perfect then!

iamsanjay:
I have another question regarding failOnPassedAfterRetry. Right now, it’s set to false, meaning the build passes if failed test cases succeed on retry. However, that’s not what I intended—I actually want the build to fail in such cases. I’m not planning to enable this just yet. So I disabled it.

iamsanjay:
Few more changes!

1. Updated API: Changed retry to develocity.testRetry due to deprecation.
2. Separated Configuration: Moved the retry configuration to a separate file for better organization.

dsmiley (approved):
+1 to merge -- thanks again.

iamsanjay:
https://ge.apache.org/s/qcptlo6laqo5g#info 
The build scan above is associated with branch_9_7, and the retry logic seems to have been introduced shortly after the branch cut-off.

dsmiley:
Ah; thanks Sanjay!

PR #2666, opened 2024-08-26
SOLR-17158 Terminate distributed processing quickly when query limit is reached

gus-asf:
In the case where partial results are not desirable, there is no reason to await any remaining requests, or to return any partial results. This change causes the &partialResults=false query parameter to short circuit processing and omit any results. This condition is indicated with a partialResults="omitted" value in the response (prior behavior was a value of true or absent)

https://issues.apache.org/jira/browse/SOLR-17158

This PR replaces https://github.com/apache/solr/pull/2379 which had become onerously stale.


gus-asf:
looking back at https://github.com/apache/solr/pull/2493 I'm exploring if there are ways to also shorten the case where partial results are desired, but possibly due to super short CpuTimeAllowed in some tests, I'm occasionally (but not consistently) hitting cases where response builder hasn't done any work yet so various bits (such as requestIds) are null or otherwise uninitialized when we get to handleRequestBody. Exactly why is unclear, but some bits of responsebuilder maybe should be initialized by default. (i.e requestIds can be set to Collections.emptyMap() initially, since dozens of places don't check it for null)

solr/CHANGES.txt
```diff
@@ -109,8 +109,11 @@ New Features
 
 Improvements
 ---------------------
+* SOLR-17158: Users for whom partial results are uninteresting may set partialResultsAllowed=false. This allows Solr to reduce time spent processing partial results and omit them from the response. (Gus Heck, Andrzej Bialeki, hossman)
```
sigram, line 112: Typo - my last name is Bialecki :) or actually Białecki where non-ascii is allowed, but I've been using both forms.

solr/core/src/java/org/apache/solr/response/SolrQueryResponse.java
```diff
@@ -61,7 +64,7 @@
  */
 public class SolrQueryResponse {
   public static final String NAME = "response";
-  public static final String RESPONSE_HEADER_PARTIAL_RESULTS_KEY = CommonParams.PARTIAL_RESULTS;
```
sigram, line 64: Originally I did this on purpose to avoid getting these two strings out of sync.

solr/core/src/java/org/apache/solr/response/SolrQueryResponse.java
```diff
@@ -139,30 +142,63 @@ public ReturnFields getReturnFields() {
 
   /**
    * If {@link #getResponseHeader()} is available, set {@link #RESPONSE_HEADER_PARTIAL_RESULTS_KEY}
-   * flag to true.
+   * attribute to true or "omitted" as required.
    */
-  public void setPartialResults() {
+  public void setPartialResults(SolrQueryRequest req) {
```
sigram, line 147: Should this be `static` now?

solr/core/src/java/org/apache/solr/response/SolrQueryResponse.java
```diff
@@ -139,30 +142,63 @@ public ReturnFields getReturnFields() {
...
+    if (header != null) {
+      if (header.get(SolrQueryResponse.RESPONSE_HEADER_PARTIAL_RESULTS_KEY) == null) {
+        Object value = partialResultsStatus(shouldDiscardPartials(req.getParams()));
+        header.add(SolrQueryResponse.RESPONSE_HEADER_PARTIAL_RESULTS_KEY, value);
+      }
     }
   }
 
+  public static Object partialResultsStatus(boolean discarding) {
+    return discarding ? "omitted" : Boolean.TRUE;
```
sigram, line 158: Should this be a constant somewhere?

solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-9.adoc
```diff
@@ -71,6 +71,13 @@ Due to changes in Lucene 9, that isn't possible any more.
 === Configuration
 In solrconfig.xml, the `numVersionBuckets` and `versionBucketLockTimeoutMs` settings are now obsolete and ignored; a warning will be logged if specified.
 
+=== Partial Results
+When solr is used for reporting, legal discovery or other use cases that cannot accept partial results, yet it still needs to be protected from long-running queries, users may now opt out of receiving partial results.
```
sigram, line 75: long-running queries -> queries exceeding limits ?

solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-9.adoc
```diff
@@ -71,6 +71,13 @@ Due to changes in Lucene 9, that isn't possible any more.
...
+This feature should eliminate some wasted processing and bandwidth, as well as simplifying client side logic, and reducing the chance of inappropriate responses to the user.
+This can be enabled either via a request parameter (`allowPartialResults`), environment variable or system property.
+The partialResults attribute in the request header will now have three possible states: `"true"`, `"omitted"` or absent, with the new `"omitted"` value indicating the difference between a timed out request with results omitted due to query limits and results that are merely empty because the query didn't match any docs.
```
sigram, line 78: Somehow this sentence is very confusing to me ... maybe this: the new "omitted" value indicating that query returned partial results, which due to the setting were not allowed and they were then discarded? or something more English-like ;)

solr/core/src/java/org/apache/solr/response/SolrQueryResponse.java
fsparv: Right but they aren't actually the same string. One is a request parameter and the other is a response header key. The fact they are textually similar is a coincidence (and packages now make this difficult)

solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-9.adoc
gus-asf: > long-running queries -> queries exceeding limits ?

yes... this could be clearer, will wordsmith.

solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-9.adoc
gus-asf: Yeah probably needs to be broken into 2 sentences at least.

solr/core/src/java/org/apache/solr/response/SolrQueryResponse.java
gus-asf: It could be. Though it only gets used here and in one assert in one test right now.

solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-9.adoc
sigram: Hmm, maybe using "discarded" instead of "omitted", both in the docs AND in the response header, would make it clearer?

solr/core/src/java/org/apache/solr/response/SolrQueryResponse.java
gus-asf: There's not really a point to making it static, the object already exists in every location it's used.

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
```diff
@@ -62,7 +69,14 @@ public class HttpShardHandler extends ShardHandler {
   private HttpShardHandlerFactory httpShardHandlerFactory;
   private Map<ShardResponse, CompletableFuture<LBSolrClient.Rsp>> responseFutureMap;
   private BlockingQueue<ShardResponse> responses;
+
+  /**
+   * The number of pending requests. This must be incremented before a {@link ShardResponse} is
+   * added to {@link #responses}, and decremented after a ShardResponse is removed from {@code
+   * responses}. We cannot rely on responses.size() bec
```
dsmiley, line 76: missing explanation

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
```diff
@@ -42,13 +45,17 @@
...
 import org.apache.solr.common.util.NamedList;
+import org.apache.solr.common.util.StrUtils;
 import org.apache.solr.core.CoreDescriptor;
 import org.apache.solr.request.SolrQueryRequest;
 import org.apache.solr.request.SolrRequestInfo;
 import org.apache.solr.security.AllowListUrlChecker;
 
 @NotThreadSafe
 public class HttpShardHandler extends ShardHandler {
+  /** */
```
dsmiley, line 56: either say something or don't :-)

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
```diff
@@ -80,6 +94,39 @@ public HttpShardHandler(HttpShardHandlerFactory httpShardHandlerFactory) {
...
+   * </code>.
+   *
+   * <p>By default, returns <code>false</code> when {@value ShardParams#SHARDS_TOLERANT} is not set
+   * in <code>
+   * params</code>.
+   */
+  public static boolean getShardsTolerantAsBool(SolrQueryRequest req) {
+    String shardsTolerantValue = req.getParams().get(ShardParams.SHARDS_TOLERANT);
+    if (null == shardsTolerantValue
+        || shardsTolerantValue.trim().equals(ShardParams.REQUIRE_ZK_CONNECTED)) {
```
dsmiley, line 109: we don't normally trim our params.  I think it's very haphazard to do it on-read (like once we do it here and there, then everywhere we wonder, should we do here too?  a mess IMO)

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
```diff
@@ -80,6 +94,39 @@ public HttpShardHandler(HttpShardHandlerFactory httpShardHandlerFactory) {
...
+      return false;
+    } else {
+      boolean tolerant = StrUtils.parseBool(shardsTolerantValue.trim());
+      if (tolerant && shouldDiscardPartials(req.getParams())) {
```
dsmiley, line 113: this word "discard" is new here (I think) and I find it confusing.  Maybe `shouldErrorInsteadOfPartialResults`

Or flip the boolean -- allowPartialResults

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
```diff
@@ -139,48 +186,54 @@ public void submit(
...
-      responses.add(srsp);
-      return;
-    }
-
-    long startTime = System.nanoTime();
-    SolrRequestInfo requestInfo = SolrRequestInfo.getRequestInfo();
-    if (requestInfo != null) {
-      req.setUserPrincipal(requestInfo.getReq().getUserPrincipal());
-    }
+      long startTime = System.nanoTime();
```
dsmiley, line 204: please encode time units into time var names, like an "Ns" suffix

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
```diff
@@ -191,7 +244,7 @@ protected QueryRequest makeQueryRequest(
   }
 
   /** Subclasses could modify the Response based on the shard */
-  protected ShardResponse transfomResponse(
+  protected ShardResponse transformResponse(
```
dsmiley, line 247: I merged a PR that corrects that this wasn't being called above -- 391cdd04277b83a010f1d33288f9bc344a7ba2d0

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
```diff
@@ -216,9 +269,18 @@ public ShardResponse takeCompletedOrError() {
 
   private ShardResponse take(boolean bailOnError) {
     try {
+      // although nothing in this class guarantees that pending has been incremented to the total
+      // number of expected requests, actual usage in SearchHandler results in this method never
+      // being called until all requests have been added in a prior loop over
+      // ShardRequest.actualShards in the same thread that invokes take() (I haven't checked but
+      // hopefully other handlers do the same) The net effect is we shouldn't arrive here with
```
dsmiley, line 276: Would an assert be useful?

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
```diff
@@ -62,7 +69,14 @@ public class HttpShardHandler extends ShardHandler {
   private HttpShardHandlerFactory httpShardHandlerFactory;
   private Map<ShardResponse, CompletableFuture<LBSolrClient.Rsp>> responseFutureMap;
```
dsmiley, line 70: It seems responseFutureMap is guarded by RESPONSE_CANCELABLE_LOCK; could use a comment to say so

solr/core/src/java/org/apache/solr/request/SolrQueryRequest.java
```diff
@@ -39,6 +40,36 @@
...
+   * true, and solr will omit results when any shard fails due query execution limits (time, cpu
+   * etc.). By default, this is set to true. Setting it to false will reduce processing, cpu and
+   * network associated with collecting and transmitting partial results. This setting can be
+   * overridden (in either direction) on a per-request basis with {@code
+   * &allowPartialResults=[true|false]}. When results have been omitted the response header should
+   * contain a partialResults element with the value "omitted"
+   */
+  boolean ALLOW_PARTIAL_RESULTS_DEFAULT =
+      System.getProperty(SOLR_ALLOW_PARTIAL_RESULTS_DEFAULT) == null
+          || Boolean.getBoolean(SOLR_ALLOW_PARTIAL_RESULTS_DEFAULT);
```
dsmiley, line 58: EnvUtils

solr/core/src/java/org/apache/solr/request/SolrRequestInfo.java
```diff
@@ -66,6 +67,22 @@ public static SolrRequestInfo getRequestInfo() {
...
 
+  public static Optional<SolrRequestInfo> getReqInfo() {
+    return Optional.ofNullable(getRequestInfo());
+  }
+
+  public static Optional<SolrQueryRequest> getRequest() {
+    return getReqInfo().map(i -> i.req);
+  }
+
+  public static boolean shouldDiscardPartials() {
```
dsmiley, line 78: this method feels out of place here

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
fsparv: https://issues.apache.org/jira/browse/SOLR-6572

dsmiley: SOLR-6572 is for XML, and had a different motivating use-case that doesn't apply here.

gus-asf: Sorry, let me be clearer. That .trim() was retained only because it was part of the (recent) changes for that ticket that I received in a merge from main: (https://github.com/apache/solr/pull/2311/files) If you are confident that the change made in SOLR-6572 was in error, we can remove it. I simply trusted the change made by @janhoy

solr/core/src/java/org/apache/solr/handler/component/ParallelHttpShardHandler.java
```diff
@@ -45,53 +48,84 @@ public class ParallelHttpShardHandler extends HttpShardHandler {
...
     final List<String> urls = getURLs(shard);
     final var lbReq = prepareLBRequest(sreq, shard, params, urls);
     final var srsp = prepareShardResponse(sreq, shard);
     final var ssr = new SimpleSolrResponse();
     srsp.setSolrResponse(ssr);
-    pending.incrementAndGet();
 
     if (urls.isEmpty()) {
       recordNoUrlShardResponse(srsp, shard);
       return;
```
sigram, line 78: Not sure if I fully understand the flow here ... wouldn't this throw off the balance of attemptStart / attemptCount (that we check in `responsesPending()`) because attemptStart has already been incremented here but the corresponding attemptCount will never be incremented?

solr/solr-ref-guide/modules/query-guide/pages/common-query-parameters.adoc
```diff
@@ -324,9 +324,20 @@ The default value of this parameter is blank, which causes no extra "explain inf
...
+If incomplete results are returned the `partialResults` response header will be set to `true`
+
+WARNING: These results may be incomplete in a non-deterministic way (e.g. only some matching documents, documents without fields, missing facets or pivots, no spellcheck results, etc).
+
+
+When this parameter is set to `false` then upon reaching a limit, Solr will stop collecting results and will not return any partial results already collected.
+In this case, the partialResults response header will be set to `omitted` and no result documents are returned.
+
+NOTE: The response will be 200 OK because the system has correctly provided the requested behavior.
+This is not an error condition.
```
sigram, line 339: I like this wording now :) it's clear to me what happens in each case.

gus-asf:
Also, I've decided to presume that the failure to call transform in the ParallelHttpShardHandler was an oversight.

gus-asf:
This is now feeling complete, I'll be looking to commit it tomorrow afternoon or perhaps Friday morning to main if nobody objects or requests more time to review. 

I've had to back off any changes that mimic https://github.com/apache/solr/pull/2493 

I had code for that, but I got lucky and hit a rare exception on one of my test runs, I found that ~1 in 20 times I was getting a CancellationException, and something less than 1 in 100 times (once it ran 800 iterations before failing) I was seeing a test fail where "omitted" was not properly reported with partialResults=false. I think I solved the CancellationException but ran out of time/energy to chase down whatever scenario was causing the rare test fail. Without that code (as the PR is now) I ran the CPU timeout test > 1500 times with no failures, so I'm confident that the problem is not transmitted in this PR, and additional enhancement can be done later. (I'll probably post a branch off of this branch showing what I had there for future reference).

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
janhoy: So, this param may come from http url param or it may come from default section on a request handler in solrcobfig. If the latter, the `true` string may contain white space if the xml is «indented».

We could not trim() all xml values, since sometimes white space is significant.

Perhaps a more readable code would be to not do plain string comparison here but use a utility method boolCompareTrim() or similar? 

I have not looked into the code base this time around, but I seem to remember there was no obvious clean location to normalize these bool-strings during config parse..

janhoy: I feel the pain too. A question would be, why do we not also trim `REQUIRE_ZK_CONNECTED`?

solr/core/src/java/org/apache/solr/handler/component/ParallelHttpShardHandler.java
```diff
@@ -41,57 +39,78 @@
...
 public class ParallelHttpShardHandler extends HttpShardHandler {
 
+  @SuppressWarnings("unused")
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   private final ExecutorService commExecutor;
 
+  /*
+   * Unlike the basic HttpShardHandler, this class allows us to exit submit before
+   * pending is incremented and the responseFutureMap is updated. If the runnables that
```
gerlowskija, line 49: [Q] Taking the "this class allows us to exit submit before pending is incremented" behavior as a given, the use of these additional AtomicInt's makes sense.  And the descriptive comment here is lovely.

But why does the PR introduce this complication in the first place?  Is it required by SOLR-17158?  Is it fixing a concurrency bug that pre-exists SOLR-17158?

gus-asf: The concurrency fixes here and changes in HttpShardHandler come from me needing to implement this and being worried about the effects of limits interrupting/canceling requests. I felt compelled to ensure that canceling futures from within the call back (to facilitate a faster short circuit once we know the request is hitting a limit) wouldn't cause concurrency issues. ... so I put effort into understanding exactly what's going on and what the controls are. I found that (before this ticket) there was reliance on the orderly update of pending, a map of futures and the request queue object. Converting the map to a synchronized object didn't seem sufficient since the logic relies on coordination among objects... so essentially I had to make sure everything was going to handle a new short circuit point initiated by the callback and concurrency became "on-topic"... 

At a high level this ticket works by letting the callbacks [signal to stop all other callbacks](https://github.com/apache/solr/pull/2666/files#diff-ce2580113ae9a9d50f308985229130f094868e83d35a8727c5683e6a2f3f44daR543) if one of them is failing and partial results are not desired. There's another opportunity to quit early in the take() method but that caused too many problems (rare test failures likely due to further introduced concurrency issues).

I do suspect some of this could be simplified, particularly what I note here: https://github.com/apache/solr/pull/2666/files#diff-ce2580113ae9a9d50f308985229130f094868e83d35a8727c5683e6a2f3f44daR98 

But I decided to only guard and de-duplicate the existing logic. I'm leaving revising the logic to a later effort in order to control scope. 

In the case of your class (which came in as a merge), this was the simplest thing I could think of (as in simple to understand, not simple/elegant). Basically we need to know that everything we scheduled has been attempted (fail or not) before we let the loop that takes responses exit or we might miss one, but we also have to ensure that a failure won't leave us looping forever. So the quick solution was to track what's started, and track what's been attempted, and make sure we don't quit waiting for responses unless they are equal.  Note that attempts may not result in a a future if `this.lbClient.requestAsync(lbReq);` fails, so the count of futures is not good enough.

dsmiley: Ooof, the additional complexity I'm reading about is concerning.  Without looking at this as closely as either of you, admittedly, I've naively wondered why doesn't HttpShardHandler use ExecutorCompletionService.  It's a really nice JDK utility that seems to do conceptually what HttpShardHandler is doing.  If it would be helpful, it would reduce the complexity of what this Solr class has to deal with.  Thoughts?

> we need to know that everything we scheduled has been attempted (fail or not) before we let the loop that takes responses exit or we might miss one

That doesn't sound right.  Meaning, I hope we can cancel if we haven't actually sent all requests.

solr/CHANGES.txt
gus-asf: uninteresting, undesirable, unacceptable, aren't acceptable .... any of those seem roughly equivalent to me. The point is the user doesn't want them. For example a legal discovery search that omits some results could cause a user to miss an important precedent and loose a case. They would set this parameter to avoid wasting bandwidth, and also forestall the possibility that badly coded UI could show incomplete results get used for an operation requiring complete results.

solr/core/src/java/org/apache/solr/handler/component/ParallelHttpShardHandler.java
gus-asf: That quote sounds scary because the code is complex and hard to describe. I am totally in agreement that we could refactor things further. (it was actually just refactored recently to use CompleteableFutures). However, I'm trying very hard to resist the (strong) temptation to completely redesign everything since I'm hoping a version of this can go in to 9x.

Here's a quick(?!?!) attempt to summarize (and possible first seed material of a blog post :) ):

The SearchHandler currently has 2 modes that it alternates until deciding it is done... First it enters a while loop that I will call [sending mode](https://github.com/apache/solr/blob/d9d2af471610c66d56b43981215afec8e41ff65b/solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java#L517C11-L552C1) where it sends every outgoing request it knows about using `shardHandler.submit(sreq, shard, params);`. Then it enters a while loop that I will call [receiving mode ](https://github.com/apache/solr/blob/d9d2af471610c66d56b43981215afec8e41ff65b/solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java#L557C1-L577C14) where it uses `ShardResponse srsp = tolerant ? shardHandler.takeCompletedIncludingErrors() : shardHandler.takeCompletedOrError()` to drain the responses that it has received unless something new shows up in outgoing. 

These two modes are contained in a while loop that is checking that there are no more outgoing requests. So the code reverts to send mode any time there's something to send, and exits if it finishes receiving with nothing further to send. This outer loop is needed because the receive process may queue new requests, but if we are done receiving no more requests can be queued, so at that point if nothing needs sending it's safe to continue.

In the classic single threaded HttpShardHandler, we can use a single atomic boolean incremented every time we send something and decremented when we receive something to keep track of how many things it was asked to send, and if we've received them all. 

In the new ParallelHttpShardHandler the send loop only schedules the send (to be performed by a thread in an executor before passing the thread, and the send mode loop will exit without any guarantee that even one http request is in flight yet. The Runnable that is passed to the executor calls `this.lbClient.requestAsync(lbReq);` and that method calls out into the actual http communication libraries which might be jetty http, jdk http, or whatever the latest fad for handling http is in the future. As such it doesn't seem wise to trust that it completes without exception. The callback is added to the future on the next line, but we can't be sure that that will happen.

Key take away: With parallel shard submit, we have two levels of tracking, one for initial intentions to make a request (at the SearchHandler level) and one for "has the Runnable been processed at all (at the shard handler level). 

This is important because [the `take(boolean)` based methods](https://github.com/apache/solr/blob/d9d2af471610c66d56b43981215afec8e41ff65b/solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java#L310C1-L318C4) have a loop that depends on obtaining a response for every thing we attempted. If we don't **track** what we attempt and **track** that the *attempt was made* (regardless of success/fail) then we run the risk of exiting that loop **before** we have ***started*** any of our http requests (in the executor threads which could delay/pause any amount of time)

So the sentence you quote about "needing to know everything we scheduled has been attempted", is referring specifically the new ParallelHttpShardHandler class question of whether or not the Runnable has been attempted by the Executor. So it's about _request initiation_ (remeber it's calling run**Async**()!), and has nothing to do with request completion.

Not discussed for simplicity:
 - tests of responses received vs the shard count that go hay wire the pending request count never exceeds the shard count.
 - fake responses created to satisfy the shard count test when shards are down.
 - empty strings as shard urls that inflate the shard count when shards are down necessitating the fake response.

solr/CHANGES.txt
dsmiley: If partial results are not allowed but the timeAllowed limit is reached, is the search effectively cancelled altogether then?  I recall this has been the case with timeAllowed & partialResults=false param (`CommonParams.PARTIAL_RESULTS`) already.  If this CHANGES.txt entry is trying to communicate the objective of this PR, I'm very confused and I think other users might be too, who I thought have already leveraged this behavior (or thought they were).

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
gus-asf: shards.tolerant is boolean. It just is incompatible with partialResults = false. The behavior with respect to REQUIRE_ZK_CONNECTED seems suspicious (silently ignore the value of shards.tolerant), but that's a change for a different ticket.

solr/core/src/java/org/apache/solr/handler/component/ParallelHttpShardHandler.java
gerlowskija: > As such it doesn't seem wise to trust that it completes without exception. The callback is added to the future on the next line, but we can't be sure that that will happen.

I followed you up until about here: what callback/lines are we talking about?  What failure/concurrency scenario are you envisioning with that?

I might be misunderstanding you entirely - but as I understand what you've said - that'd be the crux of the matter since it's what lead you to make the "pending" increment async, which in turn necessitates some of the other concurrency controls, etc.

gus-asf: https://github.com/apache/solr/blob/d9d2af471610c66d56b43981215afec8e41ff65b/solr/core/src/java/org/apache/solr/handler/component/ParallelHttpShardHandler.java#L91C11-L93C88

We schedule the async future with requestAsync, and then we add the whenComplete callback. This code should not trust the lbHttpClient method to return gracefully. An exception would leave us with no whenComplete() call, and then, the take loop would wait forever for a request that will never get added to requests.

gus-asf:
As a sanity check I ran the jmh benchmarks on main and this PR branch. Here's the comparison: 

https://docs.google.com/spreadsheets/d/1goRvpW1XEuLMrN_1BkB0OGHFYRDLUqo7E8XTCVvp9sU/edit?usp=sharing

Looks like most changes are noise, and the biggest changes are improvements I'm comfortable with this.

solr/core/src/java/org/apache/solr/handler/component/ParallelHttpShardHandler.java
gus-asf: This code works, and it doesn't seem to have an impact on performance. If you can think of a more elegant solution, perhaps that can be a follow on ticket?

gus-asf:
Ok I think this is ready, and I think I've answered the questions/feedback so far, so I'm putting it in main, but if issues arise or there's a strong feeling something needs to be reworked let me know, and we can fix it or revert if necessary. I want to move on to SOLR-17151. If things remain quiet until this ticket on Friday, I'll start looking at pushing a version of it to 9x

solr/core/src/java/org/apache/solr/handler/component/ParallelHttpShardHandler.java
gerlowskija: > We schedule the async future with requestAsync, and then we add the whenComplete callback. This code should not trust the lbHttpClient.requestAsync method to return gracefully.

I see the bug or gap you're highlighting, but aren't there ways to handle that case without all of the additional counters?

e.g. Wrap the [lbHttpClient.requestAsync](https://github.com/apache/solr/blob/b748207bc331b5eeae284ee7602626dbe5e3ff50/solr/core/src/java/org/apache/solr/handler/component/ParallelHttpShardHandler.java#L91) in a try-catch that decrements the `pending` counter if the request-sending failed.  Or if you prefer, give the CompletableFuture returned by [CompletableFuture.runAsync](https://github.com/apache/solr/blob/b748207bc331b5eeae284ee7602626dbe5e3ff50/solr/core/src/java/org/apache/solr/handler/component/ParallelHttpShardHandler.java#L114C5-L114C31) a hook to handle that exception case.

Even seeing the gap, all these counters just don't seem needed to me.

fsparv: I'm not sure I've been clear. My response was explaining the section you quoted, but that's not the part that actually motivates the counters...  The key case those two counters guard against is this:

1. send/recieve [loop starts](https://github.com/apache/solr/blob/d9d2af471610c66d56b43981215afec8e41ff65b/solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java#L514)
2. send mode [loop starts](https://github.com/apache/solr/blob/d9d2af471610c66d56b43981215afec8e41ff65b/solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java#L517)
3. submit shard req 1 to Executor (waits for thread avail)
4. submit shard req 2 to Executor (waits for thread avail)
5. submit shard req 3 to Executor (waits for thread avail)
6. send mode loop exits, nothing left to send
7. receive mode [loop starts](https://github.com/apache/solr/blob/d9d2af471610c66d56b43981215afec8e41ff65b/solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java#L557)
8. recieve mode calls take(boolean) (via whichever route)
9. the [loop in take()](https://github.com/apache/solr/blob/d9d2af471610c66d56b43981215afec8e41ff65b/solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java#L322) sees pending == 0 (because no reqs have executed yet) <<< SEED of PROBLEM
10. take loop exits, take(boolean) returns null.
11. receive mode loop exits, (nothing received so nothing new has been added to outgoing)
12. (receive/send alternating loop sees no further outgoing requests exits) <<< THE PROBLEM MANIFESTS
13. Main thread starts attempting to handle the (non-existent) responses
14. Thread becomes available, and Executor finally executes shard req 1 

The results from the parallel execution will be added to responses after some or all of the response processing has completed leading to nondeterministic results or errors, or in the worst case, we'd possibly return zero results (pretty sure current code would error with NPE, but that could get "fixed" later on, possibly when this shows up in someone's logs).

If we want to ensure that `pending` and `responseFutureMap` updates are synchronized for atomicity and "happens before" relationships, we have to do them both in the callback that runs asynchronously.

gerlowskija: Gotta love a 14-point list; thanks for being thorough!

> the loop in take() sees pending == 0 (because no reqs have executed yet) <<< SEED of PROBLEM

This all makes sense as things are written, but doesn't really explain why a simpler solution isn't possible. 
 If we need other counters because ParallelHttpShardHandler.submit() doesn't increment `pending` before returning...well, couldn't we modify submit() to behave that way?

Have `submit()` (or more accurately, `makeShardRequest`) increment the 'pending' counter before returning.  The runnable we create in `PHSH.makeShardRequest` would also have to decrement the counter in the error cases you've highlighted elsewhere in this thread (e.g. if the LB client returns an error).  But that'd work, seemingly (and be less complex than having multiple counters)?  Or is there still a gap with that approach that I'm missing?

gus-asf: It partly depends on what you want to simplify. One of the original difficulties I identified was the fact that the three key objects: responses responseFutureMap and pending had an intricate and delicate relationship that was very difficult to reason about in the presence of multi-threading, and possible cancellation as well as exceptions being thrown. There are questions of order of operations, as well as memory model happens before relationships to consider, and the more code that stands between these objects and their coordinated modification, the more complicated the logic. Concurrency is hard, so I'm trying to simplify THAT aspect as much as possible. 

My strategy was to create minimal synchronized blocks that group updates to these objects, any time the responseFutureMap is in play, and generally move unrelated code out from between interactions with these objects. 

So at a high level, your suggestion would exchange 2 counters (relatively easy to reason about) with a separation across threads between the update to pending and the update to responseFutureMap.

That separation introduces the possibility that pending might be updated but if the the intervening code fails, there is no guarantee that a response will ever be placed in responses (note that just decrementing pending isn't enough, because THREE objects have to stay coordinated). The thread could stall indefinitely under load, die with OOM, or hit a bug in our code or a library... 

If that happens, the original request thread will be stuck in the loop in take() forever.

In this PR's code, **pending will never be updated unless the future has been created, and a whenComplete callback added** _which implies that a response will always be added to the responseFutureMap regardless of what happens down stream_, because the response is something we have already created and passed into the callback. (I'm trusting the prior code that there's no possibility of rsp=null and exception=null simultaneously here). 

So... 

- if we remove the counters with no other change we can skip past our requests and fail or give incorrect response.
- if we remove the counters and move pending as you suggest we avoid skipping past, but might poll forever in take(boolean) (or further reverting to prior code with blocking responses.take(), we would wait forever).

All of this suggests that the best solution is, as David mentioned, to rework all of this to use CompletionService. But that's too big a change for this PR IMHO.

dsmiley: bq.  as David mentioned, to rework all of this to use CompletionService. 

Thanks for finally responding to this point; I wasn't sure if my idea resonated with anyone or not.

bq. But that's too big a change for this PR IMHO

It could fundamentally change the code and render your approach obsolete.  I hope we haven't dug ourselves a bigger tech-debt hole here; I dunno.  It'll be scarier for whoever takes this on, if it happens.

gus-asf: > It could fundamentally change the code and render your approach obsolete. I hope we haven't dug ourselves a bigger tech-debt hole here; I dunno. It'll be scarier for whoever takes this on, if it happens.

Well to that end I've tried to leave comments clarifying things it took me hours to tease out and understand. If I can find free time or even sponsorship for that task, I'd certainly be happy to tackle it (soon would be excellent since I have my head wrapped around the issues now), but it also seems like something that radical shouldn't get ported to 9x, whereas I hope to move portions of this not pertaining to the parallel handler down to 9x (unless that got ported to 9x... I haven't actually checked yet).

PR #2398, opened 2024-04-10
chore(deps): update dependency commons-io:commons-io to v2.17.0 - autoclosed

solrbot:
This PR contains the following updates:

| Package | Type | Update | Change |
|---|---|---|---|
| [commons-io:commons-io](https://commons.apache.org/proper/commons-io/) ([source](https://gitbox.apache.org/repos/asf?p=commons-io.git), [changelog](https://commons.apache.org/proper/commons-io/changes-report.html)) | dependencies | minor | `2.15.1` -> `2.17.0` |

---

### Configuration

📅 **Schedule**: Branch creation - "* * * * *" (UTC), Automerge - At any time (no schedule defined).

🚦 **Automerge**: Disabled by config. Please merge this manually once you are satisfied.

♻ **Rebasing**: Whenever PR becomes conflicted, or you tick the rebase/retry checkbox.

🔕 **Ignore**: Close this PR and you won't be reminded about this update again.

---

 - [ ] If you want to rebase/retry this PR, check this box

---

This PR has been generated by [Renovate Bot](https://redirect.github.com/solrbot/renovate-github-action)

PR #2669, opened 2024-08-26
Refactor: Use InputStream.transferTo

dsmiley:
Instead of longer code that accomplishes the same.


solr/core/src/java/org/apache/solr/cli/PostTool.java
```diff
@@ -1054,16 +1055,12 @@ public static InputStream stringToStream(String s) {
...
-    int read = 0;
-    while ((read = source.read(buf)) >= 0) {
-      if (null != dest) {
-        dest.write(buf, 0, read);
-      }
-    }
-    if (null != dest) {
-      dest.flush();
+    if (dest == null) {
+      dest = NullOutputStream.INSTANCE;
```
dsmiley, line 1059: A commons-io utility

solr/core/src/java/org/apache/solr/handler/BlobHandler.java
```diff
@@ -261,6 +263,16 @@ public void write(OutputStream os) throws IOException {
     }
   }
 
+  private static InputStream boundedInputStream(final InputStream is, final long maxLength)
+      throws IOException {
+    return new BoundedInputStream(is, maxLength) {
```
dsmiley, line 268: A commons-io utility

solr/solrj/src/java/org/apache/solr/common/util/BytesOutputStream.java
```diff
@@ -22,6 +22,7 @@
 import java.io.UnsupportedEncodingException;
 import java.util.Arrays;
 
+@Deprecated
```
dsmiley, line 25: unused

solr/solrj/src/java/org/apache/solr/common/util/Utils.java
```diff
@@ -1175,29 +1175,10 @@ public byte[] getbuf() {
...
-   * Reads an input stream into a byte array
-   *
-   * @param is the input stream
-   * @return the byte array
-   * @throws IOException If there is a low-level I/O error.
-   */
-  public static ByteBuffer toByteArray(InputStream is, long maxSize) throws IOException {
     try (BAOS bos = new BAOS()) {
-      long sz = 0;
-      int next = is.read();
```
dsmiley, line 1192: this code was very sad, reading one char at a time!

risdenk (approved):

solr/core/src/java/org/apache/solr/handler/designer/DefaultSampleDocumentsLoader.java
```diff
@@ -65,14 +64,11 @@ public class DefaultSampleDocumentsLoader implements SampleDocumentsLoader {
   private static final int MAX_STREAM_SIZE = (5 * 1024 * 1024);
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
-  public static byte[] streamAsBytes(final InputStream in) throws IOException {
-    ByteArrayOutputStream baos = new ByteArrayOutputStream();
-    byte[] buf = new byte[1024];
-    int r;
+  /** Reads input completely into a byte array. Closes the stream. */
```
madrob, line 67: This feels super trappy to have this method close the stream and Utils.toByteArray leave it open, but not something we can fix here.

madrob:
We're on commons-io 2.15.1? We should upgrade to 2.16.1 in this PR, since you're already looking at it and I remember that some stream closing behavior changed between these two versions. Otherwise somebody is going to be very confused without context when tests break.

solr/core/src/java/org/apache/solr/handler/designer/DefaultSampleDocumentsLoader.java
dsmiley: Yes; it's unusual for a method to accept an InputStream and also close it.  I at least documented it.  And I reduced its scope from public.  I was tempted to rename the method to something like streamAsBytesAndClose so that the closing is super apparent at the caller.  Can still do.

epugh: we have the `andClose` pattern in other places...

solr/core/src/java/org/apache/solr/cli/PostTool.java
epugh: Love it!

solr/solrj/src/java/org/apache/solr/common/util/BytesOutputStream.java
epugh: could we add the since tag to help us remember when to remove this deprecated method?   There are so many deprecated methods that just kind of hang around because we don't have good mechanisms for deciding/remembering to remove them!

epugh (approved):
LGTM.  The upgrade to commons-io would be nice.

janhoy:
> We're on commons-io 2.15.1? We should upgrade to 2.16.1 in this PR

Already open PR for it, but blocked on forbidden-apis: #2398 and https://github.com/policeman-tools/forbidden-apis/pull/249

solr/solrj/src/java/org/apache/solr/common/util/BytesOutputStream.java
dsmiley: I deleted it; we don't need to wait for internal utilities

dsmiley:
@madrob I made further changes based on your comments in the schema designer.  LMK what you think.

Otherwise I think it's mergeable (no JIRA).

madrob (approved):

PR #2670, opened 2024-08-27
SOLR-17422: Remove v2 /api/cluster (CLUSTERSTATUS)

dsmiley:
It's a kitchen sink API that is redundant with finer grain endpoints.

https://issues.apache.org/jira/browse/SOLR-17422

I generated the ref guide to review what the modified page looked like.  Not ideal but it was formatted the same before SOLR-15748 added the API that this PR removes.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -143,6 +143,8 @@ Other Changes
 
 * SOLR-17399: Replace the use of the deprecated java.util.Locale constructor with Locale Builder API. (Sanjay Dutt)
 
+* SOLR-17422: Remove V2 /api/cluster
+
 ==================  9.7.0 ==================
 New Features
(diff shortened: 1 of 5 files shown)


epugh:
Would adding some pointers to the specific api's be helpful in RefGuide and javadocs?  I can imagine a lot of user requests saying "what are the specific end points that replace this"....

gerlowskija:
> Would adding some pointers to the specific api's be helpful in RefGuide and javadocs?

+1 on documenting the more granular APIs that can achieve this.

Are we sure there aren't any gaps btw - I can't think of anything offhand that returns "live node" data, for instance?  Ditto for listing cluster properties.  I'm all for removing the kitchen-sink endpoint, but only if there's another way for users to fetch everything it offers...

PR #2678, opened 2024-08-29
SOLR-17427: Fix url building in cli

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17427

### DIFF
--- a/solr/core/src/java/org/apache/solr/cli/PostTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/PostTool.java
@@ -694,13 +694,14 @@ public class PostTool extends ToolBase {
    * @param link the absolute or relative link
    * @return the string version of the full URL
    */
-  protected static String computeFullUrl(URL baseUrl, String link) throws MalformedURLException {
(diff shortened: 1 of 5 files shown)


solr/core/src/java/org/apache/solr/cli/PostTool.java
```diff
@@ -379,37 +379,32 @@ private void doArgsMode(String[] args) {
   private void doWebMode() {
     reset();
     int numPagesPosted;
```
epugh, line 381: i love it when we can elminate a try and the nesting of code!

solr/core/src/test/org/apache/solr/cli/PostToolTest.java
```diff
@@ -211,30 +211,25 @@ public void testNormalizeUrlEnding() {
   }
 
   @Test
-  public void testComputeFullUrl() throws IOException {
-
-    PostTool webPostTool = new PostTool();
```
epugh, line 216: Nice!

epugh (approved):
LGTM!   I appreciate the bug fixes and the bits of clean up!

PR #2680, opened 2024-08-29
Use separate nodeProjectDir for each subproject

tylerbertrand:
# Description

Sometimes, `NpmTask`s are failing with being unable to find files/commands under `npmProjectDir`. [Here's an example of such a failure](https://ge.solutions-team.gradle.com/s/edxfitmmecv7i/failure#1). The failing tasks always execute after the `nodeSetup` task for that project, so in theory, everything should be in place.

However, in this case, multiple subprojects share the same `npmProjectDir`. The way the `gradle-node-plugin` works, having multiple projects use the same `npmProjectDir` doesn't save work, and can actually cause timing issues like the failures we are seeing. This is because [each `nodeSetup` task clears out the old directory before it unpacks it again](https://github.com/node-gradle/gradle-node-plugin/blob/main/src/main/kotlin/com/github/gradle/node/task/NodeSetupTask.kt#L50-L85). [In the Timeline view of the same Build Scan linked above](https://ge.solutions-team.gradle.com/s/edxfitmmecv7i/timeline?details=wa5cqs7lpiyvq&end=1724359674712&outcome=failed&start=1724359674660), we can see that the `:solr:solr-ref-guide:downloadAntoraCli` task (an `NpmTask`), is executing at the same time as `:solr:packaging:nodeSetup`, and `downloadAntoraCli` can't find `node` because it's been cleared out by another suproject's `nodeSetup` task.

# Solution

This PR updates each subproject to use its own individual `nodeProjectDir` to avoid these timing issues. Separating these out doesn't result in `node` being downladed twice - despite unpacking the same version multiple times, it's only ever downloaded once.  So no extra work is being done, since sharing the same `nodeProjectDir` doesn't save work in the first place, as explained above.

# Tests

No new tests were added, and no existing tests were modified, but the `check` and `integrationTests` tasks consistently pass when run locally with these changes.

# Checklist


epugh (commented):
It seems to make sense, I wish I knew more to weigh in one way or the other, hopefully someone else gives a +1!

janhoy:
@tylerbertrand Will you resolve the conflicts, then me or @epugh will get it merged?

epugh:
thanks for reminding us @janhoy

tylerbertrand:
@janhoy Sure, sounds good

tylerbertrand:
Hi @janhoy and @epugh, the conflicts have been resolved

epugh:
Thanks for the contribution @tylerbertrand!

dsmiley:
I'm going to move the CHANGES.txt entry from "Bug fix" into the "Other" category because Solr itself does not have a bug.

Also note the Lucene project uses "GITHUB#1234" pattern; we might want to do similarly.  It's more future-proof as some references are to issues not PRs specifically.  I'm making this change as well.

PR #2683, opened 2024-08-30
SOLR-17423: use long form, -h now means --help, not --host in solr.cmd

epugh:
https://issues.apache.org/jira/browse/SOLR-17423

Follow to original PR that fixes for windows.

### DIFF
--- a/solr/bin/solr.cmd
+++ b/solr/bin/solr.cmd
@@ -582,7 +582,7 @@ IF "%firstChar%"=="-" (
 )
 
 set SOLR_HOST=%~2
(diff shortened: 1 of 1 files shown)


epugh:
@malliaridis @rahulgoswami FYI, here is the fix on windows.   @rahulgoswami  I know you are on vacation, when you get back, this will likely be merged, but if you see any "suss" as the young people say, please holler!

malliaridis (approved):
Thanks for updating this so quickly. I hope we haven't overseen any other use case of `-h`.

PR #2687, opened 2024-09-02
SOLR-17432: Allow OTel Java Agent

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17432

Admittedly there's a lot going on here; the PR is more of a refactoring / minor improvements that also happens to support the Otel agent too, as it's a small aspect.  I can separate this PR.  A number of things were gnawing at me, which I started to deal with.

API Change:  TracerConfigurator subclasses now create an OpenTelemetry not a Tracer.  Since we depend on GlobalOpenTelemetry, if a configurator were to return some custom Tracer thing (independent of an OpenTelemetry), it's not clear how we would install it into a GlobalOpenTelemetry.  Maybe possible (?) but doesn't seem a good direction.  We could stop our use of GlobalOpenTelemetry easily as it's only used in one spot that could actually get the Tracer from elsewhere.  Maybe we should do that anyway, but nonetheless installing it is something we should continue to do.

Refactorings / changes:
* removed all tracing config from MiniSolrCloudCluster
* SolrTestCase resets tracing now
* Only TracerConfigurator should call GlobalOpenTelemetry.set (removed from SimplePropagator & OtelTracerConfigurator) and it does so in exactly one spot.  
* Tests examining spans now use OpenTelemetryRule. Removed CustomTestOtelTracerConfigurator.java

todo: test Custom TracerConfigurator ?
Do we even have a test actually using OtelTracerConfigurator.java?  
Admittedly there's no test here for the Agent approach.  Would need a bats test to do right, and it'd need to go download a 20MB jar.  I tested it manually when running a test (to debug a test) and I have used it in Solr 9.


solr/core/src/java/org/apache/solr/core/TracerConfigurator.java
```diff
@@ -38,33 +41,53 @@
...
+      if (otel == null && TRACE_ID_GEN_ENABLED) {
+        otel = OpenTelemetry.propagating(ContextPropagators.create(SimplePropagator.getInstance()));
+        log.info("OpenTelemetry loaded with simple propagation only.");
+      }
+      if (otel == null) {
+        otel = OpenTelemetry.noop();
+      }
+
+      try {
+        GlobalOpenTelemetry.set(otel); // throws IllegalStateException if already set
```
dsmiley, line 78: this is the one spot where we call this.  Previously it was in multiple places which was confusing and hard to reason about.

solr/core/src/java/org/apache/solr/core/TracerConfigurator.java
```diff
@@ -38,33 +41,53 @@
...
+        // skips needless ExecutorUtil.addThreadLocalProvider below
+        if (otel == OpenTelemetry.noop()) return otel.getTracer("solr");
+      } catch (IllegalStateException e) {
+        log.info("GlobalOpenTelemetry was already initialized (a Java agent?); using that");
```
dsmiley, line 82: Note that its common tests will log "OpenTelemetry loaded with simple propagation only." but then log immediately after "GlobalOpenTelemetry was already initialized (a Java agent?); using that" if the test uses more than one CoreContainer (Solr node).  I so wish GlobalOpenTelemetry had a way to know if it's been set already.

solr/core/src/java/org/apache/solr/core/TracerConfigurator.java
```diff
@@ -38,33 +41,53 @@
...
+      }
     }
-    if (TRACE_ID_GEN_ENABLED) {
-      ExecutorUtil.addThreadLocalProvider(new ContextThreadLocalProvider());
-      return SimplePropagator.load();
-    }
-    return TraceUtils.getGlobalTracer();
+
+    ExecutorUtil.addThreadLocalProvider(new ContextThreadLocalProvider());
```
dsmiley, line 86: Now called in exactly one spot, which I find easier to read than before

solr/core/src/java/org/apache/solr/core/TracerConfigurator.java
```diff
@@ -38,33 +41,53 @@
...
-          loader.newInstance(info.className, TracerConfigurator.class);
-      configurator.init(info.initArgs);
-      ExecutorUtil.addThreadLocalProvider(new ContextThreadLocalProvider());
-      return configurator.getTracer();
-    }
-    if (shouldAutoConfigOTEL()) {
-      return autoConfigOTEL(loader);
+  /** Initializes {@link io.opentelemetry.api.GlobalOpenTelemetry} and returns a Tracer. */
+  public static synchronized Tracer loadTracer(SolrResourceLoader loader, PluginInfo info) {
+    // synchronized to avoid races in tests starting Solr concurrently setting GlobalOpenTelemetry
```
dsmiley, line 54: I didn't personally encounter an issue here but saw existing comments (from Alex?) that suggested to me making this synchronized would be helpful.  I don't think there's a true issue; ultimately GlobalOpenTelemetry will be set exactly once so it doesn't much matter if others try and race to do so.

solr/core/src/java/org/apache/solr/core/TracerConfigurator.java
```diff
@@ -38,33 +41,53 @@
...
-      return SimplePropagator.load();
-    }
-    return TraceUtils.getGlobalTracer();
+
+    ExecutorUtil.addThreadLocalProvider(new ContextThreadLocalProvider());
+    return TraceUtils.getGlobalTracer(); // GlobalOpenTelemetry.get...
   }
 
-  protected abstract Tracer getTracer();
+  protected abstract OpenTelemetry createOtel();
```
dsmiley, line 90: this is the API change

solr/modules/opentelemetry/src/java/org/apache/solr/opentelemetry/OtelTracerConfigurator.java
```diff
@@ -49,14 +48,13 @@ public OtelTracerConfigurator() {
   }
 
   @Override
-  public Tracer getTracer() {
-    return TraceUtils.getGlobalTracer();
```
dsmiley, line 53: A needless GlobalOpenTelemetry reference

solr/modules/opentelemetry/src/java/org/apache/solr/opentelemetry/OtelTracerConfigurator.java
```diff
@@ -49,14 +48,13 @@ public OtelTracerConfigurator() {
...
+  public void init(NamedList<?> args) {
+    prepareConfiguration(args);
   }
 
   @Override
-  public void init(NamedList<?> args) {
-    prepareConfiguration(args);
-    AutoConfiguredOpenTelemetrySdk.initialize();
+  public OpenTelemetry createOtel() {
+    return AutoConfiguredOpenTelemetrySdk.builder().build().getOpenTelemetrySdk();
```
dsmiley, line 57: notice the change to *not* call initialize.  That method arranges for GlobalOpenTelemetry to be set.  I only want that set in TracerConfigurator.

solr/modules/opentelemetry/src/test-files/solr/solr.xml
```diff
@@ -34,8 +34,6 @@
     <int name="connTimeout">${connTimeout:15000}</int>
   </shardHandlerFactory>
 
-  <tracerConfig name="tracerConfig" class="org.apache.solr.opentelemetry.CustomTestOtelTracerConfigurator" />
```
dsmiley, line 37: Perhaps this class could return but only for a test explicitly testing that tracerConfig works and in a separate solr.xml for exactly that test as well (not for the other tracing tests).  Not to be used for other tests or trying to collect spans for examination.

solr/modules/opentelemetry/src/test/org/apache/solr/opentelemetry/BasicAuthIntegrationTracingTest.java
```diff
@@ -31,23 +33,23 @@
...
-import org.junit.AfterClass;
 import org.junit.BeforeClass;
+import org.junit.ClassRule;
 import org.junit.Test;
 
 public class BasicAuthIntegrationTracingTest extends SolrCloudTestCase {
 
   private static final String COLLECTION = "collection1";
 
+  @ClassRule public static OpenTelemetryRule otelRule = OpenTelemetryRule.create();
```
dsmiley, line 44: I was about to write an equivalent of OpenTelemetryRule when I found that it already exists :-)

FYI note it has docs that it doesn't work in a ClassRule yet I see it works here.

solr/modules/opentelemetry/src/test/org/apache/solr/opentelemetry/CustomTestOtelTracerConfigurator.java
dsmiley, line 1: this thing was strange; didn't need to exist, at least not as it existed

solr/core/src/java/org/apache/solr/core/TracerConfigurator.java
dsmiley: A possible hacky work-around would be to `get()` initially and check for noop. If not noop, do nothing (keep current impl).  If noop,  create one according to our logic and then call `resetForTest()` then `set()`.  Hacky because we'd be calling resetForTest in a production scenario.

janhoy (commented):
I don't have the bandwidth to do a full review and test now. But I agree on the goals and according to PR description, this is a nice evolution of our API to allow more ways to configure tracing and get more spans from various libraries free.

I'd perhaps like to see some screenshot of a Jaeger trace with the existing otel module before and after the change, to verify that we don't get unintentional changes? I suppose, if you use an agent, our manual instrumentation will still kick in, in addition?

dsmiley:
When I run the agent, I provide an Otel properties file with: (an excerpt)
```
otel.metrics.exporter=none
otel.logs.exporter=none
otel.instrumentation.common.default-enabled=false
otel.instrumentation.opentelemetry-api.enabled=true
otel.instrumentation.opentelemetry-instrumentation-annotations.enabled=true
otel.instrumentation.aws-sdk.enabled=true
otel.instrumentation.java-http-client.enabled=true
```
Notice I disable auto-instrumentation by default and selectively enable the specific instrumentations useful to me.  This works with Solr's manual instrumentation of itself as well because of enabling "opentelemetry-api".  All this is documented quite well in Otel.

I don't see how this PR would have unintentional changes on existing traces/spans since they are tested and the PR doesn't change them.  Any way, I'm happy to share a screenshot later.

This PR misses documentation to let users know they don't have to enable the opentelemetry module just to use opentelemetry :-). (an aside -- nor do we tell users they don't have to have an HDFS cluster to make use of the HDFS Directory).  Also I'd like to somewhere document a tip on how to use tracing during development / debugging of a unit test.

dsmiley:
I'm a little unhappy with keeping TracerConfigurator named this while having it return an OpenTelemetry.  Really; we're embracing OpenTelemetry here.  I would love to use OpenTelemetry for metrics eventually, to be honest.

janhoy:
I just saw SOLR-17455 coming in. While mentioning OpenTracing is correct for 9.x, for 10.0 it is gone, so perhaps when you add refGuide docs for this PR, you can also `s/OpenTracing/OpenTelemetry/` in the guide?

dsmiley:
I very much want to get back to this for Solr 10... need to maybe reformulate what's happening into multiple things.
CC @mlbiscoc

solr/core/src/java/org/apache/solr/core/TracerConfigurator.java
mlbiscoc: Maybe out of scope  but I feel like this whole setting of `GlobalOpenTelemetry.set(otel)` should sit in its own function called `loadGlobalOpenTelemetry` and be called as [early as possible as per its documentation](https://opentelemetry.io/docs/languages/java/api/#:~:text=If%20GlobalOpenTelemetry.get,by%20any%20instrumentation.) maybe inside the constructor of CoreContainer. Then on initialization of Solr core container, to set it's tracer just call `TraceUtils.getGlobalTracer()`.  But also this whole file is called `TracerConfigurator` when a lot of work is configuring OTel then just returning a static global tracer. This same logic is still used for metrics as it uses the same Java Agent and OTLP exporter from the module not just Tracer loading.

solr/core/src/java/org/apache/solr/core/TracerConfigurator.java
stillalex: hmm. if I remember correctly this is why I created `CustomTestOtelTracerConfigurator` so tests that race to set this would not fail. it felt awkward to have to catch `IllegalStateException` for problems that are triggered only via testing code.
I also remember there was an issue with checking for noop? there were some docs that were stating you should not check if the impl is noop, but I can't find a reference now.

solr/core/src/java/org/apache/solr/core/TracerConfigurator.java
stillalex: not sure `synchronized` is needed. the trouble I had seen were from tests trying to race to init this and failing because otel was already set. I think it was happening when setting up a cluster with a few nodes, but I need to revisit some of this to check.

solr/modules/opentelemetry/src/java/org/apache/solr/opentelemetry/OtelTracerConfigurator.java
stillalex: sorry, could you clarify what this means?

solr/modules/opentelemetry/src/java/org/apache/solr/opentelemetry/OtelTracerConfigurator.java
stillalex: this diff would have been easier to read if the 2 methods were not swapped.

solr/core/src/java/org/apache/solr/util/tracing/SimplePropagator.java
```diff
@@ -54,19 +46,6 @@ public class SimplePropagator implements TextMapPropagator {
 
   private static final AtomicLong traceCounter = new AtomicLong(0);
 
-  private static volatile boolean loaded = false;
-
-  public static synchronized Tracer load() {
```
stillalex, line 59: +1 moving this to TracerConfigurator is a good change.

solr/core/src/java/org/apache/solr/core/TracerConfigurator.java
stillalex: coming back to this. not sure what your plans are, but I will reshare this approach I had in the `SimplePropagator` 
https://github.com/apache/solr/pull/2687/files#diff-e61b6fa689b156777ba2ecb37be366c4c312199302a2c90425acb460dbdb3f0eL57 - I think using a flag to keep track of status works to get around the issue of multiple init calls.

stillalex:
+1 this looks good to me. when you have some time, could you rebase? curious to see what errors are left.
I think the open question is the issue of multiple init calls causing that OTEL exception during tests.

dsmiley:
I'll come back to this today/tomorrow.  I really appreciate the close look @stillalex :-)
My biggest question is really scoping.  This PR is a bit all over the place; it will take some work to break it into more meaningful chunks.  I'll think about this further when I look closer again.

dsmiley (commented):
(nevermind; wrong PR)

dsmiley:
Dusted this off with Claude; things have been renamed/changed since when I last touched this.  Also, I updated the ref guide to have a dedicated opentelemetry page to avoid re-discussing aspects in metrics & tracing.

It's a bit of a misnomer to title this as "Allow OTEL Java Agent" because it's already allowed, albeit with an ugly ClassNotFoundException that can be ignored.  Overall this makes it more first-class / documented.

solr/solr-ref-guide/modules/deployment-guide/deployment-nav.adoc
dsmiley, line 1: re-org a bit + add opentelemetry.adoc

PR #2689, opened 2024-09-02
SOLR-16503: New home for default http2 client

iamsanjay:

Spin Off #2351 

Many HTTP requests currently rely on the default client from UpdateShardHandler. However, it has been noted that UpdateShardHandler is no longer the appropriate location for this functionality. It has been suggested to migrate the default Http2SolrClient to a different class. In this PR, we propose introducing a class named ServerSolrClientCache, which extends SolrCacheClient, to host the default Http2SolrClient.


iamsanjay:
I still need to adjust how the HTTP client is initialized, but is this aligned with what you had in mind when you suggested extending SolrClientCache in [this discussion](https://github.com/apache/solr/pull/2351#discussion_r1595413371)?

solr/core/src/java/org/apache/solr/security/PKIAuthenticationPlugin.java
```diff
@@ -347,22 +348,27 @@ PublicKey fetchPublicKeyFromRemote(String nodename) {
...
-      Map<?, ?> m = (Map<?, ?>) Utils.fromJSON(bytes);
-      String key = (String) m.get("key");
-      if (key == null) {
-        log.error("No key available from {}{}", url, PublicKeyHandler.PATH);
-        return null;
-      } else {
-        log.info("New key obtained from node={}, key={}", nodename, key);
+
+      String key;
+      try (Http2SolrClient solrClient = new Http2SolrClient.Builder(url).build()) {
```
dsmiley, line 361: why create a new client when you can use the default client?

solr/core/src/java/org/apache/solr/security/PKIAuthenticationPlugin.java
```diff
@@ -347,22 +348,27 @@ PublicKey fetchPublicKeyFromRemote(String nodename) {
     HttpEntity entity = null;
     try {
       String uri = url + PublicKeyHandler.PATH + "?wt=json&omitHeader=true";
+      ModifiableSolrParams solrParams = new ModifiableSolrParams();
```
dsmiley, line 351: glad to see you convert to use a SolrClient :-).  It'll be nice to have the observability benefits.

dsmiley (commented):
CoreContainer.getSolrClientCache should change it's type to be of server.

iamsanjay:
I don't think we should extend SolrClientCache. While we're looking for a new home for the default HTTP client and have agreed that UpdateShardHandler isn't the right place, the same applies to SolrClientCache. Extending SolrClientCache implies that we intend to use some of its functionality, but it operates quite differently. `SolrClientCache` creates clients based on URLs and checks if a client already exists whenever a class needs one; if not, it creates a new one—hence the suffix "Cache." With the move to HTTP/2, we don't really need that functionality anymore. Therefore, I don't think extending SolrClientCache is a good idea. Additionally, the name doesn't suit the current HTTP/2 client since there's no cache—it's more of a provider, in my opinion.

iamsanjay:
An alternative would be to introduce another class and add a method in CoreContainer to return that class. However, we already have the `UpdateShardHandler` class which exactly does the same thing. While I understand that we don't want to place the new HTTP/2 client in there, extending `SolrClientCache` doesn't seem like the right approach either.

solr/core/src/java/org/apache/solr/security/PKIAuthenticationPlugin.java
iamsanjay: I was setting up the base path and initially used the `SolrRequest.setBasePath` method, which is now deprecated. I then tried to find a way to set the base path without using the deprecated API and remembered that we can recreate the builder with a URL to handle it. However, I know the consensus was to avoid repeatedly creating Http2SolrClient. Is there another way I can set the base path?

dsmiley: I deprecated setBasePath but should have better clarified that there is no substitute yet and thus for code inside Solr (what we're talking about here) definitely should continue to use it.

iamsanjay: Wait, looks like I did not passed the Http2SolrClient to this builder.  Hmmmmm, I need to pass that here.

dsmiley:
OK RE not subclassing SolrClientCache.

Despite CoreContainer's code size, I think it's reasonable to have a getHttpSolrClient on it as it's important and used widely enough (and I expect so by 3rd party plugins).

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -2563,6 +2581,11 @@ public PlacementPluginFactory<? extends PlacementPluginConfig> getPlacementPlugi
     return this.distributedCollectionCommandRunner;
   }
 
+  /** Returns default http solr client. */
+  public SolrClient getDefaultHttpClient() {
```
dsmiley, line 2585: Can return the specific type; there are async methods there which can be useful.

solr/core/src/java/org/apache/solr/core/HttpSolrClientProvider.java
```diff
@@ -0,0 +1,151 @@
...
+import org.apache.solr.metrics.SolrMetricManager;
+import org.apache.solr.metrics.SolrMetricsContext;
+import org.apache.solr.security.HttpClientBuilderPlugin;
+import org.apache.solr.update.UpdateShardHandlerConfig;
+import org.apache.solr.util.stats.HttpClientMetricNameStrategy;
+import org.apache.solr.util.stats.InstrumentedHttpListenerFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Internal use only. Provider of the default SolrClient implementation. */
```
dsmiley, line 38: flip the sentence order.  Javadocs should always start with the summary of what it is/does.  Internal Use only comment can be replaced with `@lucene.internal` which we use a lot and probably should use more.  But it's not that pertinent outside SolrJ.   Glad you were thinking of documenting internal-ness.  Maybe the class doesn't need to be public, which would imply its internal-ness without even needing an annotation/docs.

solr/core/src/java/org/apache/solr/core/HttpSolrClientProvider.java
```diff
@@ -0,0 +1,151 @@
...
+public class HttpSolrClientProvider implements SolrInfoBean {
```
dsmiley, line 39: Why does it implement SolrInfoBean?  It's for mostly "pluggable" things, unlike this.

solr/core/src/java/org/apache/solr/core/ServerSolrClientCache.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.http.client.HttpClient;
+import org.apache.solr.client.solrj.SolrClient;
+import org.apache.solr.client.solrj.impl.Http2SolrClient;
+import org.apache.solr.client.solrj.io.SolrClientCache;
+import org.apache.solr.common.util.IOUtils;
+import org.apache.solr.security.HttpClientBuilderPlugin;
+import org.apache.solr.update.UpdateShardHandlerConfig;
+import org.apache.solr.util.stats.InstrumentedHttpListenerFactory;
+
+/** New home for default http client. */
```
dsmiley, line 30: Do you intend to undo this?

dsmiley (commented):
> If recreating the client for a different baseUrl or core name is not recommended, we should consider removing such methods from Http2SolrClientBuilder and instead include the baseUrl and collection/core name in the request constructor to ensure these details are provided during initialization.

I recommend furthering the discussion and/or volunteering here: https://issues.apache.org/jira/browse/SOLR-17256

solr/core/src/java/org/apache/solr/core/CoreContainer.java
iamsanjay: We could recreate Http2SolrClient without casting, but I aimed to follow the principle of using base classes to allow for future implementation changes without altering the method signature. If requestAsync isn't used often, we could cast it when needed. or, Would that make it less clean?

solr/core/src/java/org/apache/solr/core/HttpSolrClientProvider.java
iamsanjay: Yeah agreed to make it package private. I guess I wanted to use it in UpdateShardHandler as well, like moving clients to this class. But yeah let's not make it too much complex for now. 

But in future, I wonder sometimes, Do we even need multiple http clients -- default, update or recovery?In past,  due to some issues with http1 we created multiple, however, is that reason even valid anymore with http2? If not, there is a possibility that having one client could improve the performance.  I guess benchmarking could have the answer to it.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
dsmiley: Since this is specifically an HttpClient (it's in the method name!) I think it's totally reasonable to return this type.  Casting is sad; it's a code smell.  reqeustAsync is somewhat new so you want see it called too many times.  It remains to be seen what the solution to [SOLR-17256](https://issues.apache.org/jira/browse/SOLR-17256) is but I could imagine a new request method containing the URL; if so then yet another reason to return a specific type.

solr/core/src/java/org/apache/solr/core/HttpSolrClientProvider.java
dsmiley: The different ones were added intentionally to tune timeouts, which matters at high scale when Solr is pushed to extremes.

solr/core/src/java/org/apache/solr/core/HttpSolrClientProvider.java
dsmiley: SolrMetricsProducer is probably not needed as we are likely directly/manually initializing producing metrics, so the interface accomplishes nothing.  I think.

iamsanjay: We have two options for passing the parent context used to initialize the HttpListenerFactory for HttpClient: either pass it through the constructor or create a method. We're using the second approach, where SolrMetricProducer requires subclasses to implement initializeMetrics. This method handles the initialization of InstrumentedHttpListenerFactory and also serves as a marker.

iamsanjay: It also take cares of closing of solrMatricContext. However, I forgot to call the super close method which deal with it. Will be adding it.

solr/core/src/java/org/apache/solr/core/HttpSolrClientProvider.java
iamsanjay: We can always re-set connection/socket timeout by re-creating the Http2SolrClient but using the same jetty client. Or is there any other setting you are talking about? I mean yes we do have many different settings at Jetty level as well, however looks like so far everything is similar for all three Http2SolrClients. 

I remember coming across a comment from Mark when the new code update client was introduced. 

https://issues.apache.org/jira/browse/SOLR-12293

> Currently the pool is shared too broadly - for example during replication we don't guarantee we read the full streams when downloading index files and we don't necessarily want to, emptying the stream for a huge file due to error or abort is too expensive. We can't have these connections pollute the update connection pool.

dsmiley (commented):
Shouldn't the ReindexCollection be using the "update" SolrClient instead of the default one?  At least I think it's more in line with that.  CC @sigram

dsmiley:
> Shouldn't the ReindexCollection be using the "update" SolrClient 

It's so trivial actually because ReindexCollectionCmd is merely issuing a request to the target node to stream the data; the command is not itself streaming the data (of course).  So might as well use a default client.  We might then wonder what "/stream" does but that's really out of scope here; IMO shouldn't change that now.

solr/core/src/java/org/apache/solr/core/HttpSolrClientProvider.java
iamsanjay: Actually I removed the interface. I thought about it and It is only valid If we want to expand the scope further. But for HttpSolrClientProvider that is not needed, So i dropped it.

solr/core/src/java/org/apache/solr/core/HttpSolrClientProvider.java
```diff
@@ -40,107 +34,55 @@
...
-
-  @Override
-  public void close() {
+  void close() {
     try {
-      SolrMetricProducer.super.close();
+      trackHttpSolrMetrics.close();
     } catch (Exception e) {
       log.error("Error closing SolrMetricProducer", e);
     }
```
dsmiley: we have IOUtils.closeQuietly

solr/core/src/test/org/apache/solr/core/TestCoreContainer.java
```diff
@@ -1131,6 +1132,48 @@ public void testCoreInitFailuresOnReload() throws Exception {
     cc.shutdown();
   }
 
+  public void testCustomSocketTimeoutForDefaultHttpClient() throws Exception {
```
dsmiley, line 1135: hmm; why this test here?

dsmiley (commented):
Overall, what do you think of the state of this PR is?  Not sure if we should switch the default client users to this in this PR or do separately.  The scope of that could be large.

iamsanjay:
> Overall, what do you think of the state of this PR is? Not sure if we should switch the default client users to this in this PR or do separately. The scope of that could be large.

Yeah +1 to do separately. I will remove the extra change.

solr/core/src/test/org/apache/solr/core/TestCoreContainer.java
iamsanjay: I want to test where we can change the SocketTimeout via UpdateShardHandlerCfg, Now `TestCoreContainer` seems to have that code where we can initlialize the CoreContainer with custom SolrXML file. So I added those.

iamsanjay: Additionally, I've noticed that the documentation does not cover how to pass UpdateShardHandlerConfig or how it affects the timeouts for the HTTP/2 client—specifically, the update, recovery, and default timeouts.

iamsanjay: https://solr.apache.org/guide/solr/latest/configuration-guide/configuring-solr-xml.html

solr/core/src/java/org/apache/solr/core/HttpSolrClientProvider.java
iamsanjay: `IOUtils.closeQuietly` takes Closeable, and InstrumentedHttpListenerFactory is a sub-class of AutoCloseable. Do we have any other utility that takes AutoCloseable?

solr/core/src/test/org/apache/solr/core/TestCoreContainer.java
dsmiley: I could see _learning_ from TestCoreContainer but putting your test there just to do something that could be done elsewhere doesn't seem like a good reason to put the test there.  Granted I'm also not sure where a good place for this is either :-).  So maybe here.

Must this thing be configured in solr.xml; can't we rely simply on EnvUtils ?  I have a wish-list JIRA issue to unify things in a more automatic way.  But in the mean time, I recommend preferring EnvUtils over solr.xml for simple (not structural / flexible) configuration.

iamsanjay: We can also just remove this test. The mock that I written is almost same what I am trying to test here via integration. I think I don't really need integration test in this case.

solr/core/src/java/org/apache/solr/core/HttpSolrClientProvider.java
dsmiley: closeQuiety could be modified to widen the parameter definition to AutoCloseable.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
iamsanjay: I was thinking about this one, I believe sticking to the parent class would benefit us in a long, I mean returning SolrClient. Instead we should introduce the empty/abstract method `requestAsync` in SolrClient, and override in all the sub classes, and where it does not support just throw an exception.

solr/core/src/java/org/apache/solr/core/HttpSolrClientProvider.java
dsmiley: That's a great reason to keep it separate, even if the configuration is identical.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
dsmiley: Okay for requestAsync even if debatable but what about SOLR-17256?  The solution for SOLR-17256 is not determined yet but the likely solution IMO would result in a URL specific request method on the client.  That doesn't belong on SolrClient.

iamsanjay: What If we introduced an requestAsyn with an empty body instead of making it abstract in the SolrClient that way It would not enforce others to implement it.

iamsanjay: I've observed that some classes take implementation classes as references, which makes them harder to change. For example, SolrClientCache initially used the Apache Http Client, but when Http2SolrClient was introduced, a new constructor was needed. Similarly, SolrClientCloudManager uses CloudLegacySolrClient, which complicates future changes. We might eventually need to introduce a SolrClientCloudManager2 to accommodate CloudSolrClient.

dsmiley: _If need be_ we can create an interface like `HttpSolrClient` (we have 3 impls) but the only backend impl we care about is Http2SolrClient (jetty) and I'm skeptical we're going to see another migration.

BTW sorry but a no-op/empty requestAsync would be nasty... think of it:  You call a method and it does absolutely nothing when it should have issues the request!

iamsanjay: I got it! That is ugly. Anyways I am okay with this method returning Http2SolrClient. 

@gerlowskija I already worked on changing it for two files PKIAuth and ReindexCollectionCmd, however we decided not to in this one and deal in separate PR. Take a look at this one, If everything's looking good then I will merge this one first and then I will make another.

gerlowskija (commented):
Great work on this overall Sanjay!  I left some comments inline, but they're all minor tweaks (Javadoc improvements, method naming, etc.)  Feel free to take or leave them as you see fit.

I do have one concern I feel a little more strongly about: the interface or API or whatever that Solr devs will be able to use.

IMO the current entry point (`CoreContainer.getDefaultHttpClient()`) sets a big "trap" for callers - the returned client can only be used in very select ways.  That is: consumers will get an exception unless they somehow know to call the relatively obscure `SolrRequest.setBasePath` method!  Even canny users would be unable to use most SolrClient methods, as only a few SolrClient methods take in a `SolrRequest` in the first place: `add()`, `commit()`, `deleteById()`, `deleteByQuery()`, `query()`, etc. would all be unusable in the returned instance!

Wdyt about saving `getDefaultHttpClient()` take in a baseUrl, so that the returned instance can take advantage of all SolrClient methods "out of the box"?

(Also - am I missing some existing discussion on this, maybe in 'resolved' GH comments?  The [comment here](https://github.com/apache/solr/pull/2689#issuecomment-2329584805) gives me the impression that this has already come up and you've put some thought into it, but I'm having trouble digging up that context...)

solr/core/src/java/org/apache/solr/core/CoreContainer.java
gerlowskija: Awesome, thanks for the heads up @iamsanjay !

solr/core/src/java/org/apache/solr/core/HttpSolrClientProvider.java
```diff
@@ -0,0 +1,88 @@
...
+import org.apache.solr.util.stats.InstrumentedHttpListenerFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Provider of the default SolrClient implementation.
+ *
+ * @lucene.internal
+ */
+class HttpSolrClientProvider {
```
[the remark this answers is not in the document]
iamsanjay: We were looking for a new home for this client and didn't want to add the burden of creating and managing it to CoreContainer. So, we created a new class for that purpose. However, since we're still in the initial phase and experimenting with it, we don't want to expose this class publicly. That's why it's package-private and marked with @lucene.internal. In fact, I'm considering making this class final as well.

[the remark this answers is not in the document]
iamsanjay: It takes `UpdateShardHandlerConfig` for setting such as IdleTimeout, connectionTimeout and maxConnectionPerDestination. However, I do think that this client should have its own configuration to configure those values. Because right now all of three Http2SolrClient -- update, recovery and default -- are sharing same configuration.

gerlowskija:
All the approaches have downsides for sure, but for me personally, object allocation and the 'Closeable'-burden seem like the lesser evils.

> unnecessary object creation

"Cloud" clients are a bit different because of their ongoing ZK connections, but otherwise we've long assumed that our "Http" SolrClients are cheap and can/should be instantiated as-needed.  That assumption underlies a lot of other decisions in SolrJ - e.g. the push to [make client impl's immutable and remove their setters](https://issues.apache.org/jira/browse/SOLR-8975).  And IMO that makes sense - object allocations have a cost, but in terms of bottom-line performance, those are almost surely negligible compared to the much heavier, slower network requests the client is about to do.

> responsible for closing 
 
And on the "they are responsible for closing" side - it is a bit of a burden, but it's one that Solr devs are used to paying attention to.  And even for new devs, there's a lot of tooling built around enforcing this: IDEs will flag the unclosed Closeable, Solr's test-harness detects unclosed resources, etc.  Requiring users to close out the client is a non-zero burden - I'm not waving that away.  It's just a much smaller one IMO (and ["less surprising"](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)) than giving callers a client that they immediately need to clone or know a special incantation to actually use.

Not going to -1 anything here - I'd hate to come in late and blow up your hard work.  But those are my 2c at least, from the perspective of "how would a caller likely expect this thing to work"

dsmiley:
I'm in favor of a CoreContainer.newHttpSolrClient(String url) method that the caller should close and that which uses a common underlying HttpClient instance / connection-pool underneath.

iamsanjay:
Of course I am not suggesting this change for now. But still want to share my concern here. 

The ability to add a listener factory after initialization introduces the potential for a race condition. Currently, both HttpSolrClientProvider and UpdateShardHandler call setSecurityBuilder later in the process, which delays object creation. If any other class recreates the client during this time, it won't be in a complete state, as it will be missing the security listeners.

It would be beneficial to incorporate pkiAuthenticationSecurityBuilder into the previously stated Solr client providers ctora and make it available to Http2SolrClient during construction, rather than at a later stage.

iamsanjay:
We have already taken a step towards achieving immutability for Http2SolrClient by deprecating addListenerFactory. As I mentioned earlier, the next step should be to ensure that the security listener is available during the construction of Http2SolrClient.

iamsanjay:
Now that the new method (`Http2SolrClient.requestWithBaseUrl`) is going to be introduced ion #2714. I am just re-thinking design of `CoreContainer#getDefaultHttpClient` method signature. May it should just returns the Http2SolrClient without re-creating it. And caller will call the new method to setup the baseUrl.

iamsanjay:
This one is ready and I am going to merge it in a day.

dsmiley (approved):

solr/solrj/src/java/org/apache/solr/common/util/IOUtils.java
```diff
@@ -16,15 +16,14 @@
...
-import java.io.Closeable;
 import java.lang.invoke.MethodHandles;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class IOUtils {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
-  public static void closeQuietly(Closeable closeable) {
+  public static void closeQuietly(AutoCloseable closeable) {
```
mkhludnev, line 26: fwiw it was not fully backward compatible https://github.com/SearchScale/dataimporthandler/issues/91 sigh

dsmiley: Ugh.  Fixing: https://github.com/apache/solr/pull/3199

PR #2694, opened 2024-09-06
SOLR-17441 MetricUtils: Skip Unreadable Properties

heythm:
https://issues.apache.org/jira/browse/SOLR-17441

# Description

Enhance the property introspection logic by skipping unreadable properties early in the process. By checking if desc.getReadMethod() is null before attempting to access the property, we can avoid unnecessary attempts to access properties that do not have a read method. This will reduce the number of exceptions thrown and the overhead associated with handling those exceptions, improving overall performance and stability.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/util/stats/MetricUtils.java
+++ b/solr/core/src/java/org/apache/solr/util/stats/MetricUtils.java
@@ -787,10 +787,16 @@ public class MetricUtils {
                 }
               });
 
+      // if BeanInfo retrieval failed, return early
       if (beanInfo == null) {
         return;
       }
       for (final PropertyDescriptor desc : beanInfo.getPropertyDescriptors()) {
+        // skip properties without a read method
+        if (desc.getReadMethod() == null) {
+          continue;
+        }
+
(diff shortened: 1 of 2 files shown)


dsmiley (approved):

solr/core/src/java/org/apache/solr/util/stats/MetricUtils.java
```diff
@@ -787,10 +787,16 @@ public static <T extends PlatformManagedObject> void addMXBeanMetrics(
                 }
               });
 
+      // if BeanInfo retrieval failed, return early
       if (beanInfo == null) {
         return;
       }
       for (final PropertyDescriptor desc : beanInfo.getPropertyDescriptors()) {
+        // skip properties without a read method
+        if (desc.getReadMethod() == null) {
```
dsmiley, line 796: this is fine but I think it'd be nicer to save the result and use it later rather than double-lookup.  Also nicer to get the read method inside the try-finally in case there's some issue even calling getReadMethod

dsmiley (approved):

PR #2697, opened 2024-09-08
SOLR-17444: Add Passage Sorting Option in UnifiedSolrHighlighter

Seunghan-Jung:
https://issues.apache.org/jira/browse/SOLR-17444

# Description

We will add the `passageSortComparator` option, introduced in the `UnifiedHighlighter` in the Lucene 9.11 project, to the Solr highlighter parameters, enabling its functionality within Solr.

For reference, here is the related Lucene project commit:

https://github.com/apache/lucene/commit/8773725ac0f15962b39bcc1f6fc5fb331b7a6d62
 

# Solution

1. Added the parameter name `PASSAGE_SORT` to HighlightParams to support the related functionality.
2. In the `UnifiedSolrHighlighter.SolrExtendedUnifiedHighlighter` class, I overrode the `getPassageSortComparator` method from the superclass `UnifiedHighlighter`. The `PASSAGE_SORT` parameter from the request is parsed and the corresponding `Comparator<Passage>` object is used.

# Tests

# Checklist


solr/core/src/java/org/apache/solr/highlight/UnifiedSolrHighlighter.java
```diff
@@ -314,6 +343,24 @@ protected PassageFormatter getFormatter(String fieldName) {
       return new DefaultPassageFormatter(preTag, postTag, ellipsis, "html".equals(encoder));
     }
 
+    @Override
+    protected Comparator<Passage> getPassageSortComparator(String fieldName) {
+      PassageSort passageSort =
+          PassageSort.parse(
+              params.getFieldParam(fieldName, HighlightParams.PASSAGE_SORT, "startOffset"));
```
dsmiley, line 350: could you instead reference the START_OFFSET.key instead of this constant string?

dsmiley (commented):
Nice!  Thanks for your contribution.

solr/core/src/java/org/apache/solr/highlight/UnifiedSolrHighlighter.java
Seunghan-Jung: Thank you for the review. The feedback you provided has been addressed in commit 210a5ec.

solr/core/src/java/org/apache/solr/highlight/UnifiedSolrHighlighter.java
```diff
@@ -314,6 +343,24 @@ protected PassageFormatter getFormatter(String fieldName) {
...
+
+      switch (passageSort) {
+        case START_OFFSET:
+          return Comparator.comparingInt(Passage::getStartOffset);
+        case END_OFFSET:
+          return Comparator.comparingInt(Passage::getEndOffset);
+        case SCORE:
+          return Comparator.comparingDouble(Passage::getScore);
+        default:
+          throw new AssertionError();
```
dsmiley, line 361: If a user supplies a bogus value, we should throw a SolrException with BAD_REQUEST as you can see elsewhere in this file.

dsmiley (commented):
It's somewhat unfortunate to have no Solr level test... but not strictly necessary.  You could probably add a query & assertion to a test method in TestUnifiedSolrHighlighter.  Just showing the param has an effect for one of the non-default options.

dsmiley:
Just a thought -- is this enum actually buying us anything at all?  It's a bunch of lines of code but only ultimately used for a set lookup and nothing else?  If so, lets just have less lines of code and do away with it.  switch-case on String literals and null if no arg provided.

Seunghan-Jung:
@dsmiley 
> Just a thought -- is this enum actually buying us anything at all? It's a bunch of lines of code but only ultimately used for a set lookup and nothing else? If so, lets just have less lines of code and do away with it. switch-case on String literals and null if no arg provided.

The model I followed is the HighlightComponent.getHighlighter method. 
https://github.com/apache/solr/blob/34f77886104a15f9e478cf7b4a5f6b0691d79eb7/solr/core/src/java/org/apache/solr/handler/component/HighlightComponent.java#L174-L194
I adopted this approach. This is also why I trigger an AssertionError in the default case of the switch statement.

Seunghan-Jung:
> It's somewhat unfortunate to have no Solr level test... but not strictly necessary. You could probably add a query & assertion to a test method in TestUnifiedSolrHighlighter. Just showing the param has an effect for one of the non-default options.

I was writing test code and discovered a bug that required a fix on the Lucene side, which I have now resolved. Here is the relevant PR: https://github.com/apache/lucene/pull/13832.

solr/core/src/java/org/apache/solr/highlight/UnifiedSolrHighlighter.java
Seunghan-Jung: It was fixed in this commit: https://github.com/apache/solr/pull/2697/commits/deafe902701bf33013a5c694fc1e85e9c7530818.

solr/core/src/java/org/apache/solr/highlight/UnifiedSolrHighlighter.java
```diff
@@ -314,6 +343,26 @@ protected PassageFormatter getFormatter(String fieldName) {
...
 
+    @Override
+    protected Comparator<Passage> getPassageSortComparator(String fieldName) {
+      String value = params.get(HighlightParams.PASSAGE_SORT, PassageSort.START_OFFSET.key);
+
+      PassageSort passageSort = PassageSort.parse(value);
+
+      switch (passageSort) {
+        case START_OFFSET:
+          return Comparator.comparingInt(Passage::getStartOffset);
```
Seunghan-Jung, line 354:
```suggestion
      switch (passageSort) {
        case "startOffset:
          return Comparator.comparingInt(Passage::getStartOffset);
```

Is this the direction you were aiming for? Are you suggesting that we avoid creating unnecessary enum class and use string literals instead?

Seunghan-Jung (commented):
@dsmiley 
> Just a thought -- is this enum actually buying us anything at all? It's a bunch of lines of code but only ultimately used for a set lookup and nothing else? If so, lets just have less lines of code and do away with it. switch-case on String literals and null if no arg provided.

The model I followed is the HighlightComponent.getHighlighter method. 
https://github.com/apache/solr/blob/34f77886104a15f9e478cf7b4a5f6b0691d79eb7/solr/core/src/java/org/apache/solr/handler/component/HighlightComponent.java#L174-L194
I adopted this approach. This is also why I trigger an AssertionError in the default case of the switch statement.

solr/core/src/java/org/apache/solr/highlight/UnifiedSolrHighlighter.java
dsmiley: Yes, because the enum isn't providing much of any value.  Isn't a useful "type".

Seunghan-Jung: It has been reflected in commit d5618e5.

solr/solr-ref-guide/modules/query-guide/pages/highlighting.adoc
```diff
@@ -509,6 +509,20 @@ Furthermore, phrases will be highlighted as a whole instead of word by word.  Cu
...
 If either `hl.usePhraseHighlighter` or `hl.multiTermQuery` are set to `false`, then this setting is effectively `false` no matter what you set it to.
 
+`hl.passageSort`::
++
+[%autowidth,frame=none]
+|===
+|Optional |Default: `startOffset`
+|===
++
+The UH uses the score to select the candidates for the number of snippets defined by `hl.snippets`,
```
dsmiley, line 519:
```suggestion
The UH calculates a score for each passage, and it uses it to pick the best snippets defined by `hl.snippets`,
```
(Users probably don't know passages have a score; they can't even see it after all)

solr/solr-ref-guide/modules/query-guide/pages/highlighting.adoc
```diff
@@ -509,6 +509,20 @@ Furthermore, phrases will be highlighted as a whole instead of word by word.  Cu
...
+but the final sorting of these selected candidates is determined by this option.
+The available choices are: `startOffset`, `endOffset`, and `score`.
```
dsmiley, line 521: Honestly is a user going to ever choose endOffset?  Could we simplify to only startOffset and maybe not even use that geeky internal word; maybe "natural" would be best understood?

dsmiley (approved):
+1 love the simplicity without the enum.

cpoerschke (approved):

dsmiley:
Proposed CHANGES.txt for 9.8, New Feature:

> SOLR-17444: The UnifiedHighlighter now supports an hl.passageSort option to control passage ordering per document startOffset|endOffset|score.  (Seunghan-Jung)

I plan to merge with that in a day or two.

dsmiley:
With Claude I tidied up some things and added a test.  If no further input to the contrary, I'll merge Monday.  Will be released in 10.1.

PR #2700, opened 2024-09-09
remove deprecated SolrParams methods

cpoerschke:
Noticed whilst code reading. The methods appear to have been deprecated in Solr 7.4 -- https://issues.apache.org/jira/browse/SOLR-11914 -- and so the removal here can be both for `main` and `branch_9x` branches.

### DIFF
--- a/solr/core/src/test/org/apache/solr/search/TestMinHashQParser.java
+++ b/solr/core/src/test/org/apache/solr/search/TestMinHashQParser.java
@@ -417,8 +417,7 @@ public class TestMinHashQParser extends SolrTestCaseJ4 {
     QParser qparser =
         h.getCore()
             .getQueryPlugin("minhash")
-            .createParser(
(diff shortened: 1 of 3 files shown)


solr/core/src/test/org/apache/solr/search/TestMinHashQParser.java
```diff
@@ -417,8 +417,7 @@ public void testBandsWrap() throws SyntaxError {
     QParser qparser =
         h.getCore()
             .getQueryPlugin("minhash")
-            .createParser(
-                "1, 2, 3, 4, 5, 6, 7, 8, 9, 10", SolrParams.toSolrParams(par), null, null);
+            .createParser("1, 2, 3, 4, 5, 6, 7, 8, 9, 10", par.toSolrParams(), null, null);
```
epugh, line 420: "par"?   Sort of a nitpick, but maybe a more descriptive variable name?  "params'?

epugh (approved):
LGTM.  Nice to see some deprecated code that has lingered get removed!

PR #2701, opened 2024-09-09
SOLR-17414: Multi-threaded search: don't throw RejectedExecutionException

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17414

This Executor is a bit special, and as such I don't think we should modify ExecutorUtil for the needs of one executor.   Backed out Christine's and Gus's change.  Instead, MDCAwareThreadPoolExecutor can be subclasses wherever -- did so in SolrIndexSearcher.

If we imagine this Executor is of general use, then its initialization should remain in CoreContainer.  I put it in SolrIndexSearcher but could move it.  IMO it is of general use; the multi-threaded faceting should use the same.  But that hasn't happened yet and I suspect nobody cares enough to do it so I'm leaving it be.


dsmiley:
Barring benchmarks & CHANGES.txt, I think this is ready to merge.

I changed to using a blocking strategy instead of caller-runs.  It should be easier to manage a Solr node's CPU consumption; more queries aren't going to tip it over (within reason).  But to maximize a machine's resources, it may be necessary to configure a multiple of the number of available CPUs; we'll see I guess.

I could see follow-on work to ensure there's prioritization of an earlier request's submissions over that of another concurrently happening slightly after.  But I suppose it doesn't *really* matter.  Also, maybe the search *itself* (i.e. thread processing SearchHandler) should use this executor?  Although it's not necessarily CPU-bound (does distributed-search) whereas segment-search is.  But we could do the distrib=false requests here.  Hmm.  An experiment for another day.

dsmiley:
CHANGES.txt: When searching with multiThreaded=true, the internal tasks may now block instead of enqueuing with a risk of rejection.  Solr will use less resources under stress but to get the most of your machine, you may want to increase the thread pool.

PR #1765, opened 2023-07-07
SOLR-16878: Add JUL Log4J manager for lucene logs

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-16878

This can be tested relatively easily, **using Java 20**

```
$ ./gradlew dev
$ cd solr/packaging/build/dev
$ ./bin/solr start -e films
```

In the solr logs (not the console logs), you will see:

```
INFO  (coreLoadExecutor-10-thread-1) [ x:films] o.a.l.s.MemorySegmentIndexInputProvider Using MemorySegmentIndexInput with Java 20; to disable start with -Dorg.apache.lucene.store.MMapDirectory.enableMemorySegments=false
```

For the new Vector logs, after introducing the option in Solr in #1758, no logging is done by default. But we can use the C1 compiler, which makes Lucene log that it can't use the Panama Vector API.

```
$ ./gradlew dev
$ cd solr/packaging/build/dev
$ ./bin/solr start -e films -a "-XX:TieredStopAtLevel=1"
```
In the solr logs (again, not the console logs), you will see:

```
WARN  (qtp1027495011-41) [ x:films] o.a.l.u.VectorUtilProvider C2 compiler is disabled; Java vector incubator API can't be enabled
```

PR #2703, opened 2024-09-10
SOLR-16976: Remove log4j-jul jar and use slf4j bridge for JUL

szhou1998:
https://issues.apache.org/jira/browse/SOLR-16976

# Description

https://github.com/apache/solr/pull/1765 introduced a change that causes an exception when remote JMX is enabled (stack trace included in Jira description).

# Solution

@elyograg provided a patch which resolves this issue by removing log4j-jul and using the slf4j bridge for JUL. I removed the NOCOMMIT from the patch.

# Tests

@elyograg tested the patch. Added a test that there's no exception on Solr startup with remote JMX enabled.

# Checklist


epugh:
I don't have much knowledge in how logging works.  One thing that would help, would you mind creating a BATS style test that demonstrates the bug and the fix?  I believe you could write a test that starts up solr with remote JMX, and then we see the bug, and then the fix.   The bats tests are in `./solr/packaging/test`.    Let me know if you have any questions!

szhou1998:
> I don't have much knowledge in how logging works. One thing that would help, would you mind creating a BATS style test that demonstrates the bug and the fix? I believe you could write a test that starts up solr with remote JMX, and then we see the bug, and then the fix. The bats tests are in `./solr/packaging/test`. Let me know if you have any questions!

I've removed the patch and added a test to start solr with remote JMX enabled. It should fail with the exception seen in the Jira.

epugh:
Humm...  I ran the "Solr Script Tests" expecting it to fail...    I can run it locally tomorrow and see what I see!

szhou1998:
Ah my mistake, Solr starts even though the exception is logged. Added a refute_output for the exception.

epugh:
okay, the refute wasn't actually matching..   i found that we need to consult the log file.  Also, I was getting a RMI Port exception..    Can you validate that this test fails for you, and hten add in your patch?

solr/bin/solr
```diff
@@ -1076,10 +1076,8 @@ if [ "${SOLR_HOME:0:${#EXAMPLE_DIR}}" = "$EXAMPLE_DIR" ]; then
   SOLR_LOGS_DIR="$SOLR_HOME/../logs"
 fi
 
-# Set the logging manager by default, so that Lucene JUL logs are included with Solr logs.
-LOG4J_CONFIG=("-Djava.util.logging.manager=org.apache.logging.log4j.jul.LogManager")
 if [ -n "${LOG4J_PROPS:-}" ]; then
-  LOG4J_CONFIG+=("-Dlog4j.configurationFile=$LOG4J_PROPS")
+  LOG4J_CONFIG=("-Dlog4j.configurationFile=$LOG4J_PROPS")
 fi
```
cpoerschke, line 1081: Wondering if `else` is needed to ensure `LOG4J_CONFIG` is always initialised?
```suggestion
else
  LOG4J_CONFIG=
fi
```

epugh: Do we have a bats test that would catch the `LOG4J_CONFIG` not being initialized and show if this is needed?  

There is a lot of tricky logic in the start command, and I'd love to have more tests so that as we refactor/change it we have more confidence that the tests are meaningful!

epugh:
Okay, so, after putting in the slf4j bridge, I see that the bats tests is now failing (and thats a good thing) because it means we don't see the error:

```
not ok 85 SOLR-16976 solr starts with remote JMX enabled # in 7641 ms
# (from function `assert_output' in file /home/runner/work/solr/solr/.gradle/node/packaging/node_modules/bats-assert/src/assert.bash, line 247,
#  in test file test/test_start_solr.bats, line 40)
#   `assert_output --partial 'java.lang.ClassNotFoundException: org.apache.logging.log4j.jul.LogManager'' failed
```

@szhou1998 would you mind flipping the `assert_output` to a `refute_output` and lets see if the test now passes!

solr/bin/solr
szhou1998: I added a commit to initialize LOG4J_CONFIG=() which is what it was originally https://github.com/apache/solr/pull/1765/files#diff-49c29c8f653341f48008c38f5f2cf970fa430cdca29181c819d7bdcfcc722980

solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
```diff
@@ -133,6 +134,9 @@ public SolrDispatchFilter() {}
   @Override
   public void init(FilterConfig config) throws ServletException {
     try {
+      SLF4JBridgeHandler.removeHandlersForRootLogger();
+      SLF4JBridgeHandler.install();
```
dsmiley, line 138: This logic should ideally be done as early as possible.  A ServletFilter's init method is not actually so early as we can do.  We could do CoreContainerProvider.init() which will start up CoreContainer _before_ the ServletFilter.  But we could _do earlier than that_ by **adding some XML based configuration in jetty.xml, which I think makes more sense**.  It could also mean we add the dependency JAR without formerly making it a dependency of solr-core, thus wouldn't leak to other/external things projects that might want to depend on solr-core.  This is a deployment matter so it's best done at the Jetty layer.

szhou1998: Would that change be in this file? https://github.com/apache/solr/blob/457ca3668a8906277f8f638af9c740b57f2a2a20/solr/server/etc/jetty.xml#L8
Do you know what it would look like?

dsmiley: A quick google search turned up [this](https://github.com/rapla/rapla/blob/master/src/test/etc/jetty.xml) immediately.

dsmiley:
Please stop force-pushing to PRs -- it resets the review state in GitHub.
Any way, I will merge this soon; thank you!

PR #2646, opened 2024-08-14
SOLR-17406: Introduce Version Catalogs

malliaridis:
https://issues.apache.org/jira/browse/SOLR-17406

> **IMPORTANT** Due to the extensive scope with the dependency updates made in this PR, a new PR without the updates was created, see #2706.

# Description

With Gradle it is possible to manage versions and dependencies in gradle files via version catalogs. This allows us to cleanup dependency resolution and move various versions from across the project to a single file.

# Solution

This PR introduces version catalogs and removes the palantir consistent versions plugin, since it is no longer needed. Additionally, the build module(s) are restructured and further changes from https://github.com/apache/lucene/pull/13484 were applied to reflect the same behavior.

# Tests

Minor adjustments have been made to some test files that fix some references (due to restructuring of build files), and besides that spotless and tidy are added to run on build files too.

# Open TODOs

- [X] Resolve dependency resolution conflicts
- [X] Update documentation about dependencies
  - [X] Include new notes for reviewing license and notice files (see [discussion thread](https://lists.apache.org/thread/0yhtfdlv4mlqfyzk7j19l1ojf84oh9ol))
- [X] Update license and notice files of updated dependencies
- [ ] Reapply forbidden APIs' signature logic in `forbidden-apis.gradle` (depends on https://github.com/policeman-tools/forbidden-apis/pull/249)
- [ ] Review dependency bot for breaking functionality
- [ ] Update Changelog file
- [ ] Verify the validity of the steps for "Update Lucene prerelease" (ASF account needed)

# Checklist

Please review the following and check all that apply:

- [X] I have reviewed the guidelines for [How to Contribute](https://github.com/apache/solr/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.
- [X] I have created a Jira issue and added the issue ID to my pull request title.
- [X] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [X] I have developed this patch against the `main` branch.
- [x] I have run `./gradlew check`.
- [ ] I have added tests for my changes.
- [ ] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

# Dependecy Updates and Related PRs

[...]

PR #2706, opened 2024-09-11
SOLR-17406: Introduce Version Catalogs

malliaridis:
https://issues.apache.org/jira/browse/SOLR-17406

# Description

With the version catalogs all versions and references to them are managed in a single file and updated from there.

# Solution

This PR introduces version catalogs and removes the palantir consistent versions plugin, since it is no longer needed. 

For syncing versions a new gradle file (dependencies.gradle) is introduced, where the version of a dependency (either transitive or non-transitive) is set via constraints. This part of the changes should be improved in the future if possible to reduce the maintenance overhead.

This PR is split from https://github.com/apache/solr/pull/2646.

This PR is following the practices from https://github.com/apache/lucene/pull/13484.

**⚠️ IMPORTANT ⚠️** Merging this PR will likely break Solrbot functionality and will require additional actions.

Note that this PR is not tested for backwards compatibility (9x).

# Tests

No tests should be affected by this.

# Checklist


help/dependencies.txt
```diff
@@ -29,60 +29,126 @@ testImplementation - makes a dependency only available for test  classes.
...
+foo-bar-baz = { module = "foo.bar:baz", version.ref = "foo-bar-baz" }
+
+Note that the used names separated by dashes are later referenced with dots
+instead of dashes, but more on that later.
+
+The chosen name for the module should more or less reflect the module's
+group name and module id in a way that it groups related dependencies under
+the same "prefix" (see below). There is no specific convention here and
+group prefixes for domain names like "com" and "io" are avoided, as they
+do not add any value and increase the size of the reference / alias name.
```
janhoy, line 53: Is this format of `libs.versions.toml` compatible with Dependabot or Renovatebot?

malliaridis: As far as I know both support gradle version catalogs, and the naming conventions in the catalogs should not affect the compatibility.

build.gradle
```diff
@@ -139,7 +143,7 @@ apply from: file('gradle/maven/defaults-maven.gradle')
 
 // IDE support, settings and specials.
 apply from: file('gradle/ide/intellij-idea.gradle')
-apply from: file('gradle/ide/eclipse.gradle')
+// apply from: file('gradle/ide/eclipse.gradle')
```
dsmiley, line 146: do we remove it then?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,432 @@
...
+ltgt-errorprone = { id = "net.ltgt.errorprone", version.ref = "ltgt-errorprone" }
+morethan-jmhreport = { id = "io.morethan.jmhreport", version.ref = "morethan-jmhreport" }
+nodegradle-node = { id = "com.github.node-gradle.node", version.ref = "nodegradle-node" }
+openapi-generator = { id = "org.openapi.generator", version.ref = "openapi" }
+owasp-dependencycheck = { id = "org.owasp.dependencycheck", version.ref = "owasp-dependencycheck" }
+swagger3-core = { id = "io.swagger.core.v3.swagger-gradle-plugin", version.ref = "swagger3" }
+thetaphi-forbiddenapis = { id = "de.thetaphi.forbiddenapis", version.ref = "thetaphi-forbiddenapis" }
+udnercouch-download = { id = "de.undercouch.download", version.ref = "undercouch-download" }
+
+[libraries]
```
dsmiley, line 188: ah well, we've got to follow what Gradle wants us to do now but IMO coming up with a unique library string for each library is annoying.  Maintaining this is annoying.

gradle/node.gradle
```diff
@@ -16,7 +16,7 @@
  */
 
 configure([project(":solr:packaging"), project(":solr:solr-ref-guide"), project(":solr:webapp")]) {
-    apply plugin: "com.github.node-gradle.node"
+    apply plugin: libs.plugins.nodegradle.node.get().pluginId
```
dsmiley, line 19: TBH I liked it as it was -- a well known string instead of some dotted namespace from the toml that we invent for each dependency.

help/dependencies.txt
dsmiley, line 53: to use libs.versions.toml; are we actually forced to add such invented names here and use them as such as well?

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/SolrZkClient.java
```diff
@@ -85,7 +85,7 @@ public class SolrZkClient implements Closeable {
   private int minStateByteLenForCompression;
 
   public MapWriter getMetrics() {
-    return metrics::writeMap;
+    return metrics;
```
dsmiley, line 88: Not needed in this PR, I assume

dsmiley (commented):
Huge effort on your part!  Even considering a ton of this was lifted from Lucene.

I have some dislike for having to come up with lib names in the TOML and referencing them everywhere instead of well-known "group:artifact" strings.  I'm not sure if we are forced into this?

To validate this, I'd want to look at the tar.gz contents before & after for ground-truth on what we ship.

dsmiley:
> spotless and tidy are added to run on build files too

Awesome!

gradle/libs.versions.toml
dweiss: You have to come up with your own name. Intellij will help you with suggestion prompts, if you're using it but otherwise you're down to using those custom names.

help/dependencies.txt
malliaridis: This is a good point and I agree with you. It is clearer if we use the "group:artifact" notation in most cases and just not provide the version information in the string. This would then pick the version from the version catalog as far as I understand it. I tried it with the strings before and it worked just fine.

There is however one risk with this approach that may also be a feature we want: if we do not force the developers to use the aliases, they may add at some point version information to dependencies without adding them to version catalogs. This could potentially lead to different versions of the same library (if the current plugins allow it, not sure), which is the case at the moment for some plugins for example (not checked by palantir's version plugin).

I am open to use strings instead, just want a few more opinions / confirmations for that.

build.gradle
malliaridis: Good catch, this slipt into the PR without me noticing it before I create the PR. I have to see if I can build the project with the `eclipse.gradle` file loaded, it caused some issues on my environment.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/SolrZkClient.java
malliaridis: It was causing linting errors in the CI/CD, but not locally, which is why I had to fix it. Same case for `p::writeMap` in another commit.

help/dependencies.txt
dweiss: > This would then pick the version from the version catalog as far as I understand it.

It would surprise me if it did it automatically... You can have a version catalog with multiple versions of the same artifact (with identical coordinates) - how would it choose the right one?

The rare upside of using those aliases is that if the coordinates of a library change (which infrequently happens), you only need to change it in one place. ;)

gradle/libs.versions.toml
janhoy: I agree with this. Lack of namespacing will get us into trouble. Would it be better if we agree on some simple convention to use the original coordinate with dots replaced by space? E.g.

```toml
[versions]
org-apache-commons_commons-compress = "1.26.1"
org-apache-lucene = "9.12.0"

[libraries]
org-apache-commons_commons-compress = { module = "org.apache.commons:commons-compress", version.ref = "org-apache-commons_commons-compress" }
org-apache-lucene_lucene-foo = { module = "org.apache.lucene:lucene-foo", version.ref = "org-apache-lucene" }
org-apache-lucene_lucene-bar = { module = "org.apache.lucene:lucene-bar", version.ref = "org-apache-lucene" }

```

build.gradle
malliaridis: So I uncommented it fixed some version references and a null error that was thrown on my side. Hopefully this works the same now for eclipse users.

gradle/libs.versions.toml
malliaridis: I think before we run into namespace limits we would have other issues first. 😅 However, even if a collection of libraries use the same group and eventually similar artifact namings, we always can add more words in the alias, like prefixes and suffixes. This is already the case for some libraries.

Using the fully qualified name would require renaming the alias and all places where it is used if a library changes group or artifact name (like when splitting into multiple modules). So I would still prefer a simplified alias, rather than the fully qualified name.

dweiss: You can do that, although there is a meaning to prefixes like this in gradle catalogs: they form artifact groups [1]. When I started using them, I followed a convention similar to yours... but over time I resigned from doing that. It wasn't really that helpful (or intuitive, with those dashes everywhere). 

[1] https://docs.gradle.org/current/userguide/platforms.html#sub:mapping-aliases-to-accessors

help/dependencies.txt
malliaridis: >  You can have a version catalog with multiple versions of the same artifact (with identical coordinates) - how would it choose the right one?

That's interesting, I didn't know that. 🤔 I guess it would follow the version resolution strategy like usual in that case, but I think even if that is the case, this is not what we want probably.

I guess going with the aliases is the best option we have so far.

dsmiley:
I checked out this branch and did `gw assemble` then renamed `solr/packaging/build/solr-10.0.0-SNAPSHOT/` to have an "-after" suffix.  Then I checked out a recent main (before cross-DC addition) and did `gw assemble`. 
Then compared: `diff -qr solr-10.0.0-SNAPSHOT/ solr-10.0.0-SNAPSHOT-after/`

There were many differences.  Just one example: the opentelemetry module didn't have netty libs but now it does.
Surely some differences are because main has moved along since when you started.

Can you please re-sync with main and repeat this exercise so we can hopefully get to no differences, or at least those differences that exist needs to be spelled out specifically and why.  Then lets merge this; I'll approve.

malliaridis:
@dsmiley thanks for the instructions, I didn't know before how to check the builds for consistency.

I am using now IntelliJ for a better overview and I see that the configurations that were included in `gradle/validation/dependencies.gradle` are not sufficient. I have now extended the list locally to the following:

```
def consolidatedConfigurations = project.configurations.matching {
    it.name in [
            "annotationProcessor",
            "api",
            "apiElements",
            "compileClasspath",
            "compileOnly",
            "libExt",
            "implementation",
            "runtimeClasspath",
            "runtimeElements",
            "runtimeOnly",
            "runtimeLibs",
            "serverLib",
            "testCompileClasspath",
            "testCompileOnly",
            "testImplementation",
            "testRuntimeClasspath",
            "testRuntimeOnly"
    ]
}
```
@dweiss any recommendations how to constellate this list of configurations?

Now, besides a few new libraries included in some modules that I need to figure out where they are coming from, the main differences are in `netty` libraries (consistently now `2.0.65.Final`),  and `google-api-grpc-proto` libraries (consistently now `2.42.0`).

The affected modules by these upgrades are:
- modules/gcs-modules
- modules/opentelemetry
- modules/s3-repository
- server/solr-webapp

I will pick up work again and try to solve this before the next review.

henrik242:
> > spotless and tidy are added to run on build files too
> 
> Awesome!

It's cool, but I really would have preferred to have those in a separate PR. It clutters this PR quite a lot.

malliaridis:
> It's cool, but I really would have preferred to have those in a separate PR. It clutters this PR quite a lot.

I totally agree with you. I should have tried to separate these two changes. The reason I ended up doing it in this PR is the reference I used, which was https://github.com/apache/lucene/pull/13484. I couldn't separate one from the other, so I decided to do it in a similar way.

About the configurations and dependencies: I am still not very sure about the configurations that we should sync the dependencies / include in the consolidated configurations list, so I ended up with the below state.

**Updated dependencies**
- `hamcrest:hamcrest-core` is replaced with `hamcrest:hamcrest` - module was renamed and included twice with different names before
- `netty-tcnative-broingssl-static` was updated through version syncing from `2.0.61.Final` to `2.0.65.Final`
- `netty-tcnative-classes` was updated through version syncing from `2.0.61.Final` to `2.0.65.Final`
- `proto-google-common-protos` was updated through  version syncing from `2.41.0` to `2.42.0`

**module: gcs-repository**
- uses the updated `proto-google-common-protos` library (version `2.42.0`)

**module: opentelemetry**
- uses the updated `proto-google-common-protos` library (version `2.42.0`)

**server/solr-webapp**
- `netty-tcnative-broingssl-static` was updated through version syncing from `2.0.61.Final` to `2.0.65.Final`
- `netty-tcnative-classes` was updated through version syncing from `2.0.61.Final` to `2.0.65.Final`

@dsmiley the diff output should now show only the above changes if I have done everything correct.

dsmiley:
Thanks Christos.

I suggest a separate PR for the version bumping listed.  It'd also be nice to backport that to 9x.

May I suggest that we have a separate PR that merely formats our gradle files?  Could be just a quick automated thing.  It'll make this PR much easier to read and moreover, over time, we'll see what this change here did, not obscured by formatting.  A sweeping formatting change like this would not go to 9x.

Finally, this PR here is then just version catalogs refactoring, and only happens in v10.

dsmiley:
BTW if you are worried about how to update this PR if those other things merge; I could figure that out.  The end-state will be no different, so it should be possible to do a bit of git magic on the merge from main into this to say that this side (the PR) "wins" for all differing files.

malliaridis:
Thanks for the input and considering handling non-relevant changes in separate PRs. I feel more at ease if we do so, so I will continue with that and create these PRs as suggested before continuing work on this.

Changing this to a draft PR for now.

dsmiley:
I think the next step is just some dependency version alignment?

malliaridis:
I think it would be a good idea to first resolve the current solrbot PRs before aligning the versions. I am already looking into some of the PRs to see how they can be resolved.

malliaridis:
The current state should differ now only in hamcrest-core, which is being merged with hamcrest in the `dependencies.gradle`.

solr/licenses/netty-NOTICE.txt
```diff
@@ -162,9 +170,9 @@ This product optionally depends on 'JBoss Marshalling', an alternative Java
 serialization API, which can be obtained at:
 
   * LICENSE:
-    * license/LICENSE.jboss-marshalling.txt (GNU LGPL 2.1)
+    * license/LICENSE.jboss-marshalling.txt (Apache License 2.0)
```
dsmiley, line 173: I checked this -- looks accurate.  Then I got curious and searched out entire licenses folder for "LGPL".  netty-transport-native-epoll-NOTICE.txt (and unix-common-NOTICE.txt) also makes a reference to LGPL but it's via the same jboss-marshalling line you are correcting here.  So I think they can also be updated.

dsmiley (approved):
What prompted you to update the NOTICE files of existing dependencies?  -- very thorough of you but it would not have occurred to me.  I found a couple misattributed LGPL references to correct but no other issue.  

I ran the bats/integration tests and docker tests, and build the ref guide, just to exercise some parts of our build that might not otherwise in the PR validation.  I think we should merge this very soon.  A shout-out of the change arriving to our dev list would be helpful for general awareness and its impacts (e.g. dependency upgrades are now different and won't backport).

solr/core/src/java/org/apache/solr/update/processor/ClassificationUpdateProcessorFactory.java
```diff
@@ -92,8 +92,7 @@ public void init(final NamedList<?> args) {
       String algorithmString = params.get(ALGORITHM_PARAM);
       Algorithm classificationAlgorithm;
       try {
-        if (algorithmString == null
-            || Algorithm.valueOf(algorithmString.toUpperCase(Locale.ROOT)) == null) {
```
risdenk, line 96: This doesn't look related to version catalogs?

solr/core/src/test/org/apache/solr/handler/admin/IndexSizeEstimatorTest.java
```diff
@@ -227,8 +227,9 @@ public void testIntegration() throws Exception {
           (k, v) -> {
             double size = fromHumanReadableUnits((String) v);
             double sampledSize = fromHumanReadableUnits((String) sampledFieldsBySize.get(k));
-            assertNotNull(
```
risdenk, line 230: Why did this test change?

solr/core/src/test/org/apache/solr/handler/admin/api/ReloadCoreAPITest.java
```diff
@@ -57,7 +57,6 @@ public void setUp() throws Exception {
   public void testValidReloadCoreAPIResponse() throws Exception {
     SolrJerseyResponse response = reloadCoreAPI.reloadCore(coreName, new ReloadCoreRequestBody());
     assertEquals(0, response.responseHeader.status);
-    assertNotNull(response.responseHeader.qTime);
```
risdenk, line 60: Why did this test change?

solr/core/src/test/org/apache/solr/handler/admin/api/UnloadCoreAPITest.java
```diff
@@ -56,7 +56,6 @@ public void setUp() throws Exception {
   public void testValidUnloadCoreAPIResponse() throws Exception {
     SolrJerseyResponse response = unloadCoreAPI.unloadCore(coreName, getUnloadCoreRequestBodyObj());
     assertEquals(0, response.responseHeader.status);
-    assertNotNull(response.responseHeader.qTime);
```
risdenk, line 59: Why did this test change?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,460 @@
...
+commons-collections = "3.2.2"
+commons-io = "2.15.1"
+cutterslade-analyze = "1.10.0"
+cybozulabs-langdetect = "1.1-20120112"
+diffplug-spotless = "6.5.2"
+dropwizard-metrics = "4.2.26"
+eclipse-ecj = "3.39.0"
+eclipse-jetty = "10.0.22"
+eclipse-jettytoolchain = "4.0.6"
+# @keep jgit version used by git-statuts.gradle
```
risdenk, line 87: typo `git-status.gradle`

solr/core/src/test/org/apache/solr/handler/admin/IndexSizeEstimatorTest.java
malliaridis: According to error-prone:

> error: [ImpossibleNullComparison] This value cannot be null, and comparing it to null may be misleading.

This is because primitive data types have always a default value. Resolving this to focus on the root discussion.

solr/core/src/test/org/apache/solr/handler/admin/api/ReloadCoreAPITest.java
malliaridis: According to error-prone:

> error: [ImpossibleNullComparison] This value cannot be null, and comparing it to null may be misleading.

This is because primitive data types have always a default value. Resolving this to focus on the root discussion.

solr/core/src/test/org/apache/solr/handler/admin/api/UnloadCoreAPITest.java
malliaridis: According to error-prone:

> error: [ImpossibleNullComparison] This value cannot be null, and comparing it to null may be misleading.

This is because primitive data types have always a default value. Resolving this to focus on the root discussion.

gradle/libs.versions.toml
malliaridis: Good spot, thank you :D

solr/core/src/test/org/apache/solr/handler/admin/api/UnloadCoreAPITest.java
risdenk: We need to be careful with errorprone updates. See https://github.com/apache/solr/blob/main/gradle/validation/error-prone.gradle#L68 we want to make sure we list and explicitly enable/disable checks. Just upgrading means we aren't going to handle all the checks we may not want to run since they are either covered by other tools (ie: forbidden apis) or they are not necessary for our codebase.

solr/core/src/java/org/apache/solr/update/processor/ClassificationUpdateProcessorFactory.java
risdenk: We need to be careful with errorprone updates. See https://github.com/apache/solr/blob/main/gradle/validation/error-prone.gradle#L68 we want to make sure we list and explicitly enable/disable checks. Just upgrading means we aren't going to handle all the checks we may not want to run since they are either covered by other tools (ie: forbidden apis) or they are not necessary for our codebase.

malliaridis: I agree that we should not take error-prone updates lightly. But we still have to update and sync error-prone versions one way or another.

Would it suffice in your opinion if we update error-prone to 2.28.0 in a separate PR first before proceeding here?

risdenk: I agree that should be done first so we can make sure that all the rules are put into https://github.com/apache/solr/blob/main/gradle/validation/error-prone.gradle#L68

malliaridis: With the changes from https://github.com/apache/solr/pull/2862, this should no longer be an issue in this PR. As mentioned in the PR, further rule evaluation is necessary to enable a few more useful rules for the future.

risdenk (approved):

dev-docs/dependency-upgrades.adoc
```diff
@@ -16,30 +16,34 @@
...
 1. Identify the available versions from e.g. https://search.maven.org[Maven Central]
-2. Update the version in `versions.props` file
-3. Run `./gradlew --write-locks` to re-generate `versions.lock`. Note that this may cause a cascading effect where
+2. Update the version in `gradle/libs.versions.toml` file
+3. Run `./gradlew writeLocks` to re-generate `versions.lock`. Note that this may cause a cascading effect where
    the locked version of other dependencies also change.
-4. Run `./gradlew updateLicenses` to re-generate SHA1 checksums of the new jar files.
-5. Once in a while, a new version of a dependency will transitively bring in brand-new dependencies.
+4. In case of a conflict, resolve the conflict according to `help/dependencies.txt`
+5. Check if there are any constraints that are obsolete after the dependency update
```
gus-asf, line 36: Check where? How does one identify an obsolete constraint?

help/dependencies.txt
```diff
@@ -100,12 +166,12 @@ If you want to upgrade Lucene to a newer build proceed like the following:
   queued)
 - remember the build number of Jenkins (left side, first build in list,
   prefixed by '#')
-- Edit ./versions.props and change Lucene's version to '9.0.0-prereleaseX',
+- Edit gradle/libs.versions.toml and change Lucene's version to '9.0.0-prereleaseX',
```
gus-asf, line 169: When I look at versions.props we don't seem to be doing  this. Is this a legacy from pre-split? Maybe this section should go away since Lucene definitely marches ahead on it's own now? (Main is not using 10.0 let alone a pre-release)

dev-docs/dependency-upgrades.adoc
```diff
@@ -16,30 +16,34 @@
...
+6. Update the license and notice files of the changed dependencies if necessary. See `help/dependencies.txt` for
```
gus-asf, line 37: Define necessary...

gus-asf (commented):
Just looked at some of the instructions didn't have time to review all, but the upgrade steps look like they could use some clarification, and I think there's an obsolete section in dependencies.txt that you are updating instead of removing.

dev-docs/dependency-upgrades.adoc
malliaridis: Generally said, any difference in the LICENSE or NOTICE file of the affected dependencies is a change we have to include.

Updating LICENSE and NOTICE files is also subject to change, see [SOLR-15929](https://issues.apache.org/jira/browse/SOLR-15929).

gus-asf: This is detail that should be provided in the doc so that there's no variation in what people think is "necessary", even if it's just a reference to the doc mentioned in that ticket.

malliaridis:
I have updated the documentation to hopefully be clearer, but it is still not perfect yet. I have created https://issues.apache.org/jira/browse/SOLR-17563 to address outdated documentation.

PR #2708, opened 2024-09-12
SOLR-17150: Create MemAllowedLimit

sigram:
This is a refreshed PR using the latest QueryLimit APIs and a thread-safe accumulation of value, similar to the approach in `CpuAllowedLimit`. However, since that class uses a thread-local managed by `ThreadCpuTimer` this implementation adds its own thread-local.


solr/core/src/java/org/apache/solr/search/MemAllowedLimit.java
```diff
@@ -0,0 +1,130 @@
...
+  @Override
+  public boolean shouldExit() {
+    if (exitedAt > 0L) {
+      return true;
+    }
+
+    try {
+      long currentAllocatedBytes = (Long) GET_BYTES_METHOD.invoke(threadBean);
+      AtomicLong threadMem = threadLocalMem.get();
+      threadMem.compareAndExchange(-1L, currentAllocatedBytes);
```
gus-asf, line 104: compareAndSet is a more common method with an intuitive name that the reader may not have to look up. Since no return is checked here it would be sufficient.

solr/core/src/java/org/apache/solr/search/MemAllowedLimit.java
```diff
@@ -0,0 +1,130 @@
...
+      long lastAllocatedBytes = threadMem.get();
+      accumulatedMem.addAndGet(currentAllocatedBytes - lastAllocatedBytes);
```
gus-asf, line 106: On the first pass this appears to be zero, meaning that you are ignoring bytes allocated before the first invocation of the limit check? Is that the intent?

This class needs javadoc clearly stating what it intends to measure, so that future maintainers know what not to break (or if they want to break what you did on purpose). As it is I find it hard to evaluate the logic, and I am left shrugging and saying "I guess he meant to do that"

solr/solr-ref-guide/modules/query-guide/pages/common-query-parameters.adoc
```diff
@@ -382,11 +382,20 @@ This parameter specifies the amount of CPU time, in milliseconds, allowed for a
...
+Solr nodes (e.g. getting top document id-s, retrieving their fields, additional stages may be required for faceting, grouping, etc).
 For example, setting `cpuAllowed=500` gives a limit of at most 500 ms of CPU time for each of these stages - meaning that the total CPU usage by the query may reach a multiple of the `cpuAllowed` value depending on the number of stages.
 
 All other considerations regarding partial results listed for the `timeAllowed` parameter apply here, too.
 
+== memAllowed Parameter
+
+This parameter specifies the amount of memory (a float value, in MiB) allowed for a search thread to allocate
+during query execution. Similarly to the `cpuAllowed` this parameter monitors the additional memory
+allocated to the thread that executes the current query. Similarly, the same hard limit is applied to
```
gus-asf, line 395: Similarly used twice... which is strange (bad writing style) and that leaves me wondering if there's something edited incorrectly here?

Also I'm not sure if comparing it to `cpuAllowed` is helpful here. Might be clearer to just describe what it does directly.

solr/solr-ref-guide/modules/query-guide/pages/common-query-parameters.adoc
```diff
@@ -382,11 +382,20 @@ This parameter specifies the amount of CPU time, in milliseconds, allowed for a
...
+the query coordinator as to each replica that participates in the distributed search. As is the case with
+the `cpuAllowed` also here the limit is applied separately to each stage of the distributed query processing.
```
gus-asf, line 397: stage is a term not necessarily well defined from the reader's perspective.

gus-asf (changes requested):
Lack of doc leaves me with some doubt as to the efficacy of my review. Biggest suggestion is to add Javadoc and clarify the asciidoc

solr/core/src/java/org/apache/solr/search/MemAllowedLimit.java
sigram: > On the first pass this appears to be zero, meaning that you are ignoring bytes allocated before the first invocation of the limit check? Is that the intent?

Fixed. Added javadoc.

solr/solr-ref-guide/modules/query-guide/pages/common-query-parameters.adoc
sigram: But it's already used in other places. What should we use instead, in order to indicate that this limit is applied separately to each .. well, stage? phase? ... in distributed query processing?

solr/solr-ref-guide/modules/query-guide/pages/common-query-parameters.adoc
sigram: The section under `cpuAllowed` already describes what are the examples of stages, I wanted to avoid repeating this explanation here.

solr/core/src/java/org/apache/solr/search/MemAllowedLimit.java
```diff
@@ -76,11 +93,31 @@ public MemAllowedLimit(SolrQueryRequest req) {
...
+    if (!supported) {
+      throw new IllegalArgumentException(
+          "Per-thread memory allocation monitoring not available in this JVM.");
+    }
     limitBytes = Math.round(memLimit * MEBI);
+    // init the tread-local
+    initAndGetCurrentAllocatedBytes();
+  }
+
+  private long initAndGetCurrentAllocatedBytes() {
```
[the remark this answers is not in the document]
sigram: This method is private, nothing can override it.

gus-asf (approved):

PR #2710, opened 2024-09-13
bin/solr in branch_9_7 has commands that fail

epugh:

# Description

Fixes some bugs that existing in branch_9_7.
* -p port numbers
* bin/solr healthcheck is broken

This has good stuff to port back to main and branch_9x...

# Tests

New Bats tests added

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cli/CreateTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/CreateTool.java
@@ -370,25 +370,25 @@ public class CreateTool extends ToolBase {
           "Failed to create collection '" + collectionName + "' due to: " + sse.getMessage());
     }
 
-    if (cli.hasOption(SolrCLI.OPTION_VERBOSE.getOpt())) {
+    if (isVerbose()) {
       // pretty-print the response to stdout
       CharArr arr = new CharArr();
       new JSONWriter(arr, 2).write(response.asMap());
       echo(arr.toString());
-    } else {
-      String endMessage =
-          String.format(
-              Locale.ROOT,
-              "Created collection '%s' with %d shard(s), %d replica(s)",
-              collectionName,
-              numShards,
-              replicationFactor);
-      if (confName != null && !confName.trim().isEmpty()) {
-        endMessage += String.format(Locale.ROOT, " with config-set '%s'", confName);
-      }
-
-      echo(endMessage);
     }
+
+    String endMessage =
+        String.format(
+            Locale.ROOT,
+            "Created collection '%s' with %d shard(s), %d replica(s)",
+            collectionName,
+            numShards,
+            replicationFactor);
+    if (confName != null && !confName.trim().isEmpty()) {
+      endMessage += String.format(Locale.ROOT, " with config-set '%s'", confName);
+    }
+
+    echo(endMessage);
   }
 
   private Path getFullConfDir(Path solrInstallDir, Path confDirName) {
--- a/solr/core/src/java/org/apache/solr/cli/CreateTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/CreateTool.java
@@ -45,6 +45,7 @@ import org.apache.solr.cloud.ZkConfigSetService;
 import org.apache.solr.common.cloud.ZkStateReader;
 import org.apache.solr.common.params.CollectionAdminParams;
 import org.apache.solr.common.params.CommonParams;
+import org.apache.solr.common.util.EnvUtils;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.core.ConfigSetService;
 import org.noggit.CharArr;
@@ -148,6 +149,19 @@ public class CreateTool extends ToolBase {
             .required(false)
             .desc("Configuration name; default is the collection name.")
             .build(),
+        Option.builder("p")
+            .longOpt("port")
+            .deprecated(
+                DeprecatedAttributes.builder()
+                    .setForRemoval(true)
+                    .setSince("9.7")
+                    .setDescription("Use --solr-url instead")
+                    .get())
+            .argName("PORT")
+            .hasArg()
+            .required(false)
+            .desc("Port of a local Solr instance where you want to create the new core.")
+            .build(),
         SolrCLI.OPTION_SOLRURL,
         SolrCLI.OPTION_SOLRURL_DEPRECATED,
         SolrCLI.OPTION_ZKHOST,
@@ -157,6 +171,12 @@ public class CreateTool extends ToolBase {
   @Override
   public void runImpl(CommandLine cli) throws Exception {
     SolrCLI.raiseLogLevelUnlessVerbose(cli);
+
+    // If a port is provided, inject it into the environment variables so
+    // that {@link SolrCLI#getDefaultSolrUrl()} can use it instead of the default 8983.
+    if (cli.hasOption("p")) {
+      EnvUtils.setProperty("jetty.port", cli.getOptionValue("p"));
+    }
 
     try (var solrClient = SolrCLI.getSolrClient(cli)) {
       if (SolrCLI.isCloudMode(solrClient)) {
--- a/solr/core/src/java/org/apache/solr/cli/CreateTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/CreateTool.java
@@ -165,7 +165,8 @@ public class CreateTool extends ToolBase {
         SolrCLI.OPTION_SOLRURL,
         SolrCLI.OPTION_SOLRURL_DEPRECATED,
         SolrCLI.OPTION_ZKHOST,
-        SolrCLI.OPTION_ZKHOST_DEPRECATED);
+        SolrCLI.OPTION_ZKHOST_DEPRECATED,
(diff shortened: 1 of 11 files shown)


epugh:
Look at that!  in 9.6, `-V` is passed in as verbose to the CreateTool.   https://github.com/apache/solr/blob/branch_9_6/solr/bin/solr#L1185

epugh:
Found that HealthcheckTool isn't working in 9.7 either..   and our Bats test doesn't refute the right "ERROR", only "error" ;-).

epugh:
Okay, I think this is ready for review and commit.  I am thinking that I move the bats tests in another PR to `branch_9x` to see what breaks there, and then do a PR to fix `branch_9x`.  Then, I take the good stuff here and apply it to `main`...   Versus doing it the other way...

solr/core/src/java/org/apache/solr/cli/HealthcheckTool.java
```diff
@@ -170,7 +172,15 @@ protected void runCloudTool(CloudSolrClient cloudSolrClient, CommandLine cli) th
           q = new SolrQuery("*:*");
           q.setRows(0);
           q.set(DISTRIB, "false");
-          try (var solrClientForCollection = SolrCLI.getSolrClient(coreUrl)) {
+
+          // SolrCLI.getSolrClient converts the coreUrl back into a root Solr url, but we need to
+          // talk to specific core so manually creating the client.
```
janhoy, line 177: Hmm, would you say this is a bug, or is it well documented in the javadoc?

solr/core/src/java/org/apache/solr/cli/ToolBase.java
```diff
@@ -34,18 +34,29 @@ protected ToolBase(PrintStream stdout) {
...
-    if (cli.hasOption(SolrCLI.OPTION_VERBOSE.getOpt())) {
+    if (verbose) {
       echo(msg);
     }
   }
 
+  /**
+   * Is this tool being run in a verbose mode?
+   *
+   * @return verbose or not
```
janhoy, line 45: Unnecessary `@return` doc as the method name clearly indicates what is returned? :)

solr/packaging/test/test_create.bats
```diff
@@ -34,8 +34,31 @@ teardown() {
...
+  run solr create -c COLL_NAME -p ${SOLR2_PORT}
+  assert_output --partial "Created new core 'COLL_NAME'"
+}
+
+@test "ensure -V verbose parameter is supported" {
+  solr start 
+  solr assert --started http://localhost:${SOLR_PORT} --timeout 5000
+  
+  run solr create -c COLL_NAME -V
+  assert_output --partial "Created new core 'COLL_NAME'"
```
janhoy, line 63: Should not this assert a string that is only printed in verbose mode?

solr/packaging/test/test_delete_collection.bats
```diff
@@ -77,3 +77,15 @@ teardown() {
   solr delete -c "COLL_NAME" --delete-config false
   assert config_exists "COLL_NAME"
 }
+
+@test "ensure -p port parameter is supported" {
+
+  #solr start -c -p ${SOLR2_PORT}
+  #solr assert --started http://localhost:${SOLR2_PORT} --timeout 5000
```
janhoy, line 84: Remove dead code

solr/packaging/test/test_healthcheck.bats
```diff
@@ -29,8 +29,9 @@ teardown() {
 }
 
 @test "healthcheck on cloud solr" {
-  solr start -c -e films
-  run solr healthcheck -c films
+  solr start -c -e films -p ${SOLR_PORT}
+  run solr healthcheck -c films -zkhost localhost:${ZK_PORT} -V
   refute_output --partial 'error'
```
janhoy, line 34: Did you say that the refute should be case insensitive? Is that even supported in BATS?

janhoy (commented):
Thanks for quick PR. Looking good, not taken it for a spin, but very nice with the BATS tests.

Will we need edit to `solr.cmd` you think?

solr/packaging/test/test_healthcheck.bats
epugh: so, when I added a `--partial 'ERROR'` it didn't help because we actually put that message out in a legitimate run!   So I backed out that change.   There is a path where the healthcheck fails at the Java layer, and writes out "ERROR", and we don't see that due to our refute.   I guess a better approach is to look for positive things ;-)

solr/packaging/test/test_create.bats
epugh: Good question...    I think for core, you got this message, and for collection, you got a different dump (that was more like the api response).   I changed it so that both core and collection output a short message like this, and then if you add the `-v` then you ALSO see the api responses...   Does this work for you.

solr/core/src/java/org/apache/solr/cli/ToolBase.java
epugh: sure...  my javadoc looks pretty skinny with no `@return`

solr/core/src/java/org/apache/solr/cli/HealthcheckTool.java
epugh: I looked a bit more at the code, and decided to go with the javadoc approach!

epugh:
> Thanks for quick PR. Looking good, not taken it for a spin, but very nice with the BATS tests.
> 
> Will we need edit to `solr.cmd` you think?

Checked and yes, we had to fix healthcheck in `solr.cmd`.

solr/packaging/test/test_create.bats
janhoy: Sure, I just expected the assert to fail without the `-V`, which it won't now since the text asserted is always printed?

janhoy (approved):

solr/packaging/test/test_create.bats
epugh: I finally understand what you mean, and am looking at it and then will merge.

solr/core/src/java/org/apache/solr/cli/CreateTool.java
```diff
@@ -148,16 +149,36 @@ public List<Option> getOptions() {
...
+            .hasArg()
+            .required(false)
+            .desc("Port of a local Solr instance where you want to create the new core.")
+            .build(),
         SolrCLI.OPTION_SOLRURL,
         SolrCLI.OPTION_SOLRURL_DEPRECATED,
         SolrCLI.OPTION_ZKHOST,
-        SolrCLI.OPTION_ZKHOST_DEPRECATED);
+        SolrCLI.OPTION_ZKHOST_DEPRECATED,
+        SolrCLI.OPTION_VERBOSE_DEPRECATED);
```
malliaridis, line 169: Should this be eventually
```suggestion
        SolrCLI.OPTION_ZKHOST_DEPRECATED,
        SolrCLI.OPTION_VERBOSE,
        SolrCLI.OPTION_VERBOSE_DEPRECATED);
```

solr/core/src/java/org/apache/solr/cli/DeleteTool.java
```diff
@@ -122,16 +124,36 @@ public List<Option> getOptions() {
...
+            .hasArg()
+            .required(false)
+            .desc("Port of a local Solr instance where you want to create the new core.")
+            .build(),
         SolrCLI.OPTION_SOLRURL,
         SolrCLI.OPTION_SOLRURL_DEPRECATED,
         SolrCLI.OPTION_ZKHOST,
-        SolrCLI.OPTION_ZKHOST_DEPRECATED);
+        SolrCLI.OPTION_ZKHOST_DEPRECATED,
+        SolrCLI.OPTION_VERBOSE_DEPRECATED);
```
malliaridis, line 144: Should this be eventually
Same here:
```suggestion
        SolrCLI.OPTION_ZKHOST_DEPRECATED,
        SolrCLI.OPTION_VERBOSE,
        SolrCLI.OPTION_VERBOSE_DEPRECATED);
```

solr/packaging/test/test_create.bats
```diff
@@ -34,8 +34,31 @@ teardown() {
...
+  
+  run solr create -c COLL_NAME -p ${SOLR2_PORT}
+  assert_output --partial "Created new core 'COLL_NAME'"
+}
+
+@test "ensure -V verbose parameter is supported" {
+  solr start 
+  solr assert --started http://localhost:${SOLR_PORT} --timeout 5000
+  
+  run solr create -c COLL_NAME -v
```
malliaridis, line 62: `-V` (uppercase) is mentioned, but `-v` (lowercase) is tested. I am not sure if case-sensitivity is a thing at the current CLI parser, but it should be at least synced here.

```suggestion
  run solr create -c COLL_NAME -V
```

malliaridis (changes requested):
A few small additions from my side.

gus-asf:
Can we update the title of this? is it just about CLI stuff? hard to know what this is about...

epugh:
This was committed into branch_9x in #35ae5f7c55c27f463db048b975d06ec070fbbeef

PR #2712, opened 2024-09-13
SOLR-17450 StatusTool with pure Java code

janhoy:
https://issues.apache.org/jira/browse/SOLR-17450

This moves the meat of `bin/solr status` command to StatusTool, removing some `ps | grep java` logic in both scripts.

Instead, we use the Java 9 `ProcessHandle` API to interact with list of operating system processes, so we can find running instances of Solr.

The new `SolrProcessMgr` class is made standalone since it can also be used elsewhere to interact with solr processes...

I updated `test_status.bats`, but did not keep the old logic of allowing no arguments. Instead we expose botn `--solr-url` and `-p` so you can ask for status for another Solr. If you supply no args, it behaves like before, scanning for all running solrs. I added these to bats.

The output is probably not 100% same as it used to.

Windows script changes not tested at all...


solr/core/src/java/org/apache/solr/cli/SolrProcessMgr.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.Optional;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.solr.common.util.EnvUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Class to interact with Solr OS processes */
+public class SolrProcessMgr {
```
epugh, line 41: Nit picky, but how about `SolrProcessManager`?   Is `Mgr` a short cut we say a lot...?

solr/core/src/java/org/apache/solr/cli/StatusTool.java
```diff
@@ -65,37 +74,119 @@ public String getName() {
...
-        // therefore we don't use the SolrCLI.OPTION_SOLRURL.
-        Option.builder()
-            .argName("URL")
-            .longOpt("solr-url")
-            .hasArg()
-            .required(false)
-            .desc("Property set by calling scripts, not meant for user configuration.")
-            .build(),
-        OPTION_MAXWAITSECS);
+    return List.of(OPTION_SOLRURL, OPTION_MAXWAITSECS, OPTION_PORT, OPTION_SHORT);
```
epugh, line 96: this is a nice clean up..  I love that we flip the script and embrace the argument.  the whole "hiding" it thing was a total pain in the past..   And now we get MORE capability out of the tool!

solr/core/src/java/org/apache/solr/cli/StatusTool.java
```diff
@@ -65,37 +74,119 @@ public String getName() {
...
   }
 
   @Override
   public void runImpl(CommandLine cli) throws Exception {
-    // Override the default help behaviour to put out a customized message that only list user
```
epugh, line 86: and great to see all of this go away as well!

solr/core/src/java/org/apache/solr/cli/StatusTool.java
```diff
@@ -65,37 +74,119 @@ public String getName() {
...
-    // settable Options.
-    if ((cli.getOptions().length == 0 && cli.getArgs().length == 0)
-        || cli.hasOption("h")
-        || cli.hasOption("help")) {
-      final Options options = new Options();
-      options.addOption(OPTION_MAXWAITSECS);
-      new HelpFormatter().printHelp("status", options);
-      return;
+    String solrUrl = cli.getOptionValue(OPTION_SOLRURL);
```
epugh, line 101: we need to check the deprecated option as welL?   Actually, I guess not!!!!

solr/core/src/java/org/apache/solr/cli/StatusTool.java
```diff
@@ -65,37 +74,119 @@ public String getName() {
...
+    Integer port =
+        cli.hasOption(OPTION_PORT) ? Integer.parseInt(cli.getOptionValue(OPTION_PORT)) : null;
+    boolean shortFormat = cli.hasOption(OPTION_SHORT);
+
+    if (port != null && solrUrl != null) {
```
epugh, line 106: it really does seem like commons-cli should help us on this logic..  sigh..  we have other places where this is the case.

solr/core/src/java/org/apache/solr/cli/StatusTool.java
```diff
@@ -65,37 +74,119 @@ public String getName() {
...
+      CLIO.err("Please provide either --port or --solr-url, not both");
```
epugh, line 107: you know, other places we throw an exception instead of doing this...   and that percolates up and does the exit...  Something about IllegalStateException maybe...

solr/core/src/test/org/apache/solr/cli/SolrProcessMgrTest.java
epugh, line 1: Wow!  Some great tests..   I don't know that I could ever have written, but I loved reading it!

solr/packaging/test/test_help.bats
```diff
@@ -60,10 +60,8 @@ setup() {
 
 @test "status help flag prints help" {
   run solr status --help
-  assert_output --partial 'usage: status'
+  assert_output --partial 'usage: bin/solr status [--max-wait-secs <SECS>] [-p <PORT>] [--short] [-url <HOST>]'
```
epugh, line 63: love it!

epugh (approved):
Loving it...

epugh:
We can put a call out to the community for testing on Windows...  @Willdotwhite would you be interested in taking this PR for a spin on Windows?

Willdotwhite:
> We can put a call out to the community for testing on Windows... @Willdotwhite would you be interested in taking this PR for a spin on Windows?

I'd be happy to, although I'm away from my Windows machine until next weekend at the earliest I'm afraid! Is there any particular time pressure on this?

epugh:
> > We can put a call out to the community for testing on Windows... @Willdotwhite would you be interested in taking this PR for a spin on Windows?
> 
> I'd be happy to, although I'm away from my Windows machine until next weekend at the earliest I'm afraid! Is there any particular time pressure on this?

It's open source and volunteer driven ;-).   So any time is a great time!   

I do appreciate you and @rahulgoswami stepping up to do the Windows verification work.  It makes me a lot more confident about being able to evolve the CLI.

solr/core/src/java/org/apache/solr/cli/SolrProcessMgr.java
janhoy: Agree. Also perhaps it should not be manager at all, unless it perhaps grows in the future to also be able to kill processes? Listing / inspecting them only would perhaps qualify for a name like `RunningSolrProcesses`?

solr/core/src/java/org/apache/solr/cli/StatusTool.java
janhoy: Probably not, as it has never been advertised to end users?

solr/core/src/java/org/apache/solr/cli/StatusTool.java
janhoy: Sure, can do. What is the benefit of throwing an exception over doing an `exit(1)`?

solr/core/src/java/org/apache/solr/cli/SolrProcessMgr.java
janhoy: I think I'll go with `SolrProcessManager`...

janhoy:
I removed the RuntimePermission for `manageProcess` for runtime since I suspect it is only needed for tests, where we actually do spawn processes... Will revert that commit if the test fails...

janhoy:
There's a reproducible failure since `BasicAuthIntegrationTest` actually uses StatusTool and expects it to output pure JSON with the `--solr-url` option, and for it to also handle basic auth credentials correctly from `basicauth` sysProp. It does not..

janhoy:
So the reason is that in `BasicAuthIntegrationTest` it calls the tool directly with a url to query the status endpoint of the local Solr being tested. However, our test suite does not actually spawn processes, it just spins up a test jetty in a thread, so `SolrProcessManager` does not detect it. And we rely on info from running processes in that code path now. So either I'll change BasicAuth code to some other means of querying or will adapt the `--solr-url` code path of StatusTool to not rely on running process map...

epugh:
Okay, I see the challenge with StatusTool.   StatusTool *should* support `--credentials`.  Look at the tests, you can see we do:

```
    "--credentials",
      SecurityJson.USER_PASS,
```

I think this could be a perfectly good Solr 10 feature, and we should add the credentials support to StatusTool.   Do you want me to do that in this PR or in a seperate one?   All tools *should* support basic auth ;-).  and some day all the auths!

janhoy:
Feel free to jump in and commit to this PR if you have ideas. I'll eventually come around to it and can do a pull...

Willdotwhite:
Ahhh, this completely slipped my mind, apologies team - new baby brain and all that! I'll get on testing this after work today, are there any particular areas of the status checks you want me to focus on?

epugh:
@Willdotwhite I think we are looking for confidence that moving to the Java code for system information works well in both windows and linux.  I often look at the .bats test and repeat those steps in my manual testing...    I suppose also testing the pass in --solr-url would be interesting.   You could test against this demo Solr: 

http://quepid-solr.dev.o19s.com:8985/solr/media/select?q=Example is Solr 8.11 w/ images and Basic Auth setting solr:SolrRocks and http links

So...  `bin/solr.cmd status --solr-url http://quepid-solr.dev.o19s.com:8985 -u solr:SolrRocks' would be interesting, especially cause it's 8.11 not 10!

epugh:
@janhoy  would this Java code approach also be useful in the AuthTool?  It checks that you are running `bin/solr auth` on a server with a local Solr running in order to get the port number...

janhoy:
> @janhoy would this Java code approach also be useful in the AuthTool? It checks that you are running `bin/solr auth` on a server with a local Solr running in order to get the port number...

Absolutely, that's why I extracted `SolrProcessManager` as its own class to it can be used anywhere we want to list running solr processes, get a process by pid or port, check if solr is running on a certain port etc.

malliaridis:
I've tried to run this on Windows 11, but I don't get a correct output. This is what I did:

```console
[PATH_TO_PROJECT]\solr\packaging\build\dev>bin\solr.cmd start       
Java 17 detected. Enabled workaround for SOLR-16463
OpenJDK 64-Bit Server VM warning: JVM cannot use large page memory because it does not have enough privilege to lock pages in memory.
WARNING: A command line option has enabled the Security Manager
WARNING: The Security Manager is deprecated and will be removed in a future release

Found 1 Solr nodes: 

No Solr running on port 8983.

# After solr is running and accessible at http://localhost:8983
[PATH_TO_PROJECT]\solr\packaging\build\dev>bin\solr.cmd status
Looking for running processes...

Found 0 Solr nodes: 

No Solr nodes are running.
```

On MacOS `solr status` works correctly, current `main` as well. I will try to figure out the root cause of the output these days if noone else runs into the same issue and reports back before me.

rahulgoswami:
Can confirm @malliaridis's observation on Windows. solr status responds with a "No solr nodes are running".

Trying to debug this to see why processMgr.getAllRunning() in StatusTool doesn't return any values on Windows, but facing issues with trying to get the remote debugger to stick atm. Will give this another shot tomorrow.

janhoy:
Thanks for contributing Windows hardening. Could be the PID-file pattern at https://github.com/apache/solr/pull/2712/files#diff-1c8756285d3b596e65a2d20534ac75173349c4b117db59711e9121c05d0012ffR47 which *tries* to accomodate different PID-file name for win. 

Or perhaps the `java` command on windows is not pure lowercase? https://github.com/apache/solr/pull/2712/files#diff-1c8756285d3b596e65a2d20534ac75173349c4b117db59711e9121c05d0012ffR55

malliaridis:
It seems that `p.info().commandLine()` is always empty and falls back to the empty string, filtering out all the java processes. This means that `-Djetty.port=` is not maintained and only present on `bin/solr start` on first execution, resulting to this mixed output of found and not found nodes.

malliaridis:
After some lookup, I found this through [StackOverflow](https://stackoverflow.com/a/46768046/3329810) this:

> The command line arguments are not available via a non-privileged API for other processes
and so the Optional is always empty. The API is explicit that the values are OS specific.
If in the future, the arguments are available by a Window APIs, the implementation can be updated.
>
> BTW, the info structure is filled by native code; the assignments to the fields do not appear in the Java code.

Source: [JDK-8176725](https://bugs.openjdk.org/browse/JDK-8176725?focusedId=14061174&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14061174)

malliaridis:
If this is the only use-case I would avoid adding an entire library and prefer the workaround. I had to make three changes though to make it work:

- Define `isWindows` in `getCommandLine(ProcessHandle)`, for example as:
  ```java
  boolean isWindows = System.getProperty("os.name").startsWith("Windows");
  ```
- Replace any call to `ProcessHandle.Info#commandLine()` with `commandLine(ProcessHandle)` 
- Replace any call to `ProcessHandle.Info#arguments()` with `getCommandLine(ProcessHandle).orElse("").split(" ")`

Additional comments documenting the reason for the workaround should be added as well of course.

However, I am not sure if the process is the right place to lookup information like the port number. We may have changes in the future that eventually do not require / provide the port as an argument. At least that's what I think if we further simplify the CLI scripts.

janhoy:
Pushed some commits
* Fixes `BasicAuthIntegrationTest` by allowing status test to check status by URL even if Solr is not an OS process
* Workaround for Windows. Tested on Windows 11 in a VM

malliaridis (commented):
Tested on both Unix and Windows 11 and I face some errors with file permissions on Windows (see comments below).

Additionally, there are 2 use cases that we may consider changing how the tool behaves.

solr/core/src/java/org/apache/solr/cli/StatusTool.java
```diff
@@ -65,37 +75,132 @@ public String getName() {
...
+      CLIO.out(pidUrl);
+    } else {
+      statusFromRunningSolr(pidUrl, cli, maxWaitSecs);
+    }
+    CLIO.out("");
+  }
+
+  private Integer portFromUrl(String solrUrl) {
+    try {
+      return new URI(solrUrl).getPort();
```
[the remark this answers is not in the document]
janhoy: I adapted it to return 443 or 80.

gradle/testing/randomization/policies/solr-tests.policy
```diff
@@ -250,6 +252,8 @@ grant {
 
   // expanded to a wildcard if set, allows all networking everywhere
   permission java.net.SocketPermission "${solr.internal.network.permission}", "accept,listen,connect,resolve";
+
+  permission java.io.FilePermission "${java.home}${/}-", "execute";
```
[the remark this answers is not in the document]
janhoy: Added a few extra policy lines to allow execution of `wmic.exe`.

solr/server/etc/security.policy
```diff
@@ -228,4 +228,7 @@ grant {
 
   // expanded to a wildcard if set, allows all networking everywhere
   permission java.net.SocketPermission "${solr.internal.network.permission}", "accept,listen,connect,resolve";
+
+  // Used by SolrProcessManager to find commandLine of a running Windows process
+  permission java.io.FilePermission "C:\\Windows\\*\\wmic.exe", "execute";
```
janhoy, line 233: This is not yet tested..

gradle/testing/randomization/policies/solr-tests.policy
```diff
@@ -109,6 +109,8 @@ grant {
   permission java.lang.RuntimePermission "writeFileDescriptor";
   // needed by hadoop http
   permission java.lang.RuntimePermission "getProtectionDomain";
+  // SolrProcessMgr to list processes
+  permission java.lang.RuntimePermission "manageProcess";
```
janhoy, line 113: Note that we need to grant "manageProcess" permission only for tests, since tests run under the security manager and list and spawn processes. But the only current use of SolrProcessManager is from StatusTool which runs in SolrCLI which is not invoked with security manager.

If we want Solr's runtime to use SolrProcessManager in the future, we'd need to add this permission to `security.policy` as well. Not doing that since it would grant a potential attacker the right to not only list processes started by others users, but also terminate them.

epugh: Doesn't the security manager go away when we update to java 21 and lucene 10?

janhoy: You're right. But since this targets 9.x, we still need it. When someone rips out security manager related stuff in main before the 10.0 release, this will be gone of course. But not right now..

epugh: I would be fine with this as 10 if that makes your life easier..   kind of like the basic auth for CLI being 10x only...   I am assuming 10x is only months away!

solr/server/etc/security.policy
janhoy: Thinking about it, since status tool is only used from CLI and not from Solr server, we don't need this permission, since CLI runs without security mgr. I'll remove this.

gradle/testing/randomization/policies/solr-tests.policy
janhoy: No big deal, it will anyway land in main first. And if security manager is removed in main branch before this is merged, we'll of course have to take action..

janhoy:
Did more testing. The output when starting Solr on Win was broken and printed the json status text on Win, while on Linux it still printed `Started Solr server on port 3333 (pid=6838). Happy searching!`. So for `bin/solr` we still print that text in the shell script, but on Windows, we delegate it to StatusTool.

solr/core/src/java/org/apache/solr/cli/StatusTool.java
```diff
@@ -65,72 +73,224 @@ public String getName() {
...
-        echo("Started Solr server on port " + solrPort + ". Happy searching!");
-      } catch (TimeoutException timeout) {
-        throw new Exception(
-            "Solr at " + solrUrl + " did not come online within " + maxWaitSecs + " seconds!");
+    if (solrUrl != null) {
+      // URL provided, do not consult local processes, as the URL may be remote
+      if (maxWaitSecs > 0) {
+        // Used by Windows start script when starting Solr
+        try {
+          waitForSolrUpAndPrintStatus(solrUrl, cli, maxWaitSecs);
```
janhoy, line 115: I don't like the fact that the output text changes completely based on a flag like `--max-wait-secs`, but that's what I added now to get the windows print correct. @epugh I believe you added this at some point.

I'd suggest that `--max-wait-secs` only triggers the wait-and-retry logic and still prints the JSON output. Then we could use the exit code (0 or 1) from the statusTool to choose what to print in `solr.cmd`. Comments?

epugh: I think that makes sense...

janhoy: Perhaps best way is to keep the `—max-wait-secs` arg secret and undocumented. Then I have no issue with it having such behavior.

epugh: Okay, I think I need to re-read your notes.   Why isn't `max-wait-secs` a useful flag for both linux and windows?   Is it required on windows because of how we check the data requires some polling?

janhoy: `max-wait-secs` was an undocumented flag, only used in windows start script, and causes the tool to output same startup message as on linux.

It would have been useful in general for use of the tool, but I think I'll hide it for now and then we can do another round of dev to make it public later.

janhoy: Done. I invented a new "stealth" option mode: A deprecated option without a description :)

epugh:
I love how instead of making StatusTool help output more complex, you added the "stealth" idea that honestly we may use in other places!

janhoy:
Did another pass at cleaning up some convoluted code, and remove some duplicate printout. Here I have copied the command response from some local tests. Will probably commit soon if you think it looks good:

```
bin/solr status -h
usage: bin/solr status [-h] [-p <PORT>] [-s <HOST>] [--short] [--verbose]

List of options:
 -h,--help              Print this message.
 -p,--port <PORT>       Port on localhost to check status for
 -s,--solr-url <HOST>   Base Solr URL, which can be used to determine the zk-host if that's not known; defaults to:
                        http://localhost:8983.
    --short             Short format. Prints one URL per line for running instances
    --verbose           Enable verbose command output.

Please see the Reference Guide for more tools documentation:
https://solr.apache.org/guide/solr/latest/deployment-guide/solr-control-script-reference.html
```

```
bin/solr status --short
http://localhost:9999/solr
http://localhost:8888/solr
```

```
bin/solr status
Looking for running processes...

Found 2 Solr nodes:

Solr process 16318 running on port 9999
{
  "solr_home":"/Users/janhoy/git/solr/solr/packaging/build/dev/server/solr",
  "version":"10.0.0-SNAPSHOT 58f7d877726a0274665c13ea5c89bf57f14c0f2b [snapshot build, details omitted]",
  "startTime":"Fri Oct 18 12:38:37 CEST 2024",
  "uptime":"0 days, 0 hours, 20 minutes, 27 seconds",
  "memory":"209.9 MB (%41) of 512 MB",
  "cloud":{
    "ZooKeeper":"127.0.0.1:10999",
    "liveNodes":"1",
    "collections":"0"}}


Solr process 16472 running on port 8888
{
  "solr_home":"/Users/janhoy/git/solr/solr/packaging/build/dev/server/solr",
  "version":"10.0.0-SNAPSHOT 58f7d877726a0274665c13ea5c89bf57f14c0f2b [snapshot build, details omitted]",
  "startTime":"Fri Oct 18 12:38:45 CEST 2024",
  "uptime":"0 days, 0 hours, 20 minutes, 20 seconds",
  "memory":"82.7 MB (%16.1) of 512 MB",
  "cloud":{
    "ZooKeeper":"127.0.0.1:9888",
    "liveNodes":"1",
    "collections":"0"}}
```

```
bin/solr status -p 8888

Solr process 16472 running on port 8888
{
  "solr_home":"/Users/janhoy/git/solr/solr/packaging/build/dev/server/solr",
  "version":"10.0.0-SNAPSHOT 58f7d877726a0274665c13ea5c89bf57f14c0f2b [snapshot build, details omitted]",
  "startTime":"Fri Oct 18 12:38:45 CEST 2024",
  "uptime":"0 days, 0 hours, 20 minutes, 37 seconds",
  "memory":"83.2 MB (%16.2) of 512 MB",
  "cloud":{
    "ZooKeeper":"127.0.0.1:9888",
    "liveNodes":"1",
    "collections":"0"}}
```

```
bin/solr status -p 1111
Could not find a running Solr on port 1111
```

```
bin/solr status --solr-url http://localhost:8888/solr
{
  "solr_home":"/Users/janhoy/git/solr/solr/packaging/build/dev/server/solr",
  "version":"10.0.0-SNAPSHOT 58f7d877726a0274665c13ea5c89bf57f14c0f2b [snapshot build, details omitted]",
  "startTime":"Fri Oct 18 12:38:45 CEST 2024",
  "uptime":"0 days, 0 hours, 21 minutes, 6 seconds",
  "memory":"84.2 MB (%16.4) of 512 MB",
  "cloud":{
    "ZooKeeper":"127.0.0.1:9888",
    "liveNodes":"1",
    "collections":"0"}}
```

```
bin/solr status --solr-url http://localhost:1111/solr
Solr at http://localhost:1111/solr not online.
```

```
bin/solr status --solr-url http://localhost:8888/solr --max-wait-secs 1
Waiting up to 1 seconds to see Solr running on port 8888
Started Solr server on port 8888. Happy searching!
```

```
bin/solr status --solr-url http://localhost:1111/solr --max-wait-secs 1
Waiting up to 1 seconds to see Solr running on port 1111
Solr at http://localhost:1111/solr did not come online within 1 seconds!
```

janhoy:
Note that until now, the only supported syntax has been `bin/solr status` for user-facing, and `bin/solr status --solr-url <url>  --max-wait-secs <secs>` for scripts. So that's the only we need to worry about back compat in output.

janhoy:
I compared with outputs from v9.7:

## `solr status` with no running nodes:
```

+Looking for running processes...
+
+Found 0 Solr nodes:
+
No Solr nodes are running.
```

I.e. this PR adds two extra lines of output. Should I remove them?

## `solr status` with two nodes running:
Same two extra lines as above. In addition 9.7 printed this extraline:
```
WARNING: URLs provided to this tool needn't include Solr's context-root (e.g. "/solr"). Such URLs are deprecated and support for them will be removed in a future release. Correcting from [http://localhost:8888/solr/] to [http://localhost:8888/].
```

## `solr start`
Output almost identical between `bin/solr` script output and the output from statusTool, used on Windows.
```
Started Solr server on port 8888 (pid=1720). Happy searching!
Started Solr server on port 8888. Happy searching!
```

Difference is that StatusTool does not print the `(pid=xxxx)` information.

malliaridis:
@janhoy Since this PR got merged I have multiple tests that fail on Windows with `access denied ("java.io.FilePermission" "<<ALL FILES>>" "execute")`.

I'd like to see the security manager being removed in `main`, but the issues will still be present in `branch_9x`, so we should address them.

PR #2714, opened 2024-09-16
SOLR-17256: Introduce SolrRequest.setBasePath alternative

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-17256
# Description

Mutating SolrRequest objects as setBasePath does is a bad practice, as the modification isn't documented and many users expect to reuse these request objects across multiple requests.

# Solution

This PoC commit demoes one potential alternative: a `SolrClient.requestWithBaseUrl` method that temporarily overrides the base URL and then executes a caller-provided lambda.  e.g.

```
  var updateRsp = client.requestWithBaseUrl(url, (c) -> req.process(c))
```

This is a bit more involved than simply adding a new URL-consuming 'request' method, but helps us steer clear of further scope-creep in the SolrClient interface (as described in SOLR-12502).

# Tests

Additional unit tests in Http2SolrClientTest

# Checklist


solr/solrj/src/java/org/apache/solr/client/solrj/util/ClientUtils.java
```diff
@@ -102,6 +111,24 @@ public static String buildRequestUrl(
...
+    public BaseUrlOverride(String baseUrl) {
+      baseUrlOverride.set(baseUrl);
+    }
+
+    public void close() {
+      baseUrlOverride.remove();
+    }
+  }
+
+  public static final ThreadLocal<String> baseUrlOverride = new ThreadLocal<>();
```
dsmiley, line 130: Use of a ThreadLocal is pretty weird here IMO.  ThreadLocal's have their use to communicate something across API layers that can't or shouldn't include the extra information.  That doesn't apply here; a requestWithUri would work without such a hack.  I don't recall seeing an API embrace a ThreadLocal to accomplish something at it's front door / prominently.

gerlowskija: > ThreadLocal's have their use to communicate something across API layers that can't or shouldn't include the extra information.

Maybe I'm misreading your comment, but are you suggesting that this POC differs from the valid ThreadLocal use-case you described?

Here we have API layers that we don't want to be (explicitly) concerned with the baseURL - namely SolrRequest, and SolrClient's request-y methods.  So we use the lambda/thread-local combo to tunnel the information down to where the path building occurs.  Which feels to me very much like the valid use case of ThreadLocal's you described above?

Or maybe you're saying that this POC uses ThreadLocals in a valid/useful way, it's just not necessary because other options are out open?

> I don't recall seeing an API embrace a ThreadLocal to accomplish something at it's front door / prominently.

This variable can (and should) be private.  That's a PoC/code-cleanup mistake on my part that's easily remedied.

But beyond that, the thread-local **isn't** in the API anywhere - not in the nested `BaseUrlOverride` class, not in `SolrClient`, not even of any method in `ClientUtils`.  Can you elaborate a bit on how the API is "embracing" the thread-local prominently?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -533,6 +534,25 @@ public NamedList<Object> request(SolrRequest<?> solrRequest, String collection)
     }
   }
 
+  public <R> R requestWithBaseUrl(String baseUrl, SolrClientFunction<SolrClient, R> clientFunction)
```
gerlowskija, line 537: Conceptually, it probably makes sense for this method to be available on all of our "Http" clients.  But in practice, only Http2SolrClient makes it easy to quickly clone an existing instance, which this method implementation leans on pretty heavily.

Ambivalent on whether this is worth the effort to make available more widely.  Curious what others think...

gerlowskija:
Alright, I've updated the PR to get rid of the thread-local which folks didn't like.

For now the method is available on Http2SolrClient specifically - I'd love to see this spread to other "Http" clients eventually, but it's probably more work than it's worth currently.

Next steps here:

- some Javadoc wordsmithing for the new method(s)
- add a SolrRequest-consuming method to cover the simple case as David requested on the JIRA ticket
- add some tests, tidy, check.

If everyone's OK with the overall approach at this point, I'll aim to get this in sometime in the near future and then start working on PRs to update all the places we currently use `SolrRequest.setBasePath`.

iamsanjay:
IMO, We should stick to keep on using interface reference whereas this change force to use concrete classes. In #2689, Something similar we faced as we are still deciding on the type of SolrClient that `CoreContainer#getDefaultHttpClient` should return. Initially I proposed that it should return `SolrClient`, but because `requestAsyn` is not available in SolrClient. Therefore, it was recommended to change the type to Http2SolrClient.

In #2724, a new PR which would basically move `requestAsync` to SolrClient and by default throws UnsupportedOpeartionException. Same approach can be used here as well, add `requestBaseUrl` to SolrClient and default implementation should throw UnsupportedOperationException.

dsmiley:
I'm a bit uncomfortable on adding methods to SolrClient that *mostly* throw UnsupportedOperationException for most implementations.
I added the overloaded method I spoke of in the latest commit here.  No javadocs yet; just wanted to show it off.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -533,6 +535,47 @@ public NamedList<Object> request(SolrRequest<?> solrRequest, String collection)
...
+   * Temporarily modifies the client to use a different base URL and runs the provided lambda
+   *
+   * @param baseUrl the base URL to use on any requests made within the 'clientFunction' lambda
+   * @param clientFunction a Function that consumes a Http2SolrClient and returns an arbitrary value
+   * @return the value returned after invoking 'clientFunction'
+   * @param <R> the type returned by the provided function (and by this method)
+   */
+  public <R> R requestWithBaseUrl(String baseUrl, SolrClientFunction<SolrClient, R> clientFunction)
+      throws SolrServerException, IOException {
+    try (final var derivedClient = new NoCloseHttp2SolrClient(baseUrl, this)) {
```
dsmiley, line 553: why try-finally here?     ;-)

gerlowskija: Two reasons, both dumb:

1. To prevent IDE's or static analysis from complaining that we're not closing a Closeable here.
2. Because `NoCloseHttp2SolrClient.close()` is mostly a no-op, but it does have one tiny bit of logic: it calls `ObjectReleaseTracker.release(this)` so that it doesn't get (falsely) picked up by the resource-leak tracking our tests do.

dsmiley: fair enough; lets just add a one-liner excuse so this doesn't look so silly

gerlowskija:
> add requestBaseUrl to SolrClient and default implementation should throw UnsupportedOperationException.

I think I agree with David - it feels weird to add methods that would throw UOE in the majority of implementations.  It'd be right at home in other languages I think (Python, for instance), but cuts against expectation here IMO.

That said, I share @iamsanjay 's dislike for having the concrete type creeping into a lot of these usage locations.  Would a sub-class / sub-interface here be helpful?  e.g.

```
interface UrlOverrideableSolrClient {
  ...
  SolrResponse requestWithUrl(String, SolrRequest);
}
```

Something like the above would let users convey that they need "one of those clients that supports alternate URLs" without needing to specify the exact concrete type.  Something to consider for a followup JIRA/PR if you guys think it makes sense...

solr/solrj/src/java/org/apache/solr/client/solrj/SolrClient.java
```diff
@@ -1194,6 +1194,17 @@ public final NamedList<Object> request(final SolrRequest<?> request)
     return request(request, null);
   }
 
+  /**
```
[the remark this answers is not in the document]
gerlowskija: Gonna move this to its own file based on rationale above...

solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
```diff
@@ -376,10 +373,8 @@ public NamedList<Object> getLatestVersion() throws IOException {
     params.set(CommonParams.WT, JAVABIN);
     params.set(CommonParams.QT, ReplicationHandler.PATH);
     QueryRequest req = new QueryRequest(params);
-    req.setBasePath(leaderBaseUrl);
-    // TODO modify to use shardhandler
     try {
-      return solrClient.request(req, leaderCoreName);
+      return solrClient.requestWithBaseUrl(leaderCoreUrl, req).getResponse();
```
gerlowskija, line 377: [-1] Starting with [SOLR-17066](https://issues.apache.org/jira/browse/SOLR-17066) we're trying to move away from the use of core-URLs in SolrJ.  So we'll want to rework the usage that crept back in here.

We could address this by:

- adding a 'collection' arg to the lambda-less `requestWithBaseUrl`, or...
- creating a `requestWithBaseUrl` override with the 'collection' arg, or...
- using the existing lambda method:

    ```
        return solrClient.requestWithBaseUrl(leaderBaseUrl, (c) -> {
        return c.request(leaderCoreName, req);
        });
    ```

Any preference among those options @dsmiley, since you tweaked this bit recently?  If I don't hear from you in a few days I'll add a collection arg to the existing method, but I'm happy to go either route if you chime in.

dsmiley: The whole collection ambiguity aspect duplicating our SolrJ methods is really a shame, but it's a problem for another JIRA.  In the mean time, doubling up requestWithBaseUrl (lambda-less) is my preference.

Also note: `org.apache.solr.client.solrj.SolrRequest#getCollection` but only used for CloudSolrClient.  What a mess IMO.

dsmiley: or don't even double it; just have the collection variant and caller passes null if to node level.

gerlowskija: > The whole collection ambiguity aspect duplicating our SolrJ methods is really a shame

Agreed - I think we're in a better position to do something about that for 10.0 now that core-urls are dead (and `SolrRequest.setBasePath` soon will be as well).  But another JIRA for sure.

For now I'll go with the single method, `requestWithBaseUrl, baseUrl, collection, SolrRequest)`, and note in the Javadocs that collection may be null for "node" requests or "admin" requests or whatever we're calling them.

gerlowskija: (Done)

gerlowskija:
Alright - I think I've addressed the comments up to this point, and I've added the missing tests, etc, so I've pulled this out of "Draft" mode.

I'm going to start work on a separate PR that will remove a bunch of `setBasePath` usage, and make sure that doesn't flush out any bugs or improvements that would fit better into this PR.  But pending issues with that or any additional reviews here, I'll aim to merge at the end of this week-ish?

solr/solrj/src/java/org/apache/solr/client/solrj/util/ClientUtils.java
```diff
@@ -82,6 +82,8 @@ public static String buildRequestUrl(
       String serverRootUrl,
       String collection)
       throws MalformedURLException {
+
+    // TODO remove getBasePath support here prior to closing SOLR-17256
```
dsmiley, line 86: a nocommit; no? 
I've observed we devs have not used this practice much nowadays.  Maybe because we don't want to actually see "precommit" fail and similarly don't want the PR validations to fail for it either.  I suppose if there was a separate dedicated validation, it'd be okay.  Any way; just thinking out loud here.

gerlowskija: I don't want to remove this line until all of the setBasePath usages are removed (which will be in a separate PR).  That'll take at least one more PR, since I'm not bundling all of that into this PR, which is more about introducing the new APIs.

So not a "nocommit"

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -533,6 +536,40 @@ public NamedList<Object> request(SolrRequest<?> solrRequest, String collection)
...
+   * used by this client
+   *
+   * @param baseUrl a URL to a root Solr path (i.e. "/solr") that should be used for this request
+   * @param collection an optional collection or core name used to override the client's "default
+   *     collection". May be 'null' for any requests that don't require a collection or wish to rely
+   *     on the client's default
+   * @param req the SolrRequest to send
+   */
+  public final <T extends SolrResponse> T requestWithBaseUrl(
+      String baseUrl, String collection, SolrRequest<T> req)
```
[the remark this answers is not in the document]
gerlowskija: Gah, I always forget Github supports "Suggested Changes" now; so cool.

solr/solrj/src/java/org/apache/solr/client/solrj/SolrClientFunction.java
```diff
@@ -0,0 +1,30 @@
...
+import java.io.IOException;
+
+/**
+ * A lambda intended for invoking SolrClient operations
+ *
+ * <p>Exceptions commonly thrown by SolrClient operations, IOException and SolrServerException, are
+ * permitted to be thrown by lambda instances.
+ */
+@FunctionalInterface
+public interface SolrClientFunction<T, R> {
```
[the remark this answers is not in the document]
gerlowskija: I wasn't intentionally keeping it general; fixed.

gerlowskija:
> I'm going to [...] remove a bunch of `setBasePath` usage, and make sure that doesn't flush out any bugs or improvements

Still working through this, but it has surfaced at least one potential issue: this PR only provides a `setBasePath` replacement for our "Http2" client.  The large majority of `setBasePath` calls involve a Http2 client, but not all of them.  One or two places, most notably LBSolrClient, call `setBasePath` even when a HttpSolrClient is used.

There are a few workarounds available that are only slightly hacky, so it's not the end of the world.  And maybe the Apache clients are going away for 10.0 anyway and this is 100% moot - not sure where that effort stands?  But just wanted to mention it in case folks had opinions about LBSolrClient or knew what the plan around the Apache-based client was...

dsmiley:
We can ignore LBSolrClient; it's deprecated and I think a good chance of this going away in 10.

gerlowskija:
> We can ignore LBSolrClient; it's deprecated and I think a good chance of this going away in 10.

Hot dog; that'd be excellent.  I've worked around the problem in LBSolrClient with a minor hack (an `instanceof` check and cast), but I accompanied that with a TODO comment on a potential resolution, on the off-chance that LBSolrClient does survive longer-term.

gerlowskija:
I was a little surprised that we didn't add a 'deprecation' notice in CHANGES.txt when set/getBasePath were initially deprecated, but I've gone ahead and done that here:

> * SOLR-17256: Deprecate SolrRequest `setBasePath` and `getBasePath` methods.  SolrJ users wishing to temporarily
>   override an HTTP client's base URL may use `Http2SolrClient.requestWithBaseUrl` instead. (Jason Gerlowski,
>   Sanjay Dutt, David Smiley)

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -113,7 +115,7 @@ public class Http2SolrClient extends HttpSolrClientBase {
   private final HttpClient httpClient;
 
   private List<HttpListenerFactory> listenerFactory = new ArrayList<>();
-  private final AsyncTracker asyncTracker = new AsyncTracker();
+  protected AsyncTracker asyncTracker = new AsyncTracker();
```
dsmiley, line 118: I missed why this was related; what's going on here @gerlowskija ?

gerlowskija: It's a visibility change to allow the NoCloseHttp2SolrClient to access this instance var, so that the "NoClose" client can use the same instance as the "normal" Http2SolrClient it derives from.  See [here](https://github.com/apache/solr/pull/2714/files#diff-db379c74efc9ff28acfd7571b07e16dbcdaa510783f2bee835b4a288927b5b3dR841)

iamsanjay:
> Task :solr:solrj:compileTestJava FAILED
```
/home/runner/work/solr/solr/solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java:199: warning: [NonCanonicalType] The type `SolrClient.RemoteSolrException` was referred to by the non-canonical name `BaseHttpSolrClient.RemoteSolrException`. This may be misleading.
      } catch (BaseHttpSolrClient.RemoteSolrException rse) {
                                 ^
    (see https://errorprone.info/bugpattern/NonCanonicalType)
  Did you mean '} catch (SolrClient.RemoteSolrException rse) {'?
/home/runner/work/solr/solr/solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java:210: warning: [NonCanonicalType] The type `SolrClient.RemoteSolrException` was referred to by the non-canonical name `BaseHttpSolrClient.RemoteSolrException`. This may be misleading.
      } catch (BaseHttpSolrClient.RemoteSolrException rse) {
                                 ^
    (see https://errorprone.info/bugpattern/NonCanonicalType)
  Did you mean '} catch (SolrClient.RemoteSolrException rse) {'?
error: warnings found and -Werror specified
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
1 error
```

gerlowskija:
Huh, thanks for the heads up - fixing now.

PR #2721, opened 2024-09-23
SOLR-17442: Resolve -v flag conflict (version, value, verbose)

epugh:
https://issues.apache.org/jira/browse/SOLR-17442

# Description

Deprecating -v in favour of -d.

# Solution

# Tests

# Checklist


malliaridis (changes requested):
I forgot to mention in the ticket that introducing `-d` for `debug` may require further deprecations where `-d` is used for other options as well.

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
```diff
@@ -148,9 +154,21 @@ public class SolrCLI implements CLIO {
   public static final Option OPTION_VERBOSE =
```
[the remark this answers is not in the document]
epugh: hey, I hadn't thought about that use....    Good catch, and we should.

epugh:
> I forgot to mention in the ticket that introducing `-d` for `debug` may require further deprecations where `-d` is used for other options as well.

I am wondering if we should just stick to `--debug`, I saw a `-d` for directory...    it's not a common command to run, so maybe having a short cut isn't that valuable....     And dealing with the `-d`, well that seems really tough....

epugh:
> > I forgot to mention in the ticket that introducing `-d` for `debug` may require further deprecations where `-d` is used for other options as well.
> 
> I am wondering if we should just stick to `--debug`, I saw a `-d` for directory... it's not a common command to run, so maybe having a short cut isn't that valuable.... And dealing with the `-d`, well that seems really tough....

I checked what Docker does, just for run, and they use `-D`:

```
-D, --debug              Enable debug mode
```

I am kind of leaning towards avoiding the whole `-d` topic by removing it

malliaridis:
`-d` is one of the conflicting flags that we have to resolve anyway. I would prefer only the long variant in that case, which may be either `debug` or `verbose`, both are fine and when the user uses either of them he is already comitted to be more explicit with the CLI, if that makes sense.

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
```diff
@@ -137,6 +137,12 @@ public class SolrCLI implements CLIO {
   public static final Option OPTION_SOLRURL =
       Option.builder("url")
           .longOpt("solr-url")
+          .deprecated(
+              DeprecatedAttributes.builder()
+                  .setForRemoval(true)
+                  .setSince("9.7")
+                  .setDescription("Use --solr-url instead")
+                  .get())
```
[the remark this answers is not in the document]
epugh: interesting, I actually don't know why I added this...    It doesn't seem to fit the rest of this PR...

solr/bin/solr.cmd
```diff
@@ -362,23 +362,21 @@ goto done
...
+@echo Usage: solr stop [-k key] [-p port] [-d]
 @echo.
 @echo  -k key         Stop key; default is solrrocks
 @echo.
 @echo  -p port        Specify the port the Solr HTTP listener is bound to
 @echo.
 @echo  --all          Find and stop all running Solr servers on this host
 @echo.
-@echo  -V/--verbose   Verbose messages from this script
+@echo  -d/--debug     Debug messages from this script
```
malliaridis, line 379: If we go in the CLI classes with `--debug` without `-d`, then we should reflect the changes here as well to avoid any inconsistencies.

solr/bin/solr.cmd
```diff
@@ -398,10 +396,11 @@ IF "%1"=="-usage" goto usage
 IF "%1"=="/?" goto usage
 IF "%1"=="-f" goto set_foreground_mode
 IF "%1"=="--foreground" goto set_foreground_mode
-IF "%1"=="-V" goto set_verbose
-IF "%1"=="--verbose" goto set_verbose
+IF "%1"=="-V" goto set_debug
+IF "%1"=="--verbose" goto set_debug
```
malliaridis, line 400: `--debug` (and if `-d` is kept `-d` as well) should probably also added to the list.

solr/bin/solr
```diff
@@ -425,21 +425,19 @@ function print_usage() {
...
     echo ""
-    echo "  -v and -q             Verbose (-v) or quiet (-q) logging. Sets default log level of Solr to DEBUG or WARN instead of INFO"
-    echo ""
-    echo "  -V/--verbose           Verbose messages from this script"
+    echo "  -d/--debug or -q/--quiet Debug (-d) or quiet (-q) logging. Sets default log level of Solr to DEBUG or WARN instead of INFO"
     echo ""
   elif [ "$CMD" == "stop" ]; then
     echo ""
-    echo "Usage: solr stop [-k key] [-p port] [-V]"
+    echo "Usage: solr stop [-k key] [-p port] [-d]"
```
malliaridis, line 432: support for `-d` is not covered but mentioned in the output. Should probably be removed here as well.

solr/bin/solr
```diff
@@ -425,21 +425,19 @@ function print_usage() {
...
     echo ""
     echo "  -k <key>      Stop key; default is solrrocks"
     echo ""
     echo "  -p <port>     Specify the port the Solr HTTP listener is bound to"
     echo ""
     echo "  --all         Find and stop all running Solr servers on this host"
     echo ""
-    echo "  -V/--verbose   Verbose messages from this script"
+    echo "  -d/--debug    Debug messages from this script"
```
malliaridis, line 440: Same for removing `-d` applies here too.

solr/core/src/java/org/apache/solr/cli/AssertTool.java
```diff
@@ -131,7 +131,9 @@ public List<Option> getOptions() {
 
   @Override
   public int runTool(CommandLine cli) throws Exception {
-    verbose = cli.hasOption(SolrCLI.OPTION_VERBOSE.getOpt());
+    debug =
+        cli.hasOption((SolrCLI.OPTION_DEBUG.getOpt()))
+            || cli.hasOption(SolrCLI.OPTION_VERBOSE.getOpt());
```
malliaridis, line 136: With an `OptionGroup` this could be simplified to `debug = cli.hasOption(SolrCLI.OPTION_DEBUG)`, where `SolrCLI.OPTION_DEBUG` is the `OptionGroup` that includes verbose and debug options.

Note that the help outputs may be different if using option groups.

solr/core/src/java/org/apache/solr/cli/ConfigTool.java
```diff
@@ -92,12 +92,24 @@ public List<Option> getOptions() {
...
             .required(false)
             .desc("Set the property to this value; accepts JSON objects and strings.")
             .build(),
+        Option.builder()
+            .longOpt("value")
+            .argName("VALUE")
+            .hasArg()
+            .required(false) // should be true when -v is removed.
+            .desc("Set the property to this value; accepts JSON objects and strings.")
+            .build(),
```
malliaridis, line 112: `OptionGroup` could also be used here and the group may be marked as required, forcing the user to provide either option. This way, later checks (`if (value == null)`)are not necessary and `SolrCLI.getOptionWithDeprecatedAndDefault` may be replaced with `cli.getOption(optionGroup, null)`.

solr/core/src/java/org/apache/solr/cli/CreateTool.java
```diff
@@ -224,17 +224,18 @@ protected void createCore(CommandLine cli, SolrClient solrClient) throws Excepti
...
     }
 
-    echoIfVerbose("\nCreating new core '" + coreName + "' using CoreAdminRequest", cli);
+    echoIfDebug("\nCreating new core '" + coreName + "' using CoreAdminRequest", cli);
 
     try {
       CoreAdminResponse res = CoreAdminRequest.createCore(coreName, coreName, solrClient);
-      if (cli.hasOption(SolrCLI.OPTION_VERBOSE.getOpt())) {
+      if (cli.hasOption((SolrCLI.OPTION_DEBUG.getOpt()))
+          || cli.hasOption(SolrCLI.OPTION_VERBOSE.getOpt())) {
```
malliaridis, line 238: Same improvement with `OptionGroup` applies here as well.

solr/core/src/java/org/apache/solr/cli/CreateTool.java
```diff
@@ -362,7 +363,8 @@ protected void createCollection(CloudSolrClient cloudSolrClient, CommandLine cli
           "Failed to create collection '" + collectionName + "' due to: " + sse.getMessage());
     }
 
-    if (cli.hasOption(SolrCLI.OPTION_VERBOSE.getOpt())) {
+    if (cli.hasOption((SolrCLI.OPTION_DEBUG.getOpt()))
+        || cli.hasOption(SolrCLI.OPTION_VERBOSE.getOpt())) {
```
malliaridis, line 367: ... and here (`OptionGroup`).

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
```diff
@@ -642,8 +665,13 @@ private static void printHelp() {
     print(
         "  Omit '-z localhost:2181' from the above command if you have defined ZK_HOST in solr.in.sh.");
     print("");
-    print("Pass --help or -h after any COMMAND to see command-specific usage information,");
-    print("such as:    ./solr start --help or ./solr stop -h");
+    print("Global Options:");
+    print("  -v,  --version           Print version information and quit");
+    print("  -d,  --debug             Enable debug mode");
```
malliaridis, line 670: `-d` should probably not included here.

solr/core/src/java/org/apache/solr/cli/ToolBase.java
```diff
@@ -45,7 +46,9 @@ protected void echo(final String msg) {
 
   @Override
   public int runTool(CommandLine cli) throws Exception {
-    verbose = cli.hasOption(SolrCLI.OPTION_VERBOSE.getOpt());
+    debug =
+        cli.hasOption((SolrCLI.OPTION_DEBUG.getOpt()))
+            || cli.hasOption(SolrCLI.OPTION_VERBOSE.getOpt());
```
malliaridis, line 51: `OptionGroup` simplification.

solr/core/src/java/org/apache/solr/cli/ToolBase.java
```diff
@@ -33,8 +33,9 @@ protected ToolBase(PrintStream stdout) {
     this.stdout = stdout;
   }
 
-  protected void echoIfVerbose(final String msg, CommandLine cli) {
-    if (cli.hasOption(SolrCLI.OPTION_VERBOSE.getOpt())) {
+  protected void echoIfDebug(final String msg, CommandLine cli) {
+    if (cli.hasOption((SolrCLI.OPTION_DEBUG.getOpt()))
+        || cli.hasOption(SolrCLI.OPTION_VERBOSE.getOpt())) {
```
malliaridis, line 38: `OptionGroup` simplification.

solr/packaging/test/test_bats.bats
```diff
@@ -35,7 +35,7 @@ teardown_file() {
   # Conversely, on shutdown, we do need this to execute strictly
   # because using "run" will eat failing test exit codes
   solr stop --all
-  # DEBUG : (echo -n "# " ; solr stop -V --all) >&3
+  # DEBUG : (echo -n "# " ; solr stop -d --all) >&3
```
malliaridis, line 38: We should use the extended form (`--debug`) here instead

malliaridis (changes requested):
I think we should start sooner using `OptionGroup`s. I realized in this review and after testing it that we could simplify a lot of logic with them.

Many comments but mostly minor or repetitive things.

epugh:
@malliaridis I was able to integrate OptionGroup for the --debug and --verbose options in this PR and it works.   What I haven't been able to grok is how to simplify the checking for the flags.  You still have to check for each it apepars!!  If you find some code that avoids that, let me know.   

I am going to open a fresh PR that uses optiongroups more widely.

solr/core/src/java/org/apache/solr/cli/AssertTool.java
epugh: I think I need to dig more in and write a unit test to prove that works!

epugh:
I was at the point of merging and the --debug syntax just bugged me.   I am backing it out and just keep --verbose which rings better for me, and still accomplishs the goal of making -v no longer overlap.

malliaridis:
> You still have to check for each it apepars!! If you find some code that avoids that, let me know.

I see that you are using the `Option#getOpt()` and `Option#getLongOpt()` when you use `CommandLine#getOptionValue(String option)` or `CommandLine#hasOption(String option)`.

There is also a `CommandLine#getOptionValue(Option option)` and `CommandLine#getOptionValue(OptionGroup optionGroup)` that you could use instead. Same applies for `hasOption`. That way it is possible to use the `OptionGroup` and simplify logic as proposed before.

You can also set things like `required` (exactly one of the options in the group is present) at the `OptionGroup` via `OptionGroup#setRequired(boolean required)`. This is useful for cases where required options are split intot two due to deprecation, made optional and checked later for presence explicitly, like in [ConfigTool.java](https://github.com/apache/solr/pull/2721/files#diff-4dc9e0c436824dac317a2023038313dacbb6cd6d5aa592cc8c3e06d071335a87):

```java
Option.builder("v")
    .longOpt("value")
    .deprecated(...)
    .required(false)
    //...
Option.builder()
    .longOpt("value")
    .required(false) // should be true when -v is removed.
    //...

String value = SolrCLI.getOptionWithDeprecatedAndDefault(cli, "value", "v", null);
//...
if (value == null) {
    throw new MissingArgumentException("'value' is a required option.");
}
```

Using `OptionGroup`s will probably render `SolrCLI#getOptionWithDeprecatedAndDefault(...)` obsolete.

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
```diff
@@ -145,9 +145,16 @@ public class SolrCLI implements CLIO {
...
+
   public static final Option OPTION_VERBOSE =
       Option.builder("v")
           .longOpt("verbose")
+          .deprecated(
+              DeprecatedAttributes.builder()
+                  .setForRemoval(true)
+                  .setSince("9.8")
+                  .setDescription("Use --debug instead")
+                  .get())
```
malliaridis, line 157: @epugh could you review this file again? I think this should be undone if we won't introduce `--debug` and deprecate `-v` for version (below too).

PR #2722, opened 2024-09-24
SOLR-17459: Fix unrecognized option -scheme

malliaridis:
https://issues.apache.org/jira/browse/SOLR-17459

# Description

This PR is a fix for the "unrecognized option: -scheme" exception thrown when using the `bin/solr config` tool.

# Solution

The PR fully removes the support of `-s`, `-scheme` and `--scheme`, which was only used in `ConfigTool` before the migration (see #2540). Note that the argument could be set by the users before the migration. With the migration, the URL scheme can no longer be set via the scheme argument and is now set via the `--solr-url` argument or the environment variable `solr.url.scheme` (in case the fallback URL is used).

Note that this bug occurs only in `branch_9x` and `branch_9_7` and does not affect `main`.

# Tests

_No tests were modified or added.__

# Checklist

### DIFF
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -965,14 +965,6 @@ if [[ "$SCRIPT_CMD" == "config" ]]; then
               ZK_HOST="$2"
               shift 2
           ;;
-          -s|--scheme|-scheme)
-              if [[ -z "$2" || "${2:0:1}" == "-" ]]; then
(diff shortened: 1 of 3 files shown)


epugh (approved):
LGTM.   I went back and re-read the original posting to the mailing list, and this makes sense.

PR #2724, opened 2024-09-27
Introduce requestAsync Method in SolrClient for API Consistency and Extensibility

iamsanjay:
This PR introduces an abstract requestAsync method in the SolrClient class to provide a unified and symmetrical API for asynchronous requests across different implementations of SolrClient. The method will be available at the interface level, promoting consistency in API usage. For implementation classes that do not support asynchronous operations, invoking requestAsync will throw an `UnsupportedOperationException`.

# Checklist


solr/solrj/src/java/org/apache/solr/client/solrj/SolrClient.java
```diff
@@ -1225,4 +1228,30 @@ public SolrRequest.SolrClientContext getContext() {
...
+  public abstract CompletableFuture<NamedList<Object>> requestAsync(
+      final SolrRequest<?> request, String collection);
+
+  /**
+   * Execute an asynchronous request against a Solr server using the default collection.
+   *
+   * @param request the request to execute
+   * @return a {@link CompletableFuture} see {@link #requestAsync(SolrRequest, String)}.
+   */
+  public CompletableFuture<NamedList<Object>> requestAsync(final SolrRequest<?> request) {
```
dsmiley, line 1254: should be final

solr/solrj/src/java/org/apache/solr/client/solrj/SolrClient.java
```diff
@@ -1225,4 +1228,30 @@ public SolrRequest.SolrClientContext getContext() {
...
+   * @param collection the collection to execute the request against
+   * @return a {@link CompletableFuture} that tracks the progress of the async request. Supports
+   *     cancelling requests via {@link CompletableFuture#cancel(boolean)}, adding callbacks/error
+   *     handling using {@link CompletableFuture#whenComplete(BiConsumer)} and {@link
+   *     CompletableFuture#exceptionally(Function)} methods, and other CompletableFuture
+   *     functionality. Will complete exceptionally in case of either an {@link IOException} or
+   *     {@link SolrServerException} during the request. Once completed, the CompletableFuture will
+   *     contain a {@link NamedList} with the response from the server.
+   */
+  public abstract CompletableFuture<NamedList<Object>> requestAsync(
```
dsmiley, line 1245: why not add a default impl throwing UnsupportedOperationException here?

dsmiley:
IMO it's a questionable practice to add a method that will often throw UnsupportedOperationException.

iamsanjay:
As you must have seen now that we are doing the migration we are facing some challenges relating to HttpSolrClient, in some cases It also exposed HttpClient -- `SolrClientCloudManager`, `HttpSolrCall` are some of the examples. Of course we have not exposed HttpClient but once we start using concrete classes reference, It started to evolve on their own. For instance, adding requesAsync was the first example, and now in the new PR we are adding another method `requestWithUrl`. As our codebase evolve we will start seeing more and more Http2SolrClient reference, and during that we will keep on adding more and more to `Http2SolrClient` only. 

Yes; adding such method which throws `UnsupportedOpeartionException` is ugly. However, Lucene and our own codebase has good number of reference to UnsupportedOperationException. Even, Jdk uses it -- The collection framework, and lots of other places as well. 

But it is either choosing between this, or adding more methods to the concrete classes. If we decide to stick to `SolrClient`, we will definitely think of something.

jdyer1:
I think the right way to do this is to add a new interface solely for "requestAsync" and only those clients that support async would implement it.  If Java's collection framework were being designed today I doubt it would throw UOE.

iamsanjay:
Yeah, or we can also add another interface at the level above `HttpSolrClientBase`, and add all these new methods to it. As for name, `HttpSolrClient` would be good, however that is already taken. But we can think of some other name.

PR #2725, opened 2024-09-28
SOLR-17544: Prevent conflicting connection flags from being used via OptionGroup

epugh:
https://issues.apache.org/jira/browse/SOLR-17544

# Description

Change Options to being actual objects instanciatied, and use the Java object instead of the string representing the option in all the logic.

# Solution

Introduce OptionGroup for -s and -z.

# Tests


solr/core/src/java/org/apache/solr/cli/Tool.java
```diff
@@ -52,5 +53,7 @@ default String getFooter() {
 
   List<Option> getOptions();
 
+  List<OptionGroup> getOptionGroups();
```
malliaridis, line 56: I know this PR is a draft, but is it eventually better to go for `Options getOptions()` instead of `getOptions()` and `getOptionGroups()`? If a list is needed `Options#getOptions()` may be used instead.

`Options#addOptions(Options options)` does also exist if necessary.

epugh: definitly!   I didn't realize....   Also, if you want to push up to it, please do!

malliaridis:
@epugh I think I don't have the permissions to directly push to it. But I have done some untested work how we could further improve the implementation with `OptionGroup`s (and there may be more optimizations now possible).

Have a look at https://github.com/epugh/solr/compare/experiment_option_group...malliaridis:solr:experiment_option_group

I also found a few cases where deprecated options are no longer supported in this branch, so I might file a ticket for bringing back support in 9.8 (just in case) if this is an issue there.

epugh:
> @epugh I think I don't have the permissions to directly push to it. But I have done some untested work how we could further improve the implementation with `OptionGroup`s (and there may be more optimizations now possible).
> 
> Have a look at [epugh/solr@experiment_option_group...malliaridis:solr:experiment_option_group](https://github.com/epugh/solr/compare/experiment_option_group...malliaridis:solr:experiment_option_group)
> 
> If you want to add the changes to your branch I can create a PR, just let me know. :)
> 
> I also found a few cases where deprecated options are no longer supported in this branch, so I might file a ticket for bringing back support in 9.8 (just in case) if this is an issue there.

This is really cool...    I just added you to my epugh:solr repo, hopefully I did it right!

My head is thinking that OptionGroups, and using the explicit objects instead of strings is the way to go.  In terms of timing, I'm sort of thinking that maybe this all goes in 10 (`main`) only.   I'm really looking forward to getting rid of all the deprecated code, which will simplify a lot of the complex logic on params, simplify our testing, and open the door to more refactorings like embracing OptionGroups.    WDYT?   I don't believe 10 will be that far off anyway, so if `main` and `branch_9x` diverage a bit, that is okay.  Maybe we open a draft PR for OptionGroup in 10, and push hard to finish the conflicting params work that is in 10 and 9, and then finish off the OptionGroup work?

malliaridis:
@epugh this one is ready I believe. The merge with main was ugly. >.<

I ended up with more lines added than removed, even though I have simplified and removed redundant elements. This is probably only because of the extraction of options to a separate file or variables.

I made a few cleanups and migrations for consistency and also fixed a bug in StatusTool (see 31b1becd357e3f6d22950e9868cd1ff07686b129). Now we are using `getOptionValue` only with `Option` as parameter, not strings. And I also migrated to `getParsedOptionValue` wherever possible (booleans are not supported and the File parsing seems buggy).

malliaridis:
I didn't find any options that are mutually exclusive to use groups. This was relevant for the deprecated options to simplify the `getOptionValue` with a single option group. But since we have removed the deprecated options, we are back again using only `Option`s.

solr/core/src/java/org/apache/solr/cli/ApiTool.java
```diff
@@ -36,6 +36,16 @@
  * <p>Used to send an arbitrary HTTP request to a Solr API endpoint.
  */
 public class ApiTool extends ToolBase {
+
+  private static final Option SOLR_URL_OPTION =
+      Option.builder()
```
malliaridis, line 41: We should probably add the `-s` option here.

epugh:
I go back and forth on making everything a Java object.  `cli.getOption("my-option")` to me reads better than `cli.getOption(MY_OPTION)`, having said that, I think the enhanced IDE integration is the way to go...!    Glad to see the tests get fixed!   I will review in the AM and merge.   

Is this for branch_9x as well???

malliaridis (commented):
I am also not sure if we should keep the private options static or if it is more efficient if not static. They are only accessed by the same class instance and the static came from the options that live outside the class.

solr/core/src/java/org/apache/solr/cli/Tool.java
```diff
@@ -50,7 +50,19 @@ default String getFooter() {
     return "\nPlease see the Reference Guide for more tools documentation: https://solr.apache.org/guide/solr/latest/deployment-guide/solr-control-script-reference.html";
   }
 
-  List<Option> getOptions();
+  /**
```
epugh, line 53: I find the getAllOptions and getOPtions very confusing, especially because we have logic in ToolBase that even goes through and uses getAllOptions and then iterates and checks that if they are deprecated and removes them.  I also think that the only use of the "getOptions" is in the help..   I'll try to refactor this and we can see if we like the commit or not!

solr/core/src/java/org/apache/solr/cli/ToolBase.java
```diff
@@ -48,9 +52,30 @@ protected void echo(final String msg) {
...
+        options.addOption(toolOpt);
+      }
+    }
+    return options;
+  }
+
   @Override
   public int runTool(CommandLine cli) throws Exception {
-    verbose = cli.hasOption(SolrCLI.OPTION_VERBOSE.getLongOpt());
+    verbose = cli.hasOption(CommonCLIOptions.VERBOSE_OPTION);
```
epugh, line 77: This does read nicer!

solr/core/src/java/org/apache/solr/cli/Tool.java
malliaridis: I was considering merging them too. You mentioning it now seems like the way to go. :)

epugh:
Okay, I think this may be finally ready to merge to `main` (only).

solr/core/src/java/org/apache/solr/cli/AuthTool.java
```diff
@@ -569,7 +564,7 @@ private void updateIncludeFileEnableAuth(
         includeFileLines.add("REM The following lines added by solr.cmd for enabling BasicAuth");
         includeFileLines.add("set SOLR_AUTH_TYPE=kerberos");
         includeFileLines.add(
-            "set SOLR_AUTHENTICATION_OPTS=\"-Dsolr.httpclient.config=" + basicAuthConfFile + "\"");
+            "set SOLR_AUTHENTICATION_OPTS=\"-Dsolr.httpclient.config=basicAuthConfFile\"");
```
malliaridis, line 567: What exactly is the effect of this change? Is the string `basicAuthConfFile` here expected as value? We have two more references in AuthTool that mention `SOLR_AUTHENTICATION_OPTS`. Do they maybe also need ot be updated?

malliaridis (commented):
Looks good to me, with only a question about the last change that I do not understand.

solr/core/src/java/org/apache/solr/cli/AuthTool.java
epugh: so, if you have a null basicAuthConfFile variable, you go this route, and it's spitting out help info about lines in your file you need to comment out...  instead of inlinling a null in the messaage, it will at lease say "basicAuthConfFile"

malliaridis: Ah, I see. I paid little attention to the surrounding code. Maybe it is better then to add brackets or leave it blank to make it clear that this has to be filled out? Because a value like `basicAuthConfFile` looks like an enum value.

malliaridis (approved):

solr/core/src/java/org/apache/solr/cli/ToolBase.java
malliaridis: I would probably prefer the option group to be placed in `CommonCLIOptions` as another "option", mainly because of context, class responsibilities and for later referencing in other classes. I believe that there won't be any overrides either in implementations of `ToolBase`.

```java
// If CommonCLIOptions only imported
super.getOptions()
    .addOptionGroup(CommonCLIOptions.CONNECTION_OPTIONS)

...

if (cli.hasOption(CommonCLIOptions.CONNECTION_OPTIONS) { ... }
```

`getConnectionOptions` being placed in `ToolBase` reminds me on `getOptions()` and I would expect `Options` to be returned, rather than an `OptionGroup`.

solr/CHANGES.txt
```diff
@@ -26,6 +26,8 @@ Improvements
 * SOLR-16116: Apache Curator is now used to manage all Solr Zookeeper interactions. This should provide more stability in the Solr-Zookeeper interactions.
   The solrj-zookeeper module, now has a dependency on curator. (Houston Putman, Kevin Risden, Mike Drob, David Smiley)
 
+* SOLR-17544: Solr CLI will now warn you when you use multiple options and only one is a valid choice.  Combining -s and -z is most common example.  (Eric Pugh, Christos Malliaridis)
```
malliaridis, line 29: The OptionGroup interrupts as far as I know, "warn" is mildly expressed.

```suggestion
* SOLR-17544: Solr CLI will stop when you combine mutually exclusive options.  Combining -s and -z is most common example.  (Eric Pugh, Christos Malliaridis)
```

malliaridis (commented):
A few thoughts / suggestions, nothing of importance. :)

PR #2733, opened 2024-09-30
SOLR-17464

jdyer1:
Ths PR uses the passed-in client's executor if none already exists, no cast required.  This could still NPE in the case the neither the new client nor the passed-in client have an executor, but this seems an unlikely scenario.


solr/solrj/src/test/org/apache/solr/client/solrj/impl/HttpJdkSolrClientTest.java
```diff
@@ -194,7 +194,11 @@ public void testGetById() throws Exception {
 
   @Test
   public void testAsyncGet() throws Exception {
-    super.testQueryAsync();
+    String url = getBaseUrl() + DEBUG_SERVLET_PATH;
+    ResponseParser rp = new XMLResponseParser();
+    HttpSolrClientBuilderBase<?, ?> b =
```
dsmiley, line 199: "var" would be nice here

dsmiley (approved):

PR #2734, opened 2024-09-30
SOLR-16470: Create V2 equivalent of V1 Replication: Get files/{filePath}

mlbiscoc:

# Description

Create a V2 equivalent API for the replication "GET indexversion"

V1
`/solr/{coreName}/replication?command=filecontent&file=_0_Lucene99_0.tmd&wt=filestream`

V2
`/api/cores/{coreName}/replication/files/{filePath}?dirType=file`

Other sample requests

`/api/cores/{coreName}/replication/files/./tlog.0000000000000000000?dirType=tlogFile`
`/api/cores/{coreName}/replication/files/./solrconfig.xml?dirType=cf`

# Solution

Moved getFileStream() logic into ReplicationAPIBase class and updated the V1 endpoint to use the CoreReplicationAPI class so that the ReplicationHandler can be deprecated in the future.

For the V2 api, made `dirType` a required parameters specifying what file directory (cf/tlogFile/file) to fetch a file from and better expressed the optional parameters that exist for the api
* `offset`
* `compression`
* `checksum`
* `maxWriteMBPerSec`
* `generation`

# Tests

Wrote new unit test in CoreReplicationAPITest class testFetchFile()

# Checklist


mlbiscoc:
Hey @gerlowskija, should be the last API for SOLR-16470 whenever you get time to review. Thanks!

gerlowskija:
Awesome - thanks @mlbiscoc .  Sorry for the delay getting to this - took me awhile to catch back up after the travel last week.  Reviewing now...

solr/core/src/java/org/apache/solr/handler/admin/api/CoreReplicationAPI.java
```diff
@@ -68,6 +71,45 @@ public FileListResponse fetchFileList(
     return doFetchFileList(gen);
   }
 
+  @GET
```
gerlowskija, line 74: This is all correct and in line with the v2 APIs you did back in 2023 (PR #1620 PR #1704).

But we've made a slight tweak since then to support the [OpenAPI code generation stuff](https://issues.apache.org/jira/browse/SOLR-16825).  In short - we needed the JAX-RS annotations to live somewhere that gets compiled *before* "core", so that "core" can use all the nice generated code.  We managed this by creating a new Gradle module - "api" - where the JAX-RS annotations live on interfaces.  (Code in "core" can then implement those interfaces with methods containing the API logic).

Long story short - we'll want to do the same here.  There's some instructions [here](https://github.com/apache/solr/blob/main/dev-docs/apis.adoc#writing-jax-rs-apis) that might be helpful.  Though in your case most of the steps are already done - the pieces just need moved around a bit.

(Slightly my fault for not calling out more explicitly on SOLR-16470 that the setup had changed since your earlier commits, so I'm willing to make the necessary changes here.  You may beat me to it, but if not I'll try and update this eventually.)

solr/core/src/java/org/apache/solr/handler/admin/api/CoreReplicationAPI.java
```diff
@@ -68,6 +71,45 @@ public FileListResponse fetchFileList(
...
+  @Path("/files/{filePath}")
+  @Produces({MediaType.TEXT_PLAIN, BINARY_CONTENT_TYPE_V2})
+  public String fetchFile(
```
gerlowskija, line 77: [?] Can this really return a `String`?  Since it can return raw files I would've expected this to have to return `InputStream` or `Object` or something more generic in order to handle the potentially-binary content coming back?

gerlowskija (changes requested):

solr/core/src/java/org/apache/solr/handler/admin/api/CoreReplicationAPI.java
mlbiscoc: Ah good to know. No worries though, I just got back from vacation so will find some bandwidth soon to make the necessary adjustments.

gerlowskija: OK, I'll leave you to it, but if you have any questions or get stuck, lmk and I'll try to help out!

(Hope you had a great vacation!)

mlbiscoc:
Hey @gerlowskija made the changes but ended up kind of inflating this PR to support the OpenAPI code generation and moving files. Lmk if this went the right direction

gerlowskija:
Alright, merged in the latest 'main' and synced it with the changes here.  The only real change of note is that the interface file is now named 'ReplicationApis.java' (I went with the name on 'main' since it was already there.  The merge conflicts were a bit hairy, so if anything other than that seems "off", it's likely a mistake so please point it out 😆 

Hoping to dig into the StreamingOutput thing shortly - thanks @mlbiscoc !

mlbiscoc:
> Alright, merged in the latest 'main' and synced it with the changes here. The only real change of note is that the interface file is now named 'ReplicationApis.java' (I went with the name on 'main' since it was already there. The merge conflicts were a bit hairy, so if anything other than that seems "off", it's likely a mistake so please point it out 😆
> 
> Hoping to dig into the StreamingOutput thing shortly - thanks @mlbiscoc !

Great, thanks Jason. Will go back through it and take a look at the merge soon. Also need to look that precommit failure…

gerlowskija:
After a bit of reading, I agree with Matt that StreamingOutput seems to be the "preferred" approach.

The best explanation I could find on "why" comes from [this SO post](https://stackoverflow.com/questions/38632895/streamingoutput-in-jax-rs-usecases).  To summarize: StreamingOutput is easier for various JAX-RS supported interceptor hooks (MessageBodyWriter, ResponseFilter, etc.).  The example given is an interceptor that implements gzip-encoding - apparently that's tough to do on an OutputStream, but easier in some way for JAX-RS to handle when wrapped as a 'StreamingOutput'.

That said - supporting this type in our Java code-generation might take a bit more effort than we want to bite off here and now.  The API should be fine as far as Javascript and Python clients are concerned, but the custom template we're using for Java makes that one a bit trickier.

I'll exempt `fetchFile` from Java code generation (we have an annotation for that purpose), and create a followup JIRA ticket around standardizing on and supporting StreamingOutput.

gerlowskija:
(To close the loop, I created SOLR-17562 to cover the codegen StreamingOutput support)

Not sure what's up with the precommit and tests GHA's, but they both pass locally for me.  Will take a look and fix if I can.

Other than that (and CHANGES.txt, which I'll do right before merge) - is there anything else that needs done here @mlbiscoc ?

fsparv:
Not seeing documentation updates?

solr/core/src/java/org/apache/solr/handler/ReplicationHandler.java
```diff
@@ -1513,257 +1544,11 @@ public void postSoftCommit() {}
     };
   }
 
-  /** This class is used to read and send files in the lucene index */
-  private class DirectoryFileStream implements SolrCore.RawWriter {
```
dsmiley, line 1517: minor: as you are moving things and implementing a new interface, it's an opportunity to consider the ordering of the methods follows a logical flow.  Disclaimer:  I don't know this code enough to know what that is; it may be ideal already.  We like to read top-down generally so if it flows that way then great.

gerlowskija:
> Not seeing documentation updates?

Ah, good thought.  Yeah, `modules/deployment-guide/pages/user-managed-index-replication.adoc` should probably be updated to mention the v2 version of the API.

(If you haven't done a ref-guide update yet - we tend to use a tabbed display to group v1 and v2 examples of a particular API together. There are lots of sections in the ref-guide that could serve as an example here, but I'll pick "Create Collection" as an example.  For create-collection [the asciidoc snippet here](https://github.com/apache/solr/blob/main/solr/solr-ref-guide/modules/deployment-guide/pages/collection-management.adoc?plain=1#L35-L63) gets parsed and displayed into the nice looking "tabbed" display [here](https://solr.apache.org/guide/solr/latest/deployment-guide/collection-management.html#create).)

Other than the docs though, this LGTM and I'll aim to merge when they get added.

mlbiscoc:
Added in some documentation to the ref-guide for the filecontent api and it's parameters. Also made a small change to fix precommit. Went back through the merge and rebase and nothing sticks out to me that looks out of the ordinary.

solr/solr-ref-guide/modules/deployment-guide/pages/user-managed-index-replication.adoc
```diff
@@ -457,6 +457,51 @@ http://_host:port_/api/cores/_core_name_/replication/files?generation=<_generati
...
+Directory type is required for both V1 and V2 with 3 parameter supported types:
+
+* `file` Read from Lucene index file directory
+
+* `cf` Read from Configuration file directory
+
+* `tlogFile` Read from Tlog file directory
+
++
+There are several supported optional request parameters:
```
epugh, line 493: maybe "server optional"

PR #2737, opened 2024-10-02
SOLR-17453: Leverage waitForState() instead of busy waiting

psalagnac:
https://issues.apache.org/jira/browse/SOLR-17453

# Description

For some changes done to the cluster state via overseer, the code actively refreshes the seen cluster state until the change is fully done. This is a waste of resources, and this can be replaced by a ZK watch.

# Solution

Replace busy waiting (mostly done with `TimeOut` class) by Zookeeper watches and `ZkStateReader.waitForState()`.

# Tests

No new tests added.
Some tests updated to also remove busy waiting.

# Checklist


solr/core/src/test/org/apache/solr/cloud/TestPullReplicaErrorHandling.java
```diff
@@ -118,7 +118,7 @@ public void tearDown() throws Exception {
       log.info("tearDown deleting collection");
       CollectionAdminRequest.deleteCollection(collectionName).process(cluster.getSolrClient());
       log.info("Collection deleted");
-      waitForDeletion(collectionName);
+      TestPullReplica.waitForDeletion(collectionName);
```
psalagnac, line 121: This method is already invoked in `TestPullReplicaWithAuth` (no change in this PR). Since they are related tests, it makes sense to shared it instead of duplicating the code.

solr/core/src/test/org/apache/solr/cloud/TestRebalanceLeaders.java
```diff
@@ -558,14 +534,6 @@ private void concentratePropByRestartingJettys() throws Exception {
     // then start them again to concentrate the leaders. It's not necessary that all shards have a
     // leader.
 
-    ExecutorService executorService = ExecutorUtil.newMDCAwareCachedThreadPool("Start Jetty");
-
-    for (JettySolrRunner jetty : jettys) {
-      cluster.stopJettySolrRunner(jetty);
-    }
-
-    ExecutorUtil.shutdownAndAwaitTermination(executorService);
```
psalagnac, line 567: Side change: the executor was started and stopped for no reasons. It is actually used right after.

solr/test-framework/src/java/org/apache/solr/cloud/AbstractDistribZkTestBase.java
```diff
@@ -242,45 +240,15 @@ public static void waitForCollectionToDisappear(
...
-        fail(
-            "Could not find new leader even after waiting for "
-                + timeOut.timeElapsed(MILLISECONDS)
-                + "ms");
-      }
-
-      Thread.sleep(100);
-    }
 
+    long start = System.nanoTime();
```
psalagnac, line 248: The code with `waitForState()` was existed already. We were checking the condition twice.

solr/modules/hdfs/src/test/org/apache/solr/hdfs/cloud/SharedFileSystemAutoReplicaFailoverTest.java
```diff
@@ -31,6 +31,7 @@
 import java.util.concurrent.Future;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
```
psalagnac, line 34: I haven't been able to make this test pass, same with and without the fix. It is flagged `@BadApple` for some time.
My understanding is it was testing the "autoAddReplica" property, which is not supported anymore since removal of autoscaling framework.
See https://github.com/apache/solr/commit/cc0c111949d5039a0c7cb67cad55c63e2f761298#diff-ec4c71bca5a91aa5a49e509688891dc2d21fdda62981add8fb240b256e8d1a80L165
Should this test class just be deleted?

solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
```diff
@@ -221,24 +223,19 @@ public void call(ClusterState clusterState, ZkNodeProps message, NamedList<Objec
...
               SolrException.ErrorCode.SERVER_ERROR,
-              "Could not fully create collection: " + collectionName);
+              "Could not fully create collection: " + collectionName,
+              e);
         }
 
         // refresh cluster state (value read below comes from Zookeeper watch firing following the
         // update done previously, be it by Overseer or by this thread when updates are distributed)
         clusterState = ccc.getSolrCloudManager().getClusterState();
-        newColl = clusterState.getCollection(collectionName);
```
dsmiley, line 241: Why remove this line?  I think it's supposed to be here.

solr/core/src/java/org/apache/solr/cloud/api/collections/ReindexCollectionCmd.java
```diff
@@ -360,22 +362,17 @@ public void call(ClusterState clusterState, ZkNodeProps message, NamedList<Objec
...
-        clusterState = ccc.getSolrCloudManager().getClusterState();
-        created =
-            clusterState.hasCollection(targetCollection)
-                && clusterState.hasCollection(chkCollection);
-        if (created) break;
-      }
-      if (!created) {
+      try {
+        for (String col : List.of(targetCollection, chkCollection)) {
+          ccc.getZkStateReader().waitForState(col, 30, TimeUnit.SECONDS, Objects::nonNull);
```
dsmiley, line 367: so much nicer :-)

solr/core/src/test/org/apache/solr/cloud/TestRebalanceLeaders.java
```diff
@@ -604,74 +572,61 @@ private void forceUpdateCollectionStatus() {
...
+
+    waitForState(
+        "Waiting for all replicas to become inactive",
+        COLLECTION_NAME,
+        (liveNodes, docCollection) -> {
+          boolean expectedInactive = true;
+
+          for (Slice slice : docCollection.getSlices()) {
+            for (Replica rep : slice.getReplicas()) {
+              if (!downJettyNodes.contains(rep.getNodeName())) {
```
dsmiley, line 591: I don't recommend flipping this wherever you see it, but you are quite welcome to use either style.  Over 10 years ago, Lucene (which included Solr) developers started preferring the higher clarity of == false over an exclamation mark, I recall related to the visual subtleness of a single char.  Nevertheless there's no code standard on this.

solr/modules/hdfs/src/test/org/apache/solr/hdfs/cloud/SharedFileSystemAutoReplicaFailoverTest.java
dsmiley: autoAddReplica was ported to work without the autoscaling framework; it survived the deletion.  I recommend not removing the test right now; deserves a separate discussion.

solr/test-framework/src/java/org/apache/solr/cloud/AbstractDistribZkTestBase.java
```diff
@@ -242,45 +240,15 @@ public static void waitForCollectionToDisappear(
...
-          log.info(
-              "Old leader {}, new leader {}. New leader got elected in {} ms",
-              oldLeader,
-              slice.getLeader(),
-              timeOut.timeElapsed(MILLISECONDS));
-        }
-        break;
-      }
-
-      if (timeOut.hasTimedOut()) {
```
dsmiley, line 269: the loss of the diagnostics seems unfortunate; we could retain that if we time out

solr/test-framework/src/java/org/apache/solr/cloud/AbstractDistribZkTestBase.java
dsmiley: I suggest naming this var "startNs" so it has units

solr/test-framework/src/java/org/apache/solr/cloud/AbstractFullDistribZkTestBase.java
```diff
@@ -2217,83 +2218,60 @@ public static SolrInputDocument getDoc(Object... fields) throws Exception {
...
+    AtomicReference<String> message = new AtomicReference<>();
+    try {
+      reader.waitForState(
+          collectionName,
+          120,
+          TimeUnit.SECONDS,
+          c -> {
+            int expectedSlices = numShardsNumReplicaList.get(0);
+            // The Math.min thing is here, because we expect replication-factor to be reduced to if
+            // there
```
dsmiley, line 2235: reflow this

dsmiley (commented):
Wow, there were so many places to use waitForState; more than I realized!  So much better now; thanks for this.

BTW sorry for my delayed review; I didn't notice it till now

solr/core/src/test/org/apache/solr/cloud/TestRebalanceLeaders.java
epugh: We really should establish a code standard and have it be enforced....

epugh (commented):
I really like this refactoring.  I don't have anything to add past what @dsmiley  has said.

One thought, is there a way to enforce the use of `waitForState()` pattern via any of our code quality tools?   How can we make sure we don't backslide...   Maybe just converting everything is enough, assuming folks pretty much cut'n'paste these patterns of code....

dsmiley:
A line of javadocs on TimeOut recommending waitForState would be good but perhaps arguably a distraction to a general utility -- :shrug: whatever.  I don't think we'll backslide much if all/most cluster state waiting is done using the correct API.

solr/core/src/test/org/apache/solr/cloud/TestRebalanceLeaders.java
dsmiley: There doesn't have to be a code standard for everything Eric :-)

epugh: ;-).  I think we will have to agree to disagree ;-).   I really rely on the tools to encourage me in the right way to do things, or we end up with multiple ways.  I also have the memory of a gold fish, so if it isn't enforced by a tool, I may say "Hey, it should be X" for a year or two, and then swap to "Hey, it should be Y" just because I can't remember what the agreement was.   I can imagine that too strict standards becomes a strait jacket too.

solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
psalagnac: `newColl` is assigned from result of `waitForState()`, line 227. We're sure we already have the latest available version.
It would be a dupe to get it again here.

solr/test-framework/src/java/org/apache/solr/cloud/AbstractDistribZkTestBase.java
psalagnac: Good point. But this is at the cost of potentially slower test execution since we don't unblock the test thread when we receive the ZK watch.

We would get the same if we call `logThreadDumps(...)` and `printLayoutToStream(...)` in case of error. I can do such a change.

solr/core/src/test/org/apache/solr/cloud/TestRebalanceLeaders.java
psalagnac: I flipped this because it is reported by my IDE. I didn't know such a convention was embraced at some point. Will change back to `== false` form.

psalagnac:
> One thought, is there a way to enforce the use of waitForState() pattern via any of our code quality tools?

Not sure how we can automate decision on whether usages of `Timeout` are legit or not. We should use `waitForState()` instead of busy waiting for changes in Zookeeper, so we leverage the registered watchers. There are other cases, mostly when doing Solr-to-Solr requests, where we should keep `Timeout`.

dsmiley:
Probably just a one-liner left and I'll merge away :-)

solr/test-framework/src/java/org/apache/solr/cloud/AbstractDistribZkTestBase.java
psalagnac: Good catch, thanks. I replaced the exception by a call to `fail()` with similar message.
I also removed the `InterruptedException`. This one is not supposed to happen, and that's better to not consider it in the test.

PR #2741, opened 2024-10-04
SOLR-16503: Replace default USH apache client with Http2SolrClient

iamsanjay:

Replaced the default apache Http client provided by `UpdateShardHandler` with the Http2SolrClient provided by CoreContainer#getDefaultHttpSolrClient which will be introduced in #2689. As we are still deciding on what is the best way moving forward to set the url. Here, for now, almost everywhere a new Http2SolrClient is being re-created.

```
try (var solrClient =
          new Http2SolrClient.Builder(baseUrl)
              .withHttpClient(coreContainer.getDefaultHttpSolrClient())
              .build()) {
    // code goes here
}
```

Of course, If we decided to go with #2714 then setting up url and closing the instance would be replaced with a new abstraction.
```
var updateRsp = client.requestWithBaseUrl(url, (c) -> req.process(c))
```

# Checklist


iamsanjay:
**SplitShardWithNodeRoleTest.testSolrClusterWithNodeRoleWithPull** failed! Even in the past this one failed. If I look at the test, It seem simple, However the complexity that split operation involves is rather too much. This is how it runs.


1. Create a collection with one shard containing one NRT and one PULL replica.
2. waitForState to achieve above configuration
3. Index 10 documents
4. Commit collection
5. Perform split operation
6. waitForState for 2 active sub-shards (about 45 seconds, and this is where it failed.)

I am not sure If increasing Timeout will help us. But we can try it! In the logs, one can see that just before the assertion failed, the IndexFetcher was running. May be the PULL replicas were downloading the index, and the time finished and the sub-shard never really recovered. 

Question: Do both NRT and PULL replicas need to be active for sub-shards to be active?

solr/core/src/java/org/apache/solr/util/stats/InstrumentedHttpListenerFactory.java
```diff
@@ -133,4 +134,18 @@ public void initializeMetrics(SolrMetricsContext parentContext, String scope) {
   public SolrMetricsContext getSolrMetricsContext() {
     return solrMetricsContext;
   }
+
+  public static NameStrategy getNameStrategy(String name) {
+    InstrumentedHttpListenerFactory.NameStrategy nameStrategy =
```
dsmiley, line 139: `var` would be way shorter and still clear

solr/core/src/test/org/apache/solr/core/TestHttpSolrClientProvider.java
```diff
@@ -0,0 +1,81 @@
...
+          HttpClientUtil.DEFAULT_SO_TIMEOUT);
+    } finally {
+      httpSolrClientProvider.close();
+    }
+  }
+
+  @Test
+  public void test_when_updateShardHandler_cfg_is_not_null() {
+    var idleTimeout = 10000;
+    UpdateShardHandlerConfig cfg = new UpdateShardHandlerConfig(-1, -1, idleTimeout, -1, null, -1);
```
dsmiley, line 60: assert != the default

solr/core/src/test/org/apache/solr/core/TestHttpSolrClientProvider.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.solr.client.solrj.impl.HttpClientUtil;
+import org.apache.solr.metrics.SolrMetricsContext;
+import org.apache.solr.update.UpdateShardHandlerConfig;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+public class TestHttpSolrClientProvider extends SolrTestCase {
+
+  HttpSolrClientProvider httpSolrClientProvider;
```
dsmiley, line 34: why is this a field?  It could be local vars and could then do try-with-resources normally

solr/core/src/java/org/apache/solr/security/PKIAuthenticationPlugin.java
```diff
@@ -346,23 +347,24 @@ PublicKey fetchPublicKeyFromRemote(String nodename) {
...
-      String key = (String) m.get("key");
+      final var solrParams = new ModifiableSolrParams();
+      solrParams.add("wt", "json");
+      solrParams.add("omitHeader", "true");
+
+      final var request = new GenericSolrRequest(GET, PublicKeyHandler.PATH, solrParams);
+      log.debug("Fetching fresh public key from: {}", url);
+      var solrClient = cores.getDefaultHttpSolrClient();
+      NamedList<Object> resp =
+          solrClient.requestWithBaseUrl(url, client -> client.request(request));
```
dsmiley, line 358: don't need the lambda form

solr/core/src/java/org/apache/solr/pkg/PackageAPI.java
```diff
@@ -418,13 +439,26 @@ private void syncToVersion(int expectedVersion) {
...
-      Utils.executeGET(
-          coreContainer.getUpdateShardHandler().getDefaultHttpClient(),
-          coreContainer.getZkController().zkStateReader.getBaseUrlV2ForNodeName(liveNode)
-              + "/cluster/package?wt=javabin&omitHeader=true&expectedVersion"
-              + expected,
-          Utils.JAVABINCONSUMER);
+      var baseUrl = coreContainer.getZkController().zkStateReader.getBaseUrlV2ForNodeName(liveNode);
+      try {
+        var solrClient = coreContainer.getDefaultHttpSolrClient();
+        solrClient.requestWithBaseUrl(baseUrl, client -> client.request(solrRequest));
```
dsmiley, line 455: don't need the lambda form

solr/core/src/java/org/apache/solr/pkg/PackageAPI.java
```diff
@@ -250,13 +255,29 @@ public void refresh(PayloadObj<String> payload) {
...
-        Utils.executeGET(
-            coreContainer.getUpdateShardHandler().getDefaultHttpClient(),
-            coreContainer.getZkController().zkStateReader.getBaseUrlV2ForNodeName(liveNode)
-                + "/cluster/package?wt=javabin&omitHeader=true&refreshPackage="
-                + p,
-            Utils.JAVABINCONSUMER);
+        final var baseUrl =
+            coreContainer.getZkController().zkStateReader.getBaseUrlV2ForNodeName(liveNode);
+        try (var solrClient =
+            new Http2SolrClient.Builder(baseUrl)
```
dsmiley, line 271: why create a new client here?

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -484,14 +520,21 @@ public List<FileDetails> list(String path, Predicate<String> predicate) {
...
       String baseUrl =
           coreContainer.getZkController().getZkStateReader().getBaseUrlV2ForNodeName(node);
-      String url = baseUrl + "/cluster/files" + path + "?localDelete=true";
-      HttpDelete del = new HttpDelete(url);
-      // invoke delete command on all nodes asynchronously
-      coreContainer.runAsync(() -> Utils.executeHttpMethod(client, url, null, del));
+      try (var solrClient =
+          new Http2SolrClient.Builder(baseUrl)
+              .withHttpClient(coreContainer.getDefaultHttpSolrClient())
+              .build()) {
```
dsmiley, line 534: why a new client?

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -484,14 +520,21 @@ public List<FileDetails> list(String path, Predicate<String> predicate) {
...
+        // invoke delete command on all nodes asynchronously
+        solrClient.requestAsync(solrRequest);
+      }
```
dsmiley, line 537: nice... although no attention is payed to the result.

solr/core/src/java/org/apache/solr/cloud/Overseer.java
```diff
@@ -856,12 +856,13 @@ private void doCompatCheck(BiConsumer<String, Object> consumer) {
...
     try (CloudSolrClient client =
-        new CloudLegacySolrClient.Builder(
+        new CloudHttp2SolrClient.Builder(
                 Collections.singletonList(getZkController().getZkServerAddress()), Optional.empty())
-            .withSocketTimeout(30000, TimeUnit.MILLISECONDS)
-            .withConnectionTimeout(15000, TimeUnit.MILLISECONDS)
-            .withHttpClient(updateShardHandler.getDefaultHttpClient())
+            .withZkClientTimeout(30000, TimeUnit.MILLISECONDS)
+            .withZkConnectTimeout(15000, TimeUnit.MILLISECONDS)
+            .withHttpClient(getCoreContainer().getDefaultHttpSolrClient())
```
dsmiley, line 865: withSocketTimeout was on the HTTP connection at the LBHttpSolrClient layer not ZK one.  The Http2/Jetty side doesn't quite work the same way.  It creates an LBHttp2SolrClient without such customizations.  Maybe they could bet set at the HttpClient (Jetty) layer, albeit this means you're not passing in   AFAICT, there's no direct substitute for our Http2 builder; instead you'd have to create the Jetty HttpClient with those settings and then pass it in.

Aaaaanyway.... I don't think it's worth retaining such particulars here.  I don't know why these specific timeouts were put here; they were added by @sigram in relation to ReindexCollection.  I recommend we simply use the new getCoreContainer().getDefaultHttpSolrClient().

solr/core/src/java/org/apache/solr/cloud/api/collections/ReindexCollectionCmd.java
```diff
@@ -729,170 +726,153 @@ private void waitForDaemon(
...
-      // now kill it - it's already stopped, this simply removes its status
-      q.set("action", "kill");
-      req = new QueryRequest(q);
-      solrClient.request(req);
     }
+    // now kill it - it's already stopped, this simply removes its status
+    executeDaemonAction("kill", daemonName, daemonReplica);
+  }
+
+  private NamedList<Object> executeDaemonAction(
```
dsmiley, line 862: love this refactoring, by the way

solr/core/src/test/org/apache/solr/core/TestHttpSolrClientProvider.java
dsmiley: I meant this up front to clarify that 10_000 isn't the default.  An _additional_ assertion.  Your existing one was good.

solr/core/src/java/org/apache/solr/cloud/Overseer.java
```diff
@@ -857,13 +857,18 @@ private void doCompatCheck(BiConsumer<String, Object> consumer) {
...
     }
 
-    try (CloudSolrClient client =
-        new CloudHttp2SolrClient.Builder(
-                Collections.singletonList(getZkController().getZkServerAddress()), Optional.empty())
-            .withZkClientTimeout(30000, TimeUnit.MILLISECONDS)
-            .withZkConnectTimeout(15000, TimeUnit.MILLISECONDS)
-            .withHttpClient(getCoreContainer().getDefaultHttpSolrClient())
-            .build()) {
+    try (var solrClient =
```
dsmiley, line 860: since this client is embedded into the Cloud one, I think you should name this var maybe internalClient or something and don't refer to it in the block of code for the try-finally except for passing into the CloudSolrClient.

BTW; it really is painful to customize these particular timeouts as we're forced to do it at an inner layer.  Ugh.  @epugh not sure if you worked on some of the SolrClient building methods and saw this issue before.  Maybe we should add the same methods to the CloudHttp2SolrClient.Builder

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -192,7 +192,7 @@ private boolean fetchFileFromNodeAndPersist(String fromNode) {
       try {
         GenericSolrRequest request = new GenericSolrRequest(GET, "/node/files" + getMetaPath());
         request.setResponseParser(new InputStreamResponseParser(null));
-        var response = solrClient.requestWithBaseUrl(baseUrl, client -> client.request(request));
+        var response = solrClient.requestWithBaseUrl(baseUrl, request::process).getResponse();
```
dsmiley, line 195: This is rather nice.  @gerlowskija, I think we can do away with the lambda-less requestWithBaseUrl I was suggesting.  WDYT?

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -484,14 +512,21 @@ public List<FileDetails> list(String path, Predicate<String> predicate) {
...
+    final var solrRequest = new GenericSolrRequest(DELETE, "/cluster/files" + path, solrParams);
+
     for (String node : nodes) {
       String baseUrl =
           coreContainer.getZkController().getZkStateReader().getBaseUrlV2ForNodeName(node);
-      String url = baseUrl + "/cluster/files" + path + "?localDelete=true";
-      HttpDelete del = new HttpDelete(url);
-      // invoke delete command on all nodes asynchronously
-      coreContainer.runAsync(() -> Utils.executeHttpMethod(client, url, null, del));
+      try (var solrClient =
```
dsmiley, line 523: why create a new client?

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -180,23 +184,35 @@ private boolean fetchFileFromNodeAndPersist(String fromNode) {
...
+        request.setResponseParser(new InputStreamResponseParser(null));
+        var response = solrClient.requestWithBaseUrl(baseUrl, request::process).getResponse();
+        is = (InputStream) response.get("stream");
         metadata =
-            Utils.executeGET(
-                coreContainer.getUpdateShardHandler().getDefaultHttpClient(),
-                baseUrl + "/node/files" + getMetaPath(),
-                Utils.newBytesConsumer((int) MAX_PKG_SIZE));
+            Utils.newBytesConsumer((int) MAX_PKG_SIZE).accept((InputStream) response.get("stream"));
         m = (Map<?, ?>) Utils.fromJSON(metadata.array(), metadata.arrayOffset(), metadata.limit());
```
dsmiley, line 198: Please replace needless use of the special InputStreamResponseParser with a standard JsonMapResponseParser.  Perhaps don't even do that; we don't have to use JSON, I think; Solr will by default negotiate the format and get you a NamedList. (Map like thing).

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -221,17 +242,18 @@ boolean fetchFromAnyNode() {
...
-          Object nl =
-              Utils.executeGET(
-                  coreContainer.getUpdateShardHandler().getDefaultHttpClient(),
-                  reqUrl,
-                  Utils.JAVABINCONSUMER);
-          if (Utils.getObjectByPath(nl, false, Arrays.asList("files", path)) != null) {
+          var solrClient = coreContainer.getDefaultHttpSolrClient();
+          var resp = solrClient.requestWithBaseUrl(baseUrl, request::process).getResponse();
+
+          if (Utils.getObjectByPath(resp, false, Arrays.asList("files", path)) != null) {
```
dsmiley, line 254: there are convenience methods on NamedList avoiding the need to refer to this awkward utility method.  Try findRecursive.

solr/modules/cross-dc/src/java/org/apache/solr/crossdc/update/processor/MirroringUpdateProcessor.java
```diff
@@ -233,7 +223,9 @@ public void processDelete(final DeleteUpdateCommand cmd) throws IOException {
         boolean done = false;
         while (!done) {
           q.set(CursorMarkParams.CURSOR_MARK_PARAM, cursorMark);
-          QueryResponse rsp = client.query(collection, q);
+          var baseUrl = cmd.getReq().getCore().getCoreContainer().getZkController().getBaseUrl();
+          var client = cmd.getReq().getCoreContainer().getDefaultHttpSolrClient();
+          QueryResponse rsp = client.requestWithBaseUrl(baseUrl, c -> c.query(collection, q));
```
dsmiley, line 228: @HoustonPutman  wouldn't it be cool if this could talk directly without HTTP?  I suppose EmbeddedSolrServer might not be appropriate since it doesn't handle collection arguments.  But the collection in this case appears to be the same collection as the present core, thus we could address this to the same core?

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
dsmiley: ping @gerlowskija

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
dsmiley: @iamsanjay you said at the meetup that this doesn't work sometimes.  I'm not surprised but I'd like to hear more about specifically why it doesn't work for particular callers that we might like to use, such as here.

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -223,15 +219,12 @@ private boolean fetchFileFromNodeAndPersist(String fromNode) {
         filedata.reset();
         persistToFile(filedata, metadata);
         return true;
-
-      } catch (SolrServerException | IOException e) {
+      } catch (SolrServerException e) {
         throw new SolrException(SolrException.ErrorCode.SERVER_ERROR, "Error fetching data", e);
+      } catch (IOException ioe) {
+        throw new SolrException(SERVER_ERROR, "Error persisting file", ioe);
```
dsmiley, line 225: notice the needless lopsidedness in both throw statements.  Use the import.

+1 to your changes today BTW.

iamsanjay:
@dsmiley I think it doesn't even matter If we pass only zkAddress, or the combination of both zkHost and zkChroot.  

```
void testZkClientClusterStateProviderCtor() throws IOException {
        try (
        var stateProvider = new ZkClientClusterStateProvider(Collections.singletonList("localhost:2181"), "/solr");
        var stateProvider1 = new ZkClientClusterStateProvider(Collections.singletonList("localhost:2181/solr"), null);
        var stateProvider2 = new ZkClientClusterStateProvider("localhost:2181/solr");){
            assertEquals(stateProvider.getZkHost(), stateProvider2.getZkHost());
            assertEquals(stateProvider.getZkHost(), stateProvider1.getZkHost());
        }
    }
```

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
dsmiley: I now understand that we must do this because we need the metadata in ByteBuffer form to persist it later, therefore we need the InputStream.

dsmiley (approved):

solr/modules/cross-dc/src/java/org/apache/solr/crossdc/update/processor/MirroringUpdateProcessor.java
dsmiley: I filed #2963

solr/core/src/java/org/apache/solr/cloud/Overseer.java
dsmiley: @iamsanjay this is my only un-addressed comment.  albeit minor.

dsmiley:
@iamsanjay is anything stopping this from getting merged?  Tests still pass?  Need a CHANGES.txt saying something like: "Most remaining usages of Apache HttpClient in Solr switched to Jetty HttpClient (HTTP 2)."

epugh:
Thank you for this wonderful effort..   Getting through this migration gets us much closer to Solr 10!

PR #2742, opened 2024-10-04
SOLR-17477: Support custom aggregates in plugin code

timatbw:
https://issues.apache.org/jira/browse/SOLR-17477

# Description

Better support for writing custom aggregates in 3rd party plugin code, outside of Solr.

# Solution

Opened up visibility of certain classes that are necessary to write custom aggregates, and allow them to register by type so that they are supported just the same as field/query/range/heatmap etc

# Tests

Adjusted existing test that confirms a custom aggregate can be written outside the solr facet package

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/facet/DocValuesAcc.java
+++ b/solr/core/src/java/org/apache/solr/search/facet/DocValuesAcc.java
@@ -37,7 +37,7 @@ import org.apache.solr.schema.SchemaField;
  * org.apache.lucene.index.DocValues}
  */
 public abstract class DocValuesAcc extends SlotAcc {
-  SchemaField sf;
+  protected SchemaField sf;
 
   public DocValuesAcc(FacetContext fcontext, SchemaField sf) throws IOException {
     super(fcontext);
--- a/solr/core/src/java/org/apache/solr/search/facet/DocValuesAcc.java
+++ b/solr/core/src/java/org/apache/solr/search/facet/DocValuesAcc.java
@@ -37,7 +37,7 @@ import org.apache.solr.schema.SchemaField;
  * org.apache.lucene.index.DocValues}
  */
 public abstract class DocValuesAcc extends SlotAcc {
-  protected SchemaField sf;
+  protected final SchemaField sf;
 
   public DocValuesAcc(FacetContext fcontext, SchemaField sf) throws IOException {
     super(fcontext);
--- a/solr/core/src/java/org/apache/solr/search/facet/FacetModule.java
+++ b/solr/core/src/java/org/apache/solr/search/facet/FacetModule.java
@@ -423,7 +423,7 @@ public class FacetModule extends SearchComponent {
   }
 
   public static class FacetLongMerger extends FacetSortableMerger {
-    long val;
+    private long val;
 
     @Override
     public void merge(Object facetResult, Context mcontext) {
@@ -444,7 +444,7 @@ public class FacetModule extends SearchComponent {
   // base class for facets that create buckets (and can hence have sub-facets)
   public abstract static class FacetBucketMerger<FacetRequestT extends FacetRequest>
       extends FacetMerger {
-    FacetRequestT freq;
+    protected final FacetRequestT freq;
 
     public FacetBucketMerger(FacetRequestT freq) {
       this.freq = freq;
@@ -485,7 +485,7 @@ public class FacetModule extends SearchComponent {
   }
 
   public static class FacetQueryMerger extends FacetBucketMerger<FacetQuery> {
-    FacetBucket bucket;
+    private FacetBucket bucket;
 
     public FacetQueryMerger(FacetQuery freq) {
       super(freq);
--- a/solr/core/src/java/org/apache/solr/search/facet/FacetModule.java
+++ b/solr/core/src/java/org/apache/solr/search/facet/FacetModule.java
@@ -370,7 +370,7 @@ public class FacetModule extends SearchComponent {
   }
 
   // base class for facet functions that can be used in a sort
-  abstract static class FacetSortableMerger extends FacetMerger {
+  public abstract static class FacetSortableMerger extends FacetMerger {
     public void prepareSort() {}
 
     @Override
@@ -386,7 +386,7 @@ public class FacetModule extends SearchComponent {
     public abstract int compareTo(FacetSortableMerger other, FacetRequest.SortDirection direction);
   }
 
-  abstract static class FacetDoubleMerger extends FacetSortableMerger {
+  public abstract static class FacetDoubleMerger extends FacetSortableMerger {
     @Override
     public abstract void merge(Object facetResult, Context mcontext);
 
@@ -422,7 +422,7 @@ public class FacetModule extends SearchComponent {
     }
   }
 
-  static class FacetLongMerger extends FacetSortableMerger {
+  public static class FacetLongMerger extends FacetSortableMerger {
     long val;
 
     @Override
@@ -442,7 +442,8 @@ public class FacetModule extends SearchComponent {
   }
 
   // base class for facets that create buckets (and can hence have sub-facets)
-  abstract static class FacetBucketMerger<FacetRequestT extends FacetRequest> extends FacetMerger {
+  public abstract static class FacetBucketMerger<FacetRequestT extends FacetRequest>
+      extends FacetMerger {
     FacetRequestT freq;
 
     public FacetBucketMerger(FacetRequestT freq) {
@@ -483,7 +484,7 @@ public class FacetModule extends SearchComponent {
     }
   }
 
-  static class FacetQueryMerger extends FacetBucketMerger<FacetQuery> {
+  public static class FacetQueryMerger extends FacetBucketMerger<FacetQuery> {
     FacetBucket bucket;
 
     public FacetQueryMerger(FacetQuery freq) {
--- a/solr/core/src/java/org/apache/solr/search/facet/FacetParser.java
+++ b/solr/core/src/java/org/apache/solr/search/facet/FacetParser.java
@@ -22,7 +22,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import org.apache.solr.common.SolrException;
 import org.apache.solr.common.params.SolrParams;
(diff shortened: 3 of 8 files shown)


mkhludnev:
I like it. Thanks @timatbw. I'm able to merge, if there's an approve from the second 👀.

solr/core/src/java/org/apache/solr/search/facet/DocValuesAcc.java
```diff
@@ -37,7 +37,7 @@
  * org.apache.lucene.index.DocValues}
  */
 public abstract class DocValuesAcc extends SlotAcc {
-  SchemaField sf;
+  protected SchemaField sf;
```
cpoerschke, line 40: maybe
```suggestion
  protected final SchemaField sf;
```

solr/core/src/java/org/apache/solr/search/facet/FacetModule.java
```diff
@@ -422,7 +422,7 @@ public static int compare(double a, double b, FacetRequest.SortDirection directi
     }
   }
 
-  static class FacetLongMerger extends FacetSortableMerger {
+  public static class FacetLongMerger extends FacetSortableMerger {
     long val;
```
cpoerschke, line 426:
```suggestion
    private long val;
```

timatbw: yeah good point, there's not much point making the class public if subclasses can't access the field, I think I need to review these and make them protected if subclasses could reasonably need to access them.

solr/core/src/java/org/apache/solr/search/facet/FacetParser.java
```diff
@@ -138,21 +139,30 @@ public Object parseFacetOrStat(String key, Object o) throws SyntaxError {
...
+    ParseHandler fieldParser = (p, k, a) -> new FacetFieldParser(p, k).parse(a);
+    REGISTERED_TYPES.put("field", fieldParser);
+    REGISTERED_TYPES.put("terms", fieldParser);
+    REGISTERED_TYPES.put("query", (p, k, a) -> new FacetQueryParser(p, k).parse(a));
+    REGISTERED_TYPES.put("range", (p, k, a) -> new FacetRangeParser(p, k).parse(a));
+    REGISTERED_TYPES.put("heatmap", (p, k, a) -> new FacetHeatmap.Parser(p, k).parse(a));
+    REGISTERED_TYPES.put("func", (p, k, a) -> p.parseStat(k, a));
+  }
+
+  public static void registerParseHandler(String type, ParseHandler parseHandler) {
```
cpoerschke, line 158: Thanks for the https://issues.apache.org/jira/browse/SOLR-17477 comment w.r.t. how this is called!

mkhludnev (changes requested):
please proceed with access modifiers

mkhludnev (approved):

solr/core/src/java/org/apache/solr/search/facet/FacetParser.java
mkhludnev: Not sure where to discuss it. But registering anything in static initializer is asking for trouble, imho. I'd rather hookup a component via solrconfig.xml (or a plugin??), which just calls `registerParserHandler()` for custom handlers.
Anyway it's a matter of taste, and should block this PR, unless someone propose to make  `FP.registerParserHandler` `protected`. WDYT?

solr/core/src/java/org/apache/solr/search/facet/FacetParser.java
```diff
@@ -138,21 +139,30 @@ public Object parseFacetOrStat(String key, Object o) throws SyntaxError {
     return parseFacetOrStat(key, type, args);
   }
 
+  public interface ParseHandler {
+    Object doParse(FacetParser<?> parent, String key, Object args) throws SyntaxError;
```
mkhludnev, line 143: btw, is there an other exemplar of such a naming convention like `doSometh` ? why not just `parse`?

mkhludnev:
We need to add changes here. I made a https://github.com/mkhludnev/solr/commit/292efa541a42353000a68b3445fd95f2dc03acf5 adding one more demo test. I don't know why, but I couldn't push it into this PR, nor open PR against `[timatbw:](https://github.com/timatbw/solr/)`. 
@timatbw please merge that commit into this PR. Also would be great to improve registering parsers. WDYT?

solr/core/src/java/org/apache/solr/search/facet/FacetParser.java
timatbw: Yeah it's perhaps not the best way of hooking things up, especially as the whole process is already kicked off from solrconfig.xml where you declare a valueSourceParser. I was relying on the point at which that custom parser object is instantiated to run a static initialiser to also register it for json parsing too. But maybe it's safer to invert that and have the Solr code register the custom parser if it implements json parsing. I'll have another look and think about the alternatives.

My main reason for suggesting custom code uses a static block to register is when you have many hundreds of SolrCores in your CoreContainer and each one would be instantiated and call the register method, when really you might only need it done once. But now thinking if you run a mixed workload with different solrconfig and they're not homogenous, you don't want static registering maybe (each SolrCore is distinct)

timatbw: Although the 'standard set' of registered types in the static block above probably is appropriate as those are built-in and common to all SolrCores.

solr/core/src/java/org/apache/solr/search/facet/FacetParser.java
timatbw: No particular strong feelings on this, I tend to use `doSomething` when it's an inner helper method called from an outer/wrapper method that is the `something(..)` but in this case it's not really directly called. In practice I think people would use lambda anyway and never actually write this method name. But happy to change it to `parse`!

solr/core/src/java/org/apache/solr/search/facet/FacetParser.java
mkhludnev: `REGISTERED_TYPES` are shared across cores. It means:
- one core might inject a custom parser, it appears across all cores. If it hooks up some large state it remains in heap after core unloading (it might be called as a leak)
- `registerPH` might be called with one of the standard name bringing some surprises to other cores.

mkhludnev:
oh gosh.. I figure it out how to update PR as a maintainer. The problem was I did it by https auth from Idea, but it seems it requested token only for my account and can't push into someone else's even if `Maintainers are allowed to edit this pull request.` is turn on. Old good ssh did a thing.

solr/core/src/java/org/apache/solr/search/facet/FacetParser.java
timatbw: Yes all good points, I'm going to change this to remove the statics and make it per-core which probably fits better with how it's hooked in too

solr/core/src/java/org/apache/solr/search/facet/FacetParser.java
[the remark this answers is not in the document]
timatbw: This is already how it is, the calling code allows the parser to return either a FacetRequest or a AggValueSource hence that Object type was needed here too. It's not ideal, and perhaps in practice most people will want to produce a AggValueSource only. Creating new subclasses of FacetRequest outside the package is harder, and I've not yet looked at how feasible it is to do that from a plugin (it's not the focus of this issue). But it would be interesting if eventually you can define custom types of faceting too.

timatbw:
I've just realised the registration part of this work is not necessary! All that's really needed is the public modifier changes. I'm going to simplify the PR down to just that bit.

The reason the registration isn't needed is because you can already achieve the desired results using the existing mechanism `type:func` which lets you access the structured json args via local params. So a ValueSourceParser you might have in your plugin code can be configured this way by pulling values from the local params accessible in the `FunctionQParser` object, instead of positional arguments. All you do is declare your json facet like so

```
{ mycalculation:{ type:func, func:myaggregate, myfield:something, mylimit:2 } }
```

then all those key-value params are available inside your implementation of `ValueSourceParser`

```
    @Override
    public ValueSource parse(FunctionQParser fp) throws SyntaxError {
        if (fp.getLocalParams() != null) {
            String fieldName = fp.getLocalParams().get("myfield");
            Integer limit = fp.getLocalParams().getInt("mylimit");
            return new MyAgg(fieldName, limit);
        } else {
            return new MyAgg(fp.parseArg(), fp.parseArg());
        }
    }
```

timatbw:
I'm going to (force) push my local branch to this PR again, which will overwrite/replace the commits you've added on top @mkhludnev just so you know. But I think the previous approach we were working on isn't necessary so I'm simplifying this PR a lot

mkhludnev:
> you can already achieve the desired results using the existing mechanism `type:func`

I'm made a simple experiment. ~~this time, you won't drop my commit~~ Mind the file location, I wish to build in a code snippet into RefGuide! Meanwhile, strawman approach with funcs, doesn not let custom parser to read a nested map (ok it let it read,but only stringified). Also, custom parser has no access to a parent facet parser, it might be necessary for parsing nested facets, but maybe FacetParser might be instantiated right there, haven't check it yet.  
https://github.com/mkhludnev/solr/commit/85870c467a0ffb038f45280c7a4bde58a4237b0e#diff-01a34eb1905d3f728ae9cb3d433945541b5a69fbe02e3f8da21ccd38ebe859edR148
The q no. 1 are we ok with the scope limited by `type:func` hook, in this case, are facet parser changes necessary?

timatbw:
Good points, yes I did notice that the parent parser would not be available via the simple localParams approach, but I didn't realise that SolrParams only support String values not arbitrary nested objects (that's a shame).

For my use-cases, I'm OK with just the String-typed map and no need for the parent parser, but I guess a more general solution could support those as well.

The solution I had been working on before I found `type:func` was to add a 2nd `AggValueSource parse(FacetParser parent, Object args)` method to `ValueSourceParser` (with a default returning null), and then in FacetParser after the switch cases, do a `getSolrRequest().getCore().getValueSourceParser(type)` and if there's one, call the new method and use that as the return value. That ended up being fairly minimal changes.

mkhludnev:
I'm not sure it may work, since ValueSourceParser is invoked by FuncQParser with fixed interface.

mkhludnev:
Here's my idea:

 - What's already done is fine and really needed, let's just add a refguide section with a good example and go ahead with piggyback on AggValueSourceParser. 
 
Then, in another PR we may choose: 

- plan B.1: introduce a new entity: FacetParser in solrconfig.xml and /config api. That's doable, unless vetoed. 
- plan B.2: introduce factory methods in FacetModule instead of constructors' calls for all parsers and facets. And maybe even extract them into a super factory for FacetModule. Then demo how users may to override the standard FacetModule by own descendant extending a parsing routine. It let us to stick with existing extensions/config system. 

@timatbw, wdyt?

mkhludnev:
> you can already achieve the desired results using the existing mechanism `type:func`

Wait a sec. I read here https://solr.apache.org/guide/solr/latest/query-guide/json-facet-api.html#stat-facet-functions 

> Unlike all the facets discussed so far, Aggregation functions (also called facet functions, analytic functions, or metrics) do not partition data into buckets. Instead, they calculate something over all the documents in the domain.

Can the proposed approach with `type:func` yield multiple buckets? What we want to achieve here?

timatbw:
Correct, the work in this Jira+PR is only about custom metrics i.e. calculating a value for each bucket. It's not about creating custom faceting types for making buckets in the first place. That would be interesting too, but is quite a different bit of work because the FacetModule is not currently set up to do that (a lot more classes would need to be open for extension).

mkhludnev:
> I don't know much about the RefGuide as I've not written for it before, will look.

made a short clue https://github.com/mkhludnev/solr/commit/c1c1a3e662eb63db6baf859bc6684e6c42631fe2

mkhludnev:
> PR is only about custom metrics i.e. calculating a value for each bucket. It's not about creating custom faceting types for making buckets in the first place.

does it mean `FacetParser` change is not necessary?

timatbw:
> does it mean `FacetParser` change is not necessary?

The PR now only changes that class to make it public but that was probably only needed in the original changes around registering parsers, so yes it may not need to be changed at all now.

PR #2753, opened 2024-10-10
SOLR-17490: Check for existence of perl executable and skip if it doesnt exist

epugh:
https://issues.apache.org/jira/browse/SOLR-17490

# Description

Make the first time developer experience better with Solr.    Today, everytime someone clones Solr for the first time, especially on Windows, they get bitten by the need to install perl..   just to product "Changes.html".   

# Solution

Add a check for the perl executable and log a warning if it isn't there, and continue on your Gradle process...

# Tests

Manual

### DIFF
--- a/gradle/documentation/changes-to-html.gradle
+++ b/gradle/documentation/changes-to-html.gradle
@@ -78,14 +78,13 @@ class ChangesToHtmlTask extends DefaultTask {
     def output = new ByteArrayOutputStream()
     
     // Check if the perl executable exists
-    def perlExecutable = project.externalTool("perl")
-    if (!file(perlExecutable).exists()) {
-        logger.warn("WARNING: Perl is not installed or could not be found at: ${perlExecutable}, skipping creating Changes.html")
+    if (!perlExists()) {
+        logger.warn("WARNING: Perl is not installed, skipping creating Changes.html")
         return
     }
-    
+
     def result = project.exec {
-      perlExecutable
+      executable project.externalTool("perl")
       standardInput changesFile.newInputStream()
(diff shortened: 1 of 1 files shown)


epugh:
@patrbraun this one is for you!

patrbraun (commented):
It skips the Perl now. Although it also skips and shows the warning even when Perl is installed.

dsmiley:
I so want this; let's just require Java!
But we actually also require "npm" as well, albeit it'll get auto-downloaded.  Any way, one thing at a time.

gradle/documentation/changes-to-html.gradle
```diff
@@ -76,8 +76,16 @@ class ChangesToHtmlTask extends DefaultTask {
 
   def toHtml(File versionsFile) {
     def output = new ByteArrayOutputStream()
+    
+    // Check if the perl executable exists
+    def perlExecutable = project.externalTool("perl")
+    if (!new File(perlExecutable).exists()) {
```
janhoy, line 82: I believe it can be simplified a bit:
```suggestion
    if (!file(perlExecutable).exists()) {
```

janhoy (approved):
I like this. I actually made a stab at porting the perl script with ChatGPT, but it was too large for context window :) Would probably be a good candidate for a groovy script too, or simply a Java class..

gradle/documentation/changes-to-html.gradle
epugh: i tried gpt as well and no joy!

epugh:
@patrbraun  would you mind taking another stab with the latest change?

patrbraun:
It's working for me now.

patrbraun (approved):

PR #2754, opened 2024-10-10
SOLR-17488: CLI: Resolve -d conflicts

epugh:
https://issues.apache.org/jira/browse/SOLR-17488

# Description

see jira

# Solution

   * Deprecate -d for server-dir in RunExmapleTool. 
   *  Support --server-dir in bin/solr and  bin/solr.cmd 
   * Deprecate -d for delay in PostTool to avoid any confusion with conf-dir.

# Tests

bats and java tests

### DIFF
--- a/solr/bin/solr.cmd
+++ b/solr/bin/solr.cmd
@@ -1202,7 +1202,7 @@ REM Run the requested example
   -Dlog4j.configurationFile="file:///%DEFAULT_SERVER_DIR%\resources\log4j2-console.xml" ^
   -Dsolr.install.symDir="%SOLR_TIP%" ^
   -classpath "%DEFAULT_SERVER_DIR%\solr-webapp\webapp\WEB-INF\lib\*;%DEFAULT_SERVER_DIR%\lib\ext\*" ^
-  org.apache.solr.cli.SolrCLI run_example --script "%SDIR%\solr.cmd" -e %EXAMPLE% -d "%SOLR_SERVER_DIR%" ^
+  org.apache.solr.cli.SolrCLI run_example --script "%SDIR%\solr.cmd" -e %EXAMPLE% --server-dir "%SOLR_SERVER_DIR%" ^
   --url-scheme !SOLR_URL_SCHEME! !PASS_TO_RUN_EXAMPLE!
 
 REM End of run_example
--- a/solr/bin/solr.cmd
+++ b/solr/bin/solr.cmd
@@ -1521,7 +1521,9 @@ for %%a in (%*) do (
    ) else (
       set "option!option!=%%a"
       if "!option!" equ "-d" set "SOLR_SERVER_DIR=%%a"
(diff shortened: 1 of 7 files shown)


epugh:
@malliaridis this is ready, I avoided deciding what to do about keep versus --delete-config since deciding that wasn't requried to resolve -d conflicts.

malliaridis (changes requested):
Thanks for creating a PR this quick. I have found a couple `-d` references, mainly in documentation, that need to be updated as well:
- [ ] bin/solr.cmd#408: `IF "%1"=="--server-dir" goto set_server_dir` should be included
- [ ] bin/solr.cmd#1522-1524 (around): In the code block only `-d` is checked, but `--server-dir` should be included as well probably
- [ ] bin/solr#1375: `-d` in error message should be `--server-dir`
- [ ] solr-control-script-reference.adoc#99: `-d` should be `--server-dir`
- [ ] solr-control-script-reference.adoc#110: `-d` should be `--server-dir`
- [ ] solr-control-script-reference.adoc#206: `-d` should be `--server-dir`
- [ ] solr-control-script-reference.adoc#281: `-d` should be `--server-dir`
- [ ] solr-control-script-reference.adoc#418: Is `-d` a valid option here?
- [ ] solr-control-script-reference.adoc#800: `-deleteConfig` should be `--delete-config`
- [ ] solr-control-script-reference.adoc#824: `-deleteConfig` should be `--delete-config`
- [x] solr-control-script-reference.adoc#983: `-d` should be `--server-dir`
- [x] post-tool.adoc#46: `-d` should be removed

epugh:
Thanks for the additional review @malliaridis and I think I've got them sorted.  I love how hte docs look with hte long form of the variables, much easier to read.

solr/bin/solr.cmd
```diff
@@ -1520,7 +1521,9 @@ for %%a in (%*) do (
    ) else (
       set "option!option!=%%a"
       if "!option!" equ "-d" set "SOLR_SERVER_DIR=%%a"
+      if "!option!" equ "--server-dir" set "SOLR_SERVER_DIR=%%a"
       if "!option!" equ "-s" set "SOLR_HOME=%%a"
+      if "!option!" equ "--solr-home" set "SOLR_HOME=%%a"      
       if not "!option!" equ "-s" if not "!option!" equ "-d" (
         set "AUTH_PARAMS=!AUTH_PARAMS! !option! %%a"
       )
```
malliaridis, line 1529: Is this code block still gonna work correctly if `--server-dir` and `--solr-home` is used instead?

I feel like we have to somehow get rid of these lists of options, they seem very error-prone and can be easily forgotten when we make changes.

epugh: Yeah, good darn question.  I also think we have extra options parsing in the `.cmd` that doesn't exist in `bin/solr`...   I think we have places where we could be leaning more on the existing java parsing...

epugh: I opened this: https://github.com/apache/solr/pull/2760

PR #2756, opened 2024-10-11
SOLR-17489: CLI: Deprecate variations on solr urls options

epugh:
https://issues.apache.org/jira/browse/SOLR-17489

# Description

See JIRA

# Solution

Deprecate old options in favour of `--solr-url` and `--name`.   I am really liking how this looks after going through the code...

Debating if we should be using `--collection` instead of `--name`....   That can be a seperate ticket.

# Tests

bats and java

### DIFF
--- a/solr/core/src/test/org/apache/solr/cli/PostToolTest.java
+++ b/solr/core/src/test/org/apache/solr/cli/PostToolTest.java
@@ -85,8 +85,10 @@ public class PostToolTest extends SolrCloudTestCase {
 
     String[] args = {
       "post",
-      "--solr-update-url",
-      cluster.getJettySolrRunner(0).getBaseUrl() + "/" + collection + "/update",
+      "--solr-url",
+      cluster.getJettySolrRunner(0).getBaseUrl().toString(),
+      "--name",
+      collection,
       "--credentials",
       SecurityJson.USER_PASS,
       jsonDoc.getAbsolutePath()
--- a/solr/core/src/test/org/apache/solr/cli/TestExportTool.java
+++ b/solr/core/src/test/org/apache/solr/cli/TestExportTool.java
@@ -243,8 +243,10 @@ public class TestExportTool extends SolrCloudTestCase {
 
       String[] args = {
         "export",
-        "-url",
-        cluster.getJettySolrRunner(0).getBaseUrl() + "/" + COLLECTION_NAME,
+        "--solr-url",
+        cluster.getJettySolrRunner(0).getBaseUrl().toString(),
+        "--name",
+        COLLECTION_NAME,
         "--credentials",
         SecurityJson.USER_PASS,
         "-out",
--- a/solr/core/src/test/org/apache/solr/cloud/SolrCloudExampleTest.java
+++ b/solr/core/src/test/org/apache/solr/cloud/SolrCloudExampleTest.java
@@ -115,8 +115,10 @@ public class SolrCloudExampleTest extends AbstractFullDistribZkTestBase {
 
     String[] argsForPost =
         new String[] {
-          "--solr-update-url",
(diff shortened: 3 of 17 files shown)


epugh:
@malliaridis  okay, this one is ready for another set of eyes!

epugh:
After merging this, we can then sort out the `-s` in the SolrExporter.java class.

solr/core/src/test/org/apache/solr/cli/PostToolTest.java
```diff
@@ -85,8 +85,10 @@ public void testBasicRun() throws Exception {
 
     String[] args = {
       "post",
-      "--solr-update-url",
-      cluster.getJettySolrRunner(0).getBaseUrl() + "/" + collection + "/update",
+      "--solr-url",
+      cluster.getJettySolrRunner(0).getBaseUrl().toString(),
```
malliaridis, line 89: `getBaseUrl()` extends v1 API path `/solr`, and the new logic does also add `/solr/` if `--solr-url` is used. This results to a path like `/solr/solr/collection/update`.

solr/core/src/test/org/apache/solr/cli/TestExportTool.java
```diff
@@ -243,8 +243,10 @@ public void testWithBasicAuth() throws Exception {
 
       String[] args = {
         "export",
-        "-url",
-        cluster.getJettySolrRunner(0).getBaseUrl() + "/" + COLLECTION_NAME,
+        "--solr-url",
+        cluster.getJettySolrRunner(0).getBaseUrl().toString(),
```
malliaridis, line 247: Same issue with path here.

solr/core/src/test/org/apache/solr/cloud/SolrCloudExampleTest.java
```diff
@@ -115,8 +115,10 @@ public void testLoadDocsIntoGettingStartedCollection() throws Exception {
 
     String[] argsForPost =
         new String[] {
-          "--solr-update-url",
-          solrUrl + "/" + testCollectionName + "/update",
+          "--solr-url",
+          solrUrl,
```
malliaridis, line 119: Same problem with path applies here as well, but this uses different logic.

malliaridis (changes requested):
I like the fact that we can use the same params in all tools now. I hope this change is preferred by the others as well. A few more references:

- [ ] The error message in PostTool#runImpl (line 309) could be updated to reflect the none-deprecated options
- [x] post-tool.adoc#27: -url should be split into --solr-url and --name
- [x] post-tool.adoc#115: -url should be split into --solr-url and --name
- [x] post-tool.adoc#124: -url should be split into --solr-url and --name
- [ ] post-tool.adoc#131: -url should be split into --solr-url and --name
- [x] post-tool.adoc#138: -url should be split into --solr-url and --name
- [x] post-tool.adoc#154: -url should be split into --solr-url and --name
- [x] post-tool.adoc#166: -url should be split into --solr-url and --name
- [x] post-tool.adoc#173: -url should be split into --solr-url and --name
- [x] post-tool.adoc#180: -url should be split into --solr-url and --name
- [x] post-tool.adoc#189: -url should be split into --solr-url and --name
- [x] post-tool.adoc#210: -url should be split into --solr-url and --name
- [x] post-tool.adoc#219: -url should be split into --solr-url and --name
- [x] spatal-search.adoc#74: -url should be split into --solr-url and --name

solr/core/src/test/org/apache/solr/cli/PostToolTest.java
epugh: i think I fixed it with the use of `SolrCLI.normalizeSolrUrl() in the tool`

epugh:
Will add CHANGES.txt all at once.

PR #2758, opened 2024-10-11
SOLR-17494: Remove language specific writer types

epugh:
https://issues.apache.org/jira/browse/SOLR-17494

# Description

Language specific writers types are "evaled" which isn't secure.  And not a pattern to be used.

# Solution

Remove them from main.


solr/core/src/test/org/apache/solr/response/JSONWriterTest.java
```diff
@@ -53,24 +53,17 @@ private void jsonEq(String expected, String received) {
...
 
   @Test
-  public void testTypes() throws IOException {
+  public void testSimpleJson() throws IOException {
     SolrQueryRequest req = req("q", "dummy", "indent", "off");
     SolrQueryResponse rsp = new SolrQueryResponse();
-    QueryResponseWriter w = new PythonResponseWriter();
+    QueryResponseWriter w = new JSONResponseWriter();
+    new StringWriter();
+    StringWriter buf;
```
gerlowskija, line 61: [-1] I think L60-L61 is a typo and should be the single line:

```
StringWriter buf = new StringWriter();
```

gerlowskija (commented):
One test-side typo that needs fixed, but when that's fixed this LGTM.

Oh, and maybe this should have a 10.0 CHANGES.txt entry?  (Or maybe two entries - one for 9.8 to signal the deprecation, and one for 10.0 to announce the removal?)

epugh:
Thoughts on how to mark it "deprecated" in 9x?   In some places it makes sense to put out a log message.   What would being deprecated mean here?   Just a PR that adds `@deprecated` in the `branch_9x` branch?

janhoy:
Deprecate as notice in ref guide. Also we have a separate deprecation logging mechanism.

epugh:
Can I get a LGTM and then I'll add CHANGES.txt and merge....

gerlowskija:
Woohoo, thanks @epugh !

PR #2759, opened 2024-10-12
Websocket module - starter poc

gus-asf:
DO NOT MERGE

This is a quick starter POC that I did for fun on a plane while traveling It doesn't actually do any solr things yet, just demonstrates an ability to get a socket server port up and running, and vaguely sketches how the messages might look at a high level. 

This is here for discussion only at this point, and a new PR would likely be made if this got serious.


gus-asf:
Outdated and can be recreated later whenever someone finds time to play with the idea again. Will keep branch around.

PR #2761, opened 2024-10-12
SOLR-17495: Change CLI delete command to not delete configs by default.

epugh:
https://issues.apache.org/jira/browse/SOLR-17495

# Description

See jira

# Solution

Makes --delete-config a opt in flag that specifies the user wants to delete their config along with their collection.   

This would be a 10x only change since we are changing the feature.

# Tests

Bats


epugh:
@malliaridis

solr/core/src/java/org/apache/solr/cli/DeleteTool.java
```diff
@@ -198,13 +187,13 @@ protected void deleteCollection(CloudSolrClient cloudSolrClient, CommandLine cli
...
+    boolean deleteConfig = false;
     if (cli.hasOption("delete-config")) {
-      deleteConfig = "true".equals(cli.getOptionValue("delete-config"));
+      deleteConfig = true;
     } else if (cli.hasOption("d")) {
-      deleteConfig = "true".equals(cli.getOptionValue("d"));
+      deleteConfig = true;
     } else if (cli.hasOption("deleteConfig")) {
-      deleteConfig = "true".equals(cli.getOptionValue("deleteConfig"));
+      deleteConfig = true;
```
malliaridis, line 196: Is it more readable if replaced by the following?
```suggestion
    boolean deleteConfig = cli.hasOption("delete-config")
        || cli.hasOption("deleteConfig")
        || cli.hasOption("d");
```

malliaridis (changes requested):
Thanks for the PR. I would like to notify somehow the user about the change of the default behavior by showing a message when using the delete tool, but I am not sure if there is a specific way to do so. This would only affect 9x.

We definitely have to mention that in the migration guide from 9 to 10.

A few more points to look into the bevario change:

- [x] solr-control-script-reference.adoc#799-800: the behavior described is outdated
- [x] solr/packaging/test/bats_helper.bash#74: Previously `delete_all_collections` delete configs as well if no longer used (old behavior)
- [x] solr/packaging/test/test_basic_auth.bats#60: we should probably delete the config explicitly here and check for its deletion
- [ ] solr/solr-ref-guide/modules/getting-started/pages/tutorial-films.adoc#491: we could probably delete the config explicitily here as well
- [ ] solr/solr-ref-guide/modules/getting-started/pages/tutorial-techproducts.adoc#484: we could probably delete the config explicitily here as well

solr/core/src/java/org/apache/solr/cli/DeleteTool.java
epugh: Love it...  I can't merge directly from the UI for some reason so applying manually.

epugh:
Leaving "tutorial-films.adoc" alone, as it uses `_default` and we wouldn't want to delete it.

malliaridis (approved):

epugh:
@janhoy  would you mind weighing in ont this PR since it changes a default behavior?

janhoy (commented):
I think it is A Good Thing™ to de-couple these by default. I tested a collection delete in Admin UI and it also only deletes the collection, not the configset.

For all breaking changes there should be an entry in ref-guide "Major changes" chapter...

PR #2762, opened 2024-10-14
SOLR-17456: TransactionLog ctor integrity

dsmiley:
The TransactionLog constructor can't handle an existing file being present; it shouldn't be there.  Should throw an exception in this case, NOT log a warning, leaving the object in a partially constructed state.

https://issues.apache.org/jira/browse/SOLR-17456

BTW I experienced this on a fork of Solr and I don't know the root cause.  Since applying the fix in this PR, I have not seen it since (perhaps by chance).  It should never happen!

Not sure if needs CHANGES.txt

### DIFF
--- a/solr/core/src/java/org/apache/solr/update/TransactionLog.java
+++ b/solr/core/src/java/org/apache/solr/update/TransactionLog.java
@@ -38,6 +38,7 @@ import java.util.TreeMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.lucene.util.BytesRef;
 import org.apache.solr.common.SolrException;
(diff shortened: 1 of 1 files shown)


solr/core/src/java/org/apache/solr/update/TransactionLog.java
```diff
@@ -238,9 +240,6 @@ protected TransactionLog(
       }
 
       success = true;
-
-      assert ObjectReleaseTracker.track(this);
```
dsmiley, line 242: In addition to the scope of the PR, I did a little move of this line to the bottom where it's supposed to be.

dsmiley:
I'm going to merge this in a couple days without further review.  It's a super simple one.

PR #2763, opened 2024-10-14
SOLR-17485 Remove CheckLoggingConfiguration; unnecessary

dsmiley:
Also note we extend HttpFilter instead of ServletFilter.

https://issues.apache.org/jira/browse/SOLR-17485


epugh (approved):
LGTM.   Nice to see some extra classes go away and us get back closer to just using the direct classes.

One nit...   We have a mix of `_request` and `request` for variable names, can we eliminate the `_request` naming?

dsmiley:
Addressing your nit on renaming `_request` to `request` wasn't fully possible for various reasons at first, so I started pulling on that thread, doing some refactoring that ultimately led to me doing that rename.  CC @gus-asf for review on my changes (commit messages explain what I did too).  Happy to decouple such changes to another PR; it's out of scope.  

Granted the scope has been somewhat flexible, maybe too much... this started with me wondering, why can't we just use HttpServletRequest instead of the old ServletRequest.  Oh hey look, we have some old base class that isn't what I expect, doing some old logging stuff that's obsolete / useless (IMO).  So let's just subclass HttpFilter.

An aside, the excludedPath thing seems bolted on and shouldn't have been added.

epugh:
Naming looks great, and some simplifications.   I tried to understand from javadocs what `excludedPath` was all about and didn't grok it but that isn't important.

gus-asf (approved):
Looks good.

PR #2777, opened 2024-10-17
Add an option in the gradle.properties for html junit reports

gus-asf:
This is a simple enhancement that makes it easy to enable the html test report that gradle would normally produce (but we disable to save a few seconds). I frequently have a NO_COMMIT change list containing this, and so I've finally got around to formalizing it into a gradle.properties option so others can use it if they like. 

I find it far faster to click on the link displayed at the end of a failed build. YMMV, but now you won't have to edit anything that's version controlled to try it.


gus-asf:
This is small, and build only, So I intend to merge/port in a day or two unless I get feedback.

PR #2778, opened 2024-10-18
Fix assert tool url normalization

janhoy:
Draft PR to get a green state in branch_9x

Tests used to pass on branch_9x a few days ago. Some of the many `SolrCLI` changes touched the urlNormalization stuff so that you needed to add `/solr` at the end of some URLs passed to Solr.

In particular this was true for `AssertTool`, which I fix here by getting the SolrClient from SolrCLI util method instead of constructing an `Http2HttpSolrClient` manually. The difference is that the former adds a /solr before constructing the client, and this is consistent with other client initializations in AssertTool.

The whole Solr Base Path and `/solr` suffix is a mess, and it is different in main than in branch_9x. So this will keep cropping up for as long as we backport stuff from 10 to 9 :(

### DIFF
--- a/solr/core/src/java/org/apache/solr/cli/AssertTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/AssertTool.java
@@ -355,16 +355,16 @@ public class AssertTool extends ToolBase {
       ret += assertSolrNotRunning(cli.getOptionValue("not-started"));
     }
     if (cli.hasOption("c")) {
-      ret += assertSolrRunningInCloudMode(cli.getOptionValue("c"));
+      ret += assertSolrRunningInCloudMode(SolrCLI.normalizeSolrUrl(cli.getOptionValue("c"), false));
     }
     if (cli.hasOption("cloud")) {
-      ret += assertSolrRunningInCloudMode(cli.getOptionValue("cloud"));
+      ret += assertSolrRunningInCloudMode(SolrCLI.normalizeSolrUrl(cli.getOptionValue("cloud"), false));
     }
     if (cli.hasOption("C")) {
-      ret += assertSolrNotRunningInCloudMode(cli.getOptionValue("C"));
+      ret += assertSolrNotRunningInCloudMode(SolrCLI.normalizeSolrUrl(cli.getOptionValue("C"), false));
     }
     if (cli.hasOption("not-cloud")) {
-      ret += assertSolrNotRunningInCloudMode(cli.getOptionValue("not-cloud"));
+      ret += assertSolrNotRunningInCloudMode(SolrCLI.normalizeSolrUrl(cli.getOptionValue("not-cloud"), false));
     }
     return ret;
   }
--- a/solr/core/src/java/org/apache/solr/cli/AssertTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/AssertTool.java
@@ -32,7 +32,6 @@ import org.apache.commons.cli.Option;
 import org.apache.solr.client.solrj.SolrClient;
 import org.apache.solr.client.solrj.SolrRequest;
 import org.apache.solr.client.solrj.SolrServerException;
-import org.apache.solr.client.solrj.impl.Http2SolrClient;
 import org.apache.solr.client.solrj.request.CollectionAdminRequest;
 import org.apache.solr.client.solrj.request.HealthCheckRequest;
 import org.apache.solr.client.solrj.response.CollectionAdminResponse;
@@ -355,16 +354,16 @@ public class AssertTool extends ToolBase {
       ret += assertSolrNotRunning(cli.getOptionValue("not-started"));
     }
     if (cli.hasOption("c")) {
-      ret += assertSolrRunningInCloudMode(SolrCLI.normalizeSolrUrl(cli.getOptionValue("c"), false));
+      ret += assertSolrRunningInCloudMode(cli.getOptionValue("c"));
     }
     if (cli.hasOption("cloud")) {
-      ret += assertSolrRunningInCloudMode(SolrCLI.normalizeSolrUrl(cli.getOptionValue("cloud"), false));
+      ret += assertSolrRunningInCloudMode(cli.getOptionValue("cloud"));
     }
     if (cli.hasOption("C")) {
-      ret += assertSolrNotRunningInCloudMode(SolrCLI.normalizeSolrUrl(cli.getOptionValue("C"), false));
+      ret += assertSolrNotRunningInCloudMode(cli.getOptionValue("C"));
     }
     if (cli.hasOption("not-cloud")) {
-      ret += assertSolrNotRunningInCloudMode(SolrCLI.normalizeSolrUrl(cli.getOptionValue("not-cloud"), false));
+      ret += assertSolrNotRunningInCloudMode(cli.getOptionValue("not-cloud"));
     }
     return ret;
   }
@@ -531,7 +530,7 @@ public class AssertTool extends ToolBase {
   }
 
   private static boolean runningSolrIsCloud(String url) throws Exception {
-    try (final SolrClient client = new Http2SolrClient.Builder(url).build()) {
+    try (final SolrClient client = SolrCLI.getSolrClient(url)) {
       final SolrRequest<CollectionAdminResponse> request =
           new CollectionAdminRequest.ClusterStatus();
       final CollectionAdminResponse response = request.process(client);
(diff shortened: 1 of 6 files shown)


solr/core/src/java/org/apache/solr/cli/AssertTool.java
```diff
@@ -531,7 +530,7 @@ private static boolean isSolrRunningOn(String url) throws Exception {
   }
 
   private static boolean runningSolrIsCloud(String url) throws Exception {
-    try (final SolrClient client = new Http2SolrClient.Builder(url).build()) {
+    try (final SolrClient client = SolrCLI.getSolrClient(url)) {
```
janhoy, line 533: This single change will make `solr assert --cloud` resilient for `/solr` suffix on the URL, so can be nice to include anyway.

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
```diff
@@ -792,11 +792,10 @@ public static String normalizeSolrUrl(CommandLine cli) throws Exception {
 
           String firstLiveNode = liveNodes.iterator().next();
           solrUrl = ZkStateReader.from(cloudSolrClient).getBaseUrlForNodeName(firstLiveNode);
-          // solrUrl = normalizeSolrUrl(solrUrl, false);
+          solrUrl = normalizeSolrUrl(solrUrl, false);
         }
       }
     }
-    solrUrl = normalizeSolrUrl(solrUrl, false);
```
janhoy, line 799: Not sure about this - whether all use of this. Will caller always expect to get `/solr` stripped after calling this? Guess they do, but started by reverting some changes to see if I could get to previous state..

solr/core/src/java/org/apache/solr/cli/PostTool.java
```diff
@@ -315,8 +315,7 @@ public void runImpl(CommandLine cli) throws Exception {
         throw new IllegalArgumentException(
             "Must specify -c / --name parameter with --solr-url to post documents.");
       }
-      String url =
-          SolrCLI.normalizeSolrUrl(cli) + "/solr/" + cli.getOptionValue("name") + "/update";
+      String url = SolrCLI.normalizeSolrUrl(cli) + "/" + cli.getOptionValue("name") + "/update";
```
janhoy, line 318: Cannot always blindly add `/solr`. The tests use randomized `contextPath` which may be `/xy/zx/` for a running Jetty, and then the test will pass in a `--solr-url` such as `http://localhost:9999/xy/zx/`, and the normalization will then return the same path which is correct. Question is whether the `normalizeSolrUrl` command should append `/solr` if no context path is given at all, the same way that `SolrCLI.getSolrClient(cli)` does.

epugh:
@janhoy thanks for taking a stab at this...  I definitly have been playing "whack a mole" of adding and removing /solr/'s in various places.   I will go through this today....  I appreciate your digging in with fresh eyes on this problem.

solr/core/src/java/org/apache/solr/cli/PostTool.java
epugh: I think, and I'm tagging @gerlowskija as well, that the overall idea is that we want bare Solr urls, and then add the appropriate `/solr` or `/api` as needed.

janhoy: That's how it is on main. But the "`/` as baseUrl" code is not (fully) backPorted to 9x, so there are differences. In 99% of cases our context path is `/solr` but some tests spin up Jetty's with other paths and they will break if we start creeping in too many hardcoded `/solr`. Should be safe to do in a test here and there though.

malliaridis:
> The whole Solr Base Path and /solr suffix is a mess, and it is different in main than in branch_9x. So this will keep cropping up for as long as we backport stuff from 10 to 9 :(

I would love to have less classes touching the URL and base path. This was also one of the goals of the merge in https://github.com/apache/solr/pull/2756. Unifying the options that accept a URL should reduce the possible input variations, allowing us to simplify the logic internally.

About the remaining failing test `PostToolTest#testBasicRun`: currently passes a URL with `--solr-url` that includes `/solr`, but this is probably wrong. A `/solr` path in the URL is only allowed in `--solr-update-url`, which was deprecated and replaced in https://github.com/apache/solr/pull/2756.

solr/core/src/java/org/apache/solr/cli/AssertTool.java
epugh: Makes sense.   Also, since we have a SolrCLI.getSolrCLient, let's use it!

solr/core/src/java/org/apache/solr/cli/PostTool.java
epugh: Oh right..  It was 10x only change in SOLR-16911 that did this.

epugh:
Okay, so now the tests all passed @janhoy..   I kind of hacked in something to work for PostTool, I actually thought it would be needed on more other tools...

solr/core/src/java/org/apache/solr/cli/PostTool.java
```diff
@@ -315,8 +320,17 @@ public void runImpl(CommandLine cli) throws Exception {
...
+
+      String solrUrl = cli.getOptionValue("solr-url");
+
+      String hostContext = System.getProperty("hostContext", "/solr");
+      if (hostContext.isBlank()) {
+        log.warn("Invalid hostContext {} provided, setting to /solr", hostContext);
+        hostContext = "/solr";
+      }
+
+      solrUrl = SolrCLI.normalizeSolrUrl(solrUrl, true, hostContext) + hostContext;
```
janhoy, line 332: I thought of the same workaround, and guess I'm medium happy about it. Would of course prefer a `SolrCLI` util method that hides this such as `SolrCLI.getAndNormalizeSolrUrl(cli), and consider using it wherever we do raw `cli.getOptionValue("solr-url")`, since we'd potentially run into same issue elsewhere.

malliaridis (approved):
LGTM, although I would like to find a better way of handling URL composition. Using the v1 API should also be discouraged in the long run.

Note that I am unable to manually test on Windows, as the script is currently broken. `gradlew check` finishes without errors.

janhoy:
@epugh, @malliaridis Turns out this still fails half the time, when `hostContext` is set to `/`. I can reproduce locally, about half the runs.

See https://github.com/apache/solr/blob/branch_9x/solr/test-framework/src/java/org/apache/solr/BaseDistributedSearchTestCase.java#L137-L145

So we need some kind of extra check that we don't get two `//` in a row I suppose.

malliaridis:
@janhoy For some reason I am not surprised. I guess a `hostContext` that contains only `/` may be valid from the user's perspective, but we cannot handle it without additional workarounds and the current logic.

I would prefer to use a URL builder which we can feed the path segments with and without leading and trailing slashes and it will take care of everything. Otherwise we would have to add just another workaround, like

```java
if (hostContext.equals("/")) {
  hostContext = "";
}
```

right after the current workaround `if (hostContext.isBlank()) {...}`. Maybe it's time to reconsider the normalization logic?Concatenating URL paths should definitely not be done by simply appending the paths, as we cannot guarantee the presence / absence of slashes.

_Using URL builders_
I looked into the implementation of `org.apache.http.client.utils.URIBuilder` and `jakarta.ws.rs.core.UriBuilder` (both present right now) and only the jakarta implementation can take leading and trailing slashes into account, if used properly. This means that code like:

```java
solrUpdateUrl = UriBuilder.fromUri(SolrCLI.normalizeSolrUrl(solrUrl, true, hostContext))
    .path(hostContext)
    .path(cli.getOptionValue("name")) // "testBasicRun"
    .path("/update")
    .path("/with/")
    .path("slashes/")
    .path("/again")
    .build();
// generates http://127.0.0.1:52371/solr/testBasicRun/update/with/slashes/again

```
would correctly generate the path.

But using `UriBuilder#segment` which looks like a more suitable function would escape the slashes:
 ```java
solrUpdateUrl = UriBuilder.fromUri(SolrCLI.normalizeSolrUrl(solrUrl, true, hostContext) + hostContext)
    .segment(cli.getOptionValue("name"), "/update", "/with/", "slashes/", "/again")
    .build();
// generates http://127.0.0.1:52215/solr/testBasicRun/%2Fupdate/%2Fwith%2F/slashes%2F/%2Fagain
```

Maybe we could use a more robust solution like that?

epugh:
Of the two URI builders, I think we would want the jakarta since at some point Apache HTTP Client goes away.   

What about just looking for `//` and removing one of them?  In the 9x line, does that get us through the problem.   

I remember when we looked at the hostContext a while ago and the reason we changed to just requiring `/solr` was because we thought it wasn't properly supported to use anything else anyway in Solr.   Like, it might pass in the tests, but in the real world it would fail.   What if we just "fix" the `BaseDistributedSolrTestCase` to not do anything but `/solr`.....

PR #2786, opened 2024-10-21
SOLR-17504: CoreContainer calls UpdateHandler.commit.

bruno-roustant:
https://issues.apache.org/jira/browse/SOLR-17504

# Description

This proposal is about unifying calls to UpdateHandler.commit.

CoreContainer.reload may call directly IndexWriter.commit if the core to reload is readonly (property of the Collection). This is an issue because it bypasses some logic in UpdateHandler.commit. In addition, the current code commits without taking the commit lock.

DirectUpdateHandler2.closeWriter may commit if the update log is not empty (not committed yet). But it does not call DirectUpdateHandler2.commit(), it copies commit() code instead for some reasons (comments in the code mention test failures). 

# Solution

Make CoreContainer.reload call UpdateHandler.commit.

Add the same call to DirectUpdateHandler2.shouldCommit() as in the commit() method, otherwise commit metadata are lost. In addition shouldCommit() can be extended correctly, for example in the solr-sandbox encryption module which needs to intercept calls to shouldCommit().

# Tests

A new DirectUpdateHandlerWithUpdateLogTest is added to test specifically that the UpdateHandler.shouldCommit hook is called. It's a new test because it requires the update log to test the UpdateHandler.closeWriter logic, but DirectUpdateHandlerTest does not support the update log.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -95,6 +95,7 @@ import org.apache.solr.common.cloud.Replica;
 import org.apache.solr.common.cloud.Replica.State;
 import org.apache.solr.common.cloud.SolrZkClient;
 import org.apache.solr.common.cloud.ZkStateReader;
+import org.apache.solr.common.params.ModifiableSolrParams;
 import org.apache.solr.common.util.CollectionUtil;
 import org.apache.solr.common.util.ExecutorUtil;
 import org.apache.solr.common.util.IOUtils;
@@ -138,6 +139,8 @@ import org.apache.solr.metrics.SolrMetricManager;
 import org.apache.solr.metrics.SolrMetricProducer;
 import org.apache.solr.metrics.SolrMetricsContext;
 import org.apache.solr.pkg.SolrPackageLoader;
+import org.apache.solr.request.LocalSolrQueryRequest;
+import org.apache.solr.request.SolrQueryRequest;
 import org.apache.solr.request.SolrRequestHandler;
 import org.apache.solr.request.SolrRequestInfo;
 import org.apache.solr.search.CacheConfig;
@@ -153,6 +156,8 @@ import org.apache.solr.security.PKIAuthenticationPlugin;
 import org.apache.solr.security.PublicKeyHandler;
 import org.apache.solr.security.SecurityPluginHolder;
 import org.apache.solr.security.SolrNodeKeyPair;
+import org.apache.solr.update.CommitUpdateCommand;
+import org.apache.solr.update.DirectUpdateHandler2;
 import org.apache.solr.update.SolrCoreState;
 import org.apache.solr.update.UpdateShardHandler;
 import org.apache.solr.util.OrderedExecutor;
@@ -2061,13 +2066,16 @@ public class CoreContainer {
           RefCounted<IndexWriter> iwRef = core.getSolrCoreState().getIndexWriter(null);
           if (iwRef != null) {
             IndexWriter iw = iwRef.get();
-            // switch old core to readOnly
-            core.readOnly = true;
             try {
               if (iw != null) {
-                iw.commit();
+                SolrQueryRequest req = new LocalSolrQueryRequest(core, new ModifiableSolrParams());
+                CommitUpdateCommand cmd =
+                    new DirectUpdateHandler2.ClosingCommitUpdateCommand(req, false);
+                core.getUpdateHandler().commit(cmd);
               }
             } finally {
+              // switch old core to readOnly
+              core.readOnly = true;
               iwRef.decref();
             }
           }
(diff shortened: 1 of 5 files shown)


solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -2061,13 +2066,16 @@ public void reload(String name, UUID coreId) {
           RefCounted<IndexWriter> iwRef = core.getSolrCoreState().getIndexWriter(null);
           if (iwRef != null) {
             IndexWriter iw = iwRef.get();
-            // switch old core to readOnly
-            core.readOnly = true;
             try {
               if (iw != null) {
-                iw.commit();
+                SolrQueryRequest req = new LocalSolrQueryRequest(core, new ModifiableSolrParams());
+                CommitUpdateCommand cmd =
```
dsmiley, line 2072: nit:  "var" would be clear and short

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -2061,13 +2066,16 @@ public void reload(String name, UUID coreId) {
           RefCounted<IndexWriter> iwRef = core.getSolrCoreState().getIndexWriter(null);
           if (iwRef != null) {
             IndexWriter iw = iwRef.get();
-            // switch old core to readOnly
-            core.readOnly = true;
```
dsmiley, line 2065: I suppose doing this before was to protect changes happening after the commit sneaking it.  Why did you move it?

solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java
```diff
@@ -1115,4 +1117,23 @@ public CommitTracker getCommitTracker() {
...
     return softCommitTracker;
   }
+
+  /**
+   * The purpose of this {@link CommitUpdateCommand} extension is to indicate that the {@link
+   * #commit(CommitUpdateCommand)} caller is closing the core and does not want to open any
+   * searcher. Because even if {@link CommitUpdateCommand#openSearcher} is false, a new real-time
+   * searcher will be opened with a standard {@link CommitUpdateCommand}. This internal {@link
+   * ClosingCommitUpdateCommand} is a way to not add another public flag to {@link
+   * CommitUpdateCommand}.
```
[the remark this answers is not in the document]
bruno-roustant: I like this enum idea very much.
If we replace two public boolean fields by an enum field, would it be a public API change, so limited to main branch?
Ideally, I would like to have this change in 9.x to get it in solr-sandbox.

solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java
```diff
@@ -971,8 +971,10 @@ public void closeWriter(IndexWriter writer) throws IOException {
 
             // todo: refactor this shared code (or figure out why a real CommitUpdateCommand can't
             // be used)
```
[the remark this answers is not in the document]
bruno-roustant: Yes, also unclear to me.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
bruno-roustant: Hum, good point. I moved it because I didn't realize your point and UpdateHanlder.commit() checks the code is not readonly, otherwise stops with an exception.
Actually, with your nice idea of an enum in CommitUpdateCommand, the intent would be clear and commit() would not reject if the core is closed after the commit. And we could restore the core.readonly = true at the right location here.

solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java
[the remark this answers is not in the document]
dsmiley: Definitely merely an internal change in terms of just the CommitUpdateCommand internals (I'm aware its fields are all public :-(.  No backwards-compatibility concern.  

Of course users could pass those booleans at an HTTP params level and it still needs to work, ultimately being converted to the appropriate enum.  Not sure if you want to add a new http param for the enum as an alternative... I suppose you might.

bruno-roustant:
I tried to refactor to have an enum in CommitUpdateCommand instead of 3 booleans, but the change in callings classes was becoming large and odd with dependent classes. I finally reverted to keep the 2 booleans and add a private 3rd one. There is no refactoring, but it keeps clarity. Tell me what you think.

solr/core/src/java/org/apache/solr/update/CommitUpdateCommand.java
```diff
@@ -21,9 +21,21 @@
 
 /** A commit index command encapsulated in an object. */
 public class CommitUpdateCommand extends UpdateCommand {
+
```
dsmiley, line 24: Awesome docs!

dsmiley (approved):

PR #2787, opened 2024-10-21
Resolve CLI tool issues for hostContext "/"

malliaridis:
# Description

This is the continuation of #2778, as it seems there are some special cases that cause URL parsing issues.

# Solution

The changes leverage jakarta's `UriBuilder` to avoid custom logic for URL path concatenation, which seems to cause some issues for cases where path segments like the `hostContext` may be `/`. Additionally, the currently used `SolrCLI#normalizeSolrUrl`is updated to behave differently for the special case where `hostContext` is `/`.

Any usage of `PostTool#appendPath` is replaced to avoid custom concatenation.

# Tests

Since the `appendPath` method is removed, the tests for that method are also removed.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cli/PostTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/PostTool.java
@@ -19,6 +19,7 @@ package org.apache.solr.cli;
 import static java.nio.charset.StandardCharsets.US_ASCII;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
+import jakarta.ws.rs.core.UriBuilder;
 import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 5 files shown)


malliaridis:
@janhoy I had to override the system property for `hostContext` to run into test failures, but I am not sure if this just executed a different execution path. If you have the time to check this solution let me know.

epugh:
Should this PR be marked "Ready for Review" versus Draft?

solr/core/src/java/org/apache/solr/cli/PostTool.java
```diff
@@ -330,15 +331,24 @@ public void runImpl(CommandLine cli) throws Exception {
       }
 
       solrUrl = SolrCLI.normalizeSolrUrl(solrUrl, true, hostContext) + hostContext;
-      String url = solrUrl + "/" + cli.getOptionValue("name") + "/update";
-      solrUpdateUrl = new URI(url);
+
+      solrUpdateUrl =
+          UriBuilder.fromUri(SolrCLI.normalizeSolrUrl(solrUrl, true, hostContext))
+              .path(hostContext)
```
janhoy, line 337: What if the `--solr-url` is a remote URL? Should we not then require the baseUrl to include the correct `hostContext`? Or have we documented a way for the user to pass in the `-DhostContext=/foo` to `bin/solr post`? I have not seen such doc, meaning hostContext will always be `/solr`, unless `bin/solr` is executed from within an installation that has overridden this in `bin/solr.in.sh`?

solr/core/src/java/org/apache/solr/cli/PostTool.java
```diff
@@ -330,15 +331,24 @@ public void runImpl(CommandLine cli) throws Exception {
...
 
     } else if (cli.hasOption("solr-update-url")) {
       String url = cli.getOptionValue("solr-update-url");
       solrUpdateUrl = new URI(url);
     } else if (cli.hasOption("name")) {
-      String url = SolrCLI.getDefaultSolrUrl() + "/solr/" + cli.getOptionValue("name") + "/update";
-      solrUpdateUrl = new URI(url);
+      solrUpdateUrl =
+          UriBuilder.fromUri(SolrCLI.getDefaultSolrUrl())
+              .path("solr")
```
janhoy, line 348: Here, when `--solr-url` or `--solr-update-url` is not given, meaning it will be localhost, we always use `/solr`. But what if the user is at an installation with customized `contextPath` (I doubt that happens very often)?

janhoy (approved):
Thanks for jumping in here. I think the URI building makes sense and is much safer.

Put in a few questions.

> I had to override the system property for hostContext to run into test failures

Not sure I understand where you mean here.

All in all, I doubt we'd hear a single complaint if we hardcoded this to `/solr` in the tools, but you never know, since we still formally support customizing `contextPath` in 9.x.

solr/core/src/java/org/apache/solr/cli/PostTool.java
malliaridis: The `hostContext` is indeed barely documented. The user should be able to pass a custom `hostContext` as you described via the system properties, and the `SolrCLI` class executes some custom logic to preserve these arguments as well (due to special treatment of `-D`).

I also avoided to make any changes in the bahavior and only replaced the concatenation with a URI builder.

solr/core/src/java/org/apache/solr/cli/PostTool.java
malliaridis: That is a good question. Eventually out of context of this PR, but a case we should eventually test. Seems like custom contexts in general may not have been tested and documented enough.

PR #2789, opened 2024-10-22
SOLR-17467: Fix Windows start script on 9x and optimize deprecation messages

malliaridis:
https://issues.apache.org/jira/browse/SOLR-17467

# Description

The Windows start script on 9x behaves differently than on main / 10.0, and fails with the message "mode. was unexpected at this time.". This mesage is coming from the deprecation warnings.

# Solution

The fix moves the deprecation warnings to the dedicated mode setter functions and uses `echo` instead of `@echo`, which probably caused the execution error.

Additionally, this commit adds support for the `--user-managed` mode that allows the user to suppress the deprecation warning by being explicit about the current mode.

# Tests

Existing tests were updated to reflect the updated messages and confirm the behavior for `--user-managed`, `--cloud` and no option (default fallback).

# Checklist

### DIFF
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -775,7 +775,15 @@ if [ $# -gt 0 ]; then
         -c|--cloud|-cloud)
             SOLR_MODE="solrcloud"
             PASS_TO_RUN_EXAMPLE+=("-c")
+
+            echo -e "Solr will start in SolrCloud mode by default in version 10, and you will no longer need to pass in -c or --cloud flag.\n"
             shift
+        ;;
+        --user-managed)
+          # Allow the user to use the user-managed flag to suppress the warning about th change
+          # of the default mode in v10.
+          SOLR_MODE="user-managed"
+          shift
         ;;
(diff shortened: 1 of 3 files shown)


epugh (commented):
LGTM.   Do we need the same thing then for the bin/solr linux version?   Or does this only apply to windows?

malliaridis:
Good point, I only checked the Windows script because of the bug, but I will have a look about the --user-managed option support.

janhoy:
LGTM

malliaridis:
I've now updated the behavior for `bin/solr` as well, so the script prints messages for `--cloud/-c` and when not using `--user-managed` (and falls back to default / user-managed mode). That way the user is notified when he uses a combination that will either no longer be supported (usage of `-c`) or the default behavior changes.

janhoy (approved):

janhoy (commented):
Ehm, bats test 87 failing:
> not ok 87 start provides warning about SolrCloud mode # in 7750 ms

PR #2792, opened 2024-10-24
SOLR-17321: Cleanup JDK version related logic in scripts

malliaridis:

# Description

With the min JDK 21 requirement some code snippets in `bin/solr.cmd` and `bin/solr` can be cleaned up.

# Solution

Cleanup obsolete code in `bin/solr.cmd` and `bin/solr`.

**To the reviewers**: Please review carefully the GC-related changes, as I am not sure what is supported and what not in the newer versions. I tried to remove only obsolete cases, but I am not sure how to test the configurations.

# Tests

No tests modified. Some of the tests currently fail when running `./gradlew check` (currently inspecting).

# Checklist

### DIFF
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -1322,8 +1322,7 @@ function start_solr() {
     exit 1
   fi
 
-  # Vector optimizations are only supported for Java 20 and 21 for now.
-  # This will need to change as Lucene is upgraded and newer Java versions are released
+  # Add vector optimizations module
   SCRIPT_SOLR_OPTS+=("--add-modules" "jdk.incubator.vector")
 
   SOLR_START_OPTS=('-server' "${JAVA_MEM_OPTS[@]}" "${GC_TUNE_ARR[@]}" "${GC_LOG_OPTS[@]}" "${IP_ACL_OPTS[@]}" \
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -1054,31 +1054,13 @@ fi
 
 # Establish default GC logging opts if no env var set (otherwise init to sensible default)
 if [ -z "${GC_LOG_OPTS}" ]; then
-  if [[ "$JAVA_VER_NUM" -lt "9" ]] ; then
-    GC_LOG_OPTS=('-verbose:gc' '-XX:+PrintHeapAtGC' '-XX:+PrintGCDetails' \
-                 '-XX:+PrintGCDateStamps' '-XX:+PrintGCTimeStamps' '-XX:+PrintTenuringDistribution' \
-                 '-XX:+PrintGCApplicationStoppedTime')
-  else
-    GC_LOG_OPTS=('-Xlog:gc*')
-  fi
-else
-  # TODO: Should probably not overload GC_LOG_OPTS as both string and array, but leaving it be for now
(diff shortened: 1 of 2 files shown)


solr/bin/solr
```diff
@@ -1054,31 +1054,13 @@ fi
...
 fi
 
 # if verbose gc logging enabled, setup the location of the log file and rotation
 if [ "${#GC_LOG_OPTS[@]}" -gt 0 ]; then
-  if [[ "$JAVA_VER_NUM" -lt "9" ]] || [ "$JAVA_VENDOR" == "OpenJ9" ]; then
-    gc_log_flag="-Xloggc"
-    if [ "$JAVA_VENDOR" == "OpenJ9" ]; then
-      gc_log_flag="-Xverbosegclog"
-    fi
-    if [ -z ${JAVA8_GC_LOG_FILE_OPTS+x} ]; then
```
epugh, line 1077: what does line 1077 mean?  Is it possible that it saying that if this variable "JAVA8_GC_LOG_FILE_OPTS" isn't set then we do the the GC_LOG_OPTS?

malliaridis: From my understanding that is right. The reason I removed this check is because we do not support Java 8, and as the environment variable indicates, this is only relevant for users running SOlr with Java 8 and setting this value.

I couldn't find any references in the documentation, so since it is not documented, I assume that this value is always not set for any Solr version 10+ that only supports Java 21+, and the condition is therefore always `true`.

If a condition is here relevant, we should definitely use a suitable name other than `JAVA8_*` and reintroduce the check.

solr/bin/solr
```diff
@@ -1054,31 +1054,13 @@ fi
...
-                 '-XX:+PrintGCDateStamps' '-XX:+PrintGCTimeStamps' '-XX:+PrintTenuringDistribution' \
-                 '-XX:+PrintGCApplicationStoppedTime')
-  else
-    GC_LOG_OPTS=('-Xlog:gc*')
-  fi
-else
-  # TODO: Should probably not overload GC_LOG_OPTS as both string and array, but leaving it be for now
-  # shellcheck disable=SC2128
-  GC_LOG_OPTS=($GC_LOG_OPTS)
+  GC_LOG_OPTS=('-Xlog:gc*')
```
epugh, line 1057: I googled to confirm the `-Xlog:gc*` syntax...    This is some nice cleanup!

solr/bin/solr
epugh: That makes sense.   I do sort of wish we had a bit more ways of testing these settings in our `.bats` tests, especially as with Java major versions coming out every six months, we may find we are using more of these configuration options.   Your cleanup makes sense to me!

epugh (approved):
LGTM.  Thanks for answering questions!  Let's get this in!   I noticed you changed back to Draft?

solr/bin/solr
malliaridis: Yeah, I think this requires a combination of benchmarking and testing to also tweak the configurations in the future. But that's not (yet) my expertise.

malliaridis:
> I noticed you changed back to Draft?

I switched it into draft to investigate the test that failed. I am not sure if it is related.

PR #2796, opened 2024-10-25
Fix release wizard to remove from Solr space before attempting Lucene

anshumg:
We should just remove attempting to cleanup the Lucene space once we do the Solr 10.0 release. Right now the wizard fails when it tries to cleanup the Lucene space because the 9x Solr releases are not found there.


janhoy:
We have a VOTE out to decide that 8.11.4 is our latest supported 8.x release, so I think we safely can remove the "lucene" deletion step both on main and branch_9x already. Then we can go manually delete the solr 8.x artifacts from cdn at some point..

PR #2801, opened 2024-10-27
SOLR-17151 Check limits between calls to components in SearchHandler

gus-asf:
https://issues.apache.org/jira/browse/SOLR-17151

This change seeks to ensure that a request that has exceeded a query limit will not continue processing (potentially expensive) components before returning partial results in response to a query. At the point of the creation of this PR, there's a couple failing tests to track down, but there's a good chance those will just be some places where the tests expect book-keeping in debugMap and shardsInfo that has now been short-circuited (and may or may not actually need to be retained). Happy to accept feedback while I track those down.


solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -626,44 +697,97 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
+    return true;
+  }
 
-      int pos = rb.shortCircuitedURL.indexOf("://");
-      String shardInfoName =
-          pos != -1 ? rb.shortCircuitedURL.substring(pos + 3) : rb.shortCircuitedURL;
-      shardInfo.add(shardInfoName, nl);
-      rsp.getValues().add(ShardParams.SHARDS_INFO, shardInfo);
+  private static String stageInEnglish(int nextStage) {
+    // This should probably be a enum, but that change should be its own ticket.
```
sigram, line 732: +1 to both.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -626,44 +697,97 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
+        .maybeExitWithPartialResults(
+            () ->
+                "["
+                    + when
+                    + "] Limit(s) exceeded prior to "
+                    + c.getName()
+                    + " in "
+                    + components.stream()
+                        .map(SearchComponent::getName)
+                        .collect(Collectors.toList()));
```
sigram, line 790: Maybe we should build a list of all component names in `initComponents()` and keep it around?

solr/core/src/test/org/apache/solr/core/ExitableDirectoryReaderTest.java
```diff
@@ -70,9 +70,12 @@ public void testPrefixQuery() throws Exception {
 
     // this time we should get a query cache hit and hopefully no exception?  this may change in the
     // future if time checks are put into other places.
-    assertJQ(req("q", q, "timeAllowed", "1", "sleep", sleep), assertionString);
+    // 2024-4-15: it did change..., and now this fails with 1 or 2 ms and passes with 3ms... I see
+    // no way this won't be terribly brittle. Maybe TestInjection of some sort to bring this back?
```
sigram, line 74: Maybe use `CallerSpecificQueryLimit` (example in `TestQueryLimits.testQueryLimits()`.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBSolrClient.java
```diff
@@ -306,9 +306,12 @@ public synchronized Endpoint nextOrError(Exception previousEx) throws SolrServer
         suffix = ":" + zombieServers.keySet();
       }
       // Skipping check time exceeded for the first request
-      if (numServersTried > 0 && isTimeExceeded(timeAllowedNano, timeOutTime)) {
+      // Ugly string based hack but no live servers message here is VERY misleading :(
+      if ((previousEx != null && previousEx.getMessage().contains("Limits exceeded!"))
+          || (numServersTried > 0 && isTimeExceeded(timeAllowedNano, timeOutTime))) {
         throw new SolrServerException(
-            "Time allowed to handle this request exceeded" + suffix, previousEx);
+            "The processing limits for to this request were exceeded, see cause for details",
```
sigram, line 313: Yes!

sigram (approved):
LGTM.

solr/core/src/test/org/apache/solr/core/ExitableDirectoryReaderTest.java
fsparv: I don't think that helps. The test wants to verify that the result was cached and is hoping that the request **_won't_** time out because of that, but that only worked because timeAllowed was only measured/triggered in a small area of code sensitive to the availability of the cache. Now we check it all over the place. If we raise the limit too much it's not a test, and if this test gets starved for resources it will fail spuriously.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
fsparv: The supplier is only evaluated if the exception is thrown.

solr/core/src/test/org/apache/solr/core/ExitableDirectoryReaderTest.java
fsparv: What we really need for this test is a cache admin command that can list the contents of the a cache (and one would expect a manual clear capability too, though that's not needed here) One might be able to get by with a metrics call, looking for an increment in cache hits, but that's pretty heavy reliance on coincidence rather than explicit testing of the actual desired behavior.

fsparv: I just looked at the ability to check directly. There is some hope since we could 
```
    SolrCores solrCores = ExitableDirectoryReaderTest.h.getCoreContainer().solrCores;
    List<SolrCore> cores = solrCores.getCores();
    for (SolrCore core : cores) {
      // identify core we need here...
    }
```

but assuming we can identify the appropriate core, there isn't even an internal api to check caches. Would look for something steming from `core.getSearcher().get().???` but we would need to add an accessor for the queryCache  and then on SolrCache a way to inspect the keys of the cache... that seems way beyond what this change should attempt.

gus-asf:
Crave seems to be stalled entirely. Tests pass locally, merging

PR #2805, opened 2024-10-28
SOLR-17516:  LBHttpSolrClient: support HttpJdkSolrClient (multiple class verson)

jdyer1:
This is a minimal PR to add Load Balancing functionality for the HttpJdkSolrClient.  Included is test refactoring, adding the *JdkClient as a random choice for the integration test.

The result is not ideal:  We have two layers of abstract classes and 3 implementations.  Of the 3 implementations 1 is deprecated and the other two share all code.  If we really are getting rid of the deprecated client for 10.0, we could easily consolidate in the future.  A generic LB Solr Client is likely best but users may find it an annoying API change that only yields marginal improvement.


solr/solrj/src/test/org/apache/solr/client/solrj/impl/LBHttp2SolrClientIntegrationTest.java
```diff
@@ -119,22 +114,45 @@ public void tearDown() throws Exception {
...
+  }
+
   public void testSimple() throws Exception {
     final var baseSolrEndpoints = bootstrapBaseSolrEndpoints(solr.length);
-    try (LBHttp2SolrClient client =
-        new LBHttp2SolrClient.Builder(httpClient, baseSolrEndpoints)
-            .withDefaultCollection(solr[0].getDefaultCollection())
-            .setAliveCheckInterval(500, TimeUnit.MILLISECONDS)
-            .build()) {
+    try (var h = client(baseSolrEndpoints)) {
```
dsmiley, line 150: h?  the type isn't clear enough to use 'var' if the name is merely 'h'.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/LBHttp2SolrClientIntegrationTest.java
```diff
@@ -337,8 +349,27 @@ public void startJetty() throws Exception {
         fail("TESTING FAILURE: could not grab requested port.");
       }
       this.port = newPort;
-      //      System.out.println("waiting.........");
-      //      Thread.sleep(5000);
+    }
+  }
+
+  private static class LBClientHolder implements AutoCloseable {
```
dsmiley, line 355: Wouldn't closing lbClient close the delegate?  If it doesn't, shouldn't it?   Then LBClientHolder could go away.

dsmiley (commented):
Thanks for doing this!

solr/solrj/src/test/org/apache/solr/client/solrj/impl/LBHttp2SolrClientIntegrationTest.java
jdyer1: Well, looking at LBSolrClient#close (from 2018) -- from which all 3 implementations inherit -- it doesn't seem to.  And before I modified this test, it had both the inner and outer clients in the autoclosable try/catch.  I agree that this seems wrong and perhaps was an oversight long ago.  But then again, because we make the user create their own inner client as a separate step, perhaps it was thought it is the opener's responsibility to explicitly close.

dsmiley: > we make the user create their own inner client as a separate step, perhaps it was thought it is the opener's responsibility to explicitly close

Ah; okay that makes sense; I didn't look previously.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttpSolrClientBase.java
```diff
@@ -0,0 +1,288 @@
...
+import org.apache.solr.client.solrj.SolrClient;
+import org.apache.solr.client.solrj.SolrServerException;
+import org.apache.solr.client.solrj.request.IsUpdateRequest;
+import org.apache.solr.client.solrj.request.RequestWriter;
+import org.apache.solr.common.SolrException;
+import org.apache.solr.common.util.NamedList;
+import org.slf4j.MDC;
+
+/**
+ * LBHttp2SolrClient or "LoadBalanced LBHttp2SolrClient" is a load balancing wrapper around {@link
```
dsmiley, line 40: references to Http2SolrClient here should be replaced; could just refer to an Http based SolrClient.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttpSolrClientBase.java
```diff
@@ -0,0 +1,288 @@
...
+ *
+ * <p><b>When to use this?</b><br>
+ * This can be used as a software load balancer when you do not wish to set up an external load
+ * balancer. Alternatives to this code are to use a dedicated hardware load balancer or using Apache
+ * httpd with mod_proxy_balancer as a load balancer. See <a
+ * href="http://en.wikipedia.org/wiki/Load_balancing_(computing)">Load balancing on Wikipedia</a>
+ *
+ * @since solr 8.0
+ */
+public abstract class LBHttpSolrClientBase<C extends HttpSolrClientBase> extends LBSolrClient {
```
dsmiley, line 98: Do we need subclassing?  In other words, can we have a general (not abstract) LB SolrClient that accepts an HttpSolrClientBase ?

solr/solrj/src/test/org/apache/solr/client/solrj/impl/LBHttp2SolrClientIntegrationTest.java
dsmiley: instead of a holder, could client return an anonymous subclass that propagates the close?

dsmiley:
Instead of subclasses of LB....SolrClient, ...Cloud...SolrClient, ...Http...ClusterStateProvider, we can have concrete impls that delegate to a SolrClient that is provided.  It's a vision anyway; hopefully achievable?

CC @janhoy @gerlowskija as well

janhoy:
> Instead of subclasses of LB....SolrClient, ...Cloud...SolrClient, ...Http...ClusterStateProvider, we can have concrete impls that delegate to a SolrClient that is provided. It's a vision anyway; hopefully achievable?

Isn't this a bit much to aim for in 10.0? I'd rather get a `CloudJdkSolrClient` ready for 10.0 with our current class hierarchy to offer a no-deps soljr, than getting stuck in a large refactoring.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/LBHttp2SolrClientIntegrationTest.java
jdyer1: The holder is just to avoid copy/paste in these test methods.  I'd rather test the actual class and not a subclass.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttpSolrClientBase.java
jdyer1: Yes, we do not need to subclass and I agree a generic LBHttpSolrClient would be better than this.  Doing this would be an API change, and although we can change APIs with 10.0 coming, I wonder if it is worth annoying our users for what seems like a very slight improvement?

jdyer1:
I am torn about what to do about the lengthy doc-comment which was copy/pasted from `LBHttpSolrClient` to `LBHttp2SolrClient`.  I suppose the right thing to do at this point is to copy/paste it once again for `LBHttpJdkSolrClient`, and not to have anything elaborate in the abstract class doc-comment.  This is yet another argument for having a generic LB Client and not subclasses, but I also feel this is beyond the scope of this PR.

dsmiley:
>  this is beyond the scope of this PR

I'm not sure I agree; it's *how* the issue is approached.  Maybe there's a sunk-cost issue here

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttpSolrClientBase.java
dsmiley: I think yes; if we can't change major things in a major API then.... ? wow.
Also, again, few users use LbSolrClient; it's more of an internal SolrJ component.  I say this from experience only; we don't officially communicate such.

janhoy:
> Users generally don't refer to the LB client nor ClusterStateProvider; we shouldn't constrain ourselves, especially in a major release.

Agree, wouldn't care too much about breaking these two APIs. Also, if there is will and energy to refactor things for 10.0, it is better for users to take the hit in v10 than doing yet another change for v11.

jdyer1:
I opened a better PR #2828 with a generic client.  Please review/compare.

jdyer1:
superseded by #2828.

PR #2808, opened 2024-10-29
SOLR-17524: Call out a few changes in 10.

epugh:
https://issues.apache.org/jira/browse/SOLR-17524

# Description

We need to let folks know the CLI has changed, but I don't want to do an exhaustive list..   especially because we have bin/solr tool --help command to remind folks.  

I did call out a few big changes that folks might see to illustrate the changes.

# Solution

Some ref guide docs.  This get's committed to `main` only.

### DIFF
--- a/solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-10.adoc
+++ b/solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-10.adoc
@@ -40,7 +40,8 @@ Some key changes that you may run into are:
  * Commands that interact with Solr now all use `--solr-url` (or `-s`) plus a `--name` (or `-c`) to specify the Solr to interact with.
  * You may use `--zk-host` (or `-z`) in place of the `--solr-url` option to lookup the Solr host.
  * Basic Authentication support has been added to bin/solr via the `--credentials` (or `-u`) option.
-
+ * Some short and single-letter options have been removed to avoid conflicts or in favor to other options.
+ 
 To learn about the updated options in each CLI tool, use the `--help` option or look up the tool in the documentation.
 
 === SolrJ
--- a/solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-10.adoc
+++ b/solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-10.adoc
@@ -41,7 +41,7 @@ Some key changes that you may run into are:
  * You may use `--zk-host` (or `-z`) in place of the `--solr-url` option to lookup the Solr host.
  * Basic Authentication support has been added to bin/solr via the `--credentials` (or `-u`) option.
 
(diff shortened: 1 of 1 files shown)


solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-10.adoc
```diff
@@ -35,6 +35,14 @@ Minimum Java version for Solr 10.x is Java 21.
 The Solr CLI has gone through some significant renovations to reduce technical debt, and now functions more consistently and predictably. Most notably, CLI commands now use unix-style options, e.g. `--help` instead of `-help`.
 Users are urged to review all use of the `bin/solr` command, such as in scripts and documentation, to ensure that the correct options are being used.
 
+Some key changes that you may run into are:
+
+ * Commands that interact with Solr now all use `--solr-url` (or `-s`) plus a `--name` (or `-c`) to specify the Solr to interact with.
+ * You may use `--zk-host` (or `-z`) in place of the `--solr-url` option to lookup the Solr host.
+ * Basic Authentication support has been added to bin/solr via the `--credentials` (or `-u`) option.
+
+To learn about specific changes to each CLI tool use the `--help` option.
```
malliaridis, line 44: The `--help` does not point out the changes, but lists only the new features. So eventually we could say something like

```suggestion
To learn about the updated options in each CLI tool, use the `--help` option or look up the tool in the documentation.
```

solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-10.adoc
```diff
@@ -35,6 +35,14 @@ Minimum Java version for Solr 10.x is Java 21.
 The Solr CLI has gone through some significant renovations to reduce technical debt, and now functions more consistently and predictably. Most notably, CLI commands now use unix-style options, e.g. `--help` instead of `-help`.
 Users are urged to review all use of the `bin/solr` command, such as in scripts and documentation, to ensure that the correct options are being used.
 
+Some key changes that you may run into are:
+
+ * Commands that interact with Solr now all use `--solr-url` (or `-s`) plus a `--name` (or `-c`) to specify the Solr to interact with.
+ * You may use `--zk-host` (or `-z`) in place of the `--solr-url` option to lookup the Solr host.
+ * Basic Authentication support has been added to bin/solr via the `--credentials` (or `-u`) option.
+
```
malliaridis, line 43: Believing that we will also remove the deprecated options in 10, this could also be mentioned:

```suggestion
 * Some short and single-letter options have been removed to avoid conflicts or in favor to other options.
 
```

malliaridis (commented):
I am not sure if it is worth mentioning each change explicitly. In my opinion, this information is sufficient to get suspicious and look up all the options I am currently using, but some users may simply ignore it unless it is a clear "Replace [this] with [that]".

So for me personally it's sufficient, but maybe ask some folks that haven't worked on the migration as well?

PR #2812, opened 2024-10-29
SOLR-17526: Deprecate --out in favour of --verbose.

epugh:
https://issues.apache.org/jira/browse/SOLR-17526

# Description

--verbose already means write out solr api responses in other Solr CLI tools.  

# Solution

Deprecate --out in favour of the existing --verbose option.

# Tests
added a bats test.


malliaridis (changes requested):
A few `-out` references I've found:
- [x] post-tool.adoc#180
- [x] post-tool.adoc#193
- [ ] indexing-with-tikca.adoc#606

I think the references in `test_post.bats` can be ignored, as they will be replaced with the removal of `--out`.

PR #2815, opened 2024-10-29
SOLR-17531: Use new unified GC log options

malliaridis:
https://issues.apache.org/jira/browse/SOLR-17531

# Description

This is a follow-up to https://github.com/apache/solr/pull/2792 for fixing issues and cleaning up `GC_LOG_OPTS` configurations.

# Solution

Use the new unified GC logging configuration.

# Tests

No tests were added due to it's complexity. If you have any recommendations for how to test GC configurations, please let me know.

# Checklist


epugh (commented):
LGTM.  I love that it feels like a lot of "weight" has been taken out of these scripts.   Having said that, I wonder if you should drop a note to dev@solr.apache.org asking for another set of eyes?   Since GC is a dark art!  With no tests!

gerlowskija:
Is this PR mostly about removing redundancy, or is it changing the effective defaults to be more in line with best-practices?

e.g. G1GC - are we moving away from it towards something else, or is it merely disappearing from our scripts because the JVM itself uses it as a default, etc.

malliaridis:
I tried to keep the settings the same and remove redundancy / update to new notation. G1GC is the default GC since Java 9. With Java 21 we also have [Generational ZGC](https://inside.java/2023/11/28/gen-zgc-explainer/) which could be a nice feature to enable, but I am not familiar enough to tune the configuration. This is also why I was looking for GC experts to improve the existing configuration and make sure the config I "migrated" is free of bugs.

In JDK 17 support for some JDK 8 log-related configs has been dropped. See https://confluence.atlassian.com/confkb/unrecognized-jvm-gc-options-when-using-jdk-11-17-1002472841.html for details.

risdenk (commented):
I think I prefer the GC being explicit instead of relying on whatever JDK value is the default. The logging cleanup configs look good, but I would keep the G1GC configs.

malliaridis:
> I think I prefer the GC being explicit instead of relying on whatever JDK value is the default. The logging cleanup configs look good, but I would keep the G1GC configs.

I was thinking the same when I was editing it, thinking that if the defaults may change under certain circumstances, then we will not have a "reproducible" state. But for the same reason we would have to include all other defaults as well.

I tried to understand why this was added in the first place, and it seemed to be some pre JDK 9 stuff, where G1GC was not the default GC and had to be configured explicitly. Since we had a lot of legacy JDK configuration, I decided to remove it during the cleanup. Not sure if there were other reasons though, as I never had to configure the JDK before (besides memory).

risdenk:
> > I think I prefer the GC being explicit instead of relying on whatever JDK value is the default. The logging cleanup configs look good, but I would keep the G1GC configs.
> 
> I was thinking the same when I was editing it, thinking that if the defaults may change under certain circumstances, then we will not have a "reproducible" state. But for the same reason we would have to include all other defaults as well.
> 
> I tried to understand why this was added in the first place, and it seemed to be some pre JDK 9 stuff, where G1GC was not the default GC and had to be configured explicitly. Since we had a lot of legacy JDK configuration, I decided to remove it during the cleanup. Not sure if there were other reasons though, as I never had to configure the JDK before (besides memory).

https://issues.apache.org/jira/browse/SOLR-13394 is where g1 was changed to be the default. I still think we should be explicit about using G1GC and make an explicit change away from it not just to what the JDK defaults to.

dsmiley:
I agree with what Kevin said.  Basically, explicitly enable G1GC because it's a rather important choice to not make an explicit decision on.  But otherwise, no need to configure what doesn't need to be configured.

PR #2819, opened 2024-10-29
SOLR-17383: Fix support for deprecated options in 9x

malliaridis:
https://issues.apache.org/jira/browse/SOLR-17383

# Description

During the migration of the CLI options some situations do not cover both (old and new) options and therefore compatibility is not granted.

# Solution

This PR focuses on the options listed in `getOptions()` and adds missing checks and reads to places where they should be included.

Note that this PR focuses only on deprecated options and is not relevant for `main`. For `main` see https://github.com/apache/solr/pull/2725 and future PRs that remove deprecated options.

# Tests

No tests modified.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
@@ -260,7 +260,7 @@ public class RunExampleTool extends ToolBase {
         "techproducts".equals(exampleName) ? "sample_techproducts_configs" : "_default";
 
     boolean isCloudMode = cli.hasOption('c');
-    String zkHost = SolrCLI.getZkHost(cli);
+    String zkHost = SolrCLI.getOptionWithDeprecatedAndDefault(cli, "z", "zkHost", null);
     int port =
         Integer.parseInt(
             cli.getOptionValue('p', System.getenv().getOrDefault("SOLR_PORT", "8983")));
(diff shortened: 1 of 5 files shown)


solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
```diff
@@ -260,7 +260,7 @@ protected void runExample(CommandLine cli, String exampleName) throws Exception
         "techproducts".equals(exampleName) ? "sample_techproducts_configs" : "_default";
 
     boolean isCloudMode = cli.hasOption('c');
-    String zkHost = cli.getOptionValue('z');
+    String zkHost = SolrCLI.getZkHost(cli);
```
epugh, line 263: Nice one!   I don't think I realized this was here.

epugh (approved):
LGTM.   This is much cleaner!

epugh:
Looks like some tests need some love!

solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
malliaridis: Apparently this does not work for the specific use case, so I we have to fallback to `SolrCLI#getOptionWithDeprecatedAndDefault`.

epugh:
This ticket will also close https://issues.apache.org/jira/browse/SOLR-17409!

malliaridis:
> This ticket will also close https://issues.apache.org/jira/browse/SOLR-17409!

I avoided applying it in other places as well to keep the changes to the minimum. Should I do that @epugh in this PR?

epugh:
maybe just get this one in...  honestly, i think [SOLR-17409](https://issues.apache.org/jira/browse/SOLR-17409) may not be needed if we are in a stable 9x place, since I want to focus on main and getting us closer to 10!

malliaridis:
I was considering backporting this to `branch_9_7` but it seems a bit late for fixing. @epugh what do you think? Changes are compatible.

epugh:
If this is important to a 9.7.1, then yes.  I think what we normally do is communicate in the release thread on dev@ about these things.  Otherwise, being on branch_9x is fine because it'll be picked up in 9.8....

PR #2816, opened 2024-10-29
SOLR-17483: Cleanup zk-related logic in bin/solr.cmd

malliaridis:
https://issues.apache.org/jira/browse/SOLR-17483

# Description

With the migration to Java classes many parts of the scripts can be removed or simplified. This PR focuses on the ZK related parts.

# Solution

Removes completely all ZK related logic from `bin/solr.cmd` and fixes an issue with `AuthTool`.

# Tests

Please describe the tests you've developed or run to confirm this patch implements the feature or solves the problem.

# Checklist

Please review the following and check all that apply:

- [X] I have reviewed the guidelines for [How to Contribute](https://github.com/apache/solr/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.
- [ ] I have created a Jira issue and added the issue ID to my pull request title.
- [X] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended, not available for branches on forks living under an organisation)
- [X] I have developed this patch against the `main` branch.
- [ ] I have run `./gradlew check`.
- [ ] I have added tests for my changes.
- [ ] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

PR #2823, opened 2024-10-30
SOLR-17483: Backport cleanup zk-related logic in bin/solr.cmd

malliaridis:
https://issues.apache.org/jira/browse/SOLR-17483

# Description

See https://github.com/apache/solr/pull/2816 for details.

Additionally, this PR also fixes a bug introduced in 78920854a9530109086df0d4ba67fc05162735f6 that lead to "mode." form an echo message being executed as a command.

# Solution

See https://github.com/apache/solr/pull/2816 for details.

# Tests

Windows tests still pending. See [SOLR-17508](https://issues.apache.org/jira/browse/SOLR-17508).

# Checklist


epugh:
I know we want to add windwos testing, but I wouldn't let that hold this up.

malliaridis:
Since there are bugfixes included as well, I think we are good to go. There is no ETA for the Windows testing, so it should not be a blocker.

epugh:
I just saw a bug report (with a patch!) reported to the dev list: https://issues.apache.org/jira/browse/SOLR-17551.   Looking a bit closer, I wonder if we apply this PR to branch_9x and branch_9_7, then does this fix the reported bug as well, or if we only do the bug fix in branch_9_7, and leave this one to go to branch_9x?

malliaridis:
It seems to be backwards compatible, so there is probably no issues backport this to branch_9_7 as well. :)

malliaridis:
I've further looked into backporting this, but it would break compatibility to other deprecated commands, which we probably don't want to mess with again. I would therefore apply the patch from https://issues.apache.org/jira/browse/SOLR-17551 instead.

epugh:
> I've further looked into backporting this, but it would break compatibility to other deprecated commands, which we probably don't want to mess with again. I would therefore apply the patch from https://issues.apache.org/jira/browse/SOLR-17551 instead.

Awesome.    Could I talk you into applying it?  I don't have a windows box and it's tricky enough I don't want to break things worse.  Thanks for making a decision on which way to go!!!!

malliaridis:
Leaving this as a note: Mistakenly changed the commit message with vim during cherrypick. The correct issue is SOLR-17483, not SOLR-17484.

PR #2824, opened 2024-10-30
Actually make collections PRS enabled in tests

HoustonPutman:
Fixes issues with https://github.com/apache/solr/pull/2230, where collections didn't have the PRS flag enabled when the default was enabled.

This does break tests, so it can't be merged without fixing the issues with PRS.

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
+++ b/solr/core/src/java/org/apache/solr/cloud/api/collections/CreateCollectionCmd.java
@@ -121,6 +121,9 @@ public class CreateCollectionCmd implements CollApiCmds.CollectionApiCommand {
           message.getStr(CollectionStateProps.PER_REPLICA_STATE),
           isPRS);
     }
+    if (isPRS && !message.getBool(CollectionStateProps.PER_REPLICA_STATE, false)) {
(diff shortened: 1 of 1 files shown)


dsmiley:
So the change is to non-test code but it's actually about tests?  Would it be more accurate to say that there was a partial/broken attempt to make a cluster-wide default for PRS but this PR completes that?

nit: assuming this system property isn't really advertised/documented, I'd rather it be "solr.cloud.prs" (maybe ".enabled" suffix) and not have the word "default" in it.  No sys props needs "default" as a system prop by its very nature is a kind of default.

HoustonPutman:
It is not advertised at all, I believe it is just used for testing. But I think default in the name is fine. `Solr.cloud.prs=true`, doesn't really tell me what I'm setting to true. Maybe enabled, but that sounds like you are allowing/disallowing something. Maybe we put "test" in the name so that it's clear this is just for testing?

dsmiley:
Even if today we only use it for testing (because only we know of its secret existence), isn't the same mechanism appropriate for a cluster?  It needn't be the *only* way to set it, of course; I could imagine additional layers like cluster props or solr.xml.

`solr.cloud.prs.enabled` then

PR #2825, opened 2024-10-30
Ability to run only SolrCloud tests

dsmiley:
Is this test annotation useful?  Or maybe it's enough to run all tests in org.apache.solr.cloud ?  Albeit not all SolrCloud related tests are in this package!  (SolrJ)


janhoy (commented):
If it scratches an itch for you, then LGTM.

I have wished for a skip-integrationtests flag, but I'm afraid our test suite relies so heavily on those that running only unit tests won't be good for much.

PR #2826, opened 2024-10-31
SOLR-17534 ClusterState.getCollectionNames

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17534

This PR also calls ClusterState.size() in some places where one of the map methods were used as an intermediary.  Added javadocs to that method as well, and indicated it's an approximation.


solr/core/src/test/org/apache/solr/cloud/OverseerTest.java
```diff
@@ -758,7 +758,7 @@ private void waitForCollections(ZkStateReader stateReader, String... collections
     while (0 < maxIterations--) {
 
       final ClusterState state = stateReader.getClusterState();
-      Set<String> availableCollections = state.getCollectionsMap().keySet();
+      Set<String> availableCollections = Set.copyOf(state.getCollectionNames());
```
aparnasuresh85, line 761: `state.getCollectionNames()` might already return an unmodifiable view of keys from `immutableCollectionStates`. Is the `Set.copyOf()` necessary?

aparnasuresh85:
+1. Nice and simple. There is a failing test that needs to be addressed: TestLBHttp2SolrClient.testTwoServers

aparnasuresh85 (approved):

solr/core/src/test/org/apache/solr/cloud/OverseerTest.java
dsmiley: state.getCollectionNames returns a Collection that just-so-happens to be a Set but may end up being a List in the future.  I didn't want to choose a Set on the first introduction of this method unless we think this choice is stable.

Since this spot here is merely a test, maybe I'll simply cast it to a Set.  If one day it becomes a List or something... whatever; easy enough to switch to Set.copyOf.  I wish the JDK had a convenient option to cast-or-create.  Didn't seem worth it in this class to hand-role a few lines of code for it.

dsmiley:
testTwoServers [has been unstable lately](https://ge.apache.org/scans/tests?search.rootProjectNames=solr-root&search.timeZoneId=America%2FNew_York&tests.container=org.apache.solr.client.solrj.TestLBHttp2SolrClient&tests.test=testTwoServers) (<-- linked from build failures, conveniently); certainly isn't this PR

After maybe flipping that Set.copyOf to a Set cast in the test, I will wait till Monday to merge.  CHANGES.txt in Other section would be "Added ClusterState.getCollectionNames"

iamsanjay:
> testTwoServers [has been unstable lately](https://ge.apache.org/scans/tests?search.rootProjectNames=solr-root&search.timeZoneId=America%2FNew_York&tests.container=org.apache.solr.client.solrj.TestLBHttp2SolrClient&tests.test=testTwoServers) (<-- linked from build failures, conveniently); certainly isn't this PR

Really want to disable this one and another one from the same class (or, find better solution).  They failed so frequently that It wastes lot of time. (May be discussion for dev-list)

PR #2834, opened 2024-11-01
SOLR-17535 ClusterState.collectionStream() in lieu of getCollectionStates()

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17535

Deprecated getCollectionStates.

This PR does not include getCollectionNames, but assuming that one is merged, then after merging this one then there are no more callers of getCollectionStates.


solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -634,7 +634,7 @@ private void constructState(Set<String> changedCollections) {
           collectionWatches.watchedCollections(),
           collectionWatches.activeCollections(),
           lazyCollectionStates.keySet(),
-          clusterState.getCollectionStates());
+          clusterState.collectionStream().toList());
```
dsmiley, line 637: nice little JDK 17 convenience

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -1432,8 +1432,7 @@ private DocCollection fetchCollectionState(String coll, Watcher watcher)
                 zkClient,
                 Instant.ofEpochMilli(stat.getCtime()));
 
-        ClusterState.CollectionRef collectionRef = state.getCollectionStates().get(coll);
-        return collectionRef == null ? null : collectionRef.get();
+        return state.getCollectionOrNull(coll);
```
dsmiley, line 1435: LOL

dsmiley (commented):
So far just doing getCollectionStates.  The getCollectionsMap can be a separate PR I guess; I have that one mostly done and it has many more callers.

solr/solrj/src/java/org/apache/solr/common/cloud/ClusterState.java
```diff
@@ -382,6 +383,7 @@ void setLiveNodes(Set<String> liveNodes) {
    * Be aware that this may return collections which may not exist now. You can confirm that this
    * collection exists after verifying CollectionRef.get() != null
    */
+  @Deprecated // see collectionStream()
```
murblanc, line 386: Use `@deprecated` tag in Javadoc instead of dangling comment

solr/core/src/java/org/apache/solr/handler/designer/SchemaDesignerConfigSetHelper.java
```diff
@@ -168,24 +167,12 @@ Map<String, Object> analyzeField(String configSet, String fieldName, String fiel
...
-          collections.add(coll);
-        }
-      } catch (Exception exc) {
-        log.warn("Failed to get config name for {}", coll, exc);
-      }
-    }
-    return collections;
+    return zkStateReader()
+        .getClusterState()
+        .collectionStream()
```
murblanc, line 172: Performance regression: this forces reading all `state.json` of all collections.
I think `getCollectionNames()` should be part of this PR and used here instead.

solr/solrj/src/java/org/apache/solr/common/cloud/ClusterState.java
murblanc: And IMO this can't be deprecated for performance reasons until `getCollectionNames()` is added to replace it.

solr/solrj/src/java/org/apache/solr/common/cloud/ClusterState.java
```diff
@@ -400,29 +402,14 @@ public Set<String> getHostAllowList() {
     return hostAllowList;
   }
 
-  /**
-   * Iterate over collections. Unlike {@link #getCollectionStates()} collections passed to the
-   * consumer are guaranteed to exist.
-   *
-   * @param consumer collection consumer.
-   */
+  /** Streams the resolved DocCollections. Use this sparingly in case there are many collections. */
```
murblanc, line 405: Better use Javadoc tags to refer to other classes

solr/solrj/src/java/org/apache/solr/common/cloud/ClusterState.java
```diff
@@ -400,29 +402,14 @@ public Set<String> getHostAllowList() {
...
+  public Stream<DocCollection> collectionStream() {
+    return collectionStates.values().stream().map(CollectionRef::get).filter(Objects::nonNull);
+  }
+
+  /** Streams the resolved DocCollections. Use this sparingly in case there are many collections. */
```
murblanc, line 410: Seems that Javadoc comment was copied from above method and needs updating

dsmiley: It was deliberate as I thought it was good enough, but I'll add more words

solr/core/src/java/org/apache/solr/handler/designer/SchemaDesignerConfigSetHelper.java
dsmiley: You are saying this because of the collection name filter that was there.  As I indicated in my self-PR review, I think that was bogus/erroneous.

murblanc: My comment is unrelated to the name filter.
This code collects collection names. The change forces it to read the `state.json` of the collections even though it doesn't need the additional info. The behavior also changes, partially created collections will no longer be considered when previously they were.

I don't know what this specific class does with the collection name, but since this PR deprecates a method without providing an alternate way of achieving the same result (with comparable performance), I stand by my comment to add `getCollectionNames()` in this PR.

dsmiley: The DocCollection was being read before as well.  It's where the configSet name is.

solr/solrj/src/java/org/apache/solr/common/cloud/ClusterState.java
dsmiley: Indeed; this PR shall not be merged until getCollectionNames is (today!)

dsmiley:
Assuming tests pass, I'm hoping to merge tomorrow.

PR #2835, opened 2024-11-02
SOLR-17540: Remove Hadoop Auth Module

epugh:
https://issues.apache.org/jira/browse/SOLR-17540

# Description
Remove Hadoop Auth

# Solution

no more Hadoop Auth

# Tests

Just removing things

# Tasks

- [x] Look at `solr-tests.policy`
- [x] Do we still need `useShortName` feature, maybe only supported by hadoop-auth?
- [x] remove licenses
- [x] update versions.lock
- [ ] Look at `javax.security.auth.kerberos` in `package-list` file in docs render dir
- [x] Is Kerb stuff in Solr clients part of hadoop-auth, or to work with other setups?


epugh:
Kerb stuff appears to still work!  All tests ran.

dev-tools/scripts/refguide/htaccess.txt
```diff
@@ -1,6 +1,6 @@
...
+RedirectMatch 301 ^/guide/(aliases|audit-logging|authentication-and-authorization-plugins|basic-authentication-plugin|circuit-breakers|client-apis|cloud-screens|cluster-node-management|collection-management|collections-core-admin|configuring-logging|enabling-ssl|indexupgrader-tool|installing-solr|jvm-settings|jwt-authentication-plugin|mbean-request-handler|metrics-reporting|performance-statistics-reference|ping|plugins-stats-screen|replica-management|rule-based-authorization-plugin|securing-solr|security-ui|shard-management|solr-control-script-reference|solrcloud-recoveries-and-write-tolerance|solrcloud-with-legacy-configuration-files|taking-solr-to-production|thread-dump|upgrading-a-solr-cluster|zookeeper-access-control|cert-authentication-plugin)\.html /guide/solr/latest/deployment-guide/$1.html
```
janhoy, line 3: Instead of this, you should edit `mappings.csv` and `gen-refguide-redirects.py` similar to how the `_8_11` links for removed features in 9.x will then generate a `htaccess.txt` that redirect those pages to latest 8.x. So everything removed in 9.x should have the "latest" link redirect to a concrete 9.x version...

solr/webapp/web/partials/login.html
```diff
@@ -60,23 +60,6 @@ <h1>Basic Authentication</h1>
...
-      <a href="https://solr.apache.org/guide/solr/latest/deployment-guide/kerberos-authentication-plugin.html">
-        Solr's Kerberos documentation
-      </a>.
-    </p>
-    The response from the server was:
-    <hr/>
-    <pre>HTTP 401 {{statusText}}
-WWW-Authenticate: {{wwwAuthHeader}}</pre>
-    <hr/>
-  </div>
```
janhoy, line 78: There is also traces of `Negotiate` authScheme in `login.js`

janhoy (changes requested):
Fantastic to see all those files go, and all those external deps removed! Just a few comments..

risdenk:
There are some more hadoop auth cleanup in the security.policy

* https://github.com/apache/solr/blob/main/solr/server/etc/security.policy#L134
* https://github.com/apache/solr/blob/main/gradle/testing/randomization/policies/solr-tests.policy#L103

epugh:
> There are some more hadoop auth cleanup in the security.policy
> 
>     * https://github.com/apache/solr/blob/main/solr/server/etc/security.policy#L134
> 
>     * https://github.com/apache/solr/blob/main/gradle/testing/randomization/policies/solr-tests.policy#L103

Thanks for that!  I hope I got them all out...

solr/webapp/web/partials/login.html
epugh: thanks!

epugh:
@janhoy I tried running that script, but couldn't quite grok it.  COuld you give an example of what that scripts should be?   And let's add an example to the readme or the to script itself!

janhoy:
> @janhoy I tried running that script, but couldn't quite grok it. COuld you give an example of what that scripts should be? And let's add an example to the readme or the to script itself!

It's some time ago, and the script was made to make sure we had redirects for the old regfuide structure, and we also added in page removals. The gist is to mainain the csv file with metadata of all changes, and edit the py script to output the correct htacces. 

We likely need to make another CSV section for pages removed in 10.0 guide, and then generate the correct redeirects..

I don't remember where / how the generated htaccess file is checked in though. And I know Antora has some built-in support for generating htaccess as well, should look into it..

To be pragmatic and unblock this, I'd make a Blocker JIRA for 10.0 and maintain a list of the removed pages there, so someone can update htaccess in proper way in due time.

epugh:
I have added https://issues.apache.org/jira/browse/SOLR-17546 to track the blocker task, I don't know if I did the "label" right, there were a lot to choose from!

epugh:
Okay, I think I'm looking for an LGTM...    I've also emailed the dev list with the PR..   No one has said "don't remove it"....   Is that enought to click merge when I get an LGTM?

epugh:
Got a little too frisky, broke something in the `hdfs` module:

```
2> 4749 ERROR (SUITE-HdfsCollectionsApiTest-seed#[54C70D1C3623A1D7]-worker) [n: c: s: r: x: t:] o.a.h.h.s.n.FSNamesystem FSNamesystem initialization failed.
  2>           => org.apache.hadoop.security.KerberosAuthException: failure to login: javax.security.auth.login.LoginException: Security Exception
  2>    at org.apache.hadoop.security.UserGroupInformation.doSubjectLogin(UserGroupInformation.java:2064)
  2> org.apache.hadoop.security.KerberosAuthException: failure to login: javax.security.auth.login.LoginException: Security Exception
  2>    at org.apache.hadoop.security.UserGroupInformation.doSubjectLogin(UserGroupInformation.java:2064) ~[hadoop-client-api-3.4.0.jar:?]
  2>    at org.apache.hadoop.security.UserGroupInformation.createLoginUser(UserGroupInformation.java:733) ~[hadoop-client-api-3.4.0.jar:?]
  2>    at org.apache.hadoop.security.UserGroupInformation.getLoginUser(UserGroupInformation.java:683) ~[hadoop-client-api-3.4.0.jar:?]

```

risdenk (approved):

dsmiley:
Please do not merge this until after #2706 so we don't give Malliaridis a tough time over there.  That one should be merged soon.

dsmiley:
So Kerberos Auth is going away with this.  Be sure to highlight this in CHANGES.txt and in JIRA.  To me, Hadoop-Auth is a forgettable implementation detail (I didn't know _really_ know what this is) to the effective industry feature (I know of Kerberos).  If you search our ref guide for "Kerberos", it still exists in this PR.  See `authentication-and-authorization-plugins.adoc`.

epugh:
@malliaridis any chance you could look at this with me?   I believe the `versions.lock` and `version.props` are actually the same because I copied from `main` over into my branch the files.   I don't quite know how to tell it that `hadoop-auth/build.gradle` is actually gone~

malliaridis:
With `hadoop-auth` gone, we have only one more blocker (s3) before migrating to jetty12. :D

epugh:
> So Kerberos Auth is going away with this. Be sure to highlight this in CHANGES.txt and in JIRA. To me, Hadoop-Auth is a forgettable implementation detail (I didn't know _really_ know what this is) to the effective industry feature (I know of Kerberos). If you search our ref guide for "Kerberos", it still exists in this PR. See `authentication-and-authorization-plugins.adoc`.

@dsmiley  I wanted to follow up on this.  Is it possible that Kerberos still exists without Hadoop Auth?   I assumed Hadoop Auth was different thatn Kerberos, in that with Kerberos you could authenticate in a Kerberos environment, and that didn't require Hadoop Auth..   That Hadoop Auth was a place to validate kerberous credentials, but you might validate elsewhere?   There do seem to be lots of references to Kerberos that aren't tied to Hadoop Auth..   I think I need some guidence on where to go...   Should I just remove the rest of the Kerberos code?

dsmiley:
Hadoop Auth and Kerberos aren't the same.  [Kerberos](https://en.wikipedia.org/wiki/Kerberos_(protocol)) is an authentication protocol, [Hadoop-Auth](https://hadoop.apache.org/docs/current/hadoop-auth/index.html) is framework that implements Kerberos and more.  Hadoop-Auth seems to power several options, LDAP is another (I'm learning this now with some basic Google searches).  @risdenk [wrote about it](https://risdenk.github.io/2018/11/20/apache-solr-hadoop-authentication-plugin-ldap.html) in 2018.

The underlying point is we should communicate the effective impact to users in terms they might understand.  They might not know about the specific libraries that implement the functionality that are going away.  Granted, a user using such stuff likely knows they are using this module since they had to opt-in to get the functionality (as of Solr 9 but not prior) but it may be an overlooked detail nonetheless.

solr/modules/hadoop-auth/src/java/org/apache/solr/security/hadoop/KerberosPlugin.java
dsmiley, line 1: I assumed you were aware of what you were removing, but maybe not.

epugh: I started thinking that there were TWO things...   Kerberos in the Solr side, and then a seperate Kerberos in SolrJ capablity that somehow stays...   I'm going to check what files I may have missed...

epugh:
I ran `find . -not -path '*/\.*' -type f -exec grep -l -i "kerberos" {} +` and wow...   There are a LOT more files...

I think what I am hearing is that just removing the module and the light changes to the webapp isn't enough. There are a whole bunch of OTHER places that rely in one way or another on `hadoop-auth`?   I had some thoughts of moving it to the sandbox, but if it's this ingrained in various locations in Solr, then that really doesn't make sense.

Does this align with other folks expectations that removing Hadoop Auth would touch all these files?

```
./solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java
./solr/core/src/java/org/apache/solr/core/CoreContainer.java
./solr/core/src/java/org/apache/solr/security/AuthorizationContext.java
./solr/core/src/java/org/apache/solr/cli/AuthTool.java
./solr/core/src/java/org/apache/solr/request/json/RequestUtil.java
./solr/bin/solr
./solr/bin/solr.cmd
./solr/bin/solr.in.sh
./solr/solrj/src/test/org/apache/solr/client/solrj/impl/BasicHttpSolrClientTest.java
./solr/solrj/src/java/org/apache/solr/client/solrj/impl/Krb5HttpClientBuilder.java
./solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClient.java
./solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-8.adoc
./solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-9.adoc
./solr/solr-ref-guide/modules/deployment-guide/pages/authentication-and-authorization-plugins.adoc
./solr/solr-ref-guide/modules/deployment-guide/pages/zookeeper-access-control.adoc
./solr/solr-ref-guide/modules/deployment-guide/pages/solr-on-hdfs.adoc
./solr/solr-ref-guide/modules/deployment-guide/pages/solr-control-script-reference.adoc
./solr/solr-ref-guide/modules/deployment-guide/pages/rule-based-authorization-plugin.adoc
./solr/solr-ref-guide/modules/getting-started/pages/solr-admin-ui.adoc
./solr/webapp/web/js/angular/controllers/index.js
./solr/server/etc/security.policy
./solr/CHANGES.txt
./solr/modules/hdfs/src/java/org/apache/solr/hdfs/HdfsDirectoryFactory.java
./gradle/documentation/render-javadoc/java11/package-list
./gradle/testing/randomization/policies/solr-tests.policy
./dev-tools/scripts/refguide/new-guide.txt
./dev-tools/scripts/refguide/old-guide.txt
./dev-tools/scripts/refguide/htaccess.txt

```

dsmiley:
I'm not surprised there's a blast radius that exceeds the module itself.  For everything outside the module, we'll need to review/edit with some care and it'll be on a case-by-case basis. 

_For example_ HttpSolrClient was a surprising reference you made.  I see one method withKerberosDelegationToken but it's only called by certain tests.  Surprising to me; should have been documented as such and maybe not public.  Furthermore the functionality itself doesn't *really* reference Kerberos; it references an abstract "delegation token" concept that perhaps might be meaningful outside Kerberos -- I really don't know about that.

dsmiley:
To answer some of these questions, it's helpful to do git-blame (I use my IDE actually) sometimes with history for the selected code to find the people/issues and then follow up with anyone.

epugh:
Would someone be willing to look at what is going on in SolrDispatchFilter?   Around https://github.com/apache/solr/blob/620175a5626e69823b5aec8e20734a6352195000/solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java#L336 it appears that MAYBE there is an opportunity for some refactoring to simplify the flow...   Especially since we specifically mention the Hadoop Auth as the reason for the extra complexity..   I do not understand this bit and would love another set of eyes....   I could also see a path to updating the very lengthy comments to say "This has complexity due to hadoop auth partially, and now that it is gone there may be an opportunity for improvement"...

gus-asf:
> Would someone be willing to look at what is going on in SolrDispatchFilter? Around
> 
> https://github.com/apache/solr/blob/620175a5626e69823b5aec8e20734a6352195000/solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java#L336
> it appears that MAYBE there is an opportunity for some refactoring to simplify the flow... Especially since we specifically mention the Hadoop Auth as the reason for the extra complexity.. I do not understand this bit and would love another set of eyes.... I could also see a path to updating the very lengthy comments to say "This has complexity due to hadoop auth partially, and now that it is gone there may be an opportunity for improvement"...

I wrote that comment to memorialize several hours of digging I did back when I moved startup to a context listener. One of the things I found perplexing about SolrDispatchFilter when I first tried to understand it for that task was the lack of a call to doFilter(req,resp,filterchain) ... note that our custom version with the boolean retry doesn't count, because it doesn't make the normal call to the method specified by javax.servlet.Filter. Normally filter implementations look like:

````
  public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
      throws IOException, ServletException {

      // Do stuff, that seems important here

      if (importantStuffSeemsHappy) {
        chain.doFilter(request, response, chain);
      } else {
        // do unhappy error type stuff.
      }
     // add try/finally if there is mandatory cleanup.
    }
  }

````

So it was very weird not to find a call to doFilter in the doFilter method, nor in our custom version of it. EVENTUALLY I figured out that that call is made either in the dispatch method, OR in our auth filter (I haven't tried to prove it can't get called twice, but with just SolrDispatchFilter in play that is not currently going to cause a problem since chain.doFilter is a no-op for the final filter). One of the long term goals I have is to start pulling stuff that we are doing in this monster filter out int a series of filters, which will make the individual missions easier to understand and put the cleanup code near the instantiation code where, again it would be much easier to understand (and nesting can be easily seen to be correct). 

My impulse (not yet informed by actual attempts) is to rework our auth plugin to be auth filters. The other thing I'm pointing out in that comment is that the HadoopAuthFilter is what seems to stand in the way of writing an if block such as:

````
      if (authPlugin.authenticate(req, resp)) {            // <<< note the lack of filterchain arg

        // do searchy stuff here

        chain.doFilter(request, response, chain);
      } else {
        // do unhappy 401 error type stuff.
      }
     // add try/finally for mandatory cleanups.
    }
````

That is of course the first step to breaking auth out to it's own filter where it becomes

````
  public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
      throws IOException, ServletException {
        if (authPlugin.authenticate(req, resp)) {            
          chain.doFilter(request, response, chain);   // <<< dispatch filter is later in the chain.
        } else {
          // do unhappy 401 error type stuff.
        }
       // add try/finally for mandatory cleanups.
    }
````

The particular issue with the hadoop auth  plugin that complicates the transition is that chain.doFilter() comes before a switch statement and other code...

https://github.com/apache/solr/blob/6f94c505fc923c452c58c2a3946ee4408ea06fac/solr/modules/hadoop-auth/src/java/org/apache/solr/security/hadoop/HadoopAuthPlugin.java#L247

At least at the time of that comment it seemed that all the other plugins called chain.doFilter() at the end (or possibly in a shortcut followed by an immediate return statement). Only Hadoop auth seemed to have mandatory actions AFTER doFilter(). If it disappears, we can possibly remove the filterchain argument and make a simpler use of the return value from authenticate().

**Comment by epugh** [comment]
I am going to not touch `SolrDispatchFilter` as I don't have a good game plan to move forward with it!   Everything else is green.

For CHANGES.txt    "Remove Kerberos authentication support from Solr.  This in turn removes the Hadoop Auth module".   <-- @dsmiley ???

**Comment by dsmiley** [comment]
Cause and effect is inverted.  I suggest:

> Removed the Hadoop Auth module, and thus Kerberos authentication and other exotic options.

File: solr/core/src/java/org/apache/solr/security/RuleBasedAuthorizationPlugin.java [hunk 3]
```diff
@@ -41,17 +40,11 @@ public void init(Map<String, Object> initInfo) {
       String roleName = e.getKey();
       usersVsRoles.put(roleName, Permission.readValueAsSet(map, roleName));
     }
-    useShortName =
-        Boolean.parseBoolean(initInfo.getOrDefault("useShortName", Boolean.FALSE).toString());
```
dsmiley, line 45: what's this about?

solr/modules/hdfs/src/java/org/apache/solr/hdfs/HdfsDirectoryFactory.java
dsmiley, line 1: It's not clear that you actually needed to touch this at all.  I think the the Hadoop Auth module and HDFS's Kerberos support are completely decoupled.  WDYT @risdenk

gradle/testing/randomization/policies/solr-tests.policy
dsmiley, line 1: solr-tests.policy and security.policy need to be kept in sync, mostly.  It's annoying.  So I see you touched one in one way and the other in another way.

solr/solr-ref-guide/modules/deployment-guide/pages/solr-on-hdfs.adoc
```diff
@@ -210,41 +210,6 @@ NRTCachingDirectory max cache size.
 +
 Pass the location of HDFS client configuration files - needed for HDFS HA for example.
 
-=== Kerberos Authentication Settings
```
dsmiley, line 213: as I mentioned, I think this can stay

solr/server/etc/security.policy
```diff
@@ -131,16 +131,6 @@ grant {
...
-  permission javax.security.auth.AuthPermission "doAs";
-  permission javax.security.auth.AuthPermission "getLoginConfiguration";
-  permission javax.security.auth.AuthPermission "setLoginConfiguration";
-  permission javax.security.auth.AuthPermission "modifyPrivateCredentials";
-  permission javax.security.auth.AuthPermission "modifyPublicCredentials";
-  permission javax.security.auth.PrivateCredentialPermission "org.apache.hadoop.security.Credentials * \"*\"", "read";
-
   // needed by hadoop security
   permission java.security.SecurityPermission "putProviderProperty.SaslPlainServer";
   permission java.security.SecurityPermission "insertProvider";
```
dsmiley, line 136: can these be removed too?
(note: to test, edit solr-tests.policy, not this policy (which is production), and run tests)

epugh: okay, aligning the two files, and running tests..

solr/solr-ref-guide/modules/deployment-guide/pages/solr-on-hdfs.adoc
epugh: Restored these changes....   I really don't have a sense one way or the other...   If it's not correct, hopefully some of our Hadoop HDFS users will speak up!

solr/core/src/java/org/apache/solr/security/RuleBasedAuthorizationPlugin.java
epugh: we have this "useShortName" feature, but it turns out it's ONLY ever supported by Kerberos.   So it is dead code/unused feature with that...

solr/server/etc/security.policy
epugh: okay, removing things broke things...

solr/modules/hdfs/src/java/org/apache/solr/hdfs/HdfsDirectoryFactory.java
risdenk: the modules/hdfs should not be touched. that is totally separate.

solr/modules/hdfs/src/java/org/apache/solr/hdfs/HdfsDirectoryFactory.java
```diff
@@ -16,8 +16,6 @@
  */
 package org.apache.solr.hdfs;
 
-import static org.apache.hadoop.fs.CommonConfigurationKeysPublic.HADOOP_SECURITY_AUTHENTICATION;
```
risdenk, line 19: These changes should be reverted. modules/hdfs is totally separate from hadoop-auth.

epugh: will do!   Aside from these changes, how does the rest look?

risdenk:
I think there might be a few more places to cleanup based on running the following on your branch

```
git grep -nFi kerber | grep -Fv -e 'solr/modules/hdfs' -e 'solr-on-hdfs.adoc' -e 'solr/CHANGES.txt' -e 'solr/benchmark/src/resources/words.txt'
```

Specifically these findings:

```
solr/bin/solr.in.sh:296:# Solr internally doesn't use cookies other than for modules such as Kerberos/Hadoop Auth. If you don't need any of those
solr/core/src/java/org/apache/solr/cli/AuthTool.java:124:        + "  bin/solr auth enable --type kerberos --config \\\"<kerberos configs>\\\" [--update-include-file-only <true|false>] [-v]\n"
solr/core/src/java/org/apache/solr/core/CoreContainer.java:602:      // this caused plugins like KerberosPlugin to register its intercepts, but this intercept
solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java:328:        // obviously don't care. Kerberos plugins seem to mostly use it to satisfy the api of a
solr/core/src/java/org/apache/solr/servlet/SolrDispatchFilter.java:351:    // if it should short circuit, e.g. the Kerberos Authentication Filter will send an error and
solr/webapp/web/js/angular/controllers/index.js:27:        // Needed for Kerberos, since this is the only place from where
solr/webapp/web/js/angular/controllers/index.js:28:        // Kerberos username can be obtained.
```

It would be awesome to be able to cleanup the security policy files but I know there is some overlap with the Hadoop hdfs tests too.

risdenk:
Some added context about delegation tokens - these were a Hadoop construct at one point and expanded elsewhere to avoid hitting the KDC (kerberos server) too much so the delegation token was used in place after the initial authentication happened. Basically it was a secure token passed around instead of doing the whole roundtrip to the KDC for each call. There are some other things the delegation token can do as well (impersonation if needed). 

As David said the Hadoop authentication framework is not just Kerberos, but has a whole framework for authentication. Its similar to how Hadoop filesystem support isn't just HDFS but also S3 and some other backends. 

Jetty does have Kerberos/SPNEGO support if we want to go down that route later. The Hadoop implementation for Kerberos support was better than most other Java support out there since not many Kerberos and Java implementations historically and lots of bugs across implementations (Active Directory vs Kerby vs others). 

I do think its time to remove this module and make it fully opt in (via a plugin or separately supported module). I haven't had time to keep up with the Hadoop side development of this and don't use it anymore.

solr/core/src/java/org/apache/solr/security/RuleBasedAuthorizationPlugin.java
risdenk: Yea the short name in Kerberos is without the realm. So if the principal was `USERNAME@REALM` it would just be `USERNAME` or if the principal was `USERNAME/HOST@REALM` it would just be `USERNAME` and is a Kerberos construct to break down the whole principal

epugh:
@risdenk  I see in `solr.in.sh` the reference to `-Dsolr.http.disableCookies=true` and that it highlights the use of it in either hadoop auth or maybe in a load balancer.   Do you think i should remove that capablity?   Or just remove the text referencing hadooop auth..   I worry that i'm going to pull on a thread and break the http clients...   I could just edit it.  If you have a sense of what the cookie change *shouild* be, please do push to the PR or add a patch and I'll add it...  I'm thinking of just changing the text and leaving the rest of the cookie stuff...    Maybe add another follow on JIRA...

dsmiley:
There are ~4 classes (+ 5 tests) containing "DelegationToken".  @risdenk are these now obsolete/useless?  Would that belong here in this issue (for Hadoop Auth) or would you recommend another issue?  If another issue, can you please create it so that the wording is correct?  These classes were added in [SOLR-9200](https://issues.apache.org/jira/browse/SOLR-9200)

PR #2854, opened 2024-11-08
SOLR-17535: Deprecate ClusterState.forEachCollection

dsmiley:
Use collectionStream() instead.  Redirect callers.  A simple refactoring.

https://issues.apache.org/jira/browse/SOLR-17535

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/common/cloud/ClusterState.java
+++ b/solr/solrj/src/java/org/apache/solr/common/cloud/ClusterState.java
@@ -425,7 +425,10 @@ public class ClusterState implements MapWriter {
   /**
    * Calls {@code consumer} with a resolved {@link DocCollection}s for all collections. Use this
    * sparingly in case there are many collections.
+   *
+   * @deprecated see {@link #collectionStream()}
    */
+  @Deprecated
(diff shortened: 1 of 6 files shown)


solr/solrj/src/java/org/apache/solr/common/cloud/ClusterState.java
```diff
@@ -425,7 +425,10 @@ public Stream<DocCollection> collectionStream() {
   /**
    * Calls {@code consumer} with a resolved {@link DocCollection}s for all collections. Use this
    * sparingly in case there are many collections.
+   *
+   * @deprecated see {@link #collectionStream()}
```
murblanc, line 429: If this method was introduced a few days ago (can't remember), I think it can be removed rather than deprecated.

murblanc (approved):

solr/solrj/src/java/org/apache/solr/common/cloud/ClusterState.java
dsmiley: It has existed for some releases, thus I deprecate it.

dsmiley:
I'll merge this Monday in case there are any further comments, but I wouldn't anticipate any for this simple one.

PR #2857, opened 2024-11-09
Improve console output of auth command

malliaridis:
# Description

The console is currently printing the username and password when the user enables authentication via
```bash
bin/solr auth enable --prompt true
```
This defeats the purpose of a hidden password prompt.

# Solution

Do not print the password when basic authentication is enabled.

### DIFF
--- a/solr/core/src/java/org/apache/solr/cli/AuthTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/AuthTool.java
@@ -414,9 +414,8 @@ public class AuthTool extends ToolBase {
         final String successMessage =
             String.format(
                 Locale.ROOT,
-                "Successfully enabled basic auth with username [%s] and password [%s].",
-                username,
-                password);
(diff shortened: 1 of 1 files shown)


epugh (approved):
Nice!  It's almost embarrassing we did it the way we did it!

malliaridis:
@epugh Is something like this worth a ticket or added in the changes (planning to backport it)?

epugh:
> @epugh Is something like this worth a ticket or added in the changes (planning to backport it)?

Normally I would say no.  However, since it improves our security posture?   That might be worth going in CHANGES.txt.

epugh:
we should try and get this in....

malliaridis:
> do you want me to take it forward?

If you have the time please do. I am a bit busy  and this should already been merged. It should be backported to 9.x and 9.8 as well. Should also resolve [SOLR-17554](https://issues.apache.org/jira/browse/SOLR-17554), which was created after the PR.

PR #2861, opened 2024-11-12
SOLR-17556: Examples should run with the standard & recommended Solr home and solr data dir

epugh:
https://issues.apache.org/jira/browse/SOLR-17556

# Description
Except for -e cloud, all the other examples just store their data and configs as if you did just a normal `bin/solr start` and then created them by hand!

# Solution

Remove some use of exampledir

# Tests

Unit and manual testing


HoustonPutman (approved):

epugh:
I think stripping out the `<lib/>` makes this a breaking change for 10 only....???

HoustonPutman:
I mean, this is just changing the examples. I think we can back port this. People should not be using the tech products config in production.

solr/CHANGES.txt
```diff
@@ -237,6 +237,9 @@ led to the suppression of exceptions. (Andrey Bozhko)
 
 * SOLR-17545: Upgrade to Gradle 8.10 (Houston Putman)
 
+* SOLR-17556: Examples should run with the standard & recommended Solr home and solr data dir (except for
```
gerlowskija, line 240: [0] Minor nit - the "should" here might confuse users.  Is this how Solr behaves in 9.8?  Is it how it should behave? etc.

Wording like `"home" and "data" directories used by Solr examples have been updated to align with documented best practices.` might be a little clearer.

epugh:
Feedback in corporated.  When this is back ported we need to remember to remove the CREDENTIALS lookups ;-)

PR #2848, opened 2024-11-05
SOLR-17545: Upgrade to Gradle 8.10

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17545

PR #2864, opened 2024-11-13
SOLR-17545: Fix alternative jdk testing support

risdenk:
https://issues.apache.org/jira/browse/SOLR-17545 / https://github.com/apache/solr/pull/2848 inadvertently broke alternative JDK support that some Jenkins jobs use to run different JDKs. One deprecated use was replaced this replaces the other one.

This now matches what Lucene has - https://github.com/apache/lucene/blob/main/gradle/testing/alternative-jdk-support.gradle#L53


HoustonPutman (approved):
I could have sworn I did this already. No idea what happened...

PR #2866, opened 2024-11-16
forbidden-apis: Don't need java.solr.txt file

dsmiley:
I was looking through our forbidden-apis configuration, and I see "java.solr.txt" file that appears named this way because it was added at a time when Lucene & Solr co-existed and needed to be differentiated.  I moved the rules from there into "defaults.all.txt", and I removed the reference to this file in forbidden-apis.gradle.

But I see other stuff in forbidden-apis.gradle that I don't understand; it's a rather complicated gradle file IMO.  Hoping for your review @dweiss as you wrote this one.


gradle/validation/forbidden-apis.gradle
```diff
@@ -99,7 +99,6 @@ allprojects { prj ->
           "**.SuppressForbidden"
       ]
 
-      doFirst dynamicSignatures.curry(configurations.toolsCompileClasspath, "lucene")
```
dsmiley, line 102: I don't quite understand this line but the reference to "lucene" probably matches nothing.

PR #2868, opened 2024-11-17
SOLR-17321 More jdk upgrade

dsmiley:
I saw some more code to remove/change due to the JDK upgrade... see the separate commits

@hossman tagging you for review on the SSL compatibility check removal, especially
@dweiss for a couple build files
@HoustonPutman for the Lookup.ensureInitialized thing


risdenk (approved):

malliaridis (approved):

epugh:
I have had to futz around a bit with [SSLTestConfig.java](https://github.com/apache/solr/pull/2868/files#diff-2aa996581a8f8cc08e8529820f5ab548e04b6754f819b92465c920063b75ebb7) to get tests to run, so it would be great to get this in and see if it prevents me tweaking that class!

epugh:
> @epugh I was just removing checks that shouldn't matter anymore, if I recall. What error did you see?

I will have to try and recreate it...   I don't remember, but I do remember commenting out some methods.   It may be the migration to intellj,jdk upgrade, and changing how I run tests...  If I can duplicate the errors I'll speak up!

PR #2872, opened 2024-11-19
SOLR-17567: Improve Stream CLI implementation

epugh:
https://issues.apache.org/jira/browse/SOLR-17567

This is to improve StreamTool based on some assessments by @malliaridis!

- [x] Use of raw strings in Option#getOptionValue()
- [x] Overlapping options (-f and -e)
- [ ]  Multiple type suppressions that can be avoided
- [x]  A bug that causes specific bats tests to fail. **done in seperate PR by @malliaridis** ! 
- [x]  StreamTool.LocalCatStream overrides CatStream btu does not properly set commaDelimitedFilepaths (this may be a bug)
- [x]    Obsolete LetStream#getLetParams() was introduced
- [x]   Function StreamTool.constructStream seems obsolete and occurrences can be replaced with implementation
- [x]   Implementation of StreamTool.readExpression does not take leading spaces infront of comment lines into account
- [x]  Implementation of StreamTool.readExpression does not support */ at the end of line


epugh:
@malliaridis I've marked off the items I took care of.  The remaining ones I need more detail to more forward on.  

Regardless, it's nice to get a bit more polish!

epugh:
Finally!  All the junit tests related to StreamTool are passing!

epugh:
> Wow, did not have SolrOnHdfsSnapshotsTool under my radar. Did you find it throught the forbidden-api rules?

I did!!!

malliaridis:
About the remaining TODOs:

_Multiple type suppressions that can be avoided_

Suppressions like `@SuppressWarnings({"rawtypes"})` for classes like `List` can usually be fixed by simply using `List<?>` or using the known generic type if any present. 

_StreamTool.LocalCatStream overrides CatStream btu does not properly set commaDelimitedFilepaths (this may be a bug)_

In `CatStream` `commaDelimitedFilepaths` is a class property that is also used in `CatStream#toExpression`, which is called by other classes. By passing passing the property as parameter in `validateAndSetFilepathsInSandbox` and overriding the method does not update the class property, resulting to potential inconsistent behavior / values in other classes.

Instead, it is possible to simply remove the property from the parameters of `validateAndSetFilepathsInSandbox` (like it was before), and make the property `commaDelimitedFilepaths` visible (protected modifier). That way, overriding `validateAndSetFilepathsInSandbox` will be possible without changing the method signature.

If the description is a bit irritating, I can create a separate PR for that. :)

epugh:
Okay, this I think is almost ready to merge!

epugh:
Since this is a straight up follow on from the introduction of the StreamTool in `main` I am not adding CHANGES.txt entry.

PR #2876, opened 2024-11-19
SOLR-17069: Jetty12 + EE10

iamsanjay:
https://issues.apache.org/jira/browse/SOLR-17069 

Currently, Jetty supports several environments:

1. EE8 (Jetty10): Implements the Java EE8 specifications, supporting javax.servlet.
2. EE9 (Jetty11): Implements Jakarta EE9 specifications, introducing the namespace migration from javax.* to jakarta.*.
3. EE10 (Jetty12): Implements Jakarta EE10 specifications, including the Jakarta Servlet 6.0 specification, fully adopting the jakarta.* namespace.

> Jetty 12 providing ongoing support for both older and upcoming EE specifications [from https://github.com/jetty/jetty.project/issues/10485]

In my opinion, Jetty12 + EE8 is the best next step. It retains compatibility with javax.servlet without forcing the removal of modules that haven’t transitioned to jakarta.* yet, while benefiting from all the new features introduced in Jetty12.

#1509


risdenk:
This should link to https://issues.apache.org/jira/browse/SOLR-17069 and potentially https://github.com/apache/solr/pull/1509

gradle/libs.versions.toml
```diff
@@ -295,7 +296,7 @@ bouncycastle-bcprov = { module = "org.bouncycastle:bcprov-jdk18on", version.ref
 carrot2-core = { module = "org.carrot2:carrot2-core", version.ref = "carrot2-core" }
 carrotsearch-hppc = { module = "com.carrotsearch:hppc", version.ref = "carrotsearch-hppc" }
 carrotsearch-randomizedtesting-runner = { module = "com.carrotsearch.randomizedtesting:randomizedtesting-runner", version.ref = "carrotsearch-randomizedtesting" }
-# @keep transitive dependency for version alignment
```
risdenk, line 298: not sure why this changed

gradle/libs.versions.toml
```diff
@@ -305,7 +306,7 @@ commonsio-commonsio = { module = "commons-io:commons-io", version.ref = "commons
 cybozulabs-langdetect = { module = "com.cybozu.labs:langdetect", version.ref = "cybozulabs-langdetect" }
 dropwizard-metrics-core = { module = "io.dropwizard.metrics:metrics-core", version.ref = "dropwizard-metrics" }
 dropwizard-metrics-graphite = { module = "io.dropwizard.metrics:metrics-graphite", version.ref = "dropwizard-metrics" }
-dropwizard-metrics-jetty10 = { module = "io.dropwizard.metrics:metrics-jetty10", version.ref = "dropwizard-metrics" }
+dropwizard-metrics-jetty10 = { module = "io.dropwizard.metrics:metrics-jetty12-ee10", version.ref = "dropwizard-metrics" }
```
risdenk, line 309: does it matter that this is ee10 vs ee8 like the description?

risdenk: also the key should change from `dropwizard-metrics-jetty10`

solr/core/src/java/org/apache/solr/handler/admin/ShowFileRequestHandler.java
```diff
@@ -100,9 +100,14 @@ public class ShowFileRequestHandler extends RequestHandlerBase implements Permis
...
 
   static {
-    KNOWN_MIME_TYPES = new HashSet<>(MimeTypes.getKnownMimeTypes());
+    KNOWN_MIME_TYPES = new HashSet<>();
+    for (MimeTypes.Type type : MimeTypes.Type.values()) {
+      KNOWN_MIME_TYPES.add(type.toString());
+    }
     KNOWN_MIME_TYPES.add("text/xml");
     KNOWN_MIME_TYPES.add("text/javascript");
+    KNOWN_MIME_TYPES.add("text/csv");
```
risdenk, line 109: Why did csv/xhtml get added?

solr/core/src/test/org/apache/solr/cloud/OverseerStatusTest.java
```diff
@@ -24,6 +24,7 @@
 import org.junit.BeforeClass;
 import org.junit.Test;
 
+// @LogLevel("org.eclipse.jetty=DEBUG")
```
risdenk, line 27: left over from debugging?

solr/licenses/jetty-ee10-servlet-12.0.10.jar.sha1
```diff
@@ -0,0 +1 @@
+a8d087bde78f231084901553bb6829663b4a22f1
```
risdenk, line 1: this is ee10 servlet?

solr/test-framework/src/java/org/apache/solr/embedded/JettySolrRunner.java
```diff
@@ -350,7 +351,7 @@ private void init(int port) {
       connector.setIdleTimeout(THREAD_POOL_MAX_IDLE_TIME_MS);
 
       server.setConnectors(new Connector[] {connector});
-      server.setSessionIdManager(new DefaultSessionIdManager(server, new Random()));
+      // server.setSessionIdManager(new DefaultSessionIdManager(server, new Random()));
```
risdenk, line 354: commented out?

solr/test-framework/src/java/org/apache/solr/embedded/JettySolrRunner.java
```diff
@@ -414,16 +419,16 @@ public void contextInitialized(ServletContextEvent event) {
...
     }
 
     chain = injectJettyHandlers(chain);
 
     if (config.enableV2) {
       RewriteHandler rwh = new RewriteHandler();
       rwh.setHandler(chain);
-      rwh.setRewriteRequestURI(true);
-      rwh.setRewritePathInfo(false);
+      // rwh.setRewriteRequestURI(true);
```
risdenk, line 430: commented out?

iamsanjay:
`s3-repository` test cases started to failing as HandlerWrapper class not available anymore which was part of Jetty10 and now with Jetty12, it's part of ee8.nested.
However, spring-boot jetty-starter doesn't have anything for ee8. They do have support for ee10 webapps, and adding those runs the test successfully but there are multiple library conflict is there. Working on it.

risdenk:
s3mock was a problem when I last looked at Jetty 11/12 due to it needing to stay on 2.x due to Java 17 required in 3.x. However, now that we are on JDK 17 on main, we should be able to upgrade to s3mock 3.x - https://github.com/adobe/S3Mock/releases/tag/3.0.0 (latest being https://github.com/adobe/S3Mock/releases/tag/3.11.0)

I wonder if that will address the Jetty incompatibilities since this upgraded Spring as well.

This upgrade of s3mock should be done separately from any Jetty upgrade if possible.

iamsanjay:
For S3Mock, the library is upgraded to support Jetty12. Test cases are running now! For s3mock to deal in separate PR, I tried to maintain two different set of jetty(10 & 12), but there were conflict errors. So for now I just upgraded s3Mock libs to also use Jetty12. 

Facing some issues with gcs-repository. Running successfully as Intellij test but failing If run via gradlew. 

```
./gradlew :solr:modules:gcs-repository:test --tests "org.apache.solr.gcs.GCSIncrementalBackupTest" -Ptests.jvms=96 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=90E261AA14A5CB20 -Ptests.timeoutSuite=600000! -Ptests.file.encoding=US-ASCII --debug
```


```
 java.lang.IllegalStateException: Unable to locate keystore resource file in classpath: SSLTestConfig.hostname-and-ip-missmatch.keystore
        at __randomizedtesting.SeedInfo.seed([90E261AA14A5CB20]:0)
        at org.apache.solr.util.SSLTestConfig.<init>(SSLTestConfig.java:121)
        at org.apache.solr.util.SSLTestConfig.<init>(SSLTestConfig.java:83)
        at org.apache.solr.util.RandomizeSSL$SSLRandomizer.createSSLTestConfig(RandomizeSSL.java:129)
        at org.apache.solr.SolrTestCaseJ4.buildSSLConfig(SolrTestCaseJ4.java:475)
        at org.apache.solr.SolrTestCaseJ4.setupTestCases(SolrTestCaseJ4.java:298)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
```

I did changed some code in SSLTestConfig
```
 - trustStore = keyStore = Resource.newClassPathResource(resourceName);
 + trustStore = keyStore = ResourceFactory.root().newSystemResource(resourceName);
```

iamsanjay:
Recent commit fix the test cases related to V2 api. For now, it's just a workaround and I'll discuss with Jason to come up with a better solution. Earlier we were setting all the headers null, however the change in library does not provide one universal method `removeHeader` but separate methods to deal with all the headers. 

Not valid anymore:
```
response.setHeader(name, null)
```
Changed to:
```
response.getHeaderNames().stream()
            .filter(name -> !name.equalsIgnoreCase("Content-Length"))
            .forEach(name -> response.setHeader(name, ""));
        response.setContentLength(-1);
```

iamsanjay:
Next, there is one issue with the `InstrumentedHandler` which is used to track metrics for JettyHandlers. However, with the upgrade that class is not valid anymore. Though dropwizard/metrics have newer version available for Jetty12 but that is only for EE10 (`InstrumentedEE10Handler`), nothing for EE8 AFAIK.

solr/core/src/java/org/apache/solr/api/V2HttpCall.java
```diff
@@ -434,7 +434,12 @@ protected void executeCoreRequest(SolrQueryResponse rsp) {
         // SolrCore counter
         core.close();
         core = null;
-        response.getHeaderNames().stream().forEach(name -> response.setHeader(name, null));
+        // Skip specific headers
+        // workaround for response.setHeader(name, null)
+        response.getHeaderNames().stream()
```
gerlowskija, line 439: This looks good for now IMO.

Longer term I'd love to see this code go away in favor of having the v2 API stuff get handled by its own servlet, but we're pretty far from that being workable for now.  It is a little hacky in the interim though 😦

fsparv:
> In my opinion, Jetty12 + EE8 is the best next step.

> does it matter that this is ee10 vs ee8 like the description?

https://issues.apache.org/jira/browse/SOLR-17503 (also linked from the ticket referenced above ticket). I don't have anything against continued support for javax per se, but there are definitely folks I've worked with that would benefit from being able to eliminate javax classes from their stack. Specifically they got tripped up by the appearance of jakarta based jetty versions (upgraded from other dependencies), which were then providing jakarta HttpServletRequests to tests based on our cloud solr test classes... boom.  The goal with 17503 is to enable folks that integrate with our core code to use jakarta based servlet classes without explosion. If we do additional work to also continue to handle javax that's not a problem so long as Solr stops being the reason people can't upgrade their dependencies that also pull in jetty (or other jakarta based classes). One thing that's already sticky about our situation is our rest api stuff is already on jakarta packages ('jakarta.ws.rs:jakarta.ws.rs-api'). 

So IIUC,  EE10 is simple to satisfy SOLR-17503 too, if we do more ok, but how important is that? is it worth the effort? I have a suspicion that we would involve separate artifacts and/or significant config gymnastics.

iamsanjay:
I picked changes from #2835 i.e. Hadoop-auth module being removed and this one and pushed a new branch to my fork.

https://github.com/iamsanjay/solr/tree/jetty12_ee10

I started working on ee8 because hadoop-auth module does not have any jakarta api support, however If #2835 will be merged then I believe we can transitioned to ee10 without any issue.

epugh:
> I picked changes from #2835 i.e. Hadoop-auth module being removed and this one and pushed a new branch to my fork.
> 
> Branch contains code for Jetty 12 + EE10 (Jakarta servlet api) https://github.com/iamsanjay/solr/tree/jetty12_ee10
> 
> I started working on ee8 because hadoop-auth module does not have any jakarta api support, however If #2835 will be merged then I believe we can transitioned to ee10 without any issue.

I've been sitting on merging the code, for no really good reason..   Thanks for highlighting that it's a bit of a blocker to your progress, I will update the PR and merge it today.

epugh:
> > I picked changes from #2835 i.e. Hadoop-auth module being removed and this one and pushed a new branch to my fork.
> > Branch contains code for Jetty 12 + EE10 (Jakarta servlet api) https://github.com/iamsanjay/solr/tree/jetty12_ee10
> > I started working on ee8 because hadoop-auth module does not have any jakarta api support, however If #2835 will be merged then I believe we can transitioned to ee10 without any issue.
> 
> I've been sitting on merging the code, for no really good reason.. Thanks for highlighting that it's a bit of a blocker to your progress, I will update the PR and merge it today.

Merge is done!

iamsanjay:
Most of the test cases, I believe, are failing due to some issues with the Restore Command. Now in the logs, there is no much detail regarding the issue the only thing I got it.
``` 
{responseHeader={status=0, QTime=611}, STATUS=failed, response=Error handling 'RESTORECORE' action}
```
But If I try to debug I can see more elaborated exception message where I believe for some reason it was suppressed. 
```
Cannot invoke "org.eclipse.jetty.server.Request.getAttribute(String)" because the return value of "org.eclipse.jetty.ee10.servlet.ServletApiRequest.getRequest()" is null
```
There is somewhat related issue opened on Jetty's github https://github.com/jetty/jetty.project/issues/12518

iamsanjay:
Got some of the commits from the main branch. It would be great to remove those before someone would review this PR again.

iamsanjay:
The only failing test is the flaky one, I checked on develocity. 
```
Reproduce with: ./gradlew :solr:modules:ltr:test --tests "org.apache.solr.ltr.TestLTRQParserPlugin.ltr_expensiveFeatureRescoringAndPartialResultsNotTolerated_shouldRaiseException" "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=690CA0CDF5423C59 -Ptests.timeoutSuite=600000! -Ptests.file.encoding=UTF-8
```

solr/core/src/test/org/apache/solr/servlet/HttpSolrCallCloudTest.java
```diff
@@ -93,40 +114,66 @@ private void assertCoreChosen(int numCores, TestRequest testRequest) throws Unav
     assertEquals(numCores, coreNames.size());
   }
 
-  private static class TestResponse extends Response {
+  public static class TestRequest implements HttpServletRequest {
```
dsmiley, line 117: I hate mocks but as I look at this, mocking seems perfect so you don't have to implement everything.  Or maybe there's a base class as there was before that you can extend.  (Same feedback for another source file you did this in)

solr/core/src/test/org/apache/solr/util/tracing/TestHttpServletRequestGetter.java
```diff
@@ -40,8 +49,9 @@ public void test() {
             "b", Set.of("1"),
             "c", Set.of("2"));
 
+    HttpServletRequest httpServletRequestMock = mock(HttpServletRequest.class);
```
dsmiley, line 52: aha; you figured out mocking :-)

solr/cross-dc-manager/src/java/org/apache/solr/crossdc/manager/consumer/Consumer.java
```diff
@@ -21,8 +21,8 @@
 import static org.apache.solr.crossdc.common.KafkaCrossDcConf.ZK_CONNECT_STRING;
 
 import com.codahale.metrics.SharedMetricRegistries;
-import com.codahale.metrics.servlets.MetricsServlet;
-import com.codahale.metrics.servlets.ThreadDumpServlet;
+import io.dropwizard.metrics.servlets.MetricsServlet;
```
dsmiley, line 24: it's suspicious to see we import com.codahale & io.dropwizard.  Maybe okay but I wasn't expecting.

solr/modules/s3-repository/build.gradle
```diff
@@ -20,7 +20,9 @@ apply plugin: 'java-library'
 description = 'S3 Repository'
 
 dependencies {
-  api project(':solr:core')
+  api(project(':solr:core')) {
+    exclude group: 'org.ow2.asm', module: '*'
```
dsmiley, line 24: them solr:core shouldn't be publishing that dependency in the first place. At least see if that can be done.

solr/server/etc/jetty.xml
```diff
@@ -214,14 +212,16 @@
           <Set name="contexts">
             <Ref refid="Contexts" />
           </Set>
+          <!--
```
dsmiley, line 215: appears to simply be some useless metadata we can remove

solr/solrj/src/test/org/apache/solr/client/solrj/embedded/JettyWebappTest.java
```diff
@@ -60,9 +60,9 @@ public void setUp() throws Exception {
 
     server = new Server(port);
     // insecure: only use for tests!!!!
-    server.setSessionIdManager(
-        new DefaultSessionIdManager(server, new Random(random().nextLong())));
-    new WebAppContext(server, path, "/solr");
+    server.addBean(new DefaultSessionIdManager(server, new Random(random().nextLong())));
+    var webAppContext = new WebAppContext(path, "/solr");
```
dsmiley, line 64: minor: inline

solr/test-framework/build.gradle
```diff
@@ -65,16 +65,19 @@ dependencies {
   implementation libs.apache.log4j.api
   implementation libs.apache.log4j.core
   implementation libs.dropwizard.metrics.core
-  implementation libs.dropwizard.metrics.jetty10
+  implementation (libs.dropwizard.metrics.jetty12.ee10){
+    exclude group: 'jakarta.servlet', module: 'jakarta.servlet-api'
```
dsmiley, line 69: as long as we explicitly depend on `libs.jakarta.servlet.api` below (we do), I wonder do we really need to exclude it here?

dsmiley (commented):
Did you read through Jetty's release notes to see if there's stuff we should do that would otherwise not be apparen't if we only fight imports & dependencies?

BTW I have yet to review the last 6 files but GitHub (or Intellij?) is struggling right now...

solr/test-framework/src/java/org/apache/solr/embedded/JettySolrRunner.java
```diff
@@ -414,16 +415,16 @@ public void contextInitialized(ServletContextEvent event) {
 
       // Default servlet as a fall-through
       root.addServlet(Servlet404.class, "/");
-      chain = root;
+      chain.setHandler(root);
```
dsmiley, line 418: it's be clearer to do this at the front of this block.

igiguere (approved):

iamsanjay:
We're hitting a NullPointerException in async code when calling TraceUtils.getSpan(req) after upgrading to Jetty 12. The root cause is that Jetty 12 no longer guarantees that ServletApiRequest.getRequest() will return a non-null value once the request crosses an async boundary (e.g., background threads, ExecutorService, etc.).

In Solr, we set the tracing Span on the HttpServletRequest early in the request lifecycle (in traceHttpRequestExecution2()), and later retrieve it using TraceUtils.getSpan(req). This worked fine in Jetty 9–11, but in Jetty 12, calling getSpan(req) from async code can throw NPEs because the internal Jetty Request is gone.

### Fix

To avoid this, we now cache the span inside HttpSolrCall during its construction, which happens synchronously and before any async processing starts. This ensures we capture the span while the Jetty request is still valid:
```
public HttpSolrCall(HttpServletRequest req, ...) {
  this.span = Optional.ofNullable(TraceUtils.getSpan(req)).orElse(Span.getInvalid());
}
```
All downstream code — including async paths like processAndWait() or CoreAdminHandler — now reads the cached Span from the HttpSolrCall instance rather than trying to access it from the request again.
This avoids the NPE and keeps tracing safe and Jetty 12–compliant without needing to propagate servlet context manually.

(Powered by ChatGPT! 🤖) It would be great if someone else also confirms this theory — hahaha!

solr/test-framework/src/java/org/apache/solr/embedded/JettySolrRunner.java
```diff
@@ -411,6 +410,7 @@ public void contextInitialized(ServletContextEvent event) {
 
       // Default servlet as a fall-through
       root.addServlet(Servlet404.class, "/");
+      chain = root;
```
igiguere: I'm confused. Why initialize 'chain' as Handler.Wrapper only to override that here, and make 'chain' a ServletContextHandler?

solr/test-framework/src/java/org/apache/solr/embedded/JettySolrRunner.java
```diff
@@ -411,6 +410,7 @@ public void contextInitialized(ServletContextEvent event) {
...
     }
 
     chain = injectJettyHandlers(chain);
```
igiguere: ... and now 'chain' becomes what?

igiguere (approved):
Actually, line 416 could be
Handler.Wrapper chain = injectJettyHandlers((Handler.Wrapper) root); // if cast is required
and no need for lines 364 and 413

Or, there's a reason I don't understand for re-assigning the same variable (I just noticed line 425: chain can become a RewriteHandler).

Anyways.  Approved.  If it builds and works at runtime, let's move on...

dsmiley:
> ServletApiRequest.getRequest() will return a non-null value once the request crosses an async boundary (e.g., background threads, ExecutorService, etc.)

This wasn't clear to me so I dug a little further.  So, the standard servlet `HttpServletRequest`, implemented by Jetty, is `ServletApiRequest`... and many of it's methods call an internal `getRequest` that could return null, throwing an NPE.

Wow... that sounds so brittle!  Basically any HttpServletRequest usage across Solr outside of the front door might throw NPE?!  Ouch!  Your fix for the tracing issue makes sense -- +1 to that.  But it seems like there's a bigger risk.  Any Solr code can call `SolrQueryRequest.getHttpSolrCall().getReq()` or `org.apache.solr.request.SolrRequestInfo#getUserPrincipal`.  The second could be early resolved I guess but the getReq might need a big fat warning.

iamsanjay:
Can we remove the `HttpSolrCall` getter from `SolrQueryRequest` along with removing `getReq()` from `HttpSolrCall`?

Few more places.

HttpSolrCall.java
```
public SolrVersion getUserAgentSolrVersion() {
    String header = req.getHeader("User-Agent");
    if (header == null || !header.startsWith("Solr")) {
      return null;
    }
    try {
      return SolrVersion.valueOf(header.substring(header.lastIndexOf(' ') + 1));
    } catch (Exception e) {
      // unexpected but let's not freak out
      assert false : e.toString();
      return null;
    }
  } 
```

CancellableQueryTracker.java

`activeQueriesGenerated.put(queryID, req.getHttpSolrCall().getQueryString());`

dsmiley:
I'll update CancellableQueryTracker.java and HttpSolrCall.getUserAgentSolrVersion within the hour roughly

iamsanjay:
Thanks! But we should avoid propagating the HttpServletRequest further into the codebase? Ideally, it shouldn’t be part of SolrQueryRequest or exposed beyond Jetty once the request has been processed—that's what I was going for with the earlier suggestion.

> Also, can we remove the getHttpSolrCall() method from SolrQueryRequest and drop getReq() from HttpSolrCall as part of that cleanup?

It doesn’t have to be part of this change, but maybe something we can consider in a follow-up.

dsmiley:
I suspect it may be difficult to remove those methods... but maybe.  We've probably done enough for now; tests pass, after all.

Can you point to a bit of Jetty or other documentation somewhere that explains this new behavior of Jetty?  Ideally the WARNING you added would have such a link for the reader to learn more, but I'd like to learn more _now_ too.  It's a really worrisome / troublesome change; pretty surprising IMO.

dsmiley (approved):
Obviously needs a CHANGES.txt, and would go under `Dependency Upgrades` which we have for some versions.

gradle/libs.versions.toml
```diff
@@ -13,8 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 [versions]
-adobe-testing-s3mock = "2.17.0"
```
risdenk, line 16: Does this PR need to be merged with main? I would expect much fewer changes in dependencies outside of Jetty. It looks like there are other lingering changes that might also need to be up to date with main.

gradle/libs.versions.toml
```diff
@@ -80,8 +80,8 @@ decompose = "3.2.2"
 diffplug-spotless = "6.5.2"
 dropwizard-metrics = "4.2.26"
 eclipse-ecj = "3.39.0"
-eclipse-jetty = "10.0.22"
-eclipse-jettytoolchain = "4.0.6"
+eclipse-jetty = "12.0.10"
```
risdenk, line 83: Should probably go to 12.0.19 - https://github.com/jetty/jetty.project/releases/tag/jetty-12.0.19

gradle/libs.versions.toml
risdenk: Looks like this changed back again?

gradle/testing/randomization/policies/solr-tests.policy
```diff
@@ -240,8 +241,15 @@ grant {
   permission java.io.FilePermission "${aws.configFile}", "read,readlink";
   permission java.io.FilePermission "${user.home}${/}.aws${/}-", "read,readlink";
 
+  // GCS
```
risdenk, line 244: Are these changes supposed to be part of Jetty 12?

solr/core/src/java/org/apache/solr/handler/admin/ShowFileRequestHandler.java
risdenk: This should be added as a code comment most likely.

solr/CHANGES.txt
```diff
@@ -119,7 +119,8 @@ Deprecation Removals
 
 Dependency Upgrades
 ---------------------
-(No changes)
+
+* SOLR-16984: Upgraded internal Jetty server from Jetty 10 to Jetty 12, including migration from javax.* to jakarta.* namespaces for Servlet and related APIs. (Sanjay Dutt, David Smiley)
```
risdenk, line 123: I'm not fully convinced other dependencies need to be upgraded, but if this needs to upgrade other dependencies (spring, aws, etc) they should probably be called out as well

solr/modules/s3-repository/build.gradle
```diff
@@ -21,7 +21,7 @@ description = 'S3 Repository'
 
 dependencies {
```
risdenk, line 22: Are all the changes to this s3 repository build.gradle required?

risdenk (commented):
@iamsanjay thanks for the progress on this! I see there are some outstanding comments and I added a few about dependencies being changed.

iamsanjay:
Test case is failing. Not [flaky](https://develocity.apache.org/scans/tests?search.rootProjectNames=solr-root&tests.container=org.apache.solr.cli.PackageToolTest&tests.test=testPackageTool)!

```
./gradlew :solr:core:test --tests "org.apache.solr.cli.PackageToolTest.testPackageTool" "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=B8415E6AAC736FCD -Ptests.timeoutSuite=600000! -Ptests.useSecurityManager=true -Ptests.file.encoding=ISO-8859-1
```

dsmiley:
I spent some time analyzing that test failure for `PackageToolTest.testPackageTool`.  The seed is pertinent but not the other args.  A gradle executed invocation produces the error almost always but strangely inconsistently via IntelliJ test runner (same seed) -- very strange!  A possible timing issue maybe.  I discovered that there's an error that's not being reported by the client (client says null but didn't output the error).  It's the error message generated by ClusterFiles.validate about an encryption signature that doesn't match.  I set a breakpoint on two machines at this point nearby and I see that this branch (showing the problem) gets a buffer length 6733 but a successful run gets length 6734.  At a glance, the data is the same except that the longer payload (that works) has an extra leading byte of decimal 80.  I didn't look into that further (time for me to quit).  I may look again tomorrow night.

epugh:
I was working on something else yesterday afternoon, and I went and manually ran the steps in the bats test https://github.com/apache/solr/blob/main/solr/packaging/test/test_packages.bats#L64 (which is not part of the test suite due to dependency on third party github links) and saw the same error!   I wish I had reported it faster.

dsmiley:
Finally I found the root cause of `PackageToolTest.testPackageTool`!  Lots of debugging today.  It probably doesn't have to do with this PR but the new versions of Jetty seem to tickle it to happening often but not every time.  Not sure if it ever happened before; maybe not if the test has been reliable.  I filed https://issues.apache.org/jira/browse/SOLR-17740 describing the problem; I will work on a fix.  In the mean time, it's okay to add a temporary ignore on that test to unblock this PR, assuming it's only that test which fails. (?)  But it'll hopefully be fixed very soon so we can just wait.

epugh:
Thanks @dsmiley  for some next level debugging.   Does this mean we are really close on this landing?

dsmiley:
This is a huge PR; I'd like to extract from this PR the "HttpServletRequest avoidance" topic to a separate PR/commit (could be same issue) so as to reduce the scope of this PR to be simpler.  I'll do this.

gradle/libs.versions.toml
```diff
@@ -13,8 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 [versions]
-adobe-testing-s3mock = "2.17.0"
-amazon-awssdk = "2.26.19"
+adobe-testing-s3mock = "3.9.1"
+amazon-awssdk = "2.28.11"
```
risdenk, line 17: Does the AWS sdk need to be updated in this MR? Is this just to align with s3mock?

gradle/testing/randomization/policies/solr-tests.policy
risdenk: @iamsanjay is this still needed?

risdenk:
For this PR need to make sure we handle https://issues.apache.org/jira/browse/SOLR-17744 as well. Not sure if anything needs to change with Jetty 12 - @hossman implied there might be a better way with Jetty 12 using the mod itself instead of making one for Jetty 10.

dsmiley:
Looking forward to this landing still... not sure what _really_ remains

iamsanjay:
I added the graceful handler as well. Planning to merge later this week if there are no further reviews.

PR #2881, opened 2024-11-21
SOLR-16427: Enable error-prone ClassInitializationDeadlock rule

malliaridis:
https://issues.apache.org/jira/browse/SOLR-16427

# Description

With the [ClassInitializationDeadlock](https://errorprone.info/bugpattern/ClassInitializationDeadlock) rule error-prone checks for possible class initialization deadlocks. See error-prone documentation for details about the problem this rule is fixing.

# Solution

Following the suggestions of error-prone, the possible deadlocks are resolved by introducing separate classes for static members.

You may review commits individually for a better overview.

This PR is not backwards compatible and public static members are moved to separate classes, potentially breaking third-party code that uses these static members.

# Tests

Tests were not altered, only existing references to static members are updated.

# Checklist


solr/core/src/java/org/apache/solr/search/FunctionQParser.java
```diff
@@ -398,7 +398,7 @@ protected ValueSource parseValueSource(int flags) throws SyntaxError {
     if ((ch >= '0' && ch <= '9') || ch == '.' || ch == '+' || ch == '-') {
       Number num = sp.getNumber();
       if (num instanceof Long) {
-        valueSource = new ValueSourceParser.LongConstValueSource(num.longValue());
+        valueSource = new ValueSourceParsers.LongConstValueSource(num.longValue());
```
dsmiley, line 401: it appears inconsistent with the lower two lines; can you static import this?  Or move to top level if the below ones are (consistency)

solr/core/src/java/org/apache/solr/search/QParserPlugin.java
```diff
@@ -43,57 +32,11 @@ public abstract class QParserPlugin implements NamedListInitializedPlugin, SolrI
...
-    map.put(MinHashQParserPlugin.NAME, new MinHashQParserPlugin());
-    map.put(HashRangeQParserPlugin.NAME, new HashRangeQParserPlugin());
-    map.put(RankQParserPlugin.NAME, new RankQParserPlugin());
-    map.put(KnnQParserPlugin.NAME, new KnnQParserPlugin());
-    map.put(VectorSimilarityQParserPlugin.NAME, new VectorSimilarityQParserPlugin());
-
-    standardPlugins = Collections.unmodifiableMap(map);
-  }
+  @Deprecated(since = "9.8", forRemoval = true)
+  public static final Map<String, QParserPlugin> standardPlugins = QParserPlugins.standardPlugins;
```
dsmiley, line 39: no need to deprecate

solr/core/src/java/org/apache/solr/search/ValueSourceParser.java
```diff
@@ -128,1329 +44,19 @@ public abstract class ValueSourceParser implements NamedListInitializedPlugin {
...
-   * thread safe.
+   * standard functions supported by default
+   *
+   * @deprecated Use {@link ValueSourceParsers#standardValueSourceParsers} instead.
    */
-  private static ValueSourceParser addParser(String name, ValueSourceParser p) {
-    return standardVSParsers.put(name, p);
-  }
+  @Deprecated(since = "9.8", forRemoval = true)
+  public static final Map<String, ValueSourceParser> standardValueSourceParsers =
```
dsmiley, line 53: no need to deprecate

solr/solrj/src/java/org/apache/solr/common/cloud/DocRouter.java
```diff
@@ -38,14 +36,24 @@
  * @lucene.experimental
  */
 public abstract class DocRouter {
-  public static final String DEFAULT_NAME = CompositeIdRouter.NAME;
-  public static final DocRouter DEFAULT;
+  /**
+   * @deprecated Use {@link DocRouters#DEFAULT_NAME} instead.
+   */
+  @Deprecated(since = "9.8", forRemoval = true)
+  public static final String DEFAULT_NAME = DocRouters.DEFAULT_NAME;
```
dsmiley, line 43: WDYT about keeping this as a constant string?  Then almost nobody would have a need to reference the new DocRouters.

solr/solrj/src/java/org/apache/solr/common/MapWriters.java
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.solr.common;
+
+import java.util.Collections;
+
+public class MapWriters {
```
dsmiley, line 22: just for EMPTY feels sad... 
Instead, I could imagine MapWriter.empty() that references an EMPTY constant declared on MapWriterMap

dsmiley (commented):
First, thanks for working on this.  It's been a problem for Lucene & Solr over the years.  @uschindler , you have always been involved in this topic so I'm tagging you for any additional review.

I suggest limiting this PR to be only main, without any concern for deprecating stuff.  By and large, you were deprecating very internal things.

TimeSource has annoyed me; I filed [SOLR-17573](https://issues.apache.org/jira/browse/SOLR-17573) to improve that (albeit may be controversial).

solr/solrj/src/java/org/apache/solr/common/MapWriters.java
malliaridis: Moving `EMPTY` to `MapWriterMap` and providing a static `MapWriter#empty()` method looks like another possible case of ClassInitializationDeadlock. Allthough it is not identified by error-prone, I believe it is similar to the examples. But maybe I am wrong on that?

solr/core/src/java/org/apache/solr/search/FunctionQParser.java
malliaridis: This one is actually interesting. The other two value sources are from Lucene, `LongConstValueSource` is from Solr, which is why this is inconsistent. I could add an import, but perhaps something else could be considered here?

Would `ConstValueSource` be the equivalent in Lucene for `long` values? Because I can only find `ConstValueSource` for `double` in Lucene.

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -1128,9 +1131,9 @@ protected SolrCore(
       initIndex(prev != null, reload);
 
       initWriters();
-      qParserPlugins.init(QParserPlugin.standardPlugins, this);
-      valueSourceParsers.init(ValueSourceParser.standardValueSourceParsers, this);
-      transformerFactories.init(TransformerFactory.defaultFactories, this);
+      qParserPlugins.init(QParserPlugins.standardPlugins, this);
```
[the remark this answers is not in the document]
malliaridis: Some tests outside the core package currently use these registries, including:

- `org.apache.solr.cloud.TestRandomFlRTGCloud` (ref. transformers)
- `org.apache.solr.search.QueryEqualityTest` (ref. value source parsers and query parser plugins)
- `org.apache.solr.search.TestStandardQParsers` (ref. query parser plugins)
- `org.apache.solr.search.SignificantTermsQParserPluginTest` (ref. query parser plugins)

From the logical grouping perspective it makes sense to me to keep them in the current packages. From the visibility perspective (public API) I would prefer to make them package-private / internal.

Perhaps its best to keep it as is for now until we introduce an auto-discovery feature?

solr/solrj/src/java/org/apache/solr/common/cloud/DocRouter.java
malliaridis: Keeping `getDocRouter(routerName)` would probably be another case of class initialization deadlock, so I moved forward and removed the deprecation and kept `DocRouters.DEFAULT_NAME`, so that it aligns with the class reference of `DocRouters.DEFAULT`.

uschindler (approved):
I have not checked all details, but in general the approach to use a static final class with plural s is a workable approach, although it's not backwards compatible. But it is only way for static final constants.

If you have accessor methods, you can use the holder pattern used in Lucene.

malliaridis:
Thanks for your feedback @uschindler 😄

> Please be still careful to not use the plural s classes during initialization on the original class. This would reintroduce the problems.

I believe this is not the case, and I hope error-prone would identify such cases.

> If you have accessor methods, you can use the holder pattern used in Lucene.

I haven't used this pattern before. Would [`TimeSources.get(type)`](https://github.com/apache/solr/blob/b34abd5a5016f125174cd036cd7b3273efe6345a/solr/solrj/src/java/org/apache/solr/common/util/TimeSources.java#L49) (previously `TimeSource.get(type)`) be such a scenario where this pattern could be used?

I guess in cases where the instances are public and directly accessed or added to a list with `static { ... }` the holder pattern cannot be used?

uschindler:
> Thanks for your feedback @uschindler 😄
> 
> > Please be still careful to not use the plural s classes during initialization on the original class. This would reintroduce the problems.
> 
> I believe this is not the case, and I hope error-prone would identify such cases.

OK.

> > If you have accessor methods, you can use the holder pattern used in Lucene.
> 
> I haven't used this pattern before. Would [`TimeSources.get(type)`](https://github.com/apache/solr/blob/b34abd5a5016f125174cd036cd7b3273efe6345a/solr/solrj/src/java/org/apache/solr/common/util/TimeSources.java#L49) (previously `TimeSource.get(type)`) be such a scenario where this pattern could be used?

The pattern can be seen here when codec initialization is done. Lucene has a default Codec, which is a subclass of the Codec class. The Codec class has a method `Codec#getDefault()` which returns the static default. The initialization is delayed by a Holder pattern:

https://github.com/apache/lucene/blob/30d18dfbe1e67bed2d4550b2bc20b9b86ce9781c/lucene/core/src/java/org/apache/lucene/codecs/Codec.java#L40-L59

Here it is called:

https://github.com/apache/lucene/blob/30d18dfbe1e67bed2d4550b2bc20b9b86ce9781c/lucene/core/src/java/org/apache/lucene/codecs/Codec.java#L139-L147

Basically the trick is similar to your's but has more checks to really prevent codec subclasses to get the default during initialization phase (that's why the null checks are there).

> I guess in cases where the instances are public and directly accessed or added to a list with `static { ... }` the holder pattern cannot be used?

No, that's the problem of most classes seen in this PR. It is impossible to use that where the superclass has a static final field to one of the subclasses. You need to break backwards compatibility.

dsmiley:
Looking back at this a bit briefly.  In general these plural form classes should be very unreferenced except when the registry itself is needed (should be rare).  These classes should be considered `@lucene.internal` (we don't have a Solr variation yet) as we pursue ServiceProviderInterface.  Instead of `DocRouters.DEFAULT`, add `DocRouter.getDefault()`.

PR #2884, opened 2024-11-22
SOLR-17569: fix flaky test TestLBHttpSolrClient (9.x) / LBHttp2SolrClientIntegrationTest (main)

jdyer1:
1. Remove "testReliability" because it had no asserts.

2. For "testTwoServers" and "testSimple", wait for the alive check query to be logged after restarting the node.  Previously a blind "Thread.sleep" was in use.

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBSolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBSolrClient.java
@@ -71,6 +71,8 @@ public abstract class LBSolrClient extends SolrClient {
 
   protected static final String UPDATE_LIVE_SERVER_MESSAGE = "Updated alive server list";
 
+  private static final String UPDATE_LIVE_SERVER_LOG = UPDATE_LIVE_SERVER_MESSAGE + ": {}";
+
   // defaults
   protected static final Set<Integer> RETRY_CODES =
       new HashSet<>(Arrays.asList(404, 403, 503, 500));
@@ -420,7 +422,7 @@ public abstract class LBSolrClient extends SolrClient {
     synchronized (aliveServers) {
       aliveServerList = aliveServers.values().toArray(new EndpointWrapper[0]);
       if (log.isDebugEnabled()) {
-        log.debug(UPDATE_LIVE_SERVER_MESSAGE + ": {}", Arrays.toString(aliveServerList));
+        log.debug(UPDATE_LIVE_SERVER_LOG, Arrays.toString(aliveServerList));
       }
     }
   }
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBSolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBSolrClient.java
@@ -20,6 +20,7 @@ package org.apache.solr.client.solrj.impl;
 import static org.apache.solr.common.params.CommonParams.ADMIN_PATHS;
 
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


solr/solrj/src/test/org/apache/solr/client/solrj/impl/LBHttp2SolrClientIntegrationTest.java
```diff
@@ -262,6 +214,31 @@ private LBSolrClient.Endpoint[] bootstrapBaseSolrEndpoints(int max) {
...
 
+  private void startJettyAndWaitForAliveCheckQuery(SolrInstance solrInstance) throws Exception {
+    final int pollIntervalMs = 50;
+    final int timeoutMs = 10000;
+    int waitSoFarMs = 0;
+
+    try (LogListener logListener = LogListener.info().substring("distrib=false")) {
+      solrInstance.startJetty();
+      while (true) {
+        if (logListener.pollMessage() != null) {
```
dsmiley, line 225: If we tweak LogListener to use the BlockingQueue semantics of the underlying BlockingQueue, you could poll with a timeout, and thus wait *just* long enough and not longer.  In other words, expose a new pollMessage(time,timeUnit) method and then no need for this loop & interrupt & sleep.  I could imagine more tests using this technique.

dsmiley (approved):
Thanks for fixing the flakiness

solr/solrj/src/test/org/apache/solr/client/solrj/impl/LBHttp2SolrClientIntegrationTest.java
dsmiley, line 1: friggin beautiful

dsmiley (approved):

solr/solrj/src/test/org/apache/solr/client/solrj/impl/LBHttp2SolrClientIntegrationTest.java
jdyer1: well, only that the more-efficient polling showed that there is a time gap between when the alive check is logged and when the host is actually returned to the alive list.  So actuallyit  was failing for me 50% of the time!  I added a debug-level log message for when the host is added to the list and it passes every time for me now.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBSolrClient.java
```diff
@@ -60,10 +61,18 @@
...
+import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
 public abstract class LBSolrClient extends SolrClient {
 
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  protected static final String UPDATE_LIVE_SERVER_MESSAGE = "Updated alive server list";
+
+  private static final String UPDATE_LIVE_SERVER_LOG = UPDATE_LIVE_SERVER_MESSAGE + ": {}";
```
dsmiley, line 74: is this constant really necessary?

jdyer1: The first one is to prevent the test from breaking should someone alter the log message.  The second is to make precommit happy (no string concatenation is allowed in a log message it seems).

dsmiley: see validate-log-calls.gradle.  You could have added "//nowarn".
Personally I hate over-use of constants; it adds needless indirection to reading code.

PR #2885, opened 2024-11-26
SOLR-17568: SolrCloud shouldn't proxy/route a core specific request

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17568

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -36,7 +36,8 @@ Improvements
 
 Optimizations
 ---------------------
-(No changes)
+* SOLR-17568: The CLI bin/solr export tool now contacts the appropriate nodes directly for data instead of proxying through one.
+ (David Smiley)
 
 Bug Fixes
 ---------------------
@@ -127,6 +128,8 @@ Other Changes
 * SOLR-17285: SolrJ RemoteSolrException moved to SolrClient. (@samuelrivascoding)
 
 * SOLR-17321: Minimum Java version for Apache Solr is now 21, and for SolrJ, it is 17. (Sanjay Dutt, David Smiley)
+
+* SOLR-17568: SolrCloud no longer reroutes/proxies a core request to another node if not found locally. (David Smiley)
 
 ==================  9.8.0 ==================
 New Features
--- a/solr/core/src/java/org/apache/solr/cli/ExportTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/ExportTool.java
@@ -661,8 +661,8 @@ public class ExportTool extends ToolBase {
       }
 
       boolean exportDocsFromCore() throws IOException, SolrServerException {
-
-        try (SolrClient client = CLIUtils.getSolrClient(baseurl, credentials)) {
+        // reference the replica's node URL, not the baseUrl in scope, which could be anywhere
+        try (SolrClient client = CLIUtils.getSolrClient(replica.getBaseUrl(), credentials)) {
           expectedDocs = getDocCount(replica.getCoreName(), client, query);
           QueryRequest request;
           ModifiableSolrParams params = new ModifiableSolrParams();
--- a/solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
+++ b/solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
@@ -278,7 +278,7 @@ public class HttpSolrCall {
         } else {
           // if we couldn't find it locally, look on other nodes
           if (idx > 0) {
-            extractRemotePath(collectionName, origCorename);
+            extractRemotePath(collectionName);
             if (action == REMOTEQUERY) {
               path = path.substring(idx);
               return;
@@ -461,10 +461,10 @@ public class HttpSolrCall {
     }
   }
 
-  protected void extractRemotePath(String collectionName, String origCorename)
+  protected void extractRemotePath(String collectionName)
       throws KeeperException, InterruptedException, SolrException {
     assert core == null;
-    coreUrl = getRemoteCoreUrl(collectionName, origCorename);
+    coreUrl = getRemoteCoreUrl(collectionName);
     // don't proxy for internal update requests
     invalidStates = checkStateVersionsAreValid(queryParams.get(CloudSolrClient.STATE_VERSION));
     if (coreUrl != null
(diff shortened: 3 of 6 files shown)


solr/CHANGES.txt
```diff
@@ -36,7 +36,8 @@ Improvements
 
 Optimizations
 ---------------------
-(No changes)
+* SOLR-17568: The CLI bin/solr export tool now contacts the appropriate nodes directly for data instead of proxying through one.
```
dsmiley, line 39: @epugh you might be interested in this aspect as it touches the CLI.  Or perhaps not since you are working on CLI infrastructure and not the business of what they do.

gbellaton (approved):
I fail to understand why the previous code with even bothering to go through all the collections from the ClusterState.
The new code looks good to me.

solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
epugh, line 1: This is some nice simplification!

solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-10.adoc
```diff
@@ -60,6 +60,11 @@ Users who previously relied on collection-specific URLs to avoid including the c
 
 The service installer now installs a `systemd` startup script instead of an `init.d` startup script. It is up to the user to uninstall any existing `init.d` script when upgrading.
 
+=== SolrCloud request routing
+
+HTTP requests to SolrCloud that are for a specific core must be delivered to the node with that core.
+Past SolrCloud versions would scan all cores everywhere to find a node with the core, and if found then proxy the request.
```
epugh, line 66: Feels like a slightly dangling sentence...   I now understand the "past", now what is happening?   Or was this just a a inefficient bug pattern?

epugh (approved):
LGTM, seems like a gnarly pattern to spot and fix?   One nit about the major changes text.

solr/core/src/java/org/apache/solr/cli/ExportTool.java
```diff
@@ -660,8 +660,8 @@ class CoreHandler {
       }
 
       boolean exportDocsFromCore() throws IOException, SolrServerException {
-
-        try (SolrClient client = CLIUtils.getSolrClient(baseurl, credentials)) {
+        // reference the replica's node URL, not the baseUrl in scope, which could be anywhere
+        try (SolrClient client = CLIUtils.getSolrClient(replica.getBaseUrl(), credentials)) {
```
janhoy, line 664: Since this is a CLI tool, and the CLI tool is potentially being used from a computer outside the cluster itself, I wonder if a side-effect of this might be that it won't work through reverse proxy / ingress with a different external DNS name than what Solr has on the inside.

If I interpret this code correctly, `replica.getBaseUrl()` is data fetched from ZK clusterstate, while `baseUrl` was the user-supplied `--solr-url`. Thus it could be that export tool would currently work through a proxy / ingress due to the auto proxying, but may not work with this change.

solr/CHANGES.txt
dsmiley: Shrug; I found the issue above by seeing a test failure.  There's no test failure relating to what you say.

solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-10.adoc
dsmiley: > now what is happening

Will HTTP 404 if Solr can't resolve the request.  I should add that.  I don't find the last sentence to be dangling but feel free to offer a rewording suggestion.

solr/core/src/java/org/apache/solr/cli/ExportTool.java
dsmiley: It's true that this change would break for such a scenario.  Thankfully the ExportTool is using the Solr/HTTP ClusterStateProvider instead of ZK; thus the URLs may be rewritable by such a proxy thing since it'd be at the HTTP level.  But Solr the client would need to be configured to use XML instead of javabin, but I think it's javabin by default.  I could imagine adding special support for this for CLUSTERSTATE or COLSTATUS

Despite that issue, I still believe in this change.  I don't think it makes sense for SolrCloud to go through such guessing efforts to resolve collection/core ambiguity.  Ideally we'd improve on that so that we wouldn't have ambiguity.  I could imagine a /solr/collectionName!shardName/select syntax or similar.

janhoy: It’s a corner case for sure and not worth modifying until there is a user request. Just wanted to highlight the issue and raise awareness as to whether we want bin/solr to work through reverse proxy.

dsmiley:
Thanks for feedback everyone.  I'll merge in a day or two.  Reminder: this is 10.0 only.

solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
```diff
@@ -1145,40 +1123,32 @@ protected String getRemoteCoreUrl(String collectionName, String origCorename)
...
     Set<String> liveNodes = clusterState.getLiveNodes();
 
-    List<Slice> shuffledSlices;
-    if (slices.size() < 2) {
-      shuffledSlices = slices;
-    } else {
-      shuffledSlices = new ArrayList<>(slices);
-      Collections.shuffle(shuffledSlices, Utils.RANDOM);
-    }
+    Iterator<Slice> shuffledSlices = new RandomIterator<>(Utils.RANDOM, slices);
```
bruno-roustant, line 1144: I was wondering why we need to shuffle if we look for a specific core by its name. But I see that the byCoreName parameter is always false (byCoreName is now always false).
So the code wants to find any active replica of any shard in a specific collection, and then return a url with replica.getBaseUrl() + "/" + origCoreName. Do I understand correctly? Is it the intended behavior?

dsmiley: Thanks for looking closer.  I see there's some dead code to remove -- byCoreName and don't pass origCorename in either.  I'll do another update to clarify.

bruno-roustant (approved):

PR #2893, opened 2024-12-02
SOLR-17579: Refactoring suggested by IDE in ReplicationHandler and tests

epugh:
Typos, some dead methods and dead variables.

https://issues.apache.org/jira/browse/SOLR-17579

# Description

Intellij suggested some fixes.

# Solution

Do them and run tests

# Tests
existing


malliaridis (commented):
I am not sure if it is fine to remove the public static final field `LOCAL_ACTIVITY_DURING_REPLICATION`, but besides that, everything else looks fine.

The assertion was a great finding though.

solr/core/src/java/org/apache/solr/handler/ReplicationHandler.java
```diff
@@ -1532,7 +1518,8 @@ private Long readIntervalNs(String interval) {
   public static final String FETCH_FROM_LEADER = "fetchFromLeader";
 
   // in case of TLOG replica, if leaderVersion = zero, don't do commit
-  // otherwise updates from current tlog won't copied over properly to the new tlog, leading to data
+  // otherwise updates from current tlog won't be copied over properly to the new tlog, leading to
+  // data
   // loss
   // don't commit on leader version zero for PULL replicas as PULL should only get its index
```
dsmiley, line 1524: reflow

solr/CHANGES.txt
```diff
@@ -164,7 +164,8 @@ Dependency Upgrades
 
 Other Changes
 ---------------------
-(No changes)
+* SOLR-17579: Remove unused code and other refactorings in ReplicationHandler and tests.  Removed unused public
```
dsmiley, line 167: honestly, would any reader of CHANGES.txt care?  I'm not questioning your effort in this PR but ...

solr/core/src/java/org/apache/solr/handler/ReplicationHandler.java
epugh: dang it, I trusted `gw tidy` too much..

solr/CHANGES.txt
epugh: i went back and forth on it, since I'd created a JIRA...   I then thought, it should then go in Changes...    Plus removing a public variable...

We do need to add to our dev docs a concise "this is what Changes" is for...

PR #2047, opened 2023-10-29
chore(deps): update org.apache.logging.log4j:* to v2.24.1 - autoclosed

solrbot:
This PR contains the following updates:

| Package | Type | Update | Change |
|---|---|---|---|
| [org.apache.logging.log4j:log4j-web](https://logging.apache.org/log4j/2.x/) ([source](https://redirect.github.com/apache/logging-log4j2)) | dependencies | minor | `2.21.0` -> `2.24.1` |
| [org.apache.logging.log4j:log4j-slf4j2-impl](https://logging.apache.org/log4j/2.x/) ([source](https://redirect.github.com/apache/logging-log4j2)) | dependencies | minor | `2.21.0` -> `2.24.1` |
| [org.apache.logging.log4j:log4j-layout-template-json](https://logging.apache.org/log4j/2.x/) ([source](https://redirect.github.com/apache/logging-log4j2)) | dependencies | minor | `2.21.0` -> `2.24.1` |
| [org.apache.logging.log4j:log4j-core](https://logging.apache.org/log4j/2.x/) ([source](https://redirect.github.com/apache/logging-log4j2)) | dependencies | minor | `2.21.0` -> `2.24.1` |
| [org.apache.logging.log4j:log4j-api](https://logging.apache.org/log4j/2.x/) ([source](https://redirect.github.com/apache/logging-log4j2)) | dependencies | minor | `2.21.0` -> `2.24.1` |
| [org.apache.logging.log4j:log4j-1.2-api](https://logging.apache.org/log4j/2.x/) ([source](https://redirect.github.com/apache/logging-log4j2)) | dependencies | minor | `2.21.0` -> `2.24.1` |

---

### Configuration

📅 **Schedule**: Branch creation - "* * * * *" (UTC), Automerge - At any time (no schedule defined).

🚦 **Automerge**: Disabled by config. Please merge this manually once you are satisfied.

♻ **Rebasing**: Whenever PR becomes conflicted, or you tick the rebase/retry checkbox.

🔕 **Ignore**: Close this PR and you won't be reminded about these updates again.

---

 - [ ] If you want to rebase/retry this PR, check this box

---

This PR has been generated by [Renovate Bot](https://redirect.github.com/solrbot/renovate-github-action)

PR #2895, opened 2024-12-06
upgrade-log4j2-2.24.2

risdenk:

This is a follow on to https://github.com/apache/solr/pull/2047 where `biz.aQute.bnd` doesn't need to be upgraded.

However running into some test failures due to concurrent modification exceptions when updating loggers in some tests.


risdenk:
Looks like we are running into:
* https://github.com/apache/logging-log4j2/issues/3234
* https://github.com/apache/logging-log4j2/issues/3283

Once those are fixed (there is a PR it looks like) should be ok

malliaridis (commented):
Thanks for taking care of dependency updates during the downtime of our bot. 😄 

Could you update the license files of
- [slf4j-api](https://github.com/qos-ch/slf4j/blob/v_2.0.16/slf4j-api/LICENSE.txt)
- [jul-to-slf4j](https://github.com/qos-ch/slf4j/blob/v_2.0.16/jul-to-slf4j/LICENSE.txt)
- jcl-over-slf4j, this changed from MIT to ASL in [repo](https://github.com/qos-ch/slf4j/blob/v_2.0.16/jcl-over-slf4j/LICENSE.txt), but is still using the [base license](https://github.com/qos-ch/slf4j/blob/v_2.0.16/LICENSE.txt) in downloaded package. Not sure which one to use in this case. @janhoy any idea how to handle these cases? Prefer ASL?

Also you can remove the content of:
- slf4j-NOTICE.txt (as it does not exist anymore according to repo)
- jul-to-slf4j-NOTICE.txt (as it does not exist anymore according to repo)
- jcl-over-slf4j-NOTICE.txt (as it does not exist anymore according to repo)

And since you are already updating the libraries, you may consider merging the log4j-*-LICENSE.txt (6 files) into log4j-LICENSE.txt and log4j-*-NOTICE.txt (6 files) into log4j-NOTICE.txt, as they all share now the same license and notice file according to repo (see [log4j](https://github.com/apache/logging-log4j2/tree/rel/2.24.2)).

I think our license file checker does not recognize the slf4j group for the jul-to-slf4j and jcl-over-slf4j libraries, otherwise we could merge these as well.

janhoy:
I have no preference when license is ambiguous. Guess whatever is in the download tar for that version is safe.

ppkarwasz:
> * [jul-to-slf4j](https://github.com/qos-ch/slf4j/blob/v_2.0.16/jul-to-slf4j/LICENSE.txt)

Since you use Log4j Core in your binary distribution, you should consider using the [JUL-to-Log4j API bridge](https://logging.apache.org/log4j/2.x/log4j-jul.html) instead. The artifact has two _modi operandi_:

* A fast bridge, which replace JUL entirely, but requires the setting of the `java.util.logging.manager` Java system property in your startup scripts (see [Using LogManager](https://logging.apache.org/log4j/2.x/log4j-jul.html#bridge-logmanager)).
* A slower bridge, similar to the one in `jul-to-slf4j`, but it provides direct bridging between JUL and Log4j API (without passing through SLF4J). See [Using Log4jBridgeHandler](https://logging.apache.org/log4j/2.x/log4j-jul.html#bridge-handler).

> * jcl-over-slf4j, this changed from MIT to ASL in [repo](https://github.com/qos-ch/slf4j/blob/v_2.0.16/jcl-over-slf4j/LICENSE.txt), but is still using the [base license](https://github.com/qos-ch/slf4j/blob/v_2.0.16/LICENSE.txt) in downloaded package. Not sure which one to use in this case. @janhoy any idea how to handle these cases? Prefer ASL?

Since the release of `commons-logging` 1.3.0, `jcl-over-slf4j` (as well as `spring-jcl` and `log4j-jcl`) are no longer necessary. We made [a blog post about it](https://logging.apache.org/blog/2023/12/02/apache-common-logging-1.3.0.html) last year.

risdenk:
Thanks @malliaridis will fix in new branch for 2.24.3 and @ppkarwasz for the release w/ the fix

risdenk:
Closing in favor of https://github.com/apache/solr/pull/3603

PR #2899, opened 2024-12-09
SOLR-17541: LBSolrClient implementations should agree on 'getClient()' semantics

jdyer1:
With this PR makes `LBHttp2SolrClient` maintains an instance of `HttpSolrClientBuilderBase` per "Base Url".  This makes the semantics of `LBHttp2SolrClient#getClient` consistent with that of the older `LBHttpSolrClient`.

Behavior changes:
- `LBHttp2SolrClient` generates a Http Solr Client per base url
    - (1) at constructon
    - (2) whenever a previously-unseen base url is encountered
    - NOTE: The Map holding the clients may grow unbounded; there is no cleanup short of callng `close()`.
- `LBHttp2SolrClient` mutates the `baseSolrUrl` variable of the caller-supplied instance of `HttpSolrClientBuilderBase` whenever it creates a new Http Solr Client.
- Both `LBHttp2SolrClient` and `CloudHttp2SolrClient` always own the internal/delegate clients.  They are now always closed by us on `close()`.

The following class definitions are changed in the SolrJ public API:
- `LBHttp2SolrClient`
    - from: `LBHttp2SolrClient<C extends HttpSolrClientBase> extends LBSolrClient`
    - to: `LBHttp2SolrClient<B extends HttpSolrClientBuilderBase<?, ?>> extends LBSolrClient`
- `LBHttp2SolrClient.Builder`
    - from: `Builder<C extends HttpSolrClientBase>`
    - to: `Builder<B extends HttpSolrClientBuilderBase<?, ?>>`
- All methods on `LBHttp2SolrClient.Builder` return a `Builder<B>` instead of a `Builder<C>`

The following are removed from the SolrJ public API:
- method `CloudHttp2SolrClient.Builder#withHttpClient`
- constructor `LBHttp2SolrClient.Builder(C solrClient, Endpoint... endpoints)`
    - replaced with: `public Builder(B solrClientBuilder, Endpoint... endpoints)`
- method `LBHttp2SolrClient.Builder#withListenerFactory` 
    - renamed to `LBHttp2SolrClient.Builder#withListenerFactories` (replace any existing)
    - add  `LBHttp2SolrClient.Builder#addListenerFactory` (add to existing)
- method `HttpJdkSolrClient#requestWithBaseUrl` (main only, never released)

The following are marked `@lucene.experimental`
- method `Http2SolrClient#requestWithBaseUrl` (main only, never released)
- interface `SolrClientFunction` (main only, never released)


solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java
```diff
@@ -94,35 +97,74 @@
...
+  private synchronized HttpSolrClientBase buildClient(Endpoint endpoint) {
+    var client = urlToClient.get(endpoint.toString());
+    if (client == null) {
+      String tmpBaseSolrUrl = solrClientBuilder.baseSolrUrl;
+      solrClientBuilder.baseSolrUrl = endpoint.getBaseUrl();
+      client = solrClientBuilder.build();
+      urlToClient.put(endpoint.getBaseUrl(), client);
+      solrClientBuilder.baseSolrUrl = tmpBaseSolrUrl;
+    }
+    return client;
```
dsmiley, line 136: Prefer calling ConcurrentHashMap.computeIfAbsent or similar to get-then-put because of it's nice atomicity properties, and avoids synchronization needs

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java
```diff
@@ -94,35 +97,74 @@
...
     this.defaultCollection = builder.defaultCollection;
+
+    if (builder.solrClientBuilder.urlParamNames == null) {
+      this.urlParamNames = Collections.emptySet();
+    } else {
+      this.urlParamNames = Collections.unmodifiableSet(builder.solrClientBuilder.urlParamNames);
+    }
+  }
+
+  private synchronized HttpSolrClientBase buildClient(Endpoint endpoint) {
```
dsmiley, line 127: synchronized is guarding what here?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java
```diff
@@ -94,35 +97,74 @@
...
+      buildClient(endpoint);
+    }
+
     this.aliveCheckIntervalMillis = builder.aliveCheckIntervalMillis;
     this.defaultCollection = builder.defaultCollection;
+
+    if (builder.solrClientBuilder.urlParamNames == null) {
+      this.urlParamNames = Collections.emptySet();
+    } else {
+      this.urlParamNames = Collections.unmodifiableSet(builder.solrClientBuilder.urlParamNames);
```
dsmiley, line 123: probably fine but I'd prefer Set.copyOf here

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java
```diff
@@ -94,35 +97,74 @@
...
-    return solrClient.getUrlParamNames();
+    return urlParamNames;
+  }
+
+  @Override
+  public void close() {
+    super.close();
+    for (HttpSolrClientBase client : urlToClient.values()) {
+      IOUtils.closeQuietly(client);
+    }
```
dsmiley, line 167: could be a one-liner with `urlToClient.values().forEach(IOUtils::closeQuiety)`

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java
jdyer1: The `synchronized` keyword was to prevent multiple threads from adding clients for the same Base Url.  However, in taking your advice to use `computeIfAbsent` this removes the need.

jdyer1:
As a blocker for this PR, I am getting **many** seemingly-unrelated solr-core test failures.  For instance, `PostToolTest#testBasicRun`.  

Maybe this is similar to the problem I describe in #2242, in that by sharing resources between client and server in a unit test, we get automatic PKI authentication, on which the tests depend.  If this speculation is true, this would be *merely* a test bug.  

Otherwise, perhaps solr-core needs to be able to share instances of `Http2SolrClient` across its internals?  The major change here being callers can only pass a `Builder` to `LbHttp2SolrClient` and `CloudHttp2Solrclient` and not a pre-built delegate client.  

In either case I do not know how to continue to make progress so any advice is appreciated.

dsmiley:
The problem is that org.apache.solr.handler.component.HttpShardHandlerFactory#setSecurityBuilder is being called but it only affects the built client, not clients built separately (in LB).  My suspicion is that your conversion of passing the Builder as a template instead of an existing client should switch back to the client as a template and then remember to call `org.apache.solr.client.solrj.impl.Http2SolrClient.Builder#withHttpClient` from the template.

jdyer1:
>  your conversion of passing the Builder as a template instead of an existing client should switch back to the client as a template

I am not sure this ticket/PR has much value if we continue having the caller pass in a pre-built delegate client instead of a Builder.  `LBHttp2solrClient` needs to create a client per base url, but the caller only supplies one.  Yet `LBHttp2SolrClient`might not know how to correctly clone that client, keeping in mind `LBHttp2solrClient` is now generic. It seems to me that passing in the pre-built client is more of a relic of the days before we used Bulders in solrJ, and not a really great API decision.  I was hoping to improve that here.

dsmiley:
Okay with me to stay with the builder.  Probably add a new method to org.apache.solr.security.HttpClientBuilderPlugin for the builder, overloading setup.

As an aside, I'm totally confused about our authentication plugins, especially `PKIAuthenticationPlugin` as it apparently has a pivotal role, even when there's know "PKI" going on!  CC @janhoy I suspect you might know

gerlowskija:
This LGTM, pending figuring out the test-failure issue and adding a CHANGES.txt entry that highlights the deprecations/removals.

> The following were not removed, but perhaps should be considered to be marked "internal" or "Experimental"

+1 - Both seems like good candidates for a "lucene.experimental" annotation.  (Presumably there should be a "solr.experimental", but I don't think that exists...)

jdyer1:
>  Probably add a new method to org.apache.solr.security.HttpClientBuilderPlugin for the builder,

Thank you for the insight on this. I went ahead and did this, but its a bit ugly.  I'd rather not have both, but then again, I want to minimize changes to solr-core with this PR.

Besides, @dsmiley, I appreciate the change to synchronize on the Builder.  That was a pretty bad miss on my part!

I also added `@lucene.experimental` to the two places discussed.

solr/core/src/java/org/apache/solr/core/HttpSolrClientProvider.java
dsmiley, line 1: WDYT @iamsanjay ?

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandlerFactory.java
```diff
@@ -305,16 +306,16 @@ public void init(PluginInfo info) {
             sb);
     int soTimeout =
         getParameter(args, HttpClientUtil.PROP_SO_TIMEOUT, HttpClientUtil.DEFAULT_SO_TIMEOUT, sb);
-
-    this.defaultClient =
+    this.httpClientBuilder =
```
dsmiley, line 309: please name `httpSolrClientBuilder` -- so as to clarify this isn't an Apache/Jetty HttpClient; it's a SolrClient.

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandlerFactory.java
```diff
@@ -305,16 +306,16 @@ public void init(PluginInfo info) {
...
         new Http2SolrClient.Builder()
             .withConnectionTimeout(connectionTimeout, TimeUnit.MILLISECONDS)
             .withIdleTimeout(soTimeout, TimeUnit.MILLISECONDS)
             .withExecutor(commExecutor)
             .withMaxConnectionsPerHost(maxConnectionsPerHost)
-            .build();
-    this.defaultClient.addListenerFactory(this.httpListenerFactory);
-    this.loadbalancer = new LBHttp2SolrClient.Builder<Http2SolrClient>(defaultClient).build();
+            .withListenerFactory(List.of(this.httpListenerFactory));
```
dsmiley, line 315: I noticed an issue.  This overwrites the list of listeners that may already be present.  Shouldn't we be augmenting, not replacing?

janhoy:
> `PKIAuthenticationPlugin` as it apparently has a pivotal role, even when there's know "PKI" going on!

PKI plugin handles incoming internal requests from other nodes in the cluster. Then there is indeed PKI going on 😉 It is registered whenever any auth plugin is active.

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandlerFactory.java
jdyer1: I was trying to work within the bounds of the existing API, but I agree this should be an add, not a replace.  I modified the API, and broke backwards-compatibility so as to not silently introduce bugs in users' applications.

jdyer1:
I am curious to know opinions as to how to handle deprecations here.  This PR introduces several API changes (see description) but most of the replacement APIs will never exist in any `9.x` version.  We can add a @Deprecated annotation to `CloudHttp2SolrClient.Builder#withHttpClient`, with a suggestion to use `#withInternalClientBuilder` instead.  Beyond that, is there anything else we can/should do in a theoretical `Solr 9.9` to warn users of the upcoming changes?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -932,8 +931,27 @@ public Builder(String baseSolrUrl) {
       this.baseSolrUrl = baseSolrUrl;
     }
 
-    public Http2SolrClient.Builder withListenerFactory(List<HttpListenerFactory> listenerFactory) {
-      this.listenerFactory = listenerFactory;
+    /**
+     * specify a listener factory, which will be appened to any existing values.
```
dsmiley, line 935:
```suggestion
     * Specify a listener factory, which will be appended to any existing values.
```

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java
```diff
@@ -94,35 +97,63 @@
  *
  * @since solr 8.0
  */
-public class LBHttp2SolrClient<C extends HttpSolrClientBase> extends LBSolrClient {
+public class LBHttp2SolrClient<B extends HttpSolrClientBuilderBase<?, ?>> extends LBSolrClient {
```
dsmiley, line 100: an aside -- I wonder if the use of generics here is more annoying than helpful.  We aren't compelled to use generics.

dsmiley (commented):
So this would be Solr 10?

dsmiley:
RE deprecations:  A good call out but I don't think we should worry much about deprecations if we can't provide the alternative.  At least not for LBSolrClient which is rather internal, I think.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudHttp2SolrClient.java
```diff
@@ -404,23 +400,6 @@ public Builder withCollectionCacheTtl(long timeToLive, TimeUnit unit) {
...
 
-    /**
-     * Set the internal http client.
-     *
-     * <p>Note: closing the httpClient instance is at the responsibility of the caller.
-     *
-     * @param httpClient http client
-     * @return this
-     */
-    public Builder withHttpClient(Http2SolrClient httpClient) {
```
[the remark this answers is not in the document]
jdyer1: This is the most controversial part of this PR, and I do not want to merge a potentially disruptive API change like this without buy-in.  I do not think this is much of a burden though, ecause the Builder has `withInternalClientBuilder`, which users should be able to easily migrate to use.  

The purpose of forcing this is that `CloudHttp2SolrClient` in turn passes this Client(Builder) to `LBHttp2SolrClient`.  But as noted before,  `LBHttp2SolrClient` doesn't really use this client either; it clones it using `Http2SolrClient.builder#withHttpClient`.  Yet we've made  `LBHttp2SolrClient` generic so this means that any subclass of `HttpSolrClientBase` needs to also have a way to be cloned.  Personally I think clone methods like this are brittle and a maintenance burden, and I don't want to need to write one for `HttpJdkSolrClient`.  It also seems like a better API:  by telling the user we want a Builder, we signal that we will create Clients as needed.  Passing in an already-built Client signals that we will use that particular one, which is not always the case here.

solr/core/src/java/org/apache/solr/security/HttpClientBuilderPlugin.java
```diff
@@ -34,4 +34,8 @@ public interface HttpClientBuilderPlugin {
   public SolrHttpClientBuilder getHttpClientBuilder(SolrHttpClientBuilder builder);
 
   public default void setup(Http2SolrClient client) {}
+
+  public default void setup(Http2SolrClient.Builder httpClientBuilder, Http2SolrClient client) {
```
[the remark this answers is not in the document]
jdyer1: Yes, I agree separate methods would be nicer.  But the implementing classes do things on `setup`  that should only be done once.  I settled on this ugly API with hopes this is the least-trappy thing to leave for future developers.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java
jdyer1: I worry too I have gone down the wrong road with this.  On the other hand, I do not want consider this with this PR.

jdyer1:
> So this would be Solr 10?

Yes.  When I read the description on this ticket, it seemed to me a good opportunity to improve our API with the upcoming major release.  It seems wrong we clone the passed-in client.  This is especially important if we want to eventually have merely one LB Client, one Cloud Client and one Cluster State Provider that can take any subclass of `HttpSolrClientBase`.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudHttp2SolrClient.java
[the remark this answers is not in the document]
dsmiley: > it clones it using Http2SolrClient.builder#withHttpClient

which still exists, so if a user somehow has a client but not a builder but needs one due to the API change you introduce, they have this option.

> by telling the user we want a Builder, we signal that we will create Clients as needed

Well written James!

+1

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttp2SolrClient.java
dsmiley: in another PR then.

solr/core/src/java/org/apache/solr/security/PKIAuthenticationPlugin.java
```diff
@@ -374,8 +374,17 @@ PublicKey fetchPublicKeyFromRemote(String nodename) {
...
+  public void setup(Http2SolrClient.Builder httpClientBuilder, Http2SolrClient client) {
+    setup(client, httpClientBuilder);
+  }
+
   @Override
   public void setup(Http2SolrClient client) {
+    setup(client, null);
+  }
+
+  private void setup(Http2SolrClient client, Http2SolrClient.Builder builder) {
```
dsmiley, line 387: why have this separate and with different parameter order?  In other words, lets make *this* one the `@Override` with builder then client.  Remove the 4 lines earlier on 377.

solr/core/src/java/org/apache/solr/security/HttpClientBuilderPlugin.java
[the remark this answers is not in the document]
dsmiley: Can you point me at such so I can appreciate the problem better?

jdyer1:
> Can you point me at such so I can appreciate the problem better?

My apologies, I think I misstated what is going on earlier.  `PKIAuthenticationPlugin` on setup injects a `RequestResponseListener` into the passed-in Client.  I wanted to ensure the same Listener gets injected both in the Builder and the Client, in case it matters.

dsmiley:
Since the direction is SolrClient immutability, I think this auth/PKI setup method should only the builder (mutable).  Is this realistic?

jdyer1:
> Since the direction is SolrClient immutability, I think this auth/PKI setup method should only the builder (mutable). Is this realistic?

I fully agree this is the right direction to take.  Unfortunately this probably requires a minor redesign to both `UpdateShardHandler` and `HttpSolrClientProvider`, which is out-of-scope here.

solr/core/src/java/org/apache/solr/security/PKIAuthenticationPlugin.java
```diff
@@ -374,17 +374,13 @@ PublicKey fetchPublicKeyFromRemote(String nodename) {
...
-
   @Override
   public void setup(Http2SolrClient client) {
-    setup(client, null);
+    setup(null, client);
   }
 
-  private void setup(Http2SolrClient client, Http2SolrClient.Builder builder) {
+  @Override
+  public void setup(Http2SolrClient.Builder httpClientBuilder, Http2SolrClient client) {
```
dsmiley, line 383: Suggest naming this param "builder" as it was, or "httpSolrClientBuilder".  Note "client" is simply "client".

dsmiley (commented):
Okay then let's leave a TODO comment to communicate this direction that's out of scope.  Maybe even create a JIRA issue and refer to it in the TODO.  it'd be good if someone addressed that for Solr 10.

Just some nitpicks on comments / naming here but I'm +1 to merge otherwise.

dsmiley (commented):
CHANGES.txt will be interesting here; a number of things happened.

jdyer1:
I added a CHANGES.txt entry to this PR and plan to merge this early next week.  I also will separately add a deprecation on `CloudHttp2SolrClient#withHttpClient` in `branch_9x`, to be released with Solr 9.9.

dsmiley (approved):

PR #2901, opened 2024-12-10
SOLR-16390: fix flaky test ClusterPropsAPITest.testClusterPropertyOpsAllGood

psalagnac:
https://issues.apache.org/jira/browse/SOLR-16390
(linking to the Jira that introduced the test)

# Description

The test fails when SSL is enabled, because of testing framework that set 'urlScheme' cluster property.

I was looking at the failure in another PR, and it seems this test fails quite often since #2788. I don't think it is tracked somewhere else.

# Solution

 This change updates the test to ignore existing cluster properties.

# Tests

n/a

### DIFF
--- a/solr/core/src/test/org/apache/solr/handler/admin/api/ClusterPropsAPITest.java
+++ b/solr/core/src/test/org/apache/solr/handler/admin/api/ClusterPropsAPITest.java
@@ -18,6 +18,8 @@
 package org.apache.solr.handler.admin.api;
 
 import static org.apache.solr.common.util.Utils.getObjectByPath;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.not;
 
 import java.net.URL;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


epugh (approved):
LGTM.

Nice find, and it seems a bit of a shame that we have to have that logic there ;-).

solr/core/src/test/org/apache/solr/handler/admin/api/ClusterPropsAPITest.java
```diff
@@ -103,10 +109,9 @@ public void testClusterPropertyOpsAllGood() throws Exception {
       // List Properties, this time there should be 1
       o = Utils.executeGET(client.getHttpClient(), baseUrlV2ClusterProps, Utils.JSONCONSUMER);
       assertNotNull(o);
-      assertEquals(1, ((List<?>) getObjectByPath(o, true, "clusterProperties")).size());
-      assertEquals(
-          testClusterProperty,
-          (String) ((List<?>) getObjectByPath(o, true, "clusterProperties")).get(0));
+      assertEquals(initCount + 1, ((List<?>) getObjectByPath(o, true, "clusterProperties")).size());
```
dsmiley, line 112: IMO this test shouldn't bother checking the size; it need only check that the entry it's expecting is there or is not there.  Hamcrest test utils helps.  I did this a day or two ago in this test: https://github.com/apache/solr/blob/2d3a8d9954a45c0eb1bbaa77b29f51ce314c1849/solr/solrj/src/test/org/apache/solr/client/solrj/impl/ClusterStateProviderTest.java#L165

dsmiley (approved):

solr/core/src/test/org/apache/solr/handler/admin/api/ClusterPropsAPITest.java
psalagnac: Good point. I'm updating the test to check the property exists instead of counting.

epugh: much nicer!

dsmiley (approved):

dsmiley:
FYI @anshumg I pushed this fix to branch_9_8 as a number of builds fail without this fix.  It's been on branch_9x since December 10th.

anshumg:
Thanks @dsmiley

PR #2902, opened 2024-12-10
SOLR-17587: Prometheus Writer duplicate TYPE information in exposition format

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-17587

# Description

Solr's Prometheus writer duplicates `# TYPE <metric name> <prometheus metric type>` in it's exposition format for `core`registry metrics. 

This is an illegal format and depending on the technologies prometheus exposition verification for example `Telegraf`, this will fail. For Prometheus server itself, this still passes and collects the metrics just fine for some reason.

This is because the Prometheus Writer takes Dropwizard registries and exports them to Prometheus Registries to expose them in Prometheus format. Solr creates Dropwizard registry for every `core` and differentiates the metrics that way even though they have the same metric names.

For prometheus, this creates an issue in that metrics should be differentiated in it's attributes and tags. So when the metrics are output with the Prometheus response writer, it duplicates the `TYPE` information because it is a registry for every `core` and doesn't know that the other `core` registries have the same metric name and results in duplicate `TYPE` information.

# Solution

When metrics are going to be exported for prometheus, we merge all the `core` Dropwizard metric registries into a single registry and export that registry into prometheus. Duplicate metric names in a registry is not allowed in prometheus, so we will also append the core name to the Dropwizard metric to differentiate which metric belongs to what core and parse the labels accordingly.

This also allowed to clean up and simply some of the `SolrPrometheusCoreFormatter` code.

# Tests

Updated the test accordingly with the coreName existing in the Dropwizard metric names and it's parsing. 

Also added an assert in `testPrometheusStructureOutput` to confirm there is no duplicate `TYPE` information in prometheus output.

# Checklist


mlbiscoc:
Going to bump this PR. It's not a blocker for standard Prometheus server metrics collection but it can potentially block users using other exporters/collectors.

dsmiley (commented):
Correct me if I'm wrong, but had you kept the test that you had removed because hossman didn't like it (it was imperfect), a reviewer (like me but you too) would be able to see in this PR a diff against the output format to understand the impact (or lack of impact).

dsmiley:
I could say the same thing for dependencies.  People add/remove dependencies in build files (Maven/Gradle/Ivy) but ultimately what we want to know is, what JARs changed in the ultimate distribution.

mlbiscoc:
> Correct me if I'm wrong, but had you kept the test that you had removed because hossman didn't like it (it was imperfect), a reviewer (like me but you too) would be able to see in this PR a diff against the output format to understand the impact (or lack of impact).

In theory, yes you would see the difference that is happening here. But since you can't see the output, here is a basically the change to the output:

Before:
```
curl 'localhost:8983/solr/admin/metrics?wt=prometheus' | grep "#" | sort
# TYPE solr_metrics_core_average_request_time gauge
# TYPE solr_metrics_core_average_request_time gauge
# TYPE solr_metrics_core_average_searcher_warmup_time gauge
# TYPE solr_metrics_core_average_searcher_warmup_time gauge
# TYPE solr_metrics_core_cache gauge
# TYPE solr_metrics_core_cache gauge
# TYPE solr_metrics_core_highlighter_requests_total counter
# TYPE solr_metrics_core_highlighter_requests_total counter
# TYPE solr_metrics_core_index_size_bytes gauge
# TYPE solr_metrics_core_index_size_bytes gauge
# TYPE solr_metrics_core_requests_time_total counter
# TYPE solr_metrics_core_requests_time_total counter
# TYPE solr_metrics_core_requests_total counter
# TYPE solr_metrics_core_requests_total counter
# TYPE solr_metrics_core_searcher_documents gauge
# TYPE solr_metrics_core_searcher_documents gauge
# TYPE solr_metrics_core_tlog_total counter
# TYPE solr_metrics_core_tlog_total counter
# TYPE solr_metrics_core_update_handler gauge
# TYPE solr_metrics_core_update_handler gauge
# TYPE solr_metrics_jetty_dispatches_total counter
# TYPE solr_metrics_jetty_requests_total counter
# TYPE solr_metrics_jetty_response_total counter
# TYPE solr_metrics_jvm_buffers gauge
# TYPE solr_metrics_jvm_buffers_bytes gauge
# TYPE solr_metrics_jvm_gc gauge
# TYPE solr_metrics_jvm_gc_seconds gauge
# TYPE solr_metrics_jvm_heap gauge
# TYPE solr_metrics_jvm_memory_pools_bytes gauge
# TYPE solr_metrics_jvm_threads gauge
# TYPE solr_metrics_node_connections gauge
# TYPE solr_metrics_node_core_root_fs_bytes gauge
# TYPE solr_metrics_node_cores gauge
# TYPE solr_metrics_node_requests_time_total counter
# TYPE solr_metrics_node_requests_total counter
# TYPE solr_metrics_node_thread_pool_total counter
# TYPE solr_metrics_os gauge
```

After:

```
curl 'localhost:8983/solr/admin/metrics?wt=prometheus' | grep "#" | sort
# TYPE solr_metrics_core_average_request_time gauge
# TYPE solr_metrics_core_average_searcher_warmup_time gauge
# TYPE solr_metrics_core_cache gauge
# TYPE solr_metrics_core_highlighter_requests_total counter
# TYPE solr_metrics_core_index_size_bytes gauge
# TYPE solr_metrics_core_requests_time_total counter
# TYPE solr_metrics_core_requests_total counter
# TYPE solr_metrics_core_searcher_documents gauge
# TYPE solr_metrics_core_tlog_total counter
# TYPE solr_metrics_core_update_handler gauge
# TYPE solr_metrics_jetty_dispatches_total counter
# TYPE solr_metrics_jetty_requests_total counter
# TYPE solr_metrics_jetty_response_total counter
# TYPE solr_metrics_jvm_buffers gauge
# TYPE solr_metrics_jvm_buffers_bytes gauge
# TYPE solr_metrics_jvm_gc gauge
# TYPE solr_metrics_jvm_gc_seconds gauge
# TYPE solr_metrics_jvm_heap gauge
# TYPE solr_metrics_jvm_memory_pools_bytes gauge
# TYPE solr_metrics_jvm_threads gauge
# TYPE solr_metrics_node_connections gauge
# TYPE solr_metrics_node_core_root_fs_bytes gauge
# TYPE solr_metrics_node_cores gauge
# TYPE solr_metrics_node_requests_time_total counter
# TYPE solr_metrics_node_requests_total counter
# TYPE solr_metrics_node_thread_pool_total counter
# TYPE solr_metrics_os gauge
```

`#` is just a comment in Prometheus but `# TYPE` is special. Some technologies with verify there are dupes while others don't. This just removes the duplicates and merged the metrics under a single `# TYPE`

solr/core/src/java/org/apache/solr/handler/admin/MetricsHandler.java
```diff
@@ -186,10 +187,22 @@ private NamedList<Object> handlePrometheusExport(SolrParams params) {
...
+    MetricRegistry mergedCoreRegistries = new MetricRegistry();
 
     for (String registryName : requestedRegistries) {
       MetricRegistry dropwizardRegistry = metricManager.registry(registryName);
+
+      // Merge all core registries into a single registry and
+      // append the core name to the metric to avoid duplicate metrics name
+      if (registryName.startsWith("solr.core")) {
+        mergedCoreRegistries.registerAll(
+            Arrays.stream(registryName.split("\\.")).skip(1).collect(Collectors.joining("_")),
```
dsmiley, line 201: It'd be faster to do a substring on the index after the first period, and then with the result of that, replace all periods with underscores.  Having a little utility method to do this with comment would ultimately be clearer IMO

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreCacheMetric.java
```diff
@@ -38,7 +37,7 @@ public SolrCoreCacheMetric(
   public SolrCoreMetric parseLabels() {
     String[] parsedMetric = metricName.split("\\.");
     if (dropwizardMetric instanceof Gauge) {
-      labels.put("cacheType", parsedMetric[2]);
+      labels.put("cacheType", parsedMetric[3]);
```
dsmiley, line 40: Previously index 2 lined up with the sample above on line 33.  But now; I dunno.  I suppose your data example string is now out of date.
Same comment as other source files you edited similarly.

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreMetric.java
```diff
@@ -25,28 +26,26 @@
...
-      labels.put("shard", m.group(2));
-      labels.put("replica", m.group(3));
+    Matcher cloudPattern = CLOUD_CORE_PATTERN.matcher(metricName);
+    Matcher standalonePattern = STANDALONE_CORE_PATTERN.matcher(metricName);
+    if (cloudPattern.find()) {
+      coreName = cloudPattern.group(1);
+      labels.put("core", cloudPattern.group(1));
+      labels.put("collection", cloudPattern.group(2));
+      labels.put("shard", cloudPattern.group(3));
+      labels.put("replica", cloudPattern.group(4));
```
dsmiley, line 41: I believe it's possible to have named groups in a regular expression.  If you enhance your regular expressions as such, you could generically apply the pattern with Matcher and extract all named groups without hard-coding the expectations separate from the regular expression, which is annoying and error-prone to maintain.

mlbiscoc: Cool, learned something new. Added.

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreMetric.java
```diff
@@ -30,18 +30,18 @@ public abstract class SolrCoreMetric extends SolrMetric {
...
+      coreName = cloudPattern.group("core");
+      labels.put("core", cloudPattern.group("core"));
+      labels.put("collection", cloudPattern.group("collection"));
+      labels.put("shard", cloudPattern.group("shard"));
+      labels.put("replica", cloudPattern.group("replica"));
     } else if (standalonePattern.find()) {
-      coreName = standalonePattern.group(1);
-      labels.put("core", standalonePattern.group(1));
+      coreName = standalonePattern.group("core");
+      labels.put("core", standalonePattern.group("core"));
```
dsmiley, line 44: I meant for you to do this generically with a loop, thus without the code here actually referring to any names.  See `Matcher.namedGroups()`.

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreMetric.java
```diff
@@ -30,18 +30,18 @@ public abstract class SolrCoreMetric extends SolrMetric {
   public String coreName;
 
   public SolrCoreMetric(Metric dropwizardMetric, String metricName) {
-    super(dropwizardMetric, metricName);
+    super(dropwizardMetric, metricName.substring(metricName.indexOf(".") + 1));
     Matcher cloudPattern = CLOUD_CORE_PATTERN.matcher(metricName);
     Matcher standalonePattern = STANDALONE_CORE_PATTERN.matcher(metricName);
```
dsmiley, line 35: Shouldn't those patterns be defined here on SolrCoreMetric where they are used?

dsmiley (commented):
Getting much better now

solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreCacheMetric.java
dsmiley: Much improved; thanks

dsmiley (approved):
+1 looks great to me.  I forgot about this PR; sorry.  Can you suggest CHANGES.txt wording (a comment here is fine).

mlbiscoc:
> +1 looks great to me. I forgot about this PR; sorry. Can you suggest CHANGES.txt wording (a comment here is fine).

No problem, thanks for reviewing. How about under bug fix "Metrics: Prometheus response writer fix for non-compliant exposition format containing duplicate TYPE lines"

dsmiley:
Apparently, named groups of Matcher was added in JDK 20.  Can you please do a branch_9x backport PR cherry picking this change and make modifications for JDK 11?  I'm thinking, declare a static Map of the named groups for each regexp.

dsmiley:
If you haven't started; I will.  It's just slightly more work in my IDE as I'm set up for the change.

mlbiscoc:
> If you haven't started; I will. It's just slightly more work in my IDE as I'm set up for the change.

No worries, I got it. I'll push a PR to 9x soon.

mlbiscoc:
[PR](https://github.com/apache/solr/pull/3006) for 9x backport

PR #2908, opened 2024-12-13
Clarify comment since examples do not use a special directory

epugh:
https://issues.apache.org/jira/browse/SOLR-17556

Spotted this comment, that I believe is no longer valid since SOLR-17556 was first commented.   I *sort* of wanted to delete the entire bit of logic, but decided it needs to stay...?

@malliaridis would love your eyes...

### DIFF
--- a/solr/bin/solr.cmd
+++ b/solr/bin/solr.cmd
@@ -839,8 +839,7 @@ IF NOT EXIST "%SOLR_HOME%\" (
   )
 )
 
-@REM This is quite hacky, but examples rely on a different log4j2.xml
-@REM so that we can write logs for examples to %SOLR_HOME%\..\logs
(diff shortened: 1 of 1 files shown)


malliaridis (commented):
I believe the comment is not relevant anymore and can be removed in the context of this PR, as I could not find any explicit reference in the examples that make use of a logs4j.xml file.

It is worth noting that there is an issue related to `SOLR_LOGS_DIR` on Windows that I reported in [SOLR-17598](https://issues.apache.org/jira/browse/SOLR-17598).

The missing test coverage for Windows is really a pain.

epugh:
Thanks for confirming this...    Yeah, no test structure sucks.   Hopefully htough, as we work out the set of changes that we introduced in 9.7 and 9.8 that things will calm down....

PR #2912, opened 2024-12-17
SOLR-15751: Modify v2 `GET /collections/collName` to support full ColStatus response

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-15751

# Description

Solr's v1 "/admin/collections?action=COLSTATUS" API is a useful way to get a high level summary of a single collection.  This includes the collection status and metadata (which CLUSTERSTATUS also returns) as well as very detailed data core-level data about fields, disk-usage, etc.

The v2 API lacks all of this functionality.

# Solution

This PR addresses this by modifying the existing v2 `GET /collections/collName` API to incorporate the v1 "colstatus" functionality.  (Previously this v2 endpoint returned the same response as v1's /admin/collections?action=CLUSTERSTATUS&collection=collName`)

The PR also converts this endpoint to JAX-RS in the process, so that it's included in our OAS and codegen.

# Tests

Added test in CollectionsAPISolrJTest that makes use of the OAS-generated SolrJ request/response.  Existing tests continue to pass.

# Checklist


epugh:
If you think there are parameters that maybe don't make sense going forward, I would vote for skipping them.   Get this in, and then see if folks highlight legit use cases for why they need those.  If you blindly carry them forward, well, then, we probably are stuck with them forever....    It's easier to add them later then it is to remove them... ;-).    Also, we should think about YAGNI more.

solr/core/src/java/org/apache/solr/api/V2HttpCall.java
```diff
@@ -188,6 +188,8 @@ public void call(SolrQueryRequest req, SolrQueryResponse rsp) {
 
       Thread.currentThread().setContextClassLoader(core.getResourceLoader().getClassLoader());
       this.path = path = path.substring(prefix.length() + pathSegments.get(1).length() + 2);
+      parts.put(
+          COLLECTION_PROP, origCorename); // Core-level API, so populate "collection" template val
```
dsmiley, line 192: (same)

solr/api/src/java/org/apache/solr/client/api/model/CollectionStatusResponse.java
```diff
@@ -0,0 +1,164 @@
...
+ * limitations under the License.
+ */
+package org.apache.solr.client.api.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import java.util.Map;
+
+/** Response of the CollectionStatusApi.getCollectionStatus() API */
+public class CollectionStatusResponse extends SolrJerseyResponse {
```
dsmiley, line 24: I love how structured / clear this is.  Previously (v1) was all untyped/magic.

solr/core/src/java/org/apache/solr/handler/admin/api/CollectionStatus.java
```diff
@@ -0,0 +1,89 @@
...
+    // TODO Push CollectionStatusResponse down into ColStatus and avoid the intermediate NL, if all
+    // usages can be switched over.
+    final var nlResponse = new NamedList<>();
+    new ColStatus(
+            coreContainer.getSolrClientCache(),
+            coreContainer.getZkController().getZkStateReader().getClusterState(),
+            new ZkNodeProps(params))
+        .getColStatus(nlResponse);
+    // collName is prop/key for a nested NL returned by ColStatus - extract the inner NL and
+    // manually add name to resp
```
dsmiley, line 82: so much for v1 moving to be on top of v2; ehh?

dsmiley (commented):
Thanks for this

gerlowskija:
re: impotent COLSTATUS boolean flags previously discussed [here](https://github.com/apache/solr/pull/2912#issuecomment-2548874545)

I think I understand this a bit better now.  The ref-guide docs for COLSTATUS describe each of the boolean flags as being independent of one another: `segments` returns base segment data, `fieldInfo` returns detailed information about each indexed field, etc.  But in reality on main/branch_9x currently, the flags are dependent on one another.  `fieldInfo` and `sizeInfo` have no impact unless paired with `segments=true`.  And `segments=true` has no impact on its own either.

The ref-guide coverage here doesn't suggest that this is intentional in any way, so I suspect this is a bug that eluded @sigram when he first did SOLR-13292.  (Hopefully AB can confirm his intentions if he catches this notification!)

Working on that theory, I'll bundle a fix for COLSTATUS into this PR and add some extra tests if there aren't any objections?

solr/api/src/java/org/apache/solr/client/api/model/CollectionStatusResponse.java
```diff
@@ -0,0 +1,164 @@
...
+    public String nodeName;
+
+    @JsonProperty("base_url")
+    public String baseUrl;
+
+    @JsonProperty public String state; // TODO Make this an enum?
+    @JsonProperty public String type; // TODO Make this an enum?
+
+    @JsonProperty("force_set_state")
+    public Boolean forceSetState;
```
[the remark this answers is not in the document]
gerlowskija: I can omit the field from the v2 response if you'd like, but this seems like a much larger concern than this PR.  It's only in the v2 response because it's in the v1 COLSTATUS response. (And it's only there because it's being persisted to ZK!)

solr/api/src/java/org/apache/solr/client/api/model/CollectionStatusResponse.java
```diff
@@ -0,0 +1,164 @@
...
+  public static class SegmentSummary {
+    @JsonProperty public String minSegmentLuceneVersion;
+    @JsonProperty public String commitLuceneVersion;
+    @JsonProperty public Integer numSegments;
+    @JsonProperty public String segmentsFileName;
+    @JsonProperty public Integer totalMaxDoc;
+
+    @JsonProperty
+    public Map<String, String>
+        userData; // Typically keys are 'commitCommandVer' and 'commitTimeMSec'
```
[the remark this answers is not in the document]
gerlowskija: It's ugly for sure.  Tidy is well-intentioned, but often a blunt instrument.

It's probably not worth too much of our time trying to clean up after it, but I'm happy to fix these few bits you highlighted.

solr/core/src/java/org/apache/solr/api/V2HttpCall.java
gerlowskija: Done

solr/core/src/java/org/apache/solr/handler/admin/api/CollectionStatus.java
gerlowskija: That model works great anytime that the v1 and v2 responses are exactly the same (which is almost always the case).  I made the v2 response for this API slightly different from v1 COLSTATUS, so it's hard to have CollectionsHandler call `V2ApiUtils.squashIntoNamedList(new CollectionStatus(..).getCollectionStatus(...))` the way we've been doing with most of our other v1-v2 pairs.

That said, I did a bit of restructuring here so that v1 can *mostly* sit on top of v2: CollectionsHandler.COLSTATUS now calls a static method `populateColStatusData` that lives in this class. Hopefully that gets us most of the benefit.  If we eventually tackle the TODO comment at L87, we should be able to close the gap and have CollectionsHandler.COLSTATUS consume the v2 API in the normal way.

gerlowskija:
Alright - COLSTATUS bug if fixed (see comment above), added some additional tests for both v1 and v2, and tests+check pass for me locally!  Will aim to merge this in a few days pending any more feedback

solr/api/src/java/org/apache/solr/client/api/model/CollectionStatusResponse.java
```diff
@@ -82,21 +97,93 @@ public static class LeaderSummary {
...
+  // Present with segment=true || sizeInfo=true
+  public static class SingleSegmentData {
+    @JsonProperty public String name;
+    @JsonProperty public Integer delCount;
+    @JsonProperty public Integer softDelCount;
+    @JsonProperty public Boolean hasFieldUpdates;
+    @JsonProperty public Long sizeInBytes;
+    @JsonProperty public Integer size;
+    // A date string of the form "2024-12-17T17:35:18.275Z"
+    @JsonProperty public String age;
```
dsmiley, line 138: Can we use Instant?

gerlowskija:
> Will aim to merge this in a few days pending any more feedback

Actually - thinking on it a bit - the v1 `/admin/segments` API is used under the hood by COLSTATUS, and has a very similar response format, so it probably makes sense to address within this PR as well.  I'll look to add that shortly.

gerlowskija:
Alright - added the `/admin/segments` v2 API, so this is back in "almost ready" territory.

Still need to look into David's question about use of 'Instant' - I fiddled a bit and got it to work with `Date`, so hopefully it's do-able in at least some places.

solr/api/src/java/org/apache/solr/client/api/model/CollectionStatusResponse.java
```diff
@@ -0,0 +1,249 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/** Response of the CollectionStatusApi.getCollectionStatus() API */
+public class CollectionStatusResponse extends SolrJerseyResponse {
+
+  @JsonProperty public String name;
+  @JsonProperty public Integer znodeVersion;
+  @JsonProperty public Long creationTimeMillis;
```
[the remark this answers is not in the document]
gerlowskija: We can change the variable name and probably the type on the deserialized-Java side.

But we'll need to keep the same serialization-time prop name and value to avoid backcompat issues.

gerlowskija:
> Still need to look into David's question about use of `Instant`

Alright - the main hiccup for using `Instant` is that it's not supported by our serialization layer.  Jackson supports it pretty easily on the v2 side, but our v1 "ResponseWriter"-based serialization code doesn't.  (The v1 code needs support because of how our v1 APIs are built "on top of" the v2 APIs - "primitive" types used in the v2 POJO end up in the NamedList that the v1 code serializes out.)

Anyway, I think it would be nice to support this, but I don't want to squeeze it into this PR so I'm not tempted to skimp on testing, etc.  I've filed [a ticket for it](https://issues.apache.org/jira/browse/SOLR-17608), and hope to pursue that shortly.  In the meantime I added 'TODO' comments to the places David flagged as being 'Instant'-ready - we can circle back to those either as a part of SOLR-17608, or once it's done.

(FYI @dsmiley )

solr/core/src/java/org/apache/solr/handler/admin/ColStatus.java
```diff
@@ -201,32 +204,35 @@ public void getColStatus(NamedList<Object> results) {
...
-              for (Map.Entry<String, ?> entry : segs) {
-                NamedList<Object> fields =
-                    (NamedList<Object>) ((NamedList<Object>) entry.getValue()).get("fields");
-                if (fields != null) {
-                  for (Map.Entry<String, Object> fEntry : fields) {
-                    Object nc = ((NamedList<Object>) fEntry.getValue()).get("nonCompliant");
-                    if (nc != null) {
+              for (Map.Entry<String, GetSegmentDataResponse.SingleSegmentData> entry :
+                  segs.entrySet()) {
+                final var fieldInfoByName = entry.getValue().fields;
```
dsmiley, line 215: I love how typed this is now.  This is an important part of the V2 vision coming into reality.

solr/core/src/java/org/apache/solr/handler/admin/api/GetSegmentData.java
```diff
@@ -0,0 +1,501 @@
...
+ *
+ * <p>Equivalent to the v1 /solr/coreName/admin/segments endpoint.
+ */
+public class GetSegmentData extends JerseyResource implements SegmentsApi {
+
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  private static final double GB = 1024.0 * 1024.0 * 1024.0;
+
+  private static final List<String> FI_LEGEND =
+      Arrays.asList(
```
dsmiley, line 83: please avoid asList in favor of the immutable List.of

solr/core/src/java/org/apache/solr/handler/admin/api/GetSegmentData.java
```diff
@@ -0,0 +1,501 @@
...
+                  })
+              .sorted(
+                  (p1, p2) -> {
+                    if (p1.second() > p2.second()) {
+                      return -1;
+                    } else if (p1.second() < p2.second()) {
+                      return 1;
+                    } else {
+                      return 0;
+                    }
```
dsmiley, line 359: I think this can be replaced by usage of `java.util.Comparator#comparing(java.util.function.Function<? super T,? extends U>)`

solr/core/src/test/org/apache/solr/handler/admin/IndexSizeEstimatorTest.java
```diff
@@ -177,56 +180,42 @@ public void testIntegration() throws Exception {
...
+              .convertValue(segInfosRaw, CollectionStatusResponse.SegmentInfo.class);
+
+      final var rawSize = segInfos.rawSize;
       assertNotNull("rawSize missing", rawSize);
-      @SuppressWarnings({"unchecked"})
-      Map<String, Object> rawSizeMap = rawSize.asMap(10);
-      @SuppressWarnings({"unchecked"})
-      Map<String, Object> fieldsBySize =
-          (Map<String, Object>) rawSizeMap.get(IndexSizeEstimator.FIELDS_BY_SIZE);
+      Map<String, String> fieldsBySize = rawSize.fieldsBySize;
```
dsmiley, line 194: again; so much better

solr/core/src/java/org/apache/solr/handler/admin/api/GetSegmentData.java
gerlowskija: [0] This stuff actually pre-exists this PR; it previously lived in SegmentInfoRequestHandler.

I generally try to avoid unrelated cleanups as I go through these v2 PRs.  Partly for scope (if I stopped to fix every little thing, forward progress would halt entirely).  And partly because you never know what seemingly "safe" change will end up biting you.

solr/core/src/java/org/apache/solr/handler/admin/api/GetSegmentData.java
gerlowskija: [0] This stuff actually pre-exists this PR; it previously lived in SegmentInfoRequestHandler.

I generally try to avoid unrelated cleanups as I go through these v2 PRs. Partly for scope (if I stopped to fix every little thing, forward progress would halt entirely). And partly because you never know what seemingly "safe" change will end up biting you.

solr/api/src/java/org/apache/solr/client/api/model/CollectionStatusResponse.java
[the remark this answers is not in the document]
gerlowskija: Yep, `Date` can work until SOLR-17608 gets sorted out.  If we think "Instant" is generally better, I can switch to 'Date', but add in a TODO comment with the JIRA ticket, so we can find this later on...

solr/api/src/java/org/apache/solr/client/api/endpoint/SegmentsApi.java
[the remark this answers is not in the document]
gerlowskija: Fwiw, we've already got some reuse here: e.g. [these SegmentsApi constants](https://github.com/apache/solr/pull/2912/files#diff-b26451021791966580a1c1416f23f66067ccc636be6c09a8c62486b95255ec36R36-R49) get re-used in CollectionStatusApi as well.

The main thing that prevents us from doing something more with inheritance, etc. is that the method signatures for CollectionStatusApi/SegmentsApi aren't the same.  The CollectionStatusApi method takes two parameters ('collectionName' and 'segments') that have no SegmentApi equivalent.

solr/api/src/java/org/apache/solr/client/api/model/CollectionStatusResponse.java
gerlowskija: I've used `Date` for now based on PR discussion, see SOLR-17608.

solr/api/src/java/org/apache/solr/client/api/model/CollectionStatusResponse.java
```diff
@@ -0,0 +1,249 @@
...
+  // Present with segments=true
+  public static class SegmentDiagnosticInfo {
+    @JsonProperty("os.version")
+    public String osVersion;
+
+    @JsonProperty("lucene.version")
+    public String luceneVersion;
+
+    @JsonProperty public String source;
+    @JsonProperty public Long timestamp;
```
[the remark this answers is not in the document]
gerlowskija: I've used `Date` for now based on PR-level discussion, see [SOLR-17608](https://issues.apache.org/jira/browse/SOLR-17608).

gerlowskija:
Alright, thanks all for the feedback.  Merged and backported!

PR #2916, opened 2024-12-18
SOLR-17582 Stream CLUSTERSTATUS API response

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-17582

# Description

`CLUSTERSTATUS` would aggregate and build the whole response including all collection properties into an `NamedList` before finally passing it to the response writers. With many or thousands of collections, this can use up memory.

# Solution

Instead of returning a `NamedList`, return a `MapWriter`. The Mapwriter will pull an `Iterator` from the documentStream and create `SolrCollectionProperiesIterator` implementing `iterator` and `override` `next()` to create and write the response in time, instead of writing the whole thing in memory.

# Tests

Fix appropriate tests that were using `NamedList` to `Map` because of the `MapWriter`.

# Checklist


solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -368,4 +341,77 @@ public static Map<String, Object> postProcessCollectionJSON(Map<String, Object>
     collection.put("health", Health.combine(healthStates).toString());
     return collection;
   }
+
+  private class SolrCollectionProperiesIterator implements Iterator<NamedList<Object>> {
```
dsmiley, line 345: I was hoping we wouldn't need a custom Iterator.  We do need a method that takes in DocCollection (and some other context that is the same across collections) and returns a NamedList.  With such a method, we can call it via streamOfDocCollection.map(collState -> theMethod(collState, routeKey, liveNodes, etc.).iterator()

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -140,11 +139,11 @@ private ClusterState fetchClusterState(SolrClient client)
       liveNodesTimestamp = System.nanoTime();
     }
 
-    var collectionsNl = (NamedList<Map<String, Object>>) cluster.get("collections");
+    var collectionsNl = (Map<String, Map<String, Object>>) cluster.get("collections");
```
dsmiley, line 142: this var name is no longer appropriate.  Should be collectionsMap (and we assume by name).

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -185,7 +184,9 @@ private DocCollection fetchCollectionState(SolrClient client, String collection)
     SimpleOrderedMap<?> cluster =
         submitClusterStateRequest(client, collection, ClusterStateRequestType.FETCH_COLLECTION);
 
-    var collStateMap = (Map<String, Object>) cluster.findRecursive("collections", collection);
+    var collState = (Map<String, Object>) cluster.findRecursive("collections");
```
dsmiley, line 187: so findRecursive can only resolve collections and not Map?  Seems worth a small enhancement to do so but I understand if you don't bother or defer.

dsmiley (commented):
Thanks for contributing!

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -368,4 +341,77 @@ public static Map<String, Object> postProcessCollectionJSON(Map<String, Object>
...
+        DocRouter router = clusterStateCollection.getRouter();
+        Collection<Slice> slices = router.getSearchSlices(routeKey, null, clusterStateCollection);
+        for (Slice slice : slices) {
+          requestedShards.add(slice.getName());
+        }
+      }
+      if (shard != null) {
+        String[] paramShards = shard.split(",");
+        requestedShards.addAll(Arrays.asList(paramShards));
+      }
```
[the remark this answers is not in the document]
mlbiscoc: Moved it out so its not per iteration.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -368,4 +341,77 @@ public static Map<String, Object> postProcessCollectionJSON(Map<String, Object>
...
+
+      byte[] bytes = Utils.toJSON(clusterStateCollection);
+      @SuppressWarnings("unchecked")
+      Map<String, Object> docCollection = (Map<String, Object>) Utils.fromJSON(bytes);
```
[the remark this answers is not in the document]
mlbiscoc: So looked at this and I understand now why it was done this way. It wants to just write this out to the `TextWriter` as a normal POJO but that doesn't seem to be what the `DocCollection` class is. So instead what was done was to just write it to JSON `byte[]` then back to a `Map` which was the "easiest" way.

Was looking to avoid this back and forth, but there were a few options. I tried using an `ObjectMapper` to `Map.class` but there is an error for unable to cast Instant due to a missing dependency for Jackson we need to introduce.
`Java 8 date/time type `java.time.Instant` not supported by default Issue`

Other way is to introduce a some kind of `toMap` method so that the TextWriter can write this as a generic `Map`.

Another option which actually looks like the way we should go is I found that the `DocCollection` class extends ZkNodeProps which implements `MapWriter`. `DocCollection` already overrides `[writeMap](https://github.com/apache/solr/blob/9cef6e390719cbd7b55085cfef98fcb053785f77/solr/solrj/src/java/org/apache/solr/common/cloud/DocCollection.java#L427)` so we could just return this to the TextWriter! Unfortunately the ClusterStatus class does a bunch of JSON post processing such as `Health` added to the `Map` that the output is missing some things because of this `[postProcessCollectionJSON()](https://github.com/apache/solr/blob/9cef6e390719cbd7b55085cfef98fcb053785f77/solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java#L334)` method. 

I am thinking we should refactor DocCollection to so we can just return this but the changes were much more drastic but may be worth it. Maybe in a different JIRA? This scope continues to creep with me adding improvement to `NamedList`. Would be happy to refactor this and pick it up if you agree. Would clean up much more code and avoid this JSON processing for every collection iteration.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
mlbiscoc: Ah yeah don't know why I didn't do this first. Changed it appropriately.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
mlbiscoc: Made the small enhancement to NamedList and added a test to make sure it works. Lmk what you think.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -205,54 +214,32 @@ private void fetchClusterStatusForCollOrAlias(
...
+          while (it.hasNext()) {
+            Map<String, Object> props = it.next();
+            props.forEach(
+                (key, value) -> {
+                  try {
+                    ew.put(key, value);
+                  } catch (IOException e) {
+                    throw new RuntimeException(e);
+                  }
+                });
```
mlbiscoc, line 239: I am thinking, if we can change up DocCollection in another PR, we can delete all those post processing functions and condense the response to this.
```java
MapWriter collectionPropsWriter =
        ew -> {
          Iterator<DocCollection> it = collectionStream.iterator();
          while (it.hasNext()) {
            DocCollection collState = it.next();
            ew.put(collState.getName(), collState);
          }
        };
```

dsmiley (commented):
Leave the round-trip serialization -- I didn't realize it was that way before.
Leave findRecursive be... not as easy as you thought and probably deserves its own issue.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -205,54 +214,32 @@ private void fetchClusterStatusForCollOrAlias(
...
+                  .map(
+                      (collectionState) ->
+                          collectionPropsResponse(
+                              collectionState,
+                              collectionVsAliases,
+                              routeKey,
+                              liveNodes,
+                              requestedShards))
+                  .iterator();
+          while (it.hasNext()) {
```
dsmiley, line 230: Oh; now that I see you are iterating it yourself (I had thought that would be driven externally), you shouldn't bother calling .iterator then looping; you can loop directly with stream.forEach

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
dsmiley: I should mention Solr has two similar APIs, ClusterStatus & ColStatus.  See org.apache.solr.handler.admin.ColStatus#getColStatus.  It'd be awesome if there was a single method that takes a DocCollection (and some other params as needed) to produce a NamedList.  At least ColStatus's code doesn't have the sad JSON round-trip.  Maybe it's own PR or not as you wish.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -342,76 +356,47 @@ public static Map<String, Object> postProcessCollectionJSON(Map<String, Object>
...
+      Set<String> requestedShards) {
+    Map<String, Object> collectionProps = new HashMap<>();
+    Map<String, Object> collectionStatus;
+    String name = clusterStateCollection.getName();
+
+    if (routeKey != null) {
+      DocRouter router = clusterStateCollection.getRouter();
+      Collection<Slice> slices = router.getSearchSlices(routeKey, null, clusterStateCollection);
+      for (Slice slice : slices) {
+        requestedShards.add(slice.getName());
```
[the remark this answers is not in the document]
mlbiscoc: This is a bug. I moved that old logic into here without realizing that parameter is manipulated. Changed `requestedShards` to `Set.of()` for immutability and made changes accordingly.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -342,76 +356,47 @@ public static Map<String, Object> postProcessCollectionJSON(Map<String, Object>
...
+    if (collectionVsAliases.containsKey(name) && !collectionVsAliases.get(name).isEmpty()) {
+      collectionStatus.put("aliases", collectionVsAliases.get(name));
+    }
+    String configName = clusterStateCollection.getConfigName();
+    collectionStatus.put("configName", configName);
+    if (solrParams.getBool("prs", false) && clusterStateCollection.isPerReplicaState()) {
+      PerReplicaStates prs = clusterStateCollection.getPerReplicaStates();
+      collectionStatus.put("PRS", prs);
     }
+    collectionProps.put(name, collectionStatus);
```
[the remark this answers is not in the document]
mlbiscoc: Yeah taking a deeper look, this extra map isn't needed. I kind of did it because of `crossCheckReplicaStateWithLiveNodes` but instead I modified that method and just return `collectionStatus`.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
mlbiscoc: Gotcha. I'll take a look at this in a separate PR/Jira

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -183,10 +181,12 @@ private void fetchClusterStatusForCollOrAlias(
...
 
-    Set<String> requestedShards = new HashSet<>();
+    Set<String> requestedShards;
     if (shard != null) {
       String[] paramShards = shard.split(",");
-      requestedShards.addAll(Arrays.asList(paramShards));
+      requestedShards = Set.of(paramShards);
+    } else {
+      requestedShards = Set.of();
     }
```
dsmiley, line 190: I bet this whole thing could be one line using ternary operator and no line wrap. 
Also, if "shard" isn't specified, I think it's clearer to have requestedShards as null to indicate there was no request for any shards.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -356,28 +346,28 @@ public static Map<String, Object> postProcessCollectionJSON(Map<String, Object>
...
+    Set<String> shards = new HashSet<>(requestedShards);
     String name = clusterStateCollection.getName();
 
     if (routeKey != null) {
       DocRouter router = clusterStateCollection.getRouter();
       Collection<Slice> slices = router.getSearchSlices(routeKey, null, clusterStateCollection);
       for (Slice slice : slices) {
-        requestedShards.add(slice.getName());
+        shards.add(slice.getName());
       }
```
dsmiley, line 364: something seems wrong to me here.  If requestedShards is specified, we should use that.  If routeKey is specified, we should use that (to compute the shards).  Or neither (99% of users won't do either).  But we shouldn't do both if for no other reason that it's confusing.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -295,22 +288,19 @@ private Map<String, Object> getCollectionStatus(
...
-            if (!liveNodes.contains(node_name)) {
-              // node is not live, so this replica is actually down
-              replicaMap.put(ZkStateReader.STATE_PROP, Replica.State.DOWN.toString());
-            }
+    Map<String, Object> shards = (Map<String, Object>) collectionProps.get("shards");
+    for (Object nextShard : shards.values()) {
+      Map<String, Object> shardMap = (Map<String, Object>) nextShard;
+      Map<String, Object> replicas = (Map<String, Object>) shardMap.get("replicas");
+      for (Object nextReplica : replicas.values()) {
+        Map<String, Object> replicaMap = (Map<String, Object>) nextReplica;
```
dsmiley, line 296: I almost want to cry just glancing at this.
This is the poster-child for why Java introduced "var".  And there may be other approaches to improve it but that's the simplest.
(Yeah you didn't write it; I know)

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -216,28 +216,21 @@ private void fetchClusterStatusForCollOrAlias(
...
+                  ew.put(
+                      collectionState.getName(),
+                      buildResponseForCollection(
+                          collectionState,
+                          collectionVsAliases,
+                          routeKey,
+                          liveNodes,
+                          requestedShards));
+                } catch (IOException e) {
+                  throw new RuntimeException(e);
```
dsmiley, line 231: Why does buildResponseForCollection throw IOException?  That's suspicious.

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
```diff
@@ -216,28 +216,21 @@ private void fetchClusterStatusForCollOrAlias(
...
-                (key, value) -> {
-                  try {
-                    ew.put(key, value);
-                  } catch (IOException e) {
-                    throw new RuntimeException(e);
-                  }
-                });
-          }
+          collectionStream.forEach(
+              (collectionState) -> {
```
dsmiley, line 220: friggin beautiful now

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
mlbiscoc: We could throw a `SolrException(BAD_REQUEST)` higher up the stack if passed both `shard` and `_route_`. I can update the ref-docs to say you can only do one or the other. But this could break people who are doing both.

I also changed that logic into a single line and `forEach`

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
dsmiley: putNoEx then

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
dsmiley: I recommend that but understand if you leave out of scope.
(I really don't think this is going to break anyone's current usage, BTW)

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
dsmiley: totally understood.  In some old Solr code like this, there's always "and one more thing" we could/should do but ultimately snowballs the scope out of control.  I leave it to you to do as you wish.  Thank you for your contribution here; I didn't mean to get more out of you than you bargained for :-)

solr/core/src/java/org/apache/solr/handler/admin/ClusterStatus.java
mlbiscoc: I'll add it in when I come back for the refactor.

dsmiley (commented):
Uh oh; a back-compat problem is dawning on me.  Any SolrJ consumer of this API (especially using the default "javabin" format) is going to break here.  The change to BaseHttpClusterStateProvider should have made this glaringly obvious but I overlooked its implication earlier (face-palm).  

I wouldn't mind if we do this only for V2 (which doesn't yet have back-compatibility concerns) but V2 only has one collection status list returning API (/cluster) and I've debated against the very existence of /cluster for V2.  There's not yet a proposal for an alternative.  Furthermore V2 raises the bigger question of coalescing on a single DocCollection serialization method.

If we forge ahead anyway (not waiting for V2), we'll have to work around this compatibility.  I had a couple ideas but whittled down to one:
* detect the response format in this handler.  If it's Javabin, we write to a NamedList (or better, SimpleOrderedMap), defeating the point of this PR _for that format_.  If JSON/XML is used, this PR will accomplish its goal. Need to write a little test for JSON that at least partially sanity checks the result.  Such a test should pass without this PR.

mlbiscoc:
> Uh oh; a back-compat problem is dawning on me. Any SolrJ consumer of this API (especially using the default "javabin" format) is going to break here. The change to BaseHttpClusterStateProvider should have made this glaringly obvious but I overlooked its implication earlier (face-palm).
> 
> I wouldn't mind if we do this only for V2 (which doesn't yet have back-compatibility concerns) but V2 only has one collection status list returning API (/cluster) and I've debated against the very existence of /cluster for V2. There's not yet a proposal for an alternative. Furthermore V2 raises the bigger question of coalescing on a single DocCollection serialization method.
> 
> If we forge ahead anyway (not waiting for V2), we'll have to work around this compatibility. I had a couple ideas but whittled down to one:
> 
> * detect the response format in this handler.  If it's Javabin, we write to a NamedList (or better, SimpleOrderedMap), defeating the point of this PR _for that format_.  If JSON/XML is used, this PR will accomplish its goal. Need to write a little test for JSON that at least partially sanity checks the result.  Such a test should pass without this PR.

Well thats unfortunate... I'm not too familiar with SolrJ (I'll maybe play with it more to catch stuff like this) but I am assuming now the API is going to return a `Map` and no longer a `NamedList` to the client thus breaking them now?

I'll continue with this and add in your ideas to keep this backwards compatible.

dsmiley:
Yes; as the code you modified in a ClusterStateProvider shows, the NamedList->Map happened and broke that code.

You could add a check for {{params.get(CommonParams.WT)}} and check if "javabin" is the value.

mlbiscoc:
> Yes; as the code you modified in a ClusterStateProvider shows, the NamedList->Map happened and broke that code.
> 
> You could add a check for {{params.get(CommonParams.WT)}} and check if "javabin" is the value.

Added in that check for `javabin` to keep backwards compatibility with SolrJ and reverted all those SolrJ changes. Also created `clusterStatusWithCollectionAndShardJSON` to sanity check the `json` response from `MapWriter`. Happy holidays!

PR #2919, opened 2024-12-19
Deprecate MapWriter.append

dsmiley:
I don't really like MapWriter.append – undocumented and called in only 2 places.  Since MapWriter is implemented by a great many things, we should be conservative adding methods there.   append() is kind of clever but I think it's outside the scope of what MapWriter should be.

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/common/MapWriter.java
+++ b/solr/solrj/src/java/org/apache/solr/common/MapWriter.java
@@ -33,6 +33,9 @@ import org.noggit.JSONWriter;
  */
 public interface MapWriter extends MapSerializable, NavigableObject, JSONWriter.Writable {
 
+  /** Writes this object's entries out to {@code ew}. */
+  void writeMap(EntryWriter ew) throws IOException;
+
   default String jsonStr() {
     return Utils.toJSONString(this);
   }
@@ -42,6 +45,7 @@ public interface MapWriter extends MapSerializable, NavigableObject, JSONWriter.
     return Utils.convertToMap(this, map);
   }
 
+  /** For implementing Noggit {@link org.noggit.JSONWriter.Writable}. */
   @Override
   default void write(JSONWriter writer) {
     writer.startObject();
@@ -70,8 +74,7 @@ public interface MapWriter extends MapSerializable, NavigableObject, JSONWriter.
     writer.endObject();
   }
 
-  void writeMap(EntryWriter ew) throws IOException;
-
+  @Deprecated
   default MapWriter append(MapWriter another) {
     MapWriter m = this;
     return ew -> {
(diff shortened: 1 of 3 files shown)


solr/solrj/src/java/org/apache/solr/common/MapWriter.java
```diff
@@ -70,8 +74,7 @@ public MapWriter.EntryWriter put(CharSequence k, Object v) {
     writer.endObject();
   }
 
-  void writeMap(EntryWriter ew) throws IOException;
-
+  @Deprecated
```
epugh, line 77: I kind of wish we always provide a bit more detail on these `@Deprecate` tags.  I like when I see a "since" as it let's me know when the code can be deleted.   I like when there is a "use X instead" comment.

I did a pass the other week around deprecated code and a lot it wasn't clear what would trigger the deleting of it..  Or how long it had been deprecated.

epugh (approved):
I like seeing some of these various Map methods deprecated, and look forawrd to them being removed!   Especially when they aren't wildely used!

solr/solrj/src/java/org/apache/solr/common/MapWriter.java
dsmiley: for an obscure method; I think we shouldn't burden ourselves with overly helping likely non-existent users of such.  Personally I do plenty of code sleuthing to identify when lines of code are added/removed like a Deprecate or whatever.

dsmiley: > I did a pass the other week around deprecated code and a lot it wasn't clear what would trigger the deleting of it.. Or how long it had been deprecated.

Curious; in that instance, why did you care?  In this case, it's just some small utility method that any caller can do themselves.  Thus the deprecation rationale is not something I think any user would find interesting _for this method_.

epugh: I was looking at code in 10 that was marked deprecated so I could remove it.  I'd love to see a release of 10 that has NO deprecation tags in it, which may be a pipe dream...  I was hoping to see some indication that would tell me what methods to be used in place for example.

epugh:
I think I can better articulate my original question.   Now that `MapWriter.append()` is deprecated on 9x, can we remove it from 10x?   I like that `TestLegacyCircuitBreaker` has a lot of direction for future Solr developers...:

```
/**
 * Tests the original circuit breaker configuration format, which was not configurable.
 *
 * @deprecated Remove in 10.0
 */
@Deprecated(since = "9.4")
```

dsmiley:
Yes; anything deprecated in 9x is removable in 10x.
That's a general statement, and it applies here too.  I know we sometimes deprecate things that are still used.

PR #2921, opened 2024-12-20
SOLR-17525 backport to 9.x

alessandrobenedetti:
https://issues.apache.org/jira/browse/SOLR-17525

Backport to satisfy gradle


alessandrobenedetti:
So @malliaridis , I started the backporting, you see the cherry picked commit and some gradle changes that should be what we want for 9.x.

I'm waiting for the gradle checks to run and be green.
You additional validation would be really helpful to close this story :)
Thanks!

malliaridis:
I've looked into the gradle changes and they look good so far. Only one note, there is a checksum file `langchain4j-0.35.0.jar.sha1` that is being removed when running `gradlew updateLicenses`. So you may want to commit the removal as well.

Additionally, 9x workflows sometimes stuck during compilation and take up to 6h to be interrupted at the end. So you may want to restart the action if it takes more than 30 minutes on the same gradle task.

PR #2924, opened 2024-12-23
SOLR-16903: Migrate off java.io.File to java.nio.file.Path from core files

mlbiscoc:

# Description

Still in draft. Tests seem to pass locally but a lot can still be refactored better around Path.

This should remove all imports of `java.io.File` inside main core files. This excludes test files as there are too many changes to include into this PR and will make that in a separate PR.

# Solution

Found where `import java.io.File;` was being used and migrated logic to `java.nio.file.Path` as best as possible.

# Tests

Tests pass locally

# Checklist

### DIFF
--- a/solr/api/src/java/org/apache/solr/client/api/endpoint/NodeFileStoreApis.java
+++ b/solr/api/src/java/org/apache/solr/client/api/endpoint/NodeFileStoreApis.java
@@ -26,7 +26,6 @@ import jakarta.ws.rs.GET;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.PathParam;
 import jakarta.ws.rs.QueryParam;
-import java.io.IOException;
 import org.apache.solr.client.api.model.SolrJerseyResponse;
 
 /**
@@ -64,6 +63,5 @@ public interface NodeFileStoreApis {
           String getFrom,
       @Parameter(description = "Indicates that (only) file metadata should be fetched")
           @QueryParam("meta")
-          Boolean meta)
-      throws IOException;
+          Boolean meta);
 }
--- a/solr/api/src/java/org/apache/solr/client/api/endpoint/NodeFileStoreApis.java
+++ b/solr/api/src/java/org/apache/solr/client/api/endpoint/NodeFileStoreApis.java
@@ -26,6 +26,7 @@ import jakarta.ws.rs.GET;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.PathParam;
 import jakarta.ws.rs.QueryParam;
+import java.io.IOException;
 import org.apache.solr.client.api.model.SolrJerseyResponse;
 
 /**
@@ -63,5 +64,6 @@ public interface NodeFileStoreApis {
           String getFrom,
       @Parameter(description = "Indicates that (only) file metadata should be fetched")
           @QueryParam("meta")
-          Boolean meta);
+          Boolean meta)
+      throws IOException;
 }
--- a/solr/core/src/java/org/apache/solr/cloud/ZkSolrResourceLoader.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ZkSolrResourceLoader.java
@@ -120,6 +120,7 @@ public class ZkSolrResourceLoader extends SolrResourceLoader {
 
     try {
       // delegate to the class loader (looking into $INSTANCE_DIR/lib jars)
+      // Use Path to normalize path separator
       is = classLoader.getResourceAsStream(Path.of(resource).toString());
     } catch (Exception e) {
       throw new IOException("Error opening " + resource, e);
--- a/solr/core/src/java/org/apache/solr/cloud/ZkSolrResourceLoader.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ZkSolrResourceLoader.java
@@ -17,7 +17,6 @@
 package org.apache.solr.cloud;
 
 import java.io.ByteArrayInputStream;
-import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.invoke.MethodHandles;
@@ -121,7 +120,7 @@ public class ZkSolrResourceLoader extends SolrResourceLoader {
 
     try {
       // delegate to the class loader (looking into $INSTANCE_DIR/lib jars)
-      is = classLoader.getResourceAsStream(resource.replace(File.separatorChar, '/'));
+      is = classLoader.getResourceAsStream(Path.of(resource).toString());
     } catch (Exception e) {
       throw new IOException("Error opening " + resource, e);
     }
--- a/solr/core/src/java/org/apache/solr/cloud/ZkSolrResourceLoader.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ZkSolrResourceLoader.java
@@ -20,6 +20,7 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.invoke.MethodHandles;
+import java.nio.file.FileSystems;
 import java.nio.file.Path;
 import org.apache.solr.common.SolrException.ErrorCode;
 import org.apache.solr.common.cloud.ZooKeeperException;
@@ -120,8 +121,9 @@ public class ZkSolrResourceLoader extends SolrResourceLoader {
 
     try {
       // delegate to the class loader (looking into $INSTANCE_DIR/lib jars)
-      // Use Path to normalize path separator
-      is = classLoader.getResourceAsStream(Path.of(resource).toString());
+      is =
+          classLoader.getResourceAsStream(
+              resource.replace(FileSystems.getDefault().getSeparator(), "/"));
     } catch (Exception e) {
       throw new IOException("Error opening " + resource, e);
     }
--- a/solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
+++ b/solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
@@ -16,7 +16,6 @@
  */
 package org.apache.solr.core;
 
-import java.io.File;
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.nio.file.DirectoryStream;
@@ -367,10 +366,7 @@ public abstract class CachingDirectoryFactory extends DirectoryFactory {
   }
 
   private static boolean isSubPath(CacheValue cacheValue, CacheValue otherCacheValue) {
-    int one = cacheValue.path.lastIndexOf(File.separatorChar);
-    int two = otherCacheValue.path.lastIndexOf(File.separatorChar);
-
-    return otherCacheValue.path.startsWith(cacheValue.path + File.separatorChar) && two > one;
+    return Path.of(otherCacheValue.path).startsWith(Path.of(cacheValue.path));
   }
 
   @Override
--- a/solr/core/src/java/org/apache/solr/core/CoreDescriptor.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreDescriptor.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.io.Reader;
 import java.lang.invoke.MethodHandles;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.HashMap;
@@ -62,7 +63,8 @@ public class CoreDescriptor {
   public static final String CORE_CONFIGSET_PROPERTIES = "configSetProperties";
   public static final String SOLR_CORE_PROP_PREFIX = "solr.core.";
 
-  public static final Path DEFAULT_EXTERNAL_PROPERTIES_FILE = Path.of("conf/solrcore.properties");
+  public static final String DEFAULT_EXTERNAL_PROPERTIES_FILE =
+      "conf" + FileSystems.getDefault().getSeparator() + "solrcore.properties";
 
   /**
    * Whether this core was configured using a configSet that was trusted. This helps in avoiding the
@@ -94,7 +96,7 @@ public class CoreDescriptor {
           CORE_CONFIG, "solrconfig.xml",
           CORE_SCHEMA, "schema.xml",
           CORE_CONFIGSET_PROPERTIES, ConfigSetProperties.DEFAULT_FILENAME,
-          CORE_DATADIR, Path.of("data").toString(),
+          CORE_DATADIR, "data" + FileSystems.getDefault().getSeparator(),
           CORE_TRANSIENT, "false",
           CORE_LOADONSTARTUP, "true");
 
@@ -238,8 +240,7 @@ public class CoreDescriptor {
    * <p>File paths are taken as read from the core's instance directory if they are not absolute.
    */
   protected void loadExtraProperties() {
-    String filename =
-        coreProperties.getProperty(CORE_PROPERTIES, DEFAULT_EXTERNAL_PROPERTIES_FILE.toString());
+    String filename = coreProperties.getProperty(CORE_PROPERTIES, DEFAULT_EXTERNAL_PROPERTIES_FILE);
     Path propertiesFile = instanceDir.resolve(filename);
     if (Files.exists(propertiesFile)) {
       try (Reader r = Files.newBufferedReader(propertiesFile, StandardCharsets.UTF_8)) {
--- a/solr/core/src/java/org/apache/solr/core/CoreDescriptor.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreDescriptor.java
@@ -22,7 +22,6 @@ import java.lang.invoke.MethodHandles;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
@@ -64,7 +63,7 @@ public class CoreDescriptor {
   public static final String SOLR_CORE_PROP_PREFIX = "solr.core.";
 
   public static final String DEFAULT_EXTERNAL_PROPERTIES_FILE =
-      Paths.get("conf", "solrcore.properties").toString();
+      Path.of("conf/solrcore.properties").toString();
 
   /**
    * Whether this core was configured using a configSet that was trusted. This helps in avoiding the
@@ -96,7 +95,7 @@ public class CoreDescriptor {
           CORE_CONFIG, "solrconfig.xml",
           CORE_SCHEMA, "schema.xml",
           CORE_CONFIGSET_PROPERTIES, ConfigSetProperties.DEFAULT_FILENAME,
-          CORE_DATADIR, Paths.get("data/").toString(),
+          CORE_DATADIR, Path.of("data/").toString(),
           CORE_TRANSIENT, "false",
           CORE_LOADONSTARTUP, "true");
 
--- a/solr/core/src/java/org/apache/solr/core/CoreDescriptor.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreDescriptor.java
@@ -16,13 +16,13 @@
  */
 package org.apache.solr.core;
 
-import java.io.File;
 import java.io.IOException;
 import java.io.Reader;
 import java.lang.invoke.MethodHandles;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
@@ -64,7 +64,7 @@ public class CoreDescriptor {
   public static final String SOLR_CORE_PROP_PREFIX = "solr.core.";
 
   public static final String DEFAULT_EXTERNAL_PROPERTIES_FILE =
-      "conf" + File.separator + "solrcore.properties";
+      Paths.get("conf", "solrcore.properties").toString();
 
   /**
    * Whether this core was configured using a configSet that was trusted. This helps in avoiding the
@@ -96,7 +96,7 @@ public class CoreDescriptor {
           CORE_CONFIG, "solrconfig.xml",
           CORE_SCHEMA, "schema.xml",
           CORE_CONFIGSET_PROPERTIES, ConfigSetProperties.DEFAULT_FILENAME,
-          CORE_DATADIR, "data" + File.separator,
+          CORE_DATADIR, Paths.get("data/").toString(),
           CORE_TRANSIENT, "false",
           CORE_LOADONSTARTUP, "true");
 
@@ -240,7 +240,8 @@ public class CoreDescriptor {
    * <p>File paths are taken as read from the core's instance directory if they are not absolute.
    */
   protected void loadExtraProperties() {
-    String filename = coreProperties.getProperty(CORE_PROPERTIES, DEFAULT_EXTERNAL_PROPERTIES_FILE);
+    String filename =
+        coreProperties.getProperty(CORE_PROPERTIES, DEFAULT_EXTERNAL_PROPERTIES_FILE.toString());
     Path propertiesFile = instanceDir.resolve(filename);
     if (Files.exists(propertiesFile)) {
       try (Reader r = Files.newBufferedReader(propertiesFile, StandardCharsets.UTF_8)) {
--- a/solr/core/src/java/org/apache/solr/core/CoreDescriptor.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreDescriptor.java
@@ -62,8 +62,7 @@ public class CoreDescriptor {
   public static final String CORE_CONFIGSET_PROPERTIES = "configSetProperties";
   public static final String SOLR_CORE_PROP_PREFIX = "solr.core.";
 
-  public static final String DEFAULT_EXTERNAL_PROPERTIES_FILE =
-      Path.of("conf/solrcore.properties").toString();
+  public static final Path DEFAULT_EXTERNAL_PROPERTIES_FILE = Path.of("conf/solrcore.properties");
 
   /**
    * Whether this core was configured using a configSet that was trusted. This helps in avoiding the
@@ -95,7 +94,7 @@ public class CoreDescriptor {
           CORE_CONFIG, "solrconfig.xml",
           CORE_SCHEMA, "schema.xml",
           CORE_CONFIGSET_PROPERTIES, ConfigSetProperties.DEFAULT_FILENAME,
-          CORE_DATADIR, Path.of("data/").toString(),
+          CORE_DATADIR, Path.of("data").toString(),
           CORE_TRANSIENT, "false",
           CORE_LOADONSTARTUP, "true");
 
--- a/solr/core/src/java/org/apache/solr/core/DirectoryFactory.java
+++ b/solr/core/src/java/org/apache/solr/core/DirectoryFactory.java
@@ -20,13 +20,11 @@ import java.io.Closeable;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
-import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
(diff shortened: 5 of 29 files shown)


solr/core/src/java/org/apache/solr/core/CoreDescriptor.java
```diff
@@ -64,7 +63,7 @@ public class CoreDescriptor {
   public static final String SOLR_CORE_PROP_PREFIX = "solr.core.";
 
   public static final String DEFAULT_EXTERNAL_PROPERTIES_FILE =
-      "conf" + File.separator + "solrcore.properties";
+      Path.of("conf/solrcore.properties").toString();
```
epugh, line 66: is there any chance of different file seperators?   I wasn't sure if that was one reason for using the `File.separator`, like on windows?  Or is that one of the benefits of Path?

mlbiscoc: I was worried about this so started reading `Path.of()` [documentation](https://github.com/openjdk/jdk/blob/62a4544bb76aa339a8129f81d2527405a1b1e7e3/src/java.base/share/classes/java/nio/file/Path.java#L147) and [oracle documentation](https://docs.oracle.com/javase/tutorial/essential/io/pathOps.html#:~:text=Returns%20the%20string%20representation%20of%20the%20Path). `Filesystems.getDefault().getPath(String)` which should return the OS path and developers can use the standard `/` regardless. It's also a benefit of Path for Solr to no longer have to do all these File.separator chars or OS checking...

I haven't actually tested this on something like a windows operating system to confirm though...

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -1354,16 +1354,29 @@ public void initializeMetrics(SolrMetricsContext parentContext, String scope) {
...
+
+    try {
+      totalSpace = Files.getFileStore(dataDirFile).getTotalSpace();
+      useableSpace = Files.getFileStore(dataDirFile).getUsableSpace();
+    } catch (Exception e) {
+      // TODO Metrics used to default to 0 with java.io.File even if directory didn't exist
+      // Should throw an exception and initialize data directory before metrics
+      totalSpace = 0L;
+      useableSpace = 0L;
+    }
```
mlbiscoc, line 1370: This here was questionable. Upon a Solr core being initialized, metrics was trying to calculate and initialize a gauge for the `data/` directories before the directory was ever created. This always failed with `File` but it defaults to 0. `Path` on the other hand throws an `Exception`. This didn't seem right but also moving where the `initializeMetrics` is called in a bunch of different places had tests failing and I wasn't sure why so I kept it as TODO.

But also if this is initializing metrics for a `SolrCore` registry upon creation, shouldn't it always be 0 as the data directory was most likely just created?

mlbiscoc (commented):
I think this is ready for review

solr/api/src/java/org/apache/solr/client/api/endpoint/NodeFileStoreApis.java
```diff
@@ -63,5 +64,6 @@ SolrJerseyResponse getFile(
           String getFrom,
       @Parameter(description = "Indicates that (only) file metadata should be fetched")
           @QueryParam("meta")
-          Boolean meta);
+          Boolean meta)
+      throws IOException;
```
epugh, line 68: Just wanted to confirm this throw was needed?   Maybe because of something else furthur down?

solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
```diff
@@ -367,10 +366,7 @@ private void close(CacheValue val) {
   }
 
   private static boolean isSubPath(CacheValue cacheValue, CacheValue otherCacheValue) {
-    int one = cacheValue.path.lastIndexOf(File.separatorChar);
-    int two = otherCacheValue.path.lastIndexOf(File.separatorChar);
-
-    return otherCacheValue.path.startsWith(cacheValue.path + File.separatorChar) && two > one;
+    return Path.of(otherCacheValue.path).startsWith(Path.of(cacheValue.path));
```
epugh, line 369: love it

solr/core/src/java/org/apache/solr/core/CoreDescriptor.java
epugh: This is a really great benefit of the migration!

solr/core/src/java/org/apache/solr/core/DirectoryFactory.java
```diff
@@ -338,59 +335,66 @@ public String getDataHome(CoreDescriptor cd) throws IOException {
...
 
   public void cleanupOldIndexDirectories(
-      final String dataDirPath, final String currentIndexDirPath, boolean afterCoreReload) {
+      final String dataDirPath, final String currentIndexDirPath, boolean afterCoreReload)
+      throws IOException {
 
-    // TODO SOLR-8282 move to PATH
-    File dataDir = new File(dataDirPath);
-    if (!dataDir.isDirectory()) {
+    Path dataDirFile = Path.of(dataDirPath);
```
epugh, line 341: love seeing a TODO get TODONE'ed!

solr/core/src/java/org/apache/solr/core/SolrPaths.java
```diff
@@ -43,9 +42,7 @@ private SolrPaths() {} // don't create this
 
   /** Ensures a directory name always ends with a '/'. */
   public static String normalizeDir(String path) {
-    return (path != null && (!(path.endsWith("/") || path.endsWith("\\"))))
-        ? path + File.separator
-        : path;
+    return (path != null && (!(path.endsWith("/") || path.endsWith("\\")))) ? path + "/" : path;
```
epugh, line 45: wow that's gnarly..   Is there no existing `normalize()` function in any of our jars that would do this for us?   I guess fixing that is orthongonal to this effort....

solr/core/src/java/org/apache/solr/core/backup/BackupManager.java
```diff
@@ -343,7 +343,7 @@ private void downloadConfigToRepo(ConfigSetService configSetService, String conf
     // getAllConfigFiles always separates file paths with '/'
     for (String filePath : filePaths) {
       // Replace '/' to ensure that propre file is resolved for writing.
-      URI uri = repository.resolve(dir, filePath.replace('/', File.separatorChar));
+      URI uri = repository.resolve(dir, Path.of(filePath).toString());
```
epugh, line 346: could `repository.resolve()` take a path instead of a string?

solr/core/src/java/org/apache/solr/core/SolrResourceLoader.java
```diff
@@ -353,12 +353,14 @@ public ClassLoader getClassLoader() {
    */
   @Override
   public InputStream openResource(String resource) throws IOException {
-    if (resource.trim().startsWith("\\\\")) { // Always disallow UNC paths
-      throw new SolrResourceNotFoundException("Resource '" + resource + "' could not be loaded.");
+    String pathResource = Paths.get(resource).normalize().toString();
+    if (pathResource.trim().startsWith("\\\\")) { // Always disallow UNC paths
```
epugh, line 357: I saw we have a `SolrPaths.assertNotUnc(dir);`, could it be used here?

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -428,10 +428,10 @@ public boolean fetch(String path, String from) {
   @Override
   public void get(String path, Consumer<FileEntry> consumer, boolean fetchmissing)
       throws IOException {
-    File file = getRealpath(path).toFile();
-    String simpleName = file.getName();
+    Path file = getRealpath(path);
+    String simpleName = file.getFileName().toString();
     if (isMetaDataFile(simpleName)) {
-      try (InputStream is = new FileInputStream(file)) {
+      try (InputStream is = new FileInputStream(file.toString())) {
```
epugh, line 434: Could we ditch the `FileInputStream`?   I asked GPT, and it said:

```
InputStream inputStream = Files.newInputStream(path))
```

The recommended approach is using `Files.newInputStream()` because:
- It's more flexible (supports options)
- Works with any filesystem provider
- Better handles symbolic links
- Part of the modern NIO.2 API

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -458,25 +458,25 @@ public void syncToAllNodes(String path) throws IOException {
   }
 
   @Override
-  public List<FileDetails> list(String path, Predicate<String> predicate) {
-    File file = getRealpath(path).toFile();
+  public List<FileDetails> list(String path, Predicate<String> predicate) throws IOException {
+    Path file = getRealpath(path);
```
epugh, line 462: Can we rename this method to `getRealPath`?

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -458,25 +458,25 @@ public void syncToAllNodes(String path) throws IOException {
...
-          (dir, name) -> {
-            if (predicate == null || predicate.test(name)) {
-              if (!isMetaDataFile(name)) {
-                fileDetails.add(new FileInfo(path + "/" + name).getDetails());
+      try (Stream<Path> fileStream = Files.list(file)) {
+        fileStream.forEach(
+            (f) -> {
+              String fileName = f.getFileName().toString();
+              if (predicate == null || predicate.test(fileName)) {
+                if (!isMetaDataFile(fileName)) {
```
epugh, line 471: Would using Path instead of String here make sense for our Distrib File Store?  (Do Paths model distributed file system better than Files did??).

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -550,19 +550,19 @@ public void refresh(String path) {
...
   }
 
-  public static FileType _getFileType(File file) {
-    if (!file.exists()) return FileType.NOFILE;
-    if (file.isDirectory()) return FileType.DIRECTORY;
-    return isMetaDataFile(file.getName()) ? FileType.METADATA : FileType.FILE;
+  public static FileType _getFileType(Path file) {
+    if (!Files.exists(file)) return FileType.NOFILE;
+    if (Files.isDirectory(file)) return FileType.DIRECTORY;
+    return isMetaDataFile(file.getFileName().toString()) ? FileType.METADATA : FileType.FILE;
```
epugh, line 565: maybe another place where if `isMetaDAtaFile` takes a Path things are easier?

solr/core/src/java/org/apache/solr/filestore/FileStore.java
```diff
@@ -44,7 +44,7 @@ public interface FileStore {
   /** Fetch a resource from another node internal API */
   boolean fetch(String path, String from);
 
-  List<FileDetails> list(String path, Predicate<String> predicate);
+  List<FileDetails> list(String path, Predicate<String> predicate) throws IOException;
```
epugh, line 47: just confirming we DO need this throws.

solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
```diff
@@ -1061,35 +1060,37 @@ private void reloadCore() {
   }
 
   private void downloadConfFiles(
-      List<Map<String, Object>> confFilesToDownload, long latestGeneration) throws Exception {
+      List<Map<String, Object>> confFilesToDownload, long latestGeneration) {
     log.info("Starting download of configuration files from leader: {}", confFilesToDownload);
     confFilesDownloaded = Collections.synchronizedList(new ArrayList<>());
-    Path tmpConfPath =
+    Path tmpconfDir =
```
epugh, line 1066: maybe `tmpConfDir`?

solr/core/src/java/org/apache/solr/handler/admin/ShowFileRequestHandler.java
```diff
@@ -195,54 +195,69 @@ private void showFromZooKeeper(
...
-  private void showFromFileSystem(SolrQueryRequest req, SolrQueryResponse rsp) {
+  private void showFromFileSystem(SolrQueryRequest req, SolrQueryResponse rsp) throws IOException {
     Path admin = getAdminFileFromFileSystem(req, rsp, hiddenFiles);
 
     if (admin == null) { // exception already recorded
       return;
     }
-    // TODO SOLR-8282 move to PATH
-    File adminFile = admin.toFile();
+    Path adminFile = admin;
```
epugh, line 204: maybe just call `admin` `adminFile` in the first place?   Now that you fixed the todo?

solr/core/src/java/org/apache/solr/handler/admin/ShowFileRequestHandler.java
```diff
@@ -195,54 +195,69 @@ private void showFromZooKeeper(
...
-          fileInfo.add("directory", true);
-        } else {
-          // TODO? content type
-          fileInfo.add("size", f.length());
-        }
-        fileInfo.add("modified", new Date(f.lastModified()));
+      try (Stream<Path> directoryFiles = Files.list(adminFile)) {
+        directoryFiles.forEach(
+            (f) -> {
+              String path = f.getFileName().toString();
```
epugh, line 238: maybe move thid `path` variable creation down below the check in case we don't need to do it?

solr/core/src/java/org/apache/solr/handler/admin/ShowFileRequestHandler.java
```diff
@@ -195,54 +195,69 @@ private void showFromZooKeeper(
...
+              if (isHiddenFile(req, rsp, f.getFileName().toString(), false, hiddenFiles)) {
+                return;
+              }
+
+              SimpleOrderedMap<Object> fileInfo = new SimpleOrderedMap<>();
+              files.add(path, fileInfo);
+              if (Files.isDirectory(f)) {
+                fileInfo.add("directory", true);
+              } else {
+                // TODO? content type
```
epugh, line 249: lets. get rid of this todo..   this code hasn't changed in years, and if someone wants it they can add it later!  we shouldn't litter our code with todos that are mostly ideas..   a todo really implies hey, someone SHOULD go do this very soon!

solr/core/src/java/org/apache/solr/spelling/AbstractLuceneSpellChecker.java
```diff
@@ -84,7 +83,7 @@ public String init(NamedList<?> config, SolrCore core) {
     // If indexDir is relative then create index inside core.getDataDir()
     if (indexDir != null) {
       if (!Path.of(indexDir).isAbsolute()) {
-        indexDir = core.getDataDir() + File.separator + indexDir;
+        indexDir = Path.of(core.getDataDir(), indexDir).toString();
```
epugh, line 86: can indexDir be a path?

solr/core/src/java/org/apache/solr/spelling/suggest/fst/AnalyzingInfixLookupFactory.java
```diff
@@ -89,7 +88,7 @@ public Lookup create(NamedList<?> params, SolrCore core) {
     String indexPath =
         params.get(INDEX_PATH) != null ? params.get(INDEX_PATH).toString() : DEFAULT_INDEX_PATH;
     if (!Path.of(indexPath).isAbsolute()) {
-      indexPath = core.getDataDir() + File.separator + indexPath;
+      indexPath = Path.of(core.getDataDir(), indexPath).toString();
```
epugh, line 91: can indexPath be a Path?   Also, is there an opportunity to align variable naming between indexPath and the use of indexDir in in AbstractLuceneSpellChecker?

solr/core/src/java/org/apache/solr/util/VersionedFile.java
```diff
@@ -41,44 +41,47 @@ public class VersionedFile {
...
           oldFiles = new ArrayList<>();
-          for (int i = 0; i < names.length - 1; i++) {
-            oldFiles.add(new File(dir, names[i]));
+          for (int i = 0; i < fileList.size() - 1; i++) {
+            oldFiles.add(Path.of(dir.toString(), fileList.get(i).toString()));
           }
         }
 
-        is = new FileInputStream(f);
+        is = new FileInputStream(f.toString());
```
epugh, line 76: `Files.newInputStream()` instead?

solr/core/src/java/org/apache/solr/util/VersionedFile.java
```diff
@@ -41,44 +41,47 @@ public class VersionedFile {
...
       } catch (Exception e) {
         // swallow exception for now
       }
     }
 
     // allow exception to be thrown from the final try.
     if (is == null) {
-      is = new FileInputStream(f);
+      is = new FileInputStream(f.toString());
```
epugh, line 84: Files.newInputStream?

epugh (changes requested):
Really great progress here!   I've been poking a bit around the FIleStore stuff, so great to see some of the improvements in that area as well.    I put in a bunch of feedback, I hope it's useful.    Ping me when you are ready for final review!

solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
```diff
@@ -1484,28 +1484,27 @@ private void copyTmpConfFiles2Conf(Path tmpconfDir) {
...
-                ErrorCode.SERVER_ERROR, "Unable to mkdirs: " + backupFile.getParentFile());
+        try {
+          Path backupFile =
+              Path.of(
+                  oldPath
+                      + "."
+                      + getDateAsStr(new Date(Files.getLastModifiedTime(oldPath).toMillis())));
+          if (!Files.exists(backupFile.getParent())) {
+            Files.createDirectories(backupFile.getParent());
           }
```
dsmiley, line 1503: No need to check if doesn't exist; an advantage of Files.createDirectories

solr/core/src/java/org/apache/solr/handler/admin/CoreAdminOperation.java
```diff
@@ -357,7 +356,7 @@ public static NamedList<Object> getCoreStatus(
           if (core != null) {
             info.add(NAME, core.getName());
             info.add("instanceDir", core.getInstancePath().toString());
-            info.add("dataDir", normalizePath(core.getDataDir()));
+            info.add("dataDir", Path.of(core.getDataDir()).toString());
```
dsmiley, line 359: I'm suspicious of this technique.  I'd prefer we have a utility method to make certain normalizations when they are needed.

solr/core/src/java/org/apache/solr/handler/admin/ShowFileRequestHandler.java
```diff
@@ -195,54 +195,69 @@ private void showFromZooKeeper(
...
-        files.add(path, fileInfo);
-        if (f.isDirectory()) {
-          fileInfo.add("directory", true);
-        } else {
-          // TODO? content type
-          fileInfo.add("size", f.length());
-        }
-        fileInfo.add("modified", new Date(f.lastModified()));
+      try (Stream<Path> directoryFiles = Files.list(adminFile)) {
+        directoryFiles.forEach(
```
dsmiley, line 236: no try-with-resources needed

solr/core/src/java/org/apache/solr/handler/admin/ShowFileRequestHandler.java
```diff
@@ -195,54 +195,69 @@ private void showFromZooKeeper(
...
 
     // Show a directory listing
-    if (adminFile.isDirectory()) {
+    if (Files.isDirectory(adminFile)) {
       // it's really a directory, just go for it.
-      int basePath = adminFile.getAbsolutePath().length() + 1;
       NamedList<SimpleOrderedMap<Object>> files = new SimpleOrderedMap<>();
-      for (File f : adminFile.listFiles()) {
-        String path = f.getAbsolutePath().substring(basePath);
-        path = path.replace('\\', '/'); // normalize slashes
```
dsmiley, line 231: don't lose this.  I feel like we need a toForwardSlashPathString(Path) utility method.

dsmiley (commented):
I got half way through review today.

mlbiscoc (commented):
Went through all the comments and changed them as best I could. Had some questions on a few which I replied back, but overall wasn't too bad. Tests pass locally again but ready for another round of review.

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
epugh: cool

epugh (approved):
LGTM pending the tests... I just kicked precommit and solr tests off on github.   

I do hope you have the energy to do this on tests too!

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
dsmiley: I don't recommend doing that now.  Path *can* be used for systems that aren't _exactly_ a file system but probably requires a custom FileSystemProvider which is a lot of work.  We have a number of APIs that have a more abstract notion of a file system; they use String today (instead of File/Path) for this reason.

dsmiley:
As a general comment, I want to ensure we're careful converting some Strings to Paths.  It's only safe if we're certain that it's exactly a file system path.  Paths resolved inside a DirectoryFactory/Directory are not necessarily literally on the file system (it's an abstraction to data that could be in HDFS or S3 or...).  It'd be nice to use a Path for this anyway but shouldn't be without a custom FileSystem impl (very out of scope here).  Same for ZK paths.  Also some of our HTTP APIs are file-system inspired (ConfigSet, FileStore, ...) with paths in which '/' is the separator no matter what the underlying file system is Solr is running on (after all; don't want the client to care about the OS Solr is running on).  These are not *directly* convertible to a Path either, or at least take some work to do separator translation if the end location is on-disk (file store & file system ConfigSetService).

solr/core/src/java/org/apache/solr/cloud/ZkSolrResourceLoader.java
```diff
@@ -121,7 +120,7 @@ public InputStream openResource(String resource) throws IOException {
 
     try {
       // delegate to the class loader (looking into $INSTANCE_DIR/lib jars)
-      is = classLoader.getResourceAsStream(resource.replace(File.separatorChar, '/'));
+      is = classLoader.getResourceAsStream(Path.of(resource).toString());
```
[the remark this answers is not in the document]
dsmiley: But I think it doesn't work; not on Windows.  I thought I commented elsewhere in that lengthy review that all `Path.of(resource).toString()` that you are adding is probably flawed

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -80,19 +80,20 @@ public DistribFileStore(CoreContainer coreContainer) {
...
-    if (File.separatorChar == '\\') {
-      path = path.replace('/', File.separatorChar);
-    }
-    SolrPaths.assertNotUnc(Path.of(path));
-    while (path.startsWith(File.separator)) { // Trim all leading slashes
+    Path dir = Path.of(path);
+    SolrPaths.assertNotUnc(dir);
+
+    // Strip the path of from being absolute to become relative to resolve with SolrHome
+    while (path.startsWith("/")) {
```
dsmiley, line 92: I was looking through Path (and some chats with LLMs) and I think I know how to do this in a more canonical manner.  Given a Path that might be absolute, call getRoot() on it.  If null, it's relative.  Otherwise, call root.relativize(path);

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -245,14 +247,14 @@ boolean fetchFromAnyNode() {
...
     }
 
     @SuppressWarnings("unchecked")
     MetaData readMetaData() throws IOException {
-      File file = getRealpath(getMetaPath()).toFile();
-      if (file.exists()) {
-        try (InputStream fis = new FileInputStream(file)) {
+      Path file = getRealPath(getMetaPath());
+      if (Files.exists(file)) {
+        try (InputStream fis = new FileInputStream(file.toString())) {
```
dsmiley, line 257: Lets just use Files.newInputStream here

solr/core/src/java/org/apache/solr/handler/admin/CoreAdminOperation.java
dsmiley: can you bring the method back please?  In a file utility place, and documenting what sort of normalization it does (otherwise is ambiguous).  Like; normalizeToOsPathSeparator(...)

solr/core/src/java/org/apache/solr/handler/admin/ShowFileRequestHandler.java
dsmiley: Okay, I agree that the ShowFileRequestHandler ought not to be receiving backslashes.

dsmiley (commented):
I reviewed the whole thing again.

epugh:
Thank you @dsmiley for digging through things.  Your comments and eyes is probably saving us a lot of test failures down the road ;-).

solr/core/src/java/org/apache/solr/handler/admin/CoreAdminOperation.java
mlbiscoc: Found a place in `FileUtils` and brought it back. Also just switched to `FileSystems.getDefault().getSeparator()` and brought all the logic to be careful of regressions since it seems neither of us can confirm fully on a windows machine.

solr/core/src/java/org/apache/solr/core/DirectoryFactory.java
```diff
@@ -369,15 +370,17 @@ public void cleanupOldIndexDirectories(
...
+        log.info("Will not remove most recent old directory after reload {}", dirsList.getFirst());
       i = 1;
     }
 
-    log.info(
-        "Found {} old index directories to clean-up under {} afterReload={}",
-        dirsList.size() - i,
-        dataDirPath,
-        afterCoreReload);
+    if (log.isInfoEnabled())
```
dsmiley, line 378: why did you add this if log.isInfoEnabled check?  I confess I hate them.  You may see them commonly in the codebase but it's only in certain scenarios that a code checker will complains about it (`precommit` will complain if so); there's no universal requirement or style.  If it didn't complain, please leave it be.

dsmiley (approved):
Great!  You can update CHANGES.txt for the existing SOLR-16903 to combine it.  I think as other PRs happen, it'll _ultimately_ be a broad File -> Path with a list of contributors.

solr/core/src/java/org/apache/solr/core/DirectoryFactory.java
mlbiscoc: Precommit was complaining because of the `getFirst()` call in the log which was why I added it. I also don't like it either.

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -80,19 +80,21 @@ public DistribFileStore(CoreContainer coreContainer) {
...
-    while (path.startsWith(File.separator)) { // Trim all leading slashes
-      path = path.substring(1);
+  private static Path _getRealPath(String dir, Path solrHome) {
+    Path path = Path.of(dir);
+    SolrPaths.assertNotUnc(path);
+
+    if (path.isAbsolute()) {
+      // Strip the path of from being absolute to become relative to resolve with SolrHome
+      path = path.subpath(0, path.getNameCount());
     }
```
HoustonPutman, line 94: This is broken on Windows. A number of tests fail because it's not stripping leading "\" characters. Not sure if we should revert to the previous logic, or do it another way.

I made https://issues.apache.org/jira/browse/SOLR-17654, but it seems this is a new thing, so we can probably close that.

@mlbiscoc

mlbiscoc: Thanks for catching Houston. I can make a PR to revert that as the safest thing but I don't have a windows machine on hand to easily test this.

mlbiscoc: [PR](https://github.com/apache/solr/pull/3160) to bring back that old logic.

PR #2925, opened 2024-12-23
SOLR-17602: Per-Module Dependency Locking

malliaridis:
https://issues.apache.org/jira/browse/SOLR-17602

# Description

To improve the developer experience of the dependency management, we are looking for solutions to make the dependency lock files more developer-friendly / readable.

# Solution

This solution is based on https://github.com/apache/solr/pull/2915 which introduces a platform module for aligning the dependency version across our project via constraints.

This PR removes the carrotsearch dependencycheck plugin and falls back to the default gradle tasks for generating lockfiles. It defines a new task that can be used via `gradlew resoslveAndLockAll --write-locks`. This task generates or updates existing `gradle.lockfile` in each module.

Additional gradle configuration has been added to lock all module configurations by default. The task `resolveAndLockAll` will then filter and lock only the resolveable dependencies.

# Additional Notes

When generating lockfiles under Windows, the lockfiles may use CRLF instead of LF and display difference in IDE. If you have configured your Git correctly, you may ignore the changes.

Additionally, the caching in CI/CD uses `versions.lock` for index, which with this PR would not exist anymore. Since this PR uses per-module dependency lockfiles, this commit updates the workflows to use all lockfiles as keys.

I have also noticed that when running `./gradlew assemble` the lockfiles are ignored for some reason and different libraries are included. Adding the `platform` module explicitly via `implementation platform(project(":platform")` solves the issue. So it may be necessary to include the platfform module in all modules and submodules explicitly to ensure the consistency in the future.

# Tests

Effects tested with lucene version update in `main` while this PR was open.

# Checklist


malliaridis:
@dsmiley does this go to the right direction in regards to your ticket?

dsmiley (approved):
> does this go to the right direction in regards to your ticket?

Yes but definitely doesn't solve them.  I like that this PR makes it easy to know if a dependency is on a certain configuration/classpath without having to google the gradle command to accomplish the same.  And if we change the dependencies, we might notice unanticipated changes.  

The error-prone aspect is unfortunate.  I wonder if we're disabling error-prone by default in the wrong way.

+1 to this PR.

janhoy:
+1 More gradle-default behavior is better, if we can make it jive with our project requirements.. Especially when the plugin in use (carrot) is not that actively maintained.

dev-docs/dependency-upgrades.adoc
```diff
@@ -30,43 +30,40 @@ In order to upgrade a dependency, you need to run through a number of steps:
...
+When it comes to security vulnerabilities that are found in direct or transitive dependencies, the recommended way to
+address them is to update the specific library if there is a new release that solves this issue. For both direct and
+transitive dependencies, we simply have to update the version as described above.
 
-== Renovate bot Pull Requests
+In case it is a transitive dependency that is not directly used, you can simply add it to `libs.versions.toml` as you
+would with any other dependency. The dependency resolution approach defined in `:platform` will handle the rest.
+Don't forget to add a `# @keep` note with a reference to the vulnerable version and CVE that is fixed with the explicit
+definition of the library and new version. This way it is easier to keep track of unreferenced dependencies in our
+libraries toml file, and we can clean them up once the libraries using the modules are updated.
```
dsmiley, line 60: Ooooh, now I see what this is for.

solr/modules/sql/build.gradle
```diff
@@ -20,6 +20,7 @@ apply plugin: 'java-library'
 description = 'SQL Module'
 
 dependencies {
+  implementation platform(project(':platform'))
```
dsmiley, line 23: The explicit-ness of this is good, I think, even if it means an additional line everywhere.  Maybe another different approach would be for the build to insert these automatically.  Either is fine.

dsmiley (approved):

PR #2927, opened 2024-12-28
SOLR-17611 SolrJ User-Agent, pass Solr version

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17611 (read this for context)

No tests.  I suppose the best test would be something in the smoke test which is at release time where we definitely have the "specification version" in the JAR manifest.  Once we have something that detects the version to behave differently, we will need it to work in a unit test environment too, where we want the latest behavior but don't have a spec version in a manifest.
Maybe the PR should include something in SolrVersion to parse it as well?


jdyer1:
There are asserts in the unit tests for the user-agent, so changing it will cause failures I think.  See `HttpSolrClientTestBase#expectedUserAgent`.  If we put the specification version in a `protected static` variable, the unit test can easily override it, which might be adequate enough of a unit testing strategy?  I was looking around and maybe ByteBuddy's `PackageDefinitionStrategy` could be used to mock this?  Even if possible (I have no idea) I guess that would be more effort than it is worth!  I agree something in the smoke tester would be the best way to verify correctness.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -106,11 +106,15 @@
  * </ul>
  */
 public class Http2SolrClient extends HttpSolrClientBase {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
   public static final String REQ_PRINCIPAL_KEY = "solr-req-principal";
+  private static final String USER_AGENT =
+      "Solr["
```
jdyer1, line 112: Could this go in the superclass so it doesn't have to exist twice?

dsmiley: I'd hate to create constants merely for the parts of this string that we concatenate -- I hate that coding practice, to be blunt.  I could see a utility method that composes them.  But any way it's tested so I'm not sure what the risk/concern of _not_ doing what you suggest is.  This is a one-liner BTW.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpJdkSolrClient.java
```diff
@@ -73,7 +74,7 @@ public class HttpJdkSolrClient extends HttpSolrClientBase {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   private static final String USER_AGENT =
-      "Solr[" + MethodHandles.lookup().lookupClass().getName() + "] 1.0";
+      "Solr[" + MethodHandles.lookup().lookupClass().getName() + "] " + SolrVersion.LATEST_STRING;
```
madrob, line 77: It's ok to lose the distinction in version number because we still have v1 and v2 clients in the class names?

solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
```diff
@@ -1170,6 +1171,21 @@ protected Object _getHandler() {
...
   }
 
+  /** Gets the client (user-agent) SolrJ version, or null if isn't SolrJ. */
+  public SolrVersion getUserAgentSolrVersion() {
+    String header = req.getHeader("User-Agent");
+    if (header == null || !header.startsWith("Solr")) {
+      return null;
+    }
+    try {
+      return SolrVersion.valueOf(header.substring(header.lastIndexOf(' ') + 1));
```
madrob, line 1181: If this gets an older client it will return 1.0 or 2.0, which can be somewhat misleading.

dsmiley: Yeah; I thought it would have no consequence so I did nothing about it.  At least by returning a SolrVersion instead of null, we communicate indirectly that it's at least SolrJ.  I could add a javadoc comment on 1.0 and 2.0 being used prior to 9.8.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpJdkSolrClient.java
dsmiley: Yes.

solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
dsmiley: I added this to the docs.

dsmiley:
Ready to merge; will do within a couple days if I hear nothing.

solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
dsmiley: BTW that method has no callers; isn't tested.  I did manually test it.  Obviously I intend to have it be used soon.

madrob (approved):

PR #2930, opened 2025-01-01
SOLR-16391: Convert "modify-coll" API to JAX-RS

gerlowskija:

# Description

Solr's v2 API supports a "modify-collection" API, but its implementation uses the legacy (non JAX-RS) framework.

# Solution

This PR migrates the "modify-collection" API to a JAX-RS implementation.  This ensures that the APIs are covered by our v2 API OAS, and includes them in all code-generation done using the OAS.

It also makes slight tweaks to the appearance of the API to help it fit in a bit more with the current API design. In particular:

- `PUT` is now used as the HTTP verb,
- the "command" property (i.e. "modify") and additional request body nesting have been removed.
- an "unset" property has been added to the request body (see comment below).
 
# Tests

Updates to V2CollectionsAPIMappingTest.  Tweaks to various other "modify coll" related tests.

# Checklist


dsmiley:
Perhaps HTTP DELETE for deleting them one at a time and we see if this is good enough for folks?

epugh:
> Perhaps HTTP DELETE for deleting them one at a time and we see if this is good enough for folks?

This is what I was thinking as well!

solr/api/src/java/org/apache/solr/client/api/endpoint/CollectionApis.java
```diff
@@ -0,0 +1,82 @@
...
+    @Operation(
+        summary = "Creates a new SolrCloud collection.",
+        tags = {"collections"})
+    SubResponseAccumulatingJerseyResponse createCollection(CreateCollectionRequestBody requestBody)
+        throws Exception;
+  }
+
+  @Path("/collections/{collectionName}")
+  interface Modify {
+    @PUT
```
dsmiley, line 58: BTW I'm suspicious of your proposed use of HTTP **PUT** as the payload isn't an entire representation of the target resource.  It modifies an existing one that may have characteristics specified on creation that are not here.  PUT doesn't have much applicability.

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT "**replaces** a representation of the target resource with the request content"
[1] https://restfulapi.net/rest-put-vs-post/ "PUT **replaces** the resource in its entirety"

gerlowskija: I've proposed using the `PATCH` verb for cases like this - seemingly it's made for cases like this.  But I've gotten feedback in the past that it's unintuitive due to its obscurity.  Wdyt?

epugh: I've used PATCH in other contexts.    I *think* that if we embrace PATCH only here, then yeah, it will be kind of obsurcure.   But if we use it on multiple APIs, then I think it's perfectly fine.

dsmiley: [PATCH](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH) makes total sense to me.  I don't care about relative obscurity.

gerlowskija: > I think that if we embrace PATCH only here, then yeah, it will be kind of obsurcure. But if we use it on multiple APIs, then I think it's perfectly fine.

That seems like a reasonable line to walk here.  Anyone have other APIs in mind that use similar partial-update semantics?  If so, I'll switch this to 'PATCH'.  Otherwise I'll stick with 'PUT' for now.

gerlowskija:
> Perhaps HTTP DELETE for deleting them one at a time and we see if this is good enough for folks?

OK, good call - I'll pursue that route for now.

gerlowskija:
Ugh, actually, using an HTTP DELETE would conflict with existing paths.

- `DELETE /api/collections/someCollName/properties/somePropName` is my first choice, but that conflicts with an existing API for deleting collection-properties.
- `DELETE /api/collections/someCollName/somePropName` was my next thought, but this also conflicts since "properties" is a valid prop name ("modify" allows users to specify arbitrary props)

We could come up with some new path-segment to differentiate "modify" properties from "collection props", say, `DELETE /api/collections/someCollName/collStateProperties/somePropName`, but that feels hacky? idk.

Personally, I'm worried the API design issues highlight deeper confusion in the underlying functionality.  Why do we support arbitrary properties in "modify" that are somehow different from "collectionprops".  When would a user choose to use one or the other?  I don't see an answer in our docs, and I don't know myself 😕 )

HoustonPutman:
> Personally, I'm worried the API design issues highlight deeper confusion in the underlying functionality. Why do we support arbitrary properties in "modify" that are somehow different from "collectionprops". When would a user choose to use one or the other? I don't see an answer in our docs, and I don't know myself 😕 )

Yeah, this is very confusing to me. If the two have to be separate, maybe metadata is a better name for it, and we go with "Collection Properties" and "Collection Metadata", though I do not think the two are as separate as we would like them to be.

gerlowskija:
I like "metadata" from a naming perspective.

> I do not think the two are as separate as we would like them to be.

Yeah, agreed.  Worth some digging there to see what the original intention was, and whether it's worth maintaining both options (at least in terms of the arbitrary/user-specified "modify" properties).

dsmiley:
+1 on "metadata" vs "properties" for nomenclature.

I was thinking about the earlier conundrum a bit.  Let's just do PATCH to `/api/collections/myCollName` to update anything updatable.  To unset a value, why not just use null?

gerlowskija:
> Let's just do PATCH to /api/collections/myCollName to update anything updatable.

We discussed PATCH a bit [here](https://github.com/apache/solr/pull/2930#discussion_r1915243956), and the slightly preferred option seemed to be to utilize PATCH only if we can find other APIs where it makes sense as well.  Otherwise, it'd be the only place the HTTP method is used by Solr, and slightly unintuitive for that reason.

I don't care strongly and am happy to flip-flop if other folks chime in for PATCH, but I think PUT has more support at the moment (I'm counting [epugh](https://github.com/apache/solr/pull/2930#discussion_r1907133948) and [hputman](https://docs.google.com/spreadsheets/d/1HAoBBFPpSiT8mJmgNZKkZAPwfCfPvlc08m5jz3fQBpA/edit?disco=AAAAbSSbh4A)).

> to unset a value, why not just use null?

As the PR exists today, with a POJO representing the request body, Jackson has no way to differentiate between fields that were absent from the request body entirely, and fields that were present with a null-value.  In other words: if you're cramming them into a POJO at deserialization time, `{"foo": null}` and `{}` are indistinguishable on the server side.

That said, I guess we don't strictly need a POJO.  I _think_ we'd be able to distinguish these cases if we had Jackson deserialize into a Map instead.  It'd make our OAS and codegen artifacts less specific (and therefore less useful) for this API, but maybe we're OK with that if it means avoiding the need for a `DELETE /api/collections/collName/metadata/replicationFactor`.

If folks are OK dropping the request-body POJO, I can investigate going this route?

dsmiley:
If Jackson can't handle it then I think it's better to drop the POJO so we can have a better HTTP API.
I'll look for more PATCH opportunities but it's not clear to me what problem there would be if only one API used PATCH.  It has zero consequences for any caller.

gerlowskija:
> It has zero consequences for any caller.

It's not really about consequences, it's about being consistent/intuitive.  Even users who know about the `PATCH` verb and its semantics will probably reach for `PUT` if they're already used to using it on update-field, and update-config, and update-collprop, and update-configset-file, and update-clusterprop, and update-filestore-entry, and update-clusterprop, and update-noderole, and so on.

But like I said - this isn't a hill I'm willing to die on: if anyone else votes for PATCH to tilt the cumulative consensus, or we find another API where it'd work, I'm happy to go that route.

In the meantime I'm going to try out the "raw-Map" approach and follow up on SOLR-13271 regarding whether we need to support the arbitrary properties now that we have "Collection Props" as a more formalized thing.

PR #2932, opened 2025-01-03
reusing empty NamedList rather than recreating a new empty NamedList …

renatoh:
For every call with face=true which is not providing any facet.fields, we do create an empty NamedList.
Having a single instance of an empty NamedList feels like the more efficient approach to me.

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
+++ b/solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
@@ -72,39 +72,20 @@ public class SimpleOrderedMap<T> extends NamedList<T> {
 
   /**
    * Returns a shared, empty, and immutable instance of SimpleOrderedMap.
-   * @return empty SimpleOrderedMap (immutable) 
+   * @return Empty SimpleOrderedMap (immutable) 
    */
   public static SimpleOrderedMap<Object> of() {
     return EMPTY;
   }
 
   /**
-   * Returns an immutable instance of SimpleOrderedMap with a single element.
-   * @return List containing the elements
+   * Returns an immutable instance of SimpleOrderedMap with a single key-value pair.
+   * @return SimpleOrderedMap containing one key-value pair
    */
-  public static SimpleOrderedMap<Object> of(Object o1) {
-    return new SimpleOrderedMap<>(List.of(o1));
-  }
-  
-  /**
-   * Returns an immutable instance of SimpleOrderedMap with two elements.
-   * @return List containing the elements
-   */
-  public static SimpleOrderedMap<Object> of(Object o1, Object o2) {
-    return new SimpleOrderedMap<>(List.of(o1,o2));
-  }
-  
-  /**
-   * Returns an immutable instance of SimpleOrderedMap with an arbitrary number of elements. 
-   * @return List containing the elements
-   */
-  public static SimpleOrderedMap<Object> of(Object... elements) {
-	  return switch (elements.length) {
-		  case 0 -> of();
-		  case 1 -> of(elements[0]);
-		  case 2 -> of(elements[0], elements[1]);
-		  default -> new SimpleOrderedMap<>(List.of(elements));
-	  };
+
+  public static <T> SimpleOrderedMap<T> of(String name, T val)
+  {
+    return new SimpleOrderedMap<>(List.of(name, val));   
   }
   
 }
--- a/solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
+++ b/solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
@@ -17,6 +17,7 @@
 package org.apache.solr.common.util;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
@@ -35,7 +36,7 @@ import java.util.Map;
  */
 public class SimpleOrderedMap<T> extends NamedList<T> {
  
-  private static final SimpleOrderedMap<Object> emptySimpleOrderedMap = new SimpleOrderedMap<>(0);
+  private static final SimpleOrderedMap<Object> emptySimpleOrderedMap = new SimpleOrderedMap<>(Collections.emptyList());
   /** Creates an empty instance */
   public SimpleOrderedMap() {
     super();
--- a/solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
+++ b/solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
@@ -74,7 +74,37 @@ public class SimpleOrderedMap<T> extends NamedList<T> {
    * Returns a shared, empty, and immutable instance of SimpleOrderedMap.
    * @return empty SimpleOrderedMap (immutable) 
    */
-  public static SimpleOrderedMap<Object> empty() {
+  public static SimpleOrderedMap<Object> of() {
     return EMPTY;
   }
+
+  /**
+   * Returns an immutable instance of SimpleOrderedMap with a single element.
+   * @return List containing the elements
+   */
+  public static SimpleOrderedMap<Object> of(Object o1) {
+    return new SimpleOrderedMap<>(List.of(o1));
+  }
+  
+  /**
+   * Returns an immutable instance of SimpleOrderedMap with two elements.
+   * @return List containing the elements
+   */
+  public static SimpleOrderedMap<Object> of(Object o1, Object o2) {
+    return new SimpleOrderedMap<>(List.of(o1,o2));
+  }
+  
+  /**
+   * Returns an immutable instance of SimpleOrderedMap with an arbitrary number of elements. 
+   * @return List containing the elements
+   */
+  public static SimpleOrderedMap<Object> of(Object... elements) {
+	  return switch (elements.length) {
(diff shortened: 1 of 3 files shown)


renatoh:
I've investigated why the tests are now failing.
The issue is that  the method org.apache.solr.request.SimpleFacets#getFacetFieldCounts is now returning an empty NamedList, and not its subtype SimpleOrderedMap anymore. This seems fine since the return type of that method is NamedList.
The problem is that returning NamedList causes that in the response-json, the facet_counts/facet_fields will become a list:
Screenshot 2025-01-03 at 18 31 15
The method org.apache.solr.common.util.JsonTextWriter#writeNamedList distinguish between these types and writes a different data structure.

I think the root cause is the method org.apache.solr.request.SimpleFacets#getFacetFieldCounts allowing to return the super-type NamedList.
Shall I change the return type of that method to SimpleOrderedMap?

madrob:
@dsmiley this looks related to your named list vs simple map thread on dev list

renatoh:
@madrob do you have a link to the mention thread?

madrob:
https://lists.apache.org/thread/gllkpo16wotmkq0ppsq578yn7dhkyo9q

dsmiley:
My general opinion here is wondering it it worth bothering with the micro optimization of using a shared instance.  And more importantly, there is no immutable NamedList so I'm uncomfortable with introducing a shared mutable instance.

renatoh:
@dsmiley I do get your point that a shared instance should be immutable. I could instantiate the shared empty instances with an immutable list, similar to what we do here:
org.apache.solr.common.util.NamedList#getImmutableCopy.

For my understanding: What is the purpose of SimpleOrderedMap, it is just to control the output format of the response writers?

dsmiley:
Yes, it's about the response format but moreover it communicates that keys do not (well technically "should" not) repeat.  It's tempting to actually implement Map on this thing.

renatoh:
Not being very familiar with the Solr source code, this construct with NamedList and SimpleOrderedMap is pretty confusing to me.
From what I understand NamedList is basically a List of Pair<Key,Value> (Ordered and keys can repeat)  and SimpleOrderedMap is an HashMap or  LinkedHashMap (not necessarily ordered but with unique) . Is that understanding correct?

Back in 2006 NamedList had the following todo on it:
 * :TODO: In the future, it would be nice if this extended Map or Collection,
 * had iterators, used java5 generics, had a faster lookup for
 * large lists, etc...
 * It could also have an interface, and multiple implementations.
 * One might have indexed lookup, one might not.

But this one got removed at one point

dsmiley:
NamedList and thus SimpleOrderedMap is ordered (insertion order).  HashMap is called HashMap because it explains its implementation via use of the hashCode() and thus get(key) is `O(1)`.  NamedList and thus SimpleOrderedMap doesn't use a hashCode so I would say it's *not* specifically like a HashMap but I would indeed say that SimpleOrderedMap is a Map even though it doesn't implement the interface.  It has Map in its name, after all.  I recall there was concern that implementing Map would lead to people passing this off to code that makes assumptions on Map having better than `O(N)` lookup by key (even TreeMap is `O(Log N)`) but NamedList is `O(N)` leading to a performance problem for non-trivial sized NamedLists.

renatoh:
Thanks for the explanation. 
Would another  option be to use a Map internally for SimpleOrderedMap without changing the interface, that lead to better lookup performance without the need of touching a lot of code?

dsmiley:
Adding an additional HashMap inside SimpleOrderedMap misses the point of why NamedList exists.  It's for something super lean/fast that's only for transferring stuff (i.e. to be serialized), navigated by iteration once.  For that use-case, a hashCode system is a waste.  But Solr uses it for configuration as well, a questionable choice that's not in line with NamedList's advantages.  Perhaps a radical point of view is that NamedList shouldn't even have "get" methods, leaving only iteration, thus emphasizing its purpose.  For sure, Solr over-uses NamedList.  And it's [hard to stop using it](https://github.com/apache/solr/pull/2916#pullrequestreview-2518454057) because of SolrJ "javabin" compatibility.

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -34,6 +35,8 @@
  * serialized. It aims to minimize overhead and to be efficient at adding new elements.
  */
 public class SimpleOrderedMap<T> extends NamedList<T> {
+ 
+  private static final SimpleOrderedMap<Object> emptySimpleOrderedMap = new SimpleOrderedMap<>(Collections.emptyList());
```
dsmiley, line 39: call it EMPTY ?

dsmiley: Also nowadays we just do `List.of()` which is equivalent, shorter, and doesn't import another class.

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -67,4 +70,8 @@ public SimpleOrderedMap<T> clone() {
     newList.addAll(nvPairs);
     return new SimpleOrderedMap<>(newList);
   }
+  
+  public static SimpleOrderedMap<Object> emptySimpleOrderedMap() {
```
dsmiley, line 74: Needs a javadoc.  And that name is quite redundant given the type where we're declaring this.
WDYT of naming this simply `of` thus mimicking the end effect of `List.of()` or `Set.of()` or `Map.of()`?

renatoh: The method name  of() is implying, at least to me, that something can be passed in and and creates something else out of it. Like when you do List.of("String"). Maybe empty() would be the better option. What do you thing?

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
renatoh: " and doesn't import another class." Never thought of that advantages of List.of over Collections.empty(). Good point, will adjust it!

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
dsmiley: Using `of` opens the door for some more static methods for more, which you could add now if it makes `of` more attractive to you.  Just a single pair would be fine and perhaps useful.  I look to the JDK for inspiration on data structure conventions.

renatoh: ok, I see you point regarding of() vs empty(). I looked at how it is done on the List and have implemented some more of()-method for 1,2 and n-elements. what do you think?

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -67,4 +69,42 @@ public SimpleOrderedMap<T> clone() {
...
+   */
+  public static SimpleOrderedMap<Object> of(Object o1, Object o2) {
+    return new SimpleOrderedMap<>(List.of(o1,o2));
+  }
+  
+  /**
+   * Returns an immutable instance of SimpleOrderedMap with an arbitrary number of elements. 
+   * @return List containing the elements
+   */
+  public static SimpleOrderedMap<Object> of(Object... elements) {
```
dsmiley, line 101: Map.of has no similar array based argument so I suggest not doing the same here.  Presumably the key and value distinction then becomes less obvious; easier to misuse.

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -67,4 +69,42 @@ public SimpleOrderedMap<T> clone() {
...
+   */
+  public static SimpleOrderedMap<Object> of(Object o1) {
+    return new SimpleOrderedMap<>(List.of(o1));
+  }
+  
+  /**
+   * Returns an immutable instance of SimpleOrderedMap with two elements.
+   * @return List containing the elements
+   */
+  public static SimpleOrderedMap<Object> of(Object o1, Object o2) {
```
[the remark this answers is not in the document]
renatoh: of course, was confused by the fact it internally holds a List t store the values.
Do you think it still makes sense to have and of for two key-value pairs e.g. like this:
of(String name1, Object val1, String name2, Object val2)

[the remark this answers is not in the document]
dsmiley: Basically, if Map has it, we can to.
Any way, I also think we shouldn't over-invest in SimpleOrderedMap without callers of these methods.

dsmiley (approved):

dsmiley:
Also merged to branch_9x thus will appear in Solr 9.9.
Thanks for contributing!

dsmiley:
If you're interested: https://issues.apache.org/jira/browse/SOLR-17623

renatoh:
sure, i will give it a go.
so the idea is to keep the ArrayList from NamedList as internal data structure but do i implement java.util.Map, correct?
How do I get access to Jira?

dsmiley:
Yes; simply add methods to SimpleOrderedMap.

JIRA: Should be in the header of the page: https://selfserve.apache.org/jira-account.html

PR #2935, opened 2025-01-03
SOLR-17519: CloudSolrClient with HTTP ClusterState can forget live nodes and then fail

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-17519

# Description

In `BaseHttpClusterStateProvider` if all initially passed nodes except for 1 go down, then CSP will only be aware of the 1 live node. If that node were to go down and any of the other initially passed nodes were to recover, CSP would not be able to connect to get latest cluster state of live nodes because it only holds the address of the now downed node.

# Solution

CSP holds immutable `initialNodes` which is used to never remove this set from live nodes to keep it resilient. Now if the case above were to occur, CSP would still be able to get cluster state and latest set of live nodes because live nodes always has the initial set. Live nodes can also hold new nodes that are added to the cluster after CSP initialization but those are removable unlike the initial nodes.

# Tests

`testClusterStateProviderDownedInitialLiveNodes`  tests the above test case and used to fail.
`testClusterStateProviderLiveNodesWithNewHost` tests live nodes with a 3rd added new host after CSP is initialized which is removable but initial nodes are always still present to be reachable by CSP.

# Checklist


solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -51,7 +52,9 @@ public abstract class BaseHttpClusterStateProvider implements ClusterStateProvid
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   private String urlScheme;
+  private Set<String> initialNodes;
   volatile Set<String> liveNodes;
+  volatile Set<String> knownNodes;
```
dsmiley, line 57: right away, I'm surprised.  I can understand needing two sets, but 3?  IMO knownNodes doesn't need to exists; that's the same as liveNodes.  I see below you've changed many references from liveNodes to knownNodes but I recommend against doing that because the exact wording of "liveNodes" is extremely pervasive in SolrCloud (well known concept) so let's not have a vary it in just this class or any class.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -65,6 +68,8 @@ public void init(List<String> solrUrls) throws Exception {
       urlScheme = solrUrl.startsWith("https") ? "https" : "http";
       try (SolrClient initialClient = getSolrClient(solrUrl)) {
         this.liveNodes = fetchLiveNodes(initialClient);
+        this.initialNodes = Set.copyOf(liveNodes);
```
dsmiley, line 71: I think the idea of this JIRA issue is that we'll take the Solr URLs as configured and use *this* is the initial / backup liveNodes.  I think this is a very simple idea to to document/understand/implement.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/ClusterStateProviderTest.java
```diff
@@ -34,15 +35,20 @@
...
+import org.apache.solr.embedded.JettySolrRunner;
 import org.hamcrest.Matchers;
+import org.junit.After;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
 public class ClusterStateProviderTest extends SolrCloudTestCase {
 
+  private static JettySolrRunner jettyNode1;
+  private static JettySolrRunner jettyNode2;
```
dsmiley, line 47: static fields in tests that refer to Solr should be null'ed out, so there's some burden to them.  Here... I think you could avoid these altogether and simply call `cluster.getJettySolrRunner(0)` which isn't bad!

solr/solrj/src/test/org/apache/solr/client/solrj/impl/ClusterStateProviderTest.java
```diff
@@ -183,4 +204,99 @@ public void testClusterStateProvider() throws SolrServerException, IOException {
...
+
+      // New nodes are removable from known hosts
+      actualKnownNodes = cspHttp.getKnownNodes();
+      assertEquals(2, actualKnownNodes.size());
+      assertEquals(expectedKnownNodes, actualKnownNodes);
+    }
+  }
+
+  private void waitForCSPCacheTimeout() throws InterruptedException {
+    Thread.sleep(6000);
```
dsmiley, line 295: This test should set the system property `solr.solrj.cache.timeout.sec` to maybe 1ms and then you can merely sleep for like a 100 milliseconds.

dsmiley (commented):
Thanks for the PR!

solr/solrj/src/test/org/apache/solr/client/solrj/impl/ClusterStateProviderTest.java
dsmiley: could add a convenience method if you like.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
mlbiscoc: I originally did it with just `liveNodes` but liveNodes is returned to the client which from it's name I was thinking they would assume they are "live". If we just place all the nodes into liveNodes thats not necessarily true, right? If the current set hold all the initially passed nodes, it isn't guaranteed it's live which is why I switched to known nodes while live is what was actually fetched from ZooKeeper.

dsmiley: There's no guarantee that getLiveNodes is going to return a list of reachable nodes.  A moment after returning it, a node could become unreachable.  So it's "best effort" and the caller has to deal with failures by trying the other nodes in the list and/or getting a possibly refreshed list.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/ClusterStateProviderTest.java
mlbiscoc: Cache timeout setting is in seconds and pulls an int. Can't make it 1ms. This probably should have had better granularity instead of increments of seconds.

I could change cacheTimeout to take milliseconds as the int instead but that would change peoples system property not using the default. So if they set the cache timeout to 5 seconds, it now turns into 5ms if they are not aware of this change.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
mlbiscoc: Thats fair. Removed known nodes and put everything into live nodes while using the initial set inside all the time.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -61,6 +67,7 @@ public abstract class BaseHttpClusterStateProvider implements ClusterStateProvid
   private int cacheTimeout = EnvUtils.getPropertyAsInteger("solr.solrj.cache.timeout.sec", 5);
 
   public void init(List<String> solrUrls) throws Exception {
+    this.initialNodes = getNodeNamesFromSolrUrls(solrUrls);
```
mlbiscoc, line 70: > I think the idea of this JIRA issue is that we'll take the Solr URLs as configured and use this is the initial / backup liveNodes. I think this is a very simple idea to to document/understand/implement.

That makes sense. This leaves me with a few questions then. Shouldn't this take a list of `URL` or `URI` java object to verify actual non malformed URLs instead of a list of URLs? I created the functions below to convert these Strings into cluster state nodeNames for `liveNodes`

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -413,6 +418,30 @@ public String getQuorumHosts() {
...
+    return Collections.unmodifiableSet(set);
+  }
+
+  /** URL to cluster state node name (http://127.0.0.1:12345/solr to 127.0.0.1:12345_solr) */
+  public String getNodeNameFromSolrUrl(String solrUrl)
+      throws MalformedURLException, URISyntaxException {
+    URL url = new URI(solrUrl).toURL();
+    return url.getAuthority() + url.getPath().replace('/', '_');
+  }
+
```
mlbiscoc, line 444: Not sure if these methods belong here but it is required to convert the initial set of String urls into cluster state node names

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
dsmiley: great idea to do basic URL malformed checks like this

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -413,6 +418,30 @@ public String getQuorumHosts() {
     return String.join(",", this.liveNodes);
   }
 
+  /** Live nodes should always have the latest set of live nodes but never remove initial set */
+  private void setLiveNodes(Set<String> nodes) {
+    Set<String> liveNodes = new HashSet<>(nodes);
+    liveNodes.addAll(this.initialNodes);
+    this.liveNodes = Set.copyOf(liveNodes);
```
dsmiley, line 425: Why is this 3 lines of extra copy-ing instead of nothing more than: `this.liveNodes = Set.copyOf(nodes);` ?  That is, why are we touching / using initialNodes at all here?
Maybe an IllegalArgumentException if nodes.isEmpty.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -413,6 +418,30 @@ public String getQuorumHosts() {
...
+  }
+
+  public Set<String> getNodeNamesFromSolrUrls(List<String> urls)
+      throws URISyntaxException, MalformedURLException {
+    Set<String> set = new HashSet<>();
+    for (String url : urls) {
+      String nodeNameFromSolrUrl = getNodeNameFromSolrUrl(url);
+      set.add(nodeNameFromSolrUrl);
+    }
+    return Collections.unmodifiableSet(set);
```
dsmiley: Could easily be converted to a single expression with streams

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -413,6 +418,30 @@ public String getQuorumHosts() {
...
+  }
+
+  /** URL to cluster state node name (http://127.0.0.1:12345/solr to 127.0.0.1:12345_solr) */
+  public String getNodeNameFromSolrUrl(String solrUrl)
+      throws MalformedURLException, URISyntaxException {
+    URL url = new URI(solrUrl).toURL();
+    return url.getAuthority() + url.getPath().replace('/', '_');
+  }
```
dsmiley, line 442: can you find other code in Solr doing this?  Surely it's somewhere.

dsmiley: should be a static method that with a unit test

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -229,10 +236,9 @@ > getCacheTimeout()) {
       for (String nodeName : liveNodes) {
```
dsmiley, line 236: If we exhaust liveNodes, shouldn't we then try the initial configured nodes, and then only failing that, throw an exception?

dsmiley:
I simplified characterization of how I think this should be done:
* on initialization, copy the configured URLs to a field for safe keeping.  Convert to URL if you like (validates if malformed)
* on initialization, liveNodes shall be empty; we haven't fetched it yet.  Ideally we don't even try to on initialization.
* when getLiveNodes is called and it's not out of date, return it.  If it's out of date, loop live nodes _and then the initial configured URLs_ for the first server that will respond to fetch the current live nodes.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
mlbiscoc: Hmm maybe... I found `getBaseUrlForNodeName` but it's going the other way nodeName->url. Let me look around more

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
mlbiscoc: My idea was to not do another loop through if liveNodes was exhausted. `initalNodes` would always exist in liveNodes with this set method. Let me refactor again from your suggestion comment

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -226,17 +232,15 @@ public Set<String> getLiveNodes() {
...
-        }
-      }
+
+      if (updateLiveNodes(liveNodes)) return this.liveNodes;
+
+      log.warn(
+          "Attempt to fetch cluster state from all known live nodes {} failed. Trying backup nodes",
+          liveNodes);
+
+      if (updateLiveNodes(backupNodes)) return this.liveNodes;
```
mlbiscoc, line 242: Backup is checked if liveNodes is exhausted. There are a few places that liveNodes is used for fetching but backup isn't now. Was thinking of adding it, but that requires a larger refactor from just this bug and writing additional tests. Happy to do it though.

solr/solrj/src/java/org/apache/solr/common/util/Utils.java
```diff
@@ -786,6 +789,20 @@ public static String getBaseUrlForNodeName(
...
+   * @param solrUrl Given a base Solr URL string (e.g., 'https://app-node-1:8983/solr')
+   * @return URL that looks like {@code app-node-1:8983_solr}
+   * @throws MalformedURLException if the provided URL string is malformed
+   * @throws URISyntaxException if the provided URL string could not be parsed as a URI reference.
+   */
+  public static String getNodeNameFromSolrUrl(String solrUrl)
+      throws MalformedURLException, URISyntaxException {
+    URL url = new URI(solrUrl).toURL();
+    return url.getAuthority() + url.getPath().replace('/', '_');
+  }
```
mlbiscoc, line 804: I couldn't find anywhere where this is a function like this (Maybe you someone knows?). Regardless, I think if it exists somewhere, it should sit here because the `getBaseUrlForNodeName` function right above it does the conversion `nodeName->Url` and putting `url->nodeName` right below it makes sense to me. Added unit tests and just added unit tests for `getBaseUrlForNodeName` because it didn't exist before.

mlbiscoc (commented):
> I simplified characterization of how I think this should be done:

on initialization, copy the configured URLs to a field for safe keeping. Convert to URL if you like (validates if malformed)
on initialization, liveNodes shall be empty; we haven't fetched it yet. Ideally we don't even try to on initialization.
when getLiveNodes is called and it's not out of date, return it. If it's out of date, loop live nodes and then the initial configured URLs for the first server that will respond to fetch the current live nodes.

So I went back and basically implemented the way you put here and deferred the static URL and opt-in node discovery option because it probably needs more discussion and implementation details before hand.

Only thing I didn't do was remove the liveNode fetch on initialization. Curious why CSP shouldn't do that? Wouldn't we want to confirm that the URLs passed are live?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -51,6 +55,7 @@ public abstract class BaseHttpClusterStateProvider implements ClusterStateProvid
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   private String urlScheme;
+  private Set<String> backupNodes;
```
dsmiley, line 58: Can we name this configuredNodes since that's what it is?  Yes we use it as a backup but we could adapt how we use them.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -226,17 +232,15 @@ public Set<String> getLiveNodes() {
...
+
+      log.warn(
+          "Attempt to fetch cluster state from all known live nodes {} failed. Trying backup nodes",
+          liveNodes);
+
+      if (updateLiveNodes(backupNodes)) return this.liveNodes;
+
       throw new RuntimeException(
           "Tried fetching live_nodes using all the node names we knew of, i.e. "
               + liveNodes
```
dsmiley, line 246: and the configured nodes.  Also don't recommend zkHosts :-)

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -226,17 +232,15 @@ public Set<String> getLiveNodes() {
     }
```
dsmiley, line 232: Just above this, there's a liveNodes null check.  That seems bogus now

dsmiley (commented):
See [SOLR-17607](https://issues.apache.org/jira/browse/SOLR-17607) for why we shouldn't connect on creation.

There's `Utils.getBaseUrlForNodeName` for one direction so the reverse should go right next to it.

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
```diff
@@ -530,14 +530,14 @@ public JettySolrRunner stopJettySolrRunner(int index) throws Exception {
...
-   * Add a previously stopped node back to the cluster
+   * Add a previously stopped node back to the cluster and reuse its port
    *
    * @param jetty a {@link JettySolrRunner} previously returned by {@link #stopJettySolrRunner(int)}
    * @return the started node
    * @throws Exception on error
    */
   public JettySolrRunner startJettySolrRunner(JettySolrRunner jetty) throws Exception {
-    jetty.start(false);
+    jetty.start();
```
dsmiley, line 540: That makes sense... I'll call it out in the commit log.  I wonder if there's any down-side to this.

dsmiley (commented):
You know what... See URLUtil.  Shouldn't *that* be where these methods go?  The existing method can be deprecated and call the one at the better location.
Also has a test where your tests of this method can be added to.
(big code base; so many util classes)

This PR is pretty close to merging now.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/ClusterStateProviderTest.java
```diff
@@ -51,6 +54,15 @@ public static void setupCluster() throws Exception {
                 .resolve("streaming")
                 .resolve("conf"))
         .configure();
+    cluster.waitForAllNodes(30);
+    System.setProperty("solr.solrj.cache.timeout.sec", "1");
```
cpoerschke, line 58: (am random pull request browser/scanner only here) -- maybe have matching `System.clearProperty` during test cleanup to avoid carry-ever effects between tests.

dsmiley: Christine, Solr's test infrastructure auto-restores all system properties to as they were from test suite to test suite thanks to SystemPropertiesRestoreRule.  But not test-method to test-method, albeit that's not pertinent for the line you commented (as it's in a BeforeClass).

mlbiscoc:
> You know what... See URLUtil. Shouldn't _that_ be where these methods go? The existing method can be deprecated and call the one at the better location. Also has a test where your tests of this method can be added to. (big code base; so many util classes)
> 
> This PR is pretty close to merging now.

I ended up moving all those functions into `URLUril` which resulted in a much larger code change. Not sure if you meant for me to use the `Deprecated` annotation but instead just moved it and refactored the callers.

Also didn't add in [SOLR-17607](https://issues.apache.org/jira/browse/SOLR-17607) because it broke a few more tests that needed more changes so just held the changes locally and will probably just make it in another PR and link it to that JIRA unless you prefer me to do it here in one go.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -248,15 +240,19 @@ > getCacheTimeout()) {
...
     } else {
       return this.liveNodes; // cached copy is fresh enough
     }
   }
 
+  private boolean updateLiveNodes(List<String> liveNodes) {
+    return updateLiveNodes(Set.copyOf(liveNodes));
+  }
+
   private boolean updateLiveNodes(Set<String> liveNodes) {
```
dsmiley, line 253: why does updateLiveNodes take a Set as an argument; can't it simply be a Collection?  The Set is forcing you to do a conversion and that which changes the order, and the order may be deliberate.  It's not worth overloading the method for either.

solr/solrj/src/java/org/apache/solr/common/util/Utils.java
```diff
@@ -748,61 +745,6 @@ public static String applyUrlScheme(final String url, final String urlScheme) {
...
-  /**
-   * Construct a V1 base url for the Solr node, given its name (e.g., 'app-node-1:8983_solr') and a
-   * URL scheme.
-   *
-   * @param nodeName name of the Solr node
-   * @param urlScheme scheme for the base url ('http' or 'https')
-   * @return url that looks like {@code https://app-node-1:8983/solr}
-   * @throws IllegalArgumentException if the provided node name is malformed
-   */
-  public static String getBaseUrlForNodeName(final String nodeName, final String urlScheme) {
```
dsmiley, line 757: These methods are called in a bunch of places; we can't simply remove them in 9x.  They can stay and call the new location and be deprecated.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -51,6 +56,7 @@ public abstract class BaseHttpClusterStateProvider implements ClusterStateProvid
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   private String urlScheme;
+  private List<String> configuredNodes;
```
dsmiley, line 59: Perhaps configuredNodes should be a List<URL>?  Why convert to a live node format?

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
dsmiley: I suspect the results of this could be subtle, and nobody would guess this JIRA/PR-title would have anything to do with this.  Thus should be merged separately.  What did you find that led you to incorporate this change into this PR?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
HoustonPutman: Yeah, if the idea is to have 1 method that handles both, then I would instead have a `liveNodeURLs` list in addition to `liveNodes`, so that the `updateLiveNodes ` method takes in the urls.

But the method can probably split to share logic, but handle both strings and urls as inputs.

mlbiscoc: I basically tried to keep `configuredNodes` the same format as `liveNodes` as they are used the same with `updateLiveNodes`. I don't like the idea of another `liveNodeURLs` list attribute and think just changing liveNodes to be `List<URL>` would make sense but the caller now get a different list URLs and not nodeNames.

I guess I'll just go the route of updateLiveNodes functions handling both `NodeName` strings and `Base URL` inputs unless you feel strongly about having an additional `liveNodeURLs` list

HoustonPutman: yeah, keeping `liveNodes` as string and `configuredNodes` as URLs, with the method handling each sounds good to me.

dsmiley: We don't need a `liveNodeURLs` field.  We know how to convert on the fly.

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
dsmiley: When faced with curiosities like this, it's useful to do a bit of code/JIRA archeology.  `git blame` (via IntelliJ's margin, BTW) shows the current setting (to pick a random port) was *deliberate* and had its own JIRA -- https://issues.apache.org/jira/browse/SOLR-9474 and I found a related one: https://issues.apache.org/jira/browse/SOLR-9469 .  I read the conversation there.  Changing it is worthy of its own JIRA & CHANGES.txt entry, following a deeper attempt to understand the causes behind why it hasn't been reusing a port thus far.

solr/solrj/src/java/org/apache/solr/common/util/URLUtil.java
```diff
@@ -98,4 +102,59 @@ private static String removeTrailingSlashIfPresent(String url) {
...
+   *
+   * @param solrUrl Given a base Solr URL string (e.g., 'https://app-node-1:8983/solr')
+   * @return Node name that looks like {@code app-node-1:8983_solr}
+   * @throws MalformedURLException if the provided URL string is malformed
+   * @throws URISyntaxException if the provided URL string could not be parsed as a URI reference.
+   */
+  public static String getNodeNameForBaseUrl(String solrUrl)
+      throws MalformedURLException, URISyntaxException {
+    URL url = new URI(solrUrl).toURL();
+    return url.getAuthority() + url.getPath().replace('/', '_');
```
mlbiscoc, line 158: After changing configured nodes to List<URL>, this function isn't used anymore except for tests.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
mlbiscoc: Changed to `List<URL>` and changed `updateLiveNodes` accordingly. Lmk if how I changed it makes sense.

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
mlbiscoc: Thanks for find that. Thats fair. For now, I overloaded the function with a `reusePort` to not break anything and maybe change it to just reuse port later. Is it fine to do that for now, or should I just be calling jetty itself and reuse the port in the tests?

solr/solrj/src/test/org/apache/solr/client/solrj/impl/ClusterStateProviderTest.java
```diff
@@ -204,8 +204,8 @@ public void testClusterStateProviderDownedInitialLiveNodes() throws Exception {
       var jettyNode1 = cluster.getJettySolrRunner(0);
       var jettyNode2 = cluster.getJettySolrRunner(1);
 
-      String nodeName1 = getNodeNameFromSolrUrl(jettyNode1.getBaseUrl().toString());
-      String nodeName2 = getNodeNameFromSolrUrl(jettyNode2.getBaseUrl().toString());
+      String nodeName1 = getNodeNameForBaseUrl(jettyNode1.getBaseUrl().toString());
```
dsmiley, line 207: Was From -> For deliberate?  IMO it was named fine.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -387,6 +398,19 @@ public String getPolicyNameByCollection(String coll) {
             + "ZkClientClusterStateProvider can be used for this."); // TODO
   }
 
+  private List<URL> setConfiguredNodes(List<String> solrUrls) {
```
dsmiley, line 401: this doesn't actually "set" anything; it's a utility function mapping input to output converting to a URL.  It could even be static.   If you declare a static method that only converts a URL with the try-catch, it'd be simple for the caller (init) to invoke it with the streams mapping in a one-liner.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/ClusterStateProviderTest.java
mlbiscoc: Yeah it was deliberate. Was mostly to keep consistency between the names. `getNodeNameForBaseUrl` and `getBaseUrlForNodeName` but can switch it back if you prefer.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
mlbiscoc: You're right bad naming. It's only used in the `init` so Instead I just removed the function as I don't think we need another utility method and it was somewhat small enough inlined the whole thing into the stream where it was called. Can move it into a private static utility if you prefer otherwise.

solr/CHANGES.txt
```diff
@@ -176,6 +176,10 @@ Bug Fixes
 * SOLR-17629: If SQLHandler failed to open the underlying stream (e.g. Solr returns an error; could be user/syntax problem),
   it needs to close the stream to cleanup resources but wasn't. (David Smiley)
 
+* SOLR-17519: SolrJ fix to CloudSolrClient and HTTP ClusterState where it can fail to request cluster state
+  if its current live nodes list is stale. HTTP ClusterState now retains the initial configured list of passed URLs as backup
```
dsmiley, line 180: I don't think Solr users know about "HTTP ClusterState"; it's not likely in their vocabulary.  On the other hand, wording like "SolrJ CloudSolrClient configured with Solr URLs" is likely to be understood.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -216,38 +236,42 @@ private SimpleOrderedMap<?> submitClusterStateRequest(
...
+
+      if (liveNodes.stream()
+          .anyMatch((node) -> updateLiveNodes(URLUtil.getBaseUrlForNodeName(node, urlScheme))))
+        return this.liveNodes;
+
+      log.warn(
+          "Attempt to fetch cluster state from all known live nodes {} failed. Trying backup nodes",
+          liveNodes);
+
+      if (configuredNodes.stream().anyMatch((node) -> updateLiveNodes(node.toString())))
```
dsmiley, line 250: Shouldn't we check configuredNodes *first* so as to give the user the possibility of some control, basically addressing the conversation points in JIRA?

mlbiscoc: We could but to me it makes sense to try live nodes first unless this class was explicitly created as your JIRA proposal pointed, in that it only uses the passed URLs and nothing else. 

I think it should be used as backup and go for fetching from live nodes first and assume by its name it is "live". If we do the opt in as a feature flag for dynamic node discovery then that gives them control. I'm fine with flipping it with configured first though if you feel strongly about it.

dsmiley:
I'll merge this in about a day as-is.

PR #2940, opened 2025-01-06
Demonstrate roundtrip export/import works

epugh:
https://issues.apache.org/jira/browse/SOLR-13689

# Description

Trying to understand best ways of round tripping data.  Export from one collection and index into another collection.   Use our existing tooling as much as possible.

# Solution

Starting with a BATS test to demonstrate that bin/solr export and bin/solr post with .json file works.

# Tests

BATS.

### DIFF
--- a/solr/solr-ref-guide/modules/deployment-guide/pages/solr-control-script-reference.adoc
+++ b/solr/solr-ref-guide/modules/deployment-guide/pages/solr-control-script-reference.adoc
@@ -1703,7 +1703,8 @@ The default is `\*:*` which will export all documents.
 |===
 +
 Comma separated list of fields to be exported.
-By default all fields are fetched.
+By default all fields are fetched.   
+If your schema uses `copyFields` and you are reindex the data, then you probably want to specify exactly which fields are being exported to simplify the import process.
 
 `--limit <number of documents>`::
 +
--- a/solr/solr-ref-guide/modules/deployment-guide/pages/solr-control-script-reference.adoc
+++ b/solr/solr-ref-guide/modules/deployment-guide/pages/solr-control-script-reference.adoc
@@ -1704,7 +1704,7 @@ The default is `\*:*` which will export all documents.
 +
 Comma separated list of fields to be exported.
 By default all fields are fetched.   
-If your schema uses `copyFields` and you are reindex the data, then you probably want to specify exactly which fields are being exported to simplify the import process.
+If your schema uses `copyFields` and you reindex the data, then you probably want to specify exactly which fields are being exported to simplify the import process.
 
 `--limit <number of documents>`::
 +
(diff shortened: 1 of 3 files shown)


epugh:
I'd love a plus one on the this before I merge...    For a lot of multi step processes, like exporting and importing data, I find modeling them as BATS tests makes them easier to understand in context.    I know that moves these BATS tests towards being integration or even system style tests...     I suspect that we may need to split the bats tests into system/integration/unit tests in the future..

solr/solr-ref-guide/modules/deployment-guide/pages/solr-control-script-reference.adoc
```diff
@@ -1703,7 +1703,8 @@ The default is `\*:*` which will export all documents.
 |===
 +
 Comma separated list of fields to be exported.
-By default all fields are fetched.
+By default all fields are fetched.   
+If your schema uses `copyFields` and you are reindex the data, then you probably want to specify exactly which fields are being exported to simplify the import process.
```
dsmiley, line 1707:
```suggestion
If your schema uses `copyFields` and you reindex the data, then you probably want to specify exactly which fields are being exported to simplify the import process.
```

dsmiley (approved):
Cool.  Love the how-to nature of a BATS test here.

epugh:
> Cool. Love the how-to nature of a BATS test here.

I have a dream someday to re-organize all the Solr documentation along the lines of [Diátaxis](https://diataxis.fr/#).  I've used this model a few times and quite like it.

solr/solr-ref-guide/modules/indexing-guide/pages/reindexing.adoc
```diff
@@ -165,6 +165,20 @@ Once the alias is in place and you are satisfied you no longer need the old data
 One advantage of this option is that you can switch back to the old collection if you discover problems our testing did not uncover.
 Of course this option can require more resources until the old collection can be deleted.
 
+=== Exporting/Importing Data from Solr
+
+Sometimes you don't want to run your full ETL pipeline to reindex into another collection, you just want to take the data in your existing collection, export it, and then import it back.
```
gerlowskija, line 170: [-0] AFAIK `bin/solr export` uses the `/export` handler and [can only return fields that have docValues enabled](https://solr.apache.org/guide/solr/latest/query-guide/exporting-result-sets.html#field-requirements).  So it'd drop most text fields, etc.

That's a huge limitation that we should probably mention here!

solr/solr-ref-guide/modules/indexing-guide/pages/reindexing.adoc
```diff
@@ -165,6 +165,20 @@ Once the alias is in place and you are satisfied you no longer need the old data
...
+
+There are a number of third party tools that do this, see https://solr.cool/ for more information.   However, if you want to use what ships with Solr then we have some options:
+
+1. Use `bin/solr export` with the JSON output format (`.json`), and the `bin/solr post` tool to post that data back.
```
gerlowskija, line 174: [Q] How would a user choose between these three options?  Or put differently - why would they choose one over the others?

If there's no strong differentiator between the three, is there value in mentioning them all individually?

PR #2960, opened 2025-01-07
SOLR-17447 : Support to early terminate a search based on maxHitsAllowed per shard.

sijuv:
# Description

Adds the capability to early terminate a search based on maxHits parameter provided

https://issues.apache.org/jira/browse/SOLR-17447

# Solution

"maxHitsPerShard" request parameter controls how many hits the searcher should go over per shard. Once the searcher runs over the specfied number of documents, it will terminate the search with EarlyTerminatingCollectorException. This will be indicated by a new response header "terminatedEarly" also the "partialResults" will indicate that the results are partial. This parameter is supported in MT mode as well.

Though there are other mechanisms to control runaway queries with CPU usage limits and time limits, this is simpler for certain use cases esp in case high recall queries and rerank use cases.

Lucene currently supports this feature with the EarlyTerminatingCollector. There was some code in SOLR as well to support the collector, but looks like it was not completely wired up.

# Tests

Ran tests against a local solr instance in MT and single threaded mode

# Checklist

- [ X] I have reviewed the guidelines for [How to Contribute](https://github.com/apache/solr/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.
- [X ] I have run `./gradlew check`.


dsmiley (commented):
Overall this makes sense to me.  Thanks for the nice contribution!  I'd prefer if @gus-asf could take a look at some aspects since he worked on cpuAllowed recently.  BTW it's clear you need to run `./gradlew tidy`

I suggest renaming the param "maxHitsPerShard" to simply "maxHits" or "maxHitsTerminateEarly" and document that it's per-shard and best-effort; that more hits may ultimately be detected in aggregate.  But maybe others disagree.

It'd be good to consider interference with other features.  Maybe it works with cursorMark?  Does it count *after* PostFilter (e.g. CollapseQParser) or before?

sijuv:
> Overall this makes sense to me. Thanks for the nice contribution! I'd prefer if @gus-asf could take a look at some aspects since he worked on cpuAllowed recently. BTW it's clear you need to run `./gradlew tidy`
Addressed.
> 
> I suggest renaming the param "maxHitsPerShard" to simply "maxHits" or "maxHitsTerminateEarly" and document that it's per-shard and best-effort; that more hits may ultimately be detected in aggregate. But maybe others disagree.
> 
Renamed to maxHits
> It'd be good to consider interference with other features. Maybe it works with cursorMark? Does it count _after_ PostFilter (e.g. CollapseQParser) or before?
Sorry, I am not aware of cursorMark. This will count while the collector runs over the posting list so it is not a post filter. Not sure if I understood your question correctly,

sijuv:
@dsmiley can you pls relook when you get a chance ?

dsmiley:
note: you force-pushed to this PR.  Please don't do that; it resets the review state making it hard for me to see changes since my last review and if I try it's still possible an earlier commit was changed and I won't know it.  So never force-push to a PR that has been reviewed.

solr/core/src/java/org/apache/solr/search/MultiThreadedSearcher.java
```diff
@@ -255,7 +265,11 @@ public Object reduce(Collection collectors) throws IOException {
       for (Iterator var4 = collectors.iterator();
           var4.hasNext();
           maxScore = Math.max(maxScore, collector.getMaxScore())) {
-        collector = (MaxScoreCollector) var4.next();
+        Collector next = (Collector) var4.next();
```
dsmiley, line 268: could use a comment like "assume the next collector, except EarlyTerminating, is the MaxScoreCollector".

While you are changing this code, please rename `var4` which is an inexcusable variable name

solr/core/src/java/org/apache/solr/search/QueryCommand.java
```diff
@@ -194,7 +195,7 @@ public QueryCommand setNeedDocSet(boolean needDocSet) {
   }
 
   public boolean getTerminateEarly() {
-    return (flags & SolrIndexSearcher.TERMINATE_EARLY) != 0;
+    return (flags & SolrIndexSearcher.TERMINATE_EARLY) != 0 || maxHits < Integer.MAX_VALUE;
```
dsmiley, line 198: When I see these other constants about TERMINATE_EARLY, this is what leads me to think that maxHitsTerminateEarly is a good name.

solr/core/src/java/org/apache/solr/search/QueryResult.java
```diff
@@ -22,6 +22,7 @@ public class QueryResult {
   // Object for back compatibility so that we render true not "true" in json
   private Object partialResults;
   private Boolean segmentTerminatedEarly;
```
dsmiley, line 24: thinking out loud:  Perhaps we should do away with segmentTerminatedEarly as overly specific

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -338,6 +339,11 @@ private Collector buildAndRunCollectorChain(
       if (cmd.isQueryCancellable()) {
         core.getCancellableQueryTracker().removeCancellableQuery(cmd.getQueryID());
       }
+      if (collector instanceof final EarlyTerminatingCollector earlyTerminatingCollector) {
```
dsmiley, line 342: this check is brittle since it can only possibly true if the *last* collector is earlyTerminatingCollector.  We may add others or re-order at some point.  Also, can we just depend on the exception above and set `qr.setTerminatedEarly(true);` there?

dsmiley (commented):
Recommended additional reviewers based on touching similar functionality: @atris @gus-asf  @cpoerschke

solr/core/src/java/org/apache/solr/response/SolrQueryResponse.java
```diff
@@ -68,6 +68,7 @@ public class SolrQueryResponse {
   public static final String RESPONSE_HEADER_PARTIAL_RESULTS_DETAILS_KEY = "partialResultsDetails";
   public static final String RESPONSE_HEADER_SEGMENT_TERMINATED_EARLY_KEY =
       "segmentTerminatedEarly";
+  public static final String RESPONSE_HEADER_TERMINATED_EARLY_KEY = "terminatedEarly";
```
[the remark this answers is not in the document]
dsmiley: Okay you did in the ref guide... still a simple bit of javadoc here would be helpful

dsmiley:
A test is needed.

solr/core/src/java/org/apache/solr/search/EarlyTerminatingCollector.java
```diff
@@ -29,11 +30,15 @@
...
 public class EarlyTerminatingCollector extends FilterCollector {
 
+  private final int chunkSize = 100; // Check across threads only at a chunk size
+
   private final int maxDocsToCollect;
 
   private int numCollected = 0;
   private int prevReaderCumulativeSize = 0;
   private int currentReaderSize = 0;
+  private final AtomicInteger pendingDocsToCollect;
```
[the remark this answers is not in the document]
HoustonPutman: Agreed, it looks like the chunking can go away if we switch to a LongAdder. Much like the chunking implementation, we won't get absolute accuracy across threads, but we will get higher throughput. (And we could sacrifice throughput to increase accuracy if we wanted to, but I doubt we do)

solr/core/src/java/org/apache/solr/search/QueryResult.java
HoustonPutman: Agreed, I'm not sure how people are going to respond differently between the two

solr/solr-ref-guide/modules/query-guide/pages/common-query-parameters.adoc
```diff
@@ -399,6 +399,22 @@ For example, setting `cpuAllowed=500` gives a limit of at most 500 ms of CPU tim
...
+|===
+|Optional |Default: `false`
+|===
+
+This parameter specifies the max number of hits a searcher will iterate through before early terminating the search.
+The count is per shard and across all threads involved in case of multi-threaded search. This parameter works
+in conjunction with other parameters that could early terminate a search, ex: _timeAllowed_ etc. In case the search
+was early terminated due to it exceeding maxHits a _terminatedEarly_ header in the response will be set along with
+_partialResults_ to indicate the same
+Note : the hits counted will need not be exactly equal to the maxHits provided, however it will be within range of this value.
```
HoustonPutman, line 415: Can you explain here when the user would see partial results? The user will probably not be very familiar with how Lucene or the searcher works, so it should be from their point of view.

solr/core/src/java/org/apache/solr/search/QueryCommand.java
HoustonPutman: I agree, the name of the param does not really indicate that it could lead to partial results.

solr/core/src/java/org/apache/solr/response/SolrQueryResponse.java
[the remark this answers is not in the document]
sijuv: updated.

solr/core/src/java/org/apache/solr/search/EarlyTerminatingCollector.java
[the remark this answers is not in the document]
sijuv: Chunking we will still need. LongAdder add does not return the current count and doing increment and then get on every document I don't think will be optimal.

solr/core/src/java/org/apache/solr/search/QueryCommand.java
sijuv: done.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
sijuv: done.

sijuv:
> A test is needed.

added tests.

solr/solr-ref-guide/modules/query-guide/pages/common-query-parameters.adoc
sijuv: updated documentation.

sijuv:
@dsmiley @HoustonPutman updated, pls take a look when you get a chance.

solr/core/src/java/org/apache/solr/search/QueryResult.java
dsmiley: Okay; so the cause is interesting for diagnostic / observability purposes but semantically, the search has "terminated early" (for whatever reason).

gus-asf:
> > Overall this makes sense to me. Thanks for the nice contribution! I'd prefer if @gus-asf could take a look at some aspects since he worked on cpuAllowed recently. BTW it's clear you need to run `./gradlew tidy`
> > Addressed.
> > I suggest renaming the param "maxHitsPerShard" to simply "maxHits" or "maxHitsTerminateEarly" and document that it's per-shard and best-effort; that more hits may ultimately be detected in aggregate. But maybe others disagree.
> 
> Renamed to maxHits
> 

maybe shardMaxHits? Brevity is nice, but I think it's helpful to have a name that is at least slightly self documenting.

solr/core/src/java/org/apache/solr/search/QueryResult.java
gus-asf: I agree it's good to know why so that the corrective action is clear, but terminateEarly is unlike maxHits (or shardMaxHits) so the relationship takes special knowledge to appreciate. If something is going to report a reason I'd want it to include the same phrase... i.e. "maxHitsReached" and definately would not want to have to "just know" that terminateEarly has nothing to do with segmentTerminateEarly despite strong similarity in naming.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -329,12 +329,12 @@ private Collector buildAndRunCollectorChain(
       if (collector instanceof DelegatingCollector) {
         ((DelegatingCollector) collector).complete();
       }
-      throw etce;
+      qr.setPartialResults(true);
```
gus-asf, line 332: Does not throwing here effect SOLR-3240, (the original reason for ETCE) ?

gus-asf (changes requested):
Overall I think this is a great idea. For some use cases this is certainly appropriate. My primary concern is that the use of "terminate early" is getting pretty overloaded. There is code relating to lucene segmentTerminateEarly, and terminateEarly was previously used in SOLR-3240, and either you have stepped on some of that code, possibly breaking things if it isn't dead code from a previously removed feature. My suggestion is that this not add a 3rd "terminateEarly" but rather use it's functional behavior as a name... "searcherMaxHits" or maybe "shardMaxHits" and edit that more easily destinguished and self documenting name through the various code locations where we check it or name things after it. 

Finally I wonder if this and spellcheck_collate_max_docs should be doing exactly the same thing, or if they should be entirely separate. Should the spellcheck feature actually be using this under the covers now that it exists (or is this really just exposing the same idea for non-spellcheck cases)?

dsmiley:
(not thinking of backwards compatibility here, just thinking holistically)

I appreciate that there are very different causes of a terminated request, and we want to differentiate them in case the client cares and for diagnostics.  But semantically to the caller, the cause is irrelevant (right?), so I think a simple flag/enum should be returned _separate_ for the cause.  I'm not arguing the cause should be hidden / indiscernible.  I understand, once upon a time, there was a single cause of "terminate early" but as Solr has matured to do so under growing conditions, we shouldn't be held back by segment terminated early as the one and only terminate-early.  Let's step back and change names / semantics if needed.  Solr 10 is coming.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -295,7 +295,7 @@ private Collector buildAndRunCollectorChain(
 
     final boolean terminateEarly = cmd.getTerminateEarly();
     if (terminateEarly) {
-      collector = new EarlyTerminatingCollector(collector, cmd.getLen());
+      collector = new EarlyTerminatingCollector(collector, cmd.getMaxHitsTerminateEarly());
```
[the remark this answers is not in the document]
sijuv: I dont think the _EarlyTerminatingCollector_ has been used in solr yet. The code prior to my change was instantiating this collector when _TERMINATE_EARLY_ flag was set, but I dont see any references to this flag being set in the code because there are no usage references to _setTerminateEarly_.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
sijuv: I dont think _EarlyTerminatingCollector_ has been used yet.

solr/core/src/java/org/apache/solr/search/EarlyTerminatingCollector.java
```diff
@@ -61,11 +69,25 @@ public LeafCollector getLeafCollector(LeafReaderContext context) throws IOExcept
       public void collect(int doc) throws IOException {
         super.collect(doc);
         numCollected++;
-        if (maxDocsToCollect <= numCollected) {
+        terminatedEarly = maxDocsToCollect <= numCollected;
+        if (numCollected % chunkSize == 0) {
+          pendingDocsToCollect.add(chunkSize);
+          final long overallCollectedDocCount = pendingDocsToCollect.intValue();
+          terminatedEarly = overallCollectedDocCount >= maxDocsToCollect;
```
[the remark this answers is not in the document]
sijuv: the boolean is updated only every 100th time to reduce any overhead updating the thread shared adder brings in.

solr/solr-ref-guide/modules/query-guide/pages/common-query-parameters.adoc
```diff
@@ -399,6 +399,23 @@ For example, setting `cpuAllowed=500` gives a limit of at most 500 ms of CPU tim
...
+
+This parameter specifies the max number of hits a searcher will iterate through before early terminating the search.
+The count is per shard and across all threads involved in case of multi-threaded search. This parameter works
+in conjunction with other parameters that could early terminate a search, ex: _timeAllowed_ etc. In case the search
+was early terminated due to it exceeding maxHits a _terminatedEarly_ header in the response will be set along with
+_partialResults_ to indicate the same. Note that the _partialResults_ flag could be set in the absence of the _maxHits_
+parameter due to other limits like _timeAllowed_ or _cpuAllowed_.
+Note : the hits counted will need not be exactly equal to the maxHits provided, however it will be within range of this value.
+
+
```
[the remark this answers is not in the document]
sijuv: updated.

solr/core/src/java/org/apache/solr/search/EarlyTerminatingCollector.java
```diff
@@ -43,11 +48,14 @@ public class EarlyTerminatingCollector extends FilterCollector {
    * @param maxDocsToCollect - the maximum number of documents to Collect
    */
   public EarlyTerminatingCollector(Collector delegate, int maxDocsToCollect) {
-    super(delegate);
-    assert 0 < maxDocsToCollect;
```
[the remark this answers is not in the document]
sijuv: added back

solr/core/src/java/org/apache/solr/search/EarlyTerminatingCollector.java
```diff
@@ -29,11 +30,15 @@
  */
 public class EarlyTerminatingCollector extends FilterCollector {
 
+  private final int chunkSize = 100; // Check across threads only at a chunk size
```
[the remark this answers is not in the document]
sijuv: 100 was chosen arbitrarily. I don't think this needs a env prop.

sijuv:
@dsmiley @gus-asf pls relook

sijuv:
In the current implementation, the _EarlyTerminatingCollector_ is not used in any code path.
The collector is instantiated at only one place https://github.com/apache/solr/blob/ac3d349dac530cf1001d5113fc21b0fd641cc9d5/solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java#L293 . Now this is done only if _getTerminateEarly()_ returns true. It will never return true since the setter is never invoked in any code path. I think @gus-asf mentioned that as well, It is an unused method.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -1976,7 +1975,10 @@ public ScoreMode scoreMode() {
         MultiThreadedSearcher.TopDocsResult topDocsResult = searchResult.getTopDocsResult();
         totalHits = topDocsResult.totalHits;
         topDocs = topDocsResult.topDocs;
-
+        if (searchResult.isTerminatedEarly) {
```
HoustonPutman, line 1978: So `searchResult.isTerminatedEarly` could have been set to `true` because maxShardHits had been reached right? If so, shouldn't this be setting `setMaxHitsTerminatedEarly` to true?

Honestly, shouldn't `searchResult` be using `isMaxHitsTerminatedEarly` not `isTerminatedEarly`?

sijuv: Yes you are correct. It should mirror the flags here https://github.com/sijuv/solr/blob/a4f7301e4778c62b7f6d8d5b212c07b0e57bd8be/solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java#L327

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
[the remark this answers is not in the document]
HoustonPutman: Yeah, Gus is right. `SpellCheckCollator::collate` uses it: `checkResponse.setFieldFlags(f |= SolrIndexSearcher.TERMINATE_EARLY)`. So we need to refactor this to use it correctly. I'm going to refactor this to use the new functionality under the hood. Then we can truly remove the TerminateEarly stuff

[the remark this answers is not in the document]
HoustonPutman: So I've changed `SpellCheckCollator::collate` to use this feature now. We can remove terminate_early and all of that stuff in a different PR

HoustonPutman:
Ok I've renamed to `maxHitsAllowed`, which follows a lot of the other shard-specific limits. (like `timeAllowed` or `cpuAllowed`)

HoustonPutman:
I need to add a changelog entry, but I think this is good to go, and will probably merge sometime tomorrow.

HoustonPutman:
I'm planning on merging this early next week in case either of you have additional comments @dsmiley @gus-asf .

I'm going to make a separate ticket and PR to remove the `terminateEarly` stuff in 10x, because it will be unused after this PR.

HoustonPutman:
> Your changes since my last review are nice/simple. I'm not sure how to resolve maybe differing views on what should be consolidated or removed, but happy to defer to your judgement Houston (or Gus). At least hearing "remove ___(whatever)" is probably generally a good thing for simplicity :-)

I mean, this really is just a refactoring (loose interpretation of that word) of what already existed, and giving it a better name and user-facing support. This was only really used in 1 place (Spellcheck Collation), and didn't work in every way that Solr supports searching now. So "removing" isn't really true, it's really replacing an internal way of doing something. Hopefully no one disagrees about that part 😅 

As for the response difference between `segmentTerminatedEarly` and `maxHitsTerminatedEarly` in the responseHeader, I think it's probably best to keep those separate. `segmentTerminatedEarly` really only tells you that the `numFound` information is wrong, since we know that because of the sort, the best documents were found. As for `maxHitsTerminatedEarly`, numFound will likely be wrong (as with the previous flag), but the results are also probably not the most relevant, since we don't have the sorting guarantees that `segmentTerminatedEarly` gives. So I think it does give useful information back to the "caller".

I'm also not against cleaning this up in the future, having a better way of exposing the ways in which the response isn't giving the exactly correct information. That would be a big enough change affecting multiple features (maxHitsAllowed, circuitBreakers, blockMax-wand, segmentTerminatedEarly, etc)  that it probably belongs in its own ticket/PR.

PR #2963, opened 2025-01-08
SOLR-17614: cross-dc deletes can skip an HTTP hop

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17614

Note: I ran the module's tests and they passed.  But I have no experience with this module (or intention to use it) whatsoever.  I'm just scratching an itch to remove a pointless HTTP connection and to use the super-cool EmbeddedSolrServer.

### DIFF
--- a/solr/modules/cross-dc/build.gradle
+++ b/solr/modules/cross-dc/build.gradle
@@ -36,7 +36,6 @@ dependencies {
   implementation libs.apache.kafka.clients
   implementation libs.google.guava
   implementation libs.dropwizard.metrics.core
-  implementation libs.apache.httpcomponents.httpclient
   implementation libs.apache.zookeeper.zookeeper
   implementation libs.apache.zookeeper.jute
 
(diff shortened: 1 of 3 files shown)


dsmiley:
Needs a CHANGES.txt; perhaps in Optimizations.  "Cross-DC delete-by-query optimized to fetch deleted documents with one less HTTP hop per batch of IDs".   I plan to merge it this weekend after adding that.

HoustonPutman:
Is there an issue that this is just querying a core instead of a collection after this change?

dsmiley:
Not that I know of.

But, I confess IMO Solr **should** know what is being _addressed_ and treat _that_ as the scope but Solr doesn't care -- in SolrCloud mode, the whole collection (the core is assigned to) is the scope, unless the magic param `distrib=false` is passed (for queries).  Due to the spectre of back-compat... probably our best path forward is to look to the upcoming V2 API as the opportunity to rethink this.

PR #2975, opened 2025-01-08
SOLR-17599: CrossDC should not use the MirroringUpdateHandler when replaying updates from the transaction log or when doing PeerSync.

markrmiller:


sigram:
Let's reopen this, I think it's worth fixing.

solr/modules/cross-dc/src/java/org/apache/solr/crossdc/update/processor/MirroringUpdateProcessor.java
```diff
@@ -135,6 +136,13 @@ UpdateRequest createMirrorRequest() {
 
   @Override
   public void processAdd(final AddUpdateCommand cmd) throws IOException {
+    boolean isReplayOrPeersync =
```
sigram, line 139: I think it's better to have this logic as a utility method.

PR #3006, opened 2025-01-09
SOLR-17587: (9x backport) wt=prometheus fix duplicate TYPE information 

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-17587

# Description

Solr's Prometheus writer duplicates `# TYPE <metric name> <prometheus metric type>` in it's exposition format for `core`registry metrics. 

This is an illegal format and depending on the technologies prometheus exposition verification for example `Telegraf`, this will fail. For Prometheus server itself, this still passes and collects the metrics just fine for some reason.

This is because the Prometheus Writer takes Dropwizard registries and exports them to Prometheus Registries to expose them in Prometheus format. Solr creates Dropwizard registry for every `core` and differentiates the metrics that way even though they have the same metric names.

For prometheus, this creates an issue in that metrics should be differentiated in it's attributes and tags. So when the metrics are output with the Prometheus response writer, it duplicates the `TYPE` information because it is a registry for every `core` and doesn't know that the other `core` registries have the same metric name and results in duplicate `TYPE` information.

# Solution

When metrics are going to be exported for prometheus, we merge all the `core` Dropwizard metric registries into a single registry and export that registry into prometheus. Duplicate metric names in a registry is not allowed in prometheus, so we will also append the core name to the Dropwizard metric to differentiate which metric belongs to what core and parse the labels accordingly.

This also allowed to clean up and simply some of the `SolrPrometheusCoreFormatter` code.

# Tests

Updated the test accordingly with the coreName existing in the Dropwizard metric names and it's parsing. 

Also added an assert in `testPrometheusStructureOutput` to confirm there is no duplicate `TYPE` information in prometheus output.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreMetric.java
+++ b/solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreMetric.java
@@ -16,37 +16,41 @@
  */
 package org.apache.solr.metrics.prometheus.core;
 
-import static org.apache.solr.metrics.prometheus.core.PrometheusCoreFormatterInfo.CLOUD_CORE_PATTERN;
-
 import com.codahale.metrics.Metric;
 import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.apache.solr.common.SolrException;
 import org.apache.solr.metrics.prometheus.SolrMetric;
 
 /** Base class is a wrapper to export a solr.core {@link com.codahale.metrics.Metric} */
 public abstract class SolrCoreMetric extends SolrMetric {
+
+  public static Pattern CLOUD_CORE_PATTERN =
+      Pattern.compile(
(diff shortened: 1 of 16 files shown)


solr/core/src/java/org/apache/solr/metrics/prometheus/core/SolrCoreMetric.java
```diff
@@ -29,7 +30,8 @@ public abstract class SolrCoreMetric extends SolrMetric {
       Pattern.compile(
           "(?<core>^core_(?<collection>.*)_(?<shard>shard[0-9]+)_(?<replica>replica_.[0-9]+)).(.*)$");
   public static Pattern STANDALONE_CORE_PATTERN = Pattern.compile("^core_(?<core>.*?)\\.(.*)$");
-
+  public static List<String> CLOUD_LABEL_KEYS = List.of("core", "collection", "shard", "replica");
+  public static List<String> STANDALONE_LABEL_KEYS = List.of("core");
```
mlbiscoc, line 34: I did a static List of the keys for each regex expression instead of the Map and looped through these to add the labels

PR #3020, opened 2025-01-09
SOLR-17556: Fix smokeTest to use films as example for slim distribution

HoustonPutman:
fixes for - [SOLR-17556](https://issues.apache.org/jira/browse/SOLR-17556)

(cherry picked from commit 80dfe11fe4c8f123f56ec2f85870d12b7e086e9e)

The initial commit fixed the smoke test script to not use the techproducts example for the slim distribution. In main the `test_utf8` script was already removed, so I removed that here as well. I also removed the data, which required changing a few tests to fit the 1 less document.

More important, I also added the ability to set the solr home directory for examples. this is important for testing, but is much better than modifying the server directory in solr.

Will forward port the differences from the original commit to main if it's approved.


epugh (commented):
Changes all make sense.

One thought?  Should the "slim" distribution come with examples?   I could understand that the slim one has no examples...    Maybe, if you wanted to smoketest the "slim" distribution, you would just index the data from teh regular distrubtion, but not include it?

HoustonPutman:
Yeah , I guess we could do that with the examples-dir option. Originally i had removed the examples when adding the slim distribution, but left it in to not disrupt the testing methodology too much

HoustonPutman:
For 9x we shouldnt remove it though. Just for 10

anshumg (approved):
This looks good to me. Thanks for taking care of this @HoustonPutman

dsmiley (approved):

gerlowskija (approved):

PR #3024, opened 2025-01-10
Fix smoketest for Solr 10

HoustonPutman:
No jira, as this is mainly infra work.

This fixes the smoke tests to use java 21, and java 25 as the extra java version for whenever that comes out. (future proofing)
It also fixes a few files that were added and removed

Two small things needed to be fixed for the release distribution as well.
The checksum defaulted to 256 accidentally, and the solrj/api jar manifests had the wrong source/target compatibility listed.


gerlowskija (approved):
LGTM, assuming this change is for `main` only and won't be backported to branch_9x?

Also, assuming you picked Java25 in some places as it's the next LTS after Java 21?

HoustonPutman:
> LGTM, assuming this change is for `main` only and won't be backported to branch_9x?
> 
> Also, assuming you picked Java25 in some places as it's the next LTS after Java 21?

Yes to both questions!

PR #3027, opened 2025-01-12
Move DOMUtil ConfigNode methods to ConfigNode, with some name changes.

dsmiley:
ConfigNode.requiredStrAttr -> attrRequired
ConfigNode.child (ex) -> childRequired
ConfigNode.getAll move name to first param; don't use empty set

PluginInfo: don't need the XML Node constructor anymore

A Solr 10-only change

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/PluginInfo.java
+++ b/solr/core/src/java/org/apache/solr/core/PluginInfo.java
@@ -22,6 +22,7 @@ import static java.util.Collections.unmodifiableMap;
 import static org.apache.solr.common.params.CoreAdminParams.NAME;
 import static org.apache.solr.schema.FieldType.CLASS_NAME;
 
+import com.google.common.annotations.VisibleForTesting;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
@@ -30,10 +31,9 @@ import java.util.List;
 import java.util.Map;
 import org.apache.solr.common.ConfigNode;
 import org.apache.solr.common.MapSerializable;
-import org.apache.solr.common.util.DOMUtil;
 import org.apache.solr.common.util.NamedList;
+import org.apache.solr.util.DOMConfigNode;
 import org.w3c.dom.Node;
-import org.w3c.dom.NodeList;
 
 /** An Object which represents a Plugin of any type */
 public class PluginInfo implements MapSerializable {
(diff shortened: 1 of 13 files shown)


solr/solrj/src/java/org/apache/solr/common/ConfigNode.java
```diff
@@ -93,9 +107,16 @@ default String attr(String name) {
     return attributes().get(name);
   }
 
-  default String requiredStrAttr(String name, Supplier<RuntimeException> err) {
+  /**
+   * Like {@link #attr(String)} but throws an error (incorporating {@code missing_err}) if not
+   * found.
+   */
+  default String attrRequired(String name, String missing_err) {
```
epugh, line 114: nit pick, but should this bee `missingErr`?   Thanks for the docs..

epugh (commented):
Not an area I understand well, but the diff changes did make things more legible, and i apprecaited the additional comments.

solr/core/src/java/org/apache/solr/core/PluginInfo.java
```diff
@@ -100,39 +100,23 @@ public String toString() {
     }
   }
 
+  /** From XML. */
```
epugh, line 103: Maybe extend just a bit? From schema.xml? It made me look for a similar constructor that was /** From JSON. */ further down...

dsmiley: This constructor supports XML config generally; not just a specific file as you wondered.  I added it because it may not be obvious to the reader who may be unfamiliar with `ConfigNode` but who would recognize XML `Node` type. The other constructors are more general; can't be said to be from any particular format.  I could drop this but was trying to be helpful.

solr/solrj/src/java/org/apache/solr/common/ConfigNode.java
dsmiley: Not my naming choice; I can rename

epugh: Thanks!!!!

solr/core/src/java/org/apache/solr/core/PluginInfo.java
epugh: i think it's fine.   Thanks!

epugh (approved):

dsmiley:
In the description, I said Solr 10 only but actually I see no reason why these **internal** changes couldn't be applied to branch_9x as well so I'll do that too.

PR #3030, opened 2025-01-14
build: remove gradle/ant-compat

dsmiley:
Refactoring: remove gradle/ant-compat scripts, replacing with simple logic elsewhere


gradle/maven/defaults-maven.gradle
dsmiley, line 1: The only real change in this file is limiting the `gradle.projectsEvaluated` section to only the `signing` section, which is now the only part that needs to be evaluated later, since it depends on extensions in the distribution module.

gradle/ant-compat/misc.gradle
```diff
@@ -1,24 +0,0 @@
...
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-
-// Exclude inner classes from testing.
-allprojects {
-  tasks.withType(Test) { task ->
-    exclude '**/*$*'
```
dsmiley, line 22: if the test run fails for running inner classes as tests then I'll add this simple exclude line to a suitable other gradle file.  It's not deserving of a dedicated gradle script!

dsmiley (commented):
Just trying to simplify the build a little.

epugh (approved):
Thanks for taking this on!  It's been on my list of housekeeping things, but I didn't have the gradle chops to tackle.

HoustonPutman (approved):
Looks good to me! (As long as everything works 😅)

I cherry picked this on top of my smoke test fixes, and tested the whole thing. It passed, so I guess it all works!

> ```
> SUCCESS! [0:33:19.258836]
> ```

dsmiley:
The only thing I want to do before merging is spot-check some generated Maven POMs.

dsmiley:
No delta to the POM, before & after.  I see no signing info in the POM though; not sure what to make of that.  I'll merge.

PR #3034, opened 2025-01-14
build: Fix Out of space in CodeCache error

HoustonPutman:
Fixing the same problem as https://github.com/apache/lucene/issues/14090

The full error is "VirtualMachineError: Out of space in CodeCache for adapters in gradle"

The fix is to increase the reservedCodeCacheSize from the default to 256m, which is what Lucene uses now.


epugh:
Thanks for this!  I've been seeing this recently too...

dsmiley (approved):

PR #3036, opened 2025-01-15
SOLR-17581: Less ZK watches in tests.

psalagnac:
https://issues.apache.org/jira/browse/SOLR-17581

# Description

Test creates lot of useless Zookeeper watches, because all waitForState() create a watch both on the collection state and the list of live nodes. Most of calls ignore live nodes when evaluating the predicate.

# Solution

Introduce new test variant of waitForState(), that does not wait on live node changes when we're only interested in the collection state.

# Tests

This is a test only change.

# Checklist


solr/solrj/src/test/org/apache/solr/common/cloud/PerReplicaStatesIntegrationTest.java
```diff
@@ -245,50 +245,42 @@ public void testMultipleTransitions() throws Exception {
...
+              "false".equals(collectionState.getProperties().get(PER_REPLICA_STATE)));
       CollectionAdminRequest.modifyCollection(
               COLL, Collections.singletonMap(PER_REPLICA_STATE, "true"))
           .process(cluster.getSolrClient());
-      cluster
-          .getZkStateReader()
-          .waitForState(
-              COLL,
-              5,
-              TimeUnit.SECONDS,
```
dsmiley, line 280: you removed the time specificity but maybe doesn't matter or is already the default?  I see there is still an overload for this

solr/test-framework/src/java/org/apache/solr/cloud/SolrCloudTestCase.java
```diff
@@ -207,6 +210,47 @@ protected static void waitForState(
...
+          .waitForState(
+              collection,
+              timeout,
+              timeUnit,
+              c -> {
+                state.set(c);
+                return predicate.test(c);
+              });
+    } catch (Exception e) {
+      fail(message + "\n" + e.getMessage() + "\nLast available state: " + state.get());
```
dsmiley, line 250: didn't propagate 'e'; can't we throw it before logging the last available state?

dsmiley (commented):
Nice!

solr/test-framework/src/java/org/apache/solr/cloud/SolrCloudTestCase.java
psalagnac: I wanted to do the same as existing variant of `waitForState()` (the one that watches both live nodes and the collection).
I think the two methods should be aligned. I think it makes sense to raise a JUnit specific error. Most of the time, the caught exception has no interesting details, it's just a timeout.

solr/solrj/src/test/org/apache/solr/common/cloud/PerReplicaStatesIntegrationTest.java
psalagnac: Good point. Default timeout is 45 seconds.
I don't know whether this test has a specific need or not, so I'm restoring the 5 seconds.

dsmiley (commented):
ZkStateReader.waitForState has the wait unit params immediately after the collection name.  I think the new waitForState you added should be consistent.  Keeping the functional arg at the end seems like better style as well.

psalagnac:
> ZkStateReader.waitForState has the wait unit params immediately after the collection name. I think the new waitForState you added should be consistent. Keeping the functional arg at the end seems like better style as well.

Not new but fair! I dislike parameter order since I've been looking at these methods.

dsmiley (approved):

PR #3037, opened 2025-01-16
SolrTestCaseJ4: don't reset HttpClient SSL stuff

dsmiley:
Only a test or two need to actually reset this.

AfterClass no longer calls:
* HttpClientUtil.resetHttpClientBuilder();
* Http2SolrClient.resetSslContextFactory();

----
The theme here is trimming away quirky one-off activities in STCJ4 AfterClass & BeforeClass

Planning to do main only.

### DIFF
--- a/solr/core/src/test/org/apache/solr/cloud/SSLMigrationTest.java
+++ b/solr/core/src/test/org/apache/solr/cloud/SSLMigrationTest.java
@@ -37,6 +37,7 @@ import org.apache.solr.common.params.SolrParams;
 import org.apache.solr.embedded.JettyConfig;
 import org.apache.solr.embedded.JettySolrRunner;
 import org.apache.solr.util.SSLTestConfig;
+import org.junit.After;
 import org.junit.Test;
 
 /**
@@ -46,6 +47,11 @@ import org.junit.Test;
 @SuppressSSL
 @AwaitsFix(bugUrl = "https://issues.apache.org/jira/browse/SOLR-12028") // 17-Mar-2018
 public class SSLMigrationTest extends AbstractFullDistribZkTestBase {
+
+  @After
+  public void afterTest() {
+    HttpClientUtil.resetHttpClientBuilder(); // also resets SocketFactoryRegistryProvider
+  }
 
   @Test
   public void test() throws Exception {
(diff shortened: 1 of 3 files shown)


dsmiley:
Planning to merge tomorrow.  Pretty straight-forward change.

HoustonPutman:
@dsmiley this broke `ConcurrentUpdateSolrClientBuilderTest.testSocketTimeoutOnCommit()` in a fun way. This PR works for all tests that extend `SolrTestCaseJ4`, because all of those tests expect to have the SSLConfig configured by `SolrTestCaseJ4.setupTestCases()`:

```java
sslConfig = buildSSLConfig();
// based on randomized SSL config, set SocketFactoryRegistryProvider appropriately
HttpClientUtil.setSocketFactoryRegistryProvider(
    sslConfig.buildClientSocketFactoryRegistryProvider());
```

Unfortunately, other tests, such as the one that is failing, extend `SolrTestCase`, and expect to have the default SSLConfig configured by default (a good assumption IMO). This test fails because the test ran before it extends `SolrTestCaseJ4` and the randomization selected HTTPS for that run. After the test succeeded, SolrTestCaseJ4 no longer reset the SSLConfig, and then  `ConcurrentUpdateSolrClientBuilderTest.testSocketTimeoutOnCommit()` tried to connect to an HTTP endpoint using the HTTPS SSLConfig, which fails.

I think we should just revert this and keep the original logic.

dsmiley:
Very impressive debugging Houston!

Apparently I overlooked STCJ4.setupTestCases SSL stuff.  Maybe I should undo the STCJ4 change (as you suggest), although I like the explicit changes to the 2 tests I also touched here.  I ultimately want to be able to switch many tests extending STCJ4 to use SolrTestCase instead.  But I don't want STC to become so bloated as STCJ4 is.  Perhaps the SSL matters can be encapsulated into a TestRule.  Oh I just noticed JettyConfig.Builder is referencing the public static field STCJ4.sslConfig -- _evil_!

(separate but related issue; should raise to the dev list) I sense we'd have an easier time with randomization if we change our approach for some scenarios like SSL (also PRS, non-Overseer).  Instead of flip-flopping from one test to the next, it'd be easier to track issues and debug if the entire build (all the tests selected) run a consistent SSL choice, and we either have a dedicated build or randomly choose this at the start.  If it's the whole build, we'd then use Develocity tags, thus aiding tracking patterns.  The choice could be determined in gradle and passed only when enabled.  A failure would have a reproducer line showing the additional flags.

dsmiley:
I reverted the STCJ4 change; mentioning this PR number in my direct commit fix.

And I filed this: https://issues.apache.org/jira/browse/SOLR-17666

PR #2923, opened 2024-12-21
SOLR-17609: Remove HDFS module

epugh:
https://issues.apache.org/jira/browse/SOLR-17609

# Description
Explore what removing HDFS module from Solr impacts.

# Solution

Searching for references for `hdfs`, `hadoop`, and other keywords.  Running tests.

- [x] Look at security policy last, once we get good green builds, what can we remove?
- [ ] Look at the Notice.txt, it refers to Apache Hadoop
- [x] Look at `libs.versions.toml`

# Tests

Only existing tests.

# Checklist

Please review the following and check all that apply:

- [x] I have reviewed the guidelines for [How to Contribute](https://github.com/apache/solr/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.
- [x] I have created a Jira issue and added the issue ID to my pull request title.
- [x] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended, not available for branches on forks living under an organisation)
- [x] I have developed this patch against the `main` branch.
- [x] I have run `./gradlew check`.
- [ ] I have added tests for my changes.
- [x] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

PR #3041, opened 2025-01-17
SOLR-17609:  mark deprecation of HDFS module in 9x and removal in 10

epugh:
https://issues.apache.org/jira/browse/SOLR-17609

# Description

This PR updates `branch_9x` to flag that we are deprecating HDFS and removing it in Solr 10.

Pairs with https://github.com/apache/solr/pull/2923 that goes on `main`.

# Solution

Only apply to branch_9x.  We should try and get the current "latest" ref guide updated with this PR to give folks a chance to know in case 9.9 never happens.

# Tests

### DIFF
--- a/solr/solr-ref-guide/modules/deployment-guide/pages/solr-on-hdfs.adoc
+++ b/solr/solr-ref-guide/modules/deployment-guide/pages/solr-on-hdfs.adoc
@@ -20,6 +20,8 @@
 The Solr HDFS Module has support for writing and reading Solr's index and transaction log files to the HDFS distributed filesystem.
 It does not use Hadoop MapReduce to process Solr data.
 
+IMPORTANT: The HDFS Module has been deprecated and will be retired in Solr 10.
+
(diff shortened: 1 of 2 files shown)


HoustonPutman (approved):

HoustonPutman:
Ok, the test failures are quite funny. We should be able to use our own deprecated code 😆

epugh:
> Ok, the test failures are quite funny. We should be able to use our own deprecated code 😆

removing the `forRemoval` flag...

solr/solr-ref-guide/modules/deployment-guide/pages/solr-on-hdfs.adoc
```diff
@@ -20,6 +20,8 @@
 The Solr HDFS Module has support for writing and reading Solr's index and transaction log files to the HDFS distributed filesystem.
 It does not use Hadoop MapReduce to process Solr data.
 
+IMPORTANT: The HDFS Module has been deprecated and will be removed in Solr 10.
```
dsmiley, line 23: There is asciidoc markup for important notices

dsmiley (approved):

PR #3042, opened 2025-01-17
SOLR-17540: mark deprecation and removal of Hadoop Auth

epugh:
https://issues.apache.org/jira/browse/SOLR-17540

# Description

Communicate to 9x users that Hadoop Auth is removed in 10 when it comes out.

# Solution

This goes with the already committed code in https://github.com/apache/solr/pull/2835 to `main` that removed Hadoop Auth.


dsmiley (approved):
Thanks

PR #3031, opened 2025-01-14
SOLR-17852: Migrate schema designer to filestoreapi part deux

epugh:
https://issues.apache.org/jira/browse/SOLR-17852

# Description

Migrate Schema Designer calls from using Blob store to FileStore

# Solution

Swap implementation methods.   

# Tests

So far, just using existing tests.

PR #3047, opened 2025-01-20
SOLR-17351: Decompose filestore "get file" API

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-17351

# Description

Solr's filestore "get file" API actually covers 4 or 5 distinct operations.  Depending on provided query parameters, it can:

- return filestore entry metadata (when meta=true is specified)
- instruct the receiving Solr node to pull a file from another node's filestore and cache it locally (when getFrom=someOtherNode is specified)
- instruct the receiving Solr node to push its cached copy of a file out to all other Solr nodes (when sync=true is specified)
- return raw file contents
- return JSON-ified file contents

This makes the code for this endpoint somewhat complex.  It also makes it hard to model and parse the response on the client side.

# Solution

This PR splits up the "get file" endpoint into a number of different APIs.  Specifically:

- metadata-fetching has been moved out to the endpoint, `GET/api/cluster/filestore/metadata/some/path.txt`
- Filestore commands such as pushing/pulling files are now available at: ` POST /api/cluster/filestore/commands`
- Support for "JSON-ified" file data has been dropped in this PR (but will be retained but deprecated in the eventual 9.x backport)

These divisions allow us to generate SolrRequest/SolrResponse classes representing these APIs, meaning that SolrJ users no longer need to use GenericSolrRequest/GenericSolrResponse.

# Tests

Some rewriting in TestDistribFileStore.  Existing package and filestore tests continue to pass.

# Checklist


gerlowskija:
Still TODO: ref-guide updates, additional testing.

epugh:
Does this impact #3031 ?   Since I am using `cc.getFileStore` I don't think so, but wanted to check.

gerlowskija:
That PR uses the Java interface, and not the HTTP APIs directly, so it should be fine afaict.

solr/core/src/java/org/apache/solr/filestore/ClusterFileStore.java
```diff
@@ -141,6 +151,103 @@ public UploadToFileStoreResponse uploadFile(
...
+                .map(details -> convertToResponse(details))
+                .collect(Collectors.toList());
+        dirListingResponse.files = Collections.singletonMap(path, directoryContents);
+        break;
+    }
+
+    return dirListingResponse;
+  }
+
+  // TODO Modify the filestore implementation itself to return this object, so conversion isn't
```
epugh, line 231: +1 to this

epugh: i think i also ran into this in the schema designer

solr/core/src/java/org/apache/solr/filestore/ClusterFileStore.java
```diff
@@ -194,6 +301,48 @@ public SolrJerseyResponse deleteFile(String filePath, Boolean localDelete) {
     return response;
   }
 
+  @Override
+  @PermissionName(PermissionNameProvider.Name.FILESTORE_WRITE_PERM)
+  public SolrJerseyResponse executeFileStoreCommand(String path, String getFrom, Boolean sync) {
```
epugh, line 306: aren't we trying to get rid of the "command" pattern in our new APIs?

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -189,26 +187,32 @@ private boolean fetchFileFromNodeAndPersist(String fromNode) {
       var solrClient = coreContainer.getDefaultHttpSolrClient();
 
       try {
-        GenericSolrRequest request = new GenericSolrRequest(GET, "/node/files" + getMetaPath());
```
epugh, line 192: yay!

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -364,11 +360,13 @@ private void distribute(FileInfo info) {
       for (String node : nodes) {
         String baseUrl =
             coreContainer.getZkController().getZkStateReader().getBaseUrlV2ForNodeName(node);
+
+        String nodeToFetchFrom;
         if (i < FETCHFROM_SRC) {
           // this is to protect very large clusters from overwhelming a single node
           // the first FETCHFROM_SRC nodes will be asked to fetch from this node.
           // it's there in  the memory now. So , it must be served fast
-          getFrom = myNodeName;
```
epugh, line 371: i like the nicer name

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -382,16 +380,14 @@ private void distribute(FileInfo info) {
           // trying to avoid the thundering herd problem when there are a very large no:of nodes
```
epugh, line 380: can we fix the comment?  `number of `

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -507,7 +503,8 @@ public void delete(String path) {
 
     final var solrParams = new ModifiableSolrParams();
     solrParams.add("localDelete", "true");
-    final var solrRequest = new GenericSolrRequest(DELETE, "/cluster/files" + path, solrParams);
+    final var solrRequest =
+        new GenericSolrRequest(DELETE, "/cluster/filestore/files" + path, solrParams);
```
epugh, line 507: boo!  Still have a GSR.

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
```diff
@@ -76,135 +66,54 @@ public SolrJerseyResponse getFile(String path, Boolean sync, String getFrom, Boo
     final var response = instantiateJerseyResponse(SolrJerseyResponse.class);
 
     if (Boolean.TRUE.equals(sync)) {
-      try {
-        fileStore.syncToAllNodes(path);
-        return response;
-      } catch (IOException e) {
-        throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Error getting file ", e);
-      }
+      ClusterFileStore.syncToAllNodes(fileStore, path);
```
epugh, line 69: I've been a bit confused between NodeFileStore, ClusterFileStore, DistribFileStore ;-).

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
```diff
@@ -76,135 +66,54 @@ public SolrJerseyResponse getFile(String path, Boolean sync, String getFrom, Boo
...
-    if (type == FileStore.FileType.DIRECTORY) {
-      final var directoryListingResponse =
-          instantiateJerseyResponse(FileStoreDirectoryListingResponse.class);
-      final var directoryContents =
-          fileStore.list(path, null).stream()
-              .map(details -> convertToResponse(details))
-              .collect(Collectors.toList());
-      directoryListingResponse.files = Collections.singletonMap(path, directoryContents);
-      return directoryListingResponse;
+    if (type == FileStore.FileType.NOFILE
```
epugh, line 83: this feels overly finicky logic..  why is it so hard to decide when to get metadata?  What are we missing that doens't just make this simple.    A thing has a meta thing.  That should be the only way.

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
```diff
@@ -76,135 +66,54 @@ public SolrJerseyResponse getFile(String path, Boolean sync, String getFrom, Boo
...
-                try {
-                  InputStream inputStream = it.getInputStream();
-                  if (inputStream != null) {
-                    jsonResponse.response = new String(inputStream.readAllBytes(), UTF_8);
-                  }
-                } catch (IOException e) {
-                  throw new SolrException(
-                      SolrException.ErrorCode.SERVER_ERROR, "Error reading file " + pathCopy);
+    // User wants to get the "raw" file
+    // TODO Should we be trying to json-ify otherwise "raw" files in this way?  It seems like a
```
epugh, line 90: What is the use case FOR this?   Do we have one right now?

solr/core/src/java/org/apache/solr/packagemanager/PackageManager.java
```diff
@@ -173,7 +173,7 @@ public void uninstall(String packageName, String version)
         String.format(Locale.ROOT, "/package/%s/%s/%s", packageName, version, "manifest.json"));
     for (String filePath : filesToDelete) {
       DistribFileStore.deleteZKFileEntry(zkClient, filePath);
-      String path = "/api/cluster/files" + filePath;
+      String path = "/api/cluster/filestore/files" + filePath;
```
epugh, line 176: i like the more verbose path I think...  Can a filestore have anything under than files?   If not, what about just `"/api/cluster/filestore"`

solr/solr-ref-guide/modules/configuration-guide/pages/package-manager-internals.adoc
```diff
@@ -115,7 +115,7 @@ openssl dgst -sha1 -sign my_key.pem runtimelibs.jar | openssl enc -base64 | sed
 +
 [source, bash]
 ----
-curl --data-binary @runtimelibs.jar -X PUT  http://localhost:8983/api/cluster/files/mypkg/1.0/myplugins.jar?sig=<signature-of-jar>
+curl --data-binary @runtimelibs.jar -X PUT  http://localhost:8983/api/cluster/filestore/files/mypkg/1.0/myplugins.jar?sig=<signature-of-jar>
```
epugh, line 118: maybe drops the `files`?

solr/core/src/java/org/apache/solr/filestore/ClusterFileStore.java
```diff
@@ -194,6 +301,48 @@ public SolrJerseyResponse deleteFile(String filePath, Boolean localDelete) {
...
+    }
+
+    return response;
+  }
+
+  public static void syncToAllNodes(FileStore fileStore, String path) {
+    try {
+      fileStore.syncToAllNodes(path);
+    } catch (IOException e) {
+      throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, "Error getting file ", e);
```
dsmiley, line 325: We don't know if we can blame the user

solr/core/src/java/org/apache/solr/filestore/DistribFileStore.java
```diff
@@ -189,26 +187,32 @@ private boolean fetchFileFromNodeAndPersist(String fromNode) {
...
+      } catch (Exception e) {
         throw new SolrException(SolrException.ErrorCode.SERVER_ERROR, "Error fetching metadata", e);
       } finally {
         org.apache.solr.common.util.IOUtils.closeQuietly(is);
       }
 
+      ByteBuffer filedata = null;
+      try {
+        final var fileRequest = new FileStoreApi.GetFile(path);
+        final var client = coreContainer.getSolrClientCache().getHttpSolrClient(baseUrl);
```
dsmiley, line 206: Please don't use the SolrClientCache for anything other than streaming expressions.  I'm stamping out such usages here: https://issues.apache.org/jira/browse/SOLR-17630
I'm particularly surprised to see you removed a use of the new `requestWithBaseUrl`.

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
gerlowskija: Yeah, the name conflicts are unfortunate.

On the bright side - I think we can probably make "NodeFileStore" and "NodeFileStoreApi" go away as part of this PR.  (Because with these APIs changes, all of our filestore endpoints will be under the "/cluster/filestore" path).

That would just leave:

- `ClusterFileStoreApi` - annotated interface defining our filestore APIs
- `ClusterFileStore` - implementations of those APIs
- `DistribFileStore` - internal implementation of various filestore operations (syncing, deleting, creating, etc.)

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
gerlowskija: Agreed.  I think the reason this is finicky is that we cram a bunch of different cases into the "metadata" POJO.  It's trying to represent too much.  It gets used not just for file-metadata. but also directory listings, representing 404s, etc.

A good refactor IMO would to:
1. create a `class DirectoryMetadata extends Metadata`, to use in the directory case.
2. Have the 404 case `throw new SolrException(ErrorCode.NOT_FOUND, "...");`

That said - this code all pre-exists this PR, and I'm leery to wade into refactors for scope reasons.

gerlowskija (commented):
Replies to some review comments.  Still working my way through review feedback.

epugh:
I enjoyed catching up on the comments on this PR!

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
gerlowskija: Not that I know of.  I'm going to nuke it based on the discussion [here](https://github.com/apache/solr/pull/3047/files#r1929771967), and that might flush out something I don't understand.

(We'll need to make sure we're only deprecating it in the 9.x backport, but that's not a big deal...)

solr/core/src/java/org/apache/solr/packagemanager/PackageManager.java
gerlowskija: > Can a filestore have anything under than files?

It can: "metadata" and "commands" are both under the "filestore" path as siblings to "files".

solr/solr-ref-guide/modules/configuration-guide/pages/package-manager-internals.adoc
gerlowskija: (See other comments - "files" is our way to distinguish between "metadata" and "commands", which are both under the "/filestore" path.)

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
gerlowskija: There's no discussion of it either in the PR or the JIRA that introduced it (unless I'm missing something?).  My guess is that it was an attempt at convenience?

In any case; I think this entire file can now go.  See my comment [above](https://github.com/apache/solr/pull/3047/files#r1929771967) for more details.

solr/core/src/java/org/apache/solr/filestore/NodeFileStore.java
gerlowskija: Alright, the most recent iteration removes NodeFileStore/NodeFileStoreApi.  There's still some room for confusion between ClusterFileStoreApi, ClusterFileStore, and DistribFileStore, but overall this PR leaves things better than it found it in that regard 👍

gerlowskija:
Alright - I think I've addressed the feedback so far?  If I've missed anything, let me know.  I've brought it up to date with 'main' and will aim to merge in the next few days pending any objections?

PR #3048, opened 2025-01-20
SOLR-17623: Simple ordered map should implement map

renatoh:
https://issues.apache.org/jira/browse/SOLR-17623

Implementing the interface Map on SimpleOrderedMap and adding Unit-Test for the new methods


epugh:
Nitpicky, but can you title the PR `SOLR-17623:`, that is the style we use...

renatoh:
@dsmiley  Could you please have a look at the implementation oaf the method from the Map interface?
Also, I had to make a few adjustment to other parts in the code to make it work with the new methods.
I need to add java-doc and tidy up the unit-tests but would be good to get some feedback before I take care of that.
Thanks

renatoh:
I am curious if there is a way to implement values() and entrySet(0 without an unchecked cast, e.g. (T) nvPairs.get(i)

dsmiley:
As the PR template indicates, the PR title must use exactly the JIRA syntax.  Thus all-caps and hyphen separating SOLR and 17623, probably then a colon.  If you do it correctly, moments later a bot will update the JIRA to link to the PR.  (albeit sometimes the bot is delayed inexplicably)

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -86,6 +91,79 @@ public static SimpleOrderedMap<Object> of() {
    * @return SimpleOrderedMap containing one key-value pair
    */
   public static <T> SimpleOrderedMap<T> of(String name, T val) {
-    return new SimpleOrderedMap<>(List.of(name, val));
+    return new SimpleOrderedMap<T>(List.of(name, val));
+  }
+
```
dsmiley, line 96: minor: (organizational) better to place these instance methods above the static methods that you recently added

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -86,6 +91,79 @@ public static SimpleOrderedMap<Object> of() {
...
+  @Override
+  public boolean isEmpty() {
+    return nvPairs.isEmpty();
+  }
+
+  @Override
+  public boolean containsKey(final Object key) {
+    return super.get((String) key) != null;
```
dsmiley, line 104: the value can actually be null in a NamedList, so this is inaccurate.  Use indexOf

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -86,6 +91,79 @@ public static SimpleOrderedMap<Object> of() {
...
+
+  @Override
+  public T get(final Object key) {
+    return super.get((String) key);
+  }
+
+  @Override
+  public T put(final String key, final T value) {
+    T oldValue = get(key);
+    add(key, value);
```
dsmiley, line 127: add will not replace the existing value.  This is why that method is called "add" (vs put or set).  I pointed you at NamedList asShallowMap which dealt with this.  This concern also led me to recommend deferring mutability.

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -86,6 +91,79 @@ public static SimpleOrderedMap<Object> of() {
...
+  }
+
+  @Override
+  public void putAll(final Map<? extends String, ? extends T> m) {
+    m.forEach(this::put);
+  }
+
+  @Override
+  public Set<String> keySet() {
+    var keys = new HashSet<String>();
```
dsmiley, line 143: Please don't create new datastructures when calling methods you add

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -86,6 +91,79 @@ public static SimpleOrderedMap<Object> of() {
...
+    for (int i = 0; i < nvPairs.size(); i += 2) {
+      keys.add((String) nvPairs.get(i));
+    }
+    return keys;
+  }
+
+  @Override
+  @SuppressWarnings({"unchecked"})
+  public Collection<T> values() {
```
dsmiley, line 152: Please don't create new datastructures when calling methods you add

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -86,6 +91,79 @@ public static SimpleOrderedMap<Object> of() {
...
+    var values = new ArrayList<T>();
+    for (int i = 1; i < nvPairs.size(); i += 2) {
+      values.add((T) nvPairs.get(i));
+    }
+    return values;
+  }
+
+  @Override
+  @SuppressWarnings({"unchecked"})
+  public Set<Entry<String, T>> entrySet() {
```
dsmiley, line 162: Please don't create new datastructures when calling methods you add

solr/solrj/src/test/org/apache/solr/common/util/SimpleOrderedMapTest.java
```diff
@@ -0,0 +1,125 @@
+package org.apache.solr.common.util;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class SimpleOrderedMapTest extends Assert {
```
dsmiley, line 9: No; in Solr we extend SolrTestCase at a minimum for a unit test

solr/solrj/src/test/org/apache/solr/common/util/SimpleOrderedMapTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+  private final SimpleOrderedMap<Integer> map = new SimpleOrderedMap<>();
```
dsmiley, line 11: why have a field for the map instead of each test merely creating the maps it needs?  A field means need a lifecycle like `@After` to clear.

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
renatoh: How else should I collect every other entry from the nvPairs into Set?
The only other way I can think of is to use a .stream()

solr/solrj/src/test/org/apache/solr/common/util/SimpleOrderedMapTest.java
renatoh: JUnit calls every  single test-method with a fresh instance of the test class, there is no need for a lifecycle with a clear.

dsmiley:
> I am curious if there is a way to implement values() and entrySet(0 without an unchecked cast, e.g. (T) nvPairs.get(i)

Almost nobody could answer a cast question like that without literally working on it in their IDE.  I'll take a look at a future iteration of yours but meanwhile, don't worry about it -- it's not important.

renatoh:
@dsmiley Considering your improvements on NamedList..asShallowMap, could I just delegate the method keySet, values and entrySet to the shallowMap e.g. :

  public Collection<T> values() {
   return super.asShallowMap(true).values();
  }

dsmiley:
I would implement the methods here, copying bits of code from #3050 if needed so that you don't call NamedList.asShallowMap.  Take from that whatever you want.  I'm not sure NamedList.asShallowMap was a wise idea since it can't implement the Map contract if the keys repeat.  It might have pre-dated SimpleOrderedMap.  Assuming you finish the PR here, I'm inclined to maybe make NamedList.asShallowMap do an instanceof check for SimpleOrderedMap and if so return "this" but otherwise throw an exception.  Or remove it.  Ideally we only use NamedList that isn't SimpleOrderedMap for the rare cases where it's appropriate.

renatoh:
@dsmiley Please have another look, thanks!

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -68,24 +74,153 @@ public SimpleOrderedMap(Map.Entry<String, T>[] nameValuePairs) {
...
+  }
+
+  @Override
+  public boolean containsKey(final Object key) {
+    return this.indexOf((String) key) >= 0;
   }
 
   /**
-   * Returns a shared, empty, and immutable instance of SimpleOrderedMap.
+   * Returns {@code true} if this map maps one or more keys to the specified value.
```
dsmiley, line 91: "one or more" -- why more?   Any way, don't feel obliged to re-specify javadocs for interface implementations since it's automatically inherited.

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -68,24 +74,153 @@ public SimpleOrderedMap(Map.Entry<String, T>[] nameValuePairs) {
...
+  }
+
+  /**
+   * return a {@link Collection} of all values in the map.
+   *
+   * @return {@link Collection} of all values in the map
+   */
+  @Override
+  public Collection<T> values() {
+    return new InnerMap().values();
```
dsmiley, line 178: I think it'd be clearer to return a new AbstractList (and can declare the return type to be List as well)

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -68,24 +74,153 @@ public SimpleOrderedMap(Map.Entry<String, T>[] nameValuePairs) {
...
+  }
+
+  /**
+   * return a {@link Set} of all keys in the map.
+   *
+   * @return {@link Set} of all keys in the map
+   */
+  @Override
+  public Set<String> keySet() {
+    return new InnerMap().keySet();
```
dsmiley, line 168: I think it'd be clearer to return a new AbstractList (and can declare the return type to be List as well)

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -68,24 +74,153 @@ public SimpleOrderedMap(Map.Entry<String, T>[] nameValuePairs) {
...
+  /**
+   * Returns a shared, empty, and immutable instance of {@link SimpleOrderedMap}.
+   *
+   * @return Empty {@link SimpleOrderedMap} (immutable)
+   */
+  public static SimpleOrderedMap<Object> of() {
+    return EMPTY;
+  }
+
+  private class InnerMap extends AbstractMap<String, T> {
```
dsmiley, line 220: If you follow my feedback above, we won't need this.

dsmiley (commented):
As part of this PR, let's have NamedList.asShallowMap do an instanceof check for Map and return it if so.  That will lead to real usage / code coverage of the changes you've done.

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
renatoh: the java-doc is automatically inherited? isn't that only the case if I use the {@inheritDoc} tag in the java doc itself. And even then, it does not inherit the param, return and throw part of the java-doc, unless I tag them one by one. Nevertheless, I will move  to {@inheritDoc} where possible, in order not to repeat java doc.
With respect to "one or more", I did not come up with that phrasing, that is on Map#contains, and although it sounds strange, it is technically correct.

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
renatoh: I've tried using an AbstractList / AbstractSet for values() / keySet(), but on the AbstractSet I need to implement an anonymous  Iterator<String> within the anonymous implementation of AbstractSet. Hence the approach with the InnerMap is much more concise.

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -68,24 +74,153 @@ public SimpleOrderedMap(Map.Entry<String, T>[] nameValuePairs) {
...
    *
-   * @return Empty SimpleOrderedMap (immutable)
+   * @param value value whose presence in this map is to be tested
+   * @return {@code true} if this map maps one or more keys to the specified value
    */
-  public static SimpleOrderedMap<Object> of() {
-    return EMPTY;
+  @Override
+  public boolean containsValue(final Object value) {
+    int sz = size();
```
dsmiley, line 98: instead of all these lines, could be a one-liner values().contains(value).  It'll be rare if ever for any Solr code to call this.

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
dsmiley: I better appreciate why this is here.  I don't love the name but I don't recommend something else.  I think it should have a comment explaining why this is here.  I'd say "SimpleOrderedMap can't extend AbstractMap, sadly.  This inner class can, and allows us to implement some of Map's methods via one-liner calls out to this.

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
dsmiley: If we don't have anything extra to say beyond the default javadoc, then add nothing at all (no javadoc).  If you want to add a tidbit of extra info to say something O(N) then you can use `{@inheritDoc}` and add the sentence.  But I think you could just augment the class level javadocs to warn users about usage of this as a general Map for large maps.

solr/solrj/src/test/org/apache/solr/common/util/SimpleOrderedMapTest.java
dsmiley: Ha; thanks for educating me!  Jeesh; I should know that.  Any way, for a trivial line, I'd just put it at the top of every test method but it's not a big deal; do as you wish.

solr/solrj/build.gradle
```diff
@@ -81,7 +81,7 @@ dependencies {
   testRuntimeOnly(libs.eclipse.jetty.alpnjavaserver, {
     exclude group: "org.eclipse.jetty.alpn", module: "alpn-api"
   })
-  testImplementation(libs.mockito.core, {
+  testImplementation(libs.mockito.core, {            hb
```
cpoerschke, line 84:
```suggestion
  testImplementation(libs.mockito.core, {
```

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -146,43 +126,21 @@ public T remove(final Object key) {
...
-  /**
-   * Copies all of the mappings from the specified map to this map. These mappings will replace any
-   * mappings that this map had for any of the keys currently in the specified map.
-   *
-   * @param m mappings to be stored in this map
-   * @throws NullPointerException if the specified map is null
-   */
   @Override
   public void putAll(final Map<? extends String, ? extends T> m) {
     m.forEach(this::put);
```
dsmiley, line 131: if you look at NamedList's shallow Map putAll impl, there's a faster implementation if the map is empty at the start.  We could incorporate that here with a simple check that calls `this::add`

solr/solrj/build.gradle
```diff
@@ -48,7 +48,6 @@ dependencies {
   })
   implementation libs.apache.httpcomponents.httpclient
   implementation libs.apache.httpcomponents.httpcore
-  implementation libs.apache.commons.lang3
```
dsmiley: weird; what's that doing in this PR?

dsmiley (approved):
Looks good; thank you!  Just needs a CHANGES.txt entry.

My hope is for NamedList.asShallowMap's implementation to go away, ultimately mandating that if you want a Map, you must have a SimpleOrderedMap.  That can be a separate issue/PR.

renatoh:
> As part of this PR, let's have NamedList.asShallowMap do an instanceof check for Map and return it if so. That will lead to real usage / code coverage of the changes you've done.

Wouldn't it be a better approach just to override asShallowMap in SimpleOrderedMap and return itself?
Additionally,I have noticed that org.apache.solr.common.util.NamedList#asShallowMap(boolean) is not called outside NamedList, do we want to set it to private?

dsmiley:
> Wouldn't it be a better approach just to override asShallowMap in SimpleOrderedMap and return itself?

:face-palm: of course; wow I have no excuse for myself.  Not enough coffee maybe.

> Additionally,I have noticed that org.apache.solr.common.util.NamedList#asShallowMap(boolean) is not called outside NamedList, do we want to set it to private?

**Separately** in another issue/pr, NamedList needs some more attention around this and toMap.  Definitely deprecating that one.

renatoh:
> **Separately** in another issue/pr, NamedList needs some more attention around this and toMap. Definitely deprecating that one.

I interpreted you thumbs up on my comment early as OK for both of the issue mentioned, hence I committed and pushed it, was too quick:). Shall I revert it?

solr/solrj/src/java/org/apache/solr/common/util/NamedList.java
```diff
@@ -238,6 +238,14 @@ public int indexOf(String name, int start) {
     return -1;
   }
 
+  /***
+   * Scans the list sequentially beginning at index 0
```
dsmiley, line 242: nitpick:  javadoc should include clarity that it's the index of the name.  It's not great to require looking at a parameter's name to disambiguate the meaning of a method.

solr/solrj/src/java/org/apache/solr/common/util/NamedList.java
```diff
@@ -403,8 +411,8 @@ public Map<String, T> asShallowMap() {
     return asShallowMap(false);
   }
 
-  public Map<String, T> asShallowMap(boolean allowDps) {
-    return new Map<String, T>() {
+  private Map<String, T> asShallowMap(boolean allowDps) {
```
dsmiley, line 414: now's not the time to make private but feel free to deprecate.

solr/solrj/src/java/org/apache/solr/common/util/NamedList.java
```diff
@@ -845,7 +853,7 @@ public void forEachKey(Consumer<String> fun) {
     }
   }
 
-  public void forEach(BiConsumer<String, ? super T> action) {
+  public void forEach(BiConsumer<? super String, ? super T> action) {
```
dsmiley, line 856: did doing this fix something?  it's weird.

solr/solrj/src/java/org/apache/solr/common/util/SimpleOrderedMap.java
```diff
@@ -68,24 +75,123 @@ public SimpleOrderedMap(Map.Entry<String, T>[] nameValuePairs) {
...
+  public void putAll(final Map<? extends String, ? extends T> m) {
+    if (isEmpty()) {
+      m.forEach(this::add);
+    } else {
+      m.forEach(this::put);
+    }
+  }
+
+  @Override
+  public Map<String, T> asShallowMap() {
```
dsmiley, line 139: nitpick but place this up above implementing Map interfaces rather than arbitrarily somewhere in the middle.

dsmiley (commented):
I have a slight concern if there's existing code relying on quirky sad behaviors of NamedList.asShallowMap having weird/sad implementations of `keySet`, `values`, `entrySet` because it calls `asMap(1)` on them.  Had it been `asMap(0)`, probably lower/no concern.  _Maybe_ this change should be in Solr 10 only.  On the other hand... maybe it's an unlikely impact?   Hard to say... right away I see SolrXmlConfig calling `.asShallowMap().entrySet()`.

renatoh:
> Just needs a CHANGES.txt entry.

I've added that under Solr 9.9.0, is that correct?

renatoh:
> have a slight concern if there's existing code relying on quirky sad behaviors of NamedList.asShallowMap having weird/sad implementations of `keySet`, `values`, `entrySet` because it calls `asMap(1)` on them. Had it been `asMap(0)`, probably lower/no concern. _Maybe_ this change should be in Solr 10 only. On the other hand... maybe it's an unlikely impact? Hard to say... right away I see SolrXmlConfig calling `.asShallowMap().entrySet()`.

Isn't SolrXmlConfig always calling .asShallowMap on  a NamedList, which is create org.apache.solr.common.ConfigNode#forEachChild?  So the changes do not make a difference here, or do they?

dsmiley:
I removed the asShallowMap returning SimpleOrderedMap so such a decision can be in Solr 10.
I intend on merging this when the tests pass and back porting to 9x.  Today/tomorrow.
Thanks again for your contribution!

renatoh:
> I removed the asShallowMap returning SimpleOrderedMap so such a decision can be in Solr 10. I intend on merging this when the tests pass and back porting to 9x. Today/tomorrow. Thanks again for your contribution!

You are welcome, thanks for your support!

Now it is failing with:
 Detected license header issues (skip with -Pvalidation.rat.failOnError=false):
  Unknown license: /home/runner/work/solr/solr/solr/solrj/src/test/org/apache/solr/common/util/SimpleOrderedMapTest.java
  
  No idea what this means

dsmiley:
As you are a very new contributor, I can tell by your changes and your last question that you have not been running `./gradlew precommit`, which will detect issues that you can correct before pushing.  
BTW all source files need a header.  IntelliJ does it automatically for me, as I've configured it to.

renatoh:
> As you are a very new contributor, I can tell by your changes and your last question that you have not been running `./gradlew precommit`, which will detect issues that you can correct before pushing.
> BTW all source files need a header. IntelliJ does it automatically for me, as I've configured it to.

Ah yes, I've forgot about that one, was for some reason under the impression running the 'tidy' is sufficient, now I know:)

PR #3050, opened 2025-01-21
SOLR-17552: NamedList.asShallowMap improvements

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17552


dsmiley:
@renatoh -- this is some WIP I did a couple months ago.  Looking at this may prove useful in #3048

PR #3057, opened 2025-01-23
Prepare for having Renovate add separate PRs for branch_9x

janhoy:
Now that main branch uses gradle version catalogs (#2706), dependency upgrades cannot be easily back ported to branch_9x anymore, unless the version catalogs are back ported as well.

Also, many features are removed in main that still exist in branch_9x, so we will miss out on certain dep updates if we only rely on manual backports.

Therefore I propose, and will test with dryRun, that @solrbot files separate PRs for branch_9x, using the [baseBranches feature](https://docs.renovatebot.com/configuration-options/#basebranches). For that to work we define [`useBaseBranchConfig=merge`](https://docs.renovatebot.com/configuration-options/#usebasebranchconfig) so that we can use `renovate.json` on this branch as an overlay, specifying configs that differ from main and giving a PR title suffix etc. Steps:

* Place a `renovate.json` config in branch_9x, which targets old `versions.properties` and acts as overlay
* In [solrbot repo](https://github.com/solrbot/renovate-github-action/pull/2), configure `baseBranches` and `useBaseBranchConfig`
* Do a dry run and verify that it works

Once Solr 10 is released, we can redefine this to target latest `branch_9_x` which will have a long life in bugfix mode.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,91 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "description": "Runs Renovate with solrbot for branch_9x, see dev-docs/dependency-upgrades.adoc for more",
+  "enabled": true,
+  "dependencyDashboard": false,
+  "enabledManagers": ["gradle", "github-actions"],
+  "includePaths": ["versions.*", "build.gradle", ".github/workflows/*"],
+  "postUpgradeTasks": {
+    "commands": ["./gradlew updateLicenses"],
+    "fileFilters": ["solr/licenses/*.sha1"],
+    "executionMode": "branch"
+  },
+  "packageRules": [
+    {
+      "description": "Fix for non-semantic versions for older artifacts",
+      "matchDatasources": ["maven"],
+      "matchPackageNames": [
+        "commons-collections:commons-collections",
+        "commons-io:commons-io",
+        "commons-lang:commons-lang"
+      ],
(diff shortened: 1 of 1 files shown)


epugh (approved):
LGTM..  This type of configuration always looks a bit magical to me!   Reading it though makes sense...

malliaridis (approved):
Having separate PRs for 9 would be great, as many PRs cannot be backported or require a multi-step update to update all branches accordingly, polluting the main branch with unnecessary commits. With separate PRs we will be able to close PRs that depend on a JDK greater than 11 or unsupported releases of libraries like Jetty. And we will still be able to update dependencies with minor and patch releases as before.

We still have "double the effort", but Solrbot would handle a large chunk of workload this way.

With `useBaseBranchConfig` do we no longer need to include the `packageRules` defined in `main`?

janhoy:
> With `useBaseBranchConfig` do we no longer need to include the `packageRules` defined in `main`?

Yes, but the branch-specific `renovate.json` inherits from the one on `main`, we only need to define settings that should override the main ones, so we'll inherit `packageRules`. At least according to docs and ChatGPT 🤣

janhoy:
This effort was tested, but Renovatebot went crazy and autoclosed all PRs, and no PRs were created for branch_9x.

Turns out, despite not being mentioned in the docs for baseBranches, this feature is not available in "forking" mode, which we use. Problem is that the bot uses github API to fork and sync the solr repo, but that code is only built to sync the main branch, not other baseBranches. But instead of exiting cleanly, it just tries stuff. Open issue: https://github.com/renovatebot/renovate/issues/7850

PR #3059, opened 2025-01-23
Remove broken and unused ZkStateReader.getReplicaProps()

psalagnac:
Remove broken and unused implementation of `getReplicaProps()` which is error prone.


solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -1002,30 +1002,12 @@ public static String getShardLeadersElectPath(String collection, String shardId)
...
-      Replica.State mustMatchStateFilter) {
-    return getReplicaProps(collection, shardId, thisCoreNodeName, mustMatchStateFilter, null);
-  }
-
-  public List<ZkCoreNodeProps> getReplicaProps(
-      String collection,
-      String shardId,
-      String thisCoreNodeName,
-      Replica.State mustMatchStateFilter,
-      Replica.State mustNotMatchStateFilter) {
```
psalagnac, line 1021: Parameter `mustNotMatchStateFilter` was dropped and not passed to the overloading method.

PR #3053, opened 2025-01-22
SOLR-17631: Upgrade main to Lucene 10.x

chatman:
[SOLR-17631](https://issues.apache.org/jira/browse/SOLR-17631)

Upgrade main to Lucene 10.x

Right now, I've fixed all compilation issues with core and test modules. Other modules still don't compile.
On the tests front, I've fixed a few tests. There are the tests that still fail:

```
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.update.processor.PreAnalyzedUpdateProcessorTest.txt
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.TestRandomDVFaceting.txt
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.DistributedIntervalFacetingTest.txt
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.analysis.TestReversedWildcardFilterFactory.txt
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.TestGroupingSearch.txt
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.TestJoin.txt
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.analysis.PathHierarchyTokenizerFactoryTest.txt
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.TestDistributedSearch.txt
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.BasicFunctionalityTest.txt
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.TestDistributedGrouping.txt
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.TestRandomFaceting.txt
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.cloud.BasicDistributedZkTest.txt
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.cloud.BasicDistributedZk2Test.txt
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.cloud.DocValuesNotIndexedTest.txt
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.handler.TestSnapshotCoreBackup.txt
./solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.handler.component.DistributedSpellCheckComponentTest.txt
[...]

PR #3121, opened 2025-01-24
remove deprecated KnnVectorField use

cpoerschke:
Split out from the Lucene 10.x upgrade #3053 PR.

The class was marked experimental in/until Lucene 9.4 and in Lucene 9.5 became deprecated:
* https://github.com/apache/lucene/blob/releases/lucene/9.4.2/lucene/core/src/java/org/apache/lucene/document/KnnVectorField.java#L36
* https://github.com/apache/lucene/blob/releases/lucene/9.5.0/lucene/core/src/java/org/apache/lucene/document/KnnVectorField.java#L33

So the removal here can go to both `main` and `branch_9x` branches with no JIRA or solr/CHANGES.txt entry needed in my opinion.


HoustonPutman (approved):
Small documentation comment

alessandrobenedetti (approved):
It's overall ok, just thinking if it's worth expanding the references we are adding here in regards to KnnFloatVectorField with also KnnByteVectorField, aside from that, it's ok to go!

PR #3140, opened 2025-01-27
SolrRequest.getParams never null; and clarify mutability

dsmiley:
SolrJ SolrRequest.getParams lacked javadocs and thus also clarity on null & mutability.  This PR addresses that.  It can return null but I don't want it to, so I made some changes and removed needless null checks.  Too much code implicitly assumes it's non-null but that's an iffy assumption.  Additionally mutability is unclear.  So I propose that if the return type is ModifiableSolrParams, then it's safe to modify it.  UpdateRequest had issues with this; I removed it's lazy instantiation so we can always count on an instance being returned that we can modify.

This PR introduces SolrParams.of(), the widespread usage of which can occur later.  Or I could do all of this piece in another PR if asked.

Many changes here aren't strictly required but felt good with these changes. 

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/api/V2ApiUtils.java
+++ b/solr/core/src/java/org/apache/solr/handler/api/V2ApiUtils.java
@@ -91,9 +91,6 @@ public class V2ApiUtils {
   }
 
   public static String getMediaTypeFromWtParam(SolrParams params, String defaultMediaType) {
-    if (params == null) {
-      return defaultMediaType;
-    }
     final String wtParam = params.get(WT);
     if (StrUtils.isBlank(wtParam)) return defaultMediaType;
 
--- a/solr/modules/cross-dc/src/java/org/apache/solr/crossdc/common/MirroredSolrRequest.java
+++ b/solr/modules/cross-dc/src/java/org/apache/solr/crossdc/common/MirroredSolrRequest.java
@@ -59,10 +59,9 @@ public class MirroredSolrRequest<T extends SolrResponse> {
   }
 
   public static class MirroredAdminRequest extends CollectionAdminRequest<CollectionAdminResponse> {
-    private ModifiableSolrParams params;
+    private SolrParams params;
 
-    public MirroredAdminRequest(
-        CollectionParams.CollectionAction action, ModifiableSolrParams params) {
+    public MirroredAdminRequest(CollectionParams.CollectionAction action, SolrParams params) {
       super(action);
       this.params = params;
     }
@@ -72,7 +71,7 @@ public class MirroredSolrRequest<T extends SolrResponse> {
       return params;
     }
 
-    public void setParams(ModifiableSolrParams params) {
+    public void setParams(SolrParams params) {
       this.params = params;
     }
 
@@ -247,10 +246,10 @@ public class MirroredSolrRequest<T extends SolrResponse> {
   }
 
   public static void setParams(SolrRequest<?> request, ModifiableSolrParams params) {
-    if (request instanceof MirroredAdminRequest) {
-      ((MirroredAdminRequest) request).setParams(params);
-    } else if (request instanceof UpdateRequest) {
-      ((UpdateRequest) request).setParams(params);
+    if (request instanceof MirroredAdminRequest mReq) {
+      mReq.setParams(params);
+    } else if (request instanceof UpdateRequest uReq) {
+      uReq.setParams(params);
     } else {
       throw new UnsupportedOperationException("Can't setParams on request " + request);
     }
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/request/AbstractUpdateRequest.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/request/AbstractUpdateRequest.java
@@ -21,10 +21,9 @@ import org.apache.solr.client.solrj.response.UpdateResponse;
 import org.apache.solr.common.params.ModifiableSolrParams;
 import org.apache.solr.common.params.UpdateParams;
 
-/** */
 public abstract class AbstractUpdateRequest extends CollectionRequiringSolrRequest<UpdateResponse>
     implements IsUpdateRequest {
-  protected ModifiableSolrParams params;
+  protected ModifiableSolrParams params = new ModifiableSolrParams(); // TODO make final
   protected int commitWithin = -1;
 
   public enum ACTION {
@@ -53,8 +52,6 @@ public abstract class AbstractUpdateRequest extends CollectionRequiringSolrReque
 
   public AbstractUpdateRequest setAction(
       ACTION action, boolean waitFlush, boolean waitSearcher, boolean softCommit, int maxSegments) {
-    if (params == null) params = new ModifiableSolrParams();
-
     if (action == ACTION.OPTIMIZE) {
       params.set(UpdateParams.OPTIMIZE, "true");
       params.set(UpdateParams.MAX_OPTIMIZE_SEGMENTS, maxSegments);
@@ -104,14 +101,11 @@ public abstract class AbstractUpdateRequest extends CollectionRequiringSolrReque
    * @since Solr 1.4
    */
   public AbstractUpdateRequest rollback() {
-    if (params == null) params = new ModifiableSolrParams();
-
     params.set(UpdateParams.ROLLBACK, "true");
     return this;
   }
 
   public void setParam(String param, String value) {
-    if (params == null) params = new ModifiableSolrParams();
     params.set(param, value);
   }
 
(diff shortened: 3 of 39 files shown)


solr/solrj/src/java/org/apache/solr/client/solrj/request/AbstractUpdateRequest.java
```diff
@@ -21,10 +21,9 @@
 import org.apache.solr.common.params.ModifiableSolrParams;
 import org.apache.solr.common.params.UpdateParams;
 
-/** */
 public abstract class AbstractUpdateRequest extends CollectionRequiringSolrRequest<UpdateResponse>
     implements IsUpdateRequest {
-  protected ModifiableSolrParams params;
+  protected ModifiableSolrParams params = new ModifiableSolrParams(); // TODO make final
```
dsmiley, line 26: Up for debate on eventually final.  That means we lose the setter, which is called in a ton of places, in lieu of calling `getParams().add(params)`.  I actually went down that path and shelved the work.  I was motivated for clarity on the getParams value never being potentially shared with something else.

solr/solrj/src/java/org/apache/solr/client/solrj/request/DirectXmlRequest.java
```diff
@@ -20,13 +20,15 @@
...
 import org.apache.solr.client.solrj.util.ClientUtils;
+import org.apache.solr.common.params.ModifiableSolrParams;
 import org.apache.solr.common.params.SolrParams;
 
 /**
  * Send arbitrary XML to a request handler
  *
  * @since solr 1.3
  */
+@Deprecated
```
dsmiley, line 31: There's a lot I want to deprecate and remove in Solr 10; this is one.  The annotation snuck in; arguably for another PR.  On the other hand, Solr isn't using or testing this class, so let's get on with removing!

solr/solrj/src/java/org/apache/solr/client/solrj/request/DocumentAnalysisRequest.java
```diff
@@ -85,7 +86,7 @@ protected DocumentAnalysisResponse createResponse(SolrClient client) {
   }
 
   @Override
-  public ModifiableSolrParams getParams() {
+  public SolrParams getParams() {
```
dsmiley, line 89: With new clarity from the Javadocs, we should never declare the return type as ModifiableSolrParams unless we have a consistent instance to return.

solr/solrj/src/java/org/apache/solr/client/solrj/request/GenericV2SolrRequest.java
```diff
@@ -40,8 +40,7 @@ public GenericV2SolrRequest(METHOD m, String path) {
    * @param params query parameter names and values for making this request.
    */
   public GenericV2SolrRequest(METHOD m, String path, SolrParams params) {
-    super(m, path);
-    this.params = params;
+    super(m, removeLeadingApiRoot(path), params);
```
dsmiley, line 43: @gerlowskija this includes a minor bug fix (to call removeLeadingApiRoot as the other overloaded constructor does) but maybe this code wasn't used

solr/solrj/src/java/org/apache/solr/client/solrj/request/JavaBinUpdateRequestCodec.java
```diff
@@ -128,7 +127,7 @@ public UpdateRequest unmarshal(InputStream is, final StreamingUpdateHandler hand
 
     // NOTE: if the update request contains only delete commands the params
     // must be loaded now
-    if (updateRequest.getParams() == null) {
+    if (updateRequest.getParams().iterator().hasNext() == false) { // no params
```
dsmiley, line 130: This check is evil; took a while to find it as I debugged a test for hours.  Behavior is changing based on the presence of params or not.  Like nowhere is there a similar check (at least based on my adventure with this PR).  The history behind these lines of code is what I consider to be an ugly/messy solution to propagating a commitWithin param.  Tempting to add that on my long TODO list of messes to fix but our longer future is not "javabin", and the mess thankfully seems localized here.

solr/solrj/src/java/org/apache/solr/client/solrj/request/QueryRequest.java
```diff
@@ -16,36 +16,39 @@
...
  * @since solr 1.3
  */
 public class QueryRequest extends CollectionRequiringSolrRequest<QueryResponse> {
 
-  private SolrParams query;
+  private final SolrParams query;
 
   public QueryRequest() {
     super(METHOD.GET, null);
+    query = new ModifiableSolrParams(); // TODO SolrParams.of()
```
dsmiley: at some point I eventually added SolrParams.of; TODO to use it in more places

dsmiley (commented):
Solr 10 only, albeit could backport Solr.of, for example.

solr/solrj/src/java/org/apache/solr/common/params/SolrParams.java
```diff
@@ -526,4 +526,9 @@ public String toString() {
     }
     return sb.toString();
   }
+
+  /** An empty, immutable SolrParams. */
+  public static SolrParams of() {
```
gerlowskija, line 531: [0] When I see `of()` it reminds me of the various var-arg methods with the same name that Java's "Collections" classes offer ([List.of](https://docs.oracle.com/javase/9/docs/api/java/util/List.html#of--)).

It's a little jarring to see that we don't accept any args here, and that this is only for creating empty param-sets.  Maybe a name like `emptyParams()` or `empty()` would convey things more accurately?

gerlowskija (approved):
Left a few inline comments, but happy enough with things to give this a 'LGTM'

solr/core/src/java/org/apache/solr/handler/api/V2ApiUtils.java
```diff
@@ -91,9 +91,6 @@ public static void squashIntoNamedListWithoutHeader(
   }
 
   public static String getMediaTypeFromWtParam(SolrParams params, String defaultMediaType) {
-    if (params == null) {
```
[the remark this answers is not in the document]
dsmiley: Okay I'll remove.  I suppose the scope of this PR should be more focused on SolrJ QueryRequest not server side SolrQueryRequest.  Granted, I'd wish for the same for SolrQueryRequest.getParams; same rationale for doing it in SolrJ.  I _could_ increase the scope to that but maybe too much in one PR; not sure.

solr/solrj/src/java/org/apache/solr/client/solrj/request/DirectXmlRequest.java
dsmiley: I want to consolidate deprecations elsewhere so I will remove this change from the PR.  We can discuss when I post that.

solr/modules/cross-dc/src/java/org/apache/solr/crossdc/common/MirroredSolrRequest.java
```diff
@@ -247,10 +246,10 @@ public boolean equals(final Object o) {
   }
 
   public static void setParams(SolrRequest<?> request, ModifiableSolrParams params) {
-    if (request instanceof MirroredAdminRequest) {
-      ((MirroredAdminRequest) request).setParams(params);
-    } else if (request instanceof UpdateRequest) {
-      ((UpdateRequest) request).setParams(params);
+    if (request instanceof MirroredAdminRequest mReq) {
```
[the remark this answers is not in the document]
dsmiley: yeah it's a very nice Java 16 thing

solr/solrj/src/java/org/apache/solr/common/params/SolrParams.java
dsmiley: Coincidentally I recently had the same conversation for the new SimpleOrderedMap.of.    It's modelled off JDK data structure utilities using this name, thus its consistent.  Once we see additional overloads, I think you'll like it better.  I could add one or two in this PR.

solr/solrj/src/java/org/apache/solr/client/solrj/request/AbstractUpdateRequest.java
dsmiley: It's easy for me to enforce that setParams arg isn't null.  I'll do that.  setParams is called kinda a lot BTW, although I feel it's usually not needed.  I can do a follow-up draft PR to show you what it looks like if *nobody* calls it.

solr/core/src/java/org/apache/solr/handler/api/V2ApiUtils.java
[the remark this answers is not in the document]
dsmiley: It turns out SolrQueryRequest.getParams will always be non-null (because `org.apache.solr.servlet.SolrRequestParsers#parseQueryString(java.lang.String)` is what initializes it, always non-null) but this fact is not documented.  I'll increase the scope here a little to cover that since there's little to it.

dsmiley:
Not thinking a JIRA/CHANGES.txt is worth it but will do if asked.
Proposed commit message:
> Clarify getParams not null in SolrRequest, SolrQueryRequest, QParser
> Clarify mutability of params for SolrRequest (SolrJ).  UpdateRequest changed; doesn't lazy-create params anymore.
> New SolrParams.of() and of(key, value) for an empty or single-pair instance.

Planning to merge Tuesday if no further comments.

solr/core/src/java/org/apache/solr/handler/api/V2ApiUtils.java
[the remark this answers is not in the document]
dsmiley: I brought this back and audited with data flow analysis that it'll never be null.

dsmiley:
Planning on merging this tomorrow night.  LMK if I should hold up.

PR #3141, opened 2025-01-27
SolrParams.equals implementation

dsmiley:
Motivated by seeing sad intermediate conversions to a NamedList.

The details of this PR _might_ depend on my other one today on ensuring SolrRequest.getParams never returns null -- #3140 (and is what motivated me to work on that).  That one is 10 only; so this one can be considered 10 only as well.  Or we'd need a general utility like SolrParams.equals(SolrParams a, SolrParams b) that has null-safe args, but I'd not love to add it as it works around what #3140 solves.

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/common/params/SolrParams.java
+++ b/solr/solrj/src/java/org/apache/solr/common/params/SolrParams.java
@@ -24,6 +24,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 import org.apache.solr.client.solrj.util.ClientUtils;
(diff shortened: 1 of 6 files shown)


solr/solrj/src/java/org/apache/solr/common/params/SolrParams.java
```diff
@@ -526,4 +527,36 @@ public String toString() {
...
+    // iterating this params, see if other has the same values for each key
+    int count = 0;
+    for (Entry<String, String[]> thisEntry : this) {
+      String name = thisEntry.getKey();
+      if (!Arrays.equals(thisEntry.getValue(), b.getParams(name))) return false;
+      count++;
+    }
+    // does other params have the same number of keys?  It might have more but not less.
+    Iterator<String> bNames = b.getParameterNamesIterator();
+    while (bNames.hasNext()) {
```
renatoh, line 549: It looks to me, like this equals method is not symmetric:.equals(y) must return the same result as y.equals(x).

dsmiley: Can you suggest improvements to `testTrivialEquals` to prove your theory?

renatoh: I could not find a test case where x.equals(y) is not y.equals(y)
I missed the fact, that params with the same key are combined into the same Array, my bad!

dsmiley:
Planning on merging Saturday.  No JIRA/CHANGES.txt.

PR #3145, opened 2025-01-29
SOLR-17438 Resolve committer GPG key from other source

janhoy:
https://issues.apache.org/jira/browse/SOLR-17438

### DIFF
--- a/dev-tools/scripts/releaseWizard.py
+++ b/dev-tools/scripts/releaseWizard.py
@@ -65,7 +65,7 @@ except:
 import scriptutil
 from consolemenu import ConsoleMenu
 from consolemenu.items import FunctionItem, SubmenuItem, ExitItem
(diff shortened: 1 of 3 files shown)


madrob:
I've heard rumblings that this data is going to move away from whimsy and onto infra-reports.a.o, so it may behoove us to wait a bit instead of switching now and then switching again. @dfoulks1 is working on that, he said he can follow up with us about it next week.

janhoy:
Any news on final location @madrob ?

madrob:
I linked to a an infra issue on JIRA, it looks like there hasn't been any work yet. I may have been too optimistic about software development timelines, we may as well finish this now and then move again _eventually_, rather than waiting indefinitely.

janhoy:
Ok, will commit this then. It is fairly simple to modify should the file move

PR #3146, opened 2025-01-29
SOLR-XXXXX: Support different filesystem implementations with EmbeddedSolrServer

AndreyBozhko:

# Description

EmbeddedSolrServer has a public constructor `EmbeddedSolrServer(Path solrHome, String defaultCoreName)` - which suggests that any Path implementation is acceptable (e.g., ZipPath, S3Path, and not just UnixPath/WindowsPath).

# Solution

Updated ManagedIndexSchemaFactory code that deals with schema paths to be more idiomatic.

# Tests

Added a test for instantiating EmbeddedSolrServer using a ZipPath instance.

# Checklist

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -34,6 +34,8 @@ Improvements
 
 * SOLR-17516: `LBHttp2SolrClient` is now generic, adding support for `HttpJdkSolrClient`. (James Dyer)
 
+* SOLR-XXXXX: Support different filesystem implementations with EmbeddedSolrServer (Andrey Bozhko)
+
 Optimizations
 ---------------------
 * SOLR-17568: The CLI bin/solr export tool now contacts the appropriate nodes directly for data instead of proxying through one.
--- a/solr/core/src/java/org/apache/solr/schema/ManagedIndexSchemaFactory.java
+++ b/solr/core/src/java/org/apache/solr/schema/ManagedIndexSchemaFactory.java
@@ -23,7 +23,6 @@ import java.io.InputStream;
 import java.lang.invoke.MethodHandles;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.Map.Entry;
 import net.jcip.annotations.NotThreadSafe;
@@ -148,12 +147,11 @@ public class ManagedIndexSchemaFactory extends IndexSchemaFactory implements Sol
    */
   public Path lookupLocalManagedSchemaPath() {
     final Path legacyManagedSchemaPath =
-        Paths.get(
-            loader.getConfigPath().toString(),
-            ManagedIndexSchemaFactory.LEGACY_MANAGED_SCHEMA_RESOURCE_NAME);
+        loader
+            .getConfigPath()
+            .resolve(ManagedIndexSchemaFactory.LEGACY_MANAGED_SCHEMA_RESOURCE_NAME);
 
-    Path managedSchemaPath =
-        Paths.get(loader.getConfigPath().toString(), managedSchemaResourceName);
+    Path managedSchemaPath = loader.getConfigPath().resolve(managedSchemaResourceName);
 
     // check if we are using the legacy managed-schema file name.
     if (Files.exists(legacyManagedSchemaPath)) {
--- a/solr/core/src/test/org/apache/solr/client/solrj/embedded/TestEmbeddedSolrServerConstructors.java
+++ b/solr/core/src/test/org/apache/solr/client/solrj/embedded/TestEmbeddedSolrServerConstructors.java
@@ -17,7 +17,12 @@
 package org.apache.solr.client.solrj.embedded;
 
 import java.io.IOException;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+import org.apache.commons.io.file.PathUtils;
 import org.apache.solr.SolrTestCaseJ4;
 import org.apache.solr.client.solrj.SolrQuery;
 import org.apache.solr.client.solrj.request.CoreAdminRequest;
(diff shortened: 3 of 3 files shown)


solr/core/src/test/org/apache/solr/client/solrj/embedded/TestEmbeddedSolrServerConstructors.java
```diff
@@ -58,4 +63,63 @@ public void testNodeConfigConstructor() throws Exception {
...
   }
+
+  @SuppressWarnings("removal")
+  @Test
+  public void testPathConstructorZipFS() throws Exception {
+    assumeTrue(
+        """
+                    Test only works without Security Manager due to SecurityConfHandlerLocal
+                    missing permission to read /1/2/3/4/security.json file""",
+        System.getSecurityManager() == null);
```
AndreyBozhko, line 74: To make the test work regardless of the security manager, this line https://github.com/apache/solr/blob/a5b0b98a79559021e44d00b43fff6c020c54f829/solr/core/src/java/org/apache/solr/handler/admin/SecurityConfHandlerLocal.java#L43 needs to be updated to handle Path in a more idiomatic way.

However, that requires updating CoreContainer#getSolrHome to return a Path instead of a String - so it looks like a much bigger change.

solr/core/src/java/org/apache/solr/schema/ManagedIndexSchemaFactory.java
```diff
@@ -148,12 +147,11 @@ public String lookupZKManagedSchemaPath() {
    */
   public Path lookupLocalManagedSchemaPath() {
     final Path legacyManagedSchemaPath =
-        Paths.get(
```
madrob, line 151: We should try to remove usages of Paths.get generally, I remember it having some wonky behaviour in the past. Happy to see this coming in.

dsmiley: Absolutely; same with Path.of.  _Ultimately_ we should ForbiddenApi check this so that it's rare to call this except for a few authorized places.

solr/core/src/test/org/apache/solr/client/solrj/embedded/TestEmbeddedSolrServerConstructors.java
dsmiley: CoreContainer.solrHome is a Path so hopefully it won't be too huge to switch it's getter to use Path.

dsmiley (commented):
This is really about SOLR-17548 / SOLR-16903 / SOLR-8282 / SOLR-17548.  I don't think we need separate JIRAs for separate File->Path or String->Path matters, but separate PRs is great.  Not only do we need to convert File -> Path, but we need to change most String based paths to Path.  Perhaps PR, by PR we focus on one particular String.  I don't think this PR is *really* about EmbeddedSolrServer (you put that in the title).

dsmiley:
Probably need to wait for @mlbiscoc in https://github.com/apache/solr/pull/2924 which is touching lots of stuff.  Then pick a particular String based API endpoint and convert it in an isolated PR relating to that one method or class (if there are multiple).  I think it's easier to review like this rather than touching everything.

epugh (commented):
Really interesting experiment ;-). ZipFS kind of neat!

solr/CHANGES.txt
```diff
@@ -34,6 +34,8 @@ Improvements
 
 * SOLR-17516: `LBHttp2SolrClient` is now generic, adding support for `HttpJdkSolrClient`. (James Dyer)
 
+* SOLR-XXXXX: Support different filesystem implementations with EmbeddedSolrServer (Andrey Bozhko)
```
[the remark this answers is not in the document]
dsmiley: I believe he put this here as a TODO.  We have quite a number of existing JIRAs to pick from for this theme :-).

solr/core/src/test/org/apache/solr/client/solrj/embedded/TestEmbeddedSolrServerConstructors.java
AndreyBozhko: OK, I think I will keep this PR as draft for now, and work on updating CoreContainer#getSolrHome to return Path first. Is there an existing JIRA for that?

AndreyBozhko: Removed the assumption about the SM now that #3204 is merged - the tests should be able to pass now.

solr/core/src/test/org/apache/solr/client/solrj/embedded/TestEmbeddedSolrServerConstructors.java
```diff
@@ -58,4 +63,58 @@ public void testNodeConfigConstructor() throws Exception {
...
+    //
+    // configset.zip
+    // └── 1
+    //     └── 2
+    //         └── 3
+    //             └── 4
+    //                 ├── data
+    //                 │   └── core1
+    //                 │       ├── conf
+    //                 │       │   ├── managed-schema.xml
```
dsmiley, line 85: BTW if you actually did want to launch Solr from a ZIP; the choice of a "managed" schema is dubious as it implies the mutability of the schema, yet the zip is read-only.  In my experience, I've only seen people actually use a managed schema as-intended for POCs.  I suppose you worked around this with the memory based StorageIO thing.

dsmiley (commented):
I suppose this PR is really a proof-of-concept to explore the gaps?

AndreyBozhko:
Indeed, this PR came out of the POC I did some time ago.

So if people think there's some value in the unit test I added, then we can move forward with this PR.

Otherwise, I can close it, and open a separate PR just for the changes to the ManagedIndexSchemaFactory. Or perhaps do that only after accumulating more Java NIO cleanups.

dsmiley:
>  Or perhaps do that only after accumulating more Java NIO cleanups.

-- this, I think.  _Then_ maybe commit the unit test as it's kind of neat; shows how to do something, and that's it's possible.  I could imagine a use-case to automate embedded Solr index building maybe.  Nonetheless; the primary / real support for Solr NIO is the DirectoryFactory -- https://issues.apache.org/jira/browse/SOLR-17646

AndreyBozhko:
I updated the PR to only add the unit test, as a demonstration of what's possible.

AndreyBozhko:
GHA failure due to unrelated test BasicDistributedZk2Test

AndreyBozhko:
Hi @dsmiley I believe this unit test now works as expected - do you think it's OK to merge it in?

dsmiley:
Judt did; sure, why not.

PR #3147, opened 2025-01-29
SOLR-17637: Fix LBHttpSolrClient & HttpShardHandler bug

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17637


markrmiller:
Nice catch. Patch looks good.

dsmiley:
Just curious; I suppose this bug might show as an NPE?

HoustonPutman:
> Just curious; I suppose this bug might show as an NPE?

It doesnt show at all. The thread thats running the async request fails, and since its not in the catch, it doesnt pass the exception to the completion step. So the thread just goes away without logging anything. Took a while to pin it down because of this.

risdenk (approved):

anshumg (approved):

gus-asf:
Nice catch! thanks!

gus-asf (approved):

PR #3148, opened 2025-01-31
Add NavigableObject.wrap. Deprecate MapWriter.EMPTY and MapWriterMap.

dsmiley:
Introduce NavigableObject.wrap to coerce its arg to a NavigableObject.  Maybe call this method "of"?
Deprecate MapWriter.EMPTY; don't need it anymore (due to the above); was only used once.
Enhance Utils to not need MapWriterMap for its functions, and to slightly optimize for SimpleOrderedMap becoming a Map soon.
Deprecate MapWriterMap.  Ultimately may simply make package scope and maybe rename to orient itself around it's NavigableObject-ness, which is what it's only used for, after the changes here.

No JIRA: too internal; low impact.


dsmiley:
Will merge in a few days.

PR #3149, opened 2025-01-31
in tests replace deprecated IndexSearcher.doc() calls

cpoerschke:
The API is deprecated since Lucene 9.5 - https://github.com/apache/lucene/blob/releases/lucene/9.5.0/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java#L382-L391 -- and remove in Lucene 10.

So the replacement here can go to both `main` and `branch_9x` branches with no JIRA or solr/CHANGES.txt entry needed in my opinion.

### DIFF
--- a/solr/core/src/test/org/apache/solr/legacy/TestNumericRangeQuery32.java
+++ b/solr/core/src/test/org/apache/solr/legacy/TestNumericRangeQuery32.java
@@ -183,12 +183,12 @@ public class TestNumericRangeQuery32 extends SolrTestCase {
       ScoreDoc[] sd = topDocs.scoreDocs;
       assertNotNull(sd);
       assertEquals("Score doc count" + type, count, sd.length);
-      Document doc = searcher.doc(sd[0].doc);
+      Document doc = searcher.storedFields().document(sd[0].doc);
       assertEquals(
           "First doc" + type,
           2 * distance + startOffset,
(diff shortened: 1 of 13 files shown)


dsmiley (approved):
In a number (all?) places here, you have a SolrIndexSearcher so could call `getDocFetcher().doc(id)` which is the Solr way to do this, not calling `storedFields()`.  I leave it to you to contemplate which path makes the most sense based on the test.  But I think the solr-way is preferred unless the test is particularly low-level (like "uninverting" package.

solr/core/src/test/org/apache/solr/legacy/TestNumericRangeQuery32.java
```diff
@@ -183,12 +183,12 @@ private void testRange(int precisionStep) throws Exception {
       ScoreDoc[] sd = topDocs.scoreDocs;
       assertNotNull(sd);
       assertEquals("Score doc count" + type, count, sd.length);
-      Document doc = searcher.doc(sd[0].doc);
+      Document doc = searcher.storedFields().document(sd[0].doc);
```
cpoerschke, line 186: Here we have an `IndexSearcher` i.e. not a `SolrIndexSearcher` and so not using `SolrIndexSearcher.getDocFetcher()` here.

PR #3150, opened 2025-01-31
SOLR-17645: Gracefully handle exceptions in executor threads

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17645

The main one I see in our tests are threads submitted by ZkStateReader.waitForState(), and ZkShardTerms having issues while a cluster is shutting down. I'll be looking for more and trying to address in this one PR.


dsmiley (approved):
The 2nd-attempt retry makes this complicated.

dsmiley (approved):
Not sure if a warning is properly loud enough but I'll trust you.

PR #3155, opened 2025-02-02
SOLR-17648: multiThreaded=true: changed queue implementation

dsmiley:
from unlimited to 1000 max, after which the caller thread will execute.
 Didn't need the RejectedExecutionException avoidance hack anymore; Lucene 9.12 has it.

https://issues.apache.org/jira/browse/SOLR-17648

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
+++ b/solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
@@ -34,7 +34,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -228,25 +229,15 @@ public static ExecutorService initCollectorExecutor(NodeConfig cfg) {
...
-              put(runnable); // blocks
-            } catch (InterruptedException e) {
-              Thread.currentThread().interrupt();
-              throw new RuntimeException("interrupted submitting to search multi-threaded pool", e);
-            }
-            return true;
-          }
-        },
+        new LinkedBlockingQueue<>(
+            EnvUtils.getPropertyAsInteger("solr.searcherCollector.queueSize", 1000)),
```
sigram, line 240: Needs docs?

dsmiley: IMO we don't need to document _obscure_ settings (I think this is).  If you think it should be documented, let me know where it should be.  If this queue size is exceeded, the outcome is still a working system but the performance will be somewhat less than what it otherwise would be.  FWIW 1000 feels about right to me.  I reviewed the code in Lucene 9.12 that cleverly uses the parent thread.

I plan to change this property to be "solr.search.multiThreaded.queueSize" so as to have a category after solr. -- see https://cwiki.apache.org/confluence/display/SOLR/System+property+naming+structure

PR #3157, opened 2025-02-04
replace (deprecated) Top(Field,ScoreDoc)Collector.create calls

cpoerschke:
Split out from the Lucene 10.x upgrade #3053 PR.

The `create` API is deprecated since Lucene 9.10 and removed in Lucene 10:
* https://github.com/apache/lucene/blob/releases/lucene/9.10.0/lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java#L392
* https://github.com/apache/lucene/blob/releases/lucene/9.10.0/lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java#L210

So the replacement here can go to both `main` and `branch_9x` branches with no JIRA or solr/CHANGES.txt entry needed in my opinion.


epugh (approved):
I don't have a lot of Lucene level knowledge, but as a straight up refactor it LGTM.

psalagnac (commented):
There is a difference between old and new code.

Old code (with the `create()` factory method) creates instances with parameter `supportsConcurrency == false`, while new code uses constructors that implicitly set `supportsConcurrency == true`.

This goes beyond my Lucene knowledge, I'm not sure exactly what is the impact. But if that's an unwanted change, I recommend to specify this parameter using another constructor.

cpoerschke:
> There is a difference between old and new code. ...

Good catch, thank you! Yes, this is an unwanted change (and it arose from splitting off the Lucene 10 branch and since the `supportsConcurrency` argument goes away in Lucene 10, my bad) -- will update to use another constructor as suggested.

psalagnac (approved):

cpoerschke:
Thanks for the reviews here!

Just noting retrospectively that my intention here had been to (slightly) reduce the scope of the #3053 changes (and for `branch_9x` to get the changes in this PR too) but then it sort of dropped off the radar somehow, sorry. Now no-longer-applicable since PR 3053 is merged.

PR #3158, opened 2025-02-04
SOLR-17649: Fix Json faceting on  multivalue number types

thomaswoeckinger:
https://issues.apache.org/jira/browse/SOLR-17649

# Description

FacetFieldProcessorByArrayDV can not handle fields with DocValues of type DocValuesType.SORTED_NUMERIC.
If such a type is requested for faceting FacetFieldProcessorByHashDV must be used.

# Solution

Use the correct FacetFiledProcessor

# Tests

Create multi valued EnumFileds, which had DocValues enabled, to ensure the use of  DocValuesType.SORTED_NUMERIC, index a few multivalued docs and query the facets afterwards

# Checklist

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -187,7 +187,7 @@ Bug Fixes
 
 * SOLR-17652: Fix a bug that could cause long leader elections to leave PULL replicas in DOWN state forever. (hossman)
 
-* SOLR-17649: Fix a regression of faceting on multi-valued EnumFieldType, introduced by an earlier 9.x version. (Thomas Wöckinger, David Smiley, Houston Putman, Kevin Risden)
+* SOLR-17649: Fix a regression of faceting on multi-valued EnumFieldType, introduced by an earlier 9.x version. (Thomas Wöckinger, David Smiley)
 
 Dependency Upgrades
 ---------------------
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -187,6 +187,8 @@ Bug Fixes
 
 * SOLR-17652: Fix a bug that could cause long leader elections to leave PULL replicas in DOWN state forever. (hossman)
 
+* SOLR-17649: Fix a regression of faceting on multi-valued EnumFieldType, introduced by an earlier 9.x version. (Thomas Wöckinger, David Smiley, Houston Putman, Kevin Risden)
+
 Dependency Upgrades
 ---------------------
 * SOLR-17471: Upgrade Lucene to 9.12.1. (Pierre Salagnac, Christine Poerschke)
--- a/solr/core/src/java/org/apache/solr/search/facet/FacetField.java
+++ b/solr/core/src/java/org/apache/solr/search/facet/FacetField.java
@@ -20,7 +20,6 @@ import java.util.HashMap;
 import java.util.Map;
 import org.apache.solr.common.SolrException;
 import org.apache.solr.schema.FieldType;
-import org.apache.solr.schema.NumberType;
 import org.apache.solr.schema.SchemaField;
 
 public class FacetField extends FacetRequestSorted {
@@ -95,9 +94,9 @@ public class FacetField extends FacetRequestSorted {
       return new FacetFieldProcessorByArrayDV(fcontext, this, sf);
     }
 
-    NumberType ntype = ft.getNumberType();
+    final boolean isNumber = ft.getNumberType() != null;
     // ensure we can support the requested options for numeric faceting:
-    if (ntype != null) {
+    if (isNumber) {
       if (prefix != null) {
         throw new SolrException(
             SolrException.ErrorCode.BAD_REQUEST,
@@ -139,11 +138,11 @@ public class FacetField extends FacetRequestSorted {
     // FieldType.getDocValuesType()
 
     if (!multiToken) {
-      if (mincount > 0 && prefix == null && (ntype != null || method == FacetMethod.DVHASH)) {
+      if (mincount > 0 && prefix == null && (isNumber || method == FacetMethod.DVHASH)) {
         // TODO can we auto-pick for strings when term cardinality is much greater than DocSet
         // cardinality? or if we don't know cardinality but DocSet size is very small
         return new FacetFieldProcessorByHashDV(fcontext, this, sf);
-      } else if (ntype == null) {
+      } else if (isNumber == false) {
         // single valued string...
         return new FacetFieldProcessorByArrayDV(fcontext, this, sf);
       } else {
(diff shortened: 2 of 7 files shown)


thomaswoeckinger:
@risdenk as your are one of the last commiters in this area, you may have a look at this PR

thomaswoeckinger:
@dsmiley @gerlowskija May you know who is responsible for this part of faceting code, to get a reviewer

HoustonPutman (commented):
It's unfortunate we can't check the docValues format, but can we at least add a comment explaining this. It's a bit confusing just looking at it without context.

risdenk (approved):

solr/core/src/test/org/apache/solr/search/facet/TestJsonFacets.java
```diff
@@ -4907,6 +4907,38 @@ public void testQueryJoinBooksAndPages() throws Exception {
...
+  
+  @Test
+  public void testMultivalueEnumTypes() throws Exception {
+    final Client client = Client.localClient();
+    
+    final SolrParams p = params("rows", "0");
+    
+    client.deleteByQuery("*:*", null);
+    
+    List<SolrInputDocument> docsToAdd = new ArrayList<>(6);
```
dsmiley, line 4919: Can you add them to an UpdateRequest so they are batched instead of sending them individually?  Same LOC in the end but avoids an anti-pattern.  Or is the separation pertinent to what's being tested?

thomaswoeckinger:
> It's unfortunate we can't check the docValues format, but can we at least add a comment explaining this. It's a bit confusing just looking at it without context.

FacetFieldProcessorByArrayDV does not handle DocValuesType.SORTED_NUMERIC because in method findStartAndEndOrds FieldUtil.getSortedSetDocValues is used, which can not handle this kind of DocValuesType.

In detail FacetFieldProcessorByArrayDV  extends FacetFieldProcessorByArray which seems to be designed for DocValuesType.SORTED_SET, because the method lookupOrd is only possible when having such a type.

Saying that, you may have some suggestions for a code comment.

For me it seems this was not working form the beginning of the 9x branch, there was simply not test for it.

solr/core/src/test/org/apache/solr/search/facet/TestJsonFacets.java
thomaswoeckinger: I tried but this will make the test more complicated as SolrTestCaseHS.Client does not keep a SolrClient and uses null therefore in internal API calls, may you have an hint for that, but the whole test class is not using an UpdateRequest

dsmiley (commented):
Can you add a CHANGES.txt entry to bug section of 9.9?  Basically say fixed a regression of faceting on a multi-valued EnumFieldType since an earlier 9.x.

solr/CHANGES.txt
```diff
@@ -187,6 +187,8 @@ Bug Fixes
 
 * SOLR-17652: Fix a bug that could cause long leader elections to leave PULL replicas in DOWN state forever. (hossman)
 
+* SOLR-17649: Fix a regression of faceting on multi-valued EnumFieldType, introduced by an earlier 9.x version. (Thomas Wöckinger, David Smiley, Houston Putman, Kevin Risden)
```
dsmiley, line 190: LOL if I merge this I'll just list yourself; I don't want any hint of credit.  As a reviewer my input was too trivial in this case to justify my name being listed.

dsmiley (approved):
I'll leave open for further possible feedback till tomorrow but I plan to merge tomorrow.  I plan to list only Thomas's name in CHANGES.txt.

dsmiley (commented):
I played with the PR a bit in the debugger to better understand.  Then I added extra checks so that we fail if we ask for a type of docValues that will not actually be there.  This is better than silently showing nothing!

dsmiley:
Looking for additional code review since I contributed to this PR substantially.

solr/core/src/java/org/apache/solr/search/facet/FacetField.java
```diff
@@ -139,11 +139,11 @@ public FacetProcessor<FacetField> createFacetProcessor(FacetContext fcontext) {
     // FieldType.getDocValuesType()
 
     if (!multiToken) {
-      if (mincount > 0 && prefix == null && (ntype != null || method == FacetMethod.DVHASH)) {
+      if (mincount > 0 && prefix == null && (isNumber || method == FacetMethod.DVHASH)) {
         // TODO can we auto-pick for strings when term cardinality is much greater than DocSet
         // cardinality? or if we don't know cardinality but DocSet size is very small
         return new FacetFieldProcessorByHashDV(fcontext, this, sf);
-      } else if (ntype == null) {
+      } else if (isNumber == false) {
```
mkhludnev, line 146: !isNumber

solr/core/src/java/org/apache/solr/search/facet/FieldUtil.java
```diff
@@ -44,27 +48,49 @@ public static SortedDocValues getSortedDocValues(SolrIndexSearcher searcher, Str
...
-    SolrIndexSearcher searcher = context.searcher();
-    NumericDocValues si = searcher.getSlowAtomicReader().getNumericDocValues(field.getName());
-    return si == null ? DocValues.emptyNumeric() : si;
+    var reader = context.searcher().getSlowAtomicReader();
+    var dv = reader.getNumericDocValues(field.getName());
+    checkDvType(dv, field, reader);
+    return dv == null ? DocValues.emptyNumeric() : dv;
+  }
+
+  private static void checkDvType(Object dv, SchemaField field, LeafReader reader) {
```
mkhludnev, line 73: It's not clear why to have this 20 lines. If we remove it, I think it just fails at class cast somewhere later. Why ISE is better than ClassCastException later? It's  just not obvious.

solr/core/src/java/org/apache/solr/search/facet/FacetField.java
dsmiley: Maybe you are not familiar that Lucene & Solr has recognized this style as acceptable due to its clarity

solr/core/src/java/org/apache/solr/search/facet/FieldUtil.java
dsmiley: There's wasn't a class cast; there was empty/no data which ultimately leads to no facet results.  That's *worse* than throwing an exception (of whatever type; no strong opinion from me).  This is my primary addition to the PR; don't *just* fix the wrong algorithm choice, but ensure we throw an exception in cases that would previously silently produce nothing.

epugh (approved):
I think the code is easier to read.

PR #3159, opened 2025-02-05
Deprecations

dsmiley:
Misc deprecations to things that already have substitutes or none if we feel none is needed.
The actual removal (sometimes requiring migrations of callers) can happen latter.

I intend to leave this open for a while.  Any committer can contribute to this branch thus PR.  Of course challenge me on anything.
I don't think deprecating random methods is JIRA/CHANGES.txt worthy.  Someone will find out in their IDE eventually :-). What we can do to help that user is to ensure there's a comment to help them know what to do differently.


epugh:
I ❤️ deprecations.  How can we have a strategy to make sure these removals happen?   I.e, how can we help folks see these areas of work and then move it forward?  

ARe you thinking that once this is back ported to 9, we can just go ahead in `main` and start removing the deprecated code?

In my head, any deprecated code existing in 10 is a shame.  We should rip those band aids off, but maybe I need a more nuanced view of what deprecation means.   Every time I see in a test a deprecated method (especially when it's not clear exactly how to fix it) it makes me sad ;-).

dsmiley:
> we can just go ahead in main and start removing the deprecated code?

Sure.  Some have JIRAs, even.  There's plenty of other deprecations outside this PR.

> any deprecated code existing in 10 is a shame.

I've learned to be realistic in a large code base.  The positive spin I'm at peace with is that spending just a little bit of deprecation time in advance gives us permission later to have the fun of deleting stuff when it suits us.

dsmiley:
Looking to merge this maybe tomorrow; certainly won't be the last of deprecations.  I'll end up creating another PR.
Some changes here are not just deprecations but using a new thing, but should be 9x compatible anyway.

PR #3161, opened 2025-02-06
Make PGP signing the release announcement optional

janhoy:
As discussed in slack https://the-asf.slack.com/archives/CEKUCUNE9/p1738760259166169

I opted to keep the instructions in the wizard but clearly mark it as optional, skipping the S/MIME email client stuff and focusing only on gpg command line clearsign option.

Not ran the wizard after this change so may be formatting bugs...

### DIFF
--- a/dev-tools/scripts/releaseWizard.yaml
+++ b/dev-tools/scripts/releaseWizard.yaml
@@ -116,10 +116,16 @@ templates:
 
     (( template=announce_solr_mail_body ))
     ----
-  announce_solr_sign_mail: |
+  announce_solr_announce_mail: |    
     The template below can be used to announce the Solr release to the
-    `announce@apache.org` mailing list. The mail *should be signed with PGP.*
-    and sent *from your `@apache.org` account*.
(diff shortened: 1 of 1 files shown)


epugh (approved):
I'd rather just rip out the section, jsut because it's one more thing and no one uses it.   But that's just me.  The rest looks good.   Though I have never been an RM ;-)

janhoy:
> I'd rather just rip out the section, jsut because it's one more thing and no one uses it. But that's just me. The rest looks good. Though I have never been an RM ;-)

Would it be better to play it down further by shortening the text, skipping the step by step instructions, and instead just link to gpg, e.g.?

> ...and it *can optionally* be signed with your PGP key, using the `--clearsign` option of gpg [as explained here](https://www.gnupg.org/gph/en/manual/x135.html).

janhoy:
Here is a proposal for how it will look without the detailed instructions, further downplaying PGP:
Skjermbilde 2025-02-06 kl  16 14 38
Skjermbilde 2025-02-06 kl  16 14 54

epugh:
Great!

PR #3163, opened 2025-02-06
SOLR-17635: unmarshalling Map to SimpleOrderedMap if key i of type St…

renatoh:
https://issues.apache.org/jira/browse/SOLR-17635

# Description

# Solution

# Tests

# Checklist


solr/solrj/src/java/org/apache/solr/common/util/JavaBinCodec.java
```diff
@@ -863,6 +868,26 @@ protected Map<Object, Object> readMap(DataInputInputStream dis, int sz) throws I
...
+        } else {
+          m = newMap(sz);
+        }
+      }
+      Object val = readVal(dis);
+      m.put(key, val);
+    }
+    return m;
+  }
+
```
renatoh, line 890: @dsmiley  FYI, this is marked as drafted.
We cannot simply create a SimpleOrderedMap since the key of the Map can also me something other than a String e.g. an Integer. The only way I can think of to overcome that is to read the first key and then do an instanceof check on it. Pretty ugly considering I have to use a raw Map.

solr/solrj/src/java/org/apache/solr/common/util/JavaBinCodec.java
```diff
@@ -848,9 +848,14 @@ protected Map<Object, Object> newMap(int size) {
     return size < 0 ? new LinkedHashMap<>() : CollectionUtil.newLinkedHashMap(size);
   }
 
-  public Map<Object, Object> readMap(DataInputInputStream dis) throws IOException {
+  public Map<?, Object> readMap(DataInputInputStream dis) throws IOException {
     int sz = readVInt(dis);
-    return readMap(dis, sz);
+
+    if (EnvUtils.getPropertyAsBool("solr.solrj.javabin.mapAsNamedList", true)) {
```
dsmiley, line 854: Reading this setting on _every single map decode_ is quite bad for performance.  This setting should be a field on the codec, and that which could be toggled with a setter (if useful in a test).

solr/solrj/src/java/org/apache/solr/common/util/JavaBinCodec.java
dsmiley: Oh man I didn't think of that!  Let's say we treat keys as a String no matter what.  I'm suspicious that our response formats would ever have a map whose key isn't a string.  Maybe add an assert if the key isn't already a String so that we can more readily identify problems.  (not sure if may read CharSequence but we can consider that as a String).

renatoh: Just casting the key to a String, or having an assert on it, will fail three unit test of TestJavaBinCodec. E.g. the .bin used in the tests - org.apache.solr.common.util.TestJavaBinCodec#SOLRJ_JAVABIN_BACKCOMPAT_BIN - holds a Map with Integer as key. So you are saying that this is not representing a realistic case and had to be changed?

dsmiley: Yes; consider this a hypothesis to be tested (by the results of the totality of Solr's test suite).

solr/solrj/src/test/org/apache/solr/common/util/TestJavaBinCodec.java
```diff
@@ -293,7 +306,7 @@ public void testForwardCompat() throws IOException {
         for (int i = 1;
             i < currentFormatBytes.length;
             i++) { // ignore the first byte. It is version information
-          assertEquals(newFormatBytes[i], currentFormatBytes[i]);
+          assertEquals("for i:" + i, newFormatBytes[i], currentFormatBytes[i]);
```
renatoh, line 309: After rewriting the .bin file with the changed output from org.apache.solr.common.util.TestJavaBinCodec#generateAllDataTypes tests like org.apache.solr.common.util.TestJavaBinCodec#testBackCompat are passing now. But this one testForwardCompat which compares all the bytes fails since the bytes are not the same anymore.
Is it possible I did something wrong when recreated the bin-file?
Is just took the output from generateAllDataTypes(), called  javabin.marshall(matchObj, output) with it and wrote the result via a FileOutputStream to a file. 
So the object in the unmarshalle version are the same, but the bytes are not

dsmiley: TestJavaBinCodec should test without this setting.  This is another reason to make this a setter or something, which is clearer than a system property.

renatoh: All the tests in TestJavaBinCodec do not need to pass with javabin.setMapAsNamedList=true even though it will be true by default? I would have thought all these test need to pass in order to ensure we do not break anything by changing to SOM.
In this case, there is also no need to update /solrj/javabin_backcompat.bin?

solr/solrj/src/java/org/apache/solr/common/util/JavaBinCodec.java
```diff
@@ -863,6 +871,21 @@ protected Map<Object, Object> readMap(DataInputInputStream dis, int sz) throws I
...
+  protected Map<?, Object> readMapAsSimpleOrderedMapForStringKeys(DataInputInputStream dis, int sz)
+      throws IOException {
+
+    SimpleOrderedMap<Object> entries = new SimpleOrderedMap<>(sz);
+
+    for (int i = 0; i < sz; i++) {
+      Object key = readVal(dis);
+      assert key instanceof String;
+      Object val = readVal(dis);
+      entries.put((String) key, val);
```
dsmiley, line 884:
```suggestion
      entries.add((String) key, val); // using NL.add() since key won't repeat
```

solr/solrj/src/java/org/apache/solr/common/util/JavaBinCodec.java
```diff
@@ -1436,4 +1459,8 @@ public void close() throws IOException {
       daos.flushBuffer();
     }
   }
+
```
dsmiley, line 1462: TODO javadocs.

solr/solrj/src/test/org/apache/solr/common/util/TestJavaBinCodec.java
```diff
@@ -409,12 +419,17 @@ private static byte[] getBytes(Object o, boolean readAsCharSeq) throws IOExcepti
     }
   }
 
-  private static Object getObject(byte[] bytes) throws IOException {
+  private static Object getObject(byte[] bytes, boolean mapAsNamedList) throws IOException {
```
dsmiley, line 422: don't need these getObject anymore, I think

dsmiley (commented):
This looks basically ready!  

Suggested 9.9 CHANGES.txt in "Other"

> SOLR-17635: SolrJ "javabin" format now reads all maps as SimpleOrderedMap (a NamedList, implements Map) instead of LinkedHashMap for better compatibility as Solr changes some NamedLists to Map/SimpleOrderedMap.

This is also worthy of upgrade notes, which would go in a 9.9 section on `major_changes_in_solr_9.adoc` (maybe named differently) where you can point out the system property that will toggle this behavior.

solr/solrj/src/java/org/apache/solr/common/util/JavaBinCodec.java
dsmiley: That was an `O(N^2)` performance bug, by the way 😱

solr/solrj/src/test/org/apache/solr/common/util/TestJavaBinCodec.java
renatoh: @dsmiley 
We have now 9 failing test, here my findings:
TestFastJavabinDecoder
  testSimple-> the assert on line 136 fails due to the JSON with a SOM in it has some extra line breaks

TestUpdateRequestCodec
  testBackCompat4_5() -> failing because /solrj/updateReq_4_5.bin" has a Map with SolrInputDocument as Key
  testStreamableInputDocFormat -> because in JavaBinUpdateRequestCodec.StreamingCodec#readOuterMostDocIterator line 315 instanceOf NamedList is now true, and it is checked before instanceof Map

ClusterStateProviderTest
  6 tests are failing e.g. testClusterStateProvider -> ClusterState is using org.apache.solr.common.util.Utils to deserialize the Json, which creates a LinkedHashMap and clusterStateHttp has now a SOM in it

dsmiley: TestUpdateRequestCodec: dunno what to make of that without debugging; hopefully you can make recommendations.  Not surprised to see this _kind of issue_ where some code is doing instanceof checks in a certain order, written at a time when no NamedList was a Map but now a prominent (and even most popular, I'd say) is actually one.  You might want to search the codebase for instanceof checks against NamedList to see if there are other areas to improve.  I have some WIP for Utils.java
ClusterStateProviderTest: looks like just needs to convert both maps to the same type (e.g. wrap in HashMap) ?

solr/solrj/src/java/org/apache/solr/common/util/JavaBinCodec.java
renatoh: oh, I am glad at least one of us knows what we are doing here;)

dsmiley: It underscores my concern about making SimpleOrderedMap mutable.  I still question it.

solr/solrj/src/test/org/apache/solr/common/util/TestJavaBinCodec.java
renatoh: > ClusterStateProviderTest: looks like just needs to convert both maps to the same type (e.g. wrap in HashMap
Where would I do that? There is really a lot happening in those tests and I still have not managed to fully wrap my head around it.

solr/solrj/src/java/org/apache/solr/common/util/Utils.java
```diff
@@ -290,7 +290,7 @@ public static Object fromJSON(byte[] utf8) {
   }
 
   public static Object fromJSON(byte[] utf8, int offset, int length) {
-    return fromJSON(utf8, offset, length, STANDARDOBJBUILDER);
+    return fromJSON(utf8, offset, length, SIMPLEORDEREDMAPOBJBUILDER);
```
renatoh, line 293: this breaks another 17 tests.
@dsmiley Do you think it is OK to run the test ClusterStateProviderTest with solr.solrj.javabin.mapAsNamedList=false?

solr/solrj/src/java/org/apache/solr/common/util/TextWriter.java
```diff
@@ -84,11 +84,11 @@ default void writeVal(String name, Object val, boolean raw) throws IOException {
       }
     } else if (val instanceof IteratorWriter) {
       writeIterator(name, (IteratorWriter) val, raw);
-    } else if (val instanceof MapWriter) {
+    } else if (val instanceof MapWriter && !(val instanceof SimpleOrderedMap)) {
       writeMap(name, (MapWriter) val);
     } else if (val instanceof ReflectWritable) {
       writeVal(name, Utils.getReflectWriter(val));
-    } else if (val instanceof MapSerializable) {
+    } else if (val instanceof MapSerializable && !(val instanceof SimpleOrderedMap)) {
```
dsmiley, line 91: why this line?  SOM isn't MapSerializable

solr/solrj/src/java/org/apache/solr/common/util/TextWriter.java
```diff
@@ -84,11 +84,11 @@ default void writeVal(String name, Object val, boolean raw) throws IOException {
...
       // todo find a better way to reuse the map more efficiently
       writeMap(name, ((MapSerializable) val).toMap(new LinkedHashMap<>()), false, true);
     } else if (val instanceof Map) {
```
dsmiley, line 94: this should simply be ranked higher than MapWriter, where it will then apply to SimpleOrderedMap.

solr/solrj/src/java/org/apache/solr/common/util/Utils.java
dsmiley: No; we should understand why the test or underlying functionality are breaking and then make decisions based on that.

dsmiley (commented):
LMK if you need my assistance to look at some key failures and make heads or tails out of what to do.

solr/solrj/src/java/org/apache/solr/common/util/Utils.java
renatoh: Looking at a test case ClusterStateProviderTest.testClusterStateProviderEmptySolrVersion, Http2ClusterStateProvider is using JavaBinCode to deserialize, hence we get now a SOM, ZkClientClusterStateProvider is using JSON and use the Utils.fromJson to deserialize, hence it still gets LinkedHashMap. In this test it expects the same data structure from both Http2ClusterStateProvider and ZkClientClusterStateProvider, hence it's failing now.
 If I change the Utils to return SMO instead of a LinkedHashMap, as I did in this change, we break other tests.

solr/solrj/src/java/org/apache/solr/common/util/TextWriter.java
renatoh: I've tried rearranging the order before adding the instanceof-checks, but much code, or at least many tests seem to depend on this order.
If I just move up the 'instanceof Map' check as you suggest, I have 42 new tests failing.

solr/solrj/src/java/org/apache/solr/common/util/Utils.java
renatoh: @dsmiley How do we continue on this issue? Based on what I've described above,  it seems fine to me to run these tests with solr.solrj.javabin.mapAsNamedList=false. What do you think?

renatoh:
> FYI I'm _very_ eager to examine this in detail but am on vacation this week.
No worries, enjoy your vacation, I will off  next week and limited to my phone

solr/solrj/src/java/org/apache/solr/common/util/Utils.java
dsmiley: This must be changed back; I don't think it makes sense to deserialize JSON and get a SimpleOrderedMap in there.  I debugged and the root of the problem is that SimpleOrderedMap is not actually fully implementing the contract of a Map with regards to equals & hashCode.  I'll push some changes very soon.  This part should be extracted to it's own PR against the same issue of SimpleOrderedMap implementing a Map.

renatoh: I see, now the test ClusterStateProviderTest#testClusterStateProviderEmptySolrVersion is passing. So the issue was not that one of the DocCollection had a SOM and the other had a LinkedHashMap, the issue was that they were not equal despite of having the same entries. Thanks makes sense:)

dsmiley:
Let's get #3214 merged first, thus narrowing this PR to just what it needs to do.

renatoh:
@dsmiley All tests are passing now, are we able to merge?

dsmiley (commented):
Needs a CHANGES.txt entry.  major-changes-in-solr-9.adoc should explain this in a 9.9 section, that it applies to SolrJ (not just Solr servers), and that there's a performance risk, and that it can be disabled.

I have a some doubts that gnaw at me pertaining to the performance risks of O(1) maps suddenly becoming O(N).  It doesn't sit well with me.  You had suggested changing SimpleOrderedMap to be LinkedHashMap based.  That's definitely safer, and whatever extra overhead is involved there should be fairly minor.  Maybe we should pause merging this until pursuing that.  Pursuing that probably means deprecating/removing indexed access to NamedList, requiring instead callers use iterators.

renatoh:
> I have a some doubts that gnaw at me pertaining to the performance risks of O(1) maps suddenly becoming O(N). It doesn't sit well with me. You had suggested changing SimpleOrderedMap to be LinkedHashMap based. That's definitely safer, and whatever extra overhead is involved there should be fairly minor. Maybe we should pause merging this until pursuing that. Pursuing that probably means deprecating/removing indexed access to NamedList, requiring instead callers use iterators.
But if SMO is backed by a LinkedHashMap, wouldn't then these changes become obsolete? What is the point of unmarshalling to SMO (backed by a LinkedHashMap) instead of a directly to a LinkedHashMap?

dsmiley:
> But if SMO is backed by a LinkedHashMap, wouldn't then these changes become obsolete? What is the point of unmarshalling to SMO (backed by a LinkedHashMap) instead of a directly to a LinkedHashMap?

Only an SMO is also a NamedList.  A LinkedHashMap itself is not.

I was thinking about the need for this a bit more.  The primary concern is a SolrJ 9 reading data sent from Solr 10, which will see more expanded use of SimpleOrderedMap that was previously a NamedList.  Other circumstances can leave this disabled by default.  In Solr 9 specifically, we could change BinaryResponseParser to detect the version of the writer in some way TBD and then act accordingly.  A separate issue/PR.

dsmiley:
Map.get, Map.remove, Map.put -- basic operations are O(1) in a hash based Map, but not in a dense array List that SOM is today.

renatoh:
> Map.get, Map.remove, Map.put -- basic operations are O(1) in a hash based Map, but not in a dense array List that SOM is today.

I do understand that part, but today, we are umarshalling to a LinkedHashMap, with the current changes we would unmarshall to a SMO (backed by an ArrayList), I do understand the potential performance issues of that change.
But if we now changed the SMO to a LinkedHashMap, we would then unmarshall to an LinkedHashMap in a SMO, compared to unmarshall to a LinkedHashmap, as we do today. So what is the point of changing from a LinkedHashMap to LinkedHashMap in a SMO?

dsmiley:
> So what is the point of changing from a LinkedHashMap to LinkedHashMap in a SMO?

A LinkedHashMap doesn't implement NamedList.  An SMO does.  Thus a Solr 10 server could start passing maps and MapWriter things (e.g. SolrParams) to the response in situations that a SolrJ 9 is still anticipating reading a NamedList.  With SOM; it's both things in one. The choice of LinkedHashMap vs ArrayList would be a regression for cases where the client code already expected a Map; was assuming O(1) performance of the map if lots of stuff is in it.  Because NamedList/Map is nestable, at the time of unmarshalling we have no idea what the actual code looking at the object is casting it as.  In a fantasy world if there was just exactly one thing to unmarshall, then there would be maybe readMap vs readNamedList and we wouldn't have concerns that this PR tries to solve.

That said, let's not actually change SMO to be based on a LinkedHashMap; at least not soon.  I proposed only reading maps as SMO for a more narrow circumstance (9.x talking to 10.x) that would be somewhat temporary / limited, thus ameliorating a performance risk.

dsmiley:
I merged with main and tweaked the naming a little as I asked.  But most notably, made the setting `false` by default, thus is an opt-in compatibility mechanism that can be enabled as we see a need for it.  Given the performance risk, I felt this approach is safer.

I think this is ready to merge.

renatoh:
@dsmiley Thanks

PR #3166, opened 2025-02-07
Refactor UpdateLog.LogPtr to a record

dsmiley:
And remove previousPointer (unused)


solr/core/src/java/org/apache/solr/update/UpdateLog.java
```diff
@@ -268,40 +268,13 @@ public String toString() {
   protected Meter copyOverOldUpdatesMeter;
   protected SolrMetricsContext solrMetricsContext;
 
-  public static class LogPtr {
-    final long pointer;
-    final long version;
-    // used for entries that are in-place updates and need a pointer to a previous update command
-    final long previousPointer;
```
dsmiley, line 275: This was in here since when it first showed up like 10 years ago, and even then wasn't actually *used*.  Populated but not referenced.  CC @chatman

dsmiley:
Ugh; ECJ doesn't know about Java records.  Probably need to update that.

PR #254, opened 2021-08-10
Solr 15581

markrmiller:
Includes an EnglishWOrdGenerator as an initial proof of concept.

First PR is incomplete, a variety of things to smooth out and straighten up and the EnglishWordGenerator is only single token ATM.

PR #3171, opened 2025-02-09
Fix DockMakerTest maxCardinality

dsmiley:
See this [thread](https://github.com/apache/solr/pull/254#discussion_r1917413526)
> I see a repeating failure running DockMakerTest.testBasicCardinalityAlpha with seed E0BBCB76FA600308 in which every field value is the letter 'k'. This DSL framework is extremely different from any other code most people encounter so it took some time to narrow down the problem to where I'm commenting. This code will choose consecutive seeds for subsequent processing. (Consecutiveness is not a problem I think). For a cardinality of 2 and based on the random test seed, this long seed will be either 635258717 or 635258718. But the next stage will always produce 'k' for either of those seeds; the next stage produces a random string of length from 1 to 6.

I think the fundamental fix is in the test; to recognize this is a *max*Cardinality, and weaken its assertions accordingly.

I added some comments in StringsDSL.

### DIFF
--- a/solr/benchmark/src/test/org/apache/solr/bench/DockMakerTest.java
+++ b/solr/benchmark/src/test/org/apache/solr/bench/DockMakerTest.java
@@ -17,13 +17,11 @@
 package org.apache.solr.bench;
 
 import static org.apache.solr.bench.Docs.docs;
-import static org.apache.solr.bench.generators.SourceDSL.booleans;
 import static org.apache.solr.bench.generators.SourceDSL.integers;
 import static org.apache.solr.bench.generators.SourceDSL.strings;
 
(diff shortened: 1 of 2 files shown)


solr/benchmark/src/test/org/apache/solr/bench/DockMakerTest.java
```diff
@@ -200,25 +201,4 @@ public void testWordListZipfian() {
 
     assertNotNull(field.getValue().toString());
   }
-
-  @Test
-  public void testGenDoc() {
```
dsmiley, line 205: removed because it had no assertions

dsmiley:
Will merge tomorrow.

PR #3172, opened 2025-02-10
SOLR-17662 SPI: TransformerFactory/DocTransformer

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17662

Despite me wanting this for a long time, what motivated me today specifically was looking at @malliaridis trying to solve risks of static class initialization for some of these factories in #2881.  The use of SPIs should make that concern go away for some things like static plugin registries.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -1129,7 +1129,7 @@ public class SolrCore implements SolrInfoBean, Closeable {
       initWriters();
       qParserPlugins.init(QParserPlugin.standardPlugins, this);
       valueSourceParsers.init(ValueSourceParser.standardValueSourceParsers, this);
-      transformerFactories.init(TransformerFactory.defaultFactories, this);
+      transformerFactories.init(TransformerFactory.builtIns(coreContainer.getConfig()), this);
       loadSearchComponents();
       updateProcessors.init(Collections.emptyMap(), this);
 
--- a/solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformerFactory.java
+++ b/solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformerFactory.java
@@ -18,6 +18,7 @@ package org.apache.solr.response.transform;
 
 import static org.apache.solr.schema.IndexSchema.NEST_PATH_FIELD_NAME;
 
+import jakarta.inject.Named;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import org.apache.lucene.search.BooleanClause;
@@ -56,6 +57,7 @@ import org.apache.solr.search.SyntaxError;
  *
  * @since solr 4.9
  */
+@Named("child")
 public class ChildDocTransformerFactory extends TransformerFactory {
 
   static final char PATH_SEP_CHAR = '/';
--- a/solr/core/src/java/org/apache/solr/response/transform/CoreAugmenterFactory.java
+++ b/solr/core/src/java/org/apache/solr/response/transform/CoreAugmenterFactory.java
@@ -17,9 +17,11 @@
 
 package org.apache.solr.response.transform;
 
+import jakarta.inject.Named;
 import org.apache.solr.common.params.SolrParams;
 import org.apache.solr.request.SolrQueryRequest;
 
+@Named("core")
 public class CoreAugmenterFactory extends TransformerFactory {
 
   @Override
--- a/solr/core/src/java/org/apache/solr/response/transform/TransformerFactory.java
+++ b/solr/core/src/java/org/apache/solr/response/transform/TransformerFactory.java
@@ -16,15 +16,17 @@
  */
 package org.apache.solr.response.transform;
 
-import java.util.HashMap;
+import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 import org.apache.solr.common.params.SolrParams;
 import org.apache.solr.common.util.NamedList;
+import org.apache.solr.core.NodeConfig;
 import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.util.SolrPluginUtils;
 import org.apache.solr.util.plugin.NamedListInitializedPlugin;
 
-/** New instance for each request */
+/** Factory for {@link DocTransformer}. */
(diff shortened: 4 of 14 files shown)


solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -1129,7 +1129,7 @@ protected SolrCore(
       initWriters();
       qParserPlugins.init(QParserPlugin.standardPlugins, this);
       valueSourceParsers.init(ValueSourceParser.standardValueSourceParsers, this);
-      transformerFactories.init(TransformerFactory.defaultFactories, this);
+      transformerFactories.init(TransformerFactory.builtIns(coreContainer.getConfig()), this);
```
dsmiley, line 1132: I like the name "builtIns" best so I chose it.  "implicits" would be my second choice.  I like a general name, thus not including "factories" as some places where we would use this pattern are not even "factories".  "default" isn't terrible but less good.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformerFactory.java
```diff
@@ -18,6 +18,7 @@
 
 import static org.apache.solr.schema.IndexSchema.NEST_PATH_FIELD_NAME;
 
+import jakarta.inject.Named;
```
dsmiley, line 21: To solve the problem of identifying the name for plugins, I found this annotation, which seems quite suitable.  I read further about it when looking at HK2 (a dependency of ours) lightweight dependency-injection framework that uses the annotations in that JAR/package.  This PR doesn't use HK2 BTW.

solr/core/src/java/org/apache/solr/response/transform/CoreAugmenterFactory.java
```diff
@@ -17,9 +17,11 @@
 
 package org.apache.solr.response.transform;
 
+import jakarta.inject.Named;
 import org.apache.solr.common.params.SolrParams;
 import org.apache.solr.request.SolrQueryRequest;
 
+@Named("core")
 public class CoreAugmenterFactory extends TransformerFactory {
```
dsmiley, line 25: `<RANT>` Why the hell didn't we name TransformerFactory as its rightful name, DocTransformerFactory, and name it's implementations as such as well?  I don't even need to know who/why; it's just wrong.  We have sinned and need to make amends someday.`</RANT>`

solr/core/src/java/org/apache/solr/response/transform/TransformerFactory.java
```diff
@@ -105,18 +107,17 @@ default boolean mayModifyValue() {
     }
   }
 
-  public static final Map<String, TransformerFactory> defaultFactories = new HashMap<>(9, 1.0f);
+  // loaded once, based on the node
+  private static Map<String, TransformerFactory> builtIns;
```
dsmiley, line 111: no static initializer anymore; no class loading deadlock risk.  Right @uschindler ?

solr/core/src/java/org/apache/solr/response/transform/TransformerFactory.java
dsmiley, line 1: There are some other TransformerFactory impls not listed here; 2 for QueryElevationComponent which weirdly self-register when QEC loads, and the other is LTR.  I could expand the scope of this PR but wanted to start simpler to get the conversation going.

solr/core/src/java/org/apache/solr/response/transform/TransformerFactory.java
```diff
@@ -105,18 +107,17 @@ default boolean mayModifyValue() {
...
-    defaultFactories.put("docid", new DocIdAugmenterFactory());
-    defaultFactories.put("shard", new ShardAugmenterFactory());
-    defaultFactories.put("child", new ChildDocTransformerFactory());
-    defaultFactories.put("subquery", new SubQueryAugmenterFactory());
-    defaultFactories.put("json", new RawValueTransformerFactory("json"));
-    defaultFactories.put("xml", new RawValueTransformerFactory("xml"));
-    defaultFactories.put("geo", new GeoTransformerFactory());
-    defaultFactories.put("core", new CoreAugmenterFactory());
+  /** (for internal use) */
+  public static synchronized Map<String, TransformerFactory> builtIns(NodeConfig nodeConfig) {
```
dsmiley, line 114: Since the lazy loaded list is initialized based on NodeConfig (basically CoreContainer), admittedly there's a test stability risk since NodeConfig ClassLoader classpath of the first test that runs will "win".  For the built-in plugins, let's just not do anything that would disturb that list.  If we wanted to test that someone could add a plugin in a JAR, that would be a high level integration test using Docker or BATS.  Not unit tests.

dsmiley: Anyway, the use of NodeConfig here thus adds a capability of config-less addition of a plugin (i.e. just add a JAR for a custom DocTransformer; no need to edit solrconfig.xml / configSet).

dsmiley (commented):
This is a draft for discussion of the approach & scope.
Use of `javax.inject.Named` is non-obvious but I think good.  Maybe should be `jakarta` namespace moving forward, actually.  We could do name guessing as well... but as you can see here due to terrible/inconsistent class names for TransformerFactory impls, it wouldn't work.  Even so; I like the explicitness in the annotation so lets not fight it.  It's a one-liner.

solr/core/src/java/org/apache/solr/response/transform/ChildDocTransformerFactory.java
dsmiley: Could also have added a getName to instances but it's more verbose and forces instantiation of the class.  They are instantiated here anyway but nonetheless I like the idea of a future optimization that doesn't instantiate until first use.  Could also decide a public static final String field `NAME` is the name.  Anyway, FWIW I prefer the annotation; feels more modern and is simpler.  Wouldn't be okay in SolrJ (e.g. for DocRouter) as we don't want to add dependencies there but that's fine.

dsmiley:
Adding another check, that the implementations do *not* implement `SolrCoreAware`, would also be good, as they are the builtIn ones instantiated once.  I imagine a non-issue for existing ones but I could imagine a user trying this feature out and it not working for that reason.  Of course eventually we want that to work but it's out of scope of the built-in registry.

PR #3173, opened 2025-02-10
SOLR-17663: Protect TimeOut from overflow.

bruno-roustant:
https://issues.apache.org/jira/browse/SOLR-17663

TimeOut constructor is passed a long interval parameter. If this long is large enough, the timeoutAt field overflows, making all the methods incorrect, including hasTimedOut() which returns true immediately.

TimeUnit.convert() deals with the same problem by checking the value to convert ahead of time and by setting fixed Long.MAX_VALUE to prevent overflow. We should do the same for timeoutAt in TimeOut constructor.


dsmiley (approved):
Thanks!

bruno-roustant:
Thanks for your review David.

PR #3175, opened 2025-02-11
Update enabling-ssl.adoc - fix minor typo in example

Kordishal:
# Description

The example to update the Zookeeper properties has a typo as it uses urlSchema instead of urlScheme.

# Solution

Rename property in examples.

# Tests

None.

# Checklist

### DIFF
--- a/solr/solr-ref-guide/modules/deployment-guide/pages/enabling-ssl.adoc
+++ b/solr/solr-ref-guide/modules/deployment-guide/pages/enabling-ssl.adoc
@@ -236,7 +236,7 @@ The examples below use the `bin/solr` tool that comes with Solr to do this.
 ====
 [source,terminal]
 ----
-$ bin/solr cluster --property urlSchema --value https --zk-host server1:2181,server2:2181,server3:2181
+$ bin/solr cluster --property urlScheme --value https --zk-host server1:2181,server2:2181,server3:2181
 ----
 ====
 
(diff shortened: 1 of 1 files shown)


solr/solr-ref-guide/modules/deployment-guide/pages/enabling-ssl.adoc
```diff
@@ -245,7 +245,7 @@ Windows::
 ====
 [source,powershell]
 ----
-C:\> bin/solr.cmd --property urlSchema --value https --zk-host server1:2181,server2:2181,server3:2181
```
epugh, line 248: Does this need a `bin/solr cluster` here?

epugh:
Okay, one thing to confirm, and then this looks good.   One shame is that this page https://solr.apache.org/guide/solr/latest/deployment-guide/solr-control-script-reference.html never mentions the "cluster" tool ;-(

epugh:
And after reviewing, I opened https://issues.apache.org/jira/browse/SOLR-17668...

solr/solr-ref-guide/modules/deployment-guide/pages/enabling-ssl.adoc
Kordishal: I did not notice that. As I don't have a windows system I can't really confirm this. The script is running the cluster command with the solrcli class. So probably would be documented there.

https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/cli/SolrCLI.java

and then uses:

https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/cli/ClusterTool.java

SolrCli has a cluster command so probably this needs cluster too, but I have found no documentation of SolrCLI at all.

epugh:
I searched our ref guide for "solrcli" and yeah, the main page, the "Solr Control Script" page doesn't come up!:

image


We should add a "aka SolrCLI" on that page so it comes up!

epugh:
I just triggered the builds, and then this looks ready to merge.

epugh:
thank you for the fix!

PR #3176, opened 2025-02-11
SOLR-17667: Simplify zombie logic in LBSolrClient

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17667

I made a patch for branch_9x first, since that's where the failures are occurring for TestLBHttp2SolrClient

The changes are:

- No alive map, just allServers map for the servers given to the client (not passed via a request)
- The aliveness is determined by if the zombieMap contains the server
- All modifications to allServers and zombieServers go through the same methods. These methods are synchronized, so the list and both maps will not be updated independently.
- Since all modifications go through the same methods, we don't need to worry about dropping servers as the previous logic did.

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttpSolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttpSolrClient.java
@@ -192,7 +192,7 @@ public class LBHttpSolrClient extends LBSolrClient {
    */
   @Deprecated
   @Override
-  public String removeSolrServer(String server) {
+  public synchronized String removeSolrServer(String server) {
     urlToClient.remove(server);
     return super.removeSolrServer(server);
   }
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBSolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBSolrClient.java
@@ -662,7 +662,7 @@ public abstract class LBSolrClient extends SolrClient {
 
   private void checkAZombieServer(ServerWrapper zombieServer) {
     try {
-      log.debug("Checking zombie server {}", zombieServer.getBaseUrl());
+      log.debug("Checking zombie server {}", zombieServer);
       QueryRequest queryRequest = new QueryRequest(solrQuery);
       final var responseRaw = doRequest(zombieServer.getBaseUrl(), queryRequest, null);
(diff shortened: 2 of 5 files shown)


solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBHttpSolrClient.java
```diff
@@ -192,7 +192,7 @@ protected SolrClient getClient(Endpoint endpoint) {
    */
   @Deprecated
   @Override
-  public String removeSolrServer(String server) {
+  public synchronized String removeSolrServer(String server) {
```
gerlowskija, line 195: [Q] The javadocs point to `removeSolrEndpoint` as the successor to this deprecated method: should that be `synchronized` as well?

(It probably doesn't matter terribly for branch_9x since removeSolrEndpoint currently calls removeSolrServer; mostly mentioning to make sure we remember to do it on the 'main' version of this PR)

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBSolrClient.java
```diff
@@ -612,6 +603,7 @@ private void startAliveCheckExecutor() {
     if (aliveCheckExecutor == null) {
       synchronized (this) {
         if (aliveCheckExecutor == null) {
+          log.debug("Starting aliveCheckExecutor");
```
gerlowskija, line 606: [0] Not sure if you intend to keep these debug lines, but if so, might be worth adding something to the message to indicate which client it belongs to, since apps might have multiple.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBSolrClient.java
```diff
@@ -670,60 +662,64 @@ public RequestWriter getRequestWriter() {
...
-        ServerWrapper wrapper = zombieServers.remove(zombieServer.getBaseUrl());
-        if (wrapper != null) {
-          wrapper.failedPings = 0;
-          if (wrapper.standard) {
-            addToAlive(wrapper);
-          }
-        } else {
-          // something else already moved the server from zombie to alive
-        }
+        reviveZombieServer(zombieServer);
```
gerlowskija, line 672: [+1] Very nice little cleanup 🎉

gerlowskija (commented):
Left a few comments but nothing important; LGTM

PR #3177, opened 2025-02-11
Typo in enabling-ssl.adoc

temalcode:
Correcting keystore and key trust store file paths in the examples.

# Description

Correcting SSL keystore and SSL trust store file paths in the examples.

# Solution

Added root directory to Linux and C:\ to windows configuration.

# Tests

# Checklist

### DIFF
--- a/solr/solr-ref-guide/modules/deployment-guide/pages/enabling-ssl.adoc
+++ b/solr/solr-ref-guide/modules/deployment-guide/pages/enabling-ssl.adoc
@@ -83,9 +83,9 @@ NOTE: If you setup Solr as a service on Linux using the steps outlined in xref:t
 SOLR_SSL_ENABLED=true
 # Uncomment to set SSL-related system properties
 # Be sure to update the paths to the correct keystore for your environment
-SOLR_SSL_KEY_STORE=etc/solr-ssl.keystore.p12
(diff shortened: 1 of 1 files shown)


epugh (approved):
LGTM.

janhoy:
Sorry, but I cannot see that this is a bug or why this change would be any better than the default in the docs. The docs contain the same values as commented in our `solr.in.sh/cmd` file, and the paths are resolved relative to solr root dir. I.e. if you place your keystore files in `solr/server/etc/` then this works. Of course, if you place your keystore elsewhere you need to modify those values, but I cannot see why `/etc/` would be a better default?

temalcode:
In Linux:
The current docs path set to `etc/` (without the root directory) not `/etc/`. This will give you an error (running using solr.sh) if you copy paste this configuration.

In Windows:
There is no root directory and a `etc` folder in Windows. That's why I changed it to `C:\`

epugh:
I had interpreted it the same as @temalcode  the you would want a `/` to have the `/etc` dir?  And the c:\ would be a good practive for the windows version?

janhoy:
The doc is just an example. Perhaps what’s missing is to say that the path is relative to solr install dir and can/should be modified to where user wants to place the certs. Since these certs don’t exist already, one location is not more «correct» than another.

PR #3180, opened 2025-02-12
build solr/webapp/build.gradle: remove needless excludes

dsmiley:
These exclusions don't match anything.


epugh (approved):
I suppose if the build still works, then LGTM!

HoustonPutman:
Yeah, I guess the non-minimized files used to be in source control?

malliaridis (approved):
Yep, looked up the history and we had at some point a move to minimized files, but probably forgot to remove these exclusions.

PR #3181, opened 2025-02-13
SOLR-17670: Fix unnecessary memory allocation caused by a large reRankDocs param

gaojiabao1991:
https://issues.apache.org/jira/browse/SOLR-17670

# Description

For a query containing re-ranking, such as:
{
	"start": "0",
	"rows": 10,
	"fl": "ID,score",
	"q": "*:*",
	"rq": "{!rerank reRankQuery='{!func} 100' reRankDocs=1000000000 reRankWeight=2}"
}

The current execution logic is as follows:
1. Perform normal retrieval using the q parameter.
2. Re-score all documents retrieved in the q phase using the rq parameter.

During the retrieval in phase 1 (using q), a TopScoreDocCollector is created. Underneath, this creates a PriorityQueue which contains an Object[]. The length of this Object[] continuously increases with reRankDocs without any limit. 

On my local test cluster with limited JVM memory, this can even trigger an OOM, causing the Solr node to crash. I can also reproduce the OOM situation using the SolrCloudTestCase unit test. 

# Solution

I think limiting the length of the Object[] array using searcher.getIndexReader().maxDoc() at ReRankCollector would resolve this issue. This way, when reRankDocs exceeds maxDoc, memory allocation will not continue to increase indefinitely. 

# Tests

TestRerankOnSolrCloud will reproduce the OOM situation without my fix. 

# Checklist


solr/core/src/test/org/apache/solr/search/TestRerankOnSolrCloud.java
```diff
@@ -0,0 +1,102 @@
...
+            CommonParams.Q, "*:*",
+            CommonParams.FL, "id,test_s,test_i,score",
+            CommonParams.RQ,
+                "{!rerank reRankQuery='{!func} 100' reRankDocs=1000000000 reRankWeight=2}");
+
+    QueryRequest req = new QueryRequest(params);
+    QueryResponse rsp = req.process(cluster.getSolrClient(), COLLECTION_NAME);
+
+    // Test passes if execution completes without OOM
+    Assert.assertTrue(true);
```
risdenk, line 82: I don't like this test. We shouldn't be trying to cause an OOM. There has to be a better way to test this.

risdenk (changes requested):

solr/core/src/test/org/apache/solr/search/TestRerankOnSolrCloud.java
dsmiley: IMO no test is needed.

Furthermore this functionality has nothing to do with SolrCloud (or even distributed search?), so doesn't need the overhead of a SolrCloud test (or even Jetty for that matter) if we did want to test it.

gaojiabao1991: Agreed. My unit tests are non-standard and were only intended to help others confirm the issue. They should not be included in the PR as code to be merged.

I also agree with David's point @dsmiley that this modification is relatively straightforward, so once the issue is confirmed, we can simply decide on the logic and easily fix it. Trying to specifically test this small point might actually complicate things.

Therefore, I have simply removed the unit tests from the PR and added them to a separate branch:
https://github.com/apache/solr/compare/main...gaojiabao1991:solr:17670-extra-test
The tests in this branch are only for reviewers to easily confirm the issue and will not be merged.

risdenk (approved):

dsmiley (approved):

dsmiley:
I'll be merging this nice fix into 9.8.1

gaojiabao1991:
> I'll be merging this nice fix into 9.8.1

Thanks so much, David!

PR #3182, opened 2025-02-13
SOLR-17671 Add a hook in ReplicationHandler to allow extending CoreReplication.

bruno-roustant:
While it is possible to define the replication handler class in solr.xml, it is not possible to extend CoreReplication logic to tune how index files are copied, or filter/unwrap the Directory used to open the files inputs.

The proposal is to add a hook method "createCoreReplication" in ReplicationHandler, and another "filterDirectory" in ReplicationAPIBase.

Example:
In the solr-sandbox encryption module, we would need a way to unwrap the Directory used to copy files during index fetching. Otherwise the files are decrypted by the EncryptionDirectory seamlessly during the files copy, ending up with follower replicas having cleartext index.

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/admin/api/ReplicationAPIBase.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/api/ReplicationAPIBase.java
@@ -193,7 +193,8 @@ public abstract class ReplicationAPIBase extends JerseyResource {
         fileMetaData.name = infos.getSegmentsFileName();
         fileMetaData.size = filteredDir.fileLength(infos.getSegmentsFileName());
         if (infos.getId() != null) {
-          try (final IndexInput in = filteredDir.openInput(infos.getSegmentsFileName(), IOContext.READONCE)) {
+          try (final IndexInput in =
+              filteredDir.openInput(infos.getSegmentsFileName(), IOContext.READONCE)) {
             try {
               fileMetaData.checksum = CodecUtil.retrieveChecksum(in);
             } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/handler/admin/api/ReplicationAPIBase.java
```diff
@@ -377,7 +383,9 @@ public void write(OutputStream out) throws IOException {
       try {
         initWrite();
 
-        Directory dir = solrCore.withSearcher(searcher -> searcher.getIndexReader().directory());
+        Directory dir =
+            filterDirectory(
+                solrCore.withSearcher(searcher -> searcher.getIndexReader().directory()));
```
dsmiley, line 388: This seems to be the point of this PR.  Firstly, I'd rather see us get the Directory from the SolrCore.getDirectoryFactory as it doesn't seem very searcher-related.  Secondly, I think the DirectoryFactory ought to be able to return different different instances depending on the use-case/need.  See `DirectoryFactory.DirContext`.  Imagine having a new `REPLICATE` option that's only for moving bytes, for either backups or replication (both being similar)

bruno-roustant: I like that proposal very much!
I'll close this PR and create another.

bruno-roustant: New replacement [PR](https://github.com/apache/solr/pull/3185)

PR #3183, opened 2025-02-13
SOLR-17673: Disable multithreaded search at the node level by default

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17673

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -190,6 +190,10 @@ Bug Fixes
 
 * SOLR-17652: Fix a bug that could cause long leader elections to leave PULL replicas in DOWN state forever. (hossman)
 
+* SOLR-17673: Disable multi-threaded search execution by default at the node-level.
+  Users can still opt-in by providing a "indexSearcherExecutorThreads" > 0.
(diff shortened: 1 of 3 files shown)


dsmiley (approved):
What's Varun's relation here?

HoustonPutman:
> What's Varun's relation here?

He did a lot of debugging for the first calling out of this: https://issues.apache.org/jira/browse/SOLR-17642

gerlowskija (approved):

anshumg (approved):

risdenk:
I know I missed reviewing this - but this completely removes the ability to have it figure out number of processors right?

I wonder if it would make sense to set the default to -1 in the solr.xml we ship

if the value is 0 - set to num processors, its its -1 disable, if anything else use that value.

HoustonPutman:
Ok, yeah, I was a bit lazy with the solution. But I agree that it should be possible. Though from my (probably bad) viewpoint, `0` would sound like "disabled" to me, and `-1` would sound more like "unlimited" kind of like `rows=0` vs `rows=-1`.

Let me come up with something for this.

PR #3184, opened 2025-02-14
Exclude Kotlin multiplatform tests on -x test

malliaridis:
# Description

After some CI runs we have noticed that the Kotlin multiplatform test tasks of the new UI module, namely `allTests`, `desktopTest`, `wasmJsTest` and `wasmJsBrowserTest`, are not excluded when we run `gradlew check -x test`.

# Solution

This PR explicitly disables the tasks if `test` is part of the `excludedTaskNames`.

# Tests

You can test this issue by adding a failing test like below in the UI module `commonTest`:
```kotlin
@Test
fun failingTest() {
  fail("This test should always fail")
}
```

Without the code changes from this PR, you will notice that the tests are still executed on `gradlew check -x test`. With the code changes, the tests should not be executed.

Note that `gradlew check -x test --dry-run` will have the same output in both cases, probably because the exclusion rule is not resolved at this point.

# Checklist


risdenk (changes requested):
This approach seems wrong to me. I would expect changes to happen only in the solr/ui/build.gradle.kts to determine where the lifecycle is or maybe where the plugins are configured. This seems to change the behavior in a place I wouldn't expect.

dsmiley:
Yeah; agree with Kevin.

dsmiley:
Fundamentally, I think this is a hack, wherever it's placed.  But to stop some CI builds from being on fire, maybe this is what we need to do today.  Ideally we would better understand exactly why/how these tasks got attached to "check" instead of "test" and fix it.

dsmiley (approved):

HoustonPutman:
- https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/testing/internal/KotlinTestsRegistry.kt#L33
- https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/testing/internal/KotlinTestsRegistry.kt#L40

This is the reason, and I don't think there is any way around it, for now at least. We'd have to make those lines, and the other two for desktop and browser, configurable.

Overall yeah +1. I hate the hack, but I'm not sure there is a better way around it for now.

dsmiley:
@risdenk do you accept this PR, now that the hack is contained to the offending module?

risdenk (approved):

dsmiley:
I took the liberty of merging to expedite resolving failing builds.
Thanks Christos for getting this workaround in.

malliaridis:
@HoustonPutman thanks for the links, very insightful.

> We'd have to make those lines, and the other two for desktop and browser, configurable.

Not sure what you mean by configurable here. The only alternative to this workaround I could think of is to replace `check -x test` with a custom task for selectively running only the checks we want. But since `gradlew check -x test` is frequently used by devs, it would have some larger impact.

dsmiley:
Ideally we submit an issue to get JetBrains to fix this and they do it.

PR #3185, opened 2025-02-14
SOLR-17671 Replication and Backup use an unwrapped Directory.

bruno-roustant:
Currently Replication and Backup copy files using a Directory created with the configured DirectoryFactory. This Directory may be a FilterDirectory that adds additional logic (e.g. encryption) that should not run when copying files (e.g. do not decrypt).

The proposal is to add a new REPLICATE value in the DirectoryFactory.DirContext that would be used by replication, and a new BACKUP for backup to get the Directory to use. The DirectoryFactory would unwrap the Directory in this case.
One could expect that only one REPLICATE could be enough, but backup requires more inner checksum verifications that may need to differentiate the logic between the two (this is the case for encryption).

Example:
In the solr-sandbox encryption module, we would need a way to unwrap the Directory used to copy files during index fetching. Otherwise the files are decrypted by the EncryptionDirectory seamlessly during the files copy, ending up with follower replicas having cleartext index.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
+++ b/solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
@@ -428,7 +428,7 @@ public abstract class CachingDirectoryFactory extends DirectoryFactory {
 
   /**
    * Potentially filters or unwraps the cached {@link Directory} depending on the intended use
-   * defined by the {@link DirContext}.
+   * defined by the {@link org.apache.solr.core.DirectoryFactory.DirContext}.
    */
   protected Directory filterDirectory(Directory dir, DirContext dirContext) {
     // If the DirContext is REPLICATE or BACKUP, then unwrap the Directory to allow the caller to
--- a/solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
+++ b/solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
@@ -29,10 +29,12 @@ import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.lucene.store.AlreadyClosedException;
 import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.FilterDirectory;
 import org.apache.lucene.store.LockFactory;
 import org.apache.lucene.util.IOUtils;
 import org.apache.solr.common.SolrException;
@@ -59,8 +61,8 @@ public abstract class CachingDirectoryFactory extends DirectoryFactory {
     private boolean deleteOnClose = false;
 
     public CacheValue(String path, Directory directory) {
-      this.path = path;
-      this.directory = directory;
+      this.path = Objects.requireNonNull(path);
+      this.directory = Objects.requireNonNull(directory);
       this.closeEntries.add(this);
       // for debug
       // this.originTrace = new RuntimeException("Originated from:");
@@ -391,25 +393,22 @@ public abstract class CachingDirectoryFactory extends DirectoryFactory {
   public final Directory get(String path, DirContext dirContext, String rawLockType)
       throws IOException {
     String fullPath = normalize(path);
+    Directory directory;
+    CacheValue cacheValue;
     synchronized (this) {
       if (closed) {
         throw new AlreadyClosedException("Already closed");
       }
 
-      final CacheValue cacheValue = byPathCache.get(fullPath);
-      Directory directory = null;
-      if (cacheValue != null) {
-        directory = cacheValue.directory;
-      }
-
-      if (directory == null) {
+      cacheValue = byPathCache.get(fullPath);
+      if (cacheValue == null) {
         directory = create(fullPath, createLockFactory(rawLockType), dirContext);
         assert ObjectReleaseTracker.track(directory);
         boolean success = false;
         try {
-          CacheValue newCacheValue = new CacheValue(fullPath, directory);
-          byDirectoryCache.put(directory, newCacheValue);
-          byPathCache.put(fullPath, newCacheValue);
+          cacheValue = new CacheValue(fullPath, directory);
+          byDirectoryCache.put(directory, cacheValue);
+          byPathCache.put(fullPath, cacheValue);
           log.debug("return new directory for {}", fullPath);
           success = true;
         } finally {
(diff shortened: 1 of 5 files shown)


solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
```diff
@@ -421,9 +422,21 @@ public final Directory get(String path, DirContext dirContext, String rawLockTyp
...
-      return directory;
     }
+    return filterDirectory(directory, dirContext);
+  }
+
+  /**
+   * Potentially filters or unwraps the cached {@link Directory} depending on the intended use
+   * defined by the {@link DirContext}.
+   */
+  protected Directory filterDirectory(Directory dir, DirContext dirContext) {
```
bruno-roustant, line 433: I created a new filterDirectory method for two reasons:
- The key for cached directories does not include the DirContext.
- This method can be overridden if a DirectoryFactory implementation needs to adapt the filtering logic.

solr/core/src/java/org/apache/solr/core/DirectoryFactory.java
```diff
@@ -55,10 +55,16 @@ public abstract class DirectoryFactory implements NamedListInitializedPlugin, Cl
...
   public enum DirContext {
+    /** Default is index directory. */
     DEFAULT,
-    META_DATA
+    /** Directory containing metadata. */
+    META_DATA,
+    /** Directory used to copy raw files during replication. */
+    REPLICATE,
+    /** Directory used to copy raw files during backup. */
+    BACKUP,
```
bruno-roustant, line 67: I separated REPLICATE and BACKUP because backup requires more inner checksum verifications that may need to differentiate the logic between the two. (this is the case for encryption where we will need the full-fledged EncryptionDirectory to access the checksum inside the encrypted Lucene file).

bruno-roustant:
Actually in its current state, this PR does not work.
The Directory returned by the CachedDirectoryFactory may be different than the Directory cached internally. This results in unclosed directories.

One option is to cache using a key including the DirContext.
Another option is to rollback the DirContext idea and simply let the caller unwrap the Directory based on its usage. Either by directly calling FilterDirectory.unwrap, or by calling a new method DirectoryFactory.unwrap(Directory).

Actually I prefer option two. What do you think @dsmiley ?

solr/core/src/java/org/apache/solr/core/DirectoryFactory.java
bruno-roustant: Actually this very same question was discussed when I introduced DelegatingBackupRepository. During backup, two different checksums are verified:
- The Lucene format checksum, to verify the file to backup is not corrupted. This checksum can only be verified cleartext, so for encryption this means it is verified by the EncryptionDirectory after decrypting.
- The file transfer checksum, to verify the bytes are identical at the beginning and at the end of the copy. This checksum is verified by IncrementalShardBackup, and can be verified on either raw or filtered bytes. Currently it is on filtered bytes.

I specifically asked if we could skip the Lucene checksum verification in the case of encryption. But since we want to ensure we do not backup a corrupted file (which could replace an older valid one), then we have to check the cleartext bytes.

bruno-roustant:
This new commit makes the PR work. I added DirectoryFactory.unwrapFor(Directory, DirUseContext) method, with a context to potentially differentiate replication and backup.

dsmiley (commented):
So this has been an adventure.

The double-pairing of DirContext first to get the Directory (that seems maybe is ignored) and then remembering to unwrapFor is... wow pretty awkward and begs questions for anyone who might look at it.  Like, if I just want to do search, do I call unwrapFor(DEFAULT)?  And if just need meta, do I do unwrapFor(META)?  Basically, does the API suggest unwrap should always be called?  We know it doesn't but that's weird because the approach here is general.

Your astute observation that CachingDirectory doesn't consider the DirContext is a problem, standalone from this PR.  If someone passes METADATA first, it gets cached, and then later a DEFAULT would return that.  That's bad; for example HdfsDirectory layers the BlockCache directory above the base directory that's critical for performance.  

I see two paths for something elegant (neither involving the caller using unwrapFor):
(A) Incorporate the DirContext into the cache key, as you mentioned.  But that potentially means multiple Directory instances for a path, which feels kind of sad and why I avoided considering this initially.  But that could be improved so that the DirectoryFactory works with the cache to populate the layers independently so that you'd still only have one base (e.g. MMapDirectory) instance even if there are layers above.
(B) or CachingDirectory might have to wrap absolutely all Directory instances it returns with a delegate so that it can do it's release/close tracking.  All cache entries would be fully layered, but the DirectoryFactory could unwrap as needed for the DirContext use-case.

As the PR is at the moment... I'm -0

solr/core/src/java/org/apache/solr/handler/IncrementalShardBackup.java
```diff
@@ -145,21 +145,22 @@ protected IncrementalShardSnapshotResponse backup(final IndexCommit indexCommit)
...
     Directory dir =
-        solrCore
-            .getDirectoryFactory()
-            .get(
-                solrCore.getIndexDir(),
-                DirectoryFactory.DirContext.DEFAULT,
-                solrCore.getSolrConfig().indexConfig.lockType);
+        directoryFactory.get(
+            solrCore.getIndexDir(),
+            DirectoryFactory.DirContext.DEFAULT,
```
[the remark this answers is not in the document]
bruno-roustant: In theversion 2 of the PR, I separated the DirContext (what the directory is) and DirUseContext (what the directory will be used for). I put replication and backup in DirUseContext, so they cannot be passed to DirectoryFactory.get().

bruno-roustant:
As I see it, the need to unwrap is based on what the caller will use the Directory for. So Im' not sure putting a lot of logic in CachingDirectory in the best solution. Maybe the option B you propose is ok, or we could simply change the ReplicationAPIBase to always unwrap since it copies without Lucene checksum.

dsmiley:
(A) seems kinda hard due to various entanglements that seem to assume one cache entry per path or other complexities I see there.  At least (B) would actually only need to wrap dirs that are not DEFAULT.  
Also, the word "unwrap" pre-supposes an implementation detail that the callers should ideally not be exposed to and isn't necessarily accurate.  Maybe it'll wrap not unwrap or do both or something new; who knows.  Solr code using this should just give this DirContext hint as to what the dir will be used for.  That's how it is now, albeit we see a bug and the use-cases (DirContext) listed are insufficient.

bruno-roustant:
The new commit implements option B, which works, and the code remains concise.

solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
```diff
@@ -391,25 +393,22 @@ public boolean exists(String path) throws IOException {
...
-      final CacheValue cacheValue = byPathCache.get(fullPath);
-      Directory directory = null;
-      if (cacheValue != null) {
-        directory = cacheValue.directory;
-      }
-
-      if (directory == null) {
+      cacheValue = byPathCache.get(fullPath);
+      if (cacheValue == null) {
         directory = create(fullPath, createLockFactory(rawLockType), dirContext);
```
dsmiley, line 405: In a separate issue/PR, the `create` method should not contain the dirContext leaving a TODO now.  Or if you wish, make that change here; it's rather internal.

dsmiley (approved):
So glad to see option B worked out well!  Thanks for handling my push-back for better elegance well.

solr/core/src/java/org/apache/solr/handler/admin/api/ReplicationAPIBase.java
```diff
@@ -377,7 +380,7 @@ public void write(OutputStream out) throws IOException {
       try {
         initWrite();
 
-        Directory dir = solrCore.withSearcher(searcher -> searcher.getIndexReader().directory());
+        Directory dir = getDirectory();
```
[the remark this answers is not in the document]
dsmiley: I can't tell from the PR but this code will need to ensure we release the directory.

solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
bruno-roustant: I prefer to leave that for another PR (I have another proposal to make CachingDirectoryFactory support a DelegatingDirectoryFactory).
For the TODO, I'm unsure as I see some usage of the DirContext in HdfsDirectoryFactory which seems valid (creating either a BlockDirectory or a HdfsDirectory).

bruno-roustant:
@dsmiley I added CHANGES.txt for review.

dsmiley (approved):

solr/core/src/java/org/apache/solr/core/CachingDirectoryFactory.java
bruno-roustant: Thanks! I add the todo.

PR #3188, opened 2025-02-15
Address that the word solrcli doesnt find this main page in Ref Guide

epugh:
# Description

Use "SolrCLI" on the main page for the SolrCLI.

# Solution
this is a small fix, someday we need to rewrite/reorg the Solr Control Script page.    At least now the search bar finds "Solrcli"...


malliaridis (approved):
Didn't think of such optimizations before, but makes sense to improve it with these small changes. 👍

epugh:
i suspect somewhere there must be some "SEO optimization"in Antora, but this is good enough.

PR #3194, opened 2025-02-18
@SOLR-17675 Remove DirContext param from DirectoryFactory create method.

bruno-roustant:
https://issues.apache.org/jira/browse/SOLR-17675

In a previous [PR](https://github.com/apache/solr/pull/3185), we noted that the DirContext of the DirectoryFactory.create method should be removed. It was used only by the HdfsDirectoryFactory (now removed), and there was a bug in CachingDirectoryFactory which would create a Directory with this param, and cache it without this param in the cache key.

We already added a todo in DirectoryFactory, and the proposal is to apply this todo: remove the DirContext param from the DirectoryFactory.create method.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -148,6 +148,8 @@ Other Changes
 
 * SOLR-17548: Switch all public Java APIs from File to Path. (Matthew Biscocho via Eric Pugh)
 
+* SOLR-17675: Remove DirContext param from DirectoryFactory.create method. (Bruno Roustant, David Smiley)
+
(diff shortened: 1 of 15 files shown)


bruno-roustant:
Is this a public API change that should that be only for main 10?

epugh:
Since HDFS was only removed on main 10, I *think* this refactoring should only go on main as well?   9x will still have HdfsDirectoryFactor.   LGTM.

dsmiley (approved):
Thanks for getting this done Bruno.
Agreed that this can be v10 only; it's coming up and no hurry.

PR #3197, opened 2025-02-19
SOLR-17650: Fix tests for unordered buffered updates

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17650

According to the discussion in the JIRA, the behavior here is fine. We just need to update the tests now that buffered updates can be done in parallel.

The new testing stuff is heavy handed, but mostly needed to 1) test this at all and 2) give us a good error message that we can actually use.

### DIFF
--- a/solr/test-framework/src/java/org/apache/solr/SolrTestCaseJ4.java
+++ b/solr/test-framework/src/java/org/apache/solr/SolrTestCaseJ4.java
@@ -142,6 +142,7 @@ import org.apache.solr.util.SSLTestConfig;
 import org.apache.solr.util.TestHarness;
 import org.apache.solr.util.TestInjection;
 import org.apache.zookeeper.KeeperException;
+import org.hamcrest.Matcher;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Rule;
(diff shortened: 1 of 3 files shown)


solr/test-framework/src/java/org/apache/solr/util/SolrMatchers.java
```diff
@@ -0,0 +1,81 @@
...
+ * limitations under the License.
+ */
+package org.apache.solr.util;
+
+import java.util.List;
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeDiagnosingMatcher;
+
+public class SolrMatchers {
```
dsmiley, line 24: add a bit of javadoc please.  Like reference "Hamcrest", and javadoc on the subListMatches method.

solr/test-framework/src/java/org/apache/solr/SolrTestCaseJ4.java
```diff
@@ -1027,6 +1028,64 @@ public static String assertJQ(SolrQueryRequest req, double delta, String... test
...
+   * @param message Failure message for test
+   * @param test Matcher for the given object returned from deserializing the response
+   * @return The request response as a JSON String if the test matcher passes
+   */
+  @SuppressWarnings("unchecked")
+  public static <T> String assertThatJQ(SolrQueryRequest req, String message, Matcher<T> test)
+      throws Exception {
+    SolrParams params = null;
+    try {
+      params = req.getParams();
```
dsmiley, line 1052: you could get & set at the declaration rather than split across the `try`.  Could then be final.

solr/test-framework/src/java/org/apache/solr/SolrTestCaseJ4.java
```diff
@@ -1027,6 +1028,64 @@ public static String assertJQ(SolrQueryRequest req, double delta, String... test
...
+   * <p>Please use this with care: this makes it easy to match complete structures, but doing so can
+   * result in fragile tests if you are matching more than what you want to test.
+   *
+   * @param req Solr request to execute
+   * @param message Failure message for test
+   * @param test Matcher for the given object returned from deserializing the response
+   * @return The request response as a JSON String if the test matcher passes
+   */
+  @SuppressWarnings("unchecked")
+  public static <T> String assertThatJQ(SolrQueryRequest req, String message, Matcher<T> test)
```
dsmiley, line 1048: You are adding more methods that uses the almost but not quite yet deprecated TestHarness.  Can you please instead consider changing the approach of this method to look like `org.apache.solr.SolrTestCaseHS#assertJQ` which takes a SolrClient?  Let's say you agree to do that.  My next thought is... should SolrTestCase have such extensive utility methods or should they live on a helper class (a new one I guess).  I lean to the latter but I haven't thought about that much yet.

dsmiley: That's fair

PR #3198, opened 2025-02-19
Fix race condition in TestThinCache

HoustonPutman:
Basically `onRemoval()` has a much weaker guarantee than it is assuming right now.
This method can be called asynchronously, after the key is already gone locally.
Or even after the new cache is created, and it doesn't have the value that is being removed in it's local cache.

This is a way to solve most of the issues, but one race condition still remains. (Not sure it's possible to solve)
If the cache evicts a key, then the user puts that key with its original value right afterwards, this code will still assume it should be removed. Since we don't have access to timings for removal and addition, there is no way to fix this. (I think).

But for the sake of our small test failures, this should be sufficient.

### DIFF
--- a/solr/core/src/test/org/apache/solr/search/ThinCache.java
+++ b/solr/core/src/test/org/apache/solr/search/ThinCache.java
@@ -345,10 +345,25 @@ public class ThinCache<S, K, V> extends SolrCacheBase
 
   @Override
   public void onRemoval(K key, V value, RemovalCause cause) {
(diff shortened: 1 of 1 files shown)


magibney (approved):
Looks good, thanks for fixing this.

I'll put some thought towards whether there's a way to fix it for the "same value" case.

HoustonPutman:
In the meantime I'll merge, since this is failing the tests occasionally.

PR #3199, opened 2025-02-20
IOUtils: add back closeQuietly(Closeable)

dsmiley:
To retain compatibility within 9x.

Not bringing this to main/10x.
This issue was introduced by [SOLR-17593: New home for default http2 client apache/solr#2689](https://github.com/apache/solr/pull/2689) and was seen to be a compatibility issue in the DIH for one: https://github.com/SearchScale/dataimporthandler/issues/91

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/common/util/IOUtils.java
+++ b/solr/solrj/src/java/org/apache/solr/common/util/IOUtils.java
@@ -16,6 +16,7 @@
  */
 package org.apache.solr.common.util;
 
+import java.io.Closeable;
 import java.lang.invoke.MethodHandles;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


mkhludnev (approved):

mkhludnev:
Hm.. couldn't it cause 'unambiguous error'? I mean now we have two methods with (param) overloaded signature, almost all arguments passed into these methods implement both interfaces. Just worry if javac is able to distinguish between these two methods.

mkhludnev:
Ok. ~chatting to deepseek~ referring to JLS 15.12.2.5 calms me down a little. Should be fine, since the most specific method will be chosen.

janhoy (approved):

dsmiley:
Ported to branch_9_8 so that it gets in 9.8.1

PR #3202, opened 2025-02-20
LogListener: remove no-arg convenience methods

dsmiley:
These are hazardous/sloppy.  Tests should listen to a specific class/package.

Motivated by: https://issues.apache.org/jira/browse/SOLR-17667?focusedCommentId=17928175&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17928175

### DIFF
--- a/solr/test-framework/src/java/org/apache/solr/util/LogListener.java
+++ b/solr/test-framework/src/java/org/apache/solr/util/LogListener.java
@@ -160,7 +160,6 @@ public final class LogListener implements Closeable, AutoCloseable {
   }
 
   // TODO: more factories for other levels?
-  // TODO: no-arg factory variants that use "" -- simpler syntax for ROOT logger?
 
   private static LogListener create(final Level level, final String logger) {
     final String name = createName(level);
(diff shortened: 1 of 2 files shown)


risdenk (approved):

hossman:
Not really sure i understand why you consider them hazardous?

FWIW: The intent of these was to make it as easy as possible for folks to write tests asserting that "No code run in this block should cause any code, anywhere, to log an ERROR (WARN, etc...)" but see no great loss in making people write `LogListener.error("")` instead of `LogListener.error()`

dsmiley:
For the hazard, see what I linked to. Houston agrees.

madrob:
Maybe we remove the debug (and info?) no-arg versions, leave a comment why, and leave the warn and error versions around?

hossman:
> For the hazard, see what I linked to. Houston agrees.

Ah ok -- yeah, sorry i didn't notice that jira link. 

The crux of the problem being that in order to listen to the logs at a particular level, we must enable the logs at that level, and enabling something like DEBUG at the ROOT level can really slow things down.

Totally valid point.

> Maybe we remove the debug (and info?) no-arg versions, leave a comment why, and leave the warn and error versions around?

We could, but that functionality is still available by referring to the root level logger by `""` (which is already mentioned as possible in the jdocs) for the rare case where it's useful ... I agree with David: we should "Make the common/safe thing easy, Make the rare/scary thing hard"

BTW David: skimming the code i see an out of date TODO that should also be removed..

```
  // TODO: no-arg factory variants that use "" -- simpler syntax for ROOT logger?
```

(don't want to leave a TODO suggesting someone re-add the thing you're removing)

PR #3203, opened 2025-02-20
SOLR-17677: Ensure DBQ is safe before running

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-17677

# Description

Several Solr Query implementations require a SolrIndexSearcher which is not always available.  This often causes ClassCastExceptions and other errors when the queries are run by a non-SIS IndexSearcher, such as during delete-by-query operations.
 
# Solution

This PR works around this limitation by detecting and blocking these queries during DBQs:

1. A marker interface `SolrSearcherRequirer` (not thrilled with the name), is introduced that can by used by Query implementations that want to indicate their reliance on SolrIndexSearcher.
2. A `QueryVisitor` implementation `SolrSearcherRequirementDetector` (again - could use help with naming) is introduced which uses Lucene's `visit()` support to determine if a part of a query requires a SIS.
3. The DBQ code in DirectUpdateHandler2 uses this Visitor to throw a SolrException(400) if any unsupported queries are detected.

A better, longer-term fix would be to modify the implementation of these queries to not rely on SIS, but that approach would likely miss the upcoming 9.8.1 release.  The solution above is sub-par, but at least gives us something in time for the 9.8.1 release.

# Tests

New test cases in TestCloudDeleteByQuery.  Unit tests in SolrSearcherRequirementDetectorTest.

# Checklist


gerlowskija:
Check and tests pass for me locally.

The `precommit` PR check fails with the error below, but this is almost certainly unrelated to the PR.  Will investigate a bit though to see if I can resolve...

```
/home/runner/work/solr/solr/solr/core/src/java/org/apache/solr/cloud/OverseerTaskProcessor.java:505: error: cannot find symbol
    } catch (AlreadyClosedException ignored) {
             ^
  symbol:   class AlreadyClosedException
  location: class OverseerTaskProcessor
> Task :solr:core:compileJava
```

gerlowskija:
Alright - looks like that was a problem on the 'main' I was based on - there was a revert on main and I've rebased on that now.

Should be good to go.

PR #3204, opened 2025-02-21
SOLR-16903: Switch CoreContainer#getSolrHome to return Path instead of String

AndreyBozhko:

# Description

This is only a part of SOLR-16903, prompted by the comment here: https://github.com/apache/solr/pull/3146#discussion_r1935722062

# Solution

Updated public getter to return Path.

Suggested review order commit-by-commit.

# Tests

Existing tests.

# Checklist


solr/core/src/java/org/apache/solr/core/ZkContainer.java
```diff
@@ -87,7 +87,7 @@ public void initZooKeeper(final CoreContainer cc, CloudConfig config) {
     // TODO: remove after updating to an slf4j based zookeeper
     System.setProperty("zookeeper.jmx.log4j.disable", "true");
 
-    String solrHome = cc.getSolrHome();
+    String solrHome = cc.getSolrHome().toString();
```
epugh, line 90: Just want to confirm that we are still stuck with String here?   I see below the `Paths.get`

solr/solrj/src/test/org/apache/solr/client/solrj/request/TestCoreAdmin.java
```diff
@@ -80,7 +80,7 @@ public void testCustomUlogDir() throws Exception {
       cores.getAllowPaths().add(dataDir.toPath()); // Allow the test dir
       cores.getAllowPaths().add(newCoreInstanceDir.toPath()); // Allow the test dir
 
-      File instanceDir = new File(cores.getSolrHome());
+      File instanceDir = cores.getSolrHome().toFile();
```
epugh, line 83: Could we stay in path land if we change FileUtils?

epugh (approved):
LGTM.  A comment or two, but if the tests path, this looks nice.

solr/core/src/java/org/apache/solr/core/ZkContainer.java
mlbiscoc: If I am reading this right, this can be removed and the call below doesn't need `Paths.get()` but can just be `cc.getSolrHome().resolve(zoo_data).toString()`

solr/core/src/java/org/apache/solr/handler/admin/SystemInfoHandler.java
```diff
@@ -145,7 +145,7 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
       rsp.add("zkHost", getCoreContainer(req).getZkController().getZkServerAddress());
     }
     if (cc != null) {
-      rsp.add("solr_home", cc.getSolrHome());
+      rsp.add("solr_home", cc.getSolrHome().toString());
```
mlbiscoc, line 148: `toString()` might not be necessary here because of this [textWriter](https://github.com/apache/solr/blob/d2fdb16a78195e5eb40caf48761afeb1bce041ee/solr/solrj/src/java/org/apache/solr/common/util/TextWriter.java#L78) I think but this probably needs to be confirmed.

mlbiscoc: But also that TextWriter is doing a `.toAbsolutePath()` which doesn't seem right to me. If SolrHome is always absolute this wouldn't matter, if it's relative, then possible the response would change. Maybe ignoring what I said above is the safest option.

solr/core/src/java/org/apache/solr/core/ZkContainer.java
AndreyBozhko: Thanks, updated.

solr/solrj/src/test/org/apache/solr/client/solrj/request/TestCoreAdmin.java
AndreyBozhko: Yes, it's possible to replace FileUtils here with `org.apache.commons.io.file.PathUtils`. For now I changed just the callsites for CoreContainer#getSolrHome, to simplify the review.

solr/core/src/java/org/apache/solr/handler/admin/SystemInfoHandler.java
epugh: ;-)

dsmiley:
Excellent!  I love how focused this PR is.

Can you please edit the JIRA association to be [SOLR-16903](https://issues.apache.org/jira/browse/SOLR-16903) -- PR title & CHANGES.txt.  There's already a CHANGES.txt entries for 16903.   Needn't specify every single API endpoint / category since by 10.0, I think we hope we can basically say, "Changed all Java APIs using File to NIO Path" and not get into further details.

AndreyBozhko:
Sounds good, I updated the PR.

solr/core/src/java/org/apache/solr/handler/admin/SystemInfoHandler.java
dsmiley: Good eye!  I agree with removing toString here.  Whether toAbsolutePath should be called or not is another question from this PR; I'm ambivalent.

AndreyBozhko: Makes sense - the JavaBinCodec looks to be doing the same thing. https://github.com/apache/solr/blob/76c09a35dba42913a6bcb281b52b00f87564624a/solr/solrj/src/java/org/apache/solr/common/util/JavaBinCodec.java#L411-L413

dsmiley:
I plan to merge this tonight; it's very straightforward

mlbiscoc (approved):

PR #3208, opened 2025-02-24
SOLR-17644: Fix missing auth listener

iamsanjay:
https://issues.apache.org/jira/browse/SOLR-17644

# Checklist

- [] I have added tests for my changes.

### DIFF
--- a/solr/core/src/test/org/apache/solr/security/BasicAuthIntegrationTest.java
+++ b/solr/core/src/test/org/apache/solr/security/BasicAuthIntegrationTest.java
@@ -49,8 +49,11 @@ import org.apache.solr.client.solrj.request.QueryRequest;
 import org.apache.solr.client.solrj.request.RequestWriter.StringPayloadContentWriter;
 import org.apache.solr.client.solrj.request.UpdateRequest;
 import org.apache.solr.client.solrj.request.V2Request;
+import org.apache.solr.client.solrj.request.beans.PluginMeta;
 import org.apache.solr.client.solrj.response.QueryResponse;
 import org.apache.solr.cloud.SolrCloudAuthTestCase;
+import org.apache.solr.cluster.placement.PlacementPluginFactory;
+import org.apache.solr.cluster.placement.plugins.MinimizeCoresPlacementFactory;
 import org.apache.solr.common.SolrInputDocument;
 import org.apache.solr.common.cloud.DocCollection;
 import org.apache.solr.common.cloud.Replica;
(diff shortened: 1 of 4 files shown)


iamsanjay:
This is not the final solution. I created this to initiate discussion on the best possible approach to solving this.

iamsanjay:
Test case!

```
@Test
  public void testAuth() throws Exception {
    // This succeeds with auth enabled
    CollectionAdminRequest.createCollection("c123", "conf", 1, 1)
        .setBasicAuthCredentials(SecurityJson.USER, SecurityJson.PASS)
        .process(cluster.getSolrClient());
    cluster.waitForActiveCollection("c123", 1, 1);

    // Configure placement plugin
    PluginMeta plugin = new PluginMeta();
    plugin.name = PlacementPluginFactory.PLUGIN_NAME;
    plugin.klass = MinimizeCoresPlacementFactory.class.getName();
    V2Request v2Request =
        new V2Request.Builder("/cluster/plugin")
            .forceV2(true)
            .POST()
            .withPayload(singletonMap("add", plugin))
            .build();
    v2Request.setBasicAuthCredentials(SecurityJson.USER, SecurityJson.PASS);
    v2Request.process(cluster.getSolrClient());

    // Now this will fail with a 401 !!
    CollectionAdminRequest.createCollection("c456", "conf", 1, 1)
        .setBasicAuthCredentials(SecurityJson.USER, SecurityJson.PASS)
        .process(cluster.getSolrClient());
    cluster.waitForActiveCollection("c456", 1, 1);
    /// /////////
  }
```

dsmiley (commented):
Shouldn't this PR include a test like Colvin shared -- https://github.com/colvinco/solr/blob/f2e8825c5e46180eb121f20a69f246092ad60776/solr/core/src/test/org/apache/solr/security/BasicAuthIntegrationTest.java#L398-L403 so we can verify the fix?

(Doubtful but asking anyway) Is the bug a result of ZkController.getSolrCloudManager lazy creating the client instead of eager?  I'm suspicious the lazy load saves anything of consequence.

Maybe CoreContainer.loadInternal needs to move these lines:
```java
securityConfHandler =
        isZooKeeperAware() ? new SecurityConfHandlerZk(this) : new SecurityConfHandlerLocal(this);
    reloadSecurityProperties();
```
to immediately after `zkSys.initZooKeeper....`

iamsanjay:
Screenshot 2025-02-24 at 12 58 11 PM

It's not about the lazy loading. `SolrClientProvider` object is being used inside the `ZkController#getSolrCloudManager()` but at that time auth listener is not available. 

zkSys.initZooKeeper(this, cfg.getCloudConfig()); is what triggers somewhere `ZkController#getSolrCloudManager()` 

But as you can see in the IF block we initialize the `pkiAuthenticationSecurityBuilder` after that.

https://github.com/apache/solr/blob/76c09a35dba42913a6bcb281b52b00f87564624a/solr/core/src/java/org/apache/solr/core/CoreContainer.java#L868-L880

dsmiley:
I would _prefer_ we initialize the httpClients with security as soon as possible, before they are obtained, even if the design seems to support dynamic changes after creation.  It'd be better to guarantee the initialization has happened.  But it seems to awkward to guarantee that right now.  On my mind is dependency-injection frameworks which can manage complex object graphs with dependencies for startup initialization.  Any way that's kind of a fantasy right now, and a big undertaking to embrace that.

I thought of a really simple solution here.  I'm skeptical that ZkController.getSolrCloudManager truly *needs* to create yet another Http2SolrClient instance.  It could use `getCoreContainer().getDefaultHttpSolrClient();` (coming from HttpSolrClientProvider, which is instrumented with security, albeit later but that's okay).  If we want to mess with timeout tuning then I think that should be HttpSolrClientProvider instance to consider that (which is configurable!) and have it be more global.   getSolrCloudManager isn't special.

iamsanjay:
We are creating client inside ZkController with custom timeouts
```
 http2SolrClient =
          new Http2SolrClient.Builder()
              .withHttpClient(cc.getDefaultHttpSolrClient())
              .withIdleTimeout(30000, TimeUnit.MILLISECONDS)
              .withConnectionTimeout(15000, TimeUnit.MILLISECONDS)
              .build();
```
And HttpSolrClientProvider timeouts controlled via UpdateShardHandlerConfig
```
httpClientBuilder
          .withConnectionTimeout(cfg.getDistributedConnectionTimeout(), TimeUnit.MILLISECONDS)
          .withIdleTimeout(cfg.getDistributedSocketTimeout(), TimeUnit.MILLISECONDS)
          .withMaxConnectionsPerHost(cfg.getMaxUpdateConnectionsPerHost());
```

iamsanjay:
If I look the default connection and socket timeouts.

```
HttpClientUtil.DEFAULT_SO_TIMEOUT = 60,000
HttpClientUtil.DEFAULT_CONNECT_TIMEOUT = 60,000
```
But I will make the change to ZkController to use the HttpSolrClientProvider instead of creating one. I also like this solution better. I only wonder if reducing the default timeouts was the reason in the first-place it was re-created here in ZkController.

solr/core/src/test/org/apache/solr/security/BasicAuthIntegrationTest.java
```diff
@@ -373,6 +376,32 @@ public void testBasicAuth() throws Exception {
...
+      V2Request v2Request =
+          new V2Request.Builder("/cluster/plugin")
+              .forceV2(true)
+              .POST()
+              .withPayload(singletonMap("add", plugin))
+              .build();
+      v2Request.setBasicAuthCredentials("harry", "HarryIsUberCool");
+      v2Request.process(cluster.getSolrClient());
+
+      // Now this will fail with a 401 !!
```
dsmiley, line 398: Well it won't any longer :-), once we merge a fix.  You can mention a JIRA issue where it did.

dsmiley (approved):
Assuming tests pass -- looks good!
Withought GHA Crave, try "crave run" at CLI if you have that or just run directly/manually.

dsmiley (approved):
Cool; just need a CHANGES.txt

PR #3209, opened 2025-02-25
SOLR-17682: QueryResponseWriter hierarchy refactor

dsmiley:
QRW now does write(OutputStream) and NOT write(Writer). BinaryQueryResponseWriter is removed
QRW.writeToString for tests
TextQueryResponseWriter added as base for all text QRWs QueryResponseWriterUtil is removed

https://issues.apache.org/jira/browse/SOLR-17682


solr/core/src/java/org/apache/solr/response/BinaryResponseWriter.java
```diff
@@ -45,12 +44,14 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class BinaryResponseWriter implements BinaryQueryResponseWriter {
+/** Solr's "javabin" format. TODO rename accordingly. */
```
dsmiley, line 47: worth doing in another PR; maybe same JIRA

solr/core/src/java/org/apache/solr/response/BinaryResponseWriter.java
```diff
@@ -59,12 +60,6 @@ public void write(OutputStream out, SolrQueryRequest req, SolrQueryResponse resp
     }
   }
 
-  @Override
-  public void write(Writer writer, SolrQueryRequest request, SolrQueryResponse response)
-      throws IOException {
-    throw new RuntimeException("This is a binary writer , Cannot write to a characterstream");
```
dsmiley, line 65: The existence of this was a code-smell that the hierarchy was wrong; promoting my efforts here

solr/core/src/java/org/apache/solr/response/JacksonJsonWriter.java
```diff
@@ -43,17 +47,42 @@ public JacksonJsonWriter() {
...
 
+  // let's also implement the binary version since Jackson supports that (probably faster)
   @Override
-  public void write(OutputStream out, SolrQueryRequest request, SolrQueryResponse response)
+  public void write(
+      OutputStream out, SolrQueryRequest request, SolrQueryResponse response, String contentType)
       throws IOException {
-    WriterImpl sw = new WriterImpl(jsonfactory, out, request, response);
+    out = new NonFlushingStream(out);
+    // resolve the encoding
```
dsmiley, line 56: this charSet check logic is new.  Since Jackson can specify the encoding and since the contentType (with possibly the encoding) is now passed in (new), made sense to choose the right one.

solr/core/src/java/org/apache/solr/response/QueryResponseWriter.java
```diff
@@ -39,23 +41,26 @@
...
-   * <code>req</code> but the dominant source of information should be <code>rsp</code>.
-   *
-   * <p>There are no mandatory actions that write must perform. An empty write implementation would
-   * fulfill all interface obligations.
+   * Writes the response to the {@link OutputStream}. {@code contentType} is from {@link
+   * #getContentType(SolrQueryRequest, SolrQueryResponse)}, and it's often ignored.
    */
-  public void write(Writer writer, SolrQueryRequest request, SolrQueryResponse response)
+  void write(
+      OutputStream out, SolrQueryRequest request, SolrQueryResponse response, String contentType)
```
dsmiley, line 55: Added contentType since the production code path will have this already, and we'll need it in TextQueryResponseWriter

solr/core/src/java/org/apache/solr/response/RawResponseWriter.java
```diff
@@ -88,41 +85,27 @@ protected QueryResponseWriter getBaseWriter(SolrQueryRequest request) {
...
-    if (obj != null && (obj instanceof ContentStream)) {
-      return ((ContentStream) obj).getContentType();
+    if (obj instanceof ContentStream content) {
+      return content.getContentType();
     }
     return getBaseWriter(request).getContentType(request, response);
   }
 
   @Override
-  public void write(Writer writer, SolrQueryRequest request, SolrQueryResponse response)
```
dsmiley, line 98: We don't need a Writer/Reader variant of this (I think).

solr/core/src/java/org/apache/solr/response/TextQueryResponseWriter.java
```diff
@@ -81,7 +83,8 @@ private static Writer buildWriter(OutputStream outputStream, String charset)
    *
    * <p>See SOLR-8669.
    */
-  private static class NonFlushingStream extends OutputStream {
+  // nocommit discuss moving to SolrDispatchFilter wrapper.  If keep them move?
```
dsmiley, line 86: This is an observation I had to do things better.  Basically, don't we want Solr to always prevent a flush, not just specifically here only sometimes for some QueryResponseWriters?  If so, it can easily go on the close shield thing (see elsewhere in this PR to show how.  Then we can drop this NonFlushingStream thing and not worry about wether we should use it in other scenarios.  CC @magibney wonder if you have thoughts on this one

solr/core/src/java/org/apache/solr/servlet/DirectSolrConnection.java
```diff
@@ -114,14 +111,7 @@ public String request(SolrRequestHandler handler, SolrParams params, String body
...
       // Now write it out
-      QueryResponseWriter responseWriter = core.getQueryResponseWriter(req);
-      if (responseWriter instanceof BinaryResponseWriter) {
-        return ((BinaryResponseWriter) responseWriter).serializeResponse(req, rsp);
-      } else {
-        StringWriter out = new StringWriter();
-        responseWriter.write(out, req, rsp);
-        return out.toString();
-      }
+      return req.getResponseWriter().writeToString(req, rsp);
```
dsmiley, line 114: This pattern happens in many places.  The existence of it and repeating was a code-smell that the hierarchy was wrong; promoting my efforts here

solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
```diff
@@ -125,6 +125,12 @@ public void close() {
                 : CLOSE_STREAM_MSG;
             stream = ClosedServletOutputStream.CLOSED_SERVLET_OUTPUT_STREAM;
           }
+
+          @Override
+          public void flush() throws IOException {
+            // nocommit discuss
```
dsmiley, line 131: And here's the front door of Solr; super simple to prevent flushes here too

solr/core/src/test/org/apache/solr/request/TestWriterPerf.java
```diff
@@ -179,18 +175,9 @@ void doPerf(String writerName, SolrQueryRequest req, int encIter, int decIter) t
...
     for (int i = 0; i < encIter; i++) {
-      if (w instanceof BinaryQueryResponseWriter binWriter) {
-        out = new ByteArrayOutputStream();
-        binWriter.write(out, req, rsp);
-        out.close();
-      } else {
-        out = new ByteArrayOutputStream();
-        // to be fair, from my previous tests, much of the performance will be sucked up
-        // by java's UTF-8 encoding/decoding, not the actual writing
-        Writer writer = new OutputStreamWriter(out, StandardCharsets.UTF_8);
```
dsmiley, line 190: I dropped this branch because in normal use we'll write to an OutputStream, not a Writer.

solr/core/src/java/org/apache/solr/response/SmileResponseWriter.java
```diff
@@ -24,16 +24,27 @@
...
+  public void write(
+      OutputStream out, SolrQueryRequest request, SolrQueryResponse response, String contentType)
       throws IOException {
     try (SmileWriter sw = new SmileWriter(out, request, response)) {
       sw.writeResponse();
     }
   }
 
+  @Override
+  public String getContentType(SolrQueryRequest request, SolrQueryResponse response) {
```
dsmiley, line 44: Smile was wrongly subclassing BinaryResponseWriter (JavaBin) and thus inheriting its content type.  It was an innocent mistake enabled by BinaryResponseWriter's bad name; should say "JavaBin" in there.

dsmiley (commented):
10.0 only; no deprecations

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
```diff
@@ -47,12 +47,15 @@
  * org.apache.solr.handler.admin.MetricsHandler}
  */
 @SuppressWarnings(value = "unchecked")
-public class PrometheusResponseWriter extends RawResponseWriter {
```
dsmiley, line 50: @mlbiscoc why did you choose RawResponseWriter?  It's a very special QueryResponseWriter and I don't see that the this response writer utilizes it.

mlbiscoc: I don't fully remember. I think I followed a different writer at the time and this one took an `OutputStream` which was what I was looking for but please with change if this is not correct.

solr/core/src/java/org/apache/solr/response/QueryResponseWriter.java
dsmiley: It's a debatable change... it'd be less disruptive to leave it as it was and then only the text formats will re-acquire it which is fairly trivial parsing to do a second time.  Shrug.

dsmiley:
Planning to merge in a couple days without further feedback

PR #3210, opened 2025-02-25
GitHub Action: precommit: --continue

dsmiley:
We'd like to see all errors found when the "precommit" GitHub Action runs, not just the first error.

note: I'll revert the intentionally bad commit that's there to validate it works


dsmiley:
As expected, the GHA logs for precommit show several errors (that I intentionally induced) being reported, instead of only the first.    I'll revert the intentional errors so this is mergeable.

gerlowskija (approved):
LGTM once the intentional precommit breaks are removed!

PR #3211, opened 2025-02-25
Fix get status for ping response

umut-sar:
Update the example which gets the ping status "OK" value from the response.

### DIFF
--- a/solr/solr-ref-guide/modules/deployment-guide/pages/ping.adoc
+++ b/solr/solr-ref-guide/modules/deployment-guide/pages/ping.adoc
@@ -79,7 +79,7 @@ A status=OK indicates that the nodes are responding.
 SolrPing ping = new SolrPing();
 ping.getParams().add("distrib", "true"); //To make it a distributed request against a collection
 rsp = ping.process(solrClient, collectionName);
-int status = rsp.getStatus();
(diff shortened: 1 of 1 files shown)


epugh:
This is a nice quick fix, however, could I suggest another approach that would be more robust?  In other places in the ref guide, we have java code samples.  To ensure they don't fall behind, we actually have a unit test run to test them, and then we insert the code into the ref guide.

Would you be interested in taking that approach here?

for an example, look at `solrj.adoc`:

```

[source,java,indent=0]
----
include::example$UsingSolrJRefGuideExamplesTest.java[tag=solrj-query-with-solrquery]
----

```

umut-sar:
Hi @epugh,

Thanks for the clarification! Your approach makes a lot of sense. I'll review solrj.adoc and explore implementing this method in the next few days / weeks.

Best, Umut Saribiyik

epugh:
Awesome..   Don't hestitate to ping me for help/review....   I'd like to see all our Java code samples in the Ref Guide use this approach so we don't have regressions creep in.

umut-sar:
I opened a ticket for this issue. Feel free to add your suggestions to it.

https://issues.apache.org/jira/browse/SOLR-17684

I will close this PR and recreate a new one later.

PR #3213, opened 2025-02-25
Fix flaky TestSchemaDesignerAPI

HoustonPutman:
This test currently flakes because the last command of `SchemaDesignerApi.publish()` is to persist a change to the overlay.json for the newly created collection.

In `TestSchemaDesignerAPI.testAddTechproductsProgressively()`, this is called near the end of the test. So quickly after it is called, the `@AfterClass` method deletes all of the collections.

However, persisting a change to the overlay.json will asynchronously kick-off a reload of the collection.

So there is a race condition here that if this reload starts before the collection deletion but doesn't finish before the deletion, then lots of errors will happen and ultimately the `StandardDirectoryFactory` will fail to delete directories, because they have already been deleted. And annoyingly `PathUtils.deleteDirectory()` cannot handle parallel deletes gracefully, it will throw `FileNotFoundException`s. So in this case, the CachingDirectoryFactory is not able to successfully close things, and we get an error that a `NRTCachingDirectory` is not closed when it should be.

So ultimately my answer to this is to:

- Move the new overlay.json to before the collection is created. In my mind, there is no reason to update the collection once its created, because we can just as easily create it with the settings we want.
- Change ZkContainer to interrupt the core registry thread pool, because there should be no reason to continue to register cores while the zkContainer is shutting down. The ConfigSet watching code is run via this executor, so stopping it in its tracks can only help us.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/ZkContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/ZkContainer.java
@@ -246,7 +246,7 @@ public class ZkContainer {
           zkServer.stop();
         }
       } finally {
-        ExecutorUtil.shutdownAndAwaitTermination(coreZkRegister);
+        ExecutorUtil.shutdownNowAndAwaitTermination(coreZkRegister);
       }
     }
   }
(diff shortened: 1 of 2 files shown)


dsmiley (approved):
It's a shame that SolrCloud isn't gracefully handling deleting a collection that has reloads in progress.  Doing something about that would be a root cause fix but anyway, the changes you did here make sense to me.

HoustonPutman:
Yeah, I agree about that. Will think a bit about it before merging.

PR #3214, opened 2025-02-26
SOLR-17623: SimpleOrderedMap equals, hashCode (and for Map.Entry)

dsmiley:
And ensure a NamedList is never equal to a SimpleOrderedMap.

https://issues.apache.org/jira/browse/SOLR-17623

CC @renatoh


dsmiley:
LMK if this looks good to you @renatoh

PR #3215, opened 2025-02-26
SOLR-17518: Remove deprecated XML methods from UpdateRequest

psalagnac:
https://issues.apache.org/jira/browse/SOLR-17518

This is a follow-up of #3200.
Deprecated methods are now removed in Solr 10.


epugh (approved):
LGTM.   Seems slightly odd that the one test needed some extra code?  is there a refactoring of that test that would eliminate the `getXML` helper method?

psalagnac:
Thanks a good point, thanks. All other tests use test utility code for that!

PR #3217, opened 2025-02-26
SOLR-17633 Add matomo script to ref-guide

janhoy:
https://issues.apache.org/jira/browse/SOLR-17633

Adds matomo tracking to ref guide.

### DIFF
--- a/solr/solr-ref-guide/ui-src/partials/head-icons.hbs
+++ b/solr/solr-ref-guide/ui-src/partials/head-icons.hbs
@@ -1,3 +1,20 @@
 <!-- Solr Changes - Start -->
     <link rel="icon" href="{{{uiRootPath}}}/img/favicon.ico" type="image/x-icon">
+    {{#if env.MATOMO_ENABLED}}
+    <script>
+      var _paq = window._paq = window._paq || [];
+      /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
+      _paq.push(["setDoNotTrack", true]);
+      _paq.push(["disableCookies"]);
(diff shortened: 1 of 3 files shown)


solr/solr-ref-guide/ui-src/partials/head-icons.hbs
```diff
@@ -1,3 +1,20 @@
 <!-- Solr Changes - Start -->
     <link rel="icon" href="{{{uiRootPath}}}/img/favicon.ico" type="image/x-icon">
+    {{#if env.MATOMO_ENABLED}}
+    <script>
```
janhoy, line 4: I piggy-backed on `head-icons.hbs` instead of customizing `head.hbs`. See https://gitlab.com/antora/antora-ui-default/-/blob/master/src/partials/head.hbs for the default template. I could of course fork `head.hbs`, add a new `head-matomo.hbs` to the end to separate it more. But the `head-icons` script is already the last in <head> and already customized, so an easier choice. I can be persuaded otherwise though.

janhoy: So I did some more research, and turns out the [head-scripts.hbs](https://gitlab.com/antora/antora-ui-default/-/blob/master/src/partials/head-scripts.hbs) template only contained some google-analytics stuff that we don't want anyway, so I moved our matomo script there instead.

epugh (approved):
LGTM.    Curious to see what the patterns are!  Which pages get hit a lot, but also, if we can discern which ref guide queries lead to unhappy users.  indicating content gaps!

PR #3223, opened 2025-02-27
SOLR-17685: Remove script creation of solr url based on SOLR_TOOL_HOST in favour of java code in CLI tools

epugh:
https://issues.apache.org/jira/browse/SOLR-17685

# Description

Dueling places to figure out what the default solr url should be for CLI tools, that are unevenly implemented.

Linux: bin/solr auth,   Windows: bin/solr auth and bin/solr status

# Solution

Let `CLIUtils` create a default Solr url when none is passed in from environment variables, instead of doing it in `bin/solr` or `bin/solr.cmd`.

# Tests

bats


solr/core/src/java/org/apache/solr/cli/AuthTool.java
```diff
@@ -96,7 +98,7 @@ public class AuthTool extends ToolBase {
           .argName("FILE")
           .required()
           .desc(
-              "This is where any authentication related configuration files, if any, would be placed.")
+              "This is where any authentication related configuration files, if any, would be placed.  Defaults to $SOLR_HOME.")
```
malliaridis, line 101: If I am not mistaken, this defaults to `SOLR_HOME` only when `AuthTool` is used via the CLI script, but if directly used it does not default to any path. Is that correct?

I think we had a discussion before about using the classes without the CLI scripts, but I don't remember anymore if that was a valid scenario.

solr/core/src/java/org/apache/solr/cli/AuthTool.java
```diff
@@ -75,8 +75,10 @@ public class AuthTool extends ToolBase {
           .longOpt("solr-include-file")
           .hasArg()
           .argName("FILE")
+          .required()
```
malliaridis, line 78: Looking into our source code, this is actually required and always set via the CLI script, and if not set it would likely fail with a NPE I believe?

What I am wondering though is, should this be required? Does the file always exist and does it contain information we need? Or should we simply handle cases where it is not defined?

solr/bin/solr
```diff
@@ -612,22 +612,11 @@ if [[ "$SCRIPT_CMD" == "auth" ]]; then
...
-      done
-  fi
-
-  run_tool auth $@ --solr-url "$SOLR_URL_SCHEME://$SOLR_TOOL_HOST:${AUTH_PORT:-8983}" --auth-conf-dir "$SOLR_HOME" "--solr-include-file" "$SOLR_INCLUDE"
+  run_tool auth $@ --auth-conf-dir "$SOLR_HOME" "--solr-include-file" "$SOLR_INCLUDE"
   exit $?
 fi
 
-# at this point all tools that have a custom run process, like "status" and "auth" have been run and exited.
+# at this point the only tool that has a custom run process, "auth" have been run and exited.
```
malliaridis, line 619:
```suggestion
# at this point the only tool that has a custom run process, "auth" has been run and exited.
```

solr/bin/solr
```diff
@@ -612,22 +612,11 @@ if [[ "$SCRIPT_CMD" == "auth" ]]; then
...
-      do
-        port=$(jetty_port "$ID")
-        if [ "$port" != "" ]; then
-          AUTH_PORT=$port
-          break
-        fi
-      done
-  fi
-
-  run_tool auth $@ --solr-url "$SOLR_URL_SCHEME://$SOLR_TOOL_HOST:${AUTH_PORT:-8983}" --auth-conf-dir "$SOLR_HOME" "--solr-include-file" "$SOLR_INCLUDE"
```
malliaridis, line 626: This eliminates the `AUTH_PORT` and `SOLR_TOOL_HOST` environment variables, right? I believe it is fine, as the behavior is checking the jetty port and falls back to the default Solr port, so no need to distinguish it in the scripts.

Still, this is a breaking change.

malliaridis (changes requested):
From what I understand this eliminates redundant / duplicate environment variables, namely `SOLR_TOOL_HOST` and `AUTH_PORT`, which is fine to remove and replace with existing variables, but still introduces a breaking change in the CLI.

Some leftovers that would need cleanup in this case:

- [ ] We are still having references to `AUTH_PORT` in `bin/solr.cmd`
- [ ] We are still having references to `SOLR_TOOL_HOST` in `bin/solr.cmd`, `bin/solr` and `solr.in.*`

HoustonPutman:
So instead of using SOLR_TOOL_HOST, and instead using SOLR_HOST in the java code, I think we can remove the SOLR_TOOL_HOST all together, since we set it in `bin/solr`:

```bash
SOLR_TOOL_HOST="${SOLR_HOST:-localhost}"
export SOLR_TOOL_HOST
```

So since we are already defaulting to `localhost` in the code, I think we can remove `SOLR_TOOL_HOST` altogether and change `String host = EnvUtils.getProperty("solr.tool.host", "localhost");` to `String host = EnvUtils.getProperty("solr.host", "localhost");` in `CLIUtils`

But I agree that unfortunately the AUTH_PORT is a backwards incompatible change. Maybe instead we still check for the `auth` command in the 9.x code and do `export SOLR_PORT="${AUTH_PORT}"`. And we can keep the code to get that port the same. But in 10, we can remove it and the additional export.

solr/core/src/java/org/apache/solr/cli/AuthTool.java
epugh: I don't think using these outside the CLI scripts is actually a true scenario...   In fact, I think if folks are using these tools outside of the bin/solr scripts (like some folks have their own start scripts) it points to weaknesses in bin/solr that we need to improve!

solr/core/src/java/org/apache/solr/cli/AuthTool.java
epugh: At least right now, we have logic that if this file does NOT exist or is not writable, then the AuthTool quits with a warning.   So I think making it required is the right thing.   Having said that, longer term maybe something to think about?   Becasue it makes it implicitly assumed that the bin/solr auth command runs on your Solr server!

epugh:
> So instead of using SOLR_TOOL_HOST, and instead using SOLR_HOST in the java code, I think we can remove the SOLR_TOOL_HOST all together, since we set it in `bin/solr`:
> 
> ```shell
> SOLR_TOOL_HOST="${SOLR_HOST:-localhost}"
> export SOLR_TOOL_HOST
> ```
> 
> So since we are already defaulting to `localhost` in the code, I think we can remove `SOLR_TOOL_HOST` altogether and change `String host = EnvUtils.getProperty("solr.tool.host", "localhost");` to `String host = EnvUtils.getProperty("solr.host", "localhost");` in `CLIUtils`
> 
> But I agree that unfortunately the AUTH_PORT is a backwards incompatible change. Maybe instead we still check for the `auth` command in the 9.x code and do `export SOLR_PORT="${AUTH_PORT}"`. And we can keep the code to get that port the same. But in 10, we can remove it and the additional export.

Right...  So I can remove SOLR_HOST from the `bin/solr` and `bin/solr.cmd` because we default to localhost in the Java code!

HoustonPutman:
> Right... So I can remove SOLR_HOST from the `bin/solr` and `bin/solr.cmd` because we default to localhost in the Java code!

SOLR_TOOL_HOST, not SOLR_HOST 😅 but yes

epugh:
> > Right... So I can remove SOLR_HOST from the `bin/solr` and `bin/solr.cmd` because we default to localhost in the Java code!
> 
> SOLR_TOOL_HOST, not SOLR_HOST 😅 but yes

I was thinking we **COULD** remove SOLR_HOST because we look it up as teh default in the CLIUtils, however, I see that we use SOLR_HOST in lots of other places in `bin/solr` script...    If those were all in Java code, then we might not need it.

epugh:
> So instead of using SOLR_TOOL_HOST, and instead using SOLR_HOST in the java code, I think we can remove the SOLR_TOOL_HOST all together, since we set it in `bin/solr`:
> 
> ```shell
> SOLR_TOOL_HOST="${SOLR_HOST:-localhost}"
> export SOLR_TOOL_HOST
> ```
> 
> So since we are already defaulting to `localhost` in the code, I think we can remove `SOLR_TOOL_HOST` altogether and change `String host = EnvUtils.getProperty("solr.tool.host", "localhost");` to `String host = EnvUtils.getProperty("solr.host", "localhost");` in `CLIUtils`
> 
> But I agree that unfortunately the AUTH_PORT is a backwards incompatible change. Maybe instead we still check for the `auth` command in the 9.x code and do `export SOLR_PORT="${AUTH_PORT}"`. And we can keep the code to get that port the same. But in 10, we can remove it and the additional export.

Got a green on the scripts, so now looking at the AUTH_PORT.   So, I am thinking we on this PR do the check for AUTH_PORT and do what you suggested.   Commit and backport that.  And then in a NEW PR strip out the export SOLR_PORT="${AUTH_PORT}"` that only goes on main?

epugh:
Successful build!    If anyone wants to eyeball solr.cmd, that would be great.

malliaridis (approved):
LGTM now :)

PR #3224, opened 2025-02-27
SOLR-17686: Rename stageInEnglish to stageToString and make it protected and non-static

jeanielamwh:
https://issues.apache.org/jira/browse/SOLR-17686

# Description

# Solution

# Tests

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
@@ -726,7 +726,7 @@ public class SearchHandler extends RequestHandlerBase
     return true;
   }
 
-  private static String stageInEnglish(int nextStage) {
+  protected String stageInEnglish(int nextStage) {
     // This should probably be a enum, but that change should be its own ticket.
     switch (nextStage) {
       case STAGE_START:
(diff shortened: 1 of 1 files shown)


cpoerschke (approved):

jeanielamwh:
> ... might overriding be supported and yet could the base implementation use something like "CUSTOM_42" for a custom stage with value 42?

I like this idea. Otherwise, it seems a bit excessive to me that a user with custom stages must override this method that's used for logging only.

Pushed [be8e468](https://github.com/apache/solr/pull/3224/commits/be8e46872e6e3bddaa30a4824db399643fabadff) to replace the throw with returning `CUSTOM_STAGE_<NUM>`. What do you think?

dsmiley:
Way too advanced/obscure to waste people's time mentioning in CHANGES.txt IMO.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -726,7 +726,7 @@ private static boolean prepareComponents(
     return true;
   }
 
-  private static String stageInEnglish(int nextStage) {
+  protected String stageToString(int nextStage) {
```
cpoerschke, line 729: How about dropping the `next` prefix?
```suggestion
  protected String stageToString(int stage) {
```

dsmiley (approved):

PR #3225, opened 2025-02-27
Make test connectionLoss logic the same, add waitForLoss option

HoustonPutman:
Basically some of these tests require that the client be disconnected after we run `expireSession`. However, there can be gaps of a few milliseconds before the ZK client processes the event that it has been disconnected.

So just make all tests use the same logic and add an option to wait till the client has accepted the connection loss.


dsmiley (commented):
Can/should we just always wait or recommend that we do so?  Seems safer to wait.

HoustonPutman:
> Can/should we just always wait or recommend that we do so? Seems safer to wait.

Yeah, that should be fine. I'll remove the option, and always wait.

PR #3235, opened 2025-03-04
SimpleOrderedMap MapWriter constructor

dsmiley:
* new SimpleOrderedMap(MapWriter)   ex: a SolrParams
* deprecate SolrParams.toNamedList
* optimize SolrParams.writeMap

Bigger picture is avoiding NamedList.

Can go to 9x

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/common/params/SolrParams.java
+++ b/solr/solrj/src/java/org/apache/solr/common/params/SolrParams.java
@@ -76,6 +76,7 @@ public abstract class SolrParams
         assert value[0] != null;
         ew.put(entry.getKey(), value[0]);
       } else if (value.length > 1) {
+        // values shouldn't be null; not bothering to assert it
         ew.put(entry.getKey(), value);
       }
     }
--- a/solr/solrj/src/java/org/apache/solr/common/params/SolrParams.java
+++ b/solr/solrj/src/java/org/apache/solr/common/params/SolrParams.java
@@ -69,18 +69,16 @@ public abstract class SolrParams
 
   @Override
   public void writeMap(EntryWriter ew) throws IOException {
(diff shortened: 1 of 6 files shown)


dsmiley:
CC @renatoh for review (if you have time)

Note there are some toNamedList calls that I'm leaving be because they are used in places serializing a response.  This isn't the PR to make any change to our responses; I want a PR here that can go to 9x.

solr/solrj/src/java/org/apache/solr/common/params/SolrParams.java
```diff
@@ -69,18 +69,16 @@ public String get(String param, String def) {
...
-                ew.put(k, v);
-              } catch (IOException e) {
-                throw new RuntimeException("Error serializing", e);
-              }
-            });
+    for (Entry<String, String[]> entry : this) {
+      String[] value = entry.getValue();
+      // if only one value, don't wrap in an array
+      if (value.length == 1) {
+        assert value[0] != null;
```
renatoh, line 76: this 'assert' looks a bit out of place to me.  if value.lenght > 1, we do not care about null-values in the array, but if we have only one we do check for null. Also wouldn't it be better to just ignore the null value ?

dsmiley: As an `assert`, it communicates that it shouldn't happen and it somewhat enforces it (within tests, any way).  I could further add a comment to say "values should not be null" (above both conditions).  On the single-value side, this is just a cheap line.  Perhaps 90% of parameters have a single value.  I don't want to bother with an assert of every value in the list; the brevity & simplicity of that assert I kept is all I wanted.

dsmiley:
Planning to merge Monday, backporting to 9x.  Commit message will be similar to PR description, which is the first commit's message.

PR #3240, opened 2025-03-05
SOLR-17690 Make solr zk tools read ZK_HOST environment

janhoy:
https://issues.apache.org/jira/browse/SOLR-17690

Fix for https://github.com/apache/solr-operator/issues/762

### DIFF
--- a/solr/core/src/java/org/apache/solr/cli/CLIUtils.java
+++ b/solr/core/src/java/org/apache/solr/cli/CLIUtils.java
@@ -226,6 +226,11 @@ public final class CLIUtils {
       return zkHost;
     }
 
+//    zkHost = EnvUtils.getProperty("zkHost");
+//    if (zkHost != null && !zkHost.isBlank()) {
+//      return zkHost;
+//    }
+
     try (SolrClient solrClient = getSolrClient(cli)) {
       // hit Solr to get system info
       NamedList<Object> systemInfo =
--- a/solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
+++ b/solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
@@ -507,9 +507,10 @@ public class ZkSubcommandsTest extends SolrTestCaseJ4 {
     Path file =
         tmpDir.resolve("solrtest-getfile-" + this.getClass().getName() + "-" + System.nanoTime());
 
+    // Not setting --zk-host, will fall back to sysProp 'zkHost'
     String[] args =
         new String[] {
-          "cp", "-z", zkServer.getZkAddress(), "zk:" + getNode, file.toAbsolutePath().toString()
+          "cp", "zk:" + getNode, file.toAbsolutePath().toString()
         };
 
     ZkCpTool tool = new ZkCpTool();


solr/core/src/java/org/apache/solr/cli/CLIUtils.java
```diff
@@ -226,6 +226,11 @@ public static String getZkHost(CommandLine cli) throws Exception {
       return zkHost;
     }
 
+//    zkHost = EnvUtils.getProperty("zkHost");
+//    if (zkHost != null && !zkHost.isBlank()) {
+//      return zkHost;
+//    }
```
janhoy, line 232: This is the fix, commented out to prove failing test. The `CLIUtils#getZkHost(cli)` util is used by many tools. Not sure if all tools document the support for `ZK_HOST`, but the `zk cp` tool does. Reading property `zkHost` works since `ZK_HOST` env is explicitly put into this sysProp by EnvUtils.

solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
```diff
@@ -507,9 +507,10 @@ public void testGetFile() throws Exception {
     Path file =
         tmpDir.resolve("solrtest-getfile-" + this.getClass().getName() + "-" + System.nanoTime());
 
+    // Not setting --zk-host, will fall back to sysProp 'zkHost'
     String[] args =
         new String[] {
-          "cp", "-z", zkServer.getZkAddress(), "zk:" + getNode, file.toAbsolutePath().toString()
+          "cp", "zk:" + getNode, file.toAbsolutePath().toString()
```
janhoy, line 513: Simple way of testing the fix. This test explicitly sets `zkHost` sysProp in setUp, so removing `-z` shuold cause a fallback. The test `ZkSubcommandsTest#testGetFile()` will thus fail until the fix is added.

janhoy:
Refactored a bit, to be able to use the `--zk-host foo  -->  ZK_HOST=foo` fallback other places that attempt to read only the option.

Activated the fix since the Crave tests don't run anyway to show the failing test. I'll run all the tests locally.

janhoy:
Hmm, there's no `9.8.1` section in CHANGES.txt on `main`, and not on `branch_9x`, only on `branch_9_8`. 
Is the correct way for this bugfix to skip CHANGES here and on 9x, and then add one when cherry-picking to the release branch? Then the 9.8.1 RM will reconsile changes to other branches during release?

Here is suggested changes text:
```
* SOLR-17690: Make solr CLI tools read ZK_HOST environment as documented (janhoy)
```

HoustonPutman:
> Hmm, there's no `9.8.1` section in CHANGES.txt on `main`, and not on `branch_9x`, only on `branch_9_8`. Is the correct way for this bugfix to skip CHANGES here and on 9x, and then add one when cherry-picking to the release branch? Then the 9.8.1 RM will reconsile changes to other branches during release?
> 
> Here is suggested changes text:
> 
> ```
> * SOLR-17690: Make solr CLI tools read ZK_HOST environment as documented. There was a regression 
>   in 9.8.0 where this did not work, causing issues also for solr-operator v0.9 (janhoy)
> ```

So add it to the changeling for 9.9, and then in branch_9_8, add it to 9.8.1. As a part of the release, I remove the duplicates from 9.9 on branch_9x and main.

solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
```diff
@@ -305,7 +305,7 @@ protected void runExample(CommandLine cli, String exampleName) throws Exception
         "techproducts".equals(exampleName) ? "sample_techproducts_configs" : "_default";
 
     boolean isCloudMode = !cli.hasOption(USER_MANAGED_OPTION);
-    String zkHost = cli.getOptionValue(CommonCLIOptions.ZK_HOST_OPTION);
+    String zkHost = CLIUtils.getZkHostFromCliOrEnv(cli);
```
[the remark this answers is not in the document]
HoustonPutman: I'm not sure it's too risky, since it's unlikely that it's going to do anything that the user can't undo later.

HoustonPutman (commented):
There's lots of other usages of `cli.getOptionValue(CommonCLIOptions.ZK_HOST_OPTION)` so I might go through and see what else needs to be switched to this method.


Otherwise I like how it works.

solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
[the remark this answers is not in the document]
epugh: I can't reallyu imagine anyone running the example who wants it from ZK_HOST.   That would be a surprise.   You know, I think there is a legit argument that the RunExampleTool should NOT take in a ZKHost parameter.   We don't need the RunExampleTool to do everything that `bin/solr start` does..  We need it to be really good at running an example solr.   So, if we wanted to eliminate standalone support from the RunExampleTool, I would get that.  If we wanted it to only do certain other things, I would get that too.   One big annonycance in the `bin/solr` script is all the special handling we have for just running an example, and maybe if we narrow down what an example does, we would have less special case code?   

There are 14 options we pass to the run example!!!!!!   And honestly, the only option I ever pass is "techproducts" and "films" and "schemaless", every other option could potentially be hard coded.    Which would simpilify that tool a LOT.   It's an example tool, not a "let me debug solr in thwierd configuration" tool.

solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
```diff
@@ -507,10 +507,8 @@ public void testGetFile() throws Exception {
     Path file =
         tmpDir.resolve("solrtest-getfile-" + this.getClass().getName() + "-" + System.nanoTime());
 
-    String[] args =
-        new String[] {
-          "cp", "-z", zkServer.getZkAddress(), "zk:" + getNode, file.toAbsolutePath().toString()
-        };
+    // Not setting --zk-host, will fall back to sysProp 'zkHost'
+    String[] args = new String[] {"cp", "zk:" + getNode, file.toAbsolutePath().toString()};
```
epugh, line 511: :-(.  See me rant email ;-)

epugh (commented):
I posted an email to dev@ on why in general I think the use of environment variables in our scripts is a bad idea.   

So this review isn't about if this is a good or bad idea, but only about how we are actually supporting the ZK_HOST variable.

malliaridis (approved):
I'm fine with the changes. I feel that the users may be overwhelmed with the amount of options we give them for defining a property (CLI option, env vars and file), which makes our code more complex as well, as we have to cover each one individually and have a prioritization behind the scenes, covered in both Java classes and scripts. But that's another story.

Perhaps the alternative, "simplified" way to address this is to pass `EnvUtils.getProperty("zkHost")` in
`cli.getOptionValue(CommonCLIOptions.ZK_HOST_OPTION, EnvUtils.getProperty("zkHost")`, so no new methods are needed. But both ways should be fine if they adress the issue and the method is a bit more flexible in these scenarios. :)

I was also not aware that the operator would be affected by changes in the CLI, would be nice to have those covered as well during our refactoring. 😅

HoustonPutman:
So I think there is good discussion to be had here about the use of EnvVars in the SolrCLI logic (which currently we do use for default Solr URLs). But in the meantime, I do believe that there is a fix for https://github.com/apache/solr-operator/issues/762 that will support 9.8.0 and 9.8.1 even if this PR isn't included.

I think that's probably the way to go, so we can spend the time and come up with the best strategy for CLI EnvVars.

janhoy:
> I feel that the users may be overwhelmed with the amount of options we give them for defining a property (CLI option, env vars and file), which makes our code more complex as well

Adding `ZK_HOST=foo` to solr.in.sh has the same effect as adding `export ZK_HOST=foo` to your shell. It comes very handy when e.g. running Solr as a container and you can configure a lot with just env instead of having to patch the `CMD` part. The CLI tools like `zk cp` behaves the same as when starting solr itself.

So you can set `SOLR_URL` or `ZK_HOST` variable and then run much shorter commands to work with your cluster.

Perhaps we could do a better job in SolrCLI of wrapping such common behavior in more generic utils like

```java
```

solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
janhoy: Did not find your mail. I like env vars also in cli tools :)

solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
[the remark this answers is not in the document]
janhoy: Let's improve the examples in a followup issue. Perhaps we should not take away capabilities in 9.x, bu make this a 10.0 improvement. For now, this change here brings parity with how `--zk-host` option is documented and supported in the rest of the CLI.

solr/core/src/java/org/apache/solr/cli/CLIUtils.java
```diff
@@ -216,12 +217,32 @@ public static String normalizeSolrUrl(CommandLine cli) throws Exception {
...
+   * @return the value of the option or system property or the default value
+   */
+  public static String getCliOptionOrPropValue(
+      CommandLine cli, Option option, String sysprop, String defaultValue) {
+    String value = cli.getOptionValue(option);
+    if (value == null) {
+      value = EnvUtils.getProperty(sysprop, defaultValue);
+    }
+    return value;
+  }
```
janhoy, line 235: @epugh In my last commit I generalized the util method to this `getCliOptionOrPropValue` which can be used by any Tool that wishes to read a CLI OPTION with a fallback to env/prop and also a fallback to defaultvalue. Currently I only use it for zkhost. IMO the calling code also becomes more clear, as you will see in the same line both the OPT name and the prop name. Perhaps this could be used also for SOLR_URL?

solr/core/src/java/org/apache/solr/cli/CLIUtils.java
```diff
@@ -216,12 +217,32 @@ public static String normalizeSolrUrl(CommandLine cli) throws Exception {
...
+   *
+   * @param cli the command line
+   * @param option the commons cli {@link Option}
+   * @param sysprop the system property to fall back to
+   * @param defaultValue the default value. Use null if no default value is desired
+   * @return the value of the option or system property or the default value
+   */
+  public static String getCliOptionOrPropValue(
+      CommandLine cli, Option option, String sysprop, String defaultValue) {
+    String value = cli.getOptionValue(option);
```
janhoy, line 230: @epugh In my last commit I generalized the util method to this getCliOptionOrPropValue which can be used by any Tool that wishes to read a CLI OPTION with a fallback to env/prop and also a fallback to defaultvalue. Currently I only use it for zkhost. IMO the calling code also becomes more clear, as you will see in the same line both the OPT name and the prop name. Perhaps this could be used also for SOLR_URL?

epugh: I think doing the same for SOLR_URL makes sense.   It could be another PR, but it could also be in this PR!

janhoy:
This PR lost its urgency since a solr-operator change fixed the bug at hand. Perhaps the use of env.vars in CLI tools should be decided on the mailing list and then if we want to continue reading ENV in CLI tools we come back to this one?

HoustonPutman:
@janhoy I'm assuming it's ok to move this to 9.10 and not have it block 9.9?

epugh:
Do we still need this PR?

janhoy:
> Do we still need this PR?

I tried to follow up on the topic on dev@ list with subject "Environment Variables in the Solr CLI" on August 4th, as there was never a consensus for whether CLI tools should respect env.vars as they do today or whether they should require `--foo=bar` options. Let's land a decision on dev@ first, then I'll either merge or close this. I'm in favor of merging :)

acdha:
> Adding ZK_HOST=foo to solr.in.sh has the same effect as adding export ZK_HOST=foo to your shell. It comes very handy when e.g. running Solr as a container and you can configure a lot with just env instead of having to patch the CMD part. The CLI tools like zk cp behaves the same as when starting solr itself.

This is what I expected but it wasn't working with Solr 9.9, which is very confusing since the [documentation implies that works for all `solr` commands](https://solr.apache.org/guide/solr/latest/deployment-guide/zookeeper-ensemble.html#updating-solr-include-files).

```bash
$ podman run -it --rm --network bard -e SOLR_JAVA_MEM="-Xms2g" -e ZK_HOST=zk_1:2181 --entrypoint /bin/bash --user root solr:9
root@a973aed5e1d0:/opt/solr-9.9.0# echo $ZK_HOST
zk_1:2181
root@a973aed5e1d0:/opt/solr-9.9.0# solr zk ls /
Neither --zk-host or --solr-url parameters provided so assuming solr url is http://localhost:8983.

ERROR: Server refused connection at: http://localhost:8983/solr/admin/info/system

root@a973aed5e1d0:/opt/solr-9.9.0# solr zk ls -z $ZK_HOST /
WARN  - 2025-09-03 22:15:00.082; org.apache.solr.common.cloud.SolrZkClient; Using default ZkCredentialsInjector. ZkCredentialsInjector is not secure, it creates an empty list of credentials which leads to 'OPEN_ACL_UNSAFE' ACLs to Zookeeper nodes
WARN  - 2025-09-03 22:15:00.187; org.apache.solr.common.cloud.SolrZkClient; Using default ZkACLProvider. DefaultZkACLProvider is not secure, it creates 'OPEN_ACL_UNSAFE' ACLs to Zookeeper nodes
configs
overseer
aliases.json
live_nodes
collections
overseer_elect
security.json
node_roles
clusterprops.json
root@a973aed5e1d0:/opt/solr-9.9.0# echo "ZK_HOST=$ZK_HOST" >> $SOLR_INCLUDE
root@a973aed5e1d0:/opt/solr-9.9.0# solr zk ls /
Neither --zk-host or --solr-url parameters provided so assuming solr url is http://localhost:8983.

ERROR: Server refused connection at: http://localhost:8983/solr/admin/info/system
```

I can't speak to the other discussions but it would have removed some friction for me if setting ZK_HOST in our container definitions meant that all Solr-related tools would use it.

epugh:
I think I was the voice against magic variables to the bin/Solr commands.  If ZK_HOST is supported than should all the other params be too?   However, I think I am the minority voice on this. If we move forward on this, could we also just make sure Ref Guide covers this capability clearly?

solr/solr-ref-guide/modules/deployment-guide/pages/solr-control-script-reference.adoc
```diff
@@ -1144,6 +1141,17 @@ Usage: solr zk upconfig|downconfig -d <confdir> -n <configName> [-z zkHost] [-s
...
 Once ZooKeeper is initialized, Solr doesn't need to be running on any node to use these commands.
 
+=== ZooKeeper Connection String
+
+All `bin/solr zk` sub-commands require a ZooKeeper connection string to function. You can provide this in three ways, in order of precedence:
+
+1. Explicitly pass the `-z` or `--zk-host` option: `bin/solr zk <command> -z <zkHost>`
+2. Set the `ZK_HOST` environment variable: `export ZK_HOST=<zkHost>`
+3. Define `ZK_HOST` in `solr.in.sh` (or `solr.in.cmd` on Windows)
+4. If none are provided, the tools will default to `localhost:2181`
```
janhoy, line 1151: @epugh Added this sub-paragraph both to this file and to the `zookeeper-utilities` page, to explain the "hunting order". This is true for the zookeeper tools.

Some tools (ConfigTool, ExportTool, PackageTool, PostLogsTool, PostTool and StreamTool) use `normalizeSolrUrl()` function which have a slightly different logic, in that it inspects sysprops to find the local address of solr, and will then not blindly fall back to `localhost:9983` but to `$SOLR_PORT+1000`.

solr/solr-ref-guide/modules/deployment-guide/pages/solr-control-script-reference.adoc
```diff
@@ -1191,11 +1199,10 @@ An absolute path may be supplied instead.
 +
 [%autowidth,frame=none]
 |===
-s|Required |Default: none
+|Optional |Default: `localhost:2181` if `ZK_HOST` env var not set
 |===
```
janhoy, line 1203: These are not required per se, as if you have the env var or happen to run solr on default port, it is not required. So I switched to Optional in docs.

PR #3243, opened 2025-03-05
Simplify BinaryResponseWriter.getParsedResponse

dsmiley:
This method should lean on two other classes to handle the details.  Tests pass, and there are a couple testing this method including the particulars of "omit".


dsmiley:
Will merge Monday if no further feedback.  Pretty simple change.

epugh (approved):
I like the nicer comment!

PR #3249, opened 2025-03-09
SOLR-17607: Http ClusterStateProvider, lazy connect

dsmiley:
i.e. stop eagerly connecting when CloudSolrClient is created.

https://issues.apache.org/jira/browse/SOLR-17607

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
@@ -65,20 +65,19 @@ public abstract class BaseHttpClusterStateProvider implements ClusterStateProvid
   // the liveNodes and aliases cache will be invalidated after 5 secs
   private int cacheTimeout = EnvUtils.getPropertyAsInteger("solr.solrj.cache.timeout.sec", 5);
 
-  protected void init(List<String> solrUrls) throws Exception {
+  protected void initConfiguredNodes(List<String> solrUrls) throws Exception {
     this.configuredNodes =
         solrUrls.stream()
-            .map(
-                (solrUrl) -> {
-                  try {
-                    return new URI(solrUrl).toURL();
-                  } catch (MalformedURLException | URISyntaxException e) {
-                    throw new IllegalArgumentException(
-                        "Failed to parse base Solr URL " + solrUrl, e);
-                  }
-                })
+            .map(BaseHttpClusterStateProvider::stringToUrl)
             .collect(Collectors.toList());
-    this.urlScheme = this.configuredNodes.get(0).getProtocol();
+  }
+
+  private static URL stringToUrl(String solrUrl) {
+    try {
+      return new URI(solrUrl).toURL();
+    } catch (MalformedURLException | URISyntaxException e) {
+      throw new IllegalArgumentException("Failed to parse base Solr URL " + solrUrl, e);
+    }
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


dsmiley:
CC @mlbiscoc

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -78,28 +78,12 @@ public void init(List<String> solrUrls) throws Exception {
                   }
                 })
             .collect(Collectors.toList());
+    this.urlScheme = this.configuredNodes.get(0).getProtocol();
```
mlbiscoc, line 81: Looking at this now and just noticing, this seems a bit awkward. The `urlScheme` is set based on the first url in the list for http/https but before it was the last. But what if there is a mix of http/https? Not sure if it was intentional. It seems there shouldn't be a case where there is a mix of http/https urls in the configured list but depending on the ordering now, some urls may or may not work with `Utils.getBaseUrlForNodeName` when getting the live nodes list.

What do you think of throwing an exception if the urls are mixed with http/https and enforce a single protocol?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
```diff
@@ -230,43 +214,41 @@ private SimpleOrderedMap<?> submitClusterStateRequest(
...
-  public Set<String> getLiveNodes() {
+  public synchronized Set<String> getLiveNodes() {
+    // synchronized because there's no value in multiple doing this at the same time
     if (TimeUnit.SECONDS.convert((System.nanoTime() - liveNodesTimestamp), TimeUnit.NANOSECONDS)
-        > getCacheTimeout()) {
+        <= getCacheTimeout()) {
+      return this.liveNodes; // cached copy is fresh enough
+    }
 
+    if (liveNodes != null) { // thus we've fetched liveNodes previously
```
mlbiscoc, line 224: Doesn't adding this mean we never update the node list after the first `connect` using the existing liveNodes attribute? (assuming liveNodes populates successfully upon connection to Solr). Now it will only ever update the node list by fetching from configured nodes.

What if the only liveNode is a node not in the configured list and the cache timeout hit, then the node would never be fetched from because it isn't the in configured list.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
dsmiley: I agree that the URL scheme shouldn't vary across the URLs given; that should be erroneous.  An exception would be nice.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
dsmiley: I don't follow.  `liveNodes != null` is only false in the initial state.  assuming getLiveNodes is successful, liveNodes will never be null again.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
dsmiley: In my latest commit, I ended up populating urlScheme whenever we successfully fetch liveNodes.  I think it's more elegant and allows someone to specify URLs that only vary by scheme (http vs https) which actually sounds very useful.  Hope you like it.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
mlbiscoc: Nevermind ignore what I said. I misinterpreted this...

solr/solrj/src/java/org/apache/solr/client/solrj/impl/BaseHttpClusterStateProvider.java
mlbiscoc: Cool, yeah small change and I like this a lot better!

HoustonPutman:
I think this is breaking an integration test, namely: "export fails on non cloud mode" in `test_export.bats`.

It gives the following error on Jenkins:
https://ci-builds.apache.org/job/Solr/job/Solr-Smoketest-9.x/642/
```
not ok 36 export fails on non cloud mode # in 11026 ms
# (from function `assert_output' in file .gradle/node/node_modules/bats-assert/src/assert.bash, line 247,
#  in test file test/test_export.bats, line 69)
#   `assert_output --partial "ERROR: Couldn't initialize a HttpClusterStateProvider"' failed
# 
# -- output does not contain substring --
# substring (1 lines):
#   ERROR: Couldn't initialize a HttpClusterStateProvider
# output (36 lines):
#   Option '-url','--solr-collection-url': Deprecated for removal since 9.8: Use --solr-url and -c / --name instead
#   Option '-url','--solr-collection-url': Deprecated for removal since 9.8: Use --solr-url and -c / --name instead
# 
#   ERROR: Failed fetching live_nodes from [http://localhost:40197/solr]. If you think your Solr cluster is up and is accessible, you could try re-creating a new CloudSolrClient using working solr URLs.
```

https://ci-builds.apache.org/job/Solr/job/Solr-TestIntegration-main/123/
```
not ok 35 export fails on non cloud mode # in 11893 ms
# (from function `assert_output' in file .gradle/node/node_modules/bats-assert/src/assert.bash, line 247,
#  in test file test/test_export.bats, line 69)
#   `assert_output --partial "ERROR: Couldn't initialize a HttpClusterStateProvider"' failed
# 
# -- output does not contain substring --
# substring (1 lines):
#   ERROR: Couldn't initialize a HttpClusterStateProvider
# output (2 lines):
# 
#   ERROR: Failed fetching live_nodes from [http://localhost:37591/solr]. If you think your Solr cluster is up and is accessible, you could try re-creating a new CloudSolrClient using working solr URLs.
```

dsmiley:
ouch; I'll fix today.   Pierre told me too.  Sorry about this.  I really wish that I got an email that I failed the build -- a basic software engineering practice I use elsewhere for a long time.

PR #3251, opened 2025-03-10
Reimplement CancellableQueryTracker

reubent:
I don't have access to JIRA to create a ticket but please see below. I've posted in the users list but not received any replies to I've attempted to fix myself.

# Description

We noticed increasing memory usage over long periods of time in our Solr Clouds. Running some synthetic tests against such showed that it was directly linked to the number of abnormally terminated requests received - either exceeding resource limits or client timeout closed connection.

Investigation showed that the length of the response received to the tasks/list endpoints was growing. The resource leak was minimal but the time spent processing them was rising therefore a cleanup mechanism was needed.

Further investigation of why they were not being removed showed that when code ran out of time to run the calls to `checkLimitsBefore` would cause them to return before the finaliser that removes the active query from the list runs. This issue only appears to affect 
1. Explicitly set query IDs
2. Queries sent to shards which get the node name prepended to the generated query ID when adding to the list but not when being removed

# Solution

1. Prevent CancellableQueryTracker from filling with abnormally terminated requests by adding a timeout mechanism and making it configurable
3. Improve logging in that class
4. Add tests for that class
5. Add an additional release for activeQueries when the request has been finished early

Additionally there are clearly issues with the cancellation mechanism more generally that I'm not going to tackle now but am more than happy to look at in more depth later. Namely:

1. Multithreaded queries are not cancellable
2. Cancellability is dependent on the cancellation being sent to the same node that created the task to be cancelled. It also seems there's something weird with shard selection going on here too but I've not got to the bottom of it.
3. activeTaskList is again node dependent so the same UUID can be simultaneously sent to two nodes

# Tests

I've added tests for the class and run nearly half a million "broken" and half a million non broken requests against a cloud of four nodes 

# Checklist

- [n/a] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CancellableQueryTracker.java
+++ b/solr/core/src/java/org/apache/solr/core/CancellableQueryTracker.java
@@ -16,90 +16,177 @@
  */
 package org.apache.solr.core;
 
-import static org.apache.solr.common.params.CommonParams.QUERY_UUID;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import org.apache.solr.request.SolrQueryRequest;
+import org.apache.solr.search.CancellableCollector;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.lang.invoke.MethodHandles;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.UUID;
-import java.util.concurrent.ConcurrentHashMap;
-import org.apache.solr.request.SolrQueryRequest;
-import org.apache.solr.search.CancellableCollector;
+import java.util.concurrent.Executors;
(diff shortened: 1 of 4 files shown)


cpoerschke:
Hi @reubent!

> ... I don't have access to JIRA to create a ticket but please see below. ...

If you wish to access JIRA (to create a ticket, or to comment on existing tickets, now or in future) the https://selfserve.apache.org/jira-account.html portal _should_ allow you to register for access, if you wish.

Hope that helps.

cpoerschke:
> ... posted in the users list ...

Just to cross-reference, I think this is the thread in the archives: https://lists.apache.org/thread/h48d75krfxonthv8yx2mllfkd3vwvk6o

PR #3252, opened 2025-03-10
Update taking-solr-to-production.adoc environment variable SOLR_ENV to SOLR_INCLUDE

bkysela:
Solr 9.8 start script `bin/solr` reads `SOLR_INCLUDE` environment variable not `SOLR_ENV`.


epugh (approved):
LGTM, thanks for this!  I checked and agree with you...

PR #3253, opened 2025-03-11
DefaultPackageRepository: simplify HTTP & JSON

dsmiley:
The use of a SolrClient didn't seem to make sense if the web server isn't (necessarily) Solr.  Seemed to be getting more in the way.  There was some double or even triple(?) JSON back & forth parsing / serializing that I couldn't let be on principle.

Ran tests, including integration tests.  Didn't check if this actually uses GitHub.
Not sure if we lose some Solr-to-Solr auth that's relevant/important.  If so, I could switch to Jetty HttpClient but wouldn't love that.

### DIFF
--- a/solr/core/src/java/org/apache/solr/packagemanager/DefaultPackageRepository.java
+++ b/solr/core/src/java/org/apache/solr/packagemanager/DefaultPackageRepository.java
@@ -18,26 +18,21 @@
 package org.apache.solr.packagemanager;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.core.type.TypeReference;
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.net.URI;
 import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.Collection;
+import java.util.List;
 import java.util.Map;
-import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
-import org.apache.solr.client.solrj.SolrRequest;
-import org.apache.solr.client.solrj.SolrServerException;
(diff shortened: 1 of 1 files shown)


epugh:
There is a commented out .bats test in `test_packages.bats` that you could run.  It requires a live internet connection which is why we don't run it in CI that tests out github..  
> This test is useful if you are debugging/working with packages.

epugh (commented):
LGTM.   

A while back there was a discussion about simpler ways to access JSON etc without using Http2SolrClient and there was some reasons why folks said we *should* use it.   Maybe to make it easier to support auth credentials?    Don't remember.   There are some uses of GenericSolrRequest that could be used more directly.      

I believe we have some other places where we POST/PUT and consume JSON and maybe even some utility code, maybe we could get a clean sweep and use one approach everywhere?

solr/packaging/test/test_packages.bats
```diff
@@ -58,22 +58,24 @@ teardown() {
...
 
 # This test is useful if you are debugging/working with packages.
-# We have commented it out for now since it depends on a live internet
+# We have disabled it for now since it depends on a live internet
 # connection to run.  This could be updated with a local Repo server if we had
 # a package that is part of the Solr project to use.
-# @test "deploying and undeploying a cluster level package" {
-#  run solr start -Denable.packages=true
+@test "deploying and undeploying a cluster level package" {
+  skip "For developing package infra; requires a connection to github"
```
epugh, line 65: Nice!

dsmiley:
Thanks for pointing out the commented out test.  It passes now, after I updated it (the URL didn't seem to work exactly as it was).

As you can hopefully agree/observe here, Jackson is enough to fetch parse Json from a URL; Solr doesn't need its own utility.

The code that I encountered (replaced) created a new Http2SolrClient, and thus had no security/credential configuration.  To use Solr's configured auth, it'd have to get the underlying HttpClient from elsewhere in Solr's infrastructure.  But it wasn't and I'd rather just keep the simple thing of my PR until some day (if ever) someone wants to add auth.  Again; I didn't take away auth here.

PR #3154, opened 2025-01-31
SOLR-17379: Fix date parsing in Java 23, remove Lucene TestSecurityManager

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17379

PR #3258, opened 2025-03-11
SOLR-17651: Make sure CLI unit tests don't call System.exit()

psalagnac:
https://issues.apache.org/jira/browse/SOLR-17651

# Description

End goal is to make sure unit tests for CLI tools never make a call to `System.exit()`.

This introduces class `ToolRuntime` which is initiated and passed as a context for all CLI tools. When running tests, a sub-classes is used instead and blocks calls to `System.exit()`.

For now, I did only the minimal (large enough!) change to intercept calls to `System.exit()`. The newly introduced `ToolRuntime` may be leveraged later to remove some static calls, I'm mostly thinking about IOs which are pretty messy for now. We could improve testability by always using the runtime context for outputs.

# Tests

`BasicAuthIntegrationTest` was invoking `System.exit()`, which causes a very long timeout when running with Gradle. I figured out this test was doing what it should (will comment inline). I added `StatusToolTest` to keep same coverage.


solr/core/src/test/org/apache/solr/security/BasicAuthIntegrationTest.java
```diff
@@ -298,23 +295,10 @@ public void testBasicAuth() throws Exception {
...
-        assertTrue(obj.containsKey("version"));
-        assertTrue(obj.containsKey("startTime"));
-        assertTrue(obj.containsKey("uptime"));
-        assertTrue(obj.containsKey("memory"));
-      } catch (Exception e) {
-        log.error(
-            "StatusTool failed due to: {}; stdout from tool prior to failure: {}",
-            e,
-            baos.toString(StandardCharsets.UTF_8.name())); // nowarn
+      String[] toolArgs = new String[] {"status", "--solr-url", baseUrl};
```
psalagnac, line 298: Here, I removed credential parameters to force a failure of the tool.

My understanding is these parameters were wrongly added in #3154. Before that change, first execution of the tool was failing, and the test expects that.

epugh: Do we need both the positive "the tool suceeded" test and the "the tool failed" test?   I like simplifying our tests, so if the old test with all it's checks isn't needed, great..

solr/core/src/test/org/apache/solr/cli/CLITestHelper.java
```diff
@@ -0,0 +1,113 @@
...
+import java.io.PrintWriter;
+import java.io.Reader;
+import java.io.StringReader;
+import java.io.StringWriter;
+import org.apache.commons.cli.CommandLine;
+
+public class CLITestHelper {
+
+  /**
+   * Run a tool with all parameters, as specified in the command line. Fist parameter must be the
```
epugh, line 32:
```suggestion
   * Run a tool with all parameters, as specified in the command line. First parameter must be the
```

solr/core/src/test/org/apache/solr/cli/CLITestHelper.java
```diff
@@ -0,0 +1,113 @@
...
+   * @param args Complete command line.
+   * @param clazz Expected class name of tool implementation.
+   */
+  public static int runTool(String[] args, Class<?> clazz) throws Exception {
+    ToolRuntime runtime = new ValidatingRuntime();
+    return runTool(args, runtime, clazz);
+  }
+
+  /**
+   * Run a tool with all parameters, as specified in the command line. Fist parameter must be the
```
epugh, line 44:
```suggestion
   * Run a tool with all parameters, as specified in the command line. First parameter must be the
```

solr/core/src/test/org/apache/solr/cli/CLITestHelper.java
```diff
@@ -0,0 +1,113 @@
...
+   * tool name.
+   *
+   * @param args Complete command line.
+   * @param runtime The runtime implementation specified to the tool. In test, should be usually a
+   *     test specific implementation for additional checks.
```
epugh, line 49: would it make sense to have a @link for the next person to the specific class?

solr/core/src/test/org/apache/solr/cli/CLITestHelper.java
```diff
@@ -0,0 +1,113 @@
...
+    Tool tool = findTool(args, runtime);
+
+    // Check the tool actual class
+    assertEquals(clazz, tool.getClass());
+
+    CommandLine cli = parseCmdLine(tool, args);
+    return tool.runTool(cli);
+  }
+
+  public static class BufferingRuntime extends ValidatingRuntime {
```
epugh, line 62: is a BufferingRuntime on top of a ValidatingRuntitme needed?   `UnitTestingRuntime`?   `TestingRuntime`?

solr/core/src/test/org/apache/solr/cli/CreateToolTest.java
```diff
@@ -54,13 +50,6 @@ public void testCreateCollectionWithBasicAuth() throws Exception {
...
 
-    assertEquals(0, runTool(args));
-  }
-
-  private int runTool(String[] args) throws Exception {
-    Tool tool = findTool(args);
-    assertTrue(tool instanceof CreateTool);
-    CommandLine cli = parseCmdLine(tool, args);
-    return tool.runTool(cli);
+    assertEquals(0, CLITestHelper.runTool(args, CreateTool.class));
```
epugh, line 53: nicer!  we had a lot of boiler plate

solr/core/src/test/org/apache/solr/cli/PackageToolTest.java
```diff
@@ -94,7 +94,8 @@ private <T extends SolrRequest<? extends SolrResponse>> T withBasicAuth(T req) {
 
   @Test
   public void testPackageTool() throws Exception {
-    PackageTool tool = new PackageTool();
+    ToolRuntime runtime = new CLITestHelper.ValidatingRuntime();
```
epugh, line 97: why not also the `BufferingRuntime` and use that to check the outputs?

epugh: the way we have it, could just define and pass it into the constructor for PackageTool right?

solr/core/src/test/org/apache/solr/cli/SolrCLIZkToolsTest.java
epugh, line 1: i think i like better passing the command in as one of the args..  that had caught me before while reading the code...

solr/core/src/test/org/apache/solr/cli/StatusToolTest.java
```diff
@@ -0,0 +1,51 @@
...
+public class StatusToolTest extends SolrCloudTestCase {
+
+  @BeforeClass
+  public static void setupCluster() throws Exception {
+    configureCluster(2).configure();
+  }
+
+  /** Check the tool returns expected details by specifying Solr URL on the command line. */
+  @Test
+  public void testSolrUrlStatus() throws Exception {
```
epugh, line 35: and here is the answer to my previous comment, you did preserve the positive unit test!

solr/core/src/test/org/apache/solr/cli/TestSolrCLIRunExample.java
```diff
@@ -370,13 +362,12 @@ protected void testExample(String exampleName) throws Exception {
           };
 
       // capture tool output to stdout
-      ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
epugh, line 373: Not having these `baos` is nice...

solr/core/src/test/org/apache/solr/cli/CLITestHelper.java
epugh: Later on, I see how you reference is as part of `CLITestHelper.BufferingRuntime` and that makes it clear what it's for!

epugh (approved):
i think this all makes sense.   It's definitly adding some complexity to understanding how the CLI works, but we also have a lot nicer clean up of the test infrastructure that balances out the "What is this Runtime thing I need to know about"!

epugh:
On and I really like

> runtime context for outputs

madrob:
Relevant is also how ZooKeeper handles this - https://github.com/apache/zookeeper/blob/master/zookeeper-server/src/main/java/org/apache/zookeeper/util/ServiceUtils.java

psalagnac:
> Can we instead of having two implementations of the tests (for logging and system exit capture), have one that does both?

This would make all tests capture the output even when not necessary. Or do you mean having a single class that can enable/disable capturing output?

solr/core/src/test/org/apache/solr/cli/PackageToolTest.java
psalagnac: I've been using `BufferingRuntime` only for tests that actually check the output of the executed tool. This test does not. Maybe it is worth to later extend some tests to check output, this will be now easier.

solr/core/src/test/org/apache/solr/cli/CLITestHelper.java
```diff
@@ -0,0 +1,113 @@
...
+    public Reader getReader() {
+      return new StringReader(writer.toString());
+    }
+  }
+
+  /**
+   * Runtime for test with additional validations. Mostly, we ensure tests never call {@link
+   * System#exit(int)}.
+   */
+  public static class ValidatingRuntime extends DefaultToolRuntime {
```
HoustonPutman, line 101: Can we have `ValidatingRuntime` log the print(ln) messages, instead of using System.out.

Then the `BufferingRuntime` can call `super(message)`, so that the messages are still logged.

solr/core/src/test/org/apache/solr/cli/CLITestHelper.java
HoustonPutman: I agree that `ValidatingRuntime` should be renamed to `TestingRuntime`.

HoustonPutman (changes requested):

HoustonPutman:
This will also fix our failing Windows builds. Much easier to do this, than track down which test is failing there.

solr/core/src/java/org/apache/solr/cli/ToolRuntime.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.apache.solr.cli;
+
+/**
+ * An implementation of this class is specified when executing {@link ToolBase} to access
+ * environment specific methods (mostly to differentiate test from non-test executions for now).
+ *
+ * @see ToolBase
+ */
+public abstract class ToolRuntime {
```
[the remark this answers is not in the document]
psalagnac: This could be an interface right now, but I plan to add more methods so all outputs go there.
Currently, there are many static call to class `CLIO`, which makes tool impossible to test output of these tools.

solr/core/src/test/org/apache/solr/cli/CLITestHelper.java
psalagnac: Renamed so it is now a single class (as suggested in another comment) and all outputs go to a logger instead of stdout.

epugh: cool

epugh: Maybe a nit pick, but what do you think about "TestsRuntime"...  TestingRuntime makes me think it's involved in Testing of some sort of Runtime thing, versus it's teh Test version of the RuntimeTool.

HoustonPutman (approved):
I think this is good to go. Can we quickly add System.exit to the forbidden APIs (with the ToolRuntime usage as an exception)? After that we should merge.

solr/core/src/test/org/apache/solr/cli/CLITestHelper.java
HoustonPutman: I think TestingRuntime is fine, and preferable.

HoustonPutman (approved):

HoustonPutman:
Forgot about the package manager, even better! Thanks for the great work here

PR #3259, opened 2025-03-12
SOLR-17699: Fix frange perf regression when sub-clause

dsmiley:
Since 9.0 with "frange" queries embedded in other queries.  TwoPhaseIterator must be used; was iterating all docs instead.

https://issues.apache.org/jira/browse/SOLR-17699

No test but I'm certain this is the issue.


HoustonPutman:
Nice, quick find. Are there other places this needs to be updated or is this a special case?

dsmiley:
Looking across all `ConstantScoreScorer` using a DISI, -- they are all fine because they are based on a docSet/bitSet so it's super fast to iterate.  Only this one here wasn't.  I'm very familiar with the internals of this; I wish I caught the bug here in code review of a contributor.

HoustonPutman (approved):

dsmiley:
I'll merge Monday night EST.

PR #3260, opened 2025-03-12
SOLR-17700: Use core operation lock when unloading core

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17700

The tests seem to be happy with this change. `DeleteInactiveReplicaTest` failed before I added the non-locked private `unloadWithoutCoreOp()` method, so that's a good sign.

This should fix the issues we are seeing with the `SchemaDesignerApiTest` that #3213 was attempting to fix, in which there was a race condition between concurrent collection deletion and reload operations.


dsmiley (approved):
Very straight-forward; thanks

psalagnac (approved):

PR #3261, opened 2025-03-12
Allow disabling new UI module during builds

malliaridis:
# Description

The new UI module has shown some compatibilty challenges when it comes to building the project with less common operating systems like s390x or when running with alterantive JDKs (`RUNTIME_JAVA_HOME`).

# Solution

This PR adds a gradle property called `disableNewUiModule` that can be used to disable the new UI module during builds. 

This is only a termporary workaround until we properly fix the compatibility issues or decide to remove the new UI module.

# Tests

This PR can be tested by adding the following flag to the `gradle.properties`:

```properties
disableNewUiModule=true
```

# Checklist

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -67,9 +67,9 @@ include "solr:distribution"
 include "solr:docker"
 include "solr:prometheus-exporter"
 
-// Do not load the new UI module if it is disabled explicitly
-def withUiModule = !rootProject.hasProperty('disableUiModule') || rootProject.findProperty('disableUiModule') != 'true'
-if (withUiModule) { // Exclude new UI from unsupported OS builds
+def disableUiModuleValue = providers.gradleProperty('disableUiModule').orNull
+gradle.ext.withUiModule = disableUiModuleValue == null || disableUiModuleValue != 'true'
+if (gradle.ext.withUiModule) {
     include(":solr:ui")
 }
 
--- a/settings.gradle
+++ b/settings.gradle
@@ -68,8 +68,8 @@ include "solr:docker"
 include "solr:prometheus-exporter"
 
 // Do not load the new UI module if it is disabled explicitly
-def withNewUiModule = !rootProject.hasProperty('disableNewUiModule') || rootProject.findProperty('disableNewUiModule') != 'true'
-if (withNewUiModule) { // Exclude new UI from unsupported OS builds
+def withUiModule = !rootProject.hasProperty('disableUiModule') || rootProject.findProperty('disableUiModule') != 'true'
+if (withUiModule) { // Exclude new UI from unsupported OS builds
     include(":solr:ui")
 }
 
--- a/settings.gradle
+++ b/settings.gradle
@@ -57,7 +57,6 @@ include "solr:modules:s3-repository"
 include "solr:modules:scripting"
 include "solr:modules:sql"
 include "solr:webapp"
-include "solr:ui"
 include "solr:benchmark"
 include "solr:test-framework"
 include "solr:solr-ref-guide"
@@ -67,6 +66,12 @@ include "solr:packaging"
 include "solr:distribution"
 include "solr:docker"
 include "solr:prometheus-exporter"
+
+// Do not load the new UI module if it is disabled explicitly
+def withNewUiModule = !rootProject.hasProperty('disableNewUiModule') || rootProject.findProperty('disableNewUiModule') != 'true'
+if (withNewUiModule) { // Exclude new UI from unsupported OS builds
+    include(":solr:ui")
+}
 
 // Configures development for joint Lucene/ Solr composite build.
 apply from: file('gradle/lucene-dev/lucene-dev-repo-composite.gradle')
--- a/solr/webapp/build.gradle
+++ b/solr/webapp/build.gradle
@@ -94,24 +94,24 @@ task downloadBrowserify(type: NpmTask) {
   outputs.dir("${nodeProjectDir}/node_modules/browserify")
 }
 
-tasks.register("generateUiDevFiles") {
-  description = "Generate new UI for development and add files to outputs for later referencing."
-  group = "build"
-  onlyIf { rootProject.ext.withUiModule }
-
-  // Development files are larger in size but have shorter compile times.
-  dependsOn project(":solr:ui").tasks.wasmJsBrowserDevelopmentExecutableDistribution
-  outputs.dir(project(":solr:ui").file("build/dist/wasmJs/developmentExecutable/"))
+if (gradle.ext.withUiModule) {
+  tasks.register("generateUiDevFiles") {
+    description = "Generate new UI for development and add files to outputs for later referencing."
+    group = "build"
+    dependsOn project(":solr:ui").tasks.wasmJsBrowserDevelopmentExecutableDistribution
+    outputs.dir(project(":solr:ui").file("build/dist/wasmJs/developmentExecutable/"))
+  }
 }
 
-tasks.register("generateUiProdFiles") {
-  description = "Generate new UI for production and add files to outputs for later referencing"
-  group = "build"
-  onlyIf { rootProject.ext.withUiModule }
+if (gradle.ext.withUiModule) {
+  tasks.register("generateUiProdFiles") {
+    description = "Generate new UI for production and add files to outputs for later referencing"
+    group = "build"
 
-  // Production files are smaller in size but have longer compile times.
-  dependsOn project(":solr:ui").tasks.wasmJsBrowserDistribution
-  outputs.dir(project(":solr:ui").file("build/dist/wasmJs/productionExecutable/"))
+    // Production files are smaller in size but have longer compile times.
+    dependsOn project(":solr:ui").tasks.wasmJsBrowserDistribution
+    outputs.dir(project(":solr:ui").file("build/dist/wasmJs/productionExecutable/"))
+  }
 }
 
 task generateJsClientBundle(type: NpxTask) {
@@ -153,7 +153,7 @@ war {
 
   // Include new Admin UI files in webapp
   // By default, we always build dev that does not optimize generated files for release to reduce build times.
-  if (rootProject.ext.withUiModule) {
+  if (gradle.ext.withUiModule) {
     if (rootProject.ext.development) {
       from (tasks.generateUiDevFiles) {
         into "ui"
--- a/solr/webapp/build.gradle
+++ b/solr/webapp/build.gradle
@@ -97,7 +97,7 @@ task downloadBrowserify(type: NpmTask) {
 tasks.register("generateUiDevFiles") {
   description = "Generate new UI for development and add files to outputs for later referencing."
   group = "build"
-  onlyIf { rootProject.ext.withNewUiModule }
+  onlyIf { rootProject.ext.withUiModule }
 
   // Development files are larger in size but have shorter compile times.
   dependsOn project(":solr:ui").tasks.wasmJsBrowserDevelopmentExecutableDistribution
@@ -107,7 +107,7 @@ tasks.register("generateUiDevFiles") {
 tasks.register("generateUiProdFiles") {
   description = "Generate new UI for production and add files to outputs for later referencing"
   group = "build"
-  onlyIf { rootProject.ext.withNewUiModule }
+  onlyIf { rootProject.ext.withUiModule }
 
   // Production files are smaller in size but have longer compile times.
   dependsOn project(":solr:ui").tasks.wasmJsBrowserDistribution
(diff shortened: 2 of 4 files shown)


settings.gradle
```diff
@@ -68,5 +67,11 @@ include "solr:distribution"
 include "solr:docker"
 include "solr:prometheus-exporter"
 
+// Do not load the new UI module if it is disabled explicitly
+def withNewUiModule = !rootProject.hasProperty('disableNewUiModule') || rootProject.findProperty('disableNewUiModule') != 'true'
```
HoustonPutman, line 71: Wouldn't it be easier just to add a default value here?

HoustonPutman:
Could we also disable it by default in the OS/Arch values we know are unsupported, as well as the ALT JDK option?

settings.gradle
malliaridis: I tried, but it does not allow me to set `rootProject.ext.*`. I believe it is because it is not yet initialized, since we are in `settings.gradle` / the initialization phase.

solr/webapp/build.gradle
```diff
@@ -97,6 +97,7 @@ task downloadBrowserify(type: NpmTask) {
 tasks.register("generateUiDevFiles") {
   description = "Generate new UI for development and add files to outputs for later referencing."
   group = "build"
+  onlyIf { rootProject.ext.withNewUiModule }
```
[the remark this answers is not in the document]
malliaridis: Can do. :) Is there a better way to detect if the `:solr:ui` module exists? I am only aware of something like `rootProject.subprojects.any { it.name == ":solr:ui" }`.

malliaridis:
> Could we also disable it by default in the OS/Arch values we know are unsupported, as well as the ALT JDK option?

This can be done by checking the values set in `System.getProperty("os.name")` and `System.getenv("RUNTIME_JAVA_HOME")` (or is it also just a property?) I believe? The main challenge is to disable the module during the initialization phase, which has not loaded any gradle script like `alternative-jdk-support.gradle`.

Which operating systems are not supported besides s390x? If we get a more complex logic, I would extract it to a separate script and apply it like we do with `gradle/lucene-dev/lucene-dev-repo-composite.gradle`.

It also seems that it works well and does not complain about invalid lock state, which I was expecting at first when disabling the module.

uschindler:
> > Could we also disable it by default in the OS/Arch values we know are unsupported, as well as the ALT JDK option?
> 
> This can be done by checking the values set in `System.getProperty("os.name")` and `System.getenv("RUNTIME_JAVA_HOME")` (or is it also just a property?) I believe? The main challenge is to disable the module during the initialization phase, which has not loaded any gradle script like `alternative-jdk-support.gradle`.

It is not only the env var. You can also set an alternative JDK with a sysprop or project property. I am not sure if you can check all this duing early setup.

Basically I don't like this at all. Can we not just move away from Kotlin? I am not happy that we have a mixed build. Either all in Kotlin or nothing.

> Which operating systems are not supported besides s390x? If we get a more complex logic, I would extract it to a separate script and apply it like we do with `gradle/lucene-dev/lucene-dev-repo-composite.gradle`.
> 
> It also seems that it works well and does not complain about invalid lock state, which I was expecting at first when disabling the module.

Uwe

dsmiley:
We're sticking with Kotlin for the UI module, and Java for everything else.

malliaridis:
> Can we not just move away from Kotlin? I am not happy that we have a mixed build. Either all in Kotlin or nothing.

When it comes to the build stuff, I think it would be a good idea to move from Groovy to Kotlin, step by step, as stated in [SOLR-17572](https://issues.apache.org/jira/browse/SOLR-17572). The current issues are as far as I know caused by the UI module that makes use of the Kotlin Multiplatform plugin.

I also do not like the idea of disabling a module that is in our project, and these changes are not solving the actual issue. :/

uschindler:
Hi,
ok makes sense. Generally I agree to move the Gradle build at some point to kotlin, but I don't see this as an important step. This should be done together with Lucene to again make use of each other. Generally the Lucene (and Solr build) was invented by @dweiss. I don't think that it is a simple replacement, because there is a lot of special thing especially to correctly compile MR-JARS, JavaModule-System builds (not correctly supported in Gradle at all, too much magic that breaks heavily).

I was not aware that the issues come from the kotlin compiler for the user code, sorry. If that's the only issue I think the compiler-config in the runtime-java-home support can possibly be adapted to work correct witzh alternative JDKs. The problem is maybe only that some information is not exposed by our hack that's needed by kotlin.

Uwe

uschindler:
P.S.: At moment the Policeman Jenkins Server is undergoing hardware upgrade (moves to new server at Hetzner data center). It will return back to life later this evening.

uschindler:
> P.S.: At moment the Policeman Jenkins Server is undergoing hardware upgrade (moves to new server at Hetzner data center). It will return back to life later this evening.

And it's back at work :-) While I was writing my latest mail the migration to it's final datacenter location with its old IPv4 adresses and IPv6 subnets was done.

Nayana-ibm:
We've tested these changes on the s390x architecture, and they are not working as expected.
Specifically, the current code:
`def withUiModule = !rootProject.hasProperty('disableUiModule') || rootProject.findProperty('disableUiModule') != 'true'`
always evaluates `withUiModule `as true, which means the UI module is never disabled.

To fix this, we made some changes to ensure the disableUiModule flag is properly recognized. Please see diff below: 
```
diff --git a/build.gradle b/build.gradle
index 24c2045..7901130 100644
--- a/build.gradle
+++ b/build.gradle
@@ -35,6 +35,10 @@ plugins {
   alias(libs.plugins.openapi.generator) apply false
 }

+ext {
+    withUiModule = gradle.ext.withUiModule  // Use the global setting
+}
+
 // Declare default Java versions for the entire project and for SolrJ separately
 rootProject.ext.minJavaVersionDefault = JavaVersion.toVersion(libs.versions.java.min.get())
 rootProject.ext.minJavaVersionSolrJ = JavaVersion.toVersion(libs.versions.java.solrj.get())
@@ -114,7 +118,6 @@ ext {
     logger.debug('Lucene base version (without suffix): {}', luceneBaseVersion)
     return luceneBaseVersion
   }
-
   // For testing code
   vectorIncubatorJavaVersions = [ JavaVersion.VERSION_21, JavaVersion.VERSION_22, JavaVersion.VERSION_23, JavaVersion.VERSION_24, JavaVersion.VERSION_25 ] as Set
 }
diff --git a/settings.gradle b/settings.gradle
index 14b9169..4a16234 100644
--- a/settings.gradle
+++ b/settings.gradle
@@ -57,7 +57,6 @@ include "solr:modules:s3-repository"
 include "solr:modules:scripting"
 include "solr:modules:sql"
 include "solr:webapp"
-include "solr:ui"
 include "solr:benchmark"
 include "solr:test-framework"
 include "solr:solr-ref-guide"
@@ -68,5 +67,11 @@ include "solr:distribution"
 include "solr:docker"
 include "solr:prometheus-exporter"

+def disableUiModuleValue = providers.gradleProperty('disableUiModule').orNull
+gradle.ext.withUiModule = disableUiModuleValue == null || disableUiModuleValue != 'true'
+if (gradle.ext.withUiModule) {
+    include(":solr:ui")
+}
+
 // Configures development for joint Lucene/ Solr composite build.
 apply from: file('gradle/lucene-dev/lucene-dev-repo-composite.gradle')
diff --git a/solr/webapp/build.gradle b/solr/webapp/build.gradle
index bd701b9..3ea3607 100644
--- a/solr/webapp/build.gradle
+++ b/solr/webapp/build.gradle
@@ -94,22 +94,24 @@ task downloadBrowserify(type: NpmTask) {
   outputs.dir("${nodeProjectDir}/node_modules/browserify")
 }

-tasks.register("generateUiDevFiles") {
-  description = "Generate new UI for development and add files to outputs for later referencing."
-  group = "build"
-
-  // Development files are larger in size but have shorter compile times.
-  dependsOn project(":solr:ui").tasks.wasmJsBrowserDevelopmentExecutableDistribution
-  outputs.dir(project(":solr:ui").file("build/dist/wasmJs/developmentExecutable/"))
+if (gradle.ext.withUiModule) {
+  tasks.register("generateUiDevFiles") {
+    description = "Generate new UI for development and add files to outputs for later referencing."
+    group = "build"
+    dependsOn project(":solr:ui").tasks.wasmJsBrowserDevelopmentExecutableDistribution
+    outputs.dir(project(":solr:ui").file("build/dist/wasmJs/developmentExecutable/"))
+  }
 }

-tasks.register("generateUiProdFiles") {
-  description = "Generate new UI for production and add files to outputs for later referencing"
-  group = "build"
+if (gradle.ext.withUiModule) {
+  tasks.register("generateUiProdFiles") {
+    description = "Generate new UI for production and add files to outputs for later referencing"
+    group = "build"

-  // Production files are smaller in size but have longer compile times.
-  dependsOn project(":solr:ui").tasks.wasmJsBrowserDistribution
-  outputs.dir(project(":solr:ui").file("build/dist/wasmJs/productionExecutable/"))
+    // Production files are smaller in size but have longer compile times.
+    dependsOn project(":solr:ui").tasks.wasmJsBrowserDistribution
+    outputs.dir(project(":solr:ui").file("build/dist/wasmJs/productionExecutable/"))
+  }
 }

 task generateJsClientBundle(type: NpxTask) {
@@ -151,13 +153,15 @@ war {

   // Include new Admin UI files in webapp
   // By default, we always build dev that does not optimize generated files for release to reduce build times.
-  if (rootProject.ext.development) {
-    from (tasks.generateUiDevFiles) {
-      into "ui"
-    }
-  } else {
-    from (tasks.generateUiProdFiles) {
-      into "ui"
+  if (gradle.ext.withUiModule) {
+    if (rootProject.ext.development) {
+      from (tasks.generateUiDevFiles) {
+        into "ui"
+      }
+    } else {
+      from (tasks.generateUiProdFiles) {
+        into "ui"
+      }
     }
   }
 }
```
Now, if you want to disable the UI module, you can simply run:
`./gradlew -PdisableUiModule=true` 
This update makes sure the build process correctly includes or excludes the UI module based on your settings.

dsmiley:
@Nayana-ibm do you know enough Gradle to know why @malliaridis 's first attempt didn't work?  I'm always trying to learn more Gradle.

Nayana-ibm:
> @Nayana-ibm do you know enough Gradle to know why @malliaridis 's first attempt didn't work? I'm always trying to learn more Gradle.

`def withUiModule = !rootProject.hasProperty('disableUiModule') || rootProject.findProperty('disableUiModule') != 'true' `
The condition checks whether the disableUiModule property exists in the root project. However, since this property is not present, the check !rootProject.hasProperty('disableUiModule') always returns true. As a result, withUiModule always evaluates to true, no matter what value we try to pass.  We have added  `withUiModule `to ext to have access throughout the build.

Nayana-ibm:
@malliaridis Can you please check conflicts.

malliaridis:
Sorry for the late response. I will try to apply your changes today @Nayana-ibm and merge it once the checks pass. I didn't have the time to continue working on it and test it myself.

malliaridis:
@Nayana-ibm I’ve removed the duplicated `ext` entry in `/build.gradle` and added you as a co-author. I hope that’s okay with you. :)

Could you check if the change I made on your patch didn't break anything and the solution still works on s390x machines?

uschindler:
How can I execute Gradle to disable this on command line. Is `gradlew -PdisableUiModule=true` the way to go? Because this is how I interpret this PR.

That's how I would setup Jenkins like we do for all other test props.

Or do I need to modify `gradle.properties`? Really? What a bullshit, sorry! 🤪

Nayana-ibm:
> @Nayana-ibm I’ve removed the duplicated `ext` entry in `/build.gradle` and added you as a co-author. I hope that’s okay with you. :)
> 
> Could you check if the change I made on your patch didn't break anything and the solution still works on s390x machines?
changes looks good to me

malliaridis:
I'll move forward and merge this PR. Note that there is [a discussion about maintaining the UI code in a separate repository](https://lists.apache.org/thread/g7fkrpkf3mnsnjp0gyj8wzpo6v14v6nl) under the Apache Solr suite, similar to the [solr-operator](https://github.com/apache/solr-operator). But it seems that we will continue maintaining it in here, and that we may make necessary changes to simplify and improve our gradle setup to avoid these complex constraints between modules and configurations.

Nayana-ibm:
@uschindler Can you please add flag -PdisableUiModule to s390x CI Job
https://issues.apache.org/jira/browse/SOLR-17766

uschindler:
Will do in a moment.

uschindler:
This is done. Sorry for delay, was on travel and was tooo busy.

PR #3263, opened 2025-03-13
SOLR-16903: Migrate all tests using File to NIO Path

mlbiscoc:

# Description

WIP - Leaving in draft state for now as there is still more work from refactor needed but most tests pass so can be semi-reviewed.

Migrate all imports using `File` to `Path`. There should be no more imports using `File` in the project. We are close to making `File` a forbidden API after this but still more work needed.

# Checklist


solr/core/src/java/org/apache/solr/blockcache/BlockDirectory.java
```diff
@@ -289,12 +290,12 @@ String getFileCacheName(String name) throws IOException {
 
   private long getFileModified(String name) throws IOException {
     if (in instanceof FSDirectory) {
-      File directory = ((FSDirectory) in).getDirectory().toFile();
-      File file = new File(directory, name);
-      if (!file.exists()) {
+      Path directory = ((FSDirectory) in).getDirectory();
+      Path file = directory.resolve(name);
```
epugh, line 294: Maybe just purely cosmentic, but should this really be `path` instead of `file`?  In all the places?

solr/core/src/test/org/apache/solr/cli/SolrProcessManagerTest.java
```diff
@@ -84,9 +83,11 @@ private static int findAvailablePort() throws IOException {
   private static Pair<Integer, Process> createProcess(int port, boolean https) throws IOException {
     // Get the path to the java executable from the current JVM
     String classPath =
-        Arrays.stream(System.getProperty("java.class.path").split(File.pathSeparator))
+        Arrays.stream(
+                System.getProperty("java.class.path").split(System.getProperty("path.separator")))
```
epugh, line 87: I'm suprised we don't have sort of nicer way of doing this than `System.getProperty`..  Is this common enough that we should have a helper?

solr/core/src/test/org/apache/solr/cli/StreamToolTest.java
```diff
@@ -185,11 +183,11 @@ public void testReadStream() throws Exception {
...
   @SuppressWarnings({"unchecked", "rawtypes"})
   public void testLocalCatStream() throws Exception {
-    File localFile = File.createTempFile("topLevel1", ".txt");
-    populateFileWithData(localFile.toPath());
+    Path localFile = Files.createTempFile("topLevel1", ".txt");
+    populateFileWithData(localFile);
 
     StreamTool.LocalCatStream catStream =
-        new StreamTool.LocalCatStream(localFile.getAbsolutePath(), -1);
+        new StreamTool.LocalCatStream(localFile.toAbsolutePath().toString(), -1);
```
epugh, line 190: Should we just pass in a path here?

solr/core/src/test/org/apache/solr/cli/SolrProcessManagerTest.java
epugh: Is "FileSystems.getDefault().getSeparator();" a possiblity?

solr/core/src/test/org/apache/solr/cli/TestSolrCLIRunExample.java
```diff
@@ -558,32 +555,34 @@ public void testInteractiveSolrCloudExample() throws Exception {
...
+      throw new IOException(
+          "Should have been able to create file '" + toExecute.toAbsolutePath() + "' ");
+    }
 
     String[] toolArgs =
         new String[] {
           "-e", "techproducts",
-          "--server-dir", solrServerDir.getAbsolutePath(),
-          "--example-dir", solrExampleDir.getAbsolutePath(),
+          "--server-dir", solrServerDir.toAbsolutePath().toString(),
```
epugh, line 582: Shame that this requires all these `.toString`..   Is there a smarter way of handling this toolArgs?   `var toolArgs`?  And then internally we call .toString on anything passed in?????

epugh (commented):
Wow, so much work!  In general I love the removing of the indirection that using Path is showing....

solr/core/src/java/org/apache/solr/blockcache/BlockDirectory.java
dsmiley: No.  I like 'file' to differentiate from a 'dir'.  'path' is generic if could be either.

solr/core/src/test/org/apache/solr/cli/SolrCLIZkToolsTest.java
```diff
@@ -279,12 +292,16 @@ public void testCp() throws Exception {
     // src and cp3 are valid
 
     String localNamed =
-        tmp.normalize() + File.separator + "localnamed" + File.separator + "renamed.txt";
+        tmp.normalize()
```
dsmiley, line 295: shouldn't we use resolve() instead?

solr/core/src/test/org/apache/solr/cli/SolrCLIZkToolsTest.java
```diff
@@ -265,7 +275,10 @@ public void testCp() throws Exception {
         new String[] {
           "--zk-host",
           zkAddr,
-          "file:" + srcPathCheck.normalize().toAbsolutePath() + File.separator + "solrconfig.xml",
+          "file:"
+              + srcPathCheck.normalize().toAbsolutePath()
+              + FileSystems.getDefault().getSeparator()
```
dsmiley, line 280: shouldn't we use resolve instead?  And everywhere else in this source file probably.

solr/core/src/test/org/apache/solr/cli/SolrCLIZkToolsTest.java
```diff
@@ -417,7 +434,7 @@ public void testCp() throws Exception {
     res = cpTool.runTool(SolrCLI.processCommandLineArgs(cpTool, args));
     assertEquals("Copy should have succeeded.", 0, res);
 
-    Path locEmpty = Paths.get(tmp2.toAbsolutePath().toString(), "stopwords", "emptyfile");
+    Path locEmpty = Path.of(tmp2.toAbsolutePath().toString(), "stopwords", "emptyfile");
```
dsmiley, line 437: Use resolve here instead?  and in many other places similarly

dsmiley (changes requested):
I reviewed a subset of this but I'll pause until you react to this feedback.(1) Don't use Path.of if you can instead resolve() against the existing path.
 (2) Don't rely on referring to the separator char for string concatenation if the scenario could simply be path.resolve()

solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
```diff
@@ -353,13 +376,21 @@ public void testUpConfigLinkConfigClearZk() throws Exception {
     assertEquals(0, runTool(args, configSetUploadTool));
 
     assertTrue(zkClient.exists(ZkConfigSetService.CONFIGS_ZKNODE + "/" + confsetname, true));
+    final Path confDir = Path.of(ExternalPaths.TECHPRODUCTS_CONFIGSET);
```
dsmiley, line 379: `ExternalPaths.TECHPRODUCTS_CONFIGSET` should be a Path

solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
```diff
@@ -383,7 +414,8 @@ public void testUpConfigLinkConfigClearZk() throws Exception {
 
     // test down config
     Path configSetDir =
-        tmpDir.resolve(
+        Path.of(
+            tmpDir.toString(),
```
dsmiley, line 418: why not as it was, resolve()?

solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
```diff
@@ -403,30 +435,24 @@ public void testUpConfigLinkConfigClearZk() throws Exception {
...
     Path confSetDir = configSetDir.resolve("conf");
     try (Stream<Path> filesStream = Files.list(confSetDir)) {
+      List<Path> files = filesStream.toList();
       zkFiles =
           zkClient.getChildren(ZkConfigSetService.CONFIGS_ZKNODE + "/" + confsetname, null, true);
-      long fileCount = filesStream.count();
       assertEquals(
-          "Comparing original conf files that were to be uploaded to what is in ZK",
-          fileCount,
+          "Comparing original conf files that were to be uploadedto what is in ZK",
```
dsmiley, line 442:
```suggestion
          "Comparing original conf files that were to be uploaded to what is in ZK",
```

solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
```diff
@@ -403,30 +435,24 @@ public void testUpConfigLinkConfigClearZk() throws Exception {
...
-              .filter(
-                  (path -> {
-                    if (!Files.isDirectory(path)) {
-                      return false;
-                    }
-                    return !path.getFileName().toString().startsWith(".");
-                  }))
+          stream
+              .filter(Files::isRegularFile)
+              .filter(path -> path.getFileName().toString().matches("[^\\.].*"))
```
dsmiley, line 454: what was wrong with startsWith?

solr/core/src/test/org/apache/solr/cli/SolrCLIZkToolsTest.java
mlbiscoc: Tried to remove as much as `FileSystems.getDefault().getSeparator()` around the PR as I could.

solr/core/src/test/org/apache/solr/cli/SolrCLIZkToolsTest.java
mlbiscoc: Yeah I try to move to resolve as much as possible, but I might have some mistakes like the one you found here. Tried my best to clean up some of it.

mlbiscoc (commented):
Did my own passthrough of this PR and tried to fix where I made mistakes. Might need to do another passthrough as I found other places we can improve on to just make variables `Path` instead of `String`

epugh:
Looks like some nice progress..  I just clicked the unit tests button.

solr/modules/sql/src/test/org/apache/solr/handler/sql/TestSQLHandlerNonCloud.java
```diff
@@ -32,17 +32,17 @@
...
     return workDir;
   }
 
   @BeforeClass
   public static void beforeClass() throws Exception {
-    File solrHome = createSolrHome();
-    solrHome.deleteOnExit();
-    createAndStartJetty(solrHome.getAbsolutePath());
+    Path solrHome = createSolrHome();
+    solrHome.toFile().deleteOnExit();
```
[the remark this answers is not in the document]
dsmiley: This is a test, thus any file created is in a temp directory and will be deleted _anyway_, as that's what the test infrastructure takes care to do.

solr/core/src/test/org/apache/solr/cli/StreamToolTest.java
dsmiley: agreed to leave CatStream be.  I *could* see adding a Path convenience constructor there but it's out of scope of this PR IMO.

solr/core/src/test/org/apache/solr/cli/SolrProcessManagerTest.java
dsmiley: I'm good with your replacement, albeit it'd be slightly nicer to declare a variable to hold the value as you're going to refer to it multiple times.  NIO has no alternative.  It's easier to just ban File and use this more awkward lookup because it's so rare to actually use this.

solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
```diff
@@ -398,66 +411,82 @@ public void testUpConfigLinkConfigClearZk() throws Exception {
...
+              .toList();
+      files.forEach(
+          (sourceFile) -> {
+            int indexOfRelativePath =
+                sourceFile
+                    .toAbsolutePath()
+                    .toString()
+                    .lastIndexOf(
+                        "sample_techproducts_configs"
+                            + FileSystems.getDefault().getSeparator()
```
dsmiley, line 461: Wherever we see `FileSystems.getDefault()` yet a Path var is in scope (here it's `sourceFile`), we should use `pathVar.getFileSystem()` instead.

solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
```diff
@@ -490,7 +523,7 @@ public void testGet() throws Exception {
 
     final String standardOutput2 = byteStream2.toString(StandardCharsets.UTF_8);
     assertTrue(standardOutput2.startsWith("Copying from 'zk:/getNode'"));
-    byte[] fileBytes = Files.readAllBytes(Paths.get(localFile.getAbsolutePath()));
+    byte[] fileBytes = Files.readAllBytes(localFile.toAbsolutePath());
```
dsmiley, line 526: why call toAbsolutePath?  there's no point; we're going to read it not show it or pass it off somewhere.  I observe many developers love calling getAbsolutePath but I don't get the point.

solr/core/src/test/org/apache/solr/cli/SolrCLIZkToolsTest.java
```diff
@@ -131,12 +134,11 @@ public void testDownconfig() throws Exception {
     ConfigSetDownloadTool downTool = new ConfigSetDownloadTool();
     int res = downTool.runTool(SolrCLI.processCommandLineArgs(downTool, args));
     assertEquals("Download should have succeeded.", 0, res);
-    verifyZkLocalPathsMatch(
-        Paths.get(tmp.toAbsolutePath().toString(), "conf"), "/configs/downconfig1");
+    verifyZkLocalPathsMatch(tmp.toAbsolutePath().resolve("conf"), "/configs/downconfig1");
```
dsmiley, line 138: temp directories are always absolute.  Boy developers *love* calling toAbsolutePath for no reason :laugh-cry:

dsmiley (commented):
In general, try to also remove frivilous use of "getAbsolutePath" when it will accomplish nothing, or when it's already reasonably absolute (a temp dir).  I can understand if you say that's out of scope.

Some usages of FileSystem.getDefault() should instead be to a path that's in-scope to getFileSystem()

solr/core/src/test/org/apache/solr/cloud/TestTolerantUpdateProcessorCloud.java
```diff
@@ -95,10 +95,9 @@ public class TestTolerantUpdateProcessorCloud extends SolrCloudTestCase {
   public static void createMiniSolrCloudCluster() throws Exception {
 
     final String configName = "solrCloudCollectionConfig";
-    final File configDir =
-        new File(TEST_HOME() + File.separator + "collection1" + File.separator + "conf");
+    final Path configDir = TEST_HOME().resolve("collection1").resolve("conf");
```
dsmiley, line 98: LOVE THIS.  It exemplifies the improvements in this PR, and the sad state of half/partial usage of NIO before that so needed some love.

solr/core/src/test/org/apache/solr/schema/TestSchemalessBufferedUpdates.java
```diff
@@ -57,19 +59,17 @@ public class TestSchemalessBufferedUpdates extends SolrTestCaseJ4 {
...
-    File tmpConfDir = new File(tmpSolrHome, confDir);
-    File testHomeConfDir = new File(TEST_HOME(), confDir);
-    FileUtils.copyFileToDirectory(
-        new File(testHomeConfDir, "solrconfig-schemaless.xml"), tmpConfDir);
-    FileUtils.copyFileToDirectory(
-        new File(testHomeConfDir, "schema-add-schema-fields-update-processor.xml"), tmpConfDir);
-    FileUtils.copyFileToDirectory(
-        new File(testHomeConfDir, "solrconfig.snippet.randomindexconfig.xml"), tmpConfDir);
+    Path tmpSolrHome = createTempDir();
+    Path tmpConfDir = FilterPath.unwrap(tmpSolrHome.resolve(confDir));
```
dsmiley, line 63: What's this change about?  I've never heard of FilterPath before, which you added to this test

solr/core/src/test/org/apache/solr/update/CustomTLogDirTest.java
```diff
@@ -161,13 +163,20 @@ private static void validateTlogPath(
...
-    assertEquals(1, list.length);
-    CoreContainer cc = ((EmbeddedSolrServer) client).getCoreContainer();
-    cc.unload(collectionName, true, true, true);
-    assertFalse(resolvedTlogDir.toFile().exists());
+    try (Stream<Path> files = Files.list(resolvedTlogDir)) {
+      List<Path> list =
+          files
+              .filter(
+                  file -> {
+                    return Files.isRegularFile(file)
```
dsmiley, line 171: can remove the return and braces (as before)

solr/core/src/test/org/apache/solr/AnalysisAfterCoreReloadTest.java
```diff
@@ -41,9 +41,9 @@ public class AnalysisAfterCoreReloadTest extends SolrTestCaseJ4 {
 
   @BeforeClass
   public static void beforeClass() throws Exception {
-    String tmpSolrHome = createTempDir().toFile().getAbsolutePath();
-    FileUtils.copyDirectory(new File(TEST_HOME()), new File(tmpSolrHome).getAbsoluteFile());
-    initCore("solrconfig.xml", "schema.xml", new File(tmpSolrHome).getAbsolutePath());
+    Path tmpSolrHome = createTempDir().toAbsolutePath();
+    PathUtils.copyDirectory(TEST_HOME(), tmpSolrHome, StandardCopyOption.COPY_ATTRIBUTES);
+    initCore("solrconfig.xml", "schema.xml", tmpSolrHome.toAbsolutePath());
   }
```
dsmiley, line 47: it's *already* absolute! (and temp dirs are anyway)

solr/core/src/test/org/apache/solr/schema/TestSchemalessBufferedUpdates.java
mlbiscoc: `java.nio.file.ProviderMismatchException: mismatch, expected: FilterPath, got: class sun.nio.fs.UnixPath`
`createTempDir()` returns a [filterPath](https://github.com/apache/lucene/blob/a2c6d3a1c6565b7ff2eef2a5b84a91cf8f8b1ffc/lucene/test-framework/src/java/org/apache/lucene/tests/mockfile/FilterPath.java#L38). From Lucene test-framework, seems you can only resolve a FilterPath with another FilterPath.

So I unwrap it to get the underlying Path instead of Filterpath allowing `resolve` being called inside of `PathUtils.copyFileToDirectory`

solr/core/src/test/org/apache/solr/cli/SolrCLIZkToolsTest.java
mlbiscoc: Ha after taking a look (almost) every `.toAbsolutePath()` is unnecessary with an exception of I think a few and I am sure there is more. Made those changes in [7d02c14](https://github.com/apache/solr/pull/3263/commits/7d02c14d81098338744f449e1fd3b66733c9d603)

solr/core/src/test/org/apache/solr/schema/TestSchemalessBufferedUpdates.java
dsmiley: Interesting.  What Lucene is doing makes sense.  I dug deeper and found that the root cause of the problem is in commons-io -- which I just filed a trivial jira issue for: [IO-870](https://issues.apache.org/jira/browse/IO-870).  It's easy to avoid by not calling `PathUtils.copyFileToDirectory`

dsmiley (commented):
damn this PR is massive

epugh:
> damn this PR is massive

@mlbiscoc asked me about if he should break it up via Slack, and i didn't have a good answer...  I know that breaking up PR's can be time consuming and introduce it's own bugs, and a big part of me just wants to get to "done done" on this journey...   So I left it as "it's big, but it's only one thing"...    He did bring up the "break it up" idea and I didn't encourage it, so if that is a big mistake, thats on me!   I love that work that is happening here.    I will say, as I attempted to go through it, I did wonder about the breaking it up, it's a daunting PR to review!

dsmiley:
Yeah it's too late to break up.  If we could wind back the clock with the benefit of hindsight, we could do something specific broadly.  Like remove the silly toAbsolutePaths by themselves.  But why bother break this up at this point?  Too much effort for too little gain.  Just need to struggle through a big PR review.  The **tools** really struggle (GitHub UI), GitHub plugin to IntelliJ.  Well; progress not perfection.  I admit I can be a perfectionist in my feedback but that doesn't mean this PR needs to balloon in scope either.

I would like this PR to try to not add more calls to FilterPath.unwrap if it can be helped.  And to not add more calls to Path.of or Paths.get where it's not suitable to do so.   And referencing FileSystem.getDefault() when some nearby path.getFileSystem() can be used. But these could be addressed in a follow-up aiming at that as well.

mlbiscoc:
Tests seem to fail on `TestCoreDiscovery.java` due to 
```
java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "accessUserInformation")
```

Seems `Files.getPosixFilePermissions` isn't allowed on Crave running the tests. I could try to get around it without the `getPosixFilePermissions` to make files readable or unreadable but I suspect we will still run into the same issue of `Files.setAttribute` after...

dsmiley:
A quick google suggests we should add this to our tests policy for the security manager:

    `permission java.lang.RuntimePermission "accessUserInformation";`

But it wasn't needed before... nonetheless if this fixes the matter then it's fine.  Eventually we'll be disabling the security manager and perhaps implementing other measures to stop tests from writing where they shouldn't.

mlbiscoc:
> A quick google suggests we should add this to our tests policy for the security manager:
> 
> ```
> `permission java.lang.RuntimePermission "accessUserInformation";`
> ```
> 
> But it wasn't needed before... nonetheless if this fixes the matter then it's fine. Eventually we'll be disabling the security manager and perhaps implementing other measures to stop tests from writing where they shouldn't.

I think that worked. I removed permissions from `default.policy` on my local machine to reproduce and added in the `accessUserInformation` in [6b37693](https://github.com/apache/solr/pull/3263/commits/6b376931d3d934f4ea029e49c99d0c004d0d9ca0) which looks to pass now.

solr/core/src/test/org/apache/solr/search/TestRecovery.java
```diff
@@ -1779,11 +1778,11 @@ public void testRecoveryMultipleLogs() throws Exception {
       assertU(adoc("id", "CCCCCC"));
 
       h.close();
-      String[] files = ulog.getLogList(logDir.toPath());
+      String[] files = ulog.getLogList(logDir);
       Arrays.sort(files);
       String fname = files[files.length - 1];
       byte[] content;
-      try (RandomAccessFile raf = new RandomAccessFile(new File(logDir, fname), "rw")) {
+      try (RandomAccessFile raf = new RandomAccessFile(logDir.resolve(fname).toFile(), "rw")) {
```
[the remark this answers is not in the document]
mlbiscoc: I think I'm going to leave this as is. I found some solutions to this but don't want to complicate this PR more than it already is. Will be forced to come back to this anyways because if you want to make File a forbidden API, we need to remove all the `toFile()` calls.

solr/core/src/test/org/apache/solr/core/TestFileSystemConfigSetService.java
```diff
@@ -73,8 +73,14 @@ public void testIgnoresFileUploadsOutsideOfConfigSetDirectory() throws IOExcepti
...
+            ".." + FileSystems.getDefault().getSeparator() + "escapePath",
+            "foo"
+                + FileSystems.getDefault().getSeparator()
+                + ".."
+                + FileSystems.getDefault().getSeparator()
+                + ".."
+                + FileSystems.getDefault().getSeparator()
+                + "bar");
     for (String invalidFilePath : invalidFilePaths) {
       fileSystemConfigSetService.uploadFileToConfig(configName, invalidFilePath, testdata, true);
```
dsmiley, line 85: The javadocs of `uploadFileToConfig` (which I recall writing), specify that the path string provided is forward slash since it's in the abstract namespace of the config, not on the file system (necessarily).

solr/core/src/test/org/apache/solr/core/TestLazyCores.java
```diff
@@ -896,7 +896,7 @@ private LocalSolrQueryRequest makeReq(SolrCore core, String... paramPairs) {
   }
 
   private static String makePath(String... args) {
-    return String.join(File.separator, args);
+    return String.join(FileSystems.getDefault().getSeparator(), args);
```
dsmiley, line 899: This isn't wrong but I think Path.of(.....).toString() would be more idiomatic for its purpose.  And looking at it's callers, just 4 right in a row, you could inline this if you wish (I would).

solr/core/src/test/org/apache/solr/handler/component/FacetPivot2CollectionsTest.java
```diff
@@ -65,8 +65,12 @@ public class FacetPivot2CollectionsTest extends SolrCloudTestCase {
   public static void setupCluster() throws Exception {
     // create and configure cluster
     configureCluster(1)
-        .addConfig(COLL_A, configset("different-stopwords" + File.separator + COLL_A))
-        .addConfig(COLL_B, configset("different-stopwords" + File.separator + COLL_B))
+        .addConfig(
+            COLL_A,
+            configset("different-stopwords" + FileSystems.getDefault().getSeparator() + COLL_A))
```
dsmiley, line 70: I think forward slash always works for Path.resolve, which is what will happen inside configset method with this string.  We'll verify this build on Windows before merging.

mlbiscoc (commented):
Got a VM of windows on my laptop to run the tests. The only test that failed was the file permissions tests in `TestCoreDiscovery` and fixed that accordingly. Still think would be good if someone else ran the test locally (windows and unix). I think this is in a state I'm happy with.

dsmiley:
I ran the tests last Friday, especially the Windows sensitive one you found and they all passed.  I don't recall if there's any outstanding comment of mine that is unfulfilled.  I know there's a contributor that may share his results when running Windows.

mlbiscoc:
> I ran the tests last Friday, especially the Windows sensitive one you found and they all passed. I don't recall if there's any outstanding comment of mine that is unfulfilled. I know there's a contributor that may share his results when running Windows.

Thanks for helping run the tests. Trying to be careful for something this large of a change. Did a few passthroughs of the PR since opening this. At this point I should have went through all the comments in this PR and resolved them but happy to make more changes if anyone catches something.

dsmiley:
I plan to merge this Friday afternoon.  Looks like the tests passed on Windows for another contributor.

dsmiley (commented):
Glad to have merged this finally!  Thanks again @mlbiscoc !

dsmiley:
In IntelliJ, I did a find-usages of File (methods, fields, etc.) across project files.  Excluding gradle (uses File a lot), there are still a number of usages.  For example [TestRestore](https://github.com/apache/solr/blob/86f2afded1252b69b8478a2c795a7657585ef5c2/solr/core/src/test/org/apache/solr/handler/TestRestoreCore.java#L203).  On that spot in particular, wow, the few lines there are so back & forth on Path -> File -> String -> Path ... makes my head spin

HoustonPutman:
@mlbiscoc I think we only have 1 failing test from this, so good job!

`CustomTLogDirTest.testExternal` is failing often enough on the MacOS and Linux builds. One example:  
https://jenkins.thetaphi.de/view/Solr/job/Solr-main-macOS/5363

mlbiscoc:
> @mlbiscoc I think we only have 1 failing test from this, so good job!
> 
> `CustomTLogDirTest.testExternal` is failing often enough on the MacOS and Linux builds. One example: https://jenkins.thetaphi.de/view/Solr/job/Solr-main-macOS/5363

Ah so close.... Thanks @HoustonPutman Will try and fix this today.

HoustonPutman:
Oh, well it looks like the nightly tests are less happy... https://ci-builds.apache.org/job/Solr/job/Solr-NightlyTests-main

mlbiscoc:
Thats sad. I ran these tests many times locally. Does nightly not clean up directories? Many of those tests seem to fail where a file/directory already exists. `./gradlew test` doesn't seem to run into this issue. Hoping the fix should be small.

`CustomTLogDirTest.testExternal` is an interesting one. That seed that was run (or I guess depending on the seed it generates), `LuceneTestCase.createTempDir()` is returning a `windowsPath` object instead of it's normal `filterPath` and `Path` doesn't play well with it. I ran into an issue with `LuceneTestCase.createTempDir()` earlier in this PR and fixed it with `FilterPath.unwrap` and looks like unwraping here bandaids the issue.

Will try to fix these errors today.

mlbiscoc:
[PR](https://github.com/apache/solr/pull/3300) to fix nightly tests and `CustomTLogDirTest.testExternal`

PR #3266, opened 2025-03-13
SOLR-17702: Move ZK ConnectionListener logic out of SolrZkClient

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17702

This will fix the flaky `org.apache.solr.cloud.BasicDistributedZk2Test` (class method) failures.

But in general this is a better way of taking down Solr. There's no reason for SolrZkClient to own this logic anymore. The classes that want to watch on connection state can decide if they want to be async or sync, and how to handle closing the executor themselves.


markrmiller (approved):
LGTM. That existing HashSet rather Set annoys me like a crooked napkin, but that's all I've got and isn't even part of these changes.

dsmiley (approved):

PR #3267, opened 2025-03-14
SOLR-17682: Rename BinaryResponseWriter to JavaBinResponseWriter

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17682

2nd PR for this JIRA; I wanted to do this separately.


gerlowskija (approved):
Love it; the old name always bothered me, and making it more specific will avoid confusing folks now that we have two binary formats (javabin fully, CBOR partially)

PR #3269, opened 2025-03-15
Test fix: SolrExporterTestBase, port bind

dsmiley:
This test flaps because the port is given up briefly and another process might take it.
It'd be nice if the exporter could accept a '0' and then have a getter for the port... but this retry should be fine too.

(I already reviewed this with my colleague @pmcarthur-apache)


dsmiley:
will merge tomorrow.

PR #3276, opened 2025-03-19
Replace all with configureEach to solve configuration issues

malliaridis:
# Description

One of the current build issues we are facing is a NPE when `optimizeTask` is called on an undefined task during project configuration. This and similar issues are caused in various configurations that make use of the `TaskCollection.all` method. 

The actual root cause of these errors seems to be caused by our customization for the javadoc tasks in `render-javadoc.gradle`. I am not sure why this is happening and if similar errors could occur elsewhere.

# Solution

This PR changes the method from `all` to `configureEach` for the problematic locations. They are somewhat equivalent:

_org.gradle.api.DomainObjectCollection<T>.all_

> Executes the given closure against all objects in this collection, and any objects subsequently added to this collection. The object is passed to the closure as the closure delegate. Alternatively, it is also passed as a parameter.

 _org.gradle.api.DomainObjectCollection<T>.configureEach_

 >Configures each element in this collection using the given action, as each element is required. Actions are run in the order added.

Since the collections are not altered, it should be safe to use `configureEach`.

Additionally, during the development of the [initial PR](https://github.com/apache/solr/pull/2605), `java.lang.Iterable<T>.forEach` was used in two other places, so they are changed to `configureEach` as well.

# Tests

The NPE caused when specifying an alternative JDK via `RUNTIME_JAVA_HOME` should be fixed with this change. To test it, run the current state with the environment variable `RUNTIME_JAVA_HOME` pointing to a different JDK installation. WIthout the change it should fail to build the project when running `gradlew check -x test`.

# Checklist

### DIFF
--- a/gradle/hacks/dummy-outputs.gradle
+++ b/gradle/hacks/dummy-outputs.gradle
@@ -35,7 +35,7 @@ allprojects {
         "gradlewScriptsTweaked",
         "validateSourcePatterns",
         "validateLogCalls"
-    ] }.forEach { Task task ->
+    ] }.configureEach { Task task ->
     task.logger.info("Setting up ${task.path} as no-output.")
     setupDummyOutputs(task)
   }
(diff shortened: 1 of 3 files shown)


HoustonPutman (approved):
Amazing fine. I have no idea how you found this out or how this creates the NPE but if it works, it works. It seems like a fine change regardless.

PR #3277, opened 2025-03-20
Test: UpdateRequestProcessorFactoryTest, can be flaky

dsmiley:
_(already reviewed by a colleague & contributor @pmcarthur-apache )_

I'm nearly sure that `S3MockRule.builder().silent()` will change the root logger to WARN.  Maybe for JUL; not sure if such a change federates to Log4j2.  Maybe we should remove silent() but I didn't check how bad/verbose it is.


dsmiley:
Includes removal of `S3MockRule silent()` calls that manipulating logging configuration without restoring it.  The logs are not verbose enough to bother trying to suppress IMO.

PR #3281, opened 2025-03-20
SOLR-17714: Add a FuzzyQParser for all FuzzyQuery customizations

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17714

Still need to add some tests

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/FuzzyQParserPlugin.java
+++ b/solr/core/src/java/org/apache/solr/search/FuzzyQParserPlugin.java
@@ -35,7 +35,7 @@ import org.apache.solr.schema.TextField;
  * <p>Example: <code>{!fuzzy f=myfield maxEdits=1 prefixLength=3 maxExpansions=2}foobar</code>
  *
  * @see QueryParser
- * @since 4.0
+ * @since 9.9
  */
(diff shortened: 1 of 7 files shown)


solr/core/src/java/org/apache/solr/search/FuzzyQParserPlugin.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    protected String analyzeIfMultitermTermText(String field, String part) {
+      FieldType fieldType = req.getSchema().getFieldTypeNoEx(field);
+      if (part == null
+          || !(fieldType instanceof TextField)
+          || ((TextField) fieldType).getMultiTermAnalyzer() == null) return part;
+
+      SchemaField sf = req.getSchema().getFieldOrNull((field));
+      if (sf == null || !(fieldType instanceof TextField)) return part;
```
anshumg, line 96: `!(fieldType instanceof TextField)` 
will always be false, right? The conditional check in the line above will return if this condition is true.

anshumg (approved):
Left one comment, other than that looks good to me.

anshumg (approved):

PR #3283, opened 2025-03-24
SOLR-17716: Handle interrupted exception in SolrCores.waitAddPendingCoreOps.

bruno-roustant:
https://issues.apache.org/jira/browse/SOLR-17716

SolrCores.waitAddPendingCoreOps is the main locking mechanism for operations on cores like creation/loading/removal.

In the current code, if an InterruptedException is caught while waiting for the lock, the method returns null. However returning null means basically that the waiting is complete and the caller got the lock. So the caller code proceeds normally with its critical operation on the core.

We should handle differently an InterruptedException to prevent the caller from taking the lock while another pending op is on the core. The proposal is to throw a SolrException.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -2323,16 +2323,14 @@ public class CoreContainer {
     // TestConfigSetsAPI and TestLazyCores
     if (desc == null || zkSys.getZkController() != null) return null;
 
-    // This will put an entry in pending core ops if the core isn't loaded. Here's where moving the
-    // waitAddPendingCoreOps to createFromDescriptor would introduce a race condition.
-    core = solrCores.waitAddPendingCoreOps(name);
-
-    if (isShutDown) {
-      // We're quitting, so stop. This needs to be after the wait above since we may come off the
-      // wait as a consequence of shutting down.
-      return null;
-    }
     try {
+      // This will put an entry in pending core ops if the core isn't loaded. Here's where moving
+      // the waitAddPendingCoreOps to createFromDescriptor would introduce a race condition.
+      core = solrCores.waitAddPendingCoreOps(name);
+      if (isShutDown) {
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -2323,16 +2323,14 @@ public SolrCore getCore(String name, UUID id) {
...
-
-    if (isShutDown) {
-      // We're quitting, so stop. This needs to be after the wait above since we may come off the
-      // wait as a consequence of shutting down.
-      return null;
-    }
     try {
+      // This will put an entry in pending core ops if the core isn't loaded. Here's where moving
+      // the waitAddPendingCoreOps to createFromDescriptor would introduce a race condition.
+      core = solrCores.waitAddPendingCoreOps(name);
```
psalagnac, line 2329: This is not safe to move this call inside the `try` block.
If an exception is raised for whatever reason, it means we did not have the lock. In this case, we should **not** execute the finally block that releases the lock.

bruno-roustant: You're right. I wanted to have a consistent way to try-finally between all the calls to waitAddPendingCoreOps.
That means that changing waitAddPendingCoreOps to throw an exception is either not appropriate, or needs more care with the other callers. And btw, the other callers may have currently an issue if an exception is thrown inside waitAddPendingCoreOps.

psalagnac: I think it is fine to throw an exception as long as we don't remove the lock in case of exception. I'm good with your change in `SolrCores` (the other file).

bruno-roustant: Fixing the 5 calls is straightforward, so let's keep the exception.

psalagnac (approved):

bruno-roustant:
Thanks Pierre for your review.

PR #3284, opened 2025-03-24
Do not enable security manager on JDK 24+ (#14179)

uschindler:
This is basically a cherry-pick of Chris Hegarty's commit on Lucene ([faec0f823817ca95f1f103d6b9482d26ee75cc7b](https://github.com/apache/lucene/commit/faec0f823817ca95f1f103d6b9482d26ee75cc7b), PR https://github.com/apache/lucene/pull/14179). This will run all Solr tests without security manager enabled, once the runtime JVM for testing is 24 or later.

Original issue description: "This commit avoids setting the security manager on JDK 24+ - since it is not longer possible to enable it in JDK 24+ This is the minimum required to start testing with JDK 24 EA."

This will make Policeman Jenkins builds work again with Java 24+

Thanks to @ChrisHegarty.

### DIFF
--- a/gradle/testing/randomization.gradle
+++ b/gradle/testing/randomization.gradle
@@ -92,7 +92,9 @@ allprojects {
           [propName: 'tests.asserts', value: "true", description: "Enables or disables assertions mode."],
           [propName: 'tests.infostream', value: false, description: "Enables or disables infostream logs."],
           [propName: 'tests.leaveTemporary', value: false, description: "Leave temporary directories after tests complete."],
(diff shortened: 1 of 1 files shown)


uschindler:
This only fixes the build issues. I don't know about Solr startup scripts, maybe they were updated already. So this is really only the Test-SecurityManager support.

uschindler:
I will backport this to the 9.x branch.

dsmiley (approved):

PR #3285, opened 2025-03-24
Fix cross-dc security manager usage. Port some Lucene test setup

HoustonPutman:
Currently both Cross DC modules are using the security manager for unsupported versions, even after #3284. Fix that.

Also include some test setup config that Lucene uses


uschindler (approved):

uschindler:
Thanks tis looks nice. To me it was totally unclear why only those tests failed.

solr/modules/cross-dc/build.gradle
```diff
@@ -24,7 +24,7 @@ description = 'Cross-DC Replication Plugins'
 ext {
   // Use of the security manager is not supported on windows.
   // Mockito does not work with it enabled.
-  useSecurityManager = !Os.isFamily(Os.FAMILY_WINDOWS)
+  useSecurityManager = rootProject.ext.runtimeJavaVersion <= JavaVersion.VERSION_23 && !Os.isFamily(Os.FAMILY_WINDOWS)
```
uschindler, line 27: this is basically an addition which was used later.

Maybe change this to use `useSecurityManager &= !Os.isFamily(Os.FAMILY_WINDOWS)`

This would take the setting from top-level.

uschindler (changes requested):

solr/modules/cross-dc/build.gradle
uschindler: Ah this needs to be done differently. I would do the `&` trick here:

https://github.com/apache/solr/blob/de09232c53a18a0391d7386fea21a2aaae03af66/gradle/testing/randomization.gradle#L207

Uwe

uschindler: Change it to `&=`here and it will work correctly also for possible other modules.

HoustonPutman: Agreed, this is much better

PR #3287, opened 2025-03-25
Only apply test failure output logic for Java tests

HoustonPutman:
This error has been seen on some of the Jenkins runs:

```
Caused by: groovy.lang.MissingPropertyException: Could not get unknown property 'testOutputsDir' for task ':solr:ui:desktopTest' of type org.jetbrains.kotlin.gradle.targets.jvm.tasks.KotlinJvmTest.
```

There's no reason to (currently) add the output for fucit or develocity for these tests, so let's just keep the logic to the JavaProject modules.

### DIFF
--- a/gradle/testing/defaults-tests.gradle
+++ b/gradle/testing/defaults-tests.gradle
@@ -23,7 +23,7 @@ def resources = scriptResources(buildscript)
 def verboseModeHookInstalled = false
 
 allprojects {
-  plugins.withType(JavaPlugin) {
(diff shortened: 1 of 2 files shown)


risdenk (approved):

malliaridis (approved):
Nice and simple solution. 👍

I saw the error a few times now, but couldn't reliably reproduce it. From what I remember for a different fix, all the kotlin test tasks did not depend on the `test` gradle task, but are still of the type `Test` to be covered by this fix?

PR #3291, opened 2025-03-27
SOLR-17717 Use Powershell instead of WMIC to get cmdline on Windows

janhoy:
https://issues.apache.org/jira/browse/SOLR-17717

The same information as we fetch with WMIC comamnd can be fetched with `powershell.exe`. It is slower, but more universal. Alternative would be native code via the OSHI library, but too heavy to include for just this one use.


epugh (approved):
LGTM.   

One question about the approach...  Could we JUST use powershell and not keep wmic support?   Eventually it will be gone...  Powershell has been around for quite a while now....

janhoy:
I'm not fluent enough in Windows world to know whether powershell.exe is available on all home, pro and server versions, so due to lack of integration tests on windows, it feels safer to handle it as a fallback for the existing approach. I wish Java implemented this for us in `Process` on Windows too. `ph.info().commandLine()` is so much nicer than these workarounds.

janhoy:
It is tempting to just defer this complexity to https://github.com/oshi/oshi as we discussed in the original feature issue.

uschindler:
Powershell is available in Windows 10 and 11 for all variants. Servers have it for sure. Windows 8 also has Powershell but its out of date. Windows 7 can install it. If anybody complains about Windows XP, I'm out of arguments.

epugh:
> It is tempting to just defer this complexity to https://github.com/oshi/oshi as we discussed in the original feature issue.

I thought you found a limitation that prevented oshi being viable?  It looks like an active maintained library ;-).

uschindler:
> > It is tempting to just defer this complexity to https://github.com/oshi/oshi as we discussed in the original feature issue.
> 
> I thought you found a limitation that prevented oshi being viable? It looks like an active maintained library ;-).

Please don't add new libraries (especially those with native code) just for a single function call. I'd remove the wmic support completely. It might also no longer be installed on freshly installed Windows 11 Pro or Servers. If you add libraries with native code, you need to whitelist them in later Java versions, because by default next LTS will disable native code via JNI and Panama. So it is now mandatory to add the command line option to enable it for Lucene Core (see recent upgrades to build scripts).

Uwe

janhoy:
Thanks for the review. I removed RMIC support. I have tested the PowerShell option on my UTM Win11 ARM VM and it works. However, my user is administrator, so I'd need to test with an unprivileged user as well. Feel free to give it a spin if you are on Win..

janhoy:
Did some more testing and a weakness of the solution is that the Powershell process is spawned once for every Java process running on the system, and each invocation takes some 3-5 seconds on the Win11 VM on my MacBook. Don't know if such slow invocation is representative for a real windows system, but it is an overhead. I have not done a comparison for the old wmic invocation. Manual command invocation in a PS terminal takes <1s. While manually invoking `powershell.exe -Command "foo"` from a CMD terminal takes 2+ secs, perhaps there's a startup cost in starting powershell process.

So I refactored the code to instead of calling PS once per PID, we do one full PID list on startup with
```
"Get-CimInstance -ClassName Win32_Process | Where-Object { $_.Name -like '*java*' } | Select-Object ProcessId, CommandLine | ConvertTo-Json -Depth 1"
```
and stuffs it into a `Map<Long, String>` for later lookup.

The OSHI library, using jna would probably be far more performant, but I agree that adding a 1Mb jar just for this is too much.

solr/core/src/java/org/apache/solr/cli/SolrProcessManager.java
```diff
@@ -162,42 +171,60 @@ private static Optional<String> commandLine(ProcessHandle ph) {
...
+      StopWatch stopWatch = StopWatch.createStarted();
+      Process process =
+          new ProcessBuilder(
+                  "powershell.exe",
+                  "-Command",
+                  "Get-CimInstance -ClassName Win32_Process | Where-Object { $_.Name -like '*java*' } | Select-Object ProcessId, CommandLine | ConvertTo-Json -Depth 1")
+              .redirectErrorStream(true)
+              .start();
+      try (InputStreamReader inputStreamReader =
+              new InputStreamReader(process.getInputStream(), StandardCharsets.UTF_8);
```
uschindler, line 194: Why not directly tell jackson to load the JSON from the `inputStreamReader`? To me it looks strange to read it line by line and concat it together. This should be the task of the JSON parser.

uschindler:
>  Don't know if such slow invocation is representative for a real windows system, but it is an overhead.

Powershell is slow to start. Yes. Also on real machines.

solr/core/src/java/org/apache/solr/cli/SolrProcessManager.java
uschindler: ...or just use `IOUtils.toString(reader)` or better `IOUtils.toString(stream, charset)`

janhoy: Yea, added onion layers, will take a stab on peeling.

uschindler: I am sure starting the process and reading/converting the output can at end be a 4-liner using Java streaming API ;-)

uschindler: P.S.: I like the idea to use JSON output. That's much better than applying regexes and line-by line parsing of arbitrary tabular output.

solr/core/src/java/org/apache/solr/cli/SolrProcessManager.java
```diff
@@ -162,42 +169,60 @@ private static Optional<String> commandLine(ProcessHandle ph) {
...
+   *
+   * @return a map of process IDs to command lines
+   */
+  private static Map<Long, String> commandLinesWindows() {
+    try {
+      Process process =
+          new ProcessBuilder(
+                  "powershell.exe",
+                  "-Command",
+                  "Get-CimInstance -ClassName Win32_Process | Where-Object { $_.Name -like '*java*' } | Select-Object ProcessId, CommandLine | ConvertTo-Json -Depth 1")
```
janhoy, line 187: The `Get-CimInstance` is not available in Powershell 2.0 (Windows 7). We could use `Get-WmiObject` instead, but it is slower. Don't think we need to care too much about Windows 7 these days.

janhoy:
Coming back to this. Lacking test suite on multiple windows versions makes such changes brittle. I hope the iterations we have done here and the reviews is enough to merge. And then hope the fix is not worse than what it replaces, except for perhaps added latency launching powershell. I'll go ahead and merge to main for now.

PR #3292, opened 2025-03-27
SOLR-17692: Abort ongoing fetches on core close

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-17692

# Description

I recently deleted a NRT replica that was in the middle of a full-recovery and was a bit surprised to see that the "delete" blocked waiting for the recovery to finish. This is a minor pain when the index is small, but becomes a huge waste of administrator time (and network bandwidth!) as index sizes grow.

There's some plumbing in Solr that attempts to preempt recovery during a DELETE, but it appears that it mostly comes into play during peer-sync and "background replication" scenarios (i.e. PULL and TLOG replicas that do full-recovery during normal operation). Preemption doesn't seem to work once a recovering core is in the midst of a "full recovery".

`RecoveryStrategy.close()` sets a boolean flag that is checked at various points in the recovery process, and can preempt recovery at those points.  But the flag isn't visible to `IndexFetcher`, where the bulk of a full-recovery's time gets spent.

# Solution

This commit aims to fix this by having `RecoveryStrategy.close()` invoke an existing `ReplicationHandler.abortFetch`, which sets a flag that *is* noticed by IndexFetcher.  This should ensure that long-running recovery operations don't block DELETEREPLICA or other core-shutdown paths.

# Tests

Manual testing.  May need some help figuring out a way to exercise this in an automated test, given the timing-dependent nature of the bug and fix.

# Checklist


dsmiley (commented):
+1 Looks like a straight-forward improvement.  Nice job explaining the complicated matter!

solr/core/src/java/org/apache/solr/cloud/RecoveryStrategy.java
```diff
@@ -188,6 +189,9 @@ public final void close() {
     close = true;
     cancelPrepRecoveryCmd();
     log.warn("Stopping recovery for core=[{}] coreNodeName=[{}]", coreName, coreZkNodeName);
+    if (replicationHandlerDoingFetch != null) {
```
gerlowskija, line 192: [Q] Technically, there's a possibility here that 'replicationHandlerDoingFetch' is unset after we've checked the conditional, but before we call `abortFetch()1.  If that happens, it's lead to a NPE on the `abortFetch()` call.  But it seems so remote that I'm not sure it's worth the complexity of guarding against it....

gerlowskija:
Thanks!

I'd still really like to find a way to get this covered by automated tests, but I can't come up with a good way to cover it.  A unit test should be possible but the mocking would be pretty ugly.  And idk of a good way to replicate the necessary timing in a higher-level test. (i.e. to ensure that the core-close happens _during_ IndexFetcher's work, and not before or after).  Open to suggestions if anyone has them!

PR #3295, opened 2025-03-30
SOLR-17723: Use NamedList.forEach() or iterator()

dsmiley:
In order to be more Map-like.
And avoid some needless indexed method usages.

https://issues.apache.org/jira/browse/SOLR-17723

As this is just an internal refactor, I wasn't thinking of adding to CHANGES.txt.  Also wasn't planning on backporting but could change my mind.  The changes here included some occassional nearby Java 17+ tweaks like better switch expressions.


dsmiley:
This was a lot of work... even with ClaudeMind doing many transformations (not all); I did some myself and reviewed it.  It even found a (minor?) bug that I left as a TODO for scope reasons.
I intentionally didn't touch the response format writing & reading code for performance considerations.  I want this PR to be low-risk.
I'll merge this weekend without a review suggesting the contrary.

epugh:
I spot checked a couple of places and it all seems to make sense!

PR #3300, opened 2025-04-01
SOLR-16903: Nightly test failures from Path migration

mlbiscoc:

# Description

Tests on nightly failing.

# Solution

- `TestReplicationHandler` needs to check fileName from string
- `TestStressThreadBackup.java` `toList()` is immutable. Collect with `.collect(Collectors.toList())`
- `BaseDistributedSearchTestCase.java` seems with a nightly test, the file/directory already exists. Use `StandardCopyOption.REPLACE_EXISTING` to overwrite as `FileUtils.copyDirectory` does
- CustomTLogDirTest.java When using `LuceneTestCase.createTempDir()`, certain seeds generate a [WindowsFS](https://github.com/apache/lucene/blob/main/lucene/test-framework/src/java/org/apache/lucene/tests/mockfile/WindowsFS.java#L34) and this causes try-with-resources failing to [close the file stream](https://github.com/apache/solr/blob/470ae762b883b259da6f127844855395e9eedf9f/solr/core/src/test/org/apache/solr/update/CustomTLogDirTest.java#L166). Unwrap it to just get a generic NIO Path so stream can close

# Tests

# Checklist

### DIFF
--- a/solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
+++ b/solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
@@ -26,6 +26,7 @@ import java.nio.file.Path;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Stream;
+import org.apache.lucene.tests.mockfile.FilterPath;
 import org.apache.solr.SolrTestCaseJ4;
 import org.apache.solr.cloud.AbstractDistribZkTestBase;
 import org.apache.solr.cloud.AbstractZkTestCase;
@@ -376,10 +377,11 @@ public class ZkSubcommandsTest extends SolrTestCaseJ4 {
     assertEquals(confsetname, collectionProps.getStr("configName"));
 
     // test down config
-    Path configSetDir =
-        tmpDir.resolve(
-            "solrtest-confdropspot-" + this.getClass().getName() + "-" + System.nanoTime());
-    assertFalse(Files.exists(configSetDir));
+    Path destDir =
+        FilterPath.unwrap(
+            tmpDir.resolve(
+                "solrtest-confdropspot-" + this.getClass().getName() + "-" + System.nanoTime()));
+    assertFalse(Files.exists(destDir));
 
     args =
         new String[] {
@@ -387,15 +389,14 @@ public class ZkSubcommandsTest extends SolrTestCaseJ4 {
           "--conf-name",
           confsetname,
           "--conf-dir",
-          configSetDir.toString(),
+          destDir.toString(),
           "-z",
           zkServer.getZkAddress()
         };
 
     assertEquals(0, CLITestHelper.runTool(args, ConfigSetDownloadTool.class));
 
-    Path confSetDir = configSetDir.resolve("conf");
-    try (Stream<Path> filesStream = Files.list(confSetDir)) {
+    try (Stream<Path> filesStream = Files.list(destDir.resolve("conf"))) {
       List<Path> files = filesStream.toList();
       zkFiles =
           zkClient.getChildren(ZkConfigSetService.CONFIGS_ZKNODE + "/" + confsetname, null, true);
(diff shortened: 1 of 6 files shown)


solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
```diff
@@ -376,26 +377,26 @@ public void testUpConfigLinkConfigClearZk() throws Exception {
     assertEquals(confsetname, collectionProps.getStr("configName"));
 
     // test down config
-    Path configSetDir =
-        tmpDir.resolve(
-            "solrtest-confdropspot-" + this.getClass().getName() + "-" + System.nanoTime());
-    assertFalse(Files.exists(configSetDir));
+    Path destDir =
```
dsmiley, line 380: A clearer name, as this test deals with many configSetDirs, as input and output.

solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
```diff
@@ -406,45 +407,35 @@ public void testUpConfigLinkConfigClearZk() throws Exception {
       assertEquals("Comparing downloaded files to what is in ZK", files.size(), zkFiles.size());
     }
 
-    Path sourceConfDir = ExternalPaths.TECHPRODUCTS_CONFIGSET;
-    // filter out all directories starting with . (e.g. .svn)
-    try (Stream<Path> stream = Files.walk(sourceConfDir)) {
-      List<Path> files =
-          stream
-              .filter(Files::isRegularFile)
-              .filter(path -> path.getFileName().startsWith("."))
```
dsmiley, line 415: There were a comedy of errors in this test; the "worst" one was here, having multiple flaws.  Firstly, `filter` chooses those matching the condition; someone forgot to invert this!  Thus the test didn't loop over anything.  Secondly, the interpretation of this method is for "." to be a path, since we are calling Path.startsWith *not* String.startsWith.  IMO the JDK developers chose a terribly confusing name that is forbidden-api worthy using "startsWith" taking a String.  So any way, this ends up sort of kind of being, filter to paths that are under the current directory as that's what "." resolves to (CWD)... but only if normalize().toAbsolute() were to be called; which it wasn't and we don't want any way since this wasn't the actual intention.

In the end I dropped the intended filter since I don't see that we would encounter a dot-file.

CC @epugh (original author) @mlbiscoc (latest author)

dsmiley (commented):
_(note: in GitHub, you can check a box to ignore whitespace when looking at a diff.  Highly recommended)_

I ran `org.apache.solr.cli.ZkSubcommandsTest#testUpConfigLinkConfigClearZk` in a debugger to ensure it actually executes as-expected, at least on macOs

solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
```diff
@@ -406,45 +407,35 @@ public void testUpConfigLinkConfigClearZk() throws Exception {
...
-          });
+    // compare the original source with the downloaded destination
+    try (Stream<Path> stream = Files.walk(ExternalPaths.TECHPRODUCTS_CONFIGSET)) {
+      stream
+          .filter(Files::isRegularFile)
+          .forEach(
+              (sourceFile) -> {
+                Path sourceFileRelative =
+                    ExternalPaths.TECHPRODUCTS_CONFIGSET.relativize(sourceFile);
+                Path downloadedFile = destDir.resolve("conf").resolve(sourceFileRelative);
```
dsmiley, line 418: this was a huge mistake I fixed.  The source and destination were the same!  The destination *should* have been at the temp dir but it was a generically worded "confDir" that just so happened to be `ExternalPaths.TECHPRODUCTS_CONFIGSET` (the source) (!).  Thus the test was very flawed if it even ran, but it didn't as I explained all files were filtered out.

solr/core/src/test/org/apache/solr/cli/ZkSubcommandsTest.java
mlbiscoc: Thanks for fixing this test. I need to be more mindful when modifying a block this big. I agree startsWith is very confusing with path for migration and the same issue was ran into in another part of this PR.

PR #3301, opened 2025-04-01
SOLR-17724: clarify basic auth credential docs

epugh:
https://issues.apache.org/jira/browse/SOLR-17724

10x only

### DIFF
--- a/solr/solr-ref-guide/modules/deployment-guide/pages/basic-authentication-plugin.adoc
+++ b/solr/solr-ref-guide/modules/deployment-guide/pages/basic-authentication-plugin.adoc
@@ -341,15 +341,17 @@ httpBasicAuthPassword=secretPassword
 == Using the Solr Control Script with Basic Auth
 
 Once Basic authentication is enabled, all requests to the Solr Control Script (`bin/solr`) must contain user credentials.
-To ensure this, add the following line to the `solr.in.sh` or `solr.in.cmd` file.
+While you can pass credentials into `bin/solr` directly when invoking a command, you can also globally define the credentials in the `solr.in.sh` or `solr.in.cmd` file.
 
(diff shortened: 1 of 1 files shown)


solr/solr-ref-guide/modules/deployment-guide/pages/basic-authentication-plugin.adoc
```diff
@@ -81,6 +81,14 @@ If `realm` is not defined, it will default to `solr`.
 If you are using SolrCloud, you must upload `security.json` to ZooKeeper.
 An example command and more information about securing your setup can be found at xref:authentication-and-authorization-plugins#in-a-solrcloud-cluster[Authentication and Authorization Plugins In a SolrCloud Cluster].
 
+=== Password Encoding
+
+Solr stores the passwords in the format: `base64(sha256(sha256(salt+password))) base64(salt)`.  
+
+If you edit `security.json` directly then you need to encode the password yourself.  
+You can visit https://clemente-biondo.github.io/ to use a simple web utility that does the encoding for you.
```
janhoy, line 89: Dangerous to link to 3rd party tool, we cannot know if his account gets compromised and the script modified. Would rather copy the JS and use it somewhere in our own ref-guide or website, or perhaps even as part of Solr Admin?

gus-asf: I once had a client that couldn't figure out how to give me access to their solr admin ui. After multiple failed attempts and clear intent that I *should* have access, I looked up the code in Solr, executed it in an IntelliJ groovy console and generated myself a password, which I then added it to their (unsecured) zookeeper to gain access. (and advised them of what I had done and the need to secure things further). My point is, if we are going to document this, we could easily expose a tool based on our code rather than reference a 3rd party.

PR #3303, opened 2025-04-02
Forbidden-API: Path.startsWith(String)

dsmiley:
Very error-prone; confused with String.startsWith.

CC @mlbiscoc


epugh:
Nice use of our tooling!

PR #3311, opened 2025-04-05
SOLR-17496: Remove a warning that is no longer applicable.

epugh:
https://issues.apache.org/jira/browse/SOLR-17496

At smaller scale setups, running Solr in Solr Cloud mode with embedded ZooKeeper is just fine.   1 node, 2 node, it's just fine.   

For 10.0 I plan on completing #2783 which would provide more concrte details.


dsmiley (approved):
Yes!

PR #3312, opened 2025-04-07
SOLR-17864: Migrate to dot separated naming of environment variables for a initial set of variables.

epugh:

# Description
Replace legacy property names with the modern default dot notation.

# Solution
This started out of some exploration of how to merge the `solr.in.sh` and `solr.in.cmd` files, and branched off from there.

I found that if we replaced the legacy variable names (that are mapped in `EnvToSyspropMappings.properties`) then we could simplify some of the code.   

A question, do we need a deprecation type warning if this change is 10x only?  One idea I had is that instead or removing the mappings from `EnvToSyspropMappings.properties`, we could instead have two lists.  And if you use the value from the deprecated list, then you get a loud warning in the logs, with the idea that it gets removed in 11?  

# Tests

Running existing tests.   If we add the warning idea, then will need another test.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -17,6 +17,8 @@ New Features
 * SOLR-14414: Introduce a new, experimental Admin UI that can be used aside the current Admin UI. The new module is
   loaded together with the existing Admin UI and available under the URL path /solr/compose. (Christos Malliaridis)
 
+* SOLR-17864: Standarized patterns for system property naming across Solr.  i.e enable.packages is now solr.packages.enabled.  (janhoy, Eric Pugh)
+
 Improvements
 ---------------------
 


dsmiley:
It's not clear this is needed or matters.  In the code I like seeing "solr.<module>.<name>" wehre the module and name are camel case.  But I can get over it.

epugh:
> It's not clear this is needed or matters. In the code I like seeing "solr.." wehre the module and name are camel case. But I can get over it.

Thanks for the feedback ;-).  I'm happy to elaborate on reasons why if helpful ;-)

HoustonPutman:
Yeah, I'm not sure this is better (and definitely not sure if its worth the effort)

epugh:
We should probably make sure we are all on the same page, before I keep plugging before.   I, maybe erroneously, assumed that since there was a TODO in the https://github.com/apache/solr/blob/main/solr/solrj/src/resources/EnvToSyspropMappings.properties#L6 file that this was something as a community we had agreed on.  Since Solr 10 is coming, I wanted to get this in!

epugh:
Okay, before I polish off migrating the rest of the ones in https://github.com/apache/solr/pull/3312/files#diff-012f0766c431d2622a99ac1f70addc1d1f90106816a94434a39db6945db23601, I think this should probably get merged, and then I can do a second batch...?

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1766,7 +1766,7 @@ private SolrCore createFromDescriptor(
         // this mostly happens when the core is deleted when this node is down
         // but it can also happen if connecting to the wrong zookeeper
         final boolean deleteUnknownCores =
-            Boolean.parseBoolean(System.getProperty("solr.deleteUnknownCores", "false"));
+            Boolean.parseBoolean(System.getProperty("solr.delete.unknown.cores", "false"));
```
janhoy: `solr.cloud.startup.delete.unknown.cores.enabled`?

solr/core/src/java/org/apache/solr/cli/PackageTool.java
```diff
@@ -334,7 +334,7 @@ public String getHeader() {
         "Note: (a) Please add '--solr-url http://host:port' parameter if needed (usually on Windows).");
     format(
         sb,
-        "      (b) Please make sure that all Solr nodes are started with '-Denable.packages=true' parameter.");
+        "      (b) Please make sure that all Solr nodes are started with '-Dsolr.enable.packages=true' parameter.");
```
janhoy: `solr.packages.enabled` ?

solr/core/src/java/org/apache/solr/servlet/LoadAdminUiServlet.java
```diff
@@ -39,9 +39,9 @@
  */
 public final class LoadAdminUiServlet extends HttpServlet {
 
-  // check system properties for whether or not admin UI is disabled, default is false
+  // check system properties for whether the admin UI is disabled, default is false
   private static final boolean disabled =
-      Boolean.parseBoolean(System.getProperty("disableAdminUI", "false"));
+      Boolean.parseBoolean(System.getProperty("solr.admin.ui.disabled", "false"));
```
janhoy: `solr.ui.disabled` ?

solr/core/src/java/org/apache/solr/servlet/SolrRequestParsers.java
```diff
@@ -114,8 +114,8 @@ public SolrRequestParsers(SolrConfig globalConfig) {
       formUploadLimitKB = globalConfig.getFormUploadLimitKB();
 
       // security risks; disabled by default
-      enableRemoteStreams = Boolean.getBoolean("solr.enableRemoteStreaming");
-      enableStreamBody = Boolean.getBoolean("solr.enableStreamBody");
+      enableRemoteStreams = Boolean.getBoolean("solr.enable.remote.streaming");
+      enableStreamBody = Boolean.getBoolean("solr.enable.stream.body");
```
janhoy: `solr.requests.streaming.remote.enabled` and `solr.requests.streaming.body.enabled` ?

solr/solrj/src/java/org/apache/solr/common/util/EnvUtils.java
```diff
@@ -32,28 +33,47 @@
...
+      Properties deprecatedProps = new Properties();
       try (InputStream stream =
-          EnvUtils.class.getClassLoader().getResourceAsStream("EnvToSyspropMappings.properties")) {
+              EnvUtils.class
+                  .getClassLoader()
+                  .getResourceAsStream("EnvToSyspropMappings.properties");
+          InputStream stream2 =
+              EnvUtils.class
+                  .getClassLoader()
+                  .getResourceAsStream("EnvToSyspropDeprecatedMappings.properties")) {
```
janhoy: More meaninful variable names than `stream` and `stream2`. 
The file name `EnvToSyspropDeprecatedMappings.properties` gives impression that this file maps env.variables to sys props, but the file itself says it maps sys props to sys props?

Q: Would it be possible to add deprecations as a section in the existing properties file, perhaps with a prefix `deprecated.*` to keep it in one place?

solr/core/src/java/org/apache/solr/core/CoreContainer.java
janhoy: I can see that this is a painfully long prop, and I kind of agree with Smiley that `solr.cloud.startup.deleteUnknownCores` would be a better choice. But allowing CamelCase breaks the simple translation rule that you can take an env.var and lowercase it and replace underscores with dots to get the sysprop. Perhaps we could introduce dashes like `SOLR_CLOUD_STARTUP_DELETE-UNKNOWN-CORES` but it soon gets to complex. Naming has been and is a hard thing.

I'm personally willing to embrace structure and simplified rules for parsing/mapping and sacrifice some readability or brewity. Perhaps further down the road we could build one very simple config parser that instantiates config objects without all the bolier-plate parsing code we have today.

epugh:
Thanks for letting oking at this!  I had tuna but out of steam on it but will try and pick it up again and get this batch of changes in.  Definitely 10x change.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
epugh: Humm..   I can see the additional detail being being useful in what you are proposing.    I used to want brevity, and then I started using Kubernetes and saw how long it's commands are!

epugh:
Just picking this up again and getting synced to latest...  I am thinking this would only go to Solr 10....


Thanks for the comments @janhoy ....

epugh:
Just added a JIRA.  I would like to get this batch committed, before I do more.....

solr/core/src/java/org/apache/solr/cli/PackageTool.java
epugh: sounds good!  I like that booleans are ending with `.enabled`.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
epugh: Working on it...

solr/core/src/java/org/apache/solr/servlet/LoadAdminUiServlet.java
epugh: what do you think about `solr.ui.enabled` and having it be defaulted to true??

solr/core/src/java/org/apache/solr/servlet/SolrRequestParsers.java
epugh: love it

solr/core/src/java/org/apache/solr/servlet/LoadAdminUiServlet.java
epugh: i went ahead and made this change.

dev-docs/ui/testing-and-deployment.adoc
```diff
@@ -127,8 +127,8 @@ more secure production configuration.
 
 Like other modules, the new UI is enabled with the jetty module parameter `--module=new-ui`.
 When using the Solr CLI, users can disable the new UI with
-`SOLR_ADMIN_UI_EXPERIMENTAL_DISABLED=true` or by disabling the entire UI with
-`SOLR_ADMIN_UI_DISABLED=true`.
+`SOLR_ADMIN_UI_EXPERIMENTAL_ENABLED=false` or by disabling the entire UI with
+`SOLR_UI_ENABLED=false`.
```
dsmiley: It seems wrong to *both* have "ADMIN_UI" and "UI".  Pick one.  Jan proposed "UI" which I agree with.

solr/bin/solr.in.cmd
```diff
@@ -224,12 +224,12 @@ REM set SOLR_SECURITY_MANAGER_ENABLED=true
 REM This variable provides you with the option to disable the Admin UI. if you uncomment the variable below and
 REM change the value to true. The option is configured as a system property as defined in SOLR_START_OPTS in the start
 REM scripts.
-REM set SOLR_ADMIN_UI_DISABLED=false
+REM set SOLR_UI_ENABLED=false
```
dsmiley: flip the default

solr/bin/solr.in.sh
```diff
@@ -248,14 +248,14 @@
 #SOLR_SECURITY_MANAGER_ENABLED=true
 
 # This variable provides you with the option to disable the Admin UI. If you uncomment the variable below and
-# change the value to true. The option is configured as a system property as defined in SOLR_START_OPTS in the start
+# change the value to false. The option is configured as a system property as defined in SOLR_START_OPTS in the start
 # scripts.
-# SOLR_ADMIN_UI_DISABLED=false
+# SOLR_UI_ENABLED=false
```
dsmiley: flip default

solr/core/src/java/org/apache/solr/core/NodeConfig.java
```diff
@@ -248,7 +248,7 @@ public static NodeConfig loadNodeConfig(Path solrHome, Properties nodeProperties
     //  ZkFailoverTest test case...
     String zkHost = nodeProperties.getProperty(SolrXmlConfig.ZK_HOST);
     if (StrUtils.isNotNullOrEmpty(zkHost)) {
-      int startUpZkTimeOut = 1000 * Integer.getInteger("waitForZk", 0);
+      int startUpZkTimeOut = 1000 * Integer.getInteger("solr.wait.for.zk", 0);
```
dsmiley: no; it misses a category.  "solr.cloud.wait.for.zk.seconds" (I'm proposing a time unit while we're at it)

solr/solrj/src/resources/EnvToSyspropMappings.properties
```diff
@@ -5,29 +5,21 @@
 # This way, env SOLR_FOO_BAR will also match property 'solr.foo.bar' without a mapping in this file
 # TODO: Deprecate non-standard sysprops and standardize on solr.foo.bar in Solr 10
 AWS_PROFILE=aws.profile
-DEFAULT_CONFDIR=solr.default.confdir
```
dsmiley, line 8: I'm not seeing what you replaced that with.

solr/core/src/java/org/apache/solr/core/NodeConfig.java
epugh: so, this is only used in by a test, so it appears the pattern from cwiki page would be solr.tests.wait.for.zk.seconds...

solr/bin/solr.in.sh
epugh: do you mean you prefer the _DISABLED versus teh _ENABLED naming?   AS in `SOLR_UI_DISABLED=true`?   I was thinking that instead of a mishmash of _ENABLED and _DISABLED, we would always end in _ENABLED and they you would specifiy a true or false...

dev-docs/ui/testing-and-deployment.adoc
epugh: The difference is that you can opt out of the new Kotlin based "Solr UI" and keep the classic JS based "Admin UI", hence the two parameters.

epugh:
> The upgrade guide ought to list old/deprecated properties, right? Or if not then, if I'm seeing this code correctly, we log a warning if one is used?

Correct.  There is a bats test for that.  I also added https://issues.apache.org/jira/browse/SOLR-17868 to make sure we don't forget to explicitly document the mappings.

solr/solrj/src/resources/EnvToSyspropMappings.properties
epugh: After digging through the start up scripts, I learned that we don't need this variable, it's ONLY ever used internally in the start up scripts.  There is a SOLR_DEFAULT_CONFDIR that is now used, which follows the pattern, which means we don't need it in this file.

epugh:
@dsmiley I think I have addressed your comments or asked for follow up details.  Getting excited to get this first batch in!

dev-docs/ui/testing-and-deployment.adoc
dsmiley: I understand 2 parameters, but they should share the same category as they are both UI.  Maybe the "experimental" one should be SOLR_UI_EXPERIMENTAL_ENABLED.

solr/bin/solr.in.sh
dsmiley: I don't mean that.  I mean the value in the comment should be inverted itself since you flipped the ENABLED|DISABLED.  The commented value is supposed to be the default value.

solr/core/src/java/org/apache/solr/core/NodeConfig.java
dsmiley: This one is in bin/solr and I recall discussing this setting with colleagues once.  Let's not assume it's only for tests just because you only see a test flip the value in the code.  There's no telling what users do.

solr/solrj/src/resources/EnvToSyspropMappings.properties
dsmiley: ah cool.  But "SOLR_DEFAULT_CONFDIR" is poorly named since "DEFAULT" isn't a category.  I suppose you are deferring such a matter to another PR and are now just focused on certain ones.

solr/solrj/src/java/org/apache/solr/common/util/EnvUtils.java
epugh: Need to mull on this more!

solr/bin/solr.in.sh
epugh: gotcha

solr/core/src/java/org/apache/solr/core/NodeConfig.java
epugh: I do NOT see it in bin/solr... I did find this: https://issues.apache.org/jira/browse/SOLR-14503 ticket.  I guess I can imagine that SOLR_WAIT_FOR_ZK was translated to `waitForZk`. I DO see it documented in solr.in.* files.

solr/solrj/src/resources/EnvToSyspropMappings.properties
epugh: I didn't see a proposed name in the cwiki page for this, it's just listed as `solr.default.confdir`...   What do you think about `solr.configset.default.confdir`?   Or maybe even `solr.configset.default` with a name, and that figures out the specific location of the confdir under the covers, but not something a user would change?  The key bit is what configset you are using, with the default one being `_default`.

solr/solrj/src/java/org/apache/solr/common/util/EnvUtils.java
epugh: I went with the first part, making the file name clearer... merging both formats into one file made that file look werid, and harder to name...  because it was doing two jobs.

epugh:
@dsmiley okay, I think we are down to coming up with a good name for `SOLR_DEFAULT_CONFDIR`, i proposed two options.  Or save it for antoher PR, and then can I get a LGTM assuming green build?

solr/core/src/test/org/apache/solr/cloud/ZkFailoverTest.java
```diff
@@ -47,7 +47,7 @@ public static void setupCluster() throws Exception {
 
   @AfterClass
   public static void resetWaitForZk() {
-    System.setProperty("solr.wait.for.zk", "30");
+    System.setProperty("solr.tests.cloud.wait.for.zk.seconds", "30");
```
dsmiley: this is wrong now

solr/bin/solr
```diff
@@ -114,7 +114,7 @@ fi
 
 # Export variables we want to make visible to Solr sub-process
 for var in $(compgen -v); do
-  if [[ "$var" =~ ^(SOLR_.*|DEFAULT_CONFDIR|ZK_.*|GCS_BUCKET|GCS_.*|S3_.*|OTEL_.*|AWS_.*)$ ]]; then
+  if [[ "$var" =~ ^(SOLR_.*|ZK_.*|GCS_BUCKET|GCS_.*|S3_.*|OTEL_.*|AWS_.*)$ ]]; then
```
dsmiley, line 117: You probably need to restore this if we're deferring the confdir matter

solr/bin/solr
```diff
@@ -1025,7 +1017,8 @@ elif [[ $SOLR_HOME != /* ]]; then
 fi
 
 # Set the default configset dir to be bootstrapped as _default
-: "${DEFAULT_CONFDIR:="$SOLR_SERVER_DIR/solr/configsets/_default/conf"}"
+: "${SOLR_DEFAULT_CONFDIR:="$SOLR_SERVER_DIR/solr/configsets/_default/conf"}"
+export SOLR_DEFAULT_CONFDIR
```
dsmiley: You probably need to restore this if we're deferring the confdir matter

solr/bin/solr
```diff
@@ -1327,7 +1305,7 @@ function start_solr() {
     # OOME is thrown. Program operation after OOME is unpredictable.
     "-XX:+CrashOnOutOfMemoryError" "-XX:ErrorFile=${SOLR_LOGS_DIR}/jvm_crash_%p.log" \
     "-Djetty.home=$SOLR_SERVER_DIR" "-Dsolr.solr.home=$SOLR_HOME" "-Dsolr.install.dir=$SOLR_TIP" "-Dsolr.install.symDir=$SOLR_TIP_SYM" \
-    "-Dsolr.default.confdir=$DEFAULT_CONFDIR" "${LOG4J_CONFIG[@]}" "${SCRIPT_SOLR_OPTS[@]}" "${SECURITY_MANAGER_OPTS[@]}" "${SOLR_ADMIN_UI}" "${SOLR_OPTS[@]}")
```
dsmiley, line 1330: You probably need to restore this if we're deferring the confdir matter

solr/bin/solr.cmd
```diff
@@ -1144,23 +1120,23 @@ IF NOT EXIST "%SOLR_SERVER_DIR%\tmp" (
   mkdir "%SOLR_SERVER_DIR%\tmp"
 )
 
-IF "%DEFAULT_CONFDIR%"=="" set "DEFAULT_CONFDIR=%SOLR_SERVER_DIR%\solr\configsets\_default\conf"
+IF "%SOLR_DEFAULT_CONFDIR%"=="" set "SOLR_DEFAULT_CONFDIR=%SOLR_SERVER_DIR%\solr\configsets\_default\conf"
```
dsmiley: You probably need to restore this if we're deferring the confdir matter

solr/bin/solr.cmd
```diff
@@ -1144,23 +1120,23 @@ IF NOT EXIST "%SOLR_SERVER_DIR%\tmp" (
...
 
 IF "%FG%"=="1" (
   REM run solr in the foreground
   title "Solr-%SOLR_PORT%"
   echo %SOLR_PORT%>"%SOLR_TIP%"\bin\solr-%SOLR_PORT%.port
   "%JAVA%" %SERVEROPT% %SOLR_JAVA_MEM% %START_OPTS% ^
     -Dlog4j.configurationFile="%LOG4J_CONFIG%" -DSTOP.PORT=!STOP_PORT! -DSTOP.KEY=%STOP_KEY% ^
-    -Dsolr.solr.home="%SOLR_HOME%" -Dsolr.install.dir="%SOLR_TIP%" -Dsolr.install.symDir="%SOLR_TIP%" -Dsolr.default.confdir="%DEFAULT_CONFDIR%" ^
```
dsmiley, line 1155: You probably need to restore this if we're deferring the confdir matter

solr/bin/solr.cmd
```diff
@@ -1144,23 +1120,23 @@ IF NOT EXIST "%SOLR_SERVER_DIR%\tmp" (
...
+    -Dsolr.solr.home="%SOLR_HOME%" -Dsolr.install.dir="%SOLR_TIP%" -Dsolr.install.symDir="%SOLR_TIP%" ^
     -Djetty.port=%SOLR_PORT% -Djetty.home="%SOLR_SERVER_DIR%" ^
     -Djava.io.tmpdir="%SOLR_SERVER_DIR%\tmp" -jar start.jar %SOLR_JETTY_CONFIG% "%SOLR_JETTY_ADDL_CONFIG%"
 ) ELSE (
   START /B "Solr-%SOLR_PORT%" /D "%SOLR_SERVER_DIR%" ^
     "%JAVA%" %SERVEROPT% %SOLR_JAVA_MEM% %START_OPTS% ^
     -Dlog4j.configurationFile="%LOG4J_CONFIG%" -DSTOP.PORT=!STOP_PORT! -DSTOP.KEY=%STOP_KEY% ^
     -Dsolr.log.muteconsole ^
-    -Dsolr.solr.home="%SOLR_HOME%" -Dsolr.install.dir="%SOLR_TIP%" -Dsolr.install.symDir="%SOLR_TIP%" -Dsolr.default.confdir="%DEFAULT_CONFDIR%" ^
```
dsmiley, line 1163: You probably need to restore this if we're deferring the confdir matter

solr/bin/solr.cmd
```diff
@@ -1169,7 +1145,7 @@ IF "%FG%"=="1" (
     set SOLR_START_WAIT=180
   )
   REM now wait to see Solr come online ...
-  "%JAVA%" %SOLR_SSL_OPTS% %AUTHC_OPTS% %SOLR_ZK_CREDS_AND_ACLS% %SOLR_TOOL_OPTS% -Dsolr.install.dir="%SOLR_TIP%" -Dsolr.default.confdir="%DEFAULT_CONFDIR%"^
```
dsmiley, line 1172: You probably need to restore this if we're deferring the confdir matter

solr/solrj/src/resources/EnvToSyspropMappings.properties
dsmiley: `solr.configset.default.confdir` sounds good to me.

epugh:
I think we got a nice bit of clean up in the bash scripts as well!

epugh:
@HoustonPutman I'd love to collaborate on a solid "backwards compatible" test that makes you comfortable about these changes...  What are your thoughts on a strategy?   I'd like to land this PR next week...   so...?

solr/bin/solr
```diff
@@ -1016,10 +1016,6 @@ elif [[ $SOLR_HOME != /* ]]; then
   fi
 fi
 
-# Set the default configset dir to be bootstrapped as _default
-: "${SOLR_DEFAULT_CONFDIR:="$SOLR_SERVER_DIR/solr/configsets/_default/conf"}"
```
dsmiley: how were you able to remove this altogether?  I suppose that most SolrCloud BATS integration test is likely to fail if we don't bootstrap this.  So if they pass, we didn't need this line here after all.

epugh: Turns out that this exact variable declaration is ALSO defined in Java code via ConfigSetService.getDefaultConfigDirPath.

I didn't go exhaustively thorugh all the places we might try to get a default configset dir, relying on the tests...   I have seen other places where we declare stuff in the solr/bin files, but then redeclare them in Java code, so I wasn't totally surprised.

epugh:
@HoustonPutman so this ones for you: #7a701d9366a0e3e302fbf4be38adfe7bc7db0529.  This I think actually does the backwards compatible the way we want.

epugh:
I tweaked the `DeprecatedSystemPropertiesMappings.properties` file to add in the various changed mappings, which gives us back compat between 10 and 9.....   I also highlihgted in the file which should be removed in Solr 11, and put a place holder for Solr 12...     

Going to run the tests, and assuming all greeen (and or just a flaky test from crave ;-) ) going to merge today.     @HoustonPutman thanks for the input on the meetup...    (Also, nice thing about the virtual meetup is seeing folks faces and verbal tones helps me understand issues better than just text...   A +1 or -1 is just a very binary piece of input!)

PR #3314, opened 2025-04-08
SOLR-12831: Clean up ZK nodes after shard deletion is invoked

psalagnac:
https://issues.apache.org/jira/browse/SOLR-12831

This is mostly a port of https://github.com/apache/lucene-solr/pull/1238 which was filed before the Lucene/Solr split.

# Description

Delete left over znodes on zookeeper that aren't removed after a delete shard command is issued.

# Solution

In DeleteShardCmd, use the zk client to issue a recursive delete from the specified root and its children (including that root).

# Tests

Added assertions to `DeleteShardTest` to verify the znodes no longer exist.


solr/core/src/java/org/apache/solr/cloud/api/collections/DeleteShardCmd.java
```diff
@@ -238,4 +239,33 @@ private List<ZkNodeProps> getReplicasForSlice(String collectionName, Slice slice
     }
     return sourceReplicas;
   }
+
+  /**
+   * Best effort to delete Zookeeper nodes that stored other details than the shard itself in
+   * cluster state. If we fail for any reason, we just log and the shard is still deleted.
+   */
+  private void cleanupZooKeeperShardMetadata(String collection, String sliceId)
```
dsmiley, line 247: This approach looks for very specific things, and thus we need to maintain it and may miss things.  Don't we want to delete everything?

psalagnac: Not sure what you mean by _everything_.
Unfortunately, we don't have a per shard root in Zookeeper. I wish we had, it should simpler to remove all info specific to a shard, but the transition to a different data layout will be ever complex.

The naive approach is to remove an hardcoded list of ZK node.

dsmiley: Ah; I forgot what the layout was.  Yes; this is the best we can do.

solr/core/src/test/org/apache/solr/cloud/DeleteShardTest.java
```diff
@@ -125,4 +131,19 @@ private void setSliceState(String collectionName, String shardId, Slice.State st
         collectionName,
         c -> c.getSlice(shardId).getState() == state);
   }
+
+  /** Check whether shard metadata exist in Zookeeper. */
+  private void assertShardMetadata(String collection, String sliceId, boolean shouldExist)
```
dsmiley, line 136: could this be improved to scan the ZK nodes for any ZNode that matches the shard ID?  This way, new/unexpected stuff could be detected.

PR #3317, opened 2025-04-09
Equal terms DocTransformer

dsmiley:
JIRA TODO: https://issues.apache.org/jira/browse/SOLR-XXXXX

This is a sketch of a DocTransformer  that emits a boolean field on the returned doc based on a document field being equal to an input stream, as compared by the analyzed text.
`fl=id,isEqual:[equalterms field=subject value='John Smith']`

I need this for work but looking for feedback on inclusion in Solr and naming and other things.

### DIFF
--- a/solr/core/src/java/org/apache/solr/response/transform/EqualTermsDocTransformer.java
+++ b/solr/core/src/java/org/apache/solr/response/transform/EqualTermsDocTransformer.java
@@ -23,7 +23,7 @@ import java.util.List;
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.analysis.TokenStream;
 import org.apache.lucene.analysis.tokenattributes.CharTermAttribute;
-import org.apache.lucene.document.StoredField;
+import org.apache.lucene.document.Field;
 import org.apache.solr.common.SolrDocument;
 import org.apache.solr.schema.FieldType;
 import org.apache.solr.schema.SchemaField;
@@ -86,8 +86,8 @@ public class EqualTermsDocTransformer extends DocTransformer {
       fieldValue = list.getFirst();
     }
     if (fieldValue instanceof CharSequence) return fieldValue.toString();
-    if (fieldValue instanceof StoredField storedField) {
-      return storedField.stringValue();
+    if (fieldValue instanceof Field luceneField) {
+      return luceneField.stringValue();
     }
     return null;
   }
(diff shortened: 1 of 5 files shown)


solr/core/src/java/org/apache/solr/response/transform/EqualTermsDocTransformer.java
```diff
@@ -0,0 +1,152 @@
...
+
+  @Override
+  public String getName() {
+    return name;
+  }
+
+  @Override
+  public boolean needsSolrIndexSearcher() {
+    return true;
+  }
```
dsmiley, line 60: oops; need to remove.  Claude keeps adding it back LOL

solr/core/src/java/org/apache/solr/response/transform/EqualTermsDocTransformer.java
```diff
@@ -0,0 +1,152 @@
...
+    // Compare terms on-the-fly as we iterate through the token stream
+    // This allows early termination as soon as we find a mismatch
+    boolean isEqual = compareTokensOnTheFly(analyzer, fieldValue);
+    doc.setField(name, isEqual);
+  }
+
+  /**
+   * Gets the string field value, or null if not present or if found a list of values other than 1.
+   */
+  private String getFieldValue(SolrDocument doc) {
```
dsmiley, line 87: I don't love the instance checks here... I wish there was a better typed API contract on the expected types.

solr/core/src/java/org/apache/solr/response/transform/EqualTermsDocTransformer.java
```diff
@@ -0,0 +1,152 @@
...
+      while (tokenStream.incrementToken()) {
+        // Check if we've seen more tokens than in our comparison list
+        if (!compareIter.hasNext()) {
+          return false; // More tokens in source than in comparison value
+        }
+
+        // Compare the current token with the corresponding token in our pre-analyzed list
+        String currentToken = termAttr.toString();
+        String compareToken = compareIter.next();
+        if (!currentToken.equals(compareToken)) {
```
dsmiley, line 120: could be improved to avoid producing a String... maybe there's a utility method to do such a comparison concisely

solr/core/src/test/org/apache/solr/cloud/TestRandomFlRTGCloud.java
```diff
@@ -365,26 +366,45 @@ private SolrInputDocument addRandomDocument(final int docId)
 
     final SolrInputDocument doc =
         sdoc(
-            "id", "" + docId,
```
dsmiley, line 368: the newline placements are unfortunate; I can try to restore

PR #3319, opened 2025-04-11
SOLR-16262: Add canonical links to Solr 6-8 ref guide pages

HoustonPutman:
These canonical links will make sure that Search Engines only point to the most recent version of any given page.

I've also added mappings for some old pages we did not have at 8_11. So now there are no unmapped pages.
This will fix some broken links we have missed through previous versions as well.

Some of these will need better pages to map to in the future, but we can always update these mappings and run the script again.

https://issues.apache.org/jira/browse/SOLR-16262


HoustonPutman:
I'm going to go and run this for 6.x and 7.x ref guide pages tomorrow, and see if I stop seeing those in google results.

risdenk (approved):

dev-tools/scripts/refguide/refguide-add-canonical-url.py
```diff
@@ -0,0 +1,189 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+This script processes all static html files for Solr's refernce guide
+and downloads external JS and CSS files to local folders js/ and css/ for
+each version. It also updates the HTML files to reference the local files.
+Context is that ASF policy for web sites changed to not allow external
+references to JS and CSS files, and these sites were generated long ago.
```
janhoy, line 23: Update this text :)

HoustonPutman: done!

PR #3322, opened 2025-04-13
SOLR-17740: V2 API: fix raw file uploads missing 1st byte

dsmiley:
SolrRequestParsers can consume one byte of the underlying stream, but V2 was ignoring the SolrQueryRequest content stream.

https://issues.apache.org/jira/browse/SOLR-17740

I wish there was a consistently failing test for this but on the Jetty upgrade, `PackageToolTest.testPackageTool` fails the majority of the time.  A ~unit test might be nice if I get around to it.


dsmiley:
Thank you for prodding me on the reasoning behind the PushbackInputStream being there in the first place.  After looking deeper, it was added back in SOLR-15418 (me to fix another bug for v2), I'm happy to report I can remove that, and all tests pass now.  In the last 4 years, perhaps some detail changed about v2 pertaining to why I chose that unfortunate solution then.  🎉

gerlowskija (approved):
LGTM

PR #3323, opened 2025-04-13
Optimize consumeInputFully, don't loop read single byte

dsmiley:
Do not read a single byte in a loop -- a performance anti-pattern.

I just happened to notice this during a debugging session that had me looking closely at single-byte reads.

### DIFF
--- a/solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
+++ b/solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
@@ -19,7 +19,6 @@ package org.apache.solr.servlet;
 
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.context.Context;
-import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -41,6 +40,7 @@ import javax.servlet.http.HttpServletResponseWrapper;
 import org.apache.http.HttpHeaders;
 import org.apache.solr.common.SolrException;
 import org.apache.solr.common.SolrException.ErrorCode;
+import org.apache.solr.common.util.Utils;
 import org.apache.solr.logging.MDCLoggingContext;
 import org.apache.solr.util.tracing.TraceUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/servlet/ServletUtils.java
```diff
@@ -252,8 +253,11 @@ private static void traceHttpRequestExecution2(
   static void consumeInputFully(HttpServletRequest req, HttpServletResponse response) {
     try {
       ServletInputStream is = req.getInputStream();
-      //noinspection StatementWithEmptyBody
-      while (!is.isFinished() && is.read() != -1) {}
+      if (!is.isFinished() && is.read() != -1) {
```
dsmiley, line 256: even the read here isn't needed, technically, but it's super common for an InputStream to not be finished yet the read is EOF.  skipNBytes allocates a 2k array so we can avoid that typically.

risdenk: We have `Utils.readFully(InputStream is)` would that usable here?

dsmiley: Ah yes, thanks!
Feedback addressed.

solr/solrj/src/java/org/apache/solr/common/util/Utils.java
```diff
@@ -664,8 +665,13 @@ public static void consumeFully(HttpEntity entity) {
    * @throws IOException on problem with IO
    */
   public static void readFully(InputStream is) throws IOException {
-    is.skip(is.available());
-    while (is.read() != -1) {}
+    if (is.read() != -1) { // not needed but avoids a buffer allocation soon if found data
```
madrob, line 668: I think this comment missing a word?

dsmiley: Fixed

dsmiley:
didn't take to 9x because skipNBytes doesn't exist in JDK 11

markrmiller (commented):
Nice one. Looks good.

PR #3324, opened 2025-04-14
SOLR-17069 HttpServletRequest; use eagerly to avoid inaccessibility

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17069

[Justification](https://github.com/apache/solr/pull/2876#issuecomment-2782169134)

This was split out from #2876.

A related item https://issues.apache.org/jira/browse/SOLR-17741 "Remove addHttpRequestToContext option" but that's more about removing needless option than actually addressing a hazard/bug.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CancellableQueryTracker.java
+++ b/solr/core/src/java/org/apache/solr/core/CancellableQueryTracker.java
@@ -56,7 +56,7 @@ public class CancellableQueryTracker {
       }
     }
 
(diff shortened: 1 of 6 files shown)


dsmiley:
I'll merge this tonight EST.  Seems fast I know but the code has been sitting on the other PR for a bit.
I could add a CHANGES.txt or we just embellish the Jetty upgrade one, or don't really speak of it as the changes here are so internal that a reader of CHANGES.txt wouldn't find the particulars of this PR worth mentioning.  There's always the commit message anyway :-)

PR #3325, opened 2025-04-14
SOLR-17741: Remove addHttpRequestToContext option

dsmiley:
in solrconfig.xml/requestDispatcher/requestParsers. Why: Obsolete/redundant with SolrQueryRequest.getHttpSolrCall().getReq()

https://issues.apache.org/jira/browse/SOLR-17741

### DIFF
--- a/solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-10.adoc
+++ b/solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-10.adoc
@@ -111,6 +111,8 @@ Users who that don't need to vary JAR access on a per-core basis have several op
 
 * CurrencyField has been removed.  Users should migrate to the `CurrencyFieldType` implementation.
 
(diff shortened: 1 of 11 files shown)


solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-10.adoc
```diff
@@ -111,6 +111,8 @@ Users who that don't need to vary JAR access on a per-core basis have several op
 
 * CurrencyField has been removed.  Users should migrate to the `CurrencyFieldType` implementation.
 
+* The `addHttpRequestToContext` option in `solrconfig.xml` has been removed; it's obsolete.
```
gerlowskija, line 114: [0] Might be worth augmenting these docs so that users know they can still get at the request object via `SolrQueryRequest.getHttpSolrCall().getReq()`

gerlowskija (approved):
Left one minor suggestion, but otherwise LGTM.

Will you do a branch_9x PR to deprecate things on that end?

PR #3326, opened 2025-04-14
SOLR-17743: V2 API: use correct content-type

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17743

I welcome input on testing (if needed).

### DIFF
--- a/solr/solrj/src/resources/java-template/api.mustache
+++ b/solr/solrj/src/resources/java-template/api.mustache
@@ -204,6 +204,7 @@ public class {{classname}} {
             @Override
             @SuppressWarnings("unchecked")
             public RequestWriter.ContentWriter getContentWriter(String _expectedTypeIGNORE) {
+                // v2 only supports JSON request bodies, so we ignore this type coming from the RequestWriter
                 {{#vendorExtensions.x-genericEntity}}
                 return new RequestWriter.ContentWriter() {
                   @Override
--- a/solr/solrj/src/resources/java-template/api.mustache
+++ b/solr/solrj/src/resources/java-template/api.mustache
@@ -133,7 +133,6 @@ public class {{classname}} {
                 {{#isBodyParam}}
                 {{#vendorExtensions.x-genericEntity}}
                     this.requestBody = requestBody;
-                    addHeader("Content-type", "application/json");
                 {{/vendorExtensions.x-genericEntity}}
                 {{/isBodyParam}}
                 {{^isBodyParam}}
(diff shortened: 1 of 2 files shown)


solr/solrj/src/resources/java-template/api.mustache
```diff
@@ -133,7 +133,6 @@ public class {{classname}} {
                 {{#isBodyParam}}
                 {{#vendorExtensions.x-genericEntity}}
                     this.requestBody = requestBody;
-                    addHeader("Content-type", "application/json");
```
dsmiley, line 136: This doesn't belong here.  It's the ContentWriter that's responsible for the type.

solr/solrj/src/resources/java-template/api.mustache
```diff
@@ -205,26 +203,23 @@ public class {{classname}} {
 
             @Override
             @SuppressWarnings("unchecked")
-            public RequestWriter.ContentWriter getContentWriter(String expectedType) {
+            public RequestWriter.ContentWriter getContentWriter(String _expectedTypeIGNORE) {
```
dsmiley, line 206: `getContentWriter` is passed an "expectedType" from the `SolrClient.requestWriter` which defaults to "javabin".  But V2 should ignore that requestWriter as it's opinionated on how to write -- only supports JSON via Jackson or raw writing.  Perhaps that field in SolrClient should be named `defaultRequestWriter`... although we're not choosing a different RequestWriter so.... :shrug:

dsmiley (commented):
Another thing to potentially fix is I observe that a request to GET a raw file has `wt=json` but we're not actually writing with that after all.

solr/solrj/src/resources/java-template/api.mustache
gerlowskija: I do like that better from an abstraction perspective, but I remember putting this line here only after noticing that relying on the ContentWriter was leaving the header unset in some cases.  I wish I remembered the context there better.

But let's not worry about that now - if it crops up again I'll reinvestigate at that point...

solr/solrj/src/resources/java-template/api.mustache
gerlowskija: [0] Maybe some version of this rationale should live as a comment in the code/template, so that we can notice when those preconditions/assumptions change?

e.g.

> // v2 currently only supports reading JSON request-bodies, so we ignore the expected type suggestion provided by callers.

gerlowskija (approved):
LGTM - though I left a few comments you can address or ignore as you choose!

PR #3328, opened 2025-04-16
[SOLR-17726] Fix CloudMLTQParser to support copyField in qf

ilariapet:
https://issues.apache.org/jira/browse/SOLR-17726

# Description
When using CloudMLTQParser, fields that are populated via `copyField` are not taken into account when constructing the MoreLikeThis query.

This happens because CloudMLTQParser relies on a RealTime Get (`/get`) request to retrieve the source document by ID, and the document returned by RealTime Get does not include fields generated via copyField (i.e. are not part of the original SolrIDocument).

As a result, even if the copyField target is stored and has proper termVectors configured, CloudMLTQParser skips the field silently, and the MLT query ends up empty.

# Solution
The issue was addressed by introducing a helper method `getFieldValuesIncludingCopyField` that retrieves field values from the source document returned by RealTime Get.

If a field passed in the `qf` parameter of the MLT query is not directly present in the document (as is the case for fields populated exclusively via copyField), the method looks up its copyField sources in the schema and aggregates their values. These values are then used to construct the input for MoreLikeThis#like, ensuring that all content (including that of copyField-derived fields) is properly considered when building the MLT query.

# Tests

Five tests were added to cover different scenarios involving copyField usage in the qf parameter of the MoreLikeThis query:

- `testUsesACopyFieldInQf_shouldReturnExpectResults`
Verifies that when a copyField destination is used in the qf parameter, the field values are correctly retrieved from the source field(s) and the MLT query returns the expected results.
- `testUsesACopyFieldInQf_shouldGenerateNonEmptyQuery`
Verifies that the MLT query correctly uses the content of the source field(s) when a copyField destination is specified in the qf parameter.
- `testCopyFieldSourceMissing_shouldReturnNoResults`
Ensures that no results are returned when the copyField source field is missing in the source document.
- `testCopyFieldDestinMultipleSources_shouldReturnExpectResults`
Validates that when multiple source fields map to a single copyField destination, their values are correctly combined and expected results are returned.
- `testCopyFieldDestinationMultipleSources_shouldGenerateQueryUsingMultipleSourcesValues`
Validates that when multiple source fields map to a single copyField destination, their values are correctly combined and the resulting MLT query is properly constructed.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/mlt/CloudMLTQParser.java
+++ b/solr/core/src/java/org/apache/solr/search/mlt/CloudMLTQParser.java
@@ -115,7 +115,6 @@ public class CloudMLTQParser extends SimpleMLTQParser {
   private Collection<Object> getFieldValuesIncludingCopyField(SolrDocument doc, String field) {
     Collection<Object> fieldValues = doc.getFieldValues(field);
     if (fieldValues != null) return fieldValues;
-
     // Fields created using copyField are not included in documents returned by RealTime Get.
     // So if a copyField destination is used in the MLT query (qf), we need to get the values
     // from its source field instead. If there are multiple source fields, their values must be


epugh (commented):
This all makes sense...    Two things:
1) Instead of working around how RTG works, would it make sense to have RTG return copyField values instead and not need this?   I don't know, there could be good technical reasons why RTG does it what it does.... 

2) I love the descriptive text for the tests that are in your PR.  It would be nice to preserve that as a comment in the actual `CloudMLTQParserTest.java` file for the next person!  I sometimes struggle to understand why a test exists..

epugh:
Love the description.   It *could* also be a javadoc right?  On the method?  But now I am just nit picking!

ilariapet:
> This all makes sense... Two things:
> 
> 1. Instead of working around how RTG works, would it make sense to have RTG return copyField values instead and not need this?   I don't know, there could be good technical reasons why RTG does it what it does....
> 2. I love the descriptive text for the tests that are in your PR.  It would be nice to preserve that as a comment in the actual `CloudMLTQParserTest.java` file for the next person!  I sometimes struggle to understand why a test exists..

Thanks a lot for the feedback!

Regarding the first point, according to the documentation, the goal of RealTime Get (RTG) is to return the latest version of a document. Maybe the reason why `copyField` values aren’t included is to avoid duplicating existing data... including them would just introduce redundancy.
That said, we could consider modifying RTG by introducing a parameter like `includeCopyFields` (defaulting to false) to preserve the current behavior while still offering flexibility when those fields are actually needed.
Let’s see what others think as well.

solr/core/src/test/org/apache/solr/search/mlt/CloudMLTQParserTest.java
```diff
@@ -339,4 +345,106 @@ public void testInvalidSourceDocument() {
...
+  public void testCopyFieldSourceMissing_shouldReturnNoResults() throws Exception {
+    // Ensures that no results are returned when the copyField source field is missing in the source
+    // document.
+    QueryResponse queryResponse =
+        cluster
+            .getSolrClient()
+            .query(COLLECTION, new SolrQuery("{!mlt qf=copyfield_dest mindf=0 mintf=1}30"));
+    SolrDocumentList solrDocuments = queryResponse.getResults();
+    assertEquals("Expected no results if source field is missing", 0, solrDocuments.size());
+  }
```
alessandrobenedetti, line 400: Does this mean we have no value for the source and no value for the copyfield?

ilariapet: yes

PR #3330, opened 2025-04-17
SOLR-17745: Reduce leader election optimization never occurs

mlbiscoc:
Co-authored-by: Luke Kot-Zaniewski <l.kotzaniewski@gmail.com>

https://issues.apache.org/jira/browse/SOLR-17745

# Description

Credit to @kotman12 for helping find this

It appears the optimization in [SOLR-14942](https://issues.apache.org/jira/browse/SOLR-14942) never actually happens. The reason is the seemingly innocuous check of [zkController.isClosed](https://github.com/apache/solr/blob/57de0e8dece2f333a624b6498d139af31d00038c/solr/core/src/java/org/apache/solr/cloud/ZkController.java#L2154) which actually ends up always skipping the leader election node cleanup. This is not because the ZK session is actually closed but rather a side-effect of the complex implementation of [SolrZkClient::isClosed](https://github.com/apache/solr/blob/d1498e8d5a7f4382a83a2800a0d1568bcd0f2a55/solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/SolrZkClient.java#L852-L854) which also checks its "higher level" for closure. The higherLevelIsClosed delegates to [CoreContainer::isShutDown](https://github.com/apache/solr/blob/6d84e8f97611685df7b39278737ca7bc9fc326d8/solr/core/src/java/org/apache/solr/cloud/ZkController.java#L355) which actually gets set to true in an [earlier point in the shutdown sequence](https://github.com/apache/solr/blob/6d84e8f97611685df7b39278737ca7bc9fc326d8/solr/core/src/java/org/apache/solr/core/CoreContainer.java#L1299-L1308). Thus this optimization never materializes.

# Solution

Check SolrZkClient::isConnected instead of isClosed in ZKController. Also added a warn level log inside the check.

# Tests

Didn't think a test was necessary for this, but we're happy to add one if the community feels it's warranted

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/ZkController.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ZkController.java
@@ -2151,7 +2151,8 @@ public class ZkController implements Closeable {
 
   /** Attempts to cancel all leader elections. This method should be called on node shutdown. */
   public void tryCancelAllElections() {
-    if (zkClient.isClosed()) {
+    if (!zkClient.isConnected()) {
+      log.warn("Skipping leader election node cleanup since we're disconnected from ZooKeeper.");
       return;
     }
     Collection<ElectionContext> values = electionContexts.values();


psalagnac:
That's a great analysis. Thanks @mlbiscoc and @kotman12.

> Didn't think a test was necessary for this, but we're happy to add one if the community feels it's warranted

That's very unfortunate that a broken feature was initially merged.
The very complex shutdown sequence somehow led to this broken change. Adding a test would make sure there is no future regression.
Saying this, I have no idea how hard it would be to test this. Having a test is not a requirement but a nice to have.

mlbiscoc:
> That's a great analysis. Thanks @mlbiscoc and @kotman12.
> 
> > Didn't think a test was necessary for this, but we're happy to add one if the community feels it's warranted
> 
> That's very unfortunate that a broken feature was initially merged. The very complex shutdown sequence somehow led to this broken change. Adding a test would make sure there is no future regression. Saying this, I have no idea how hard it would be to test this. Having a test is not a requirement but a nice to have.

Thanks @psalagnac! I can look into whats possible for a test. I'll try and get it to fail on main and see if this PR it passes. We would need maybe an integration test with cc.shutdown() and check that the leader election znode in zookeeper goes away during shutdown. We see this bug get really bad during heavy ingestion and the container get stuck to shutdown until we hit.
```
Timed out waiting for in-flight update requests to complete for core:
```

kotman12:
> > That's a great analysis. Thanks @mlbiscoc and @kotman12.
> > > Didn't think a test was necessary for this, but we're happy to add one if the community feels it's warranted
> > 
> > 
> > That's very unfortunate that a broken feature was initially merged. The very complex shutdown sequence somehow led to this broken change. Adding a test would make sure there is no future regression. Saying this, I have no idea how hard it would be to test this. Having a test is not a requirement but a nice to have.
> 
> Thanks @psalagnac! I can look into whats possible for a test. I'll try and get it to fail on main and see if this PR it passes. We would need maybe an integration test with cc.shutdown() and check that the leader election znode in zookeeper goes away during shutdown. We see this bug get really bad during heavy ingestion and the container get stuck to shutdown until we hit.
> 
> ```
> Timed out waiting for in-flight update requests to complete for core:
> ```

This might be challenging because I suspect that the zkSys.close which happens later on deletes the leader election node as well. So to truly trigger this condition you'd have to cut the shutdown process short or suspend it to make sure that the optimized leader node removal is the only thing that can delete the leader election node.

kotman12:
@mlbiscoc @psalagnac How about something like [this](https://github.com/kotman12/solr/commit/aaa7564cb049bcedf5472617aac1681c3fe31c2b) to test this? I don't _love_ it because it depends on an implementation detail of `CoreContainer::shutdown`, namely that the optimized election node deletion occurs _before_ the closing of caches. However, this fix addresses implementation details of that very method so I don't see how you can avoid relying on _some_ detail while also having a deterministic test.

mlbiscoc:
> @mlbiscoc @psalagnac How about something like [this](https://github.com/kotman12/solr/commit/aaa7564cb049bcedf5472617aac1681c3fe31c2b) to test this? I don't _love_ it because it depends on an implementation detail of `CoreContainer::shutdown`, namely that the optimized election node deletion occurs _before_ the closing of caches. However, this fix addresses implementation details of that very method so I don't see how you can avoid relying on _some_ detail while also having a deterministic test.

Ah this is pretty clever. Although I agree, I am not a big fan of this test being fixed in that closing of caches must happen after otherwise this whole test fails. Might confuse someone thinking something like this was intentional even though it is kind of a hack to test something not entirely relating to it. If we go with this approach, it should be commented in at the minimum this is the reason so if it ever fails due to the closing of caches moving, someone can at least understand why.

I was also curious on looking into if Mockito.spy was possible and verify the calls for zkclient.delete() actually happening. Not sure how possible this is though with lots of these objects being nested deep from coreContainer and some being private/protected we would need to inject the spy one way or another. Maybe through java reflection?

kotman12:
@psalagnac understood. @mlbiscoc had an idea of potentially using mockito spy for capturing the delete event although that has the drawback of all the pains of bytecode manipulation. The benefit is that Solr already uses this in a few spots.

The other thought I had was using [log-captor](https://github.com/Hakky54/log-captor) which is a cool little apache-licensed lib that can have other uses but maybe a bit heavyweight to bring in for one test. We now log when this optimization _doesn't_ occur so capturing the logs and validating this line never gets printed might be an ok approach.

Otherwise we are fine with no test.

kotman12:
BTW just saw your comment @mlbiscoc ... I think you can try wrapping the client exposed by `cluster.getZkClient()` from within `SolrCloudTestCase` in a spy... I think that might be the instance that calls delete on the leader elect node. Basically not sure this would have to be _that_ difficult re this:

> Not sure how possible this is though with lots of these objects being nested deep from coreContainer and some being private/protected we would need to inject the spy one way or another.

Although I admit I'm not 100% this will achieve the intended effect.

mlbiscoc:
Went and added CHANGE.txt. Not a mockito expert but couldn't get a way to verify the delete calls on zkclient because I couldn't inject it inside of electionContexts as it is a private final attribute with no getters.

psalagnac (approved):

PR #3331, opened 2025-04-17
SOLR-17717 Use Powershell instead of WMIC to get cmdline on Windows (backport 9x)

janhoy:
(cherry picked from commit 6315d3fe2ca46fe7a30a33f4b8deea906d7090c0)

https://issues.apache.org/jira/browse/SOLR-17717


janhoy:
Note, @HoustonPutman moved changes entry in main to 9.10. Should do that in this PR as well and then merge.

PR #3334, opened 2025-04-22
SOLR-16951: Add PKI Auth Caching for both generation and validation

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-16951

- Default PKI TTL has been increased to 10 seconds
- The PKI Generation cache will expire cached tokens after 1/4 of the TTL time, to give a minimum 3/4 of the TTL time for transportation.
- The PKI Generation cache will start refreshing cached tokens 1/2 of the expiration time (1/8 of the TTL time), to give time for the asynchronous refresh to complete before the token is expired. I would set this to `expirationTime - 50ms`, but CaffeineCache requires a `get()` in order to trigger the async refresh, so we want to give time between `refreshTime` and `expireTime` for a request to come in to trigger the refresh. 1/2 of the expiration time should be fine here.
- I don't expect to get more than 100 users that often, so I think this is a safe cache limit for generation. For validation, the node name is included in the token, so we need an order of magnitude bigger cache. I chose 1000 here, but this can probably be changed if we need to. The validation cache isn't as important as the generation cache, so it shouldn't be a huge deal.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -229,6 +229,9 @@ Improvements
 
 * SOLR-17732: Score-based return fields other than "score" can now be returned in distributed queries. (Houston Putman)
 
+* SOLR-16951: Add PKI Auth Caching for both generation of the PKI Auth Tokens and validation of received tokens.
+  The default PKI Auth validity TTL has been increased from 5 seconds to 10 seconds. (Houston Putman)
+
 Optimizations
 ---------------------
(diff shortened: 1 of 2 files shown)


anshumg (approved):
Thanks @HoustonPutman. This looks good.

tflobbe (approved):
LGTM!

gerlowskija (approved):
LGTM!

PR #3336, opened 2025-04-22
SOLR-17634: Remove version param

dsmiley:
Not used in forever; XML format is stable.  Today would more likely use a user-agent Solr version.

https://issues.apache.org/jira/browse/SOLR-17634


solr/solrj/src/java/org/apache/solr/client/solrj/ResponseParser.java
```diff
@@ -41,11 +41,4 @@ public abstract NamedList<Object> processResponse(InputStream body, String encod
    *     no enforcement.
    */
   public abstract Collection<String> getContentTypes();
-
-  /**
-   * @return the version param passed to solr
-   */
-  public String getVersion() {
```
dsmiley, line 48: The primary change is this method going away

solr/core/src/test/org/apache/solr/servlet/DirectSolrConnectionTest.java
```diff
@@ -22,6 +22,7 @@
 import org.apache.solr.common.params.CommonParams;
 import org.junit.BeforeClass;
 
+@Deprecated
```
gerlowskija, line 25: [Q] Why are we deprecating this?  Is this related to the "version" removal?

dsmiley: Totally unrelated to this PR; sorry.  DirectSolrConnection is deprecated so I thought deprecating its test to would be nice even though totally unnecessary.

PR #3341, opened 2025-05-01
SOLR-17750: Let S3 File downloads gracefully handle connection issues

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17750

Maybe ResumableInputStream will be useful down the road, so I put it in SolrJ


markrmiller:
I've got a review for this ill post tomorrow.

markrmiller (commented):
Left a few notes.

solr/solrj/src/java/org/apache/solr/common/util/ResumableInputStream.java
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  public int read(boolean isRetry) throws IOException {
+    if (delegate == null) {
+      delegate = nextInputStreamSupplier.apply(bytesRead, contentLength);
+    }
+    int val;
+    try {
+      val = delegate.read();
+      bytesRead += 1;
```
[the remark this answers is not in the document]
HoustonPutman: Good catch

solr/modules/s3-repository/src/java/org/apache/solr/s3/S3StorageClient.java
```diff
@@ -374,8 +377,23 @@ InputStream pullStream(String path) throws S3Exception {
...
+          getBuilder.build(),
+          ResponseTransformer.unmanaged(
+              (response, inputStream) ->
+                  new ResumableInputStream(
+                      inputStream,
+                      response.contentLength(),
+                      (bytesRead, contentLength) -> {
+                        if (bytesRead > 0) {
+                          getBuilder.range(
+                              String.format(Locale.ROOT, "bytes=%d-%d", bytesRead, contentLength));
```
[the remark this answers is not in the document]
HoustonPutman: So it's exclusive on the ending, so even though it's 0 based it should be safe. I checked other usages of this, and it's consistent with what I've done.

Oh I didn't know `"%s-"` was an option. If so, then yeah, no need to use Content Length either...

solr/solrj/src/java/org/apache/solr/common/util/ResumableInputStream.java
```diff
@@ -84,6 +83,34 @@ public int read(boolean isRetry) throws IOException {
...
+  }
+
+  public int read(byte[] b, int off, int len, boolean isRetry) throws IOException {
+    if (delegate == null) {
+      delegate = nextInputStreamSupplier.apply(bytesRead);
+    }
+    int readLen;
+    try {
+      readLen = delegate.read(b, off, len);
+      if (len >= 0) {
```
markrmiller, line 95: should be readLen rather than len?

markrmiller (approved):
LGTM other than one comment I left.

Which prompts one nice to have: would be good to make sure a test hit's the bulk read method well.

markrmiller:
One more tiny one: Might be nice to note that skip() / transferTo() are not retry aware in the javadoc.

solr/solrj/src/java/org/apache/solr/common/util/ResumableInputStream.java
HoustonPutman: Yep, dumb miss.

And unfortunately this is tested:
```java
// test both read() and read(buffer, off, len)
if (random().nextBoolean()) {
  done = input.read() == -1;
  if (!done) {
    byteCount++;
  }
} else {
  int readLen = input.read(buffer, 0, random().nextInt(maxBuffer) + 1);
  if (readLen > 0) {
    byteCount += readLen;
  } else {
    // We are done when readLen = -1
    done = true;
  }
}
```

HoustonPutman: Ahh the reason it's not a big deal is that we never call read() after getting a -1, so the `byteCount` isn't really used. So it's wrong, but our test (and most usages of InputStream) didn't see any issue, because we follow the contract of InputStream.

HoustonPutman:
> One more tiny one: Might be nice to note that skip() / transferTo() are not retry aware in the javadoc.

Good call on `skip()`, I've added an implementation for that, mentioning the limitation.

As for `transferTo()`, that uses `read(buffer, 0, DEFAULT_BUFFER_SIZE)` under the hood in `InputStream`, so I think it should be supported here. It might not be optimized with `delegate`, but there really isn't a way for us to do it another way.

PR #2685, opened 2024-08-30
SmokeTestRelease Added support for multiple JDK

iamsanjay:
This PR removes the hardcoded reference to Java and introduces support for multiple JDK versions through the new `--test-alt-java` parameter. This enhancement was largely inspired by the work in [Apache Lucene PR #13108](https://github.com/apache/lucene/pull/13108).

Credits to the contributors of the original implementation for their valuable insights.

PR #3342, opened 2025-05-06
SmokeTestRelease Added support for multiple JDK (backport 9x)

janhoy:
This is a backport attempt for the generalization of smokeTestRelease from main (#2685). Did not merge since there are still important differences between main and 9x that the script must preserve/respect. One such is java versions, another is CLI syntax for invoking Solr examples etc.

(cherry picked from commit 98aa72398a095920e89e8c86d5f8c6124afa94e5)

### DIFF
--- a/dev-tools/scripts/README.md
+++ b/dev-tools/scripts/README.md
@@ -15,37 +15,30 @@ Used to validate new release candidates (RC). The script downloads an RC from a 
 or local folder, then runs a number of sanity checks on the artifacts, and then runs
 the full tests.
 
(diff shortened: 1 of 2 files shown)


janhoy:
Time for a review @iamsanjay , @HoustonPutman ?

HoustonPutman:
In general, this looks great to me, but I haven't run it locally

janhoy:
I checked jenkins and wizard, and neither use the `--test-java17` option, so this should be safe.

PR #3344, opened 2025-05-08
SOLR-17756: Parallelize calculation of index fingerprint across segments

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-17756

# Description

The index fingerprint is currently being calculated on each segment sequentially. While this works fine, the index fingerprint calculation was noticed to be a very slow process and on leader election is blocking.

This proposes to have this calculation parallelized across segments instead. Since the fingerprint is just a cumulative sum of a hash on versions, the order in which it is added to the running sum should not matter.

# Solution

Create a dedicated threadpool (`IndexFingerprintPool`) to calculate the indexfingerprint. Created a separate theadpool instead of the common `ForkJoinPool` as this can be an expensive operation calculating across a large index

# Tests

Created `testSequentialVsParallelFingerprint` test to confirm the fingerprint is the same.

# Checklist


solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2546,24 +2554,15 @@ public Explanation explain(Query query, int doc) throws IOException {
...
-    }
+    List<Callable<IndexFingerprint>> tasks =
+        searcher.getTopReaderContext().leaves().stream()
+            .map(
+                ctx ->
+                    (Callable<IndexFingerprint>)
+                        () -> searcher.getCore().getIndexFingerprint(searcher, ctx, maxVersion))
+            .collect(Collectors.toList());
+    return ExecutorUtil.submitAllAndAwaitAggregatingExceptions(fingerprintExecutor, tasks).stream()
+        .reduce(new IndexFingerprint(maxVersion), IndexFingerprint::reduce);
```
mlbiscoc, line 2565: We are only going to parallelize the fingerprint calculation. The reduce is done sequentially still after.

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -2204,6 +2204,11 @@ public IndexFingerprint getIndexFingerprint(
     return f;
   }
 
+  /** Clear the per‐segment fingerprint cache. */
+  public void clearPerSegmentFingerprintCache() {
+    perSegmentFingerprintCache.clear();
+  }
```
mlbiscoc, line 2210: I added this in for the test so the sequential calculation doesn't affect the next with its cached hash

mlbiscoc (commented):
@gerlowskija Going to tag you in this since you volunteered on the mailing list :)

Test pass locally but going to confirm with nightly locally as well.

solr/core/src/java/org/apache/solr/core/SolrCore.java
kotman12: what do you think of [using a cache-less reader](https://github.com/mlbiscoc/solr/pull/3) for the same effect? Or at least add a `@VisibleForTesting` annotation?

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
kotman12: Makes sense. The reduce here is `O(segment_count)` .. in our testing it was less than 1% of total execution time. Forking this process again could even slow things down in some cases and any marginal benefit is certainly not worth adding a fork join pool dependency.

solr/core/src/java/org/apache/solr/core/SolrCore.java
mlbiscoc: Thanks @kotman12. Pulled in change as I actually do prefer it over having this `clearPerSegmentFingerprintCache` method in core.

solr/core/src/test/org/apache/solr/update/SolrIndexFingerprintTest.java
```diff
@@ -0,0 +1,88 @@
...
+    assertU(adoc("id", "105"));
+    assertU(adoc("id", "106"));
+    assertU(commit());
+
+    assertU(adoc("id", "107"));
+    assertU(adoc("id", "108"));
+    assertU(adoc("id", "109"));
+    assertU(commit());
+
+    var searcher = core.getSearcher().get();
```
kotman12, line 55: could you try-with-resources pattern here so the searcher always gets closed, even if the test unexpectedly throws?

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -449,6 +456,10 @@ public CoreContainer(NodeConfig config, CoresLocator locator, boolean asyncSolrC
     this.allowListUrlChecker = AllowListUrlChecker.create(config);
 
     this.indexSearcherExecutor = SolrIndexSearcher.initCollectorExecutor(cfg);
+
+    this.indexFingerprintExecutor =
+        ExecutorUtil.newMDCAwareFixedThreadPool(
```
psalagnac, line 461: Can't we have lazily initialized threads instead for memory saving? My understanding is they will no be used most of the time.

This can be achieved with
```
 ExecutorUtil.newMDCAwareCachedThreadPool(
            EXECUTOR_MAX_CPU_THREADS, Integer.MAX_VALUE, ...
```

mlbiscoc:
Closing this PR because I messed up the diff pulling in main incorrectly. Instead of trying to fix it, just remaking the PR [here](https://github.com/apache/solr/pull/3352) instead

PR #3345, opened 2025-05-10
SOLR-17704: Refactor and use curator for distributed locks

heythm:
https://issues.apache.org/jira/browse/SOLR-17704

# Description

This PR refactors Solr's distributed lock implementation to use Curator's ZK primitives for all ZooKeeper operations, while preserving the legacy Solr lock semantics. The new implementation, CuratorDistributedLocks, replaces the previous use of Curator's high-level lock recipes and closely matches the behavior of the old ZkDistributedLock, ensuring that read locks are only blocked by lower-numbered write locks and write locks are blocked by any lower-numbered lock. The lock acquisition strategy is documented and explained in the code.

# Solution

- Introduced a new CuratorDistributedLocks class that uses Curator's CuratorFramework for all ZK operations (node creation, deletion, children listing, and watchers).
- The lock acquisition logic matches the legacy Solr semantics:
  - Read locks only block on lower-numbered write locks. Write locks block on any lower-numbered lock.
  - The implementation does not sort children or only watch the immediate predecessor, but instead watches the first blocking node found, which may cause more wakeups but is simpler and matches Solr's historical behavior.
- Updated the lock factories to use this new implementation and removed the old nested lock logic
-  Improved documentation and code comments to clarify the design and trade-offs.

# Tests

- Existing distributed lock tests (e.g., CuratorDistributedLockFactoryTest) were run and pass, confirming that the new implementation preserves the expected semantics.
- Manual verification that lock acquisition and release behave as before, including correct exception handling when calling isAcquired() after release().
- No regressions observed in distributed locking behavior.

# Checklist


solr/core/src/java/org/apache/solr/cloud/CuratorDistributedCollectionLockFactory.java
```diff
@@ -18,24 +18,44 @@
...
 import org.apache.solr.common.SolrException;
-import org.apache.solr.common.cloud.SolrZkClient;
 import org.apache.solr.common.params.CollectionParams;
 
 /**
- * A distributed lock implementation using Zookeeper "directory" nodes created within a collection
- * znode hierarchy, for use with the distributed Collection API implementation. The locks are
- * implemented using ephemeral nodes placed below the "directory" nodes.
+ * A distributed lock factory for Solr collections, shards, and replicas, using Apache Curator's
+ * {@link org.apache.curator.framework.recipes.locks.InterProcessReadWriteLock} to manage
```
dsmiley: I don't see the usage of InterProcessReadWriteLock.

dsmiley (commented):
Thanks for contributing, Haythem!

I noticed that you renamed the classes to replace "Zk" with "Curator".  But Curator is more of an implementation choice.  Fundamentally, the locks are *in* ZooKeeper; the existing names made sense.  As Solr uses Curator more and more, I don't think we need to follow this renaming pattern of Curator showing up in class names.

Overall, this PR isn't what I expected.  I expected to see usage of Curator **recipes** that would thus save us from maintaining lower level code (whether it be ZK level of Curator "framework" level).  But I don't see that _at all_; although a comment hints at it but I don't see it in the code.  I just see some basic swap-outs of "framework" level Curator which is basically 1:1 to ZooKeeper, with barely any value-add.  Consequently, there isn't much point to this PR as it stands.  I take responsibility for not underscoring the intent in the vague JIRA description.

The PR description says a lot; it led me to believe that there is much more going on here than there is.  It said things that were probably true previously.  I suspect an LLM wrote it; and the LLM didn't recognize the refactoring nature here.  Even if my hunch is correct, you are always responsible for the LLM's output, be it code, PRs, or documentation _as if you wrote it yourself_.  If you had included an LLM generated disclaimer, with a horizontal line separating your commentary from the LLM, then that would help readers know the trustworthiness of the parts of the PR text.

PR #3346, opened 2025-05-10
SOLR-17370 - Fix facet IllegalAccessException for string fields with numeric values

heythm:
https://issues.apache.org/jira/browse/SOLR-17370

# Description

This PR addresses a bug in Solr's facet serialization logic that could cause a java.lang.IllegalAccessException when faceting on string fields that contain numeric-looking values. The issue was observed when a field defined as type="string" in the schema (e.g. dv_device_code) was included in the facet field list, and some of its values were numeric. The serialization code would attempt to use Java reflection on internal Lucene classes (such as LongPoint), resulting in an access violation and a 500 error.

# Solution

- Added a defensive check in Utils.getReflectWriter to detect Lucene Point types (e.g org.apache.lucene.document.LongPoint) and avoid attempting reflective serialization on them.
- If such a type is encountered, Solr now falls back to a string representation (ClassName:toString()), preventing the IllegalAccessException and ensuring robust serialization.
- The fix is minimally invasive and only affects the serialization path for objects that would otherwise trigger the reflective access error.

# Tests

# Checklist

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/common/util/Utils.java
+++ b/solr/solrj/src/java/org/apache/solr/common/util/Utils.java
@@ -909,6 +909,15 @@ public class Utils {
    * @return a serializable version of the object
    */
   public static Object getReflectWriter(Object o) {
+    // Defensive: Avoid reflective serialization for Lucene Point types (e.g., LongPoint)
+    // which are not meant to be serialized this way and can cause IllegalAccessException.
+    String lucenePointPkg = "org.apache.lucene.document.";
+    if (o != null
+        && o.getClass().getName().startsWith(lucenePointPkg)
+        && o.getClass().getSimpleName().endsWith("Point")) {
+      // Fallback to string representation
+      return o.getClass().getName() + ':' + o.toString();
+    }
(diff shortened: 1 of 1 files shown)


solr/solrj/src/java/org/apache/solr/common/util/Utils.java
```diff
@@ -909,6 +909,15 @@ public static void reflectWrite(MapWriter.EntryWriter ew, Object o) {
    * @return a serializable version of the object
    */
   public static Object getReflectWriter(Object o) {
+    // Defensive: Avoid reflective serialization for Lucene Point types (e.g., LongPoint)
+    // which are not meant to be serialized this way and can cause IllegalAccessException.
+    String lucenePointPkg = "org.apache.lucene.document.";
+    if (o != null
+        && o.getClass().getName().startsWith(lucenePointPkg)
```
dsmiley, line 916: This feel like a hack to me.  Perhaps it's the best choice but hard for me to say at this time before understanding how this is invoked.  Maybe there's a solution preventing a Point field from getting here in the first place.  A test would help!

PR #3347, opened 2025-05-11
SOLR-7962: Passing additional arguments to solr.cmd using "--jvm-opts" (formerly "-a") does not work with "-e" on Windows

rahulgoswami:
https://issues.apache.org/jira/browse/SOLR-7962

# Description
- solr.cmd wasn't passing additional arguments to SolrCLI when running an example (-e option)
- Parsing of the prepared command line in RunExampleTool was breaking due to certain characters in the additional args. Eg: -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:18983". commons-cli had to be "signaled" to ask cmd.exe to preserve the quotes while passing to start.cmd
- If an additional arg (supplied with "--jvm-opts") starts with "-D", the parsing breaks. 

# Solution
- Pass additional args to SolrCLI when running with -e
- Parse the command in RunExampleTool without the additional args to prepare excec.CommandLine and then add the addtional args separately with quotes preserved.
- If a system property (-D) is passed as an additional argument with --jvm-opts, pass it along to the RunExampleTool, unlike the other -D arguments.  

# Tests

# Checklist

### DIFF
--- a/solr/bin/solr.cmd
+++ b/solr/bin/solr.cmd
@@ -752,7 +752,7 @@ IF NOT "%EXAMPLE%"=="" (
     -Dlog4j.configurationFile="file:///%DEFAULT_SERVER_DIR%\resources\log4j2-console.xml" ^
     -Dsolr.install.symDir="%SOLR_TIP%" ^
     -classpath "%DEFAULT_SERVER_DIR%\solr-webapp\webapp\WEB-INF\lib\*;%DEFAULT_SERVER_DIR%\lib\ext\*" ^
-    org.apache.solr.cli.SolrCLI run_example --script "%SDIR%\solr.cmd" -e %EXAMPLE% --server-dir "%SOLR_SERVER_DIR%" ^
+    org.apache.solr.cli.SolrCLI run_example --script "%SDIR%\solr.cmd" -e %EXAMPLE% --server-dir "%SOLR_SERVER_DIR%" --jvm-opts "%SOLR_ADDL_ARGS%" ^
     --url-scheme !SOLR_URL_SCHEME! !PASS_TO_RUN_EXAMPLE!
 
   REM End of run_example
--- a/solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
@@ -690,7 +690,8 @@ public class RunExampleTool extends ToolBase {
     String verboseArg = isVerbose() ? "--verbose" : "";
 
     String jvmOpts = cli.getOptionValue(JVM_OPTS_OPTION);
-    String jvmOptsArg = (jvmOpts != null) ? " --jvm-opts \"" + jvmOpts + "\"" : "";
+    String jvmOptsArg =
+        (jvmOpts != null && !jvmOpts.isEmpty()) ? " --jvm-opts \"" + jvmOpts + "\"" : "";
 
     Path cwd = Path.of(System.getProperty("user.dir"));
     Path binDir = Path.of(script).getParent();
@@ -713,7 +714,7 @@ public class RunExampleTool extends ToolBase {
     String startCmd =
         String.format(
             Locale.ROOT,
-            "\"%s\" start %s -p %d --solr-home \"%s\" --server-dir \"%s\" %s %s %s %s %s %s %s %s",
+            "\"%s\" start %s -p %d --solr-home \"%s\" --server-dir \"%s\" %s %s %s %s %s %s %s",
             callScript,
             cloudModeArg,
             port,
@@ -725,12 +726,13 @@ public class RunExampleTool extends ToolBase {
             forceArg,
             verboseArg,
             extraArgs,
-            jvmOptsArg,
             syspropArg);
     startCmd = startCmd.replaceAll("\\s+", " ").trim(); // for pretty printing
 
+    String startCmdWithJvmOpts = startCmd + jvmOptsArg;
+
     echo("\nStarting up Solr on port " + port + " using command:");
-    echo(startCmd + "\n");
+    echo(startCmdWithJvmOpts + "\n");
 
     String solrUrl =
         String.format(
@@ -759,7 +761,19 @@ public class RunExampleTool extends ToolBase {
         }
       }
       DefaultExecuteResultHandler handler = new DefaultExecuteResultHandler();
-      executor.execute(org.apache.commons.exec.CommandLine.parse(startCmd), startEnv, handler);
+      org.apache.commons.exec.CommandLine cmd = org.apache.commons.exec.CommandLine.parse(startCmd);
+
+      if (!jvmOptsArg.isEmpty()) {
+        cmd.addArgument("--jvm-opts");
+        // cmd.exe launched by commons-exec tends to strip off the quotes and break up the argument
+        // value on certain characters
+        // This may break the value in unintended ways thereby passing incorrect value to start.cmd
+        // (eg: for "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:18983", it
+        // breaks apart the value at "-agentlib:jdwp" passing incorrect args to start.cmd )
+        // The 'false' here tells Exec: “don’t touch my quotes—this is one atomic argument”
+        cmd.addArgument("\"" + jvmOpts + "\"", false);
+      }
+      executor.execute(cmd, startEnv, handler);
 
       // wait for execution.
       try {
--- a/solr/core/src/java/org/apache/solr/cli/SolrCLI.java
+++ b/solr/core/src/java/org/apache/solr/cli/SolrCLI.java
@@ -131,7 +131,7 @@ public class SolrCLI implements CLIO {
     List<String> dashDList = new ArrayList<>();
     for (int a = 1; a < args.length; a++) {
       String arg = args[a];
-      if (arg.startsWith("-D")) {
+      if (!args[a - 1].equals("--jvm-opts") && arg.startsWith("-D")) {
         dashDList.add(arg);
       } else {
         toolArgList.add(arg);


solr/bin/solr.cmd
```diff
@@ -752,7 +752,7 @@ IF NOT "%EXAMPLE%"=="" (
     -Dlog4j.configurationFile="file:///%DEFAULT_SERVER_DIR%\resources\log4j2-console.xml" ^
     -Dsolr.install.symDir="%SOLR_TIP%" ^
     -classpath "%DEFAULT_SERVER_DIR%\solr-webapp\webapp\WEB-INF\lib\*;%DEFAULT_SERVER_DIR%\lib\ext\*" ^
-    org.apache.solr.cli.SolrCLI run_example --script "%SDIR%\solr.cmd" -e %EXAMPLE% --server-dir "%SOLR_SERVER_DIR%" ^
+    org.apache.solr.cli.SolrCLI run_example --script "%SDIR%\solr.cmd" -e %EXAMPLE% --server-dir "%SOLR_SERVER_DIR%" --jvm-opts "%SOLR_ADDL_ARGS%" ^
```
epugh, line 755: I find it interesting that `SOLR_ADDL_ARGS` already existed as a variable but wasn't being passed in, was this the bug?

solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
```diff
@@ -723,12 +724,13 @@ protected Map<String, Object> startSolr(
             forceArg,
             verboseArg,
             extraArgs,
-            jvmOptsArg,
             syspropArg);
     startCmd = startCmd.replaceAll("\\s+", " ").trim(); // for pretty printing
 
+    String startCmdWithJvmOpts = startCmd + jvmOptsArg;
```
epugh, line 730: I don't love this variable name...   Maybe it's just `startCmd`?   and a comment why we moved `jvmOptsArg` out of the previous format command?

solr/bin/solr.cmd
rahulgoswami: This was definitely the main reason. However even after passing it, it was breaking apart in RunExampleTool downstream. Looks like CommandLine.parse() was not preserving the quotes around the additional arguments causing cmd invoked by executor.execute() to break up the argument value in unexpected ways. Hence start.cmd called by RunExampleTool gets broken arguments. Fixed that too.

solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
epugh: Is this maybe not startCmd, but instead "About logging parameters"?

solr/core/src/java/org/apache/solr/cli/SolrCLI.java
```diff
@@ -124,7 +124,7 @@ public static CommandLine parseCmdLine(Tool tool, String[] args) {
     List<String> dashDList = new ArrayList<>();
     for (int a = 1; a < args.length; a++) {
       String arg = args[a];
-      if (arg.startsWith("-D")) {
+      if (!args[a - 1].equals("--jvm-opts") && arg.startsWith("-D")) {
```
epugh, line 127: I think we probably hsould add some tests around this kind of tricky logic?

epugh (commented):
This is looking great!   I think I'd love to see a bats test to cover the linux regression, and maybe something in java unit test around the RunExampleTool changes?

solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
rahulgoswami: Right. At this point we are simply formatting the command just for logging. Still realigned a few things around the variable name to make it seem less awkward.

rahulgoswami:
> This is looking great! I think I'd love to see a bats test to cover the linux regression, and maybe something in java unit test around the RunExampleTool changes?

Working on it!

epugh:
> > This is looking great! I think I'd love to see a bats test to cover the linux regression, and maybe something in java unit test around the RunExampleTool changes?
> 
> Working on it! Btw do we have any bats equivalent for Windows ?

Unfortuanntly we do NOT have any equivalent for Windows.   I hope, that over the 10x line of Solr, that we move MORE of the logic out of the shell scripts into Java so we can write unit tests.   The "bin/solr start" I think is ripe for a "StartTool.java" to take over a lot of the argument parsing.  Even if we move lots of business logic into Java code, that then returns a fully formatted start command that is executed by the shell script, this would reduce the need for windows specfiic tests.

epugh:
I just triggered the workflows to run!

rahulgoswami:
Introduced a new test case for testing jvm-opts in TestSolrCLIRunExample. Facing a weird bug in the test case execution. The functionality is fine since I have verified from cmd. Working on resolving the bug in the test case.

solr/core/src/test/org/apache/solr/cli/TestSolrCLIRunExample.java
```diff
@@ -325,6 +330,99 @@ public void testFilmsExample() throws Exception {
     testExample("films");
   }
 
+  @Test
+  public void testAdditionalArgs() throws Exception {
+    String testStandaloneMode = LuceneTestCase.rarely() ? "--user-managed" : "";
+    Path defaultSolrHomeDir = ExternalPaths.SERVER_HOME;
+    if (!Files.isDirectory(defaultSolrHomeDir)) {
```
epugh, line 337: I don't know if this is required...    I believe that all our tests that use `ExternalPaths.SERVER_HOME` assume it's valid..   Altenratively, I think there is a assert or assume type operation that only runs the test when a criteria is met for things that ARE variable!

solr/core/src/test/org/apache/solr/cli/TestSolrCLIRunExample.java
```diff
@@ -325,6 +330,99 @@ public void testFilmsExample() throws Exception {
...
+    try (SolrClient solrClient =
+        new Http2SolrClient.Builder("http://localhost:" + bindPort + "/solr").build()) {
+      propValue = getSystemProperty(solrClient, sysProp);
+    }
+    assertEquals("System property provided with --jvm-opts not found", "hello", propValue);
+    // stop the test instance
+    executor.execute(org.apache.commons.exec.CommandLine.parse("bin/solr stop -p " + bindPort));
+  }
+
+  public String getSystemProperty(SolrClient solrClient, String propertyName) throws Exception {
```
epugh, line 404: I expected this to call the java system propertly method, or be a utility method that we already have in Solr!  On reading it, maybe a name like "assertSolrRunningWithSystemProperty" would make it clearer?

solr/core/src/test/org/apache/solr/cli/TestSolrCLIRunExample.java
```diff
@@ -325,6 +330,99 @@ public void testFilmsExample() throws Exception {
...
+          "-D" + sysProp + "=hello"
+        };
+
+    // capture tool output to stdout
+    CLITestHelper.TestingRuntime runtime = new CLITestHelper.TestingRuntime(true);
+
+    RunExampleExecutor executor = new RunExampleExecutor();
+    closeables.add(executor);
+
+    RunExampleTool tool = new RunExampleTool(executor, System.in, runtime);
```
epugh, line 374: I wonder if a helper method would reduice duplication between methods around the running othe tool and hanlding of the sockets etc?

epugh (commented):
Great progress, and this stuff is gnarly to test!  (and easy to break, so great to get some tests)

rahulgoswami:
Thanks for taking a look Eric. I converted the PR to a draft since I am still polishing off the tests. It is currently failing, but I committed it anyway in order to be able to pull it on macOS to see if the test fails for Linux too (it does!). The bug is in the test, not the fix, and it is seems to be because of a nuance in the /admin/info/properties that got inadvertently uncovered as part of this testing. I am fairly certain this would work if I retrieve the system property from the Java Runtime of the launched instance instead of calling /admin/info/properties. Now that I know the reason, I am a little more at peace and will revisit this later today.

rahulgoswami:
> Great progress, and this stuff is gnarly to test! (and easy to break, so great to get some tests)

Sigh! Cannot get tests for this part to run correctly. At this point I have spent wayyyy more time trying to get the tests to run successfully than I spent fixing the issue. The CLI launches the Java process which launches the CLI which launches the actual Solr process. Easy peasy! ...Not! Not when I have to replicate this in tests. :(   [Just to recap, with this fix, launching the example with --jvm-opts via CLI works fine on both Windows and Solr. It's just that it's harder than I thought to replicate this behavior in tests, which is what I am working on.]
The other tests for RunExampleTool are launching Solr in embedded mode which won't work for this PR. This is since I need to test for the presence of --jvm-args which needs to be passed at the launch of the Solr instance and cannot be tested "within" an already running JVM of the test suite.

Rant over....back to fixing the tests.

rahulgoswami:
> The bug is in the test, not the fix, and it is seems to be because of a nuance in the /admin/info/properties that got inadvertently uncovered as part of this testing.

I take back the part where I said the issue with tests is because of a nuance in /admin/info/properties. I copy/pasted certain parts of the test from another test in TestSolrCLIRunExample (testExample()), which causes RunExampleTool's startSolr() method to end up calling "solr" script instead of "solr.cmd" (I would expect this to throw an error on Windows, but it doesn't. I am guessing it's got something to do with the custom executor (RunExampleExecutor) that the tests use).  

Upon further examination seems like the existing testExample() uses a custom RunExampleExecutor to execute the launching process. This launches Solr in embedded mode binding to a pre-declared server Socket. In effect, it seems like a new JVM instance is never launched. Which is why the system property sent with --jvm-opts doesn't reflect. This is not the case when actually running the example from CLI, and hence that works as expected.

epugh:
So...   There is a point of diminishing returns on tests...  If you have manually tested things on windows and linux...  then that is good...  We *do* I think have a bats style test right for LInux, so that is good...  

Progress over perfection..

epugh:
> > There is a point of diminishing returns on tests... If you have manually tested things on windows and linux... then that is good
> 
> Makes sense. And although I hate to admit it, I _do_ think I might have hit that point with this PR. I am going to revert some of the test changes and take a quick look at the Linux bats test later tonight. And hope to get this PR polished off thereafter.

Sounds like a plan!

rahulgoswami:
Ok so this is ready. When trying to run an example on MacOS with -e, I am getting into a weird AccessControlException with Jetty. I do not see this behavior on Windows. I have a feeling something broke on the main branch and when I merged main into this PR branch, it is breaking the example run here too. I ran the example on the main branch thereafter, and indeed the same issue there.

In case this is something environmental with my Mac, I would really appreciate if you could help run the bats test (test_example.bats) in this PR on your Linux/MacOS machine. And if everything looks good, kindly send these changes to the promised land!  :-)

Of course if something needs to be addressed, I'll take a stab at it in the morning. Thank you for your collaboration throughout!

solr/CHANGES.txt
```diff
@@ -291,6 +291,9 @@ Bug Fixes
 * SOLR-17758: The NumFieldLimitingUpdateRequestProcessor's "warnOnly" mode has been fixed, and now processes documents even
   when the limit has been exceeded. (Jason Gerlowski, Rahul Goswami)
 
+* SOLR-7962: Passing additional arguments to solr.cmd using "--jvm-opts" (formerly "-a") in conjunction with "-e" (examples like 'techproducts')
```
epugh, line 294: Nice description!

solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
```diff
@@ -230,17 +230,13 @@ public void runImpl(CommandLine cli) throws Exception {
...
       } else {
-        scriptFile = serverDir.getParent().resolve("bin").resolve("solr.cmd");
-        if (Files.isRegularFile(scriptFile)) {
-          script = scriptFile.toAbsolutePath().toString();
-        } else {
-          throw new IllegalArgumentException(
-              "Cannot locate the bin/solr script! Please pass --script to this application.");
-        }
+        throw new IllegalArgumentException(
+            "Cannot locate the bin/solr script! Please pass --script to this application.");
```
epugh, line 239: maybe do the same `solr.cmd : solr` thing in the error message?   nit pick ;_0

solr/packaging/test/test_example.bats
```diff
@@ -0,0 +1,41 @@
...
+
+teardown() {
+  # save a snapshot of SOLR_HOME for failed tests
+  save_home_on_failure
+
+  solr stop --all >/dev/null 2>&1
+}
+
+@test "start -e cloud works with --jvm-opts" {
+  solr start -e cloud --no-prompt --jvm-opts "-Dcustom.prop=helloworld"
```
epugh, line 32: nice that this is also a test that our cloud example starts!  we didn't have that in bats at all..

epugh (approved):
Looks all good.  I am running tests and then will merge.

solr/core/src/java/org/apache/solr/cli/RunExampleTool.java
rahulgoswami: Resolved!

PR #3348, opened 2025-05-12
SOLR-17758: Fix limiting-URP "warnOnly" mode

gerlowskija:
https://issues.apache.org/jira/browse/SOLR-17758

# Description

Due to a bug in the initial implementation, the "warnOnly" mode of the NumFieldLimiting "URP" never worked as designed.  The "warn" codepath failed to call `propagateAdd` for the next URP in the chain.  This makes the "warn only" behavior indistinguishable from the "hard limit" behavior from a user perspective.

# Solution

This commit fixes the bug by ensuring that the `processAdd` call is propagated, in all cases.  It also adds integration testing for several "warn only" mode related scenarios, to prevent against future regressions.

# Tests

See `NumFieldLimitingUpdateRequestProcessorIntegrationTest`

# Checklist


gerlowskija:
The 'test' job here flagged a failure in `NumFieldLimitingUpdateRequestProcessorIntegrationTest`.  It looked a lot like a collection-reload triggered by the test was returning before all of the involved cores themselves were reloaded, and that subsequent update requests were going to the "pre-reload" SolrCore.

Still not 100% sure if that's the case.  I made a couple tweaks to workaround this (and to make the test easier to run with `-Ptests.dups`), and that fixed the failures locally for me across a few hundred iterations.

I'll give it a few hundred more runs, and then merge if things still look good.  But I'll keep a particularly close eye on the test runs, just in case.

That said - the collection-reload semantics that I observed seem pretty concerning, and worth a follow up JIRA.  Has anyone seen anything like this before?

gerlowskija:
Eh - ran into more failures here pretty quickly, with the surprising collection-reload semantics.  Will give this more time/attention before merging.

rahulgoswami (commented):
~~Sorry if this counts as an unsolicited review comment, but this should be >= instead of > ? 
More so since the user complained about getting a 400 error code (SolrException.ErrorCode.BAD_REQUEST) despite having "warnOnly" true, likely due to flowing into the "else" logic in procesAdd(). I haven't had a chance to reproduce this myself, but ">=" seems to be the more correct check anyway (?)~~ 
Don't see a delete option when I click on the kebab menu towards top right. Please ignore this and refer to the below comment instead.

rahulgoswami (changes requested):
Sorry if this counts as an unsolicited review comment, but this should be >= instead of > ?
More so since the user complained about getting a 400 error code (SolrException.ErrorCode.BAD_REQUEST) despite having "warnOnly" true, likely due to flowing into the "else" logic in procesAdd(). I haven't had a chance to reproduce this myself, but ">=" seems to be the more correct check anyway (?)

gerlowskija:
Ooh, good catch Rahul!  I'm afk, but excited to try this out in a bit.

Pretty sure I copy/pasted that arg-parsing from elsewhere in Solr, so I bet there's a few other places with that bug that we can fix as well 🤔

gerlowskija:
Looks like this fixed my test issues - I take back what I said earlier about potential "race" issues with the reload API.

PR #3350, opened 2025-05-13
SOLR-17754 Fix race condition in overseer main loop.

psalagnac:
https://issues.apache.org/jira/browse/SOLR-17754

# Description

Stuck overseer that sometimes happens under high load, when the overseer has at least 100 running tasks.

See [Jira](https://issues.apache.org/jira/browse/SOLR-17754) for the full scenario as description is pretty long.

# Solution

This fixes the overseer main loop so we never submit more than 100 concurrent tasks to the thread pool. Instead of manually tracking when a task is complete, we check the status using a standard java `Future`.

The changes also makes sure we don't write the result to ZK response node when we should not (see [comment](https://issues.apache.org/jira/browse/SOLR-17754?focusedCommentId=17951216&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17951216)), removing the erroneous occurrences of log `"Response ZK path: <node> doesn't exist. Requestor may have disconnected from ZooKeeper"`

# Tests

Add a new test to make sure we don't fail anymore with lot of tasks.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -190,7 +190,7 @@ Other Changes
 New Features
 ---------------------
 * SOLR-17582: The CLUSTERSTATUS API will now stream each collection's status to the response,
-  fetching and computing it on the fly.  To avoid a backwards compatibilty concern, this won't work
+  fetching and computing it on the fly.  To avoid a backwards compatibility concern, this won't work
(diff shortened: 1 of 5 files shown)


psalagnac:
Going to merge this PR this week.

PR #3352, opened 2025-05-14
SOLR-17756: Parallelize calculation of index fingerprint across segments

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-17756

# Description

The index fingerprint is currently being calculated on each segment sequentially. While this works fine, the index fingerprint calculation was noticed to be a very slow process and on leader election is blocking.

This proposes to have this calculation parallelized across segments instead. Since the fingerprint is just a cumulative sum of a hash on versions, the order in which it is added to the running sum should not matter.

# Solution

Create a dedicated threadpool (`IndexFingerprintPool`) to calculate the indexfingerprint. Created a separate theadpool instead of the common `ForkJoinPool` as this can be an expensive operation calculating across a large index

# Tests

Created `testSequentialVsParallelFingerprint` test to confirm the fingerprint is the same.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -1307,7 +1307,7 @@ public class CoreContainer {
 
     ExecutorUtil.shutdownAndAwaitTermination(coreContainerAsyncTaskExecutor);
     ExecutorUtil.shutdownAndAwaitTermination(indexSearcherExecutor);
-    indexFingerprintExecutor.shutdownNow();
+    ExecutorUtil.shutdownNowAndAwaitTermination(indexFingerprintExecutor);
     ExecutorService customThreadPool =
         ExecutorUtil.newMDCAwareCachedThreadPool(new SolrNamedThreadFactory("closeThreadPool"));
 
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -458,8 +458,10 @@ public class CoreContainer {
     this.indexSearcherExecutor = SolrIndexSearcher.initCollectorExecutor(cfg);
 
     this.indexFingerprintExecutor =
-        ExecutorUtil.newMDCAwareFixedThreadPool(
-            EXECUTOR_MAX_CPU_THREADS, new SolrNamedThreadFactory("IndexFingerprintPool"));
+        ExecutorUtil.newMDCAwareCachedThreadPool(
+            EXECUTOR_MAX_CPU_THREADS,
+            Integer.MAX_VALUE,
+            new SolrNamedThreadFactory("IndexFingerprintPool"));
   }
 
   @SuppressWarnings({"unchecked"})
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -26,6 +26,7 @@ import static org.apache.solr.common.params.CommonParams.INFO_HANDLER_PATH;
 import static org.apache.solr.common.params.CommonParams.METRICS_PATH;
 import static org.apache.solr.common.params.CommonParams.ZK_PATH;
 import static org.apache.solr.common.params.CommonParams.ZK_STATUS_PATH;
+import static org.apache.solr.search.SolrIndexSearcher.EXECUTOR_MAX_CPU_THREADS;
 import static org.apache.solr.security.AuthenticationPlugin.AUTHENTICATION_PLUGIN_PROP;
 
 import com.github.benmanes.caffeine.cache.Interner;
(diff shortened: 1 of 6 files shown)


mlbiscoc:
@kotman12 @psalagnac @gerlowskija Here is the new PR after I messed up the pulling commits from main on the last one. Our commit history is still the same as before but with addition of the change from Pierres comment in `0fb9900`

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1293,6 +1307,7 @@ public void shutdown() {
 
     ExecutorUtil.shutdownAndAwaitTermination(coreContainerAsyncTaskExecutor);
     ExecutorUtil.shutdownAndAwaitTermination(indexSearcherExecutor);
+    indexFingerprintExecutor.shutdownNow();
```
psalagnac, line 1310: I would call `ExecutorUtil.shutdownNowAndAwaitTermination()` instead.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2546,24 +2547,17 @@ public Explanation explain(Query query, int doc) throws IOException {
...
-          .map(
-              ctx -> {
-                try {
-                  return searcher.getCore().getIndexFingerprint(searcher, ctx, maxVersion);
-                } catch (IOException e) {
-                  exception.set(e);
-                  return null;
-                }
-              })
-          .filter(java.util.Objects::nonNull)
```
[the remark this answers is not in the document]
kotman12: Yes it was on purpose. Ultimately all roads lead to [this](https://github.com/apache/solr/blob/40926bcdd3f279022746e46b46f124f3cc12747c/solr/core/src/java/org/apache/solr/update/IndexFingerprint.java#L110-L137). The previous implementation allowed for a null in the case of an exception [here](https://github.com/apache/solr/blob/40926bcdd3f279022746e46b46f124f3cc12747c/solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java#L2558) which was a hack to make this work with java streams. But with  [ExecutorUtil::submitAllAndAwaitAggregatingExceptions](https://github.com/apache/solr/blob/40926bcdd3f279022746e46b46f124f3cc12747c/solr/solrj/src/java/org/apache/solr/common/util/ExecutorUtil.java#L451C9-L451C16) luckily exceptions don't get planted as null values that we later have to ignore.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
mlbiscoc: I had originally looked at `shutdownNowAndAwaitTermination` because of the 2 other Executors above but I felt that `shutdownNow()` was fine and there wasn't any need to hold the shutdown from this pool since the indexFingerprint calculation doesn't affect anything if you cut it off in the middle of shutdown. I didn't think there is any benefit to waiting for this to cleanup.

dsmiley: PIerre is right: we should await termination.  Solr, particularly the tests, are very demanding that absolutely all threads complete when a test completes, which in turn them requires that shutting down CoreContainer stops all threads belonging to it.  This also aligns with the upcoming "structured concurrency" philosophy loosely.

solr/core/src/java/org/apache/solr/update/IndexFingerprint.java
```diff
@@ -219,4 +220,29 @@ public static IndexFingerprint fromObject(Object o) {
   public String toString() {
     return toMap(new LinkedHashMap<>()).toString();
   }
+
+  @Override
+  public boolean equals(Object o) {
```
dsmiley, line 225: Curious; are these needed?

IndexFingerprint appears to be ripe for being a `record`.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
mlbiscoc: Understood. I went ahead and moved over to `shutdownNowAndAwaitTermination ` for shutdown.

solr/core/src/java/org/apache/solr/update/IndexFingerprint.java
mlbiscoc: Its for the test to `assertEquals` both fingerprints for comparison if identical in attributes. `compare` didn't check `numDocs`, `maxDoc` and `maxVersionSpecified`.

dsmiley (approved):

mlbiscoc:
Will merge this later in the week.

dsmiley:
If you wish, you can merge sooner -- you got approving reviews >= 48 hours ago.
If nobody reviews your work, then you would declare an intention to merge in 48 hours (or more if you wish).
Trivial updates like docs, CHANGES.txt, don't reset the clock.

There's no hard mandate on this matter, but the above is for best behavior.

PR #3355, opened 2025-05-16
SOLR-17625 NamedList.findRecursive-> _get

wildtusker:
https://issues.apache.org/jira/browse/SOLR-17625


solr/core/src/java/org/apache/solr/cli/AssertTool.java
```diff
@@ -280,7 +281,8 @@ public int assertSolrNotRunning(String url, String credentials) throws Exception
         System.nanoTime() + TimeUnit.NANOSECONDS.convert(timeoutMs, TimeUnit.MILLISECONDS);
     try (SolrClient solrClient = CLIUtils.getSolrClient(url, credentials)) {
       NamedList<Object> response = solrClient.request(new HealthCheckRequest());
-      Integer statusCode = (Integer) response.findRecursive("responseHeader", "status");
+      Object value = response._get(List.of(new String[] {"responseHeader", "status"}), null);
+      Integer statusCode = (Integer) value;
```
dsmiley, line 285: Don't declare new variables merely to cast.

solr/core/src/java/org/apache/solr/cli/HealthcheckTool.java
```diff
@@ -173,9 +173,12 @@ protected void runCloudTool(CloudSolrClient cloudSolrClient, CommandLine cli) th
                   solrClient.request(
                       new GenericSolrRequest(
                           SolrRequest.METHOD.GET, CommonParams.SYSTEM_INFO_PATH));
-              uptime = SolrCLI.uptime((Long) systemInfo.findRecursive("jvm", "jmx", "upTimeMS"));
-              String usedMemory = (String) systemInfo.findRecursive("jvm", "memory", "used");
-              String totalMemory = (String) systemInfo.findRecursive("jvm", "memory", "total");
+              Object value2 = systemInfo._get(List.of("jvm", "jmx", "upTimeMS"), null);
+              uptime = SolrCLI.uptime((Long) value2);
+              Object value1 = systemInfo._get(List.of("jvm", "memory", "used"), null);
```
dsmiley, line 178: use _getStr, same with below

solr/core/src/java/org/apache/solr/cloud/api/collections/SplitShardCmd.java
```diff
@@ -862,14 +862,18 @@ public static void checkDiskSpace(
                 SolrRequest.METHOD.GET, "/admin/metrics", SolrRequest.SolrRequestType.ADMIN, params)
             .process(cloudManager.getSolrClient());
 
-    Number size = (Number) rsp.getResponse().findRecursive("metrics", indexSizeMetricName);
+    NamedList<Object> entries1 = rsp.getResponse();
```
dsmiley, line 865: again, no need to declare another variable

dsmiley (changes requested):
I suspect you didn't review your work / the results.

You had `List.of(new String[]{"foo", "bar"})` in all these places instead of `List.of("foo", "bar")`.  Using IntelliJ "structural replace" feature, I easily fixed all of them at once and pushed to your PR.

There are newly declared variables everywhere.  That should be completely unnecessary.  You can fix them by doing doing "inline" refactoring in IntelliJ when clicking each needless variable (use a hot key to make it go quicker).

In places a String is casted, `_getStr` should be used.

In some places, I expect the default value (last arg) to be useful, which `findRecursive` didn't support.

All this makes this change look way worse but instead we should find essentially the same amount of code as before, mostly.

wildtusker:
1)   List.of(new String[]{"foo", "bar"})

2) Made changes using Method Over loading .Got rid of List.of(new String[]{"foo", "bar"}) .Even the variables which were getting created 
3) Newly declared variables have been removed

4_getStr has not been used as of now. 
Default value will be on case to case basis or where required.

solr/core/src/java/org/apache/solr/cli/StatusTool.java
```diff
@@ -314,12 +314,12 @@ public static Map<String, Object> reportStatus(NamedList<Object> info, SolrClien
 
     String solrHome = (String) info.get("solr_home");
     status.put("solr_home", solrHome != null ? solrHome : "?");
-    status.put("version", info.findRecursive("lucene", "solr-impl-version"));
-    status.put("startTime", info.findRecursive("jvm", "jmx", "startTime").toString());
-    status.put("uptime", SolrCLI.uptime((Long) info.findRecursive("jvm", "jmx", "upTimeMS")));
+    status.put("version", info._get(List.of("lucene", "solr-impl-version"), null));
+    status.put("startTime", info._get(List.of("jvm", "jmx", "startTime"), null).toString());
```
dsmiley, line 318: these two can be `_getStr`

solr/core/src/java/org/apache/solr/cli/StatusTool.java
```diff
@@ -314,12 +314,12 @@ public static Map<String, Object> reportStatus(NamedList<Object> info, SolrClien
...
+    status.put("uptime", SolrCLI.uptime((Long) info._get(List.of("jvm", "jmx", "upTimeMS"), null)));
 
-    String usedMemory = (String) info.findRecursive("jvm", "memory", "used");
-    String totalMemory = (String) info.findRecursive("jvm", "memory", "total");
+    String usedMemory = (String) info._get(List.of("jvm", "memory", "used"), null);
+    String totalMemory = (String) info._get(List.of("jvm", "memory", "total"), null);
```
dsmiley, line 322: `_getStr`

solr/core/src/java/org/apache/solr/cloud/api/collections/SplitShardCmd.java
dsmiley: still an issue

dsmiley (commented):
I did a partial review.  I encourage you to review it fully (I do it for my own work).

I get the feeling that `_get` ought not to have a default parameter (at the end) since it's so rarely useful, judging from this PR anyway.  If that makes sense; it'd be a separate PR later.

solr/core/src/java/org/apache/solr/cli/StatusTool.java
wildtusker: Will make the change. _get is replaced with _getStr.

solr/core/src/java/org/apache/solr/cli/HealthcheckTool.java
```diff
@@ -173,9 +173,11 @@ protected void runCloudTool(CloudSolrClient cloudSolrClient, CommandLine cli) th
                   solrClient.request(
                       new GenericSolrRequest(
                           SolrRequest.METHOD.GET, CommonParams.SYSTEM_INFO_PATH));
-              uptime = SolrCLI.uptime((Long) systemInfo.findRecursive("jvm", "jmx", "upTimeMS"));
-              String usedMemory = (String) systemInfo.findRecursive("jvm", "memory", "used");
-              String totalMemory = (String) systemInfo.findRecursive("jvm", "memory", "total");
+              uptime =
+                  SolrCLI.uptime((Long) systemInfo._get(List.of("jvm", "jmx", "upTimeMS"), null));
+              String usedMemory = (String) systemInfo._get(List.of("jvm", "memory", "used"), null);
```
dsmiley, line 178: I assume you didn't finish reviewing this PR since this shows some `_getStr` opportunities.

solr/core/src/test/org/apache/solr/search/facet/SpatialHeatmapFacetsTest.java
```diff
@@ -505,14 +502,15 @@ public void testJsonFacets() throws Exception {
...
     final NamedList<?> classicResp =
-        (NamedList<?>)
-            response.getResponse().findRecursive("facet_counts", "facet_heatmaps", FIELD);
+        (NamedList<?>) entries1._get(List.of("facet_counts", "facet_heatmaps", FIELD), null);
     if (classicResp != null) {
       return classicResp;
     }
     // JSON Facet
-    return (NamedList<?>) response.getResponse().findRecursive("facets", "f1");
+    NamedList<Object> entries = response.getResponse();
```
dsmiley, line 512: inline

solr/core/src/test/org/apache/solr/search/facet/SpatialHeatmapFacetsTest.java
```diff
@@ -505,14 +502,15 @@ public void testJsonFacets() throws Exception {
 
   private NamedList<?> getHmObj(QueryResponse response) {
     // classic faceting
+    NamedList<Object> entries1 = response.getResponse();
```
dsmiley, line 505: inline

dsmiley (commented):
Please self review more next time so that you believe it to be ready.  There were dubious `entries1` that we discussed previously.

Otherwise looks good!

solr/core/src/test/org/apache/solr/search/facet/SpatialHeatmapFacetsTest.java
```diff
@@ -438,8 +436,7 @@ public void testJsonFacets() throws Exception {
                       + " } "
                       + "}"));
       {
-        NamedList<Object> entries1 = response.getResponse();
-        final NamedList<?> q1Res = (NamedList<?>) entries1._get(List.of("facets", "q1"), null);
+          final NamedList<?> q1Res = (NamedList<?>) response.getResponse()._get(List.of("facets", "q1"), null);
```
dsmiley, line 439: remember "gw tidy", I see here and elsewhere.  I think you can configure IntelliJ better so that basic indentation like this would not be incorrect so often.  I have an idea that I'll raise elsewhere that should reduce this annoyance.

PR #3356, opened 2025-05-17
SOLR-17711 index fetcher doesn't need request timeout

kotman12:
https://issues.apache.org/jira/browse/SOLR-17711

# Description

There is now a default total request timeout applied to IndexFetcher because of Jetty Http2 client migration. We ran into a few problems with this already during replication of large number of shards on a single host. Sometimes we hit machine bandwidth limits because shards are downloaded in parallel. Thus requests for the next segment of each shard compete with each other and sometimes take more than 120 seconds. The 120 second number is important because it is the default _idle_ timeout and ever since [this change](https://github.com/apache/solr/commit/e62cb1b0066132347589b5e5ca38f38fd5e668d0), the idle timeout happens to be the total request timeout in the http2 client. I don't think that is a good idea in the case of IndexFetcher, hence this change.

# Solution

Revert to the pre-`Http2SolrClient` behavior of no timeout in `IndexFetcher`.

# Tests

Not aware of an easy way to test this reasonably, i.e. to accelerate the clock.

# Checklist


dsmiley (approved):
Looks good, thanks.
Can you please add to CHANGES.txt in the improvement section?

kotman12:
@dsmiley Just saw your message to the mailing list, thanks. So it looks like what I propose here is incidentally a no-op as it currently stands? Should I just wait for your PR then? Looks like I'd have to make this change to the `recoveryOnlyClient` which may or may not be the right thing to do but the question might be moot if we apply your planned changes.

solr/CHANGES.txt
```diff
@@ -49,6 +49,8 @@ Improvements
 
 * SOLR-17739: Uploading a config sets with forbidden file types will produces a Exception. Previously it was just ignoring those files and logging a WARN. (Abhishek Umarjikar via Eric Pugh)
 
+* SOLR-17711: Remove total request timeout from Recovery Client inadvertently added during Jetty Http2 client migration. Recovery requests can take a long time. (Luke Kot-Zaniewski)
```
dsmiley: You've put this in 10.x but it belongs in 9.x IMO.

dsmiley:
Let's get this into 9.9; right?

solr/CHANGES.txt
kotman12: Updated to put this in 9.9 .. also moved it to bug-fix category

solr/CHANGES.txt
```diff
@@ -295,6 +293,8 @@ Bug Fixes
 * SOLR-17758: The NumFieldLimitingUpdateRequestProcessor's "warnOnly" mode has been fixed, and now processes documents even
   when the limit has been exceeded. (Jason Gerlowski, Rahul Goswami)
 
+* SOLR-17711: Remove total request timeout from Recovery Client inadvertently added during Jetty Http2 client migration. Recovery requests can take a long time. (Luke Kot-Zaniewski)
```
dsmiley: I don't think most users understand what this means.  The audience of CHANGES.txt is users.

kotman12: how about now?

PR #3357, opened 2025-05-19
Harmonize SolrJ timeouts

dsmiley:

https://lists.apache.org/thread/002ql2ol0dklswhskokhxopj3d9tpnwj

### DIFF
--- a/solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/SolrClientCache.java
+++ b/solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/SolrClientCache.java
@@ -205,19 +205,14 @@ public class SolrClientCache implements Closeable {
                 .withDefaultCollection(URLUtil.extractCoreFromCoreUrl(url));
     if (http2SolrClient != null) {
       builder.withHttpClient(http2SolrClient);
+      // cannot set idleTimeout or connection timeout
+      // nocommit
+    } else {
+        builder.withIdleTimeout(minSocketTimeout, TimeUnit.MILLISECONDS);
+        builder.withConnectionTimeout(minConnTimeout, TimeUnit.MILLISECONDS);
     }
     builder.withOptionalBasicAuthCredentials(basicAuthCredentials);
 
-    long idleTimeout = minSocketTimeout;
-    if (builder.getIdleTimeoutMillis() != null) {
-      idleTimeout = Math.max(idleTimeout, builder.getIdleTimeoutMillis());
-    }
-    builder.withIdleTimeout(idleTimeout, TimeUnit.MILLISECONDS);
-    long connTimeout = minConnTimeout;
-    if (builder.getConnectionTimeout() != null) {
-      connTimeout = Math.max(idleTimeout, builder.getConnectionTimeout());
-    }
-    builder.withConnectionTimeout(connTimeout, TimeUnit.MILLISECONDS);
     return builder;
   }
 
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/ConcurrentUpdateHttp2SolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/ConcurrentUpdateHttp2SolrClient.java
@@ -256,7 +256,7 @@ public class ConcurrentUpdateHttp2SolrClient extends SolrClient {
             }
 
             Response response =
-                responseListener.get(client.getIdleTimeout(), TimeUnit.MILLISECONDS);
+                responseListener.get(client.getRequestTimeoutMillis(), TimeUnit.MILLISECONDS);
             rspBody = responseListener.getInputStream();
 
             int statusCode = response.getStatus();
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
@@ -94,15 +94,12 @@ import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
 /**
- * Difference between this {@link Http2SolrClient} and {@link HttpSolrClient}:
+ * An HTTP {@link SolrClient} using Jetty {@link HttpClient}. This is Solr's most mature client for
+ * direct HTTP.
  *
- * <ul>
- *   <li>{@link Http2SolrClient} sends requests in HTTP/2
- *   <li>{@link Http2SolrClient} can point to multiple urls
- *   <li>{@link Http2SolrClient} does not expose its internal httpClient like {@link
- *       HttpSolrClient#getHttpClient()}, sharing connection pools should be done by {@link
- *       Http2SolrClient.Builder#withHttpClient(Http2SolrClient)}
- * </ul>
+ * <p>Despite the name, this client supports HTTP 1.1 and 2 -- toggle with {@link
+ * HttpSolrClientBuilderBase#useHttp1_1(boolean)}. In retrospect, the name should have been {@code
+ * HttpJettySolrClient}.
  */
 public class Http2SolrClient extends HttpSolrClientBase {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
@@ -129,6 +126,12 @@ public class Http2SolrClient extends HttpSolrClientBase {
     super(serverBaseUrl, builder);
 
     if (builder.httpClient != null) {
+      if (builder.followRedirects != null
+          || builder.connectionTimeoutMillis != null
+          || builder.idleTimeoutMillis != null) {
+        throw new IllegalArgumentException(
+            "You cannot provide the HttpClient and also specify options that are used to build a new client");
+      }
       this.httpClient = builder.httpClient;
       if (this.executor == null) {
         this.executor = builder.executor;
@@ -252,7 +255,9 @@ public class Http2SolrClient extends HttpSolrClientBase {
     httpClient.setMaxRequestsQueuedPerDestination(
         asyncTracker.getMaxRequestsQueuedPerDestination());
     httpClient.setUserAgentField(new HttpField(HttpHeader.USER_AGENT, USER_AGENT));
-    httpClient.setIdleTimeout(idleTimeoutMillis);
+    httpClient.setConnectTimeout(builder.getConnectionTimeoutMillis());
+    httpClient.setIdleTimeout(builder.getIdleTimeoutMillis());
+    // note: request timeout is set per request
 
     if (builder.cookieStore != null) {
       httpClient.setCookieStore(builder.cookieStore);
@@ -260,8 +265,6 @@ public class Http2SolrClient extends HttpSolrClientBase {
 
     this.authenticationStore = new AuthenticationStoreHolder();
     httpClient.setAuthenticationStore(this.authenticationStore);
-
-    httpClient.setConnectTimeout(builder.connectionTimeoutMillis);
 
     setupProxy(builder, httpClient);
 
@@ -327,6 +330,11 @@ public class Http2SolrClient extends HttpSolrClientBase {
 
   public void setAuthenticationStore(AuthenticationStore authenticationStore) {
     this.authenticationStore.updateAuthenticationStore(authenticationStore);
+  }
+
+  /** (visible for testing) */
+  public long getIdleTimeout() {
+    return getHttpClient().getIdleTimeout();
   }
 
   public static class OutStream implements Closeable {
@@ -514,7 +522,7 @@ public class Http2SolrClient extends HttpSolrClientBase {
     try {
       InputStreamResponseListener listener = new InputStreamReleaseTrackingResponseListener();
       req = sendRequest(makeRequest(solrRequest, url, false), listener);
-      Response response = listener.get(idleTimeoutMillis, TimeUnit.MILLISECONDS);
+      Response response = listener.get(requestTimeoutMillis, TimeUnit.MILLISECONDS);
       url = req.getURI().toString();
       InputStream is = listener.getInputStream();
       return processErrorsAndResponse(solrRequest, response, is, url);
@@ -622,11 +630,8 @@ public class Http2SolrClient extends HttpSolrClientBase {
   private void decorateRequest(Request req, SolrRequest<?> solrRequest, boolean isAsync) {
     req.headers(headers -> headers.remove(HttpHeader.ACCEPT_ENCODING));
 
-    if (requestTimeoutMillis > 0) {
-      req.timeout(requestTimeoutMillis, TimeUnit.MILLISECONDS);
-    } else {
-      req.timeout(idleTimeoutMillis, TimeUnit.MILLISECONDS);
-    }
+    req.timeout(requestTimeoutMillis, TimeUnit.MILLISECONDS);
+
     if (solrRequest.getUserPrincipal() != null) {
       req.attribute(REQ_PRINCIPAL_KEY, solrRequest.getUserPrincipal());
     }
@@ -1020,12 +1025,6 @@ public class Http2SolrClient extends HttpSolrClientBase {
       if (cookieStore == null) {
         cookieStore = getDefaultCookieStore();
       }
-      if (idleTimeoutMillis == null || idleTimeoutMillis <= 0) {
-        idleTimeoutMillis = (long) HttpClientUtil.DEFAULT_SO_TIMEOUT;
-      }
-      if (connectionTimeoutMillis == null) {
-        connectionTimeoutMillis = (long) HttpClientUtil.DEFAULT_CONNECT_TIMEOUT;
-      }
 
       if (keyStoreReloadIntervalSecs != null
           && keyStoreReloadIntervalSecs > 0
@@ -1085,17 +1084,11 @@ public class Http2SolrClient extends HttpSolrClientBase {
       if (this.basicAuthAuthorizationStr == null) {
         this.basicAuthAuthorizationStr = http2SolrClient.basicAuthAuthorizationStr;
       }
-      if (this.followRedirects == null) {
-        this.followRedirects = http2SolrClient.httpClient.isFollowRedirects();
-      }
-      if (this.idleTimeoutMillis == null) {
-        this.idleTimeoutMillis = http2SolrClient.idleTimeoutMillis;
+      if (this.requestTimeoutMillis == null) {
+        this.requestTimeoutMillis = http2SolrClient.requestTimeoutMillis;
       }
       if (this.requestWriter == null) {
         this.requestWriter = http2SolrClient.requestWriter;
-      }
-      if (this.requestTimeoutMillis == null) {
-        this.requestTimeoutMillis = http2SolrClient.requestTimeoutMillis;
       }
       if (this.responseParser == null) {
         this.responseParser = http2SolrClient.parser;
@@ -1104,8 +1097,7 @@ public class Http2SolrClient extends HttpSolrClientBase {
         this.urlParamNames = http2SolrClient.urlParamNames;
       }
       if (this.listenerFactory == null) {
-        this.listenerFactory = new ArrayList<HttpListenerFactory>();
-        http2SolrClient.listenerFactory.forEach(this.listenerFactory::add);
+        this.listenerFactory = new ArrayList<>(http2SolrClient.listenerFactory);
       }
       if (this.executor == null) {
         this.executor = http2SolrClient.executor;
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpJdkSolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpJdkSolrClient.java
@@ -86,12 +86,18 @@ public class HttpJdkSolrClient extends HttpSolrClientBase {
 
   protected HttpJdkSolrClient(String serverBaseUrl, HttpJdkSolrClient.Builder builder) {
     super(serverBaseUrl, builder);
+    HttpClient.Builder b = HttpClient.newBuilder();
 
     HttpClient.Redirect followRedirects =
         Boolean.TRUE.equals(builder.followRedirects)
             ? HttpClient.Redirect.NORMAL
             : HttpClient.Redirect.NEVER;
-    HttpClient.Builder b = HttpClient.newBuilder().followRedirects(followRedirects);
+    b.followRedirects(followRedirects);
+
+    b.connectTimeout(Duration.of(builder.getConnectionTimeoutMillis(), ChronoUnit.MILLIS));
+    // note: idle timeout isn't used for the JDK client
+    // note: request timeout is set per request
+
     if (builder.sslContext != null) {
       b.sslContext(builder.sslContext);
     }
@@ -419,11 +425,7 @@ public class HttpJdkSolrClient extends HttpSolrClientBase {
   }
 
   private void decorateRequest(HttpRequest.Builder reqb, SolrRequest<?> solrRequest) {
-    if (requestTimeoutMillis > 0) {
-      reqb.timeout(Duration.of(requestTimeoutMillis, ChronoUnit.MILLIS));
-    } else if (idleTimeoutMillis > 0) {
-      reqb.timeout(Duration.of(idleTimeoutMillis, ChronoUnit.MILLIS));
-    }
+    reqb.timeout(Duration.of(requestTimeoutMillis, ChronoUnit.MILLIS));
     reqb.header("User-Agent", USER_AGENT);
     setBasicAuthHeader(solrRequest, reqb);
     Map<String, String> headers = solrRequest.getHeaders();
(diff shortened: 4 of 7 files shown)


solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/SolrClientCache.java
dsmiley, line 1: NOCOMMIT here; it remains to be seen how we'll solve the incompatibility of withHttpClient + with timeouts.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/ConcurrentUpdateHttp2SolrClient.java
```diff
@@ -256,7 +256,7 @@ void sendUpdateStream() throws Exception {
             }
 
             Response response =
-                responseListener.get(client.getIdleTimeout(), TimeUnit.MILLISECONDS);
+                responseListener.get(client.getRequestTimeoutMillis(), TimeUnit.MILLISECONDS);
```
dsmiley, line 259: It was weird to be using the idle (aka socket) timeout in a situation where we're waiting for the request.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -129,6 +126,12 @@ protected Http2SolrClient(String serverBaseUrl, Builder builder) {
     super(serverBaseUrl, builder);
 
     if (builder.httpClient != null) {
+      if (builder.followRedirects != null
```
dsmiley, line 129: NOCOMMIT; it remains to be seen if we're going to retain HttpClient reuse but reusing it means we can't be customizing timeouts.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -261,8 +266,6 @@ private HttpClient createHttpClient(Builder builder) {
     this.authenticationStore = new AuthenticationStoreHolder();
     httpClient.setAuthenticationStore(this.authenticationStore);
 
-    httpClient.setConnectTimeout(builder.connectionTimeoutMillis);
```
dsmiley, line 264: just reorganized this upwards

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -252,7 +255,9 @@ private HttpClient createHttpClient(Builder builder) {
     httpClient.setMaxRequestsQueuedPerDestination(
         asyncTracker.getMaxRequestsQueuedPerDestination());
     httpClient.setUserAgentField(new HttpField(HttpHeader.USER_AGENT, USER_AGENT));
-    httpClient.setIdleTimeout(idleTimeoutMillis);
+    httpClient.setConnectTimeout(builder.getConnectionTimeoutMillis());
+    httpClient.setIdleTimeout(builder.getIdleTimeoutMillis());
```
dsmiley, line 259: note: these getters now resolve a default value, a primitive; do not return null.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -514,7 +522,7 @@ public NamedList<Object> request(SolrRequest<?> solrRequest, String collection)
     try {
       InputStreamResponseListener listener = new InputStreamReleaseTrackingResponseListener();
       req = sendRequest(makeRequest(solrRequest, url, false), listener);
-      Response response = listener.get(idleTimeoutMillis, TimeUnit.MILLISECONDS);
+      Response response = listener.get(requestTimeoutMillis, TimeUnit.MILLISECONDS);
```
dsmiley, line 525: was weird to be using the idle (AKA socket) timeout in a place where, obviously, the request timeout is what should be used.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -622,11 +630,8 @@ private void setBasicAuthHeader(SolrRequest<?> solrRequest, Request req) {
   private void decorateRequest(Request req, SolrRequest<?> solrRequest, boolean isAsync) {
     req.headers(headers -> headers.remove(HttpHeader.ACCEPT_ENCODING));
 
-    if (requestTimeoutMillis > 0) {
-      req.timeout(requestTimeoutMillis, TimeUnit.MILLISECONDS);
-    } else {
-      req.timeout(idleTimeoutMillis, TimeUnit.MILLISECONDS);
-    }
+    req.timeout(requestTimeoutMillis, TimeUnit.MILLISECONDS);
```
dsmiley, line 633: We can & should keep this simple.  there are 2 clients sharing some foundation, and we now resolve a default early so here we just do something obvious.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -1020,12 +1025,6 @@ public Http2SolrClient build() {
       if (cookieStore == null) {
         cookieStore = getDefaultCookieStore();
       }
-      if (idleTimeoutMillis == null || idleTimeoutMillis <= 0) {
-        idleTimeoutMillis = (long) HttpClientUtil.DEFAULT_SO_TIMEOUT;
-      }
-      if (connectionTimeoutMillis == null) {
-        connectionTimeoutMillis = (long) HttpClientUtil.DEFAULT_CONNECT_TIMEOUT;
-      }
```
dsmiley, line 1028: No longer need such fields on the SolrClient since they are the responsibility of the HttpClient, thus they were redundant here and misleading (setting them changes nothing) IMO

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -1085,27 +1084,20 @@ public Builder withHttpClient(Http2SolrClient http2SolrClient) {
       if (this.basicAuthAuthorizationStr == null) {
         this.basicAuthAuthorizationStr = http2SolrClient.basicAuthAuthorizationStr;
       }
-      if (this.followRedirects == null) {
-        this.followRedirects = http2SolrClient.httpClient.isFollowRedirects();
-      }
-      if (this.idleTimeoutMillis == null) {
-        this.idleTimeoutMillis = http2SolrClient.idleTimeoutMillis;
+      if (this.requestTimeoutMillis == null) {
+        this.requestTimeoutMillis = http2SolrClient.requestTimeoutMillis;
```
dsmiley, line 1088: only the request timeout is owned by our SolrClient and *not* HttpClient

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpJdkSolrClient.java
```diff
@@ -86,12 +86,18 @@ public class HttpJdkSolrClient extends HttpSolrClientBase {
...
     HttpClient.Redirect followRedirects =
         Boolean.TRUE.equals(builder.followRedirects)
             ? HttpClient.Redirect.NORMAL
             : HttpClient.Redirect.NEVER;
-    HttpClient.Builder b = HttpClient.newBuilder().followRedirects(followRedirects);
+    b.followRedirects(followRedirects);
+
+    b.connectTimeout(Duration.of(builder.getConnectionTimeoutMillis(), ChronoUnit.MILLIS));
+    // note: idle timeout isn't used for the JDK client
+    // note: request timeout is set per request
```
dsmiley, line 99: * organizational change
* setting the connection timeout (was forgotten!)

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpJdkSolrClient.java
```diff
@@ -419,11 +425,7 @@ private synchronized boolean maybeTryHeadRequestSync(String url) {
   }
 
   private void decorateRequest(HttpRequest.Builder reqb, SolrRequest<?> solrRequest) {
-    if (requestTimeoutMillis > 0) {
-      reqb.timeout(Duration.of(requestTimeoutMillis, ChronoUnit.MILLIS));
-    } else if (idleTimeoutMillis > 0) {
-      reqb.timeout(Duration.of(idleTimeoutMillis, ChronoUnit.MILLIS));
-    }
+    reqb.timeout(Duration.of(requestTimeoutMillis, ChronoUnit.MILLIS));
```
dsmiley, line 428: same comment as for Http2SolrClient: we now keep this simply and do defaulting logic in one common place earlier

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpJdkSolrClient.java
```diff
@@ -553,12 +555,6 @@ public Builder(String baseSolrUrl) {
 
     @Override
     public HttpJdkSolrClient build() {
-      if (idleTimeoutMillis == null || idleTimeoutMillis <= 0) {
-        idleTimeoutMillis = (long) HttpClientUtil.DEFAULT_SO_TIMEOUT;
-      }
-      if (connectionTimeoutMillis == null) {
-        connectionTimeoutMillis = (long) HttpClientUtil.DEFAULT_CONNECT_TIMEOUT;
-      }
```
dsmiley, line 561: same comment as for Http2SolrClient: these settings were redundant/duplicative and thus misleading.  Only need to be inside the underlying HttpClient.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBase.java
```diff
@@ -57,8 +57,6 @@ public abstract class HttpSolrClientBase extends SolrClient {
   /** The URL of the Solr server. */
   protected final String serverBaseUrl;
 
-  protected final long idleTimeoutMillis;
```
dsmiley, line 60: redundant with HttpClient; removed here.

solr/solrj/src/test/org/apache/solr/client/solrj/impl/Http2SolrClientTest.java
```diff
@@ -661,14 +661,16 @@ public void testIdleTimeoutWithHttpClient() {
         assertEquals(oldClient.getIdleTimeout(), onlyBaseUrlChangedClient.getIdleTimeout());
         assertEquals(oldClient.getHttpClient(), onlyBaseUrlChangedClient.getHttpClient());
       }
-      try (Http2SolrClient idleTimeoutChangedClient =
-          new Http2SolrClient.Builder("baseSolrUrl")
-              .withHttpClient(oldClient)
-              .withIdleTimeout(3000, TimeUnit.MILLISECONDS)
```
dsmiley, line 667: can't do this pairing anymore

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
dsmiley: I see the idle timeout was correct because this only waits for the HTTP headers

solr/core/src/java/org/apache/solr/cloud/Overseer.java
```diff
@@ -862,7 +862,6 @@ private void doCompatCheck(BiConsumer<String, Object> consumer) {
             new Http2SolrClient.Builder()
                 .withHttpClient(getCoreContainer().getDefaultHttpSolrClient())
                 .withIdleTimeout(30000, TimeUnit.MILLISECONDS)
-                .withConnectionTimeout(15000, TimeUnit.MILLISECONDS)
```
dsmiley, line 865: Not supported anymore if we use an existing HttpClient.  No big deal.

solr/core/src/java/org/apache/solr/cloud/SyncStrategy.java
```diff
@@ -360,7 +360,6 @@ private void requestRecovery(
           try (SolrClient client =
               new Http2SolrClient.Builder(baseUrl)
                   .withHttpClient(solrClient)
-                  .withConnectionTimeout(30000, TimeUnit.MILLISECONDS)
```
dsmiley, line 363: Not supported anymore if we use an existing HttpClient.  No big deal.

solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
```diff
@@ -287,7 +284,6 @@ public IndexFetcher(
     String compress = (String) initArgs.get(COMPRESSION);
     useInternalCompression = ReplicationHandler.INTERNAL.equals(compress);
     useExternalCompression = ReplicationHandler.EXTERNAL.equals(compress);
-    connTimeout = getParameter(initArgs, HttpClientUtil.PROP_CONNECTION_TIMEOUT, 30000, null);
```
dsmiley, line 290: Not supported anymore if we use an existing HttpClient.  No big deal.

Here this setting comes from ReplicationHandler config in solrconfig.xml but I don't think anyone specifies this (I searched for it and didn't see in all of Solr configs/tests).

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/SolrClientCache.java
```diff
@@ -205,19 +205,13 @@ private static Http2SolrClient.Builder newHttp2SolrClientBuilder(
                 .withDefaultCollection(URLUtil.extractCoreFromCoreUrl(url));
     if (http2SolrClient != null) {
       builder.withHttpClient(http2SolrClient);
+      // cannot set connection timeout
```
dsmiley, line 208: Okay?  I suppose.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBase.java
dsmiley: and isn't supported by the Jdk HttpClient

dsmiley (commented):
I think this PR is now in good shape. 

Summary in JIRA:
>If Http2SolrClient.Builder.withHttpClient is used, (and it's used more in 9.8, shifted from deprecated Apache HttpClient based clients), settings affecting HttpClient construction cannot be customized; an IllegalStateException will now be thrown if you try.  The connection timeout cannot be customized, but idle & request timeouts can be.  Callers (various places in Solr) should no longer customize the connection timeout.
The HttpJdkSolrClient wasn't setting the connection timeout as per the builder configuration.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
```diff
@@ -115,38 +121,49 @@ public B withMaxConnectionsPerHost(int max) {
...
-   */
+  /** Set a timeout for requests to receive a response. Zero means infinite. */
   @SuppressWarnings("unchecked")
   public B withRequestTimeout(long requestTimeout, TimeUnit unit) {
     this.requestTimeoutMillis = TimeUnit.MILLISECONDS.convert(requestTimeout, unit);
     return (B) this;
   }
 
+  public long getRequestTimeoutMillis() {
+    return requestTimeoutMillis != null && requestTimeoutMillis >= 0
```
kotman12: Curious why setting a negative value defaults to the idle timeout but 0 means no timeout? I guess this is to preserve the existing behavior in Solr? The underlying Jetty http request API has `<=0`  signify no timeout.

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/SolrClientCache.java
kotman12: Should we validate that the `http2SolrClient` connectionTimeout is greater than or equal to the `minConnTimeout`? Maybe this validation should be in the constructor? There seems to be an expectation from the comments that this is the "floor" for some reason... But if this is actually irrelevant then does it make more sense to just rip out minConnTimeout altogether to avoid confusion?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -1016,28 +1017,31 @@ protected <B extends HttpSolrClientBase> B build(Class<B> type) {
...
+      } else {
+        if (followRedirects != null
+            || connectionTimeoutMillis != null
+            || maxConnectionsPerHost != null
+            || useHttp1_1 != httpClient.getTransport() instanceof HttpClientTransportOverHTTP
+            || proxyHost != null
+            || sslConfig != null
+            || cookieStore != null
+            || keyStoreReloadIntervalSecs != null) {
+          throw new IllegalArgumentException(
```
kotman12: :+1:

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
```diff
@@ -115,38 +121,49 @@ public B withMaxConnectionsPerHost(int max) {
...
   @SuppressWarnings("unchecked")
   public B withIdleTimeout(long idleConnectionTimeout, TimeUnit unit) {
     this.idleTimeoutMillis = TimeUnit.MILLISECONDS.convert(idleConnectionTimeout, unit);
     return (B) this;
   }
 
-  public Long getIdleTimeoutMillis() {
-    return idleTimeoutMillis;
+  public long getIdleTimeoutMillis() {
+    return idleTimeoutMillis != null
```
kotman12: Wasn't the original logic, effectively:
`return idleTimeoutMillis != null && idleTimeoutMillis > 0 ? idleTimeoutMillis : HttpClientUtil.DEFAULT_SO_TIMEOUT`
?

In general I think we should be very careful about default-setting the idle timeout to "forever". I believe that in a majority of cases you _don't_ want this default.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
kotman12: Actually if I am understanding the original correctly:

```
    if (requestTimeoutMillis > 0) {
      req.timeout(requestTimeoutMillis, TimeUnit.MILLISECONDS);
    } else {
      req.timeout(idleTimeoutMillis, TimeUnit.MILLISECONDS);
    }
```
\+

```
    if (builder.requestTimeoutMillis != null) {
      this.requestTimeoutMillis = builder.requestTimeoutMillis;
    } else {
      this.requestTimeoutMillis = -1;
    }
```

Then I'd expect the following to preserve existing behavior. If we want to change existing behavior we should be very explicit about it and IMO we shouldn't change it to something so complicated, i.e. we either stick with the idle timeout fallback or we go to a "forever" fallback. I don't know if a three-state default solution is really beneficial. At that point you'd want to disambiguate by explicitly setting to "forever" or idleTimeout or whatever else...

`return requestTimeoutMillis != null && requestTimeoutMillis > 0 ? requestTimeoutMillis : getIdleTimeoutMillis()`

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
dsmiley: can't customize connection timeout specifically, and other non-timeout things.

solr/solrj-streaming/src/java/org/apache/solr/client/solrj/io/SolrClientCache.java
dsmiley: I dug into this... the timeouts came from https://issues.apache.org/jira/browse/SOLR-14672 which is more about making them inheritable instead of hard-coded.  We have that already -- they are inherited from the passed in client, which in turn has values that are configurable via HttpSolrClientProvider via UpdateShardHandlerConfig via solr.xml.
I think we're fine.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
dsmiley: Good eye.  I'll change this back for consistency.

dsmiley: I didn't intend to change existing behavior, but I wanted the getXXXTimeout methods to return a >0 value, so that it's  it's clear/obvious how to interpret it _and_ so that callers could do min/max against some other value sensibly.  For example the SolrClientCache could then use `max` on and it'd behave sensibly when compared to another default.  If 0 means forever, it's semantically inverted compared to the other values.

dsmiley: WDYT @kotman12 ?  
9.9 is coming next week...

kotman12: @dsmiley were you going to change this? Regarding:

> For example the SolrClientCache could then use max on and it'd behave sensibly when compared to another default. If 0 means forever, it's semantically inverted compared to the other values.

I think the existing logic does ensure that this returns greater than 0 because it defaults to idleTimeout in the case of `<=0`. Idle timeout's original logic effectively was:

```
return idleTimeoutMillis != null && idleTimeoutMillis > 0 ? idleTimeoutMillis : HttpClientUtil.DEFAULT_SO_TIMEOUT
```

so again, unless `HttpClientUtil.DEFAULT_SO_TIMEOUT` returns a value `<=0` (which should never be the case) I don't see how you would ever be inverting the meanings of these values and so I think it would be safe to call `max` when comparing them.

The current approach in this PR is more complicated because it does double-defaulting:

```
{
    timeout < 0: idleTimeout
    timeout = 0: Integer.MAX_VALUE
    timeout > 0: timeout
}
```

I don't think this is really useful? Maybe I'm still missing something.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
```diff
@@ -25,6 +25,12 @@
 
 public abstract class HttpSolrClientBuilderBase<
     B extends HttpSolrClientBuilderBase<?, ?>, C extends HttpSolrClientBase> {
+  /**
+   * About 24 days in milliseconds -- basically forever to wait for something. But not so large to
+   * cause overflow.
+   */
+  protected static final long FOREVER_MILLIS = Integer.MAX_VALUE;
```
kotman12: I think it would be useful to expose this in some way. Either by making this public or having a function in the builder which does this for you, i.e.:

```
  public B withNoIdleTimeout() {
    this.idleTimeoutMillis = FOREVER_MILLIS;
    return (B) this;
  }
```

kotman12: @dsmiley if we are really worried about overflow here, should we bite the bullet and make this an int? This fallback value doesn't really prevent overflows since I could just do:

```
.withIdleTimeout(Long.MAX_VALUE, TimeUnit.MILLISECONDS)
```
which I incidentally did [here](https://github.com/apache/solr/pull/3356/files). To a future reviewer a change like this may seem reasonable, in fact you approved it and you are reasonable 😄 . But if this will overflow in some low-level lossy cast to int then this is still dangerous and the `FOREVER_MILLIS`, although well-intentioned, just adds noise IMO without really providing much protection.

dsmiley: I'd rather not add yet another method for the ~infinite/forever case.  I did see an issue with Long.MAX_VALUE and a test failed clearly related to overflow... if I recall it was with the JDK HttpClient.  Any way, we could ignore the matter if it doesn't show up again.  If it does show up again... I could see either switching to integer or reducing the value on `setXXX(...)`

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -129,6 +128,20 @@ protected Http2SolrClient(String serverBaseUrl, Builder builder) {
     super(serverBaseUrl, builder);
 
     if (builder.httpClient != null) {
+      // Validate that no conflicting options are provided when using an existing HttpClient
+      if (builder.followRedirects != null
+          || builder.connectionTimeoutMillis != null
+          || builder.maxConnectionsPerHost != null
+          || builder.useHttp1_1
```
HoustonPutman, line 135: (Posting on the PR instead of the commit)

`useHttp1_1` behaves differently than the rest of these, because it is not a default as Null option. So if someone wants to customize `useHttp1`, while not changing the sysProp, this will fail.

I'll make a PR so that `useHttp1_1` is a Boolean and defaults to null, like the rest of these.

(This was breaking some of the benchmark suites)

dsmiley: Thanks for noticing / fixing!  This was gnawing me a little but I didn't see the bug.  I was tempted to use a `Boolean` and to resolve the correct choice in a getter, like I did for the timeouts.

PR #3358, opened 2025-05-20
Explicitly disable dev/prod tasks based on flag

malliaridis:
# Description

With the `production` build flag set to `false`, gradle may fail when running `gradlew assemble` with an error due to some invalid task dependencies coming from the Compose MP plugin. The reported message is similar to the below:

See also discussion in https://lists.apache.org/thread/1l25p9cd9h8ch32jt2ljpmdt3fc9bctq

```
...
> Task :solr:ui:wasmJsProductionExecutableCompileSync FAILED

* What went wrong:
A problem was found with the configuration of task ':solr:ui:wasmJsProductionExecutableCompileSync' (type 'DefaultIncrementalSyncTask').
  - Gradle detected a problem with the following location: '/Users/christos/workspace/apache/solr/build/js/packages/composeApp/kotlin'.
    
    Reason: Task ':solr:ui:wasmJsBrowserDevelopmentWebpack' uses this output of task ':solr:ui:wasmJsProductionExecutableCompileSync' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':solr:ui:wasmJsProductionExecutableCompileSync' as an input of ':solr:ui:wasmJsBrowserDevelopmentWebpack'.
      2. Declare an explicit dependency on ':solr:ui:wasmJsProductionExecutableCompileSync' from ':solr:ui:wasmJsBrowserDevelopmentWebpack' using Task#dependsOn.
      3. Declare an explicit dependency on ':solr:ui:wasmJsProductionExecutableCompileSync' from ':solr:ui:wasmJsBrowserDevelopmentWebpack' using Task#mustRunAfter.
    
    For more information, please refer to https://docs.gradle.org/8.10/userguide/validation_problems.html#implicit_dependency in the Gradle documentation.
```

The error is likely thrown because by default both development and production artifacts are generated on `assemble`. In the `webapp` module, we explicitly depend on tasks and include their outputs based on the variant configured by the `production` flag, leading to potentially different artifacts being included based on execution order.

The current workaround was to set the `production` flag to `true`.

# Solution

This PR solves the issue of the invalid task dependencies and allows the `production` flag to be set to `false` again without failing on `gradlew assemble`.

The `production` flag can be set to `false` again on our build servers after merging this.

# Tests

_No tests written for gradle execution_

# Checklist

### DIFF
--- a/solr/ui/build.gradle.kts
+++ b/solr/ui/build.gradle.kts
@@ -174,3 +174,21 @@ tasks.matching { task ->
     // and you will see the test tasks being listed there as well, but they will be skipped as
     // expected
 }
+
+// Explicitly enable or disable development tasks based on the build variant
+// This prevents any invalid task dependencies on assemble task execution
+tasks.matching {
+    val taskName = it.name.lowercase()
+    taskName.contains("wasmjs") && taskName.contains("development")
+}.configureEach {
+    onlyIf { rootProject.ext["development"] as Boolean }
+}
+
+// Explicitly enable or disable production tasks based on the build variant
+// This prevents any invalid task dependencies on assemble task execution
+tasks.matching {
+    val taskName = it.name.lowercase()
+    taskName.contains("wasmjs") && taskName.contains("production")
+}.configureEach {
+    onlyIf { !(rootProject.ext["development"] as Boolean) }
+}


malliaridis:
@gerlowskija I've added you as a reviewer as you have reported this issue on slack.

dsmiley:
So these tasks like `wasmJsProductionExecutableCompileSync` are not defined by our project but are defined by external plugins we use, which have the notion of "production" artifacts and maybe "development" artifacts as well?  And that which are mutually exclusive?

(just asking/stating this for confirmation)

malliaridis:
> So these tasks like wasmJsProductionExecutableCompileSync are not defined by our project but are defined by external plugins we use, which have the notion of "production" artifacts and maybe "development" artifacts as well? 

Yes, that's what I could figure out from the gradle tasks outputed with `--dry-run`.

> And that which are mutually exclusive?

I believe so, yes. During `assemble`, both development and `production` tasks are executed and artifacts are generated in that order, so that `production` override the `development` artifacts. However, when we assemble with our custom gradle flag `production=false`, we include and depend on artifacts from the `development`. In my understanding, this is causing the error.

malliaridis:
> Why would a Solr dev change the "production" setting?

_tl;dr_
The distinction between production and development artifacts is coming from the Compose gradle plugin and currently only affects the new UI / webapp. It is used for optimized builds of the new UI.

_Some Background_
I believe the two build types originate from Android, where mobile apps have a debug and a release variant. The debug variant is used for faster build times and allow developers to quickly deploy code changes to devices and emulators. The release variant is normally configured to optimize the artifacts shipped to end users, so they are smaller in size and code is obfuscated to make reverse engineering harder.

_ Our Case_
In our case, the development setting has quicker build times, whereas the production setting optimizes the shipped files for web-assembly (smaller in size and strict security rules in webapp). I believe there are also easy ways to attach a debugger during development, which is not possible when production artifacts are generated (obfuscated and shrinked code, stricter security rules makes it harder).

The initial problem started when we integrated specific build files to the webapp and wanted to optimize build times by using development artifacts during development (so that devs don't have to wait for the code shrinking of production artifacts), and production artifacts for Solr releases. However, this had as an effect, that some tasks from the build chain related to production builds were still executed and were overwriting some other files (according to gradle) when we were using development builds (and / or vice-versa). Some of the multiplatform libraries may also not work if security rules are strict (e.g. ktor).

_What we likely want_
For the webapp it is important to ship optimized files (development flag `true`) for the new UI, as the users will have longer loading times if the files are large, and also less secure webapps if the stricter rules are not applied. The distinction can also be used in future if we want to optimize other parts of our project for production  or for containers for example (so that Solr distribution is minimal). But further use cases have not been explored yet.

During development (development flag `true`), having fast build times, hot reloads (not available yet) and similar features are very useful, especially during UI development.

epugh:
@gerlowskija remember how I'd love to have Solr start with a couple of modes?  You could imagine a `bin/start --mode=developer` and in that mode, the UI would be in the "dev" version, while any other mode it would be in "prod" version.

dsmiley:
Eric, this PR is about the build having a mode, not Solr itself.

PR #3359, opened 2025-05-22
fix-tmp-symlink

risdenk:

# Description

On Mac /tmp is a symlink and this causes issues with the Solr security manager. SOLR-16457 / https://github.com/apache/solr/pull/1282 fixed a similar issue with symlinks and the security manager. This applies the same type of solution but specifically for `java.io.tmpdir` to ensure that on Macs (and Linux) the full path is used not the symlink. This makes the security manager checking the path happy since its not trying to compare to a symlink anymore.

# Checklist

### DIFF
--- a/solr/bin/solr
+++ b/solr/bin/solr
@@ -1314,6 +1314,7 @@ function start_solr() {
   SOLR_START_OPTS=('-server' "${JAVA_MEM_OPTS[@]}" "${GC_TUNE_ARR[@]}" "${GC_LOG_OPTS[@]}" "${IP_ACL_OPTS[@]}" \
     "${REMOTE_JMX_OPTS[@]}" "${CLOUD_MODE_OPTS[@]}" -Dsolr.log.dir="$SOLR_LOGS_DIR" \
     "-Djetty.port=$SOLR_PORT" "-DSTOP.PORT=$stop_port" "-DSTOP.KEY=$STOP_KEY" \
+    "-Djava.io.tmpdir=$(cd $TMPDIR; pwd -P)" \
     # '-OmitStackTraceInFastThrow' ensures stack traces in errors,
     # users who don't care about useful error msgs can override in SOLR_OPTS with +OmitStackTraceInFastThrow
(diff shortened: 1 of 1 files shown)


solr/bin/solr
```diff
@@ -1314,6 +1314,7 @@ function start_solr() {
   SOLR_START_OPTS=('-server' "${JAVA_MEM_OPTS[@]}" "${GC_TUNE_ARR[@]}" "${GC_LOG_OPTS[@]}" "${IP_ACL_OPTS[@]}" \
     "${REMOTE_JMX_OPTS[@]}" "${CLOUD_MODE_OPTS[@]}" -Dsolr.log.dir="$SOLR_LOGS_DIR" \
     "-Djetty.port=$SOLR_PORT" "-DSTOP.PORT=$stop_port" "-DSTOP.KEY=$STOP_KEY" \
+    "-Djava.io.tmpdir=$(cd $TMPDIR; pwd -P)" \
```
risdenk, line 1317: note this was enough to fix `./bin/solr start` but looks like SOLR_START_OPTS might not be used everywhere. for example in `stop_solr` or `run_tool` but we could apply a similar fix there.

epugh (approved):
So, this makes total sense.   My only thought is that it's adding a bit more complexity to this already incredibly complex script!  If we removed Security Manager, would that obviate the need for this?   And, if that statement is true, how can we make sure to back this change out in the future when that happens?

malliaridis (approved):
Tested on a Mac and this fixes the issue. It seems that you also covered all places in the solr script, good job! 👍 

Have you checked if the [cross-dc-manager](https://github.com/apache/solr/blob/8188bf5e0f37ca929e4cebce9e73e50d5d07d86d/solr/cross-dc-manager/bin/cross-dc-manager) may be also affected by this bug? I never used the tool before to know how the script works.

risdenk:
@malliaridis From a quick look, I don't think [cross-dc-manager](https://github.com/apache/solr/blob/8188bf5e0f37ca929e4cebce9e73e50d5d07d86d/solr/cross-dc-manager/bin/cross-dc-manager) sets the security manager so wouldn't be affected by this.

PR #3363, opened 2025-05-24
SOLR-17756: Explicit cast on IndexFingerprint for Java 11

mlbiscoc:
IndexFingerprint needs a separate explicit cast on Java 11 after instanceof check.


mlbiscoc:
@dsmiley Just saw you beat me to it fixing in commit [2c82ba9](https://github.com/apache/solr/commit/2c82ba9f9bc64fd1e574c042b50a79f796a73363). Thanks for that. Will close this out.

PR #3378, opened 2025-05-29
Tests for examples failing on Windows

rahulgoswami:

# Description

Running gradlew -p solr/core test --tests "org.apache.solr.cli.TestSolrCLIRunExample" on Windows causes tests for techproducts, films and other examples to fail.

# Solution

Tests initiated via TestSolrCLIRunExample use a custom executor (RunExampleExecutor). The execute(cmd) method of this executor executes one logic of starting an embedded test solr cluster when the script is bin/solr and a default alternate flow otherwise. This condition should instead check for either of bin/solr or bin/solr.cmd since starting a test cluster in embedded mode essentially makes this flow platform independent.

One could argue that the way we spawn a test cluster for these tests doesn't *exactly* replicate a CLI flow, but that could be tackled as a separate fix to make the tests *better*.

# Tests

gradlew -p solr/core test --tests "org.apache.solr.cli.TestSolrCLIRunExample" succeeds on Windows.

# Checklist

### DIFF
--- a/solr/core/src/test/org/apache/solr/cli/TestSolrCLIRunExample.java
+++ b/solr/core/src/test/org/apache/solr/cli/TestSolrCLIRunExample.java
@@ -102,7 +102,7 @@ public class TestSolrCLIRunExample extends SolrTestCaseJ4 {
       commandsExecuted.add(cmd);
 
       String exe = cmd.getExecutable();
-      if (exe.endsWith("solr")) {
+      if (exe.endsWith("solr") || exe.endsWith("solr.cmd")) {
         String[] args = cmd.getArguments();
         if ("start".equals(args[0])) {
           if (hasFlag("--user-managed", args)) {


solr/core/src/test/org/apache/solr/cli/TestSolrCLIRunExample.java
```diff
@@ -102,7 +102,7 @@ public int execute(org.apache.commons.exec.CommandLine cmd) throws IOException {
       commandsExecuted.add(cmd);
 
       String exe = cmd.getExecutable();
-      if (exe.endsWith("solr")) {
+      if (exe.endsWith("solr") || exe.endsWith("solr.cmd")) {
```
epugh, line 105: somethign I am confused by....   How would the exe ever NOT be either solr or solr.cmd???   Is there a way this code can be run that doesn't include one of those?  I am just wondering if we need this if statement?

epugh: i looked at the class...  and it's super specific to testing Solr CLI, so I can't imagine a world where anything BUT a solr or solr.cmd is used???   In fact, I think an assert would be more apprporiate, to fail the test if you used this class with anything but solr or solr.cmd?

epugh: Love it!   Thanks for the explanation and the simplification.  I love elminating if loops.

epugh:
Can you add a changes entry?  And then when the tests pass I'll merge.

PR #3379, opened 2025-05-29
SOLR-17774: Refactor initialization of GlobalOpenTelemetry

mlbiscoc:
This refactors how the Global Open Telemetry instance was set and adds `opentelemetry-exporter-prometheus` and `opentelemetry.sdk` into core as a dependency. `opentelemetry-exporter-prometheus` is needed as it gives us a MetricReader to collect metrics in Prometheus format that can be exposed to the /admin/metrics endpoint.

Initially, `GlobalOpenTelemetry` was only set for accepting tracing and spans but was not dynamic enough to configure it for metrics or even logs down the line.

I renamed `TracerConfigurator` to `OpenTelemetryConfigurator` accordingly.

I also moved the initialization to of setting the `GlobalOpenTelemetry` to the constructor of `coreContainer` so it is called as early as possible. It was initially inside of `loadInternal()` which threw an exception for metrics because there are metrics being recorded on loading of coreContainer.


solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -423,6 +433,11 @@ public CoreContainer(NodeConfig config, CoresLocator locator, boolean asyncSolrC
     this.solrHome = config.getSolrHome();
     this.solrCores = SolrCores.newSolrCores(this);
     this.nodeKeyPair = new SolrNodeKeyPair(cfg.getCloudConfig());
+    this.prometheusMetricReader = new PrometheusMetricReader(true, null);
+    initializeOpenTelemetrySdk();
+    this.tracer = TraceUtils.getGlobalTracer();
+    this.meterProvider = MetricUtils.getMeterProvider();
```
mlbiscoc, line 439: This was initially done with `TracerConfigurator.loadTracer` inside of `loadInternal` Moved it up to be called earlier.

solr/core/src/java/org/apache/solr/core/OpenTelemetryConfigurator.java
```diff
@@ -46,26 +50,53 @@ public abstract class TracerConfigurator implements NamedListInitializedPlugin {
...
-      return configurator.getTracer();
-    }
-    if (shouldAutoConfigOTEL()) {
-      return autoConfigOTEL(loader);
-    }
+  private static volatile boolean loaded = false;
+
+  public static synchronized void configureOpenTelemetrySdk(
+      SdkMeterProvider sdkMeterProvider, SdkTracerProvider sdkTracerProvider) {
+    if (loaded) return;
```
mlbiscoc, line 57: Its a bit annoying Open Telemetry doesn't have a method to call if the `GlobalOpenTelemetry` was already set. So I understand why this boolean loaded existed on `SimplePropagator` now.

solr/core/src/java/org/apache/solr/util/tracing/SimplePropagator.java
```diff
@@ -54,19 +50,6 @@ public class SimplePropagator implements TextMapPropagator {
...
-  public static synchronized Tracer load() {
-    if (!loaded) {
-      log.info("OpenTelemetry tracer enabled with simple propagation only.");
-      OpenTelemetry otel =
-          OpenTelemetry.propagating(ContextPropagators.create(SimplePropagator.getInstance()));
-      GlobalOpenTelemetry.set(otel);
-      loaded = true;
-    }
-    return TraceUtils.getGlobalTracer();
-  }
```
mlbiscoc, line 68: This basically moved into the new `OpenTelemetryConfigurator`

mlbiscoc (commented):
@dsmiley Could you take a look at this? There were a 2-3 tests failing with this change that I will take a look but this should be reviewable.  This is also going into the separate feature branch not on main.

janhoy (commented):
Some quick comments.

You may want to run `gradle updateLicenses` as well. I was not convinced that the sdk-logs lib was really excluded?

mlbiscoc (commented):
Thanks for the quick review @janhoy. I updated the licenses but I cannot fully remove the sdk-logs. Doing the exclusion only removes it as a `compileClasspath`.

Tests should all be passing locally now as well.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -465,6 +480,19 @@ public CoreContainer(NodeConfig config, CoresLocator locator, boolean asyncSolrC
             new SolrNamedThreadFactory("IndexFingerprintPool"));
   }
 
+  private void initializeOpenTelemetrySdk() {
+    PluginInfo info = cfg.getTracerConfiguratorPluginInfo();
+
+    if (OpenTelemetryConfigurator.shouldAutoConfigOTEL()) {
+      OpenTelemetryConfigurator.autoConfigureOpenTelemetrySdk(loader);
+    } else if (info != null && info.isEnabled()) {
+      OpenTelemetryConfigurator.loadOpenTelemetrySdk(loader, cfg.getTracerConfiguratorPluginInfo());
```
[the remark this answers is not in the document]
janhoy: Thanks. I like the similar naming better.

mlbiscoc:
Going to bump this if any one interesting in looking? I have another PR almost ready that implements the foundation for working OTEL metrics.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
dsmiley: A growing concern on my mind is that CoreContainer is a kitchen sink and has gotten out of control.  That said, if you need (or makes most logical sense) for a field on CoreContainer, really, then fine.  If you can keep OTEL stuff to another source file in some way, that would probably be ideal.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -465,6 +480,39 @@ public CoreContainer(NodeConfig config, CoresLocator locator, boolean asyncSolrC
             new SolrNamedThreadFactory("IndexFingerprintPool"));
   }
 
+  /**
```
dsmiley: Docs are always well intentioned but here, I think this is saying way too much.  It's about as much as there is code.  This is a private method; why bother?  Docs need to be maintained as well; it's not a free gift (from whatever person/LLM wrote it).

solr/core/src/java/org/apache/solr/core/OpenTelemetryConfigurator.java
dsmiley: Agreed; I said the same in my OTEL Agent PR.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
mlbiscoc: Thats understandable. Moved it out of `coreContainer` and into `MetricManager` which makes the most sense. `MetricManager` creates instruments for measuring metrics and so also putting the `MetricReader` and `MeterProvider` in there as well was perfect.

`initializeOpenTelemetrySdk` was moved into the `OpenTelemetryConfigurator`

solr/core/src/java/org/apache/solr/core/CoreContainer.java
janhoy: I asked for docs since there was a regression in the first pass. I agree Javadocs primarily is for public APIs, so I'm happy if the intended order of initialization is documented in a different way, e.g.  inline the if-then-else logic as oneline `//` comments. Just short and concise to make the flow easier to grok.

mlbiscoc:
Any other comments? Will merge in the next few days otherwise and open the next PR.

PR #3380, opened 2025-05-31
SOLR-17775: optimize ValueSourceAugmenter

yurkor:
https://issues.apache.org/jira/browse/SOLR-17775

# Description
Optimize ValueSourceAugmenter performance by pre-calculating function values during setContext() instead of calculating them on-demand for each document.

# Solution
Collect all document IDs from DocList during setContext(), sort them, and calculate function values sequentially by reader segment. Store results in a map for fast lookup during transform(). Include fallback for documents not in the pre-calculated set.

# Tests
Existing test suite covers ValueSourceAugmenter functionality and all tests pass. Tested in Solr Cloud setup with function queries as pseudo-fields (e.g., &fl=sum(x,y),id,score) and verified performance improvement 

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/response/transform/ValueSourceAugmenter.java
+++ b/solr/core/src/java/org/apache/solr/response/transform/ValueSourceAugmenter.java
@@ -40,16 +40,18 @@ import org.apache.solr.search.SolrIndexSearcher;
  * @since solr 4.0
  */
 public class ValueSourceAugmenter extends DocTransformer {
-
+  private static final Object NULL_SENTINEL = new Object();
   public final String name;
   public final QParser qparser;
   public final ValueSource valueSource;
-  private static final Object NULL_SENTINEL = new Object();
+  private final int maxPrefetchSize;
 
   public ValueSourceAugmenter(String name, QParser qparser, ValueSource valueSource) {
     this.name = name;
     this.qparser = qparser;
     this.valueSource = valueSource;
+    String maxPrefetchSizeRaw = qparser.getParam("preFetchDocs");
+    this.maxPrefetchSize = maxPrefetchSizeRaw != null ? Integer.parseInt(maxPrefetchSizeRaw) : 1000;
   }
 
   @Override
@@ -70,14 +72,15 @@ public class ValueSourceAugmenter extends DocTransformer {
         return;
       }
 
-      final int[] ids = new int[docList.size()];
+      final int prefetchSize = Math.min(docList.size(), maxPrefetchSize);
+      final int[] ids = new int[prefetchSize];
       int i = 0;
       var iter = docList.iterator();
-      while (iter.hasNext()) {
+      while (iter.hasNext() && i < prefetchSize) {
         ids[i++] = iter.nextDoc();
       }
       Arrays.sort(ids);
-      scores = new IntObjectHashMap<>(ids.length);
+      cachedValuesById = new IntObjectHashMap<>(ids.length);
 
       FunctionValues values = null;
       int docBase = -1;
@@ -92,7 +95,7 @@ public class ValueSourceAugmenter extends DocTransformer {
         }
         int localId = docid - docBase;
         var value = values.objectVal(localId);
-        scores.put(docid, value != null ? value : NULL_SENTINEL);
+        cachedValuesById.put(docid, value != null ? value : NULL_SENTINEL);
       }
     } catch (IOException e) {
       throw new SolrException(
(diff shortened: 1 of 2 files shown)


mkhludnev (approved):

mkhludnev:
I like it. @devs please have a glance at it. I'm ready to merge, but prefer the second opinion. 
@yurkor thank you. 
Would you mind to add it into solr/CHANGES.txt 9.9.0 Optimizations ?

yurkor:
> I like it. @devs please have a glance at it. I'm ready to merge, but prefer the second opinion. @yurkor thank you. Would you mind to add it into solr/CHANGES.txt 9.9.0 Optimizations ?

Added to CHANGES.txt

dsmiley (commented):
Using `parser.getLocalParams()` (which can be null BTW), you could check for a custom tuning param like "preFetchDocs".  This provides an advanced tuning option.  Don't have to document if it seems unlikely to be used.

yurkor:
@dsmiley added use of localParams

solr/core/src/java/org/apache/solr/response/transform/ValueSourceAugmenter.java
```diff
@@ -38,14 +40,23 @@
...
 
   public ValueSourceAugmenter(String name, QParser qparser, ValueSource valueSource) {
     this.name = name;
     this.qparser = qparser;
     this.valueSource = valueSource;
+    var localParams = qparser.getLocalParams();
+    maxPrefetchSize =
+        localParams != null
+            ? localParams.getInt("preFetchDocs", MAX_PREFETCH_DEFAULT)
+            : MAX_PREFETCH_DEFAULT;
```
dsmiley: oh I recall qparser.getParam(str) now, so you don't have to do the null check and it'll look in the top level request.  It doesn't have the integer parse but you could use `"1000"` (as a literal; no constant needed!)

solr/core/src/java/org/apache/solr/response/transform/ValueSourceAugmenter.java
```diff
@@ -38,14 +40,23 @@
  * @since solr 4.0
  */
 public class ValueSourceAugmenter extends DocTransformer {
+
   public final String name;
   public final QParser qparser;
   public final ValueSource valueSource;
+  private static final Object NULL_SENTINEL = new Object();
```
dsmiley: statics should be defined above

solr/core/src/java/org/apache/solr/response/transform/ValueSourceAugmenter.java
```diff
@@ -61,32 +67,66 @@ public void setContext(ResultContext context) {
...
+
+      final int prefetchSize = Math.min(docList.size(), maxPrefetchSize);
+      final int[] ids = new int[prefetchSize];
+      int i = 0;
+      var iter = docList.iterator();
+      while (iter.hasNext() && i < prefetchSize) {
+        ids[i++] = iter.nextDoc();
+      }
+      Arrays.sort(ids);
+      scores = new IntObjectHashMap<>(ids.length);
```
dsmiley: Why is this called "scores"?  I suggest a name like "cachedValuesById".

solr/core/src/java/org/apache/solr/response/transform/ValueSourceAugmenter.java
```diff
@@ -61,32 +67,66 @@ public void setContext(ResultContext context) {
...
-      FunctionValues values = valueSource.getValues(fcontext, rcontext);
-      int localId = docid - rcontext.docBase;
-      setValue(doc, values.objectVal(localId));
-    } catch (IOException e) {
-      throw new SolrException(
-          SolrException.ErrorCode.SERVER_ERROR,
-          "exception at docid " + docid + " for valuesource " + valueSource,
-          e);
+  public void transform(SolrDocument doc, int docid, DocIterationInfo docIterationInfo) {
+    Object scoreValue = (scores != null) ? scores.get(docid) : null;
```
dsmiley: `cacheValue` would be a better name

dsmiley:
Great! 
(BTW, stop force-pushing to PRs.  It resets the GH review state, therefore, as a reviewer, I can't as easily clearly see your changes from one change to the next).
CHANGES.txt:  In 9.9, in the optimizations section, and perhaps this simple sentence:
> Speed up returning fields ("fl" param) with DocValues referenced directly or indirectly via function queries.  (your name here)

dsmiley:
OMG I'm slipping... you already have a CHANGES.txt.  Nonetheless my feedback shows a massive difference between the actual code change (what you wrote about there) and how users understand/experience the optimization.  Only a deep diving Solr hacker would know of ValueSourceAugmenter.  An end-user has no clue what that means.  **The audience of CHANGES.txt is users of Solr**

yurkor:
@dsmiley Done. Will keep in mind - no force pushes.

dsmiley:
I mispoke/misunderstood... thinking this would work for a `fl=aDvField` when no, that's handled by `SolrDocumentFetcher` `org.apache.solr.search.SolrDocumentFetcher#decorateDocValueFields`.  That was somewhat recently optimized by @magibney (see it's use of `DocValuesIteratorCache`).  I can update this PR CHANGES.txt to say "function queries in 'fl' param".

PR #3381, opened 2025-06-09
SOLR-17678: Change matchScore to originalScore, make it a fake-function

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17678

The functionality was already added in https://github.com/apache/solr/pull/3222, but this is making the following changes before any release has been made:

- Rename from `matchScore` to `originalScore`, since that is the name that LTR queries already use for this feature. And eventually, the two implementations can hopefully be merged.
- Make the return field `originalScore()` to resemble a function. It is not currently a function, but in the future we want to use a real ValueSource function instead of hardcoding this option in Solr. The ticket that covers this is [SOLR-17784](https://issues.apache.org/jira/browse/SOLR-17784).

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -208,7 +208,7 @@ New Features
 
 * SOLR-17714: Added a FuzzyQParser to enable all FuzzyQuery customizations. (Houston Putman, Siju Varghese)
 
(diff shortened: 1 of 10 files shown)


sijuv (approved):

dsmiley (approved):
Thanks for accommodating my suggestion

PR #3382, opened 2025-06-10
SOLR-17777: Fix leaked metric registry after core unload

psalagnac:
https://issues.apache.org/jira/browse/SOLR-17777

# Description

This fixes a minor memory leak. I don't think it can have an actual impact on heap, but the fix is trivial and not risky.

# Solution

This makes sure we don't re-create a registry when none exists during core unload.

# Tests

Added a new test to validate we don't have the core registry anymore after unload.


dsmiley:
@sigram tagging you as a suitable reviewer

dsmiley (approved):

PR #3384, opened 2025-06-11
SOLR-17785: Integrate foundational OTEL meter instruments

mlbiscoc:
Integrate OTEL into `SolrMetricManager` and `SolrMetricContext` and update the`wt=prometheus` endpoint to output Prometheus metrics from OTEL. This still keeps Dropwizard in parallel for now as removing many of its functions breaks many tests and the scope of this PR gets too large in changes.

To summarize all changes below:
1. `SolrMetricManager` and `SolrMetricContext` can be used to create OTEL meter instruments such as a `LongCounter` to record metrics.
2. `SolrMetricProducer#initializeMetrics` takes a set of `Attributes` that can be used with initializing OTEL metrics eventually replacing scope.
3. Created `AttributedLongCounter`, `AttributedDoubleCounter`, etc. This binds a set of attributes to a metric working similarly to how Dropwizard initializes. This also avoids having to rebuild Attributes every time metrics were to be captured.
4. Add OTEL equivalent metric capturing from Dropwizard into `RequestHandlerBase` by updating its `initializeMetrics` and its corresponding  `RequestHandlerBaseTest`
5. `PrometheusResponseWriter` and `PrometheusFormatter` code completely removed as OTEL->Prometheus exporter already exists with the OTEL SDK.
6. A number of `TODO` comments of what and where work is still needed for OTEL migration

Sample Prometheus output from `admin/metrics?wt=prometheus` OTEL -> Prometheus

```
solr_metrics_core_requests_total{category="QUERY",collection="foobar",core="foobar_shard1_replica_n1",internal="true",otel_scope_name="solr.core.foobar.shard1.replica_n1",replica="replica_n1",scope="/select",shard="shard1",type="requests"} 0.0
solr_metrics_core_requests_total{category="QUERY",collection="foobar",core="foobar_shard1_replica_n1",internal="true",otel_scope_name="solr.core.foobar.shard1.replica_n1",replica="replica_n1",scope="/select",shard="shard1",type="serverErrors"} 0.0
solr_metrics_core_requests_total{category="QUERY",collection="foobar",core="foobar_shard1_replica_n1",internal="true",otel_scope_name="solr.core.foobar.shard1.replica_n1",replica="replica_n1",scope="/select",shard="shard1",type="timeouts"} 0.0
solr_metrics_core_requests_total{category="QUERY",collection="foobar",core="foobar_shard1_replica_n1",internal="true",otel_scope_name="solr.core.foobar.shard1.replica_n1",replica="replica_n1",scope="/terms",shard="shard1",type="clientErrors"} 0.0
solr_metrics_core_requests_total{category="QUERY",collection="foobar",core="foobar_shard1_replica_n1",internal="true",otel_scope_name="solr.core.foobar.shard1.replica_n1",replica="replica_n1",scope="/terms",shard="shard1",type="errors"} 0.0
```


solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -193,14 +203,79 @@ public static class HandlerMetrics {
     public final Timer requestTimes;
     public final Counter totalTime;
 
-    public HandlerMetrics(SolrMetricsContext solrMetricsContext, String... metricPath) {
+    public AttributedLongCounter otelRequests;
+    public AttributedLongCounter otelNumErrors;
+    public AttributedLongCounter otelNumServerErrors;
+    public AttributedLongCounter otelNumClientErrors;
+    public AttributedLongCounter otelNumTimeouts;
+    public AttributedLongTimer otelRequestTimes;
```
mlbiscoc: These are essentially equivalent to Dropwizards metrics above. Once we do deprecations, we can just change the name to the ones above and remove them.

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -193,14 +203,79 @@ public static class HandlerMetrics {
...
+          solrMetricsContext.longHistogram(
+              "solr_metrics_core_requests_times", "HTTP Solr request times", "ms");
+
+      otelRequests =
+          new AttributedLongCounter(
+              baseRequestMetric,
+              Attributes.builder()
+                  .putAll(attributes)
+                  .put(AttributeKey.stringKey("type"), "requests")
+                  .build());
```
mlbiscoc: These are the base metrics. Then you bind them to attributes so we don't need to keep rebuilding them every time we record an event.

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -193,14 +203,79 @@ public static class HandlerMetrics {
...
+                  .put(AttributeKey.stringKey("type"), "timeouts")
+                  .build());
+
+      otelRequestTimes = new AttributedLongTimer(baseRequestTimeMetric, attributes);
+
+      otelRequests.record(0L);
+      otelNumErrors.record(0L);
+      otelNumServerErrors.record(0L);
+      otelNumClientErrors.record(0L);
+      otelNumTimeouts.record(0L);
```
mlbiscoc: I record 0's so after initializing so it creates the metrics otherwise, if a handler like `export` is never called, you'll never actually see the metric. Its debatable if this should actually be here and I am thinking it shouldn't now. Metrics should probably only exist if a handler is actually being used, otherwise you bloat the response with 0's.

solr/core/src/java/org/apache/solr/handler/admin/MetricsHandler.java
```diff
@@ -127,11 +126,12 @@ private void handleRequest(SolrParams params, BiConsumer<String, Object> consume
       return;
     }
 
+    // TODO SOLR-17458: Make this the default option after dropwizard removal
     if (PROMETHEUS_METRICS_WT.equals(params.get(CommonParams.WT))) {
-      response = handlePrometheusExport(params);
-      consumer.accept("metrics", response);
+      consumer.accept("metrics", metricManager.getMetricReader().collect());
```
mlbiscoc, line 131: Instead of going through the whole dropwizard -> prometheus formatter logic. Very simply the metric reader does the OTEL -> prometheus for us through the SDK.

solr/core/src/java/org/apache/solr/handler/component/SearchHandler.java
```diff
@@ -152,18 +153,28 @@ public void init(PluginInfo info) {
...
-  public void initializeMetrics(SolrMetricsContext parentContext, String scope) {
-    super.initializeMetrics(parentContext, scope);
+  public void initializeMetrics(
+      SolrMetricsContext parentContext, Attributes attributes, String scope) {
+    super.initializeMetrics(
+        parentContext,
+        Attributes.builder()
+            .putAll(attributes)
+            .put(AttributeKey.stringKey("category"), getCategory().toString())
+            .put(AttributeKey.stringKey("internal"), "false")
```
mlbiscoc, line 165: Instead of having the handler do something like `/select[shard]`, we use a tag saying `internal` true/false

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
```diff
@@ -127,54 +131,6 @@ public void testPrometheusStructureOutput() throws Exception {
     }
   }
 
-  public void testPrometheusDummyOutput() throws Exception {
```
mlbiscoc, line 130: Pretty much all Prometheus Formmater tests go away with the exception of `testPrometheusStructureOutput` which tests for prometheus exposition format from the OTEL sdk.

solr/core/src/java/org/apache/solr/blockcache/Metrics.java
```diff
@@ -54,8 +55,10 @@ public class Metrics extends SolrCacheBase implements SolrInfoBean {
   private SolrMetricsContext solrMetricsContext;
   private long previous = System.nanoTime();
 
+  // TODO SOLR-17458: Migrate to Otel
```
dsmiley, line 58: The Lucene & Solr projects use the magic word "NOCOMMIT" not TODO.  TODO means someday maybe.  NOCOMMIT means, do NOT merge this into a release branch.  "precommit" will fail.

But since you've written this so many times already, we could leave the existing ones as-is and remember to search for "TODO SOLR-17458"

solr/core/src/java/org/apache/solr/handler/ReplicationHandler.java
```diff
@@ -837,9 +838,11 @@ private CommitVersionInfo getIndexVersion() {
   }
 
   // TODO: Handle compatibility in 8.x
```
dsmiley: This comment adjacent to what you are adding seem obsolete; feel free to replace them

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
dsmiley: please remove.  I've argued in the past that we shouldn't be dumping metrics out for stuff that isn't used.

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -193,14 +203,79 @@ public static class HandlerMetrics {
...
+
+      otelRequests =
+          new AttributedLongCounter(
+              baseRequestMetric,
+              Attributes.builder()
+                  .putAll(attributes)
+                  .put(AttributeKey.stringKey("type"), "requests")
+                  .build());
+
+      otelNumErrors =
```
dsmiley: I wonder if we're using OTEL the right way here to handle this scenario -- basically there are errors, and some are server and some are client.  I see you simply vary the "type" for each.  Maybe... instead we should only have the server & client, both have "type=errors", but differentiate by another attribute like "errorSource=client|server".  This still allows users to easily look at metrics undifferentiated if they don't care (i.e. errors of whatever source).  Presumably this would also work?

solr/core/src/java/org/apache/solr/jersey/RequestMetricHandling.java
```diff
@@ -117,13 +122,17 @@ public void filter(
...
       } else {
         log.debug("Skipping partialResults check because entity was not SolrJerseyResponse");
       }
+      final Timer.Context dropwizardTimer =
+          (Timer.Context) requestContext.getProperty("dropwizardTimer");
+      metrics.totalTime.inc(dropwizardTimer.stop());
 
-      final Timer.Context timer = (Timer.Context) requestContext.getProperty(TIMER);
-      metrics.totalTime.inc(timer.stop());
+      final AttributedLongTimer timer = (AttributedLongTimer) requestContext.getProperty(TIMER);
```
dsmiley: minor but I recommend using `var` in these cases rather then repeating the type

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
```diff
@@ -527,6 +717,7 @@ public void removeRegistry(String registry) {
    *     exist, so the swap operation will only rename the existing registry without creating an
    *     empty one under the previous name.
    */
+  // TODO SOLR-17458: Don't think we need
```
dsmiley: We might choose not to support core swapping in 10 or might have a disclaimer around metrics not swapping.

solr/core/src/java/org/apache/solr/metrics/SolrMetricsContext.java
```diff
@@ -116,21 +134,94 @@ public SolrMetricsContext getChildContext(Object child) {
...
     metricNames.add(name);
   }
 
   /** Return a snapshot of metric values that this component reports. */
+  // TODO Don't think this is needed anymore
   public Map<String, Object> getMetricsSnapshot() {
     return MetricUtils.convertMetrics(getMetricRegistry(), metricNames);
   }
 
+  public LongCounter longCounter(String metricName, String description) {
```
dsmiley, line 151: Seeing all these simply pass through to metricManager, I wonder if you plan to remove SolrMetricsContext once DropWizard is gone?  Or maybe it stays but you have yet to introduce OTEL scope, so it shows up here?

solr/core/src/java/org/apache/solr/util/stats/MetricUtils.java
```diff
@@ -858,4 +861,20 @@ public static <T extends PlatformManagedObject> void addMXBeanMetrics(
   public static MeterProvider getMeterProvider() {
     return GlobalOpenTelemetry.getMeterProvider();
   }
+
+  public static Attributes createAttributes(String... attributes) {
```
dsmiley, line 865: I'm susipicious that this is a good idea since Attributes has a nice builder.

solr/core/src/test/org/apache/solr/handler/RequestHandlerBaseTest.java
```diff
@@ -89,9 +116,16 @@ public void testEach4xxSolrExceptionIncrementsTheClientErrorCount() {
 
     RequestHandlerBase.processErrorMetricsOnException(e, metrics);
 
-    verify(metrics.numErrors).mark();
-    verify(metrics.numClientErrors).mark();
-    verifyNoInteractions(metrics.numServerErrors);
+    verify(mockLongCounter, times(1))
+        .add(eq(1L), argThat(attrs -> "errors".equals(attrs.get(AttributeKey.stringKey("type")))));
```
dsmiley: I suspect the idea here in this OTEL API is that you're expected to declare `AttributeKey.stringKey("type")` as a constant somewhere, perhaps TYPE_ATTR

solr/core/src/test/org/apache/solr/metrics/SolrCoreMetricTest.java
```diff
@@ -1,82 +0,0 @@
...
-package org.apache.solr.metrics;
-
-import com.codahale.metrics.Metric;
-import io.prometheus.metrics.model.snapshots.Labels;
-import org.apache.solr.SolrTestCaseJ4;
-import org.apache.solr.metrics.prometheus.SolrPrometheusFormatter;
-import org.apache.solr.metrics.prometheus.core.SolrCoreMetric;
-import org.junit.Test;
-
-public class SolrCoreMetricTest extends SolrTestCaseJ4 {
```
dsmiley, line 26: if you delete a test, it'd be good for your review to comment why

solr/core/src/java/org/apache/solr/metrics/SolrMetricProducer.java
```diff
@@ -40,20 +41,16 @@ static String getUniqueMetricTag(Object o, String parentName) {
...
-   * @param parentContext parent metrics context. If this component has the same life-cycle as the
-   *     parent it can simply use the parent context, otherwise it should obtain a child context
-   *     using {@link SolrMetricsContext#getChildContext(Object)} passing <code>this</code> as the
-   *     child object.
-   * @param scope component scope
+   * Deprecated entry point for initializing metrics. TODO SOLR-17458: This will be removed Change
+   * to only take attributes completely removing Dropwizard
    */
-  void initializeMetrics(SolrMetricsContext parentContext, String scope);
+  void initializeMetrics(SolrMetricsContext parentContext, Attributes attributes, String scope);
```
dsmiley, line 52: most of the lines of this PR were the impact of adding Attributes as a parameter.  Question:  could/should Attributes go into SolrMetricsContext?  It's not clear to me what the javadocs of this method should say about Attributes.    I think writing some javadocs at this stage helps clarify to you/reviewers what the intention is.  I understand OTEL metrics is attributed but it's not clear why the caller of this method would be passing attributes to the component implementing SolrMetricProducer.

solr/core/src/java/org/apache/solr/metrics/otel/instruments/AttributedLongTimer.java
```diff
@@ -0,0 +1,56 @@
...
+import io.opentelemetry.api.metrics.LongHistogram;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * This class records the elapsed time (in milliseconds) to a {@link LongHistogram}. Each thread
+ * measures its own timer allowing concurrent timing operations on separate threads..
+ */
+public class AttributedLongTimer extends AttributedLongHistogram {
+
+  private final ThreadLocal<Long> startTimeNanos = new ThreadLocal<>();
```
dsmiley: The use of ThreadLocal concerns me; seems like an over-use of ThreadLocal.  I could be convinced otherwise, though.  Instead of conceptually changing the state of this timer (scoped to this thread, granted), couldn't we return a Timer object that the caller closes?  Note that Solr has RTimer / RTimerTree you may want to see... perhaps return an RTimer subclass, for example.

solr/core/src/java/org/apache/solr/metrics/otel/OtelDoubleMetric.java
```diff
@@ -14,9 +14,9 @@
...
  */
+package org.apache.solr.metrics.otel;
 
-/**
- * The {@link org.apache.solr.metrics.prometheus.jvm.SolrPrometheusJvmFormatter} is responsible for
- * exporting solr.jvm registry metrics to Prometheus.
- */
-package org.apache.solr.metrics.prometheus.jvm;
+@FunctionalInterface
+public interface OtelDoubleMetric {
```
dsmiley: Does this serve a purpose?  Ditto for OtelLongMetric.

solr/core/src/java/org/apache/solr/blockcache/Metrics.java
mlbiscoc: Didn't know about `NOCOMMIT` I updated a few places with it and will use that instead going forward and keep the rest as TODO for now.

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
mlbiscoc: This is a good point actually and updating it with a separate type instead of inside the value didn't really make sense. Should look like this now:

```
solr_metrics_core_requests_total{category="QUERY",collection="foobar",core="foobar_shard1_replica_n1",handler="/select",internal="false",otel_scope_name="solr.core.foobar.shard1.replica_n1",replica="replica_n1",shard="shard1",source="client",type="errors"} 0.0
solr_metrics_core_requests_total{category="QUERY",collection="foobar",core="foobar_shard1_replica_n1",handler="/select",internal="false",otel_scope_name="solr.core.foobar.shard1.replica_n1",replica="replica_n1",shard="shard1",source="server",type="errors"} 0.0
```

solr/core/src/java/org/apache/solr/metrics/SolrMetricProducer.java
mlbiscoc: So before with Dropwizard since there was no attributes, `SolrMetricsContext` and `scope` were used together to determine the scope of where metrics should be inside of it's JSON. `scope` string for `RequestHandlerBase` was the handler such as `/select` but doesn't necessarily need to be a handler.

The idea now is that we won't need a `scope` string since whatever this scope string was can be inside of Attributes being passed in itself and more. So `Attributes` passed in here will be the base attributes for whatever is implementing `SolrMetricProducer`. For `RequestHandlerBase`, the attributes I felt were important to pass in as a base for every metric being recorded there was core descriptors like `coreName` `collection` etc and is then built from there with things like `type=errors`

solr/core/src/java/org/apache/solr/metrics/SolrMetricsContext.java
mlbiscoc: I'm debating if SolrMetricsContext isn't necessarily needed so I'll either delete it entirely or rename it to just `SolrMetricsScope` to pin instrument creation to the scope it wants to create to metricManager. I didn't determine that entirely in this PR yet.

In this PR, I should have mentioned it as well but the concept of `registryName` just changes to `otel scope` which is why you see when we pass `registryName` into metric manager. Also if you look at the metrics, there is a tag that OTEL created itself called `otel_scope_name` which was the registry or "scope" that this metric was created in. So we still keep the `solr.node` `solr.core.*` registries unless we want some much more granular.

```
solr_metrics_core_requests_times_milliseconds_bucket{category="QUERY", ... otel_scope_name="solr.core.foobar.shard1.replica_n1", ...} 1

```

solr/core/src/java/org/apache/solr/metrics/otel/OtelDoubleMetric.java
mlbiscoc: Not really actually. I just removed it.

solr/core/src/java/org/apache/solr/metrics/otel/instruments/AttributedLongTimer.java
mlbiscoc: I implemented a `MetricTimer` that extends the `RTimer`. This might work a bit nicer and works similar to how people might have used the Dropwizard Timer

solr/core/src/java/org/apache/solr/util/stats/MetricUtils.java
mlbiscoc: Removed

solr/core/src/test/org/apache/solr/handler/RequestHandlerBaseTest.java
mlbiscoc: I created a constant in `SolrMetricProducer`

solr/core/src/test/org/apache/solr/handler/admin/MetricsHandlerTest.java
```diff
@@ -706,459 +693,6 @@ public void testExprMetrics() throws Exception {
     assertTrue(map.toString(), map.containsKey("count"));
   }
 
-  @Test
-  @SuppressWarnings("unchecked")
-  public void testPrometheusMetricsCore() throws Exception {
```
mlbiscoc, line 711: Many of these Prometheus Tests are relating to the PrometheusFormatter features that was removed. A version of this should come back eventually further into the migration but with metric naming and attributes may change, I didn't see a reason to currently just fix it if naming is subject to change.

solr/core/src/test/org/apache/solr/metrics/SolrCoreMetricTest.java
mlbiscoc: Added in comments on many of the tests. All these `Solr*MetricTest` are specific unit tests around these wrappers that was used to parse the dropwizard output for labels. Don't need any of this anymore since it comes from OTEL.

mlbiscoc:
> The introduction of use of OTEL's "scope" will be very interesting & foundational to Solr's adoption of OTEL metrics. I think that's the next step after this PR.

This PR already has "scope" in the metrics. All metrics have a tag called `otel_scope_name`. When creating an instrument from `SolrMetricsContext`, it has to create the instrument from a `scope` to `SolrMetricManager` which is the registry name that the SolrMetricsContext holds. Unless we want scope to be something else, I am keeping it the same registries or "scope` which is `solr.node`, `solr.core.*`, `solr.jetty`...

mlbiscoc:
Actually the more I work on this, I think I am going to go with the path of renaming `SolrMetricsContext` to `SolrMetricsScope`. This is outside the scope of this PR, so will do it in another PR.

For `otel_scope_name`, we should update scope based on how the [OTEL instrumentation scope spec](https://opentelemetry.io/docs/specs/otel/common/instrumentation-scope/) recommends 

> Developers can decide what denotes a reasonable instrumentation scope. For example, they can select a module, a package, or a class as the instrumentation scope.

So for `requestHandlerBase`, the `scopeName` in `SolrMetricsScope` should maybe be the package name of `org.apache.solr.handler`.  If we really want to keep registries, then we can add a label called registry but I don't know if on an attribute based framework we still need the registry tag? Core metrics for example already have the `core` name as a label making it unique.

solr/core/src/java/org/apache/solr/metrics/SolrMetricProducer.java
dsmiley: aaaah.  Can you point me where you augment the base attributes with more so I can see what this looks like?

dsmiley:
"Context" and "Scope" are almost synonymous... "Context" just potentially holds more than a scope.  I'm not sure a rename is warranted; let's table that.  Better to do big renames at the end to make the change easier to review beforehand, then any wrote renames that we think are warranted.  Perhaps in a completely separate PR/commit.

I read about [instrumentation scope](https://opentelemetry.io/docs/specs/otel/common/instrumentation-scope/); thanks.  (I edited my previous response).  I don't think there's much point in using a Java package like thing here.  I think scope is intended to basically refer to all of Solr (e.g. be something like "org.apache.solr"), possibly with module consideration added (e.g. add ".llm").

dsmiley:
Remember AB's [comment](https://issues.apache.org/jira/browse/SOLR-17458?focusedCommentId=17941031&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17941031).  To add/clarify something he's getting at is the need to unregister metrics related to a core.  The DropWizard MetricsRegistry per core was a way to do that.  I chatted with Google Gemini about this matter, and it recommended creating a dedicated SdkMeterProvider in place of a DropWizard MetricRegistry.

solr/core/src/java/org/apache/solr/metrics/SolrCoreMetricManager.java
```diff
@@ -135,8 +144,19 @@ public void registerMetricProducer(String scope, SolrMetricProducer producer) {
...
+    producer.initializeMetrics(
+        solrMetricsContext,
+        Attributes.builder()
+            .put(CORE_ATTR, core.getCoreDescriptor().getName())
+            .put(COLLECTION_ATTR, collectionName)
+            .put(SHARD_ATTR, shardName)
+            .put(REPLICA_ATTR, replicaName)
+            .put((scope.startsWith("/")) ? HANDLER_ATTR : SCOPE_ATTR, scope)
+            .build(),
+        scope);
```
mlbiscoc, line 159: > Can you point me where you augment the base attributes with more so I can see what this looks like?

@dsmiley Here is where I initialize the base attributes for `requestHandlerBase` that was instrumented.

dsmiley: I'm looking for where you augment base attributes with more info

mlbiscoc:
> I think scope is intended to basically refer to all of Solr (e.g. be something like "org.apache.solr"), possibly with module consideration added (e.g. add ".llm").

Actually this makes sense. If you have a single place where many different metrics are stored, this scope name might be the differentiator especially if all your systems have some general metric like `http_requests_total`. So then all scope should be `org.apache.solr` technically. Do you still find value in keeping the concept of registry as a label on all metrics then? 

If not, I don't see the need to keep `SolrMetricsContext` and we should just create metrics through `SolrMetricManager` because that is currently all I am using it for. I don't see much value in aggregating on a label like `registry` except for `solr.core.*` but we have core as a label already.

solr/core/src/java/org/apache/solr/metrics/SolrCoreMetricManager.java
mlbiscoc: Look inside `RequestHandlerBase` at the `HandlerMetrics` constructor.

```
Attributes.builder()
                  .putAll(attributes)
                  .put(AttributeKey.stringKey("source"), "client")
                  .put(TYPE_ATTR, "errors")
                  .build()
```

dsmiley:
We don't need an attribute for the registry so long as our metrics are reasonably differentiated / clear.  We can change our mind in the future with basically no disturbance, I suppose.

I wouldn't get rid of SolrMetricsContext _yet_.  AB also kind of recommended keeping it to reduce the review/impact.

solr/core/src/java/org/apache/solr/metrics/SolrMetricProducer.java
mlbiscoc: > We don't need an attribute for the registry so long as our metrics are reasonably differentiated / clear. We can change our mind in the future with basically no disturbance, I suppose. I wouldn't get rid of SolrMetricsContext yet. AB also kind of recommended keeping it to reduce the review/impact.

If we keep `SolrMetricsContext` then it might actually then make sense to just place attributes in here like you said. The `Base Attributes` can still sit in here then be augmented with additional attributes on the metric no differently and it would significantly reduce number of PR line changes. Also gives `SolrMetricContext` a use besides just being a passthrough to `metricManager`. Currently Solr creates a new `SolrMetricsContext` per registry, but we will instead need to create on per implementation of `SolrMetricProducer` . I'm going to play with this and see if it makes sense otherwise I'll keep it as is.

mlbiscoc:
So I didn't move attributes into SolrMetricsContext. Ended up being a bigger refactor because currently keeping Dropwizard in parallel temporarily to incrementally migrate.  Maybe will do it closer to the end of this migration depending what happens to SolrMetricsContext.

I have another PR pretty close to ready and would like to put that up for review. 
It adds in dynamic solr core metric creation and deletion with the dedicated SdkMeterProvider and the correct `otel scope name` as well as adds a bunch of tests around these foundational components. There are some larger changes in-order to make this work that you'd be probably be interested in. 

I also have a WIP for some basic filtering but there are some "gotchas" that we need to discuss but will put it aside for a separate discussion.

If you have nothing else here, I'd like to merge this and put up the new PR.

dsmiley (approved):

PR #3385, opened 2025-06-11
SOLR-17780: Add support for scalar quantized dense vectors

liangkaiwen:
https://issues.apache.org/jira/browse/SOLR-17780

# Description

Lucene 9.9 codec includes an implementation of scalar quantized dense vectors. This is not yet supported in Solr.

# Solution

* Add a new schema type ScalarQuantizedDenseVectorField
* Add usage of Lucene99ScalarQuantizedVectorsFormat / Lucene99HnswScalarQuantizedVectorsFormat in SchemaCodecFactory. This incorporates the respective index reader/writers

# Tests

* Test for bad schema where a non valid `bits` value is provided
* Test for bad schema where a non valid `confidenceInterval` is provided
* Test for bad schema where `compress`=true but bit is not 4 byte alignment
* Test for default ScalarQuantizedDenseVectorField
* Test for 4 bit aligned ScalarQuantizedDenseVectorField
* Test for 4 bit compressed ScalarQuantizedDenseVectorField
* Test for custom `confidenceInterval` ScalarQuantizedDenseVectorField
* Test for `dynamicConfidenceInterval`=true

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java
+++ b/solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java
@@ -28,7 +28,9 @@ import org.apache.lucene.codecs.KnnVectorsWriter;
 import org.apache.lucene.codecs.PostingsFormat;
 import org.apache.lucene.codecs.lucene101.Lucene101Codec;
 import org.apache.lucene.codecs.lucene101.Lucene101Codec.Mode;
+import org.apache.lucene.codecs.lucene99.Lucene99HnswScalarQuantizedVectorsFormat;
 import org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsFormat;
+import org.apache.lucene.codecs.lucene99.Lucene99ScalarQuantizedVectorsFormat;
 import org.apache.lucene.index.SegmentReadState;
 import org.apache.lucene.index.SegmentWriteState;
 import org.apache.solr.common.SolrException;
@@ -36,10 +38,13 @@ import org.apache.solr.common.SolrException.ErrorCode;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.schema.DenseVectorField;
 import org.apache.solr.schema.FieldType;
+import org.apache.solr.schema.ScalarQuantizedDenseVectorField;
 import org.apache.solr.schema.SchemaField;
 import org.apache.solr.util.plugin.SolrCoreAware;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import static org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsFormat.DEFAULT_NUM_MERGE_WORKER;
 
 /**
  * Per-field CodecFactory implementation, extends Lucene's and returns postings format
@@ -126,16 +131,41 @@ public class SchemaCodecFactory extends CodecFactory implements SolrCoreAware {
           public KnnVectorsFormat getKnnVectorsFormatForField(String field) {
             final SchemaField schemaField = core.getLatestSchema().getFieldOrNull(field);
             FieldType fieldType = (schemaField == null ? null : schemaField.getType());
-            if (fieldType instanceof DenseVectorField vectorType) {
+            KnnVectorsFormat delegate;
+            if (fieldType instanceof ScalarQuantizedDenseVectorField scalarQuantizedVectorType) {
+              final String knnAlgorithm = scalarQuantizedVectorType.getKnnAlgorithm();
+              switch (knnAlgorithm) {
+                case DenseVectorField.FLAT_ALGORITHM:
+                  delegate = new Lucene99ScalarQuantizedVectorsFormat(
+                          scalarQuantizedVectorType.getConfidenceInterval(),
+                          scalarQuantizedVectorType.getBits(),
+                          scalarQuantizedVectorType.useCompression());
+                  break;
+                case DenseVectorField.HNSW_ALGORITHM:
+                  int maxConn = scalarQuantizedVectorType.getHnswMaxConn();
+                  int beamWidth = scalarQuantizedVectorType.getHnswBeamWidth();
+                  delegate = new Lucene99HnswScalarQuantizedVectorsFormat(maxConn,
+                          beamWidth,
+                          DEFAULT_NUM_MERGE_WORKER,
+                          scalarQuantizedVectorType.getBits(),
+                          scalarQuantizedVectorType.useCompression(),
+                          scalarQuantizedVectorType.getConfidenceInterval(),
+                          null);
+                  break;
+                default:
+                  throw new SolrException(ErrorCode.SERVER_ERROR, knnAlgorithm + " KNN algorithm is not supported");
+              }
+              return new SolrDelegatingKnnVectorsFormat(delegate, scalarQuantizedVectorType.getDimension());
+            } else if (fieldType instanceof DenseVectorField vectorType) {
               String knnAlgorithm = vectorType.getKnnAlgorithm();
               if (DenseVectorField.HNSW_ALGORITHM.equals(knnAlgorithm)) {
                 int maxConn = vectorType.getHnswMaxConn();
                 int beamWidth = vectorType.getHnswBeamWidth();
-                var delegate = new Lucene99HnswVectorsFormat(maxConn, beamWidth);
+                delegate = new Lucene99HnswVectorsFormat(maxConn, beamWidth);
                 return new SolrDelegatingKnnVectorsFormat(delegate, vectorType.getDimension());
               } else {
                 throw new SolrException(
-                    ErrorCode.SERVER_ERROR, knnAlgorithm + " KNN algorithm is not supported");
+                        ErrorCode.SERVER_ERROR, knnAlgorithm + " KNN algorithm is not supported");
               }
             }
             return super.getKnnVectorsFormatForField(field);
--- a/solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java
+++ b/solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java
@@ -131,44 +131,13 @@ public class SchemaCodecFactory extends CodecFactory implements SolrCoreAware {
           public KnnVectorsFormat getKnnVectorsFormatForField(String field) {
             final SchemaField schemaField = core.getLatestSchema().getFieldOrNull(field);
             FieldType fieldType = (schemaField == null ? null : schemaField.getType());
-            KnnVectorsFormat delegate;
-            if (fieldType instanceof ScalarQuantizedDenseVectorField scalarQuantizedVectorType) {
-              final String knnAlgorithm = scalarQuantizedVectorType.getKnnAlgorithm();
-              switch (knnAlgorithm) {
-                case DenseVectorField.FLAT_ALGORITHM:
-                  delegate = new Lucene99ScalarQuantizedVectorsFormat(
-                          scalarQuantizedVectorType.getConfidenceInterval(),
-                          scalarQuantizedVectorType.getBits(),
-                          scalarQuantizedVectorType.useCompression());
-                  break;
-                case DenseVectorField.HNSW_ALGORITHM:
-                  int maxConn = scalarQuantizedVectorType.getHnswMaxConn();
-                  int beamWidth = scalarQuantizedVectorType.getHnswBeamWidth();
-                  delegate = new Lucene99HnswScalarQuantizedVectorsFormat(maxConn,
-                          beamWidth,
-                          DEFAULT_NUM_MERGE_WORKER,
-                          scalarQuantizedVectorType.getBits(),
-                          scalarQuantizedVectorType.useCompression(),
-                          scalarQuantizedVectorType.getConfidenceInterval(),
-                          null);
-                  break;
-                default:
-                  throw new SolrException(ErrorCode.SERVER_ERROR, knnAlgorithm + " KNN algorithm is not supported");
-              }
-              return new SolrDelegatingKnnVectorsFormat(delegate, scalarQuantizedVectorType.getDimension());
-            } else if (fieldType instanceof DenseVectorField vectorType) {
-              String knnAlgorithm = vectorType.getKnnAlgorithm();
-              if (DenseVectorField.HNSW_ALGORITHM.equals(knnAlgorithm)) {
-                int maxConn = vectorType.getHnswMaxConn();
-                int beamWidth = vectorType.getHnswBeamWidth();
-                delegate = new Lucene99HnswVectorsFormat(maxConn, beamWidth);
-                return new SolrDelegatingKnnVectorsFormat(delegate, vectorType.getDimension());
-              } else {
-                throw new SolrException(
-                        ErrorCode.SERVER_ERROR, knnAlgorithm + " KNN algorithm is not supported");
-              }
+            if (fieldType instanceof DenseVectorField) {
+              final DenseVectorField vectorField = (DenseVectorField) fieldType;
+              return new SolrDelegatingKnnVectorsFormat(vectorField.buildKnnVectorsFormat(),
+                      vectorField.getDimension());
+            } else {
+              throw new SolrException(ErrorCode.SERVER_ERROR, "field is not a support KNN vector type");
             }
-            return super.getKnnVectorsFormatForField(field);
           }
         };
   }
--- a/solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java
+++ b/solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java
@@ -127,14 +127,16 @@ public class SchemaCodecFactory extends CodecFactory implements SolrCoreAware {
           public KnnVectorsFormat getKnnVectorsFormatForField(String field) {
             final SchemaField schemaField = core.getLatestSchema().getFieldOrNull(field);
             FieldType fieldType = (schemaField == null ? null : schemaField.getType());
-            if (fieldType instanceof DenseVectorField) {
-              final DenseVectorField vectorField = (DenseVectorField) fieldType;
+            if (fieldType instanceof DenseVectorField vectorField) {
+              final String knnAlgorithm =  vectorField.getKnnAlgorithm();
+              if (!DenseVectorField.HNSW_ALGORITHM.equals(knnAlgorithm)) {
+                throw new SolrException(
+                        ErrorCode.SERVER_ERROR, knnAlgorithm + " KNN algorithm is not supported");
+              }
               return new SolrDelegatingKnnVectorsFormat(
                   vectorField.buildKnnVectorsFormat(), vectorField.getDimension());
-            } else {
-              throw new SolrException(
-                  ErrorCode.SERVER_ERROR, "field is not a supported KNN vector type");
             }
+            return super.getKnnVectorsFormatForField(field);
           }
         };
   }
--- a/solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java
+++ b/solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java
@@ -30,7 +30,6 @@ import org.apache.lucene.codecs.lucene101.Lucene101Codec;
 import org.apache.lucene.codecs.lucene101.Lucene101Codec.Mode;
 import org.apache.lucene.codecs.lucene99.Lucene99HnswScalarQuantizedVectorsFormat;
 import org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsFormat;
-import org.apache.lucene.codecs.lucene99.Lucene99ScalarQuantizedVectorsFormat;
 import org.apache.lucene.index.SegmentReadState;
 import org.apache.lucene.index.SegmentWriteState;
 import org.apache.solr.common.SolrException;
@@ -38,13 +37,10 @@ import org.apache.solr.common.SolrException.ErrorCode;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.schema.DenseVectorField;
 import org.apache.solr.schema.FieldType;
-import org.apache.solr.schema.ScalarQuantizedDenseVectorField;
 import org.apache.solr.schema.SchemaField;
 import org.apache.solr.util.plugin.SolrCoreAware;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import static org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsFormat.DEFAULT_NUM_MERGE_WORKER;
 
 /**
  * Per-field CodecFactory implementation, extends Lucene's and returns postings format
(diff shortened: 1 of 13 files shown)


liangkaiwen:
I still need to add tests and update documentation

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Override
+    public void init(IndexSchema schema, Map<String, String> args) {
+        super.init(schema, args);
+
+        this.bits = ofNullable(args.get(BITS))
+                .map(Integer::parseInt)
+                .orElse(DEFAULT_BITS);
+        args.remove(BITS);
```
dsmiley: instead of "args.get" a few lines above, you should use this "remove", as it returns it.  Ditto for other args below.

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
```diff
@@ -0,0 +1,107 @@
...
+        boolean useDynamicConfidenceInterval = ofNullable(args.get(DYNAMIC_CONFIDENCE_INTERVAL))
+                .map(Boolean::parseBoolean)
+                .orElse(false);
+        args.remove(DYNAMIC_CONFIDENCE_INTERVAL);
+
+        if (useDynamicConfidenceInterval) {
+            this.confidenceInterval = Lucene99ScalarQuantizedVectorsFormat.DYNAMIC_CONFIDENCE_INTERVAL;
+        }
+
+        this.confidenceInterval = ofNullable(args.get(CONFIDENCE_INTERVAL))
```
dsmiley: this overwrites the confidenceInterval set when useDynamicConfidenceInterval==true above

solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java
```diff
@@ -126,16 +131,41 @@ public DocValuesFormat getDocValuesFormatForField(String field) {
           public KnnVectorsFormat getKnnVectorsFormatForField(String field) {
             final SchemaField schemaField = core.getLatestSchema().getFieldOrNull(field);
             FieldType fieldType = (schemaField == null ? null : schemaField.getType());
-            if (fieldType instanceof DenseVectorField vectorType) {
+            KnnVectorsFormat delegate;
```
dsmiley: The instanceof checks can be a code smell.  I look at all this and wonder: wouldn't it make sense for a method DenseVectorField.buildKnnVectorsFormat() to exist?  Or just a getter if it's built in the field type's init()... which would also mean no need to have each of those fields (e.g. no bits, ...) with their getters either.

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
liangkaiwen: good catch. I was hoping to make any explicit confidence interval override the dynamic behavior, however with both absence of arg and the default being null I think I will flip the behavior (use dynamic if specified otherwise confidence interval if specified otherwise default confidence interval if none provided)

solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java
liangkaiwen: reworked it. Let me know if this is along the lines of what you were thinking

dsmiley: Much better.  Did you consider init() also creating the codec to thus avoid the need corresponding fields, getters, setters?

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
dsmiley: Okay but then let's check if both are set and throw an exception.

solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java
liangkaiwen: I'm not quite sure I follow. You mean the ScalarQuantizedDenseVectorField init() or the SchemaCodecFactory init()?

dsmiley: `ScalarQuantizedDenseVectorField.init()` (and its subclass, DenseVectorField) creating a `KnnVectorsFormat` that later can simply be returned.

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
liangkaiwen: If neither are set, it should default to the default behavior (scaled confidence interval) - which according to lucene spec is passing a null confidence interval

solr/core/src/java/org/apache/solr/core/SchemaCodecFactory.java
liangkaiwen: That's probably true. However it does look like the getters are used in the DenseVectorField tests as KnnVectorsFormat doesn't expose the internal setting values. Personally I think it's ok to leave patterned as-is and having getters affords some level of flexibility and transparency.

But I can make this change if you feel strongly about it

solr/core/src/java/org/apache/solr/schema/DenseVectorField.java
```diff
@@ -64,6 +65,7 @@
 public class DenseVectorField extends FloatPointField {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
   public static final String HNSW_ALGORITHM = "hnsw";
+  public static final String FLAT_ALGORITHM = "flat";
```
alessandrobenedetti: 'VECTOR_STORAGE_ALGORITHM' maybe?
Is it used somewhere?

alessandrobenedetti: reading it more we are mixing up the 'knn' algorithm (only HNSW supported right now), with the 'vector storage' (flat, scalarQuantised and binaryQuantised

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.Map;
+
+import static java.util.Optional.ofNullable;
+import static org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsFormat.DEFAULT_NUM_MERGE_WORKER;
+
+public class ScalarQuantizedDenseVectorField extends DenseVectorField {
+    public static final String BITS = "bits"; //
+    public static final String CONFIDENCE_INTERVAL = "confidenceInterval";
+    public static final String DYNAMIC_CONFIDENCE_INTERVAL = "dynamicConfidenceInterval";
+    public static final String COMPRESS = "compress"; // can only be enabled when bits = 4 per Lucene codec spec
```
alessandrobenedetti: these first four lines are param names if I'm not mistaken.

Maybe we can call them '<name>_PARAM' e.g. 'BITS_PARAM'

Or add a comment line at the beginning that clearly group them as param names,
it's a minor though, but can increase code readability

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
```diff
@@ -0,0 +1,128 @@
...
+                .map(Boolean::parseBoolean)
+                .orElse(false)) {
+            this.confidenceInterval = Lucene99ScalarQuantizedVectorsFormat.DYNAMIC_CONFIDENCE_INTERVAL;
+        }
+    }
+
+    @Override
+    public KnnVectorsFormat buildKnnVectorsFormat() {
+        final String knnAlgorithm = getKnnAlgorithm();
+        if (KNN_ALGORITHM.equals(knnAlgorithm)) {
```
alessandrobenedetti: mmmm not sure about this part:

The only knn algorithm supported right now is 'HNSW'.
While for vector storage, there's flat (what currently supported), scalar quantised and binary quantised.
Effectively in Lucene HNSW vector formats define how to build and read the HNSW graph (where each vector ID is a node). 
Vector storage formats define how to store the vectors themselves.

The vector storage format is directly (and only) used by a the graph format parent ( I had in my to-do list to re-organise the package to easily differentiate graph from storage formats)

**Lucene99HnswVectorsFormat** uses **Lucene99FlatVectorsFormat**.
• Store for each field the vector metadata (.vemf) and vector data (vec).
• It has a dedicated writer and reader, dependent on the vector encoding (BYTE or
FLOAT32).
• It has its scorer (that just uses the supplied similarity).

**Lucene99HnswScalarQuantizedVectorsFormat** uses **Lucene99ScalarQuantizedVectorsFormat** (for scalar quantisation).
• Lossy compression from float32 to a number of bits (4 or 7)
• Raw disk increases (raw + quantized vectors).
• Off-heap memory decreases as quantised vectors are loaded.
• Search is faster.
• It has a dedicated writer and reader, dependent on the bits to use for quantisation.

Given that, I suspect with this contribution we want to differentiate between:
Lucene99HnswVectorsFormat and Lucene99HnswScalarQuantizedVectorsFormat (without manipulating directly the vector storage).

-> the if then else should be changed I suspect, I don't understand it as it is right now as we are mixing a graph format (Lucene99HnswScalarQuantizedVectorsFormat already using the scalar quantised storage format) and a vectore storage format (Lucene99ScalarQuantizedVectorsFormat)

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
alessandrobenedetti: to be fair, checking the original DenseVectorField, the same renaming could help there as well I suspect

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
alessandrobenedetti: From an initial review the ScalarQuantizedDenseVectorField should always return the Lucene99HnswScalarQuantizedVectorsFormat.

The traditional DenseVectorField, should always return Lucene99HnswVectorsFormat

alessandrobenedetti (changes requested):
I believe the overall idea to be ok, having the quantised format under a dedicated new field type, given the amount of parameters make sense from a Solr code perspective (the user on the schema won't see much difference).

I left some comments, the 'vectorField.buildKnnVectorsFormat' doesn't look all right to me, for the quantised one.
I think it's just a misunderstanding of how the formats work, so it should be easily solvable.
Once tests are added I think this Pull Request would be quickly good to go!

solr/solr-ref-guide/modules/query-guide/pages/dense-vector-search.adoc
```diff
@@ -240,6 +240,66 @@ client.add(Arrays.asList(d1, d2));
 ====
 ======
 
+=== ScalarQuantizedDenseVectorField
+Because dense vectors can have a costly storage footprint, it may be worthwhile to use a technique called "quantization"
+to reduce the stored representation size at the cost of some precision.
+
+This dense vector type uses a conversion that projects a 32 bit float precision feature down to an 8 bit int (or smaller)
+by linearly mapping the float range down to evenly sized "buckets" of values that fit into an int. A more detailed explanation
+can be found in this https://www.elastic.co/search-labs/blog/scalar-quantization-101[blog post].
```
alessandrobenedetti: I would encourage linking a resource that's not Elastic. I  don't think we want Elasticsearch reference in the Solr documentation, it would feel weird

solr/solr-ref-guide/modules/query-guide/pages/dense-vector-search.adoc
```diff
@@ -240,6 +240,66 @@ client.add(Arrays.asList(d1, d2));
 ====
 ======
 
+=== ScalarQuantizedDenseVectorField
+Because dense vectors can have a costly storage footprint, it may be worthwhile to use a technique called "quantization"
```
alessandrobenedetti: I would specify that's not only strictly storage (because actually the storage footprint on disk increases with quantisation).

I would stress memory/latency benefits above all.

solr/solr-ref-guide/modules/query-guide/pages/dense-vector-search.adoc
```diff
@@ -240,6 +240,66 @@ client.add(Arrays.asList(d1, d2));
...
+quantization (and therefore information gain) we can scale the quantization intervals to the middle n % of values and place the remaining
+outliers in the outermost intervals.
++
+For example: 0.9 means scale interval sizes to the middle 90% of values
++
+If this param is omitted a default is used; scaled to the number of dimensions according to `1-1/(vector_dimensions + 1)`
++
+If `0` is provided, confidence intervals will be dynamically adjusted (at segment merge time) optimized by sampling values
++
+Accepted values: `FLOAT32`  (within 0.9 and 1.0) or 0 for dynamically adjusted confidence interval
```
alessandrobenedetti: I'm not sure this is clear enough for someone not familiar with the topic.
Can I suggest involving some additional readers here?
@ilariapet , @aruggero , @liangkaiwen  can you involve some of your colleagues?
We may need to rephrase a bit this chapter

solr/solr-ref-guide/modules/query-guide/pages/dense-vector-search.adoc
```diff
@@ -240,6 +240,66 @@ client.add(Arrays.asList(d1, d2));
...
+
+`compress`::
++
+[%autowidth,frame=none]
+|===
+s|Optional |Default: `false`
+|===
++
+If set to true, this will further pack multiple dimension values within a one byte alignment. This further decreases the storage size by 50% at some decode penalty.
```
alessandrobenedetti: is it related to the quantised vector storage (both memory and disk)?
or only memory?
Maybe we should clarify that on disk we still store also the full vector

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
```diff
@@ -0,0 +1,154 @@
...
+  public static final String CONFIDENCE_INTERVAL_PARAM = "confidenceInterval";
+  public static final String DYNAMIC_CONFIDENCE_INTERVAL_PARAM = "dynamicConfidenceInterval";
+  public static final String COMPRESS_PARAM =
+      "compress"; // can only be enabled when bits = 4 per Lucene codec spec
+
+  static final int DEFAULT_BITS = 7; // use signed byte as default when unspecified
+  static final Float DEFAULT_CONFIDENCE_INTERVAL = null; // use dimension scaled confidence interval
+
+  // lucene does not expose these so we are duplicating them here
+  static final Float MINIMUM_CONFIDENCE_INTERVAL = 0.9f;
```
alessandrobenedetti: I'm not sure we need this here, looking at Lucene code, in case we go off the values we get an "llegalArgumentException(
          "confidenceInterval must be between "
              + MINIMUM_CONFIDENCE_INTERVAL
              + " and "
              + MAXIMUM_CONFIDENCE_INTERVAL
              + " or 0"
              + "; confidenceInterval="
              + confidenceInterval);"
              
              Can't we catch this exception and handle it rather than duplicating the constants?

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
```diff
@@ -0,0 +1,154 @@
...
+
+    if (confidenceInterval != null
+        && confidenceInterval != Lucene99ScalarQuantizedVectorsFormat.DYNAMIC_CONFIDENCE_INTERVAL
+        && (confidenceInterval < MINIMUM_CONFIDENCE_INTERVAL
+            || confidenceInterval > MAXIMUM_CONFIDENCE_INTERVAL)) {
+      throw new SolrException(
+          SolrException.ErrorCode.SERVER_ERROR,
+          getClass().getSimpleName()
+              + " fields must have non-dynamic confidence interval between 0.9 and 1.0 "
+              + field.getName());
```
alessandrobenedetti: as commented above, I'm wondering if we can use the Lucene exceptions here

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+    if (getBits() != 4 && getBits() != 7) {
```
alessandrobenedetti: can't we rely on Lucene checks that throw the "if (bits < 1 || bits > 8 || (ALLOWED_BITS & (1 << bits)) == 0) {
      throw new IllegalArgumentException("bits must be one of: 4, 7; bits=" + bits);
    }"

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
```diff
@@ -0,0 +1,154 @@
...
+      throw new SolrException(
+          SolrException.ErrorCode.SERVER_ERROR,
+          getClass().getSimpleName()
+              + " fields must have bit size of 4 (half-byte) or 7 (signed-byte) "
+              + field.getName());
+    }
+
+    if (useCompression() && getBits() != 4) {
```
alessandrobenedetti: Can't we rely on "if (bits > 4 && compress) {
      // compress=true otherwise silently does nothing when bits=7?
      throw new IllegalArgumentException("compress=true only applies when bits=4");
    }" Lucene side?

solr/core/src/test-files/solr/collection1/conf/schema-densevector.xml
```diff
@@ -25,7 +25,7 @@
   <fieldType name="high_dimensional_float_knn_vector" class="solr.DenseVectorField" vectorDimension="2048" similarityFunction="cosine" vectorEncoding="FLOAT32"/>
   <fieldType name="high_dimensional_byte_knn_vector" class="solr.DenseVectorField" vectorDimension="2048" similarityFunction="cosine" vectorEncoding="BYTE"/>
   <fieldType name="plong" class="solr.LongPointField" useDocValuesAsStored="false"/>
-  
+
```
alessandrobenedetti: I guess this is unwanted

solr/core/src/test-files/solr/collection1/conf/schema-densevector-quantized.xml
```diff
@@ -0,0 +1,86 @@
...
+  <field name="v_scalar_dynamic" type="knn_vector_scalar_quantized_confidence_dynamic" indexed="true" stored="true" />
+
+  <field name="string_field" type="string" indexed="true" stored="true" multiValued="false" required="false"/>
+
+  <field name="_version_" type="plong" indexed="true" stored="true" multiValued="false" />
+  <field name="_text_" type="text_general" indexed="true" stored="false" multiValued="true"/>
+  <copyField source="*" dest="_text_"/>
+  <fieldType name="text_general" class="solr.TextField" positionIncrementGap="100" multiValued="true">
+    <analyzer type="index">
+      <tokenizer class="solr.StandardTokenizerFactory"/>
```
alessandrobenedetti, line 73: can't we use one single schema file, where you add the different fields and field types with the ad hoc incorrect parameters?

alessandrobenedetti (changes requested):
I think we are on the right path, I left various comments, my main points are:

1) re-using Lucene validation and IllegalArgumentException, rather than rewrite them (so that if Lucene changes anything we get it for free)

2) restructure tests to use just one schema file (with the multiple ad hoc incorrect configs)

Let's iterate, I think we are close! Thanks for the work done so far!

solr/solr-ref-guide/modules/query-guide/pages/dense-vector-search.adoc
liangkaiwen: This accepted values section was copied and I forgot to update. For this param, it's simply a bool

solr/solr-ref-guide/modules/query-guide/pages/dense-vector-search.adoc
liangkaiwen: I believe this is on disk packing. Will add clarification

solr/core/src/test-files/solr/collection1/conf/schema-densevector-quantized.xml
liangkaiwen: I can try it. Would that not just always error on the first invalid field though?

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
liangkaiwen: I agree, it's bad practice to duplicate validation code. The lucene implementation doesn't expose any validation function, but I updated the code to just simply call the format constructor and not store the created object. Exception must be a SolrException to match the inherited interface, so I wrapped the IllegalArgumentException. Let me know if this meets your expectation?

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
```diff
@@ -0,0 +1,132 @@
...
+
+  @Override
+  public void checkSchemaField(final SchemaField field) throws SolrException {
+    super.checkSchemaField(field);
+
+    try {
+      // format does not expose any validation, however by constructing
+      // we will run lucene level argument validation
+      buildKnnVectorsFormat();
+    } catch (Exception e) {
```
alessandrobenedetti: Isn't there any more specific exception to catch?

alessandrobenedetti (approved):
I think we are nearly there for merging, just check the exception catching I mentioned and we can merge in the next few days

solr/core/src/java/org/apache/solr/schema/ScalarQuantizedDenseVectorField.java
liangkaiwen: changed type to IllegalArgumentException

alessandrobenedetti (approved):
running tests, adding changes and merging soon!

alessandrobenedetti:
Hi @liangkaiwen , I rebased, added CHANGES.txt, and now tests are broken. I don't have bandwidth now to investigate, can you? Let me know under the comments, I'm happy to finalise this and merge it soon!

dsmiley:
Just a quick comment: next time please don't rebase a PR; simply merge main into it.  I've found rebasing PRs to be a bad practice as it resets the GitHub review state, making it impossible to be certain what incremental changes happened (if any).

alessandrobenedetti:
> Just a quick comment: next time please don't rebase a PR; simply merge main into it. I've found rebasing PRs to be a bad practice as it resets the GitHub review state, making it impossible to be certain what incremental changes happened (if any).

Wasn't it mandatory to comply with Crave checks? That's the only reason I (and my colleagues at least) are doing it, has anything changed?

dsmiley:
I don't recall Crave being confused by a merge from "main".  I do recall once in a while (rare-ish), GitHub PR would show changes from the merge that it shouldn't.

liangkaiwen:
@alessandrobenedetti updated + fixed tests

alessandrobenedetti:
Checks and tests are still failing @liangkaiwen , I'll give it a try now, If I fix them in 1 hour I'll merge, if not I'll ask you to give a double check, keep you posted!

alessandrobenedetti:
I fixed the tidy problem, but the tests (the ones expected to throw exceptions) are still failing.
I gave it a very quick try, with debug, but it doesn't look obvious to me, as I'm not familiar with that type of test with an expected error string, so in case you can't manage to fix them, I'll give it a try, spending a bit more time. Let me know!

liangkaiwen:
Huh, I thought it had passed after I committed my last changes but I guess not. No problem, I will handle it

PR #3256, opened 2025-03-11
SOLR-17698: Remove deprecated EnumField

epugh:
https://issues.apache.org/jira/browse/SOLR-17698

# Description

Removing deprecated EnumField

# Solution
EnumField has been replaced by EnumFieldType.  Simplified tests and supporting test configurations.

# Tests

Existing

PR #3390, opened 2025-06-16
SOLR-17698: Fix TestUseDocValuesAsStored for enums

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17698

Fixes test broken by https://github.com/apache/solr/pull/3256


epugh (approved):
LGTM..   I thoiught there was a failure, but then I re-ran and passed, and GH actions passed, so I thought it was transient.   Thanks for this @HoustonPutman

PR #3391, opened 2025-06-16
SOLR-17092: Fix race condition for async request status

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17092

This is only for when the Overseer is used, as compared to https://github.com/apache/solr/pull/3268 which was for the `DistributedApiAsyncTracker`. But overall, the two are similar race conditions. We need to check if the request is running first, in case the Overseer ZK operations are completed in-between the "isSuccessful" and "isRunning" checks. If we check "isRunning" first, then we should not be able to return "Not found" or "submitted" because the check will either see it in the running map or the successful map, there is no possibility that it is in neither of them.


psalagnac (approved):
This looks good, thanks @HoustonPutman.

Sharing more of my understanding here:
We add the task to the completed/failed map before we remove it from the running map, so we have a time window where it is in both maps. In this case, we still return "running" status. This is fine, we will return "completed/failed" at the next call.
I don't see any race condition anymore.

PR #3392, opened 2025-06-16
SOLR-17790: Allow -j option to start with a dash

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17790

So that you can pass options to `-j` that start with a dash (e.g. (`-j "--module=..."`).


epugh:
this makes sense, but I wonder if I could talk you into a bats test so we don't keep battling it out on this?  (I think this may have been supported previously and someone (maybe me!) tightned it up!   Also, do we need to do anything in the windows side, do we have the same issue?

HoustonPutman:
Yeah, I can add a test (it would have prevented this from being committed as you mentioned). And I think windows is safe, I checked it best I could and I don't really see any of the same checks.

PR #3395, opened 2025-06-19
Solr navigable remove default

wildtusker:
https://issues.apache.org/jira/browse/SOLR-NavigableRemoveDeafult

# Description

_get(T t1,T t2) is replaced _get(T t1) in Navigable Object.All the classes which make call to this method get replaced

# Solution

Analyze and Refactor feature of IntelliJ Idea were used to carry out the task

# Tests

No tests developed.Tests Passed.

### DIFF
--- a/solr/core/src/java/org/apache/solr/packagemanager/PackageManager.java
+++ b/solr/core/src/java/org/apache/solr/packagemanager/PackageManager.java
@@ -36,6 +36,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Scanner;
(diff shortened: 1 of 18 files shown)


dsmiley:
Looks good!
But why not also do `_getStr` ?

> https://issues.apache.org/jira/browse/SOLR-NavigableRemoveDeafult

That URL is invalid

dsmiley (approved):

dsmiley:
Obviously some formatting issues (you didn't run Tidy before committing) but looks good

dsmiley:
I back-ported to branch_9x as well, and included the former method signatures with Deprecated annotations for backwards-compatibility.

Thanks for contributing Gaurav!

PR #3396, opened 2025-06-19
GHA renovation

dsmiley:
* branches (all)
* timeout-minutes
* checkout version
* harmonization across scripts (spacing / wording / labels)
* remove chmod for gradlew (unnecessary)

Caching tweaks: removed this but I'm unsure:
https://github.com/apache/lucene-solr/pull/1910 CC @madrob


.github/workflows/docker-test.yml
```diff
@@ -17,36 +16,29 @@ jobs:
...
       uses: actions/setup-java@v4
       with:
         distribution: 'temurin'
         java-version: 21
         java-package: jdk
+
     - name: Setup Gradle
       uses: gradle/actions/setup-gradle@v4
-    - name: Install ACL
-      run: sudo apt-get install acl
```
madrob, line 38: Do we not need this any more? @HoustonPutman is probably better to look

madrob (commented):
Was there something wrong with the caching? Were we running up against storage limits? The original comment said it took 5 min off the build which seems pretty significant, especially in a 15 minute timeout

dsmiley:
The elaborate caching configuration is not in Lucene, despite that same elaborate caching configuration being committed before the lucene-solr split.  Thus Lucene devs removed it.  In particular, it was @dweiss in apache/lucene#13110

I'm going to further improve this PR to use the nice `.github/actions/prepare-for-build/action.yml` (in Lucene, small tweaks for Solr) refactoring so that each GHA can share it

dsmiley:
@madrob when you first added that 5 years ago, we weren't using the GHA gradle action or it didn't exist then.  It now does with caching built-in.  My goal here is to simplify.

.github/workflows/docker-test.yml
dsmiley: in passing tests I trust :-)   (which they do)

dsmiley:
The docker build isn't taking any noticeably longer here than the test runs I see in the GHA dashboard for that script.
I want to merge this Tuesday evening.

PR #3398, opened 2025-06-20
SOLR-17792: Fix deadlocks in ParallelHttpShardHandler

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17792

Here I'm making `responseFutureMap` a ConcurrentMap, so it doesn't need to be synchronized. Synchronization blocks also won't be necessary because we won't be using `pending` anymore. We will consider ourselves pending if `responseFutureMap` or `responses` is not empty. (previously pending was increased if either something was added to `responseFutureMap`, or a no-shard-response was added to `responses`).

There is still some synchronization, but just around closing to make sure that responses aren't added to the response list after cancelling. As the response list _might_ be used by future requests down the line and we wouldn't want to mix up results from a cancelled request with the next request.

This is a blocker for 9.9, because the ParallelHttpShardHandler is unusable without fixing this.

### DIFF
--- a/solr/core/src/java/org/apache/solr/update/PeerSync.java
+++ b/solr/core/src/java/org/apache/solr/update/PeerSync.java
@@ -246,6 +246,7 @@ public class PeerSync implements SolrMetricProducer {
           if (log.isInfoEnabled()) {
             log.info("{} DONE. sync failed", msg());
           }
+          shardHandler.cancelAll();
           syncErrors.inc();
           return PeerSyncResult.failure();
         }
--- a/solr/core/src/java/org/apache/solr/update/PeerSync.java
+++ b/solr/core/src/java/org/apache/solr/update/PeerSync.java
@@ -246,7 +246,6 @@ public class PeerSync implements SolrMetricProducer {
           if (log.isInfoEnabled()) {
             log.info("{} DONE. sync failed", msg());
           }
-          shardHandler.cancelAll();
           syncErrors.inc();
           return PeerSyncResult.failure();
         }
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -30,6 +30,7 @@ github:
     branch_9_6: {}
     branch_9_7: {}
     branch_9_8: {}
+    branch_9_9: {}
     branch_9x: {}
 
   protected_tags:
--- a/dev-tools/scripts/scriptutil.py
+++ b/dev-tools/scripts/scriptutil.py
@@ -229,7 +229,7 @@ def attemptDownload(urlString, fileName):
 
 version_prop_re = re.compile(r'baseVersion\s*=\s*([\'"])(.*)\1')
 
-lucene_version_prop_re = re.compile(r'^apache-lucene\s*=\s*"([a-zA-Z0-9\.\-]+)"')
+lucene_version_prop_re = re.compile(r'apache-lucene\s*=\s*"([a-zA-Z0-9\.\-]+)"')
 
 def find_current_version():
   script_path = os.path.dirname(os.path.realpath(__file__))
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -133,6 +133,8 @@ Deprecation Removals
 
 * SOLR-17698: Removed deprecated EnumField field type. (Eric Pugh, David Smiley)
 
+* SOLR-17778: Removed no-op core admin operation FORCEPREPAREFORLEADERSHIP and its SolrJ peer OverrideLastPublished. (Pierre Salagnac)
+
 Dependency Upgrades
 ---------------------
 
@@ -192,6 +194,31 @@ Other Changes
 * SOLR-14070: Deprecate CloudSolrClient's ZooKeeper Hosts constructor. (Abhishek Umarjikar)
 
 * SOLR-17703: SolrJ - Rename BinaryResponseParser and BinaryRequestWriter including StreamingBinaryResponseParser to "JavaBin.." equivalent. (Abhishek Umarjikar)
+
+==================  9.10.0 ==================
+New Features
+---------------------
+(No changes)
+
+Improvements
+---------------------
+(No changes)
+
+Optimizations
+---------------------
+(No changes)
+
+Bug Fixes
+---------------------
+(No changes)
+
+Dependency Upgrades
+---------------------
+(No changes)
+
+Other Changes
+---------------------
+(No changes)
 
 ==================  9.9.0 ==================
 New Features
@@ -258,6 +285,10 @@ Improvements
 
 * SOLR-17776: The HttpJdkSolrClient wasn't setting the connection timeout as per the builder configuration. (David Smiley)
 
+* SOLR-17746: Provide long form --jettyconfig option to go with -j in bin/solr scripts. (Eric Pugh, Rahul Goswami)
+
+* SOLR-17628: Export metric timers via `wt=prometheus` as Prometheus summaries. (Jude Muriithi, Matthew Biscocho)
+
 Optimizations
 ---------------------
 * SOLR-17578: Remove ZkController internal core supplier, for slightly faster reconnection after Zookeeper session loss. (Pierre Salagnac)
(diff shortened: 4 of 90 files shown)


HoustonPutman:
I'm planning on merging this early next week and would love a review beforehand!

solr/core/src/java/org/apache/solr/update/PeerSync.java
```diff
@@ -246,7 +246,6 @@ public PeerSyncResult sync() {
           if (log.isInfoEnabled()) {
             log.info("{} DONE. sync failed", msg());
           }
-          shardHandler.cancelAll();
```
dsmiley: what's this removal about?

dsmiley (commented):
Jason or Gus would be ideal reviewers; this is complex stuff that both of them have touched.  I've been a bystandard to the action, and I want to share with you one point I've had about this code:
https://github.com/apache/solr/pull/2666#discussion_r1777941494
tldr; HttpShardHandler ought to use ExecutorCompletionService.  WDYT?

solr/core/src/java/org/apache/solr/update/PeerSync.java
HoustonPutman: Basically, the shardHandler will already cancel requests on a failure if that is set, so this is no longer needed. I don't have to remove it, since it's already cancelled. I'll reconsider this.

HoustonPutman:
> tldr; HttpShardHandler ought to use ExecutorCompletionService. WDYT?

So after looking into this a bit, it would save us a bit of overhead of adding responses to our queue. However there are a few issues:

- `lbClient.requestAsync(lbReq)` returns a `Future` for us, so `ExecutorCompletionService.submit()` doesn't really fit our use case, since that is expecting a `Runnable` or `Callable` to be passed in and it will create and return its own `Future`. We would need to re-architect the clients a bit to support this, and I'm not even sure it could be done.
- `ParallelHttpShardHandler` does submit its own things, but I'm not sure that the `ExecutorCompletionService` will buy us a whole lot here, as it still needs to fit into the response handling that `HttpShardHandler` uses.
- There's no easy `cancelAll` logic that `ExecutorCompletionService.submit` can provide us, so we would still need to do that all ourselves.

Overall `ExecutorCompletionService` looks like a nice little utility, that is just a bit too simple to be plug-and-play for the complexities that we have to support.

gus-asf:
Running the various Limit related tests 100+ times each was something that revealed rare issues when I worked on this the first time. Have you done anything similar yet?

HoustonPutman:
First I go and change Line 75 in `NumericSearch` (in the benchmark module):

From
```java
miniClusterState.createCollection(COLLECTION, 1, 1);
```
to
```java
miniClusterState.createCollection(COLLECTION, 5, 1);
```

Then in `MiniSolrCloudCluster.java` line 117:
from:
```java
+ "  <shardHandlerFactory name=\"shardHandlerFactory\" class=\"HttpShardHandlerFactory\">\n"
```
to:
```java
+ "  <shardHandlerFactory name=\"shardHandlerFactory\" class=\"ParallelHttpShardHandlerFactory\">\n"
```

After this we can run the numericSearch benchmark:
```bash
cd solr/benchmark
./jmh.sh NumericSearch
```

gus-asf:
IIRC the timeout tests were hitting some edge cases caused by cancellation at unexpected points. The benchmark is unlikely to be canceling any requests (I haven't looked carefully at it, but that would seem unlikely in a benchmark)

HoustonPutman:
I did beast some of the limit tests as well to test out the cancelling. And I ran `TestTolerantSearch` with some changes to control order of cancellation of the shard requests, with some logging, to make sure that everything was working as expected.

I can run these some more as well.

dsmiley:
Fantastic peer review @gerlowskija :-)

fsparv:
I also was unable to break the new code using the limit tests (100 runs each), I'm still curious if there's a reproducible method for producing the problematic behavior with the suspected deadlocks, and observing the +50ms delay. I suspect the delay may have been a memory consistency issue (perhaps causing poll() to not return despite an available item). I still haven't come up with a logic to understand any deadlock.

I had previously been reluctant to change responseFutureMap to a ConcurrentHashMap since there's a different contract for nulls and it's pretty central, so any performance cost is a big deal  (though it's usage is shard/request proportional, and not document/field/token proportional so perhaps I shouldn't worry so much?). Also, I worry that folks who have implemented custom handlers could be disrupted by changing the underlying types. 

What are your thoughts on those considerations?

gus-asf:
* above comment is me (in the wrong browser)

HoustonPutman:
> I'm still curious if there's a reproducible method for producing the problematic behavior with the suspected deadlocks

Yeah, you can use the benchmark example (with 2 lines of code changes) that I shared above: https://github.com/apache/solr/pull/3398#issuecomment-3046296265

> since there's a different contract for nulls and it's pretty central, so any performance cost is a big deal

I don't think we use nulls at all here (correct me if I'm wrong) and yeah, performance cost is a big deal, but after looking at the benchmarking, it doesn't look like it makes any meaningful difference. I'll share those numbers (Throughput, higher is better):

| Benchmark  | HttpShardHandler (Not Fixed) | HttpShardHandler (Fixed) | ParallelHttpShardHandler (Not Fixed) | ParallelHttpShardHandler (Fixed) |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| NumericSearch.doubleDvSet  | 2986  | 2887  |  1.259 | 3026  |
| NumericSearch.doubleSet  | 2637  |  2611 | 1.271  | 2650  |
| NumericSearch.floatDvSet  | 2874  | 2883  |  0.544 | 2886  |
| NumericSearch.floatSet  | 2500  | 2598  | 0.588  | 2619  |
| NumericSearch.intDvSet  | 2906 |  2901 | 0.694  |  2801 |
| NumericSearch.intSet  | 2651  | 2667  |  0.545 | 2495  |
| NumericSearch.longDvSet  |  2978 |  2848 | 1.006  | 2748  |
| NumericSearch.longSet  | 2669  |  2635 | 0.336  | 2550  |

> Also, I worry that folks who have implemented custom handlers could be disrupted by changing the underlying types.

I am somewhat sympathetic to this, but it is a very complex class and those who are using internal implementation details in a subclass are going to need to be on their toes to do rewrites on minor version upgrades. If they have written their own implantation from scratch, then there should be no issue.

HoustonPutman:
So this broke `DistributedDebugComponentTest.testTolerantSearch` and `TestJsonFacets.testTolerant`. The reason is that originally `HttpShardHandler.take()` would return on an error, cancelling all other responses. Now the logic is that on an error, outstanding requests will be cancelled, but if any responses come in in the meantime, they will be processed. However, the code currently expects that the last response will have the error. So since the last response might not have the error anymore, we have to change the code that checks the responses for errors.

However, after fixing that and beasting the tests, there is a really weird error around cancelling requests. The `DistributedDebugComponentTest.testTolerantSearch` does a non-tolerant search then does a tolerant search. The error I described above was breaking the non-tolerant search. That is easily fixable. The second part of the test, testing tolerant search fails very occasionally (but only when the non-tolerant search is done first, when that is commented out, the tolerant search does not fail).

The tolerant search fails (occasionally) because all three shard requests fail instead of just 1 of the shard requests failing (because of a non-exisistant endpoint). the bad shard has the failure that the test expects, but the good shards both fail with `java.io.IOException: cancel_stream_error/unexpected_data_frame` meaning that the requests were cancelled, even thought the request is "tolerant". I did a lot of debugging here, and noticed that Solr is behaving correctly and we are not cancelling shard requests for tolerant solr requests. And the fact that if the "non-tolerant search" request case right before the tolerant search request is commented out, the failures stop, tell us that the cancellations from the non-tolerant request are bleeding into the tolerant request. This is bad. I also confirmed this by commenting out the line that actually cancels the HTTP requests: https://github.com/apache/solr/blob/branch_9_9/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java#L570-L574

This only happens on branch_9x (and presumable branch_9_9), not on main. So I believe it's a bug in Jetty 10, which Jetty 12 has solved. So we are probably fine just fixing this part on branch_9x and branch_9_9, and leaving the request cancellation enabled on main (10.x).

Amazingly, when beasting, there is a big difference in whether the non-existent endpoint is put first or last in the list of shards. The failure rate is much higher when the bad shard is the first listed rather the last one listed.

I'm going to create another ticket for the cancellation bug. But I'll create another PR to fix what this PR broke (the first part of this long comment).

PR #3399, opened 2025-06-20
SOLR-16391: Fix CoresAPI integration in V2HttpCall

HoustonPutman:

We might be able to move this else if statement entirely, if all cores APIs are handled by JaxRS, but I may be wrong about that. It's just a guess.

### DIFF
--- a/solr/core/src/java/org/apache/solr/api/V2HttpCall.java
+++ b/solr/core/src/java/org/apache/solr/api/V2HttpCall.java
@@ -175,7 +175,7 @@ public class V2HttpCall extends HttpSolrCall {
             }
           }
         }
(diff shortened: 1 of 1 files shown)


gerlowskija (approved):
What this really needs to accompany it are some more tests for the v2 API - like those that Colvin started [here](https://github.com/apache/solr/pull/3023).  I say it as someone who's at least partially responsible, but the code in V2HttpCall is very brittle...Far too many times I've tried to fix a bug in V2HttpCall and ended up breaking some other case.  A lot of whack-a-mole style issues.

There's already a separate PR for those though, so IMO it's not worth gating this on automated tests.

LGTM!

PR #3400, opened 2025-06-20
SOLR-17776: Fix overchecking of http1 flag in Http2SolrClient

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17776

The http1 check should work like the rest of the custom option checks, and only fail if the user has explicitly set it to something (other than what the existing httpClient uses).

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
@@ -132,8 +132,9 @@ public class Http2SolrClient extends HttpSolrClientBase {
       if (builder.followRedirects != null
           || builder.connectionTimeoutMillis != null
           || builder.maxConnectionsPerHost != null
-          || builder.useHttp1_1
-              != builder.httpClient.getTransport() instanceof HttpClientTransportOverHTTP
+          || (builder.providedUseHttp1_1 != null
+              && builder.providedUseHttp1_1
+                  != builder.httpClient.getTransport() instanceof HttpClientTransportOverHTTP)
           || builder.proxyHost != null
           || builder.sslConfig != null
           || builder.cookieStore != null
@@ -273,7 +274,7 @@ public class Http2SolrClient extends HttpSolrClientBase {
 
     HttpClient httpClient;
     HttpClientTransport transport;
-    if (builder.useHttp1_1) {
+    if (builder.shouldUseHttp1_1()) {
       if (log.isDebugEnabled()) {
         log.debug("Create Http2SolrClient with HTTP/1.1 transport");
       }
@@ -335,7 +336,7 @@ public class Http2SolrClient extends HttpSolrClientBase {
       proxy = new Socks4Proxy(address, builder.proxyIsSecure);
     } else {
       final Protocol protocol;
-      if (builder.useHttp1_1) {
+      if (builder.shouldUseHttp1_1()) {
         protocol = HttpClientTransportOverHTTP.HTTP11;
       } else {
         // see HttpClientTransportOverHTTP2#newOrigin
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
@@ -38,7 +38,8 @@ public abstract class HttpSolrClientBuilderBase<
   protected Set<String> urlParamNames;
   protected Integer maxConnectionsPerHost;
   protected ExecutorService executor;
-  protected boolean useHttp1_1 = Boolean.getBoolean("solr.http1");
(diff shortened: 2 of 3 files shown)


solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
```diff
@@ -38,7 +38,8 @@ public abstract class HttpSolrClientBuilderBase<
   protected Set<String> urlParamNames;
   protected Integer maxConnectionsPerHost;
   protected ExecutorService executor;
-  protected boolean useHttp1_1 = Boolean.getBoolean("solr.http1");
+  protected final boolean defaultUseHttp1_1 = Boolean.getBoolean("solr.http1");
```
dsmiley, line 41: this could be static (and thus moved above to statics); right?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
```diff
@@ -38,7 +38,8 @@ public abstract class HttpSolrClientBuilderBase<
...
+  protected Boolean providedUseHttp1_1;
```
dsmiley: personally, I'm not a fan of "provided".  The vast majority of fields here are in-effect "provided", and they don't need this name prefix.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
```diff
@@ -167,10 +168,26 @@ public long getRequestTimeoutMillis() {
    */
   @SuppressWarnings("unchecked")
   public B useHttp1_1(boolean useHttp1_1) {
-    this.useHttp1_1 = useHttp1_1;
+    this.providedUseHttp1_1 = useHttp1_1;
     return (B) this;
   }
 
+  /**
+   * Return whether the HttpSolrClient built will prefer http1.1 over http2. This will be determined
```
dsmiley: I think this is saying too much honestly.  Docs need to be maintained.  It exposes little details and possibly repeats things elsewhere.  *every* getter (lets say this is a getter), will use what is provided if the user supplies a choice.  Let's not say the obvious, since then we'll be tempted to repeat this again and again and again...   
Secondly, any getter like method (includes this) on a builder is really for internal use, so I wouldn't bother documenting it.  Thirdly, if you want to document the semantics of what this settings means, the place to do that is on the setter (or "with" method).  It's already documented.  So just drop this IMO.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
```diff
@@ -167,10 +168,26 @@ public long getRequestTimeoutMillis() {
...
+   *
+   * @return whether the build HttpSolrClient will prefer http1.1
+   */
+  @SuppressWarnings("unchecked")
+  public boolean shouldUseHttp1_1() {
+    if (providedUseHttp1_1 != null) {
+      return providedUseHttp1_1;
+    } else {
+      return defaultUseHttp1_1;
+    }
```
dsmiley: ternary please

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
```diff
@@ -132,8 +132,9 @@ protected Http2SolrClient(String serverBaseUrl, Builder builder) {
       if (builder.followRedirects != null
           || builder.connectionTimeoutMillis != null
           || builder.maxConnectionsPerHost != null
-          || builder.useHttp1_1
-              != builder.httpClient.getTransport() instanceof HttpClientTransportOverHTTP
+          || (builder.providedUseHttp1_1 != null
+              && builder.providedUseHttp1_1
+                  != builder.httpClient.getTransport() instanceof HttpClientTransportOverHTTP)
```
dsmiley: I think we now can *and should* drop this part.  Just keep this looking _just_ like the other settings with null checks; very visually obviously consistent instead of some quirk to ponder.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
HoustonPutman: I chose provided for this one, so that people in the future maybe think twice before using it instead of the `shouldUseHttp1_1()` method to check whether or not to use http1. Not a huge concern, and I can change it.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClientBuilderBase.java
HoustonPutman: I don't think it should be static, because test cases almost certainly try to change it I imagine...

dsmiley (approved):

PR #3402, opened 2025-06-23
SOLR-17793: Create dedicated SdkMeterProviders for Solr core metrics

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-17793

OTEL metrics are immutable once created for a meter provider. This is problematic for creation and deletion of Solr Cores keeping stale metrics. To do this we do a few things:

1. Reuse Dropwizards concept of `registries` and tie it to a dedicated `SdkMeterProvider` for each Solr Core with a registered Prometheus Metric Reader.
2. With the use of dedicated SdkmeterProviders, we move away from using `GlobalOpenTelemetry` singleton instance for metrics
3. When deleting a core, close the corresponding `SdkMeterProvider` and its metric readers with it

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -2280,9 +2280,7 @@ public class CoreContainer {
     }
 
     // delete metrics specific to this core
-    // NOCOMMIT: Remove this
     metricManager.removeRegistry(core.getCoreMetricManager().getRegistryName());
-    metricManager.closeMeterProvider(core.getCoreMetricManager().getRegistryName());
 
     if (zkSys.getZkController() != null) {
       // cancel recovery in cloud mode
--- a/solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
+++ b/solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
@@ -621,10 +621,6 @@ public class SolrMetricManager {
     return set;
   }
 
-  public Set<String> providerNames() {
-    return meterProviderAndReaders.keySet();
-  }
-
   /**
    * Check whether a registry with a given name already exists.
    *
@@ -753,12 +749,14 @@ public class SolrMetricManager {
   }
 
   /**
-   * Remove a named registry.
+   * Remove a named registry and close an existing {@link SdkMeterProvider}. Upon closing of
+   * provider, all metric readers registered to it are closed.
    *
    * @param registry name of the registry to remove
    */
   // TODO SOLR-17458: You can't delete OTEL meters
   public void removeRegistry(String registry) {
+    // NOCOMMIT Remove all closing Dropwizard registries
     // close any reporters for this registry first
     closeReporters(registry, null);
     // make sure we use a name with prefix
@@ -773,17 +771,8 @@ public class SolrMetricManager {
         swapLock.unlock();
       }
     }
-  }
-
-  /**
-   * Remove and close an existing {@link SdkMeterProvider}. Upon closing of provider, all metric
-   * readers registered to it are closed.
-   *
-   * @param providerName name of the Meter Provider to remove
-   */
-  public void closeMeterProvider(String providerName) {
     meterProviderAndReaders.computeIfPresent(
-        enforcePrefix(providerName),
+        registry,
         (key, meterAndReader) -> {
           meterAndReader.sdkMeterProvider().close();
           return null;
--- a/solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
+++ b/solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
@@ -78,7 +78,6 @@ public class TestPrometheusResponseWriter extends SolrTestCaseJ4 {
 
     try (SolrClient adminClient = getHttpSolrClient(solrClientTestRule.getBaseUrl())) {
       NamedList<Object> res = adminClient.request(req);
-      assertNotNull("null response from server", res);
       String output = (String) res.get("response");
 
       Set<String> seenTypeInfo = new HashSet<>();
--- a/solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriterCloud.java
+++ b/solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriterCloud.java
@@ -28,14 +28,11 @@ import org.apache.solr.client.solrj.request.GenericSolrRequest;
 import org.apache.solr.cloud.SolrCloudTestCase;
 import org.apache.solr.common.params.ModifiableSolrParams;
 import org.apache.solr.common.util.NamedList;
-import org.apache.solr.util.SolrJettyTestRule;
 import org.junit.Before;
 import org.junit.BeforeClass;
-import org.junit.ClassRule;
 import org.junit.Test;
 
 public class TestPrometheusResponseWriterCloud extends SolrCloudTestCase {
-  @ClassRule public static SolrJettyTestRule solrClientTestRule = new SolrJettyTestRule();
 
   @BeforeClass
   public static void setupCluster() throws Exception {
@@ -80,28 +77,23 @@ public class TestPrometheusResponseWriterCloud extends SolrCloudTestCase {
             new ModifiableSolrParams().set("wt", "prometheus"));
     req.setResponseParser(new InputStreamResponseParser("prometheus"));
 
-    NamedList<Object> prometheusResponse = solrClient.request(req);
-    assertNotNull("null response from server", prometheusResponse);
-    NamedList<Object> res = solrClient.request(req);
-    InputStream in = (InputStream) prometheusResponse.get("stream");
-
-    assertNotNull("null response from server", res);
-    String output = new String(in.readAllBytes(), StandardCharsets.UTF_8);
-
-    output
-        .lines()
-        .forEach(
-            line -> {
-              if (line.startsWith("solr_metrics_core_requests_total")
-                  && line.contains("handler=\"/select\""))
-                assertTrue(
-                    "Missing expected Solr cloud mode prometheus labels in metric: " + line,
-                    line.contains("collection=\"collection1\"")
-                        && line.contains("core=\"collection1_shard1_replica_n1\"")
-                        && line.contains("replica=\"replica_n1\"")
-                        && line.contains("shard=\"shard1\"")
-                        && line.contains("type=\"requests\""));
-            });
+    NamedList<Object> resp = solrClient.request(req);
+    try (InputStream in = (InputStream) resp.get("stream")) {
+      String output = new String(in.readAllBytes(), StandardCharsets.UTF_8);
+      assertTrue(
+          "Missing expected Solr cloud mode prometheus metric with cloud labels",
+          output
+              .lines()
+              .anyMatch(
+                  line ->
+                      line.startsWith("solr_metrics_core_requests_total")
+                          && line.contains("handler=\"/select\"")
+                          && line.contains("collection=\"collection1\"")
+                          && line.contains("core=\"collection1_shard1_replica_n1\"")
+                          && line.contains("replica=\"replica_n1\"")
+                          && line.contains("shard=\"shard1\"")
+                          && line.contains("type=\"requests\"")));
+    }
   }
 
   @Test
@@ -113,58 +105,60 @@ public class TestPrometheusResponseWriterCloud extends SolrCloudTestCase {
     solrClient.query("collection1", query);
     solrClient.query("collection2", query);
 
-    var promReq =
+    var req =
         new GenericSolrRequest(
             METHOD.GET,
             "/admin/metrics",
             SolrRequestType.ADMIN,
             new ModifiableSolrParams().set("wt", "prometheus"));
-    promReq.setResponseParser(new InputStreamResponseParser("prometheus"));
+    req.setResponseParser(new InputStreamResponseParser("prometheus"));
 
-    NamedList<Object> prometheusResponse = solrClient.request(promReq);
-    assertNotNull("null response from server", prometheusResponse);
+    NamedList<Object> resp = solrClient.request(req);
 
-    InputStream in = (InputStream) prometheusResponse.get("stream");
-    String output = new String(in.readAllBytes(), StandardCharsets.UTF_8);
+    try (InputStream in = (InputStream) resp.get("stream")) {
+      String output = new String(in.readAllBytes(), StandardCharsets.UTF_8);
 
-    assertTrue(
-        "Prometheus output should contains solr_metrics_core_requests for collection1",
-        output
-            .lines()
-            .anyMatch(
-                line ->
-                    line.startsWith("solr_metrics_core_requests") && line.contains("collection1")));
-    assertTrue(
-        "Prometheus output should contains solr_metrics_core_requests for collection2",
-        output
-            .lines()
-            .anyMatch(
-                line ->
-                    line.startsWith("solr_metrics_core_requests") && line.contains("collection2")));
+      assertTrue(
+          "Prometheus output should contains solr_metrics_core_requests for collection1",
+          output
+              .lines()
+              .anyMatch(
+                  line ->
+                      line.startsWith("solr_metrics_core_requests")
+                          && line.contains("collection1")));
+      assertTrue(
+          "Prometheus output should contains solr_metrics_core_requests for collection2",
+          output
+              .lines()
+              .anyMatch(
+                  line ->
+                      line.startsWith("solr_metrics_core_requests")
+                          && line.contains("collection2")));
+    }
 
     // Delete collection and assert metrics have been removed
     var deleteRequest = CollectionAdminRequest.deleteCollection("collection1");
     deleteRequest.process(solrClient);
 
-    prometheusResponse = solrClient.request(promReq);
-    assertNotNull("null response from server", prometheusResponse);
-
-    in = (InputStream) prometheusResponse.get("stream");
-    output = new String(in.readAllBytes(), StandardCharsets.UTF_8);
-
-    assertFalse(
-        "Prometheus output should not contain solr_metrics_core_requests after collection was deleted",
-        output
-            .lines()
-            .anyMatch(
-                line ->
-                    line.startsWith("solr_metrics_core_requests") && line.contains("collection1")));
-    assertTrue(
-        "Prometheus output should contains solr_metrics_core_requests for collection2",
-        output
-            .lines()
-            .anyMatch(
-                line ->
-                    line.startsWith("solr_metrics_core_requests") && line.contains("collection2")));
+    resp = solrClient.request(req);
+    try (InputStream in = (InputStream) resp.get("stream")) {
+      String output = new String(in.readAllBytes(), StandardCharsets.UTF_8);
+      assertFalse(
+          "Prometheus output should not contain solr_metrics_core_requests after collection was deleted",
+          output
+              .lines()
+              .anyMatch(
+                  line ->
+                      line.startsWith("solr_metrics_core_requests")
+                          && line.contains("collection1")));
+      assertTrue(
+          "Prometheus output should contains solr_metrics_core_requests for collection2",
+          output
+              .lines()
+              .anyMatch(
+                  line ->
+                      line.startsWith("solr_metrics_core_requests")
+                          && line.contains("collection2")));
+    }
   }
 }
--- a/solr/core/src/test/org/apache/solr/metrics/SolrMetricManagerTest.java
+++ b/solr/core/src/test/org/apache/solr/metrics/SolrMetricManagerTest.java
@@ -518,7 +518,7 @@ public class SolrMetricManagerTest extends SolrTestCaseJ4 {
             .orElseThrow(() -> new AssertionError("LongCounter metric not found"));
     assertEquals(5, data.getDataPoints().getFirst().getValue(), 0.0);
 
-    metricManager.closeMeterProvider(METER_PROVIDER_NAME);
+    metricManager.removeRegistry(METER_PROVIDER_NAME);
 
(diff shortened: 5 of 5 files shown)


solr/core/src/java/org/apache/solr/core/OpenTelemetryConfigurator.java
```diff
@@ -71,28 +67,21 @@ public static synchronized void initializeOpenTelemetrySdk(
...
     if (TRACE_ID_GEN_ENABLED) {
       log.info("OpenTelemetry tracer enabled with simple propagation only.");
       ExecutorUtil.addThreadLocalProvider(new ContextThreadLocalProvider());
-      builder.setPropagators(ContextPropagators.create(SimplePropagator.getInstance()));
     }
 
-    if (sdkMeterProvider != null) builder.setMeterProvider(sdkMeterProvider);
-    if (sdkTracerProvider != null) builder.setTracerProvider(sdkTracerProvider);
-
-    GlobalOpenTelemetry.set(builder.build());
```
mlbiscoc, line 95: We stop using `GlobalOpenTelemetry` instance as you are fixed to 1 MeterProvider for the whole process and can't be modified. So reverted OTELs initialization of its `SimplePropogator`.

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
```diff
@@ -115,6 +115,8 @@ public class SolrMetricManager {
 
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
+  public static final String OTEL_SCOPE_NAME = "org.apache.solr";
```
mlbiscoc, line 118: All metrics in Solr will have this as OTEL scope name

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
```diff
@@ -147,25 +149,19 @@ public class SolrMetricManager {
   private final MetricRegistry.MetricSupplier<Timer> timerSupplier;
   private final MetricRegistry.MetricSupplier<Histogram> histogramSupplier;
 
-  private final MeterProvider meterProvider;
-  private final Map<String, io.opentelemetry.api.metrics.Meter> meters = new ConcurrentHashMap<>();
-  private final PrometheusMetricReader metricReader;
+  private final ConcurrentMap<String, MeterProviderAndReaders> meterProviderAndReaders =
+      new ConcurrentHashMap<>();
```
mlbiscoc, line 153: We wrap a Meter Provider and its corresponding Metric Readers together. Currently the only metric reader is Prometheus but additional metric reader could be a periodic reader to push with OTLP. This is essentially our concept of dropwizard's `registry`

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
```diff
@@ -40,12 +48,107 @@ public class PrometheusResponseWriter implements QueryResponseWriter {
...
     return CONTENT_TYPE_PROMETHEUS;
   }
+
+  /**
+   * Merge a collection of individual {@link MetricSnapshot} instances into one {@link
+   * MetricSnapshots}. This is necessary because we create a {@link SdkMeterProvider} per Solr core
+   * resulting in duplicate metric names across cores which is an illegal format if not under the
+   * same prometheus grouping.
+   */
+  private MetricSnapshots mergeSnapshots(List<MetricSnapshot> snapshots) {
```
mlbiscoc, line 72: This is a bit annoying that it had to be done. All Solr Cores have the same metric names with just different labels, but when using dedicated MeterProviders they are considered duplicated names running into [this](https://github.com/apache/solr/pull/2902) bug I fixed a while ago because they aren't under the same grouping. So we need to merge the duplicate metrics into a single prometheus grouping so we are compliant with Prometheus exposition format.

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
```diff
@@ -16,134 +16,198 @@
...
 import org.junit.AfterClass;
+import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class TestPrometheusResponseWriter extends SolrTestCaseJ4 {
+public class TestPrometheusResponseWriter extends SolrCloudTestCase {
```
mlbiscoc: I moved to SolrCloudTestCase since all of this was being tested in standalone mode. All of these should test in cloud mode

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
```diff
@@ -16,134 +16,198 @@
...
+            Float.parseFloat(actualValue);
+          } catch (NumberFormatException e) {
+            log.warn("Prometheus value not a parsable float");
+            assertTrue(VALID_PROMETHEUS_VALUES.contains(actualValue));
+          }
+        });
+  }
+
+  @Test
+  public void testCollectionDeletePrometheusOutput() throws Exception {
```
mlbiscoc: I don't know if this should be here as its a full high level test. Maybe a lower level test for this in `MetricsHandlerTest` or just a unit test in `TestSolrMetricManager` is suffice?

mlbiscoc:
With dedicated SDK Meter Providers, this will not automatically work with the Java Agent as that relies on a linking with a `GlobalOpenTelemetry` which has a single SdkMeterProvider and so we lose auto-configuration that tracing gets and OTLP. So for OTLP push we need to manually set the SDK up. I have another change that I did not add to this PR to keep it in scope, but essentially to `MetricManager` we add a `OtlpExporter` and register a `PeriodicMetricReader` with the `MeterProvider` to push tjese metrics with OTLP.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -849,6 +847,7 @@ private void loadInternal() {
         new SolrMetricsContext(
             metricManager, SolrMetricManager.getRegistryName(SolrInfoBean.Group.node), metricTag);
 
+    // NOCOMMIT: Do we need this for OTEL or is this specific for reporters?
```
dsmiley, line 850: It's not specific to otel; it's generic.  At the moment, I see it's only used for async core loading, which feels needless but anyway another topic.  I may put up a trivial PR.

solr/core/src/java/org/apache/solr/core/OpenTelemetryConfigurator.java
dsmiley: You say that but the code here continues to set GlobalOpenTelemetry :shrug:

FWIW I think we should continue to use GOT but only for the container instance, if that makes sense.

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
```diff
@@ -1587,7 +1671,18 @@ public MetricsConfig getMetricsConfig() {
...
+  /** Get a shallow copied map of {@link PrometheusMetricReader}. */
+  public Map<String, PrometheusMetricReader> getPrometheusMetricReaders() {
+    return meterProviderAndReaders.entrySet().stream()
+        .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().prometheusMetricReader()));
   }
+
+  public PrometheusMetricReader getPrometheusMetricReader(String providerName) {
+    var name = enforcePrefix(providerName);
+    if (!meterProviderAndReaders.containsKey(name)) return null;
+    return meterProviderAndReaders.get(name).prometheusMetricReader();
```
dsmiley: calling `contains` and then conditionally `get` is a minor anti-pattern since it's a double-lookup.

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
dsmiley, line 1: The switch of the name "registry" to "provider" seems needless to me.  Even if we like the new name better, I think bigger renames should be deferred.
The former has more meaning than the latter, which is super generic.

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
```diff
@@ -40,12 +48,107 @@ public class PrometheusResponseWriter implements QueryResponseWriter {
...
+    Map<String, CounterSnapshot.Builder> counterSnapshotMap = new HashMap<>();
+    Map<String, GaugeSnapshot.Builder> gaugeSnapshotMap = new HashMap<>();
+    Map<String, HistogramSnapshot.Builder> histogramSnapshotMap = new HashMap<>();
+    InfoSnapshot otelInfoSnapshots = null;
+
+    for (MetricSnapshot snapshot : snapshots) {
+      String metricName = snapshot.getMetadata().getPrometheusName();
+
+      switch (snapshot) {
+        case CounterSnapshot counterSnapshot -> {
```
dsmiley, line 82: in each `case`, you are calling computeIfAbsent yet not saving the result.  You then lookup again.  Please combine.

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
```diff
@@ -40,12 +48,107 @@ public class PrometheusResponseWriter implements QueryResponseWriter {
...
+                var base =
+                    CounterSnapshot.builder()
+                        .name(counterSnapshot.getMetadata().getName())
+                        .help(counterSnapshot.getMetadata().getHelp());
+
+                return counterSnapshot.getMetadata().hasUnit()
+                    ? base.unit(counterSnapshot.getMetadata().getUnit())
+                    : base;
+              });
+          counterSnapshot.getDataPoints().forEach(counterSnapshotMap.get(metricName)::dataPoint);
```
dsmiley: it's not clear what these lines actually accomplish.  You loop stuff but appear to do do nothing with any results.  _disclaimer_: I'm merely looking via GitHub UI.

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
dsmiley: What does it matter?  In general, tests should be lean, only testing at the highest level that's pertinent to the feature.  Metrics don't have to do with SolrCloud.

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
dsmiley: I suggest a separate test suite so that TestPrometheusResponseWriter doesn't need SolrCloud.

solr/core/src/java/org/apache/solr/core/OpenTelemetryConfigurator.java
mlbiscoc: I should have been more clear here. We are not using `GlobalOpenTelemetry` for metrics here anymore but traces still use it for the `SimplePropogator` now which is why I am still setting the `GlobalOpenTelemetry`.

Not sure what you mean using `GlobalOpenTelemetry` only for the container instance? If we continue with our manage dediated SDKMeterProviders and the GlobalOpenTelemetry for metrics things might get messy. Different metric readers and exporters managed by different things. Autoconfiguration would work for only the container instance but not our instrumented sdk providers.

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
mlbiscoc: Changed it to provider as OTEL doesn't call this a registry to be consistent. I can switch it back if you prefer in this PR. Makes the line changes smaller as well.

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
mlbiscoc: Honestly maybe my java docs weren't clear enough. Essentially we need to return a single `Snapshots` object (Not to confuse with `Snapshot` object) that has all of our metrics. All these lines combine equivalent metrics names together. With multiple metric readers because of multiple meter providers, Prometheus structures the data like so between 2 cores from 2 `Snapshot`

**Counter snapshot: core_1**
- metric name “requests_total”
**Datapoints**
- {core: “core_1”, path: “/select”}, value = 10

**Counter snapshot: core_2**
- metric name “requests_total”
**Datapoints**
- {core: “core_2”, path: “/select”}, value = 20

We then need to merge these into a single Counter snapshot otherwise we output 2 `Snapshots` with duplicate metric name. So we take the builder for that specific metric name and merge them into a single snapshot.

**Counter snapshot: merged**
- metric name “requests_total”
**Datapoints**
- {core: “core_1”, path: “/select”}, value = 10
- {core: “core_2”, path: “/select”}, value = 20

At the very end of this method, we build all the snapshots and create the final `Snapshots` object that the prometheus write can output without the duplicate TYPE and DESC information.

I tested this as well with an an internal OTLP exporting in solr to a OTEL collector and the collector seems to do some kind merging like this to avoid the duplicate information.

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
mlbiscoc: Solr defaults to Cloud mode, so felt it make sense to move to cloud mode. Lets you also get specific labels from cloud. For example, all cloud and standalone core metrics have the label `coreName` but `collection` or `shard` is cloud specific. Lets us test for that with CloudTestCase but I get where you're coming from.

I can revert it back if you prefer. I was going to make a separate test suite like you recommended below anyways and move `testCollectionDeletePrometheusOutput` that I made.

solr/core/src/java/org/apache/solr/core/OpenTelemetryConfigurator.java
dsmiley: Oh I see; nevermind my suggestion then..

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
dsmiley: > Solr defaults to Cloud mode

That's not relevant.  Would you then suggest _every_ test in Solr run SolrCloud?  It's been difficult to keep our test times down (we used to be at ~40min); I shudder to think if every test had to run SolrCloud.

> all cloud and standalone core metrics have the label coreName but collection or shard is cloud specific

Okay then I don't have a strong preference.  If you do wind up separating the higher level integration test, maybe it could have check at least one core metric has the collection & shard.

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
dsmiley: Yes, please change the name; thanks.

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
dsmiley: oh I am mistaken; the result is put into `builder`.

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
dsmiley: Thanks!  What I specifically was confused about was the line of code that had an appearance of being side-effect free, thus pointless.  But the method reference `dataPoint` _adds_ a data point.  It looks ambiguous to my fresh eyes on this API.

Having to write all this here is a bit of a PITA

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriter.java
mlbiscoc: Thats understandable. I went with meeting in the middle and wrote a test to just check a single metric that the cloud labels exist and moved `testCollectionDeletePrometheusOutput` into that test suite. Reverted everything else back.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -2281,7 +2280,9 @@ private void unloadWithoutCoreOp(
     }
 
     // delete metrics specific to this core
+    // NOCOMMIT: Remove this
     metricManager.removeRegistry(core.getCoreMetricManager().getRegistryName());
+    metricManager.closeMeterProvider(core.getCoreMetricManager().getRegistryName());
```
dsmiley: you could restore the previous name

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
```diff
@@ -578,13 +613,18 @@ public boolean matches(String s, Metric metric) {
...
   /** Return a set of existing registry names. */
+  // NOCOMMIT: Remove for OTEL
   public Set<String> registryNames() {
     Set<String> set = new HashSet<>();
     set.addAll(registries.keySet());
     set.addAll(SharedMetricRegistries.names());
     return set;
   }
 
+  public Set<String> providerNames() {
```
dsmiley: is this needed?  We're calling these registries...

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriterCloud.java
```diff
@@ -0,0 +1,170 @@
...
+import org.apache.solr.common.params.ModifiableSolrParams;
+import org.apache.solr.common.util.NamedList;
+import org.apache.solr.util.SolrJettyTestRule;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+public class TestPrometheusResponseWriterCloud extends SolrCloudTestCase {
+  @ClassRule public static SolrJettyTestRule solrClientTestRule = new SolrJettyTestRule();
```
dsmiley: what is this doing here?  It's a standalone Solr instance inside a SolrCloud test!

For your awareness, I have a plan to hopefully make a TestRule for SolrCloud, so that we need not extend SolrCloudTestCase just to test SolrCloud.

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriterCloud.java
```diff
@@ -0,0 +1,170 @@
...
+    assertNotNull("null response from server", res);
+    String output = new String(in.readAllBytes(), StandardCharsets.UTF_8);
+
+    output
+        .lines()
+        .forEach(
+            line -> {
+              if (line.startsWith("solr_metrics_core_requests_total")
+                  && line.contains("handler=\"/select\""))
+                assertTrue(
```
dsmiley: we should check that at least one line reaches this point :-)
Maybe use the `anyMatch` technique that you use later.

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriterCloud.java
```diff
@@ -0,0 +1,170 @@
...
+    var promReq =
+        new GenericSolrRequest(
+            METHOD.GET,
+            "/admin/metrics",
+            SolrRequestType.ADMIN,
+            new ModifiableSolrParams().set("wt", "prometheus"));
+    promReq.setResponseParser(new InputStreamResponseParser("prometheus"));
+
+    NamedList<Object> prometheusResponse = solrClient.request(promReq);
+    assertNotNull("null response from server", prometheusResponse);
```
dsmiley: IMO this is never needed

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriterCloud.java
```diff
@@ -0,0 +1,170 @@
...
+
+    InputStream in = (InputStream) prometheusResponse.get("stream");
```
dsmiley: these streams need to be closed

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
dsmiley, line 1: I think a tricky thing here right now is that SolrMetricManager is holding both DropWizard & OTEL metrics.  Thus only one can have the method name we want (i.e. one with registry in the name).  Seems you should make the leap to OTEL completely soon.  Could go most of the way by retaining the DropWizard pervasive methods like `counter`, etc. and return `NoopCounter` and similar until it can completely go.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
mlbiscoc: Do you mean keep the name `removeRegistry` instead of a separate? If so, sure then I'll just move the closing of the meterProviders into `removeRegistry` until we completely move off. Made that change in latest commit

solr/core/src/test/org/apache/solr/response/TestPrometheusResponseWriterCloud.java
mlbiscoc: Mistake leaving that in. Was the result of me copy and pasting some things and modifying...

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
mlbiscoc: Its mostly the tests that blow the number of changes which is why I was trying to add otel in parallel then incrementally migrating a subset of dropwizard metrics and the corresponding tests using them. Since everything goes through metric manager, if I switch it to NoopCounter then every test relying on counter will fail. I'm fine just moving over and making one large PR with the migration by returning a NoopCounter if you prefer and then moving to OTEL completely. 

But I also found a while ago there is something in solr that was parsing the dropwizard output and using those metrics. It may have been the UI or something but it was a failing a test and I found something about it making the fix not simple. Don't fully remember but I'll look for it again.

I have another somewhat small PR I wanted to put after this that add in a OTLP exporter that can be enabled. Although its debatable where that code should live which is what I wanted to bring up next.

dsmiley (approved):

PR #3403, opened 2025-06-24
SOLR-17794: Log content stream JSON payload as cs.params in SlowRequest log

mkhludnev:
https://issues.apache.org/jira/browse/SOLR-17794


mkhludnev:
I'm in doubts about a huge payload which is printed in log here. However, we definitely should dump some of parameters into SlowRequest log.

PR #3404, opened 2025-06-25
SOLR-17778: Remove unused FORCEPREPAREFORLEADERSHIP operation

psalagnac:
https://issues.apache.org/jira/browse/SOLR-17778

Trivial removal of dead code.
This removes a core admin operation that does nothing, but was still visible in SolrJ.


psalagnac:
> LGTM. Interesting no test needed removing as well, I guess that either never existed or was removed already?

Caller was removed with https://github.com/apache/lucene-solr/commit/27eb6ba but the cleanup wasn't completed. Not sure abouts tests. I found none.

PR #3405, opened 2025-06-25
Refactor: remove coreContainerWorkExecutor

dsmiley:
It's only used for one silly thing:
It's needless to create an executor just to asynchronously wait for another executor (coreLoadExecutor) to close.  We can instead call shutdownAndAwaitTermination on coreLoadExecutor when the container shuts down.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -1347,7 +1347,7 @@ public class CoreContainer {
         zkSys.zkController.tryCancelAllElections();
       }
 
-      ExecutorUtil.shutdownAndAwaitTermination(coreLoadExecutor);
+      ExecutorUtil.shutdownAndAwaitTermination(coreLoadExecutor); // actually already shutdown
 
       // First wake up the closer thread, it'll terminate almost immediately since it checks
       // isShutDown.
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -1086,14 +1086,15 @@ public class CoreContainer {
     }
 
     // setup executor to load cores in parallel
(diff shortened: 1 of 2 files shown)


solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1362,7 +1347,7 @@ public void shutdown() {
         zkSys.zkController.tryCancelAllElections();
       }
 
-      ExecutorUtil.shutdownAndAwaitTermination(coreContainerWorkExecutor);
+      ExecutorUtil.shutdownAndAwaitTermination(coreLoadExecutor);
```
psalagnac: Not harmful, but I'm not sure why we have to shutdown the executor once again here, since we systematically call shutdown() in the `loadInternal()` method.

The only case where we want to wait for its completion here is when the Solr server is being stopped before all core loads completed.

dsmiley: Oh right; I should change this to `awaitTermination`.  Thanks!

dsmiley: But it's harmless to call shutdown again.  And it's more normal to see this method call here.  Only seeing awaitTermination is unusual... the reader will wonder what's going on.  If I change it, I suggest then 4 lines of code:
```java
      if (coreLoadExecutor != null) {
        assert coreLoadExecutor.isShutdown(); // set in a finally part of loadInternal
        ExecutorUtil.awaitTermination(coreLoadExecutor);
      }
```
Which I think is too much so lets keep this?

psalagnac: The assert might fail if we shutdown the CoreContainer before `loadInternal()` completed, isn't it? I don't know if this scenario may happen in practice.

> But it's harmless to call shutdown again. And it's more normal to see this method call here. Only seeing awaitTermination is unusual..

Agreed! I would just add a comment that most of the time, this executor should be already stopped when we reach this statement.

PR #3406, opened 2025-06-26
SOLR-17641: Remove use of security manager

epugh:
https://issues.apache.org/jira/browse/SOLR-17641.

# Description
Remove the Java Security manager as the first step to supporting Java 24.

# Solution

Removing it.

# Tests

Re running all tests, only modifying them to not assume/require/not require the security manager.

# Checklist


dsmiley (commented):
Technically, it is up to Oracle to remove the SecurityManager; Solr can't remove it; this PR doesn't :-)   And this PR seems to only change the default, thus make it opt-in.  But you removed the solr-tests.policy, so I don't see how anyone would successfully run Solr with the SecurityManager for the build without the test policy.

epugh:
My goal was to remove Solr use of security manager, so no opt in or opt out....   hence removing all policy files.

Thoughts on the desirability of merging this for Solr 10?   Do we have to wait for Java 24 being required/min version before finishing migrating away from the Security Manager?

reneleonhardt:
I mean... since March nobody can install SOLR anymore with the latest JDK anyway, so maybe there's no rush to release it as a v9 "patch" 😅

epugh:
@reneleonhardt can you elaborate a bit more?  I don't think I understand?   Solr won't run on Java 24?

reneleonhardt:
Your description says
```
Remove the Java Security manager as the first step to supporting Java 24.
```

To me it sounds like SOLR can't be run since Java 24 anymore and that was the reason for you to remove the security manager.

epugh:
humm...   good question.  If you are on Java 24 and you set the security manager to be disabled, does it all work?  dunno!

dsmiley:
Based on Jan's comment on the JIRA issue, I think we can close this PR without merging.

epugh:
Agreed!

PR #3410, opened 2025-06-27
SOLR-17582: Fix ClusterState serializing for older SolrJ versions

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17582

User Agent versions < 9.9.0 work, but not the old 1.0 or 2.0 agent versions used by Solr. We should catch these versions and just change them to be semVer compliant.

### DIFF
--- a/solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
+++ b/solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
@@ -1193,7 +1193,13 @@ public class HttpSolrCall {
       return null;
     }
     try {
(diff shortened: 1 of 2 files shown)


dsmiley (approved):
Thanks!

I appreciate the effort of the bespoke test but I feel as a project we're missing general higher level integration test using Docker.  A general test would break as such breaking things happen and we wouldn't feel much need to have this unit test.  We don't have such integration tests... so my point isn't really a critique of this test so much as a wish for a better future.

HoustonPutman:
Yeah, completely agreed. That would be awesome, but definitely for a future ticket. (Though we always do say that unfortunately)

PR #3411, opened 2025-06-30
SOLR-17804: re-create SolrClient if closed.

sigram:
https://issues.apache.org/jira/browse/SOLR-17804

Use a `Supplier<CloudSolrClient>` that automatically re-creates the client if its `ClusterStateProvider` is closed. Unfortunately SolrClient doesn't implement `SolrCloseable`.

### DIFF
--- a/solr/cross-dc-manager/build.gradle
+++ b/solr/cross-dc-manager/build.gradle
@@ -32,6 +32,7 @@ dependencies {
   implementation libs.dropwizard.metrics.servlets
   implementation libs.eclipse.jetty.server
   implementation libs.eclipse.jetty.ee10.servlet
+  implementation libs.google.guava
   implementation libs.slf4j.api
   runtimeOnly libs.google.protobuf.javautils
   runtimeOnly libs.commonscodec.commonscodec
--- a/solr/cross-dc-manager/build.gradle
+++ b/solr/cross-dc-manager/build.gradle
@@ -32,7 +32,6 @@ dependencies {
   implementation libs.dropwizard.metrics.servlets
   implementation libs.eclipse.jetty.server
   implementation libs.eclipse.jetty.ee10.servlet
-  implementation libs.google.guava
   implementation libs.slf4j.api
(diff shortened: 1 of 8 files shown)


solr/cross-dc-manager/build.gradle
```diff
@@ -32,6 +32,7 @@ dependencies {
   implementation libs.dropwizard.metrics.servlets
   implementation libs.eclipse.jetty.server
   implementation libs.eclipse.jetty.ee10.servlet
+  implementation libs.google.guava
```
HoustonPutman: I don't think you need this anymore

solr/cross-dc-manager/src/java/org/apache/solr/crossdc/manager/consumer/KafkaCrossDcConsumer.java
```diff
@@ -95,6 +98,59 @@ public class KafkaCrossDcConsumer extends Consumer.CrossDcConsumer {
...
+          return solrClientRef.get();
+        }
+      }
+      if (existingClient.getClusterStateProvider().isClosed()) {
+        // lock out other threads and re-open the client if its ClusterStateProvider was closed
+        synchronized (solrClientRef) {
+          // refresh and check again
+          existingClient = solrClientRef.get();
+          if (existingClient.getClusterStateProvider().isClosed()) {
+            log.info("- re-creating Solr client because its CSP was closed");
```
HoustonPutman: Can we make more normal log messages without `-`s and not abbreviating things like "CSP"?

HoustonPutman: This applies to all added log lines.

HoustonPutman (commented):
Other than the logs this looks ok to me. Have y'all tested this to make sure the handoff is smooth?

Also can we add some TODO comments that mention that we should be doing this in the client and not in the consumer eventually?

PR #3412, opened 2025-06-30
SOLR-17805: Exception in TransactionLog constructor deletes the file and does not block subsequent updates.

bruno-roustant:
https://issues.apache.org/jira/browse/SOLR-17805

TransactionLog constructor can either open an existing log, or create a new log. When it is required to create a new log, it first creates a new empty log file and then opens an output stream on it.

After the new empty log file is created, if an exception occurs when creating the output stream (e.g. disk full, encryption key not available), the TransactionLog object is not created, but the empty file is left. The problem is that UpdateLog does not update its "id" field which should point to the next (non-existing) transaction log id, instead it continues to point to the empty log file. This completely blocks the logic in UpdateLog.ensureLog() and TransactionLog constructor will always refuse to create a new transaction log, throwing "New transaction log already exists" for each update, effectively blocking all subsequent updates, until the Solr server is restarted.

The proposed fix is to

1- Delete the created empty log file if an exception is thrown inside the TransactionLog constructor.
2- Fix the error code inconsistency between DirectUpdateHandler2.addDoc() (which currently sets a 4xx code) and RequestHandlerBase.normalizeReceivedException() (which currently asserts a 5xx code).

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -2561,6 +2561,8 @@ public class CoreContainer {
   }
 
   /**
+   * Checks whether a tragic exception was thrown during update (including update log).
+   *
    * @param solrCore the core against which we check if there has been a tragic exception
    * @return whether this Solr core has tragic exception
    * @see org.apache.lucene.index.IndexWriter#getTragicException()
@@ -2573,19 +2575,6 @@ public class CoreContainer {
       // failed to open an indexWriter
       tragicException = e;
     }
-
-    if (tragicException != null && isZooKeeperAware()) {
-      getZkController().giveupLeadership(solrCore.getCoreDescriptor());
-
-      try {
-        // If the error was something like a full file system disconnect, this probably won't help
-        // But if it is a transient disk failure then it's worth a try
-        solrCore.getSolrCoreState().newIndexWriter(solrCore, false); // should we rollback?
-      } catch (IOException e) {
-        log.warn("Could not roll index writer after tragedy");
-      }
-    }
-
     return tragicException != null;
   }
 
--- a/solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
+++ b/solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
@@ -246,7 +246,7 @@ public abstract class RequestHandlerBase
     } catch (QueryLimitsExceededException e) {
       rsp.setPartialResults(req);
     } catch (Exception e) {
-      Exception normalized = normalizeReceivedException(req, e);
+      Exception normalized = processReceivedException(req, e);
       processErrorMetricsOnException(normalized, metrics);
       rsp.setException(normalized);
     } finally {
@@ -304,15 +304,21 @@ public abstract class RequestHandlerBase
     }
   }
 
-  public static Exception normalizeReceivedException(SolrQueryRequest req, Exception e) {
+  /**
+   * Processes and normalizes any exceptions that are received from the request handler. This method is called
+   * before any error metrics are recorded.
+   * <p>
+   * If a tragic exception occurred in the index writer, this method also replaces the index writer with a new
+   * one to attempt to get out of a transient failure (e.g. disk failure).
+   */
+  public static Exception processReceivedException(SolrQueryRequest req, Exception e) {
     if (req.getCore() != null) {
       assert req.getCoreContainer() != null;
       if (req.getCoreContainer().checkTragicException(req.getCore())) {
         if (req.getCoreContainer().isZooKeeperAware()) {
           try {
             // If the error was something like a full file system disconnect, this probably won't
-            // help
-            // But if it is a transient disk failure then it's worth a try
+            // help, but if it is a transient disk failure then it's worth a try.
             req.getCore()
                 .getSolrCoreState()
                 .newIndexWriter(req.getCore(), false); // should we rollback?
--- a/solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
+++ b/solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
@@ -35,9 +35,11 @@ import org.apache.solr.common.params.ShardParams;
 import org.apache.solr.common.params.SolrParams;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.common.util.SuppressForbidden;
+import org.apache.solr.core.CoreContainer;
 import org.apache.solr.core.MetricsConfig;
 import org.apache.solr.core.PluginBag;
 import org.apache.solr.core.PluginInfo;
+import org.apache.solr.core.SolrCore;
 import org.apache.solr.core.SolrInfoBean;
 import org.apache.solr.metrics.SolrDelegateRegistryMetricsContext;
 import org.apache.solr.metrics.SolrMetricManager;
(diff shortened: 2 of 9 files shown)


solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -2573,19 +2575,6 @@ public boolean checkTragicException(SolrCore solrCore) {
       // failed to open an indexWriter
       tragicException = e;
     }
-
```
bruno-roustant, line 2576: This code is moved to RequestHandlerBase.normalizeReceivedException() (currently, before this PR, the only caller), because it should not be here. This method checks the "tragedy" status of the current index writer, it should not modify the state. But this code replaces the index writer, making next call to this method return always false because the "tragedy" state is reset.

solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java
```diff
@@ -336,11 +336,18 @@ public int addDoc(AddUpdateCommand cmd) throws IOException {
               + errorDetails;
       throw new SolrException(SolrException.ErrorCode.BAD_REQUEST, errorMsg, iae);
     } catch (RuntimeException t) {
+      SolrException.ErrorCode errorCode =
```
bruno-roustant, line 339: Previously the error code was always set to 4xx, whereas it is asserted to 5xx in RequestHandlerBase.wrapLuceneTragicExceptionIfNecessary(). So I propose to check here the "tragedy" state and set the error code correctly.

solr/core/src/java/org/apache/solr/update/TransactionLog.java
```diff
@@ -247,6 +248,9 @@ protected TransactionLog(
         } catch (Exception e) {
           log.error("Error closing tlog file (after error opening)", e);
         }
+        if (!openExisting) {
```
bruno-roustant, line 251: Main point of this fix: delete the transaction log file in case of exception. Otherwise the "id" field in UpdateLog will always point to this file and next TransactionLog constructor calls will always throw the "New transaction log already exists" exception, blocking all updates.

solr/core/src/java/org/apache/solr/update/UpdateLog.java
```diff
@@ -1592,10 +1595,45 @@ protected void deleteBufferLogs() {
...
+        String newLogName = String.format(Locale.ROOT, LOG_FILENAME_PATTERN, TLOG_NAME, id);
+        newLogPath = tlogDir.resolve(newLogName);
+        // We expect that the log file does not exist since id is designed to give the index of the
+        // next transaction log to create. But in very rare cases, the log files listed in the
+        // init() method may be stale here (file system delay?), and id may point to an existing
+        // file.
+        if (!Files.exists(newLogPath)) {
+          break;
+        }
+        // If the "new" log file already exists, refresh the log list and recompute id.
```
bruno-roustant, line 1616: Additional robustness mechanism: this method ensures we use either the current log, or create a *new* one. If the id field points to an existing one, don't be blocked here, just find the next id for the next non-existing log file.

solr/core/src/test/org/apache/solr/update/UpdateLogTest.java
```diff
@@ -218,6 +223,23 @@ public void testApplyPartialUpdatesWithDelete() throws Exception {
...
+    long lastLogId = scanLastLogId(tlogDir);
+    String newLogName =
+        String.format(
+            Locale.ROOT, UpdateLog.LOG_FILENAME_PATTERN, UpdateLog.TLOG_NAME, lastLogId + 1);
+    Path newLogPath = tlogDir.resolve(newLogName);
+    createLogFile(newLogPath);
+
+    // Add a doc and expect no "New transaction log already exists" error when creating the new
+    // transaction log.
+    ulogAdd(
```
bruno-roustant, line 239: Without the fix, this call throws the "New transaction log already exists" exception.

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
```diff
@@ -307,6 +308,18 @@ public static Exception normalizeReceivedException(SolrQueryRequest req, Excepti
     if (req.getCore() != null) {
       assert req.getCoreContainer() != null;
       if (req.getCoreContainer().checkTragicException(req.getCore())) {
+        if (req.getCoreContainer().isZooKeeperAware()) {
+          try {
+            // If the error was something like a full file system disconnect, this probably won't
+            // help
+            // But if it is a transient disk failure then it's worth a try
```
dsmiley: join lines.  Maybe a period or semicolon in-between sentences

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
dsmiley, line 1: normalizeReceivedException looks quite confusing to me.  Doesn't have javadocs.  And you are modifying it to rollback?  A method called normalizeBlahBlahBlah should do no such thing IMO; should be idempotent.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
dsmiley: I confess; I only understood half of that.

solr/core/src/java/org/apache/solr/update/UpdateLog.java
dsmiley, line 1616: I wonder if we should assert false instead of repeat the loop?

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
bruno-roustant: This rollback was previously done inside CoreContainer.checkTragicException() called line 310 (was the only caller of checkTragicException).
Now I need to also call checkTragicException() in DirectUpdateHandler2 to give the appropriate error code.

I don't see where to put this rollback code elsewhere. Rename this method processReceivedException? Unless we put it back in checkTragicException() with an additional boolean param to trigger it (I don't like it much)?

solr/core/src/java/org/apache/solr/update/UpdateLog.java
bruno-roustant: Yes, it's probably simpler.
I suppose there will still be some risk of race condition, but if it happens the TransactionLog constructor will throw the "new transaction log already exist" blocking only the current update. Next updates will try again to find a new non-existing file.

bruno-roustant: Actually I tried but the code is not simpler. The loops allows me to repeat the log path construction based on the newly computed id. I find the loop structure more concise. So I just changed the numAttempts threshold to 2 before aborting.

solr/core/src/java/org/apache/solr/handler/RequestHandlerBase.java
bruno-roustant: I renamed to processReceivedException () in the PR.

solr/core/src/java/org/apache/solr/update/UpdateLog.java
dsmiley: I was a bit unclear; I meant only to promptly fail in tests so that we can see if this happens in tests at all.  But never-mind I guess... it would make it impossible to test it as you did.  I suppose it doesn't matter.

bruno-roustant:
Thanks for the review David!

PR #3413, opened 2025-06-30
SOLR-17798: Integrate SDK OTLP metric exporter

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-17798

In `SolrMetricManager` add a `MetricExporter` for OTLP metrics to be pushed via `GRPC or HTTP`.

By default the OTLP exporter is turned off and Solr only exposes metrics as pull through the /admin/metrics endpoint.
The following configuration parameters can be passed as system properties or environment variables:

`System property | env variable`
============================
`solr.otlpMetricExporterEnabled | SOLR_OTLP_METRIC_EXPORTER_ENABLED` = `true/false`
`solr.otlpMetricExporterProtocol | SOLR_OTLP_METRIC_EXPORTER_PROTOCOL` = `grpc/http/none`
`solr.otlpMetricExporterInterval | SOLR_OTLP_METRIC_EXPORTER_INTERVAL` = `<ms time>`


solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
```diff
@@ -723,10 +732,14 @@ public SdkMeterProvider meterProvider(String providerName) {
...
-              // with an exporter
-              var provider = SdkMeterProvider.builder().registerMetricReader(reader).build();
-              return new MeterProviderAndReaders(provider, reader);
+              var builder =
+                  SdkMeterProvider.builder()
+                      .registerMetricReader(reader)
+                      .registerMetricReader(
+                          PeriodicMetricReader.builder(metricExporter)
+                              .setInterval(OTLP_EXPORTER_INTERVAL, TimeUnit.MILLISECONDS)
+                              .build());
```
mlbiscoc: We are registering both the prometheus reader for our metrics endpoint and the periodic reader to push metrics to the OTLP exporter.

solr/core/src/java/org/apache/solr/metrics/otel/OtlpExporterFactory.java
```diff
@@ -0,0 +1,65 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Factory class for creating OpenTelemetry OTLP metric exporters and its configuration properties.
+ *
+ * @see io.opentelemetry.exporter.otlp.http.metrics.OtlpHttpMetricExporter
+ * @see io.opentelemetry.exporter.otlp.metrics.OtlpGrpcMetricExporter
+ * @see NoopMetricExporter
+ */
+public class OtlpExporterFactory {
```
mlbiscoc: It's debatable if this should be here in Core. Should we actually move this into the Open Telemetry module for the OTLP exporter? OTLP tracing exporter is enabled with module only so maybe metrics as well.

solr/core/src/java/org/apache/solr/metrics/otel/OtlpExporterFactory.java
```diff
@@ -0,0 +1,65 @@
...
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+  public static final Boolean OTLP_EXPORTER_ENABLED =
+      Boolean.parseBoolean(EnvUtils.getProperty("solr.otlpMetricExporterEnabled", "false"));
+
+  public static final String OTLP_EXPORTER_PROTOCOL =
+      EnvUtils.getProperty("solr.otlpMetricExporterProtocol", "grpc");
+
+  public static final int OTLP_EXPORTER_INTERVAL =
+      Integer.parseInt(EnvUtils.getProperty("solr.otlpMetricExporterInterval", "60000"));
```
mlbiscoc: 60 seconds is the default according to otel spec so copied it [here](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#periodic-exporting-metricreader) but users can change the interval as they wish.

solr/core/src/java/org/apache/solr/metrics/otel/OtlpExporterFactory.java
```diff
@@ -0,0 +1,65 @@
...
+
+  public static MetricExporter getExporter() {
+    if (!OTLP_EXPORTER_ENABLED) {
+      log.info("OTLP metric exporter is disabled.");
+      return new NoopMetricExporter();
+    }
+
+    return switch (OTLP_EXPORTER_PROTOCOL) {
+      case "grpc" -> OtlpGrpcMetricExporter.getDefault();
+      case "http" -> OtlpHttpMetricExporter.getDefault();
```
mlbiscoc, line 47: Doing some testing, since we use the `getDefault` for both grpc and http, it actually reads some default env variables without us having to instrument our own read and configure for the exporter. For example, `OTEL_EXPORTER_OTLP_ENDPOINT` can change the endpoint where the metrics are pushed but I need to find all the different parameters that this picks up from documentation somewhere.

mlbiscoc (commented):
I tested this with a OTEL collector using the following collector config:
```
otel-config.yml
service:
  pipelines:
    metrics:
      receivers: [otlp]
      exporters: [prometheus]
  telemetry:
    metrics:
      address: 0.0.0.0:8888
      level: detailed

receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
      http:
        endpoint: 0.0.0.0:4318

exporters:
  prometheus:
    endpoint: 0.0.0.0:8889
    namespace: default
```
Solr pushes the metrics to the collectors receiver with grpc then you just curl the collector endpoint and the exporter gives you prometheus
```
curl 'localhost:8889/metrics'
```

I also think this feature branch is at a point where PRs for the migration to OTEL can start happening in parallel across handlers and different classes. Might start making additional PRs for migrations

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
```diff
@@ -723,10 +734,15 @@ public SdkMeterProvider meterProvider(String providerName) {
...
+              SdkMeterProviderUtil.setExemplarFilter(builder, ExemplarFilter.traceBased());
```
mlbiscoc, line 861: Adding in trace based exemplars. If you have Solr distributed trace on and push with OTLP to something like the OTEL collector, you can get exemplars on the metrics honored across the pipeline like so:
```
solr_metrics_core_requests_times_milliseconds_bucket{collection="demo",core="demo_shard1_replica_n1",handler="/update",job="unknown_service:java",replica="replica_n1",shard="shard1",le="50.0"} 26 1.751574499605e+09 # {trace_id="4438406367ea8ea60f53e3f6e26a409d",span_id="d797ab840ed14e6d"} 30.0 1.75157449662e+09
```

But this is directly pinned to your traces sampling rate.

mlbiscoc: Removed exemplar support here to not confuse the scope changes. Go to [this PR](https://github.com/apache/solr/pull/3427) instead.

mlbiscoc (commented):
@dsmiley tagging you in this as well. Natively has a OTLP metric exporter in Solr as another configurability option. Opens Solr to many different metric pipelines with the industry support for this protocol. I set it behind some feature flags to enable but not sure if this is the correct implementation. Would like feedback on what is the correct direction to do so.

solr/core/src/java/org/apache/solr/metrics/otel/OtlpExporterFactory.java
dsmiley: I agree; metrics exporting belongs in a module.  Rationale:  to constrain dependencies

solr/core/src/java/org/apache/solr/metrics/otel/OtlpExporterFactory.java
dsmiley: Use of EnvUtils should have a format of "solr.module.blahBlah", so you have a missing period after "otlp".  Although the module should be more major like "metrics"

solr/core/src/java/org/apache/solr/metrics/otel/OtlpExporterFactory.java
mlbiscoc: So I moved it into the Open Telemetry module and loaded it with `SolrResourceLoader` with `NoopMetricExporter` as the default. We just need the `opentelemetry.exporter.sender.okhttp` as a runtime dependency in core to use the OTLP exporter.

mlbiscoc:
@janhoy Would you be open to taking a look at this? You introduced the OTEL module so getting some feedback on this implementation would be helpful if I am going the right direction here.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1447,6 +1447,9 @@ public void shutdown() {
             SolrMetricManager.getRegistryName(SolrInfoBean.Group.jvm), metricTag);
         metricManager.unregisterGauges(
             SolrMetricManager.getRegistryName(SolrInfoBean.Group.jetty), metricTag);
+
+        // Close all OTEL meter providers and metrics
+        metricManager.closeAllRegistries();
```
mlbiscoc, line 1458: Added this to close all the registries on core container shutdown.

solr/core/src/java/org/apache/solr/metrics/otel/MetricExporterFactory.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+package org.apache.solr.metrics.otel;
+
+import io.opentelemetry.sdk.metrics.export.MetricExporter;
+import org.apache.solr.common.util.EnvUtils;
+
+public interface MetricExporterFactory {
+
+  public static final Boolean OTLP_EXPORTER_ENABLED =
```
dsmiley, line 24: Just an FYI -- be aware we can't test constants with different values in our unit tests.  If we wanted to try a different value, we'd have to add a "BATS" test.

solr/core/src/test/org/apache/solr/metrics/SolrMetricManagerTest.java
```diff
@@ -390,6 +394,32 @@ public void testObservableDoubleUpDownCounter() {
...
+  @Test
+  public void testMetricExporter() throws Exception {
+    LongCounter counter =
+        metricManager.longCounter(METER_PROVIDER_NAME, "my_counter", "desc", null);
+    counter.add(5);
+    counter.add(3);
+    InMemoryMetricExporter exporter = (InMemoryMetricExporter) metricManager.getMetricExporter();
+
+    // Wait 3 seconds for the exporter to push metrics
+    Thread.sleep(3000);
```
dsmiley: It's better to use RetryUtil or similar technique instead of sleep.  Sleep is tests for large codebases like Solr is rather evil.

solr/test-framework/src/java/org/apache/solr/SolrTestCaseJ4.java
```diff
@@ -299,6 +298,7 @@ public static void setupTestCases() {
     }
 
     ExecutorUtil.resetThreadLocalProviders();
+    OpenTelemetryConfigurator.resetForTest();
```
dsmiley, line 302: did beginning vs end matter?

mlbiscoc: TBH I moved it so long ago I don't remember anymore. It passed both ways but I'll just keep the change in case there was a good reason...

dsmiley (approved):

PR #3414, opened 2025-06-30
SOLR-17824: RecoveryStrategy.pingLeader could NPE when there's no leader.

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17824

Should be continuing its retry loop.

I found this merely by looking at a test failure, briefly looked at the stack trace, and threw this fix together quickly:
https://github.com/apache/solr/actions/runs/15980088952/job/45073441889?pr=3409

Need a JIRA; maybe there is one.


dsmiley:
I'll merge this in a couple days.  CHANGES.txt will be pretty clear.
I didn't do any deeper insights on what the ramification of this issue is, how often it might happen, etc.  I merely saw an NPE with a straightforward fix.

PR #3416, opened 2025-07-01
[SOLR-17799] Replace segment merge metrics using gauges with consistent counters

liangkaiwen:

# Description

Replace existing segment merge metrics with consistent counters

# Solution

* Removed the index merge running gauge metrics
* Added the following index merge metrics:

INDEX.merge.major.started.docs
INDEX.merge.major.started.deletedDocs
INDEX.merge.major.started.segments
INDEX.merge.minor.started.docs
INDEX.merge.minor.started.deletedDocs
INDEX.merge.minor.started.segments

INDEX.merge.major.completed.docs
INDEX.merge.major.completed.deletedDocs
INDEX.merge.major.completed.segments
INDEX.merge.minor.completed.docs
INDEX.merge.minor.completed.deletedDocs
INDEX.merge.minor.completed.segments

# Tests

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/update/SolrIndexWriter.java
+++ b/solr/core/src/java/org/apache/solr/update/SolrIndexWriter.java
@@ -25,7 +25,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import org.apache.lucene.codecs.Codec;
 import org.apache.lucene.index.IndexDeletionPolicy;
@@ -72,23 +71,33 @@ public class SolrIndexWriter extends IndexWriter {
   private Directory directory;
 
   // metrics
-  private long majorMergeDocs = 512 * 1024;
-  private Timer majorMerge;
-  private Timer minorMerge;
-  private Meter majorMergedDocs;
-  private Meter majorDeletedDocs;
-  private Counter mergeErrors;
-  private Meter flushMeter; // original counter is package-private in IndexWriter
-  private boolean mergeTotals = false;
-  private boolean mergeDetails = false;
-  private final AtomicInteger runningMajorMerges = new AtomicInteger();
-  private final AtomicInteger runningMinorMerges = new AtomicInteger();
-  private final AtomicInteger runningMajorMergesSegments = new AtomicInteger();
-  private final AtomicInteger runningMinorMergesSegments = new AtomicInteger();
-  private final AtomicLong runningMajorMergesDocs = new AtomicLong();
-  private final AtomicLong runningMinorMergesDocs = new AtomicLong();
+  private static final String MERGE_METRIC_PATH_COMPONENT = "merge";
+  private static final String MERGE_METRIC_PATH_COMPONENT_MAJOR = "major";
+  private static final String MERGE_METRIC_PATH_COMPONENT_MINOR = "minor";
+  private static final String MERGE_METRIC_PATH_COMPONENT_COMPLETED = "completed";
+  private static final String MERGE_METRIC_PATH_COMPONENT_START = "start";
 
-  private final SolrMetricsContext solrMetricsContext;
+  private static final String MERGE_DOCS_METRIC_NAME = "docs";
+  private static final String MERGE_DELETED_DOCS_METRIC_NAME = "deletedDocs";
+  private static final String MERGE_SEGMENTS_METRIC_NAME = "segments";
+  private static final String MERGE_TIMES_METRIC_NAME = "mergeTimes";
+  private static final String MERGE_FLUSH_METRIC_NAME = "flush";
+  private static final String MERGE_ERRORS_METRIC_NAME = "errors";
+
+  private static final Map<String, Counter> majorMergeStartedMetrics = new HashMap<>();
+  private static final Map<String, Counter> minorMergeStartedMetrics = new HashMap<>();
+  private static final Map<String, Counter> majorMergeCompletedMetrics = new HashMap<>();
+  private static final Map<String, Counter> minorMergeCompletedMetrics = new HashMap<>();
+
+  private long majorMergeDocs = 512 * 1024;
+  private Meter flushMeter; // original counter is package-private in IndexWriter
+  private Meter majorMergeMeter;
+  private Meter minorMergeMeter;
+  private Timer majorMergeTimer;
+  private Timer minorMergeTimer;
+  private Counter mergeErrorsCounter;
+
+  private SolrMetricsContext solrMetricsContext;
   // merge diagnostics.
   private final Map<String, Long> runningMerges = new ConcurrentHashMap<>();
 
@@ -126,8 +135,6 @@ public class SolrIndexWriter extends IndexWriter {
     numOpens.incrementAndGet();
     log.debug("Opened Writer {}", name);
     // no metrics
-    mergeTotals = false;
-    mergeDetails = false;
     solrMetricsContext = null;
   }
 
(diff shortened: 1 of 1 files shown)


liangkaiwen:
Will update documentation and add tests after we confirm this is the pattern of metrics/naming that we want to use

dsmiley:
+1
Nice!
I slightly prefer "finished" instead of "completed" as I think it better balances with "started", plus it's slightly shorter.

solr/core/src/java/org/apache/solr/update/SolrIndexWriter.java
```diff
@@ -72,23 +71,33 @@ public class SolrIndexWriter extends IndexWriter {
...
+
+  private static final String MERGES_METRIC_NAME = "merges";
+  private static final String MERGE_DOCS_METRIC_NAME = "docs";
+  private static final String MERGE_DELETED_DOCS_METRIC_NAME = "deletedDocs";
+  private static final String MERGE_SEGMENTS_METRIC_NAME = "segments";
+  private static final String MERGE_TIMES_METRIC_NAME = "mergeTimes";
+  private static final String MERGE_FLUSH_METRIC_NAME = "flushes";
+  private static final String MERGE_ERRORS_METRIC_NAME = "errors";
+
+  private static final Map<String, Counter> majorMergeStartedMetrics = new HashMap<>();
```
psalagnac: I'm confused by these static (global) maps. Shouldn't we have per core metrics?

liangkaiwen: you're correct, will make this a field

solr/core/src/java/org/apache/solr/update/SolrIndexWriter.java
mlbiscoc, line 1: A bit surprised this isn't implementer of `SolrMetricProducer` and all metrics are initialized in the constructor. Maybe the `SolrMetricProducer` was create later on and this never moved. I think we should move this to implement that interface which basically does your `initMetrics()`.

solr/core/src/java/org/apache/solr/update/SolrIndexWriter.java
```diff
@@ -72,23 +71,33 @@ public class SolrIndexWriter extends IndexWriter {
...
+  private static final String MERGE_DELETED_DOCS_METRIC_NAME = "deletedDocs";
+  private static final String MERGE_SEGMENTS_METRIC_NAME = "segments";
+  private static final String MERGE_TIMES_METRIC_NAME = "mergeTimes";
+  private static final String MERGE_FLUSH_METRIC_NAME = "flushes";
+  private static final String MERGE_ERRORS_METRIC_NAME = "errors";
+
+  private final Map<String, Counter> majorMergeStartedMetrics = new HashMap<>();
+  private final Map<String, Counter> minorMergeStartedMetrics = new HashMap<>();
+  private final Map<String, Counter> majorMergeFinishedMetrics = new HashMap<>();
+  private final Map<String, Counter> minorMergeFinishedMetrics = new HashMap<>();
```
mlbiscoc, line 91: I see you are updating these with `updateMergeMetrics`. But this isn't thread safe as these can be modified concurrently from multiple threads from a merge. This should probably be a `ConcurrentHashMap`?

solr/core/src/java/org/apache/solr/update/SolrIndexWriter.java
```diff
@@ -72,23 +71,33 @@ public class SolrIndexWriter extends IndexWriter {
...
+  private static final String MERGE_METRIC_PATH_COMPONENT_FINISHED = "finished";
+  private static final String MERGE_METRIC_PATH_COMPONENT_STARTED = "started";
+
+  private static final String MERGES_METRIC_NAME = "merges";
+  private static final String MERGE_DOCS_METRIC_NAME = "docs";
+  private static final String MERGE_DELETED_DOCS_METRIC_NAME = "deletedDocs";
+  private static final String MERGE_SEGMENTS_METRIC_NAME = "segments";
+  private static final String MERGE_TIMES_METRIC_NAME = "mergeTimes";
+  private static final String MERGE_FLUSH_METRIC_NAME = "flushes";
+  private static final String MERGE_ERRORS_METRIC_NAME = "errors";
```
mlbiscoc, line 86: I'm usually a fan of constant static strings but in this case there are so many, I think I preferred it the other way of just in-lining them as I found it easier to read that it went something like
```
solrMetricsContext.gauge(
            () -> runningMajorMerges.get(),
            true,
            "running",
            SolrInfoBean.Category.INDEX.toString(),
            "merge",
            "major");
```
`running.index.merge.major`. I don't feel strongly about it though so if you prefer this way thats fine I guess.

solr/core/src/java/org/apache/solr/update/SolrIndexWriter.java
```diff
@@ -269,54 +204,22 @@ protected void merge(MergePolicy.OneMerge merge) throws IOException {
...
-      if (major) {
-        runningMajorMerges.decrementAndGet();
-        runningMajorMergesDocs.addAndGet(-totalNumDocs);
-        runningMajorMergesSegments.addAndGet(-segmentsCount);
-      } else {
-        runningMinorMerges.decrementAndGet();
-        runningMinorMergesDocs.addAndGet(-totalNumDocs);
-        runningMinorMergesSegments.addAndGet(-segmentsCount);
-      }
+      updateMergeMetrics(totalNumDocs, deletedDocs, segmentsCount, true, context);
```
mlbiscoc, line 222: Should this be negative `totalNumDocs` and `-segmentsCount`?

solr/core/src/test/org/apache/solr/update/SolrIndexWriterTest.java
```diff
@@ -0,0 +1,171 @@
...
+    for (int i = 1; i <= 3; i++) {
+      iw.deleteDocuments(new Term("id", String.valueOf(i)));
+    }
+    iw.commit();
+    iw.forceMerge(1, true);
+
+    assertEquals(
+        "should be mulitple merge operations",
+        initialMerges + 1,
+        minorMergeStartedMetrics.getCount());
```
mlbiscoc, line 128: Isn't this technically just 1 merge from line 123? Why would initial be higher than 0? Is it because minor merges on itself with less docs?

mlbiscoc (commented):
Its a bit hard to review this without seeing the output changes in parallel. Could you post the before and after?

I have some concern for backwards compatibility. I am a fan of this technically being on by default but I think it is too late for that. We should probably keep it optional as before, otherwise we run the risk of blowing up cardinality for clouds with many cores without them knowing the moment they call the /admin/metrics endpoint without filtering.

Also as this is probably going for 9.x, we would probably need to keep the old metrics there and its format for backwards compat reasons as well. Just add the news ones in parallel.

dsmiley:
Just want to react to something:

>  otherwise we run the risk of blowing up cardinality for clouds with many cores without them knowing the moment they call the /admin/metrics endpoint without filtering.

This is already a risk.  Any real-world use of fetching /admin/metrics is going to have a filtering parameter.  It'd be reasonable to detect that no filter was given and then potentially return an error.  If the user is just kicking the tires, they may want to ignore the filter requirement so the error message might contain a trivial suggestion like to see all Solr metrics.

liangkaiwen:
Re: backwards compatibility - this is a good point, but I don't think it would be worth to push both these new metrics and the existing ones (would be confusing having so many metrics of similar names, different types, and overlapping behaviors).

I think what I will do is shelve this implementation and instead target to have an OTEL implementation of these metrics for Solr 10 release

solr/core/src/java/org/apache/solr/update/SolrIndexWriter.java
liangkaiwen: I don't know about "modified". This implementation only uses this map for get operations and never put (besides initial construction). I think with only get operations, a concurrenthashmap doesn't provide much utility no?

solr/core/src/java/org/apache/solr/update/SolrIndexWriter.java
liangkaiwen: No, it shouldn't be negative because this implementation no longer tracks the state of a current merge. The previous implementation uses negatives because the merge has finished therefore there are "no documents being currently merged"

solr/core/src/test/org/apache/solr/update/SolrIndexWriterTest.java
liangkaiwen: forgot to update the assertion message. You're right, should be single

dsmiley:
Closing, as the work continued in #3682

PR #3420, opened 2025-07-08
SOLR-17779: Remove 'roles' and 'numShards' core properties

psalagnac:
https://issues.apache.org/jira/browse/SOLR-17779

This removes old core properties that are still stored in `core.properties` files but never read.
Also cleanup the SolrJ part.


epugh (approved):
LGTM.  Nice to see the clean up!

PR #3423, opened 2025-07-09
Improve usages of GenericSolrRequest (request type, http method, requiresCollection)

dsmiley:
and deprecate GenericSolrRequest.setRequiresCollection

[SOLR-17829](https://issues.apache.org/jira/browse/SOLR-17829)

Trying to emphasize the collection-ness in the constructor, and make other aspects mandatory (request type).

In a separate PR but debatably same JIRA, I'll change many usages of QueryRequest that should not be using such; should be generic.  I have that shelved, waitng for this.  That's how I got started down this path, actually.


solr/core/src/java/org/apache/solr/cli/ExportTool.java
```diff
@@ -253,11 +253,12 @@ void fetchUniqueKey() throws SolrServerException, IOException {
               .build();
       NamedList<Object> response =
           solrClient.request(
-              new GenericSolrRequest(
-                      SolrRequest.METHOD.GET,
-                      "/schema/uniquekey",
-                      SolrParams.of("collection", coll))
```
dsmiley, line 267: I don't love that collection param; moved to the request path

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
```diff
@@ -681,8 +680,8 @@ private UpdateCommand fetchFullUpdateFromLeader(AddUpdateCommand inplaceAdd, lon
     params.set(DISTRIB, false);
     params.set("getInputDocument", id);
     params.set("onlyIfActive", true);
-    SolrRequest<SimpleSolrResponse> ur =
-        new GenericSolrRequest(METHOD.GET, "/get", params).setRequiresCollection(true);
+    var ur =
+        new GenericCollectionRequest(METHOD.GET, "/get", SolrRequest.SolrRequestType.ADMIN, params);
```
dsmiley: making the request type mandatory; force you to think about it.  RTG is used for internal purposes; maybe QUERY would get throttled?  Or maybe not; I recall the rate limiter would ignore internal requests.  LBSolrClient should retry a QUERY but not ADMIN; maybe this should be QUERY.

solr/core/src/test/org/apache/solr/cloud/MultiThreadedOCPTest.java
dsmiley, line 1: shouldn't be in this PR; I'll remove

solr/core/src/test/org/apache/solr/search/TestTaskManagement.java
```diff
@@ -185,8 +186,11 @@ private NamedList<String> listTasks() throws SolrServerException, IOException {
         cluster
             .getSolrClient(COLLECTION_NAME)
             .request(
-                new GenericSolrRequest(SolrRequest.METHOD.GET, "/tasks/list")
-                    .setRequiresCollection(true));
+                new GenericCollectionRequest(
+                    SolrRequest.METHOD.GET,
+                    "/tasks/list",
```
dsmiley: it was a little eye-opening to see we have a handler that doesn't "look" like an admin handler but is one

solr/core/src/test/org/apache/solr/util/TestCborDataFormat.java
```diff
@@ -145,51 +145,57 @@ private byte[] runQuery(String testCollection, CloudSolrClient client, String wt
...
                         + "{\"name\":\"initial_release_date\",\"type\":\"string\",\"stored\":true},\n"
                         + "{\"name\":\"directed_by\",\"type\":\"string\",\"multiValued\":true,\"stored\":true},\n"
                         + "{\"name\":\"genre\",\"type\":\"string\",\"multiValued\":true,\"stored\":true},\n"
-                        + "{\"name\":\"film_vector\",\"type\":\"knn_vector_10\",\"indexed\":true,\"stored\":true}]}",
-                    XMLResponseWriter.CONTENT_TYPE_XML_UTF8));
-
-    client.request(req, testCollection);
+                        + "{\"name\":\"film_vector\",\"type\":\"knn_vector_10\",\"indexed\":true,\"stored\":true}]}")
+                    .getBytes(StandardCharsets.UTF_8),
+                XMLResponseWriter.CONTENT_TYPE_XML_UTF8),
```
dsmiley: wrong; isn't XML

solr/solrj/src/java/org/apache/solr/client/solrj/request/GenericSolrRequest.java
```diff
@@ -85,7 +85,9 @@ public GenericSolrRequest(METHOD m, String path, SolrRequestType requestType, So
    * 'false' otherwise.
    *
    * @param requiresCollection true if a default collection should be used, false otherwise.
+   * @deprecated use {@link GenericCollectionRequest} instead
    */
+  @Deprecated
```
dsmiley: perhaps controversial as you added this @gerlowskija but the opinionated stance I'm taking is to build this into the type, not be an optional method to call or forget to call

solr/solrj/src/java/org/apache/solr/client/solrj/request/GenericCollectionRequest.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Objects;
+import org.apache.solr.client.solrj.response.SimpleSolrResponse;
+import org.apache.solr.common.params.SolrParams;
+import org.apache.solr.common.util.NamedList;
+
+/**
+ * A generic request for sending to a Solr collection or core.
```
epugh: Seems a little sad that we still have the "collection or core" concept.   I liked reading over the code, and the explicitness of the collection name being required.   I also liked that the explicit type of request with the ADMIN was being added.

solr/solrj/src/java/org/apache/solr/client/solrj/request/GenericSolrRequest.java
epugh: Aren't we attempting over time to get away fro the use of GenericSolrRequest anyway?

dsmiley: Maybe but that's a long way out.

dsmiley:
Not sure if this PR should refrain from changing RequestType or POST/GET for that matter.

dsmiley:
As long as the changes to use this are not backported to 9x, there's no back-compat concern.  I don't plan to revert any changes here then.

solr/core/src/java/org/apache/solr/update/processor/DistributedUpdateProcessor.java
dsmiley: tldr; this is fine.

Right; wouldn't get rate limited because is internal ([here's the logic](https://github.com/apache/solr/blob/7279b808628a9286def7aea8b3b00f626df23dec/solr/core/src/java/org/apache/solr/servlet/RateLimitManager.java#L102)).  And what I said about LBSolrClient is somewhat/conditionally true ([logic](https://github.com/apache/solr/blob/e08d7f0f0e66cdde389eb843db341f0eb5d833bb/solr/solrj/src/java/org/apache/solr/client/solrj/impl/LBSolrClient.java#L457)); this admin request would still be retry-able because it's a core admin request instead of node level.  Gosh I don't like the logic there; it ought to be super simple/general.

dsmiley (commented):
I propose a JIRA with summary "Introduce GenericCollectionRequest".  Could do partial backport that doesn't touch the non-test usages changed here, as I don't really want to change 9x unnecessarily.
WDYT @gerlowskija ?  If I don't hear back, I'll do exactly this and merge; perhaps this weekend.

dsmiley:
@gerlowskija I'm looking at org.apache.solr.client.solrj.SolrRequest#requiresCollection and thinking it shouldn't have a default implementation.  As the base implementation, I don't think it can assume any value here.

solr/core/src/java/org/apache/solr/cli/ExportTool.java
gerlowskija: +1, I think we've talked elsewhere about moving away from these `collection` params across the board for APIs that also put the coll/core in the path.

solr/core/src/test/org/apache/solr/search/TestTaskManagement.java
gerlowskija: What in your mind makes something "look" like an admin handler?

That it should have "admin" in the path?  Something else?  (Not objecting or disagreeing, just curious for more specifics here)

solr/solrj/src/java/org/apache/solr/client/solrj/request/GenericCollectionRequest.java
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @see GenericSolrRequest
+ */
+public class GenericCollectionRequest extends CollectionRequiringSolrRequest<SimpleSolrResponse> {
```
gerlowskija: [0] Conceptually, this seems like a subset or a special case of the larger "Generic Solr Request" umbrella, so it feels a little unintuitive to me that there's no inheritance or formal relationship between this class and `GenericSolrRequest` and/or `GenericV2SolrRequest`.

Did you consider using one of those as a parent-class, before discarding that approach?  Or was `CollectionRequiringSolrRequest` your first instinct inheritance-wise?

solr/solrj/src/java/org/apache/solr/client/solrj/request/GenericSolrRequest.java
gerlowskija: I don't mind this method going away, per se.

My only concern here is that `GenericCollectionRequest` (as-written) assumes v1.  So I'm not sure how a user would make a collection-requiring v2 API call without the use of this method.

This ties into the larger concern I posted at the PR-level: these axes (v1/v2, collection-requiring or not) cause a combinatorial problem that gets really unwieldy if we try to handle it without relying on setters or ctor params or something similar to flatten some of the dimensions.

solr/core/src/test/org/apache/solr/search/TestTaskManagement.java
dsmiley: Yes, no "/admin".

dsmiley:
I hear your concern.  Personally it doesn't seem like things are out of hand because the settings we're talking about are binary (requiresCollection, V2).  I *really* think V2 most of all needs to be separate classes (and they are).  But requiresCollection could be a mandatory constructor parameter to GenericSolrRequest.  Maybe even an enum to make the call-site clear as to what is being toggled.

gerlowskija:
> Personally it doesn't seem like things are out of hand because the settings we're talking about are binary (requiresCollection, V2). I really think V2 most of all needs to be separate classes (and they are). But requiresCollection could be a mandatory constructor parameter to GenericSolrRequest.

Yeah, I'm not saying they're out of hand currently.  Just leery of starting a pattern that might get out of hand down the road...

So, if I understand that right, we would keep the same two classes we have currently: `GenericSolrRequest` and `GenericV2SolrRequest`.  The collection-or-not axis would be handled by removing the `setRequiresCollection` setter, and replacing it with a ctor parameter which forces callers to at least consider the question?

Assuming I've got that right, that sounds reasonable to me 👍

dsmiley:
> 2. You can create a ctor-param for each dimension maybe. This starts to make the ctor's a little unwieldy to call

Right.  As I look at GenericSolrCollection, I've completely lost motivation to follow through with this.  So I don't know how to make both of us happy now.  I'll repeat that I think the collection or node level is a significant enough matter that it's worthy of being in the class hierarchy.  I wouldn't say that for the other booleans other than V2.  I guess I'll back out the class I added and live with GenericSolrCollection.  I don't like that it's easy to overlook the request type.  Maybe the constructors that don't take that should be deprecated; shrug.

gerlowskija:
> So I don't know how to make both of us happy now.

I kindof thought we'd found a compromise approach in what you suggested [here](https://github.com/apache/solr/pull/3423#issuecomment-3130084660): "I really think V2 most of all needs to be separate classes (and they are). But requiresCollection could be a mandatory constructor parameter to GenericSolrRequest. "

I'm 👍 to that, and since you suggested it I assumed you were as well?  Has your thought changed since then, so that you're uninterested in any path forward that doesn't represent "collection-or-not" in the SolrRequest class hierarchy?

gerlowskija:
Oops, forgot to tag in previous comment, so doing so now: @dsmiley

dsmiley:
I toyed with another constructor and with an enum of 2 options... and it just wasn't worth it IMO.  At least isn't if there are all these other constructors one might call.  Might as well use setRequiresCollection(true).  IMO it's still sad it's forgettable to call it but I want to move on.  I need to close the JIRA issue, remove it from the PR title, and I'm not sure what to say for this PR other than basically "Use better request types and HTTP methods".

I've got another shelved change soon to become a PR that was going to use GenericCollectionRequest but I'll update it to use GenericSolrRequest with setRequiresCollection.  That one is more interesting, changing many wrong usages of QueryRequest for non-queries (non-SearchHandler).  They were using "qt", which I'm trying to reduce/remove.

solr/CHANGES.txt
```diff
@@ -260,7 +260,8 @@ Bug Fixes
 
 Dependency Upgrades
 ---------------------
-(No changes)
+* SOLR-17829: SolrJ introducing GenericCollectionRequest for generic collection/core specific requests, and that which
```
gerlowskija: [-1] This CHANGES.txt entry is out of date now, afaict?

gerlowskija:
> I toyed with another constructor and with an enum of 2 options... and it just wasn't worth it IMO. At least isn't if there are all these other constructors one might call.

Hmm, I kindof though we'd be adding the arg to _all_ the ctors so users would be **forced** to set it.  But if you tried it and it didn't seem worthwhile, I trust that.  Bummer.  I was really optimistic about the growing consensus around that approach, particularly after Hoss [suggested it here](https://issues.apache.org/jira/browse/SOLR-17866?focusedCommentId=18015218&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-18015218)

dsmiley:
Maybe a builder would be more elegant.

PR #3424, opened 2025-07-09
SOLR-17804: re-create SolrClient if closed (cherry-pick to 9x)

sigram:
https://issues.apache.org/jira/browse/SOLR-17804. Cherry-pick from PR 3411.


sigram:
Hmm, this is strange, both this PR and 3425 are failing with the same OOM error in `compileJava` (local builds work fine). Any suggestions @HoustonPutman ?

HoustonPutman:
If it works locally for you, just merge it. We don't generally create PRs for back ports anyways.

PR #3426, opened 2025-07-09
removing check which brakes update with new dynamic field

renatoh:
Using TextToVectorUpdateProcessorFactory with dynamics fields, the removed check causes an exception if no document with the input-field has been processes yet. The schema does not know the field yet.

AlsTextToVectorUpdateProcessor#isNullOrEmpty is already checking for null and if the input field does not exist it will be skipped anyway, hence I think it is safe to remove the check.

The same could happen on the with the output field check , but it might be more likely, that the output field is a field a not a dymaicField

### DIFF
--- a/solr/modules/llm/src/java/org/apache/solr/llm/textvectorisation/update/processor/TextToVectorUpdateProcessorFactory.java
+++ b/solr/modules/llm/src/java/org/apache/solr/llm/textvectorisation/update/processor/TextToVectorUpdateProcessorFactory.java
@@ -70,10 +70,6 @@ public class TextToVectorUpdateProcessorFactory extends UpdateRequestProcessorFa
   public UpdateRequestProcessor getInstance(
       SolrQueryRequest req, SolrQueryResponse rsp, UpdateRequestProcessor next) {
     IndexSchema latestSchema = req.getCore().getLatestSchema();
-    if (!latestSchema.hasExplicitField(inputField)) {
-      throw new SolrException(
-          SolrException.ErrorCode.SERVER_ERROR, "undefined field: \"" + inputField + "\"");
-    }
     if (!latestSchema.hasExplicitField(outputField)) {
       throw new SolrException(
           SolrException.ErrorCode.SERVER_ERROR, "undefined field: \"" + outputField + "\"");


alessandrobenedetti:
hi @renatoh , can you add a test that covers the scenario? I think your proposed change makes sense, but with the additional test coverage we'll be more confident

solr/modules/llm/src/test/org/apache/solr/llm/textvectorisation/update/processor/TextToVectorUpdateProcessorTest.java
```diff
@@ -40,6 +41,13 @@ public static void cleanup() throws Exception {
     afterTest();
   }
 
+  @After
```
renatoh, line 44: the deletion within the test method will not happen if a test fails, such a failing test was then impacting all following test since the deletion was not done and  the setup is done only once for the test-class

solr/modules/llm/src/java/org/apache/solr/llm/textvectorisation/model/SolrTextToVectorModel.java
```diff
@@ -80,41 +84,40 @@ public static SolrTextToVectorModel getInstance(
...
-                    .getClass()
-                    .getMethod(paramName, String.class)
-                    .invoke(builder, params.get(paramName).toString());
+                try {
+                  builder
+                      .getClass()
+                      .getMethod(paramName, String.class)
+                      .invoke(builder, params.get(paramName).toString());
+                } catch (NoSuchMethodException e) {
+                  log.error("Parameter {} not supported by model {}", paramName, className);
```
renatoh, line 117: Implementing my own DimensionAwareEmbeddingModel, I notice that if the attribute is not supported by the model, the error was not logged at all and I had to debug to see what the issue was.
I also took the liberty tot clean up the duplicate code snippets.

renatoh:
Hi @alessandrobenedetti, 

Thanks for your feedback! One of your test was actually failing, I adjusted it so that it no longer check for the exception.
The case where we the input fields it not present or empty is already covered well by TextToVectorUpdateProcessorTest.
Please also have a look at my comments on my changes

solr/modules/llm/src/java/org/apache/solr/llm/textvectorisation/model/SolrTextToVectorModel.java
```diff
@@ -80,41 +84,40 @@ public static SolrTextToVectorModel getInstance(
...
+                  throw e;
```
alessandrobenedetti: Maybe better here to throw a SolrException wrapping it?

solr/modules/llm/src/test/org/apache/solr/llm/textvectorisation/update/processor/TextToVectorUpdateProcessorTest.java
alessandrobenedetti: It makes sense, have you checked that all is fine? I remember I struggled a bit with the after/afterClass

solr/modules/llm/src/test/org/apache/solr/llm/textvectorisation/update/processor/TextToVectorUpdateProcessorFactoryTest.java
```diff
@@ -132,7 +132,7 @@ public void init_notDenseVectorOutputField_shouldThrowExceptionWithDetailedMessa
 
   /* Following test depends on a real solr schema and depends on BeforeClass-AfterClass methods */
   @Test
-  public void init_notExistentInputField_shouldThrowExceptionWithDetailedMessage() {
+  public void init_notExistentInputField_shouldNotThrowException() {
```
alessandrobenedetti: mmmm I think this was ok, if the field is undefined I expect the exception.
You should keep this test and add a new one specific for a dynamic field (the scope of this issue).
Did I misunderstand anything?

alessandrobenedetti (changes requested):
I think we are nearly there, the contributions make sense but the tests should get some additional attention

solr/modules/llm/src/test/org/apache/solr/llm/textvectorisation/update/processor/TextToVectorUpdateProcessorTest.java
renatoh: yes, now with the @After, even if I make one test fail on purpose, the result is consistent across 5 test runs. Without the @After one failing test has unwanted side effect and other test of the class are randomly failing due to the data which got not cleaned up.

solr/modules/llm/src/java/org/apache/solr/llm/textvectorisation/model/SolrTextToVectorModel.java
renatoh: make sense, will add a wrapping SolrException

solr/modules/llm/src/test/org/apache/solr/llm/textvectorisation/update/processor/TextToVectorUpdateProcessorFactoryTest.java
renatoh: But this test will fail with the changes I did to TextToVectorUpdateProcessorFactory.java, since the exception is not thrown anymore. If we want to keep this test we need to distinguish somehow between a dynamic field and a non-dynamic field within TextToVectorUpdateProcessorFactory.java, in order to throw the exception only for dynamic fields, not sure if this would make sense

alessandrobenedetti: mmm, so what would happen if the user specifies in the input a non-existent field? For example, because of a typo?
Will it just silently fail?
If that's the case, we need to differentiate between non-existent and 'non explicitly defined but compatible with a dynamic field' I suspect

renatoh: ok, I am checking now it the input or output field is a dynamic field and the exception is now still throw if the non-dynamic-field does not exists

alessandrobenedetti (approved):
The changes look good to me, I'll wait for a green build and add the changes.txt, merging afterwards. Thanks Renato!

alessandrobenedetti:
Hi Renato, wanted to merge, but a bit in a rush now, Can you solve the conflict and push it again? (the annoying CHANGES.txt).

I'll merge after lunch, so in case you can't do it in the meantime, I'll do it later.

Cheers

renatoh:
> Hi Renato, wanted to merge, but a bit in a rush now, Can you solve the conflict and push it again? (the annoying CHANGES.txt).
> 
> I'll merge after lunch, so in case you can't do it in the meantime, I'll do it later.
> 
> Cheers

Hi Alessandro, only back at my computer on friday. 
Ah, always the annoying CHANGES. txt

PR #3427, opened 2025-07-15
SOLR-17818: Expose Open Metrics 1.0 format from /admin/metrics and exemplar support

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-17818

Adding in metrics with exemplars from distributed trace. If you have Solr distributed trace on and push with OTLP to something like the OTEL collector or pull directly from the `/admin/metrics` endpoint, you can get exemplars on the metrics honored across the pipeline like so:
```
solr_metrics_core_requests_times_milliseconds_bucket{collection="demo",core="demo_shard1_replica_n1",handler="/update",job="unknown_service:java",replica="replica_n1",shard="shard1",le="50.0"} 26 1.751574499605e+09 # {trace_id="4438406367ea8ea60f53e3f6e26a409d",span_id="d797ab840ed14e6d"} 30.0 1.75157449662e+09
```

But this is directly pinned to your traces sampling rate.

Users need to specify the format with a `format=openmetrics` from the endpoint otherwise it will default to Prometheus 0.0.4 exposition format.

### DIFF
--- a/solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
+++ b/solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
@@ -52,6 +52,8 @@ import io.opentelemetry.api.metrics.ObservableLongMeasurement;
 import io.opentelemetry.api.metrics.ObservableLongUpDownCounter;
 import io.opentelemetry.exporter.prometheus.PrometheusMetricReader;
 import io.opentelemetry.sdk.metrics.SdkMeterProvider;
+import io.opentelemetry.sdk.metrics.internal.SdkMeterProviderUtil;
+import io.opentelemetry.sdk.metrics.internal.exemplar.ExemplarFilter;
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
@@ -725,8 +727,9 @@ public class SolrMetricManager {
               var reader = new PrometheusMetricReader(true, null);
               // NOCOMMIT: We need to add a Periodic Metric Reader here if we want to push with OTLP
               // with an exporter
-              var provider = SdkMeterProvider.builder().registerMetricReader(reader).build();
-              return new MeterProviderAndReaders(provider, reader);
+              var provider = SdkMeterProvider.builder().registerMetricReader(reader);
+              SdkMeterProviderUtil.setExemplarFilter(provider, ExemplarFilter.traceBased());
+              return new MeterProviderAndReaders(provider.build(), reader);
             })
         .sdkMeterProvider();
   }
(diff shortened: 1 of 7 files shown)


mlbiscoc (commented):
@dsmiley Could you take a look at this? I'd like to push for exemplar support natively exposed from Solr's /admin/metrics endpoint with the Open Metrics format as an option. Since otel is already instrumented for distributed trace in Solr, getting exemplars is simple in this PR. Also exemplars are kept even if they use OTLP. I think it's one of the more powerful features otel is supporting from prometheus.

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
```diff
@@ -725,8 +727,9 @@ public SdkMeterProvider meterProvider(String providerName) {
               var reader = new PrometheusMetricReader(true, null);
               // NOCOMMIT: We need to add a Periodic Metric Reader here if we want to push with OTLP
               // with an exporter
-              var provider = SdkMeterProvider.builder().registerMetricReader(reader).build();
-              return new MeterProviderAndReaders(provider, reader);
+              var provider = SdkMeterProvider.builder().registerMetricReader(reader);
```
dsmiley, line 728: not a comment on this line but the javadoc of meterProvider() should not be speaking of MeterProviderAndReaders as that's very internal to the class.  Javadocs are for the callers of the method.

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
```diff
@@ -55,12 +59,19 @@ public void write(
     List<MetricSnapshot> snapshots =
         readers.values().stream().flatMap(r -> r.collect().stream()).toList();
 
-    new PrometheusTextFormatWriter(false).write(out, mergeSnapshots(snapshots));
+    boolean openMetricsFormat = "openmetrics".equals(request.getParams().get("format"));
```
dsmiley: Solr has historically used request parameters to offer clients a means of requesting different formats.  While it's convenient for hacking/playing, it's not using HTTP standards; the "Accept" header is the way to go.  That said, this doesn't mean we can't have a parameter as well.  If we did, wouldn't we use our existing param -- "wt" ?  CC @gerlowski for confirmation

dsmiley (commented):
Wow there was so little to this.  I suppose this is a benefit of embracing OTEL.

solr/core/src/java/org/apache/solr/response/PrometheusResponseWriter.java
mlbiscoc: Maybe we do both then. I think I saw in the 9.9 change log that v2 apis use `Accept` as header request if wt isn't passed in. We could follow the same logic.

mlbiscoc: Added a way for the `Accept` header and added tests for Open metrics and exemplars.

PR #3429, opened 2025-07-15
SOLR-17792: Fix error handling for shard requests

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17792

https://github.com/apache/solr/pull/3398 accidentally broke the error handling in HttpShardHandler, by accepting already received requests after cancelling. So instead of throwing those responses away, the HttpShardHandler will find a response with an error after cancelling, since that is the contract that `take()` has. And even then, for tolerant requests, the `SearchHandler` needs to find any responses that have errors to know if the results are partial or not. (because tolerant requests don't guarantee that the response returned is the one with the error)

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/component/HttpShardHandler.java
@@ -343,7 +343,7 @@ public class HttpShardHandler extends ShardHandler {
             // Otherwise return the last response that we processed.
             return previousResponse.getShardRequest().responses.stream()
                 .filter(sr -> sr.getException() != null)
-                .findAny()
+                .findFirst()
                 .orElse(previousResponse);
           }
         } else {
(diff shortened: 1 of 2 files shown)


HoustonPutman:
Ultimately I think we should redesign the ShardHandler interface, because this contract really doesn't make much sense (return 1 response that represents a request, but the user can't choose which request?). It can be made a lot better IMO.

gerlowskija:
This LGTM - is there a flaky test that this PR is expected to fix?  I can try to keep a second set of eyes to make sure it's doing what we expect post-merge.

> we should redesign the ShardHandler interface, because this contract really doesn't make much sense

Agreed, I think things could be a lot simpler if we went back to the drawing board on the design.  It's also really difficult to reason about.  Instance-nesting and code like `previousResponse.getShardRequest().responses` makes things hard to reason about at a glance (IMO)

HoustonPutman:
> is there a flaky test that this PR is expected to fix?

Yes. Both `DistributedDebugComponentTest.testTolerantSearch` and `TestJsonFacets.testTolerant`.

> Instance-nesting and code like previousResponse.getShardRequest().responses makes things hard to reason about at a glance (IMO)

I completely agree, and that's not really necessary anymore IMO. Except that the API that ShardHandler promises says that the response given back will contain the error. The SearchHandler code doesn't use that anymore, but still it's good to keep the contract until we change the APIs entirely.

dsmiley:
Agreed on redesign!  I've argued that a `CompletionService` is what's needed.

PR #3305, opened 2025-04-02
SOLR-17620 Encode the version of Solr in SolrCloud "liveNode"

QFREEE:
https://issues.apache.org/jira/browse/SOLR-17620

# Description

Please provide a short description of the changes you're making with this pull request.

Added following JSON to live node znode in [ZkController.java](https://github.com/apache/solr/pull/3305/files/83ab0971c0176f794738ff0649bad00e6277c024#diff-5b63503605ede4384429e74d1fa0c410adc5da8f3246e8c36e49feff2f3ea692).

Added access method in [ZkStateReader.java](https://github.com/apache/solr/pull/3305/files/83ab0971c0176f794738ff0649bad00e6277c024#diff-9b1918d641f78a8c245306b47ffd7a4554a12771ed4cb42293a411537fe7b54d)

```
{
  "solrVersion":"10.0.0",
  "nodeName":"localhost:8983_solr",
  "roles":{
    "data":"on",
    "overseer":"allowed",
    "coordinator":"off"}
}
```

# Solution

Please provide a short description of the approach taken to implement your solution.

# Tests

Please describe the tests you've developed or run to confirm this patch implements the feature or solves the problem.

# Checklist

Please review the following and check all that apply:

- [ ] I have reviewed the guidelines for [How to Contribute](https://github.com/apache/solr/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.
- [ ] I have created a Jira issue and added the issue ID to my pull request title.
- [ ] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended, not available for branches on forks living under an organisation)
- [ ] I have developed this patch against the `main` branch.
- [ ] I have run `./gradlew check`.
- [ ] I have added tests for my changes.
- [ ] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

PR #3431, opened 2025-07-18
SOLR-17664: Set a cluster Solr Version

dsmiley:
In cluster properties.  Solr 10 initializes this property on ZK initialization; won't touch existing clusters.

https://issues.apache.org/jira/browse/SOLR-17664

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/ZkController.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ZkController.java
@@ -52,6 +52,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import org.apache.curator.framework.api.ACLProvider;
+import org.apache.solr.client.api.util.SolrVersion;
 import org.apache.solr.client.solrj.SolrClient;
 import org.apache.solr.client.solrj.cloud.SolrCloudManager;
 import org.apache.solr.client.solrj.impl.CloudHttp2SolrClient;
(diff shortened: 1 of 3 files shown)


dsmiley:
Needs a test...

HoustonPutman:
If this is just the version that the Solr Cluster was created on, what does that give us? In the future when the Lucene upgrade issue is addressed, this version could be multiple major versions out of date.

Now that we have solr version information in live nodes (I think I remember that being added), wouldn't it be better to use that and add some utilities around it?

dsmiley:
Your initial question seems answered by JIRA, which is where the greater context is.

I wish the live nodes version mechanism might be sufficient but I worry about the possibility of an older version joining a cluster comprised of newer versions.  Should we make such a scenario disallowable (halt node start-up)?

dsmiley:
CC @psalagnac , wonder what your thoughts are.  This PR basically puts a version into ZK.  Houston supposes it'd be sufficient to use the least version of any node as reported in live nodes (assumes SOLR-17620 #3305).  The greater context is that such a thing is used to make a decision like to use Overseer, or potentially other cluster decisions that might be conceived (say PRS, ...)

dsmiley:
In #3643 , I attempted to use the version from live nodes to decide to use the overseer or not.  @HoustonPutman , you pointed out that the last upgrading node in a rolling upgrade wouldn't realize that the cluster was considered to be a lower version previously, and thus would not use the overseer when the rest of the cluster is still using the overseer.  Solving it is possible but seems awkward to me.  It's debatable.

I think this approach here wouldn't have that problem, and it's simpler to reason about.  It's safer.

Again, read the short JIRA for the greater context.

PR #3432, opened 2025-07-21
Restore logging level after package tool.

psalagnac:
This is mostly to fix following tests that run in the same JVM and check logger events.

This should make smoke tests pass more easily. That's not a blocker for 9.9, so opening a PR against main first.

### DIFF
--- a/solr/core/src/java/org/apache/solr/cli/PackageTool.java
+++ b/solr/core/src/java/org/apache/solr/cli/PackageTool.java
@@ -29,6 +29,7 @@ import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
 import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.core.LoggerContext;
 import org.apache.logging.log4j.core.config.Configurator;
 import org.apache.lucene.util.SuppressForbidden;
 import org.apache.solr.client.solrj.SolrClient;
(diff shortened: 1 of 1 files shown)


HoustonPutman:
Oh my god, thank you so much. Amazing find! Spent so long trying to figure it out.

I wonder if we could turn this into a test rule for all tests, like we do for sys props. That way no one can break tests like this in the future.

epugh (approved):
LGTM.  

I will say, I never quite understood why this stack trace thing was specific to the PackageTool?  I suspect that isn't actually true, and maybe we could fix this at the SolrCLI level, or even just not have this special handling on the PackageTool, and everything would still work???

psalagnac:
> I wonder if we could turn this into a test rule for all tests, like we do for sys props. That way no one can break tests like this in the future.

That's a great suggestion. We already have `LogLevelTestRule`, but right now it only handles levels that are overridden by test code. We could add checks so the test fails in case production code does not restore logging levels.

> and maybe we could fix this at the SolrCLI level

I personally dislike the approach of shutting down loggers. This makes debugging of CLI tools impossible in case something goes wrong.
I understand we want a clean output in the user console. Logs should not go there. Can't we have a dedicated `log4j.xml` for CLI tools? So relevant logs go to a `solr-cli.log` file or similar.

HoustonPutman:
Yeah, this should be done another way and be a forbidden API probably. The current logLevel annotation rule really does enough.

PR #3433, opened 2025-07-21
SOLR-16667: LTR Add feature vector caching for ranking

aruggero:
https://issues.apache.org/jira/browse/SOLR-16667

# Description

The current definition and usage of the QUERY_DOC_FV feature cache has been modified to support both reranking and logging.

# Solution

- The cache has been defined in the SolrConfig as the filter cache, the query result cache etc..
- Lookups and insertions in the cache have been integrated in both reranking and logging phases.

# Tests

Tests have been added in the solr/modules/ltr/src/test/org/apache/solr/ltr/TestFeatureVectorCache.java file.
The tests check for the correct cache usage and response in different scenarios, considering the ltr parameters: logAll, store, efis... and their defaults.

# Checklist


solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -716,6 +727,11 @@ public <K, V> boolean regenerateItem(
           });
     }
 
+    if (solrConfig.featureVectorCacheConfig != null
+        && solrConfig.featureVectorCacheConfig.getRegenerator() == null) {
+      solrConfig.featureVectorCacheConfig.setRegenerator(new NoOpRegenerator());
```
alessandrobenedetti: With noOpRegenerator, we don't have autowarming, do we?
Any motivation?

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRScoringQuery.java
```diff
@@ -368,17 +376,7 @@ public class ModelWeight extends Weight {
...
-     *  What is the advantage of using a hashmap here instead of an array of objects?
-     *     A set of arrays was used earlier and the elements were accessed using the featureId.
-     *     With the updated logic to create weights selectively,
-     *     the number of elements in the array can be fewer than the total number of features.
-     *     When [features] are not requested, only the model features are extracted.
-     *     In this case, the indexing by featureId, fails. For this reason,
-     *     we need a map which holds just the features that were triggered by the documents in the result set.
-     *
-     */
+    // All the features
```
alessandrobenedetti: too vague, all the features across stores? Within one feature store? Within one model?

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRScoringQuery.java
```diff
@@ -551,16 +554,117 @@ public DocIdSetIterator iterator() {
...
+        }
+
+        private void fillFeaturesInfo() throws IOException {
+          if (activeDoc == targetDoc) {
+            SolrCache<Integer, float[]> featureVectorCache = null;
+            float[] featureVector;
+
+            // Check added otherwise org.apache.solr.ltr.TestLTRScoringQuery.testLTRScoringQuery
+            // and org.apache.solr.ltr.TestSelectiveWeightCreation.testScoringQueryWeightCreation
+            // fail
```
alessandrobenedetti: definitely need to elaborate here

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRScoringQuery.java
```diff
@@ -551,16 +554,117 @@ public DocIdSetIterator iterator() {
...
+          fillFeaturesInfo();
+          return makeNormalizedFeaturesAndScore();
+        }
+
+        @Override
+        public float getMaxScore(int upTo) throws IOException {
+          return Float.POSITIVE_INFINITY;
+        }
+
+        private void fillFeaturesInfo() throws IOException {
```
alessandrobenedetti: not sure this is a 'scorer' responsibility, it feels more like 'FeatureExtractor' kind of responsibility, please investigate a new class that can be used by both the scorer and logger

solr/modules/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
```diff
@@ -75,11 +75,10 @@ public class LTRFeatureLoggerTransformerFactory extends TransformerFactory {
   // used inside fl to specify to log (all|model only) features
   private static final String FV_LOG_ALL = "logAll";
 
-  private static final String DEFAULT_LOGGING_MODEL_NAME = "logging-model";
+  public static final String DEFAULT_LOGGING_MODEL_NAME = "logging-model";
```
alessandrobenedetti, line 78: do we need this public? maybe package level? maybe we need to refactor the packages a bit

solr/modules/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
```diff
@@ -423,16 +420,13 @@ private void implTransform(SolrDocument doc, int docid, DocIterationInfo docInfo
...
-              featureLogger.makeFeatureVector(
-                  LTRRescorer.extractFeaturesInfo(
-                      rerankingModelWeight,
-                      docid,
-                      (!docsWereReranked && docsHaveScores) ? docInfo.score() : null,
-                      leafContexts));
-        }
+        String featureVector =
+            featureLogger.printFeatureVector(
+                LTRRescorer.extractFeaturesInfo(
```
alessandrobenedetti: mmmm extractFeatureInfo could actually just read from cache?
maybe we need to change the name then to something like
'fetchCacheableFeatureVector' or something similar

solr/modules/ltr/src/test-files/solr/collection1/conf/solrconfig-ltr.xml
```diff
@@ -11,7 +11,7 @@
  language governing permissions and limitations under the License. -->
 
 <config>
-    <luceneMatchVersion>${tests.luceneMatchVersion:LATEST}</luceneMatchVersion>
+ <luceneMatchVersion>${tests.luceneMatchVersion:LATEST}</luceneMatchVersion>
```
alessandrobenedetti: do we need this change?

alessandrobenedetti (changes requested):
There are some changes and some discussions to do, then I'll review the tests!

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
aruggero: From what I have seen, the docId is not the Solr one, but it would be better to check together.

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRScoringQuery.java
aruggero: Would something like "// List of features. This object has the feature store dimension but is populated only with the necessary features." be better?

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRScoringQuery.java
aruggero: I don't know if it exists a scenario where the request could be null. In my first implementation, this check (request != null) was not present.
After executing tests, I noticed that for the ones listed in the comment, the request is null; therefore, it is not possible to get the cache, and the code is raising an error...
I wasn't able to modify the test in order to populate the request.

solr/modules/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
aruggero: We can discuss this.. I need to access it inside the LTRScoringQuery.java
By now, this class is only called at the beginning to create the transformer; there is no "TransformerFactory object" I can access later on.

solr/modules/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
aruggero: The "extractFeaturesInfo" method is doing something before fetching the cache, and it uses it only if available..
What exactly do you mean here? To verify if it is possible to skip all the logic in here and directly do a cache lookup?

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRScoringQuery.java
aruggero: Also, it's better to discuss this together. By now, this method invokes the already existing "extractFeatureVector()" method, which depends on the type of scorer currently used (it can be sparse or dense, we don't know a priori).
It also uses some variables that are "model weight" variables (parent class).

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRScoringQuery.java
aruggero: Ok, comment removed.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
aruggero: The docId is the Lucene doc ID. Let's keep this without a regenerator by now. For feature enhancement, it could be considered using the unique Solr key as the key for the feature vector cache instead of the Lucene ID.

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRScoringQuery.java
aruggero: Changed

gradle/libs.versions.toml
```diff
@@ -436,9 +436,9 @@ kotlinx-coroutines-test = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-t
 kotlinx-datetime = { module = "org.jetbrains.kotlinx:kotlinx-datetime", version.ref = "kotlinx-datetime" }
 kotlinx-serialization-core = { module = "org.jetbrains.kotlinx:kotlinx-serialization-core", version.ref = "kotlinx-serialization" }
 kotlinx-serialization-json = { module = "org.jetbrains.kotlinx:kotlinx-serialization-json", version.ref = "kotlinx-serialization" }
+ktor-client-auth = { module = "io.ktor:ktor-client-auth", version.ref = "ktor" }
 ktor-client-cio = { module = "io.ktor:ktor-client-cio", version.ref = "ktor" }
 ktor-client-contentNegotiation = { module = "io.ktor:ktor-client-content-negotiation", version.ref = "ktor" }
-ktor-client-auth = { module = "io.ktor:ktor-client-auth", version.ref = "ktor" }
```
aruggero: Did by the "./gradlew tidy" command.. not related to the contribution..

solr/modules/ltr/src/java/org/apache/solr/ltr/LTRScoringQuery.java
aruggero: Created a dedicated FeatureExtractor class for this.

solr/modules/ltr/src/java/org/apache/solr/ltr/response/transform/LTRFeatureLoggerTransformerFactory.java
aruggero: Moved the method extractFeaturesInfo() inside LTRFeatureLoggerTransformerFactory since it is related to logging.

solr/modules/ltr/src/test/org/apache/solr/ltr/TestFeatureVectorCache.java
```diff
@@ -0,0 +1,369 @@
...
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class TestFeatureVectorCache extends TestRerankBase {
+  SolrCore core = null;
+
+  @Before
+  public void before() throws Exception {
+    setupFeatureVectorCachetest(false);
```
alessandrobenedetti: camel case

solr/modules/ltr/src/test/org/apache/solr/ltr/TestFeatureVectorCache.java
```diff
@@ -0,0 +1,369 @@
...
+
+    final String docs0fv_default_csv =
+        chooseDefaultFeatureVector(docs0fv_dense_csv, docs0fv_sparse_csv);
+
+    final SolrQuery query = new SolrQuery();
+    query.setQuery("*:*");
+    query.add("rows", "3");
+    query.add("fl", "[fv efi.efi_feature=3]");
+
+    // No caching, we want to see lookups, an insertions and no hits
```
alessandrobenedetti: typo 'an'

solr/modules/ltr/src/test/org/apache/solr/ltr/TestFeatureVectorCache.java
```diff
@@ -0,0 +1,369 @@
...
+            + docs0fv_default_csv
+            + "'}");
+    filterCacheMetrics = lookupFilterCacheMetrics(core);
+    assertEquals(2, (long) filterCacheMetrics.get("inserts"));
+    assertEquals(8, (long) filterCacheMetrics.get("lookups"));
+    assertEquals(6, (long) filterCacheMetrics.get("hits"));
+  }
+
+  @Test
+  public void testFeatureVectorCache_loggingAllStoreReranking() throws Exception {
```
alessandrobenedetti: loggingAllFeatureStoreAndReranking

solr/CHANGES.txt
```diff
@@ -323,6 +323,8 @@ New Features
   subset of JavaScript, pre-compiled, and that which can access the score and fields. It's powered by
   the Lucene Expressions module. (hossman, David Smiley, Ryan Ernst, Kevin Risden)
 
+* SOLR-16667: LTR Add feature vector caching for ranking. (Anna Ruggero, Alessandro Benedetti)
```
aruggero: Move to 10

aruggero:
We have finished the review iterations, and the code is open to further suggestions and revisions. In the meantime, I am running a benchmark to report the performance of the new contribution compared to the current implementation.

solr/solr-ref-guide/modules/query-guide/pages/learning-to-rank.adoc
```diff
@@ -131,18 +131,17 @@ See xref:configuration-guide:solr-modules.adoc[Solr Module] for more details.
...
-       class="solr.search.CaffeineCache"
-       size="4096"
-       initialSize="2048"
-       autowarmCount="4096"
-       regenerator="solr.search.NoOpRegenerator" />
+<featureVectorCache class="solr.CaffeineCache" size="4096" initialSize="2048" autowarmCount="0"/>
 ----
 
+[NOTE]
+The `featureVectorCache` holds Lucene Document IDs. Since they are transient, this cache is not auto-warmed.
```
alessandrobenedetti: The `featureVectorCache` key is computed using the  Lucene Document ID (necessary for document-level features). 
Since these IDs are transient, this cache does not support auto-warming.

alessandrobenedetti (approved):
Aside some minors, we are ready to merge, just waiting for main to be stable

aruggero:
As mentioned, I report here the results of the benchmark that has been done:
Screenshot 2025-10-21 at 14 08 20

We compared the current main branch containing the FV_cache for LTR with our contributed cache. We have also done a test in both situations with all caches with zero size to be sure not to slow down the query execution when not using caches at all.

The first column represents the difference in milliseconds between the first query (no cache hit - miss than insert) and the second query (cache hit). The higher, the better, since it means that the cache has a great impact on reducing the query execution time.

The second column represents the average time taken to execute the first query in milliseconds (the one that does not find any entry - miss than insert).

The third column represents the average time taken to execute the second query in milliseconds (the one that finds an entry - hit).

Finally, the last columns are the cache statistics from the Solr UI.

We executed 10 pairs of queries (the first having a miss, the second a hit in the cache), each retrieving 10000 documents.
We uploaded a feature store of 200 features and use a linear model that exploits these 200 features.

All the results are positive. We obtain a slightly slower result when doing the "miss" query during pure reranking, but we still have a comparable query execution time.

A detailed blog post about this will follow on [sease.io](https://sease.io/)

der-eismann:
Sorry for hijacking this "old" PR, but https://issues.apache.org is down right now and I don't want to forget it.
We're currently testing Solr 10 and stumbled over this change because restoring a backup fails with:
```
Error CREATEing SolrCore 'example_shard1_replica_n1': Unable to create core
 [example_shard1_replica_n1] Caused by: No setter corrresponding to
 'fvCacheName' in org.apache.solr.ltr.response.transform.LTRFeatureLoggerTransformerFactory
```
The core config looks like this:
```xml
    <cache name="QUERY_DOC_FV"
           class="solr.search.CaffeineCache"
           size="4096"
           initialSize="2048"
           autowarmCount="4096"
           regenerator="solr.search.NoOpRegenerator"
    />
    <transformer name="features" class="org.apache.solr.ltr.response.transform.LTRFeatureLoggerTransformerFactory">
        <str name="fvCacheName">QUERY_DOC_FV</str>
    </transformer>
```
In the [LTR docs](https://solr.apache.org/guide/solr/latest/query-guide/learning-to-rank.html#ltr-configuration) the entries still have the `<str name="fvCacheName">QUERY_DOC_FV</str>` entry, in the [logging features](https://solr.apache.org/guide/solr/latest/query-guide/learning-to-rank.html#feature-logging-parameters) as well. Can someone please update the docs and maybe also add an entry to the Solr 10 migration docs? I don't really know how to migrate this.

aruggero:
> Sorry for hijacking this "old" PR, but https://issues.apache.org is down right now and I don't want to forget it. We're currently testing Solr 10 and stumbled over this change because restoring a backup fails with:
> 
> ```
> Error CREATEing SolrCore 'example_shard1_replica_n1': Unable to create core
>  [example_shard1_replica_n1] Caused by: No setter corrresponding to
>  'fvCacheName' in org.apache.solr.ltr.response.transform.LTRFeatureLoggerTransformerFactory
> ```
> 
> The core config looks like this:
> 
> ```
>     <cache name="QUERY_DOC_FV"
>            class="solr.search.CaffeineCache"
>            size="4096"
>            initialSize="2048"
>            autowarmCount="4096"
>            regenerator="solr.search.NoOpRegenerator"
>     />
>     <transformer name="features" class="org.apache.solr.ltr.response.transform.LTRFeatureLoggerTransformerFactory">
>         <str name="fvCacheName">QUERY_DOC_FV</str>
>     </transformer>
> ```
> 
> In the [LTR docs](https://solr.apache.org/guide/solr/latest/query-guide/learning-to-rank.html#ltr-configuration) the entries still have the `<str name="fvCacheName">QUERY_DOC_FV</str>` entry, in the [logging features](https://solr.apache.org/guide/solr/latest/query-guide/learning-to-rank.html#feature-logging-parameters) as well. Can someone please update the docs and maybe also add an entry to the Solr 10 migration docs? I don't really know how to migrate this.


Thanks for letting me know!
I'll fix it and review all the documentation again.
I'll let you know as soon as it is done.

der-eismann:
> Thanks for letting me know! I'll fix it and review all the documentation again. I'll let you know as soon as it is done.

Thank you so much for the quick reply, highly appreciated!

alessandrobenedetti:
Hi @der-eismann, just to mention, this is not currently on our open source team's priority list.
Happy to engage with a paid commitment on this, or feel free to contribute it yourself; the committer team will gladly review!
Cheers

PR #3436, opened 2025-07-23
SOLR-17721 - null pointer fix in the doAddDistinct method

puneeetsharma:
https://issues.apache.org/jira/browse/SOLR-17721

# Description

This issue is very intermittent and depending on a number of factors.Fixed a NullPointerException in Solr's atomic update functionality when performing add-distinct operations on fields that exist in the document but have null values. The error "Cannot invoke 'java.util.Collection.size()' because 'original' is null" was occurring in the doAddDistinct method of AtomicUpdateDocumentMerger when getValues() returned null for fields with null values.

# Solution

Added a null check in the doAddDistinct method to handle the case where existingField.getValues() returns null. When a field exists in the document but has a null value, getValues() returns null, but the code was trying to call size() on this null collection. The fix initializes an empty ArrayList when getValues() returns null, ensuring the atomic update operation can proceed without throwing a NullPointerException.
Code Changes:
Modified solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java in the doAddDistinct method
Added null check: if (original == null) { original = new ArrayList<>(); }

# Tests

Added a test case testAddDistinctToNullField() in solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesTest.java that reproduces the exact scenario where:
A field exists in the document (existingField != null)
The field has a null value (getValues() returns null)
An add-distinct atomic update operation is performed on that field
The test verifies that the operation completes successfully without throwing a NullPointerException and that the new value is correctly added to the field. The test suite was run to ensure no existing functionality was broken, with all 28 tests passing (1 skipped).

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
+++ b/solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
@@ -545,8 +545,11 @@ public class AtomicUpdateDocumentMerger {
     final String name = sif.getName();
     SolrInputField existingField = toDoc.get(name);
 
-    Collection<Object> original =
-        existingField != null ? existingField.getValues() : new ArrayList<>();
+    Collection<Object> original = existingField != null ? existingField.getValues() : null;
+
+    if (original == null) {
+      original = new ArrayList<>();
+    }
 
     int initialSize = original.size();
     if (fieldVal instanceof Collection) {
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -218,6 +218,8 @@ Bug Fixes
 
 * SOLR-17824: RecoveryStrategy.pingLeader could NPE when there's no shard leader (David Smiley)
 
+* SOLR-17721: NPE can occur when doing Atomic Update using Add Distinct on documents with a null field value. (puneetSharma via Eric Pugh)
+
 Dependency Upgrades
 ---------------------
 (No changes)
--- a/solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesTest.java
+++ b/solr/core/src/test/org/apache/solr/update/processor/AtomicUpdatesTest.java
@@ -1704,4 +1704,27 @@ public class AtomicUpdatesTest extends SolrTestCaseJ4 {
         "*[count(//result/doc[1]/arr[@name='intRemove']/int)=1]",
         "//result/doc[1]/arr[@name='intRemove']/int[1][.=333]");
   }
+
+  @Test
+  public void testAddDistinctToNullField() {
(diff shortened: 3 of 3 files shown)


fructosepadre (approved):

epugh:
Is "doc.setField("cat", null); // Set field to null" a normal thing to do?

solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
```diff
@@ -545,8 +545,11 @@ protected void doAddDistinct(SolrInputDocument toDoc, SolrInputField sif, Object
     final String name = sif.getName();
     SolrInputField existingField = toDoc.get(name);
 
-    Collection<Object> original =
-        existingField != null ? existingField.getValues() : new ArrayList<>();
+    Collection<Object> original = existingField != null ? existingField.getValues() : null;
```
epugh, line 548: how is this differnet than what was previous?

puneeetsharma:
> Is "doc.setField("cat", null); // Set field to null" a normal thing to do?

I have seen people do it because, during document building, fields with null values are skipped entirely and not added to the Lucene document and does't cause any index size bloating. But getValues() returning null and causing the atomic update failures.

puneeetsharma:
Thanks for the approval, @dsmiley. Could you please help merge this PR?

epugh:
If David doesn't get to it I will..   how would you liked to be referenced I. The commit message?

puneeetsharma:
Thanks @epugh you can reference me with puneetSharma in the commit message.

epugh:
@puneeetsharma I realized I had a pending code comment from my review that I never submitted (I guess thats what I get for using my phone~)...    Can you answer the question, and then I'll add CHANGES and move it along...

solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java
puneeetsharma: Previously, if existingField was null, we were assigning original to a new ArrayList<>, which was fine. However, if existingField.getValues() returned null, then original would also be assigned null, leading to a NullPointerException when calling original.size().

In the new code, whether existingField is null or existingField.getValues() returns null, original is initially set to null. We then explicitly check for null and assign it to a new ArrayList<> if needed. This ensures that original is never null when calling size(), thus avoiding the NPE.

epugh: thank yo8u for breaking this down for me...

epugh:
Thanks!  I back ported it to branch 9x as well.   Keep the fixes coming!

PR #3439, opened 2025-07-24
Solr 17294 backport to 9x

markrmiller:


janhoy:
Heads up that a 9.10 release will happen next week. So there is still time to get this PR in.

PR #3441, opened 2025-07-25
SOLR-17715: Remove qt parameter from the final parameters sent to server

r4mercur:
https://issues.apache.org/jira/browse/SOLR-17715

# Description

# Solution

# Tests

# Checklist

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/client/solrj/request/QueryRequest.java
+++ b/solr/solrj/src/java/org/apache/solr/client/solrj/request/QueryRequest.java
@@ -19,6 +19,7 @@ package org.apache.solr.client.solrj.request;
 import java.util.Objects;
 import org.apache.solr.client.solrj.response.QueryResponse;
 import org.apache.solr.common.params.CommonParams;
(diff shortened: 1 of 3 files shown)


dsmiley:
Coming back to this, I think this is close to what we should do.  But I'd rather see the path extracted in the constructor so that QueryRequest no longer overrides getPath.

Also the placement of these tests seems arbitrary/wrong to me, but the tests themselves are very good.  Can you please create a new PR?  If you are no longer interested, I'll maybe tell an AI agent to do this.

epugh:
I was just updating a downstream plugin for Solr and it had `qt` and it would be so nice to get rid of it!

dsmiley:
And add a new constructor to QueryRequest to put the path first explicitly without resorting to "qt" param.  The others can be deprecated.

r4mercur:
@dsmiley one question before i would start a new try to tackle this. Where would you see the tests-cases best placed ?

dsmiley:
Just start a new testcase, as I see no existing one.

Thanks for your interest!

PR #3443, opened 2025-07-25
SOLR-17827: Remove JXM, SLF4J and Graphite reporters

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-17827

Deprecating JMX, SLF4J and Graphite reporters which were wrapped around Dropwizard. If users have specific ways they want to get their metrics, they should turn to using OTLP with the OTEL collector and the collectors supported output plugins or something else that accepts OTLP with their preferred backend.

This should be much better as Solr doesn't need to maintain application/vendor specific backends for metrics while also slimming down Solrs maintenance.


dsmiley (approved):
This PR title is "deprecate" but in fact you are removing/deleting, and the JIRA says that too.

Please list what you removed in a new section in solr/solr-ref-guide/modules/upgrade-notes/pages/major-changes-in-solr-10.adoc

PR #3445, opened 2025-07-29
Jetty 12 EE8 for Solr 9.x

msternermxy:
Draft of upgrading to Jetty 12 EE8 in Solr 9.x. Seems to be running fine. Some tests still fail. Requires a new module `io.dropwizard.metrics:metrics-jetty12-ee8` to be contributed separately.


janhoy:
Solr 9.10 will be released next week. If you have plans to get this in, now is the time :)

msternermxy:
I don't think there was any interest in this feature, based on the mailing list response. Looking forward to Solr 10 with non-EOL Jetty instead.

PR #3446, opened 2025-07-29
SOLR-17831: Fix the bug in QueryLimits initialization in SolrIndexSearcher.

sigram:
Improve the helper class to better track locations and counts where limits are enforced.

https://issues.apache.org/jira/browse/SOLR-17831

This PR will also improve testability of using ExitableDirectoryReader by better instrumentation in CallerSpecificQueryLimit and making the sysproperty modifiable by using EnvUtils.


solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -775,7 +792,7 @@ public QueryResult search(QueryResult qr, QueryCommand cmd) throws IOException {
   protected void search(List<LeafReaderContext> leaves, Weight weight, Collector collector)
       throws IOException {
     QueryLimits queryLimits = QueryLimits.getCurrentLimits();
-    if (useExitableDirectoryReader || !queryLimits.isLimitsEnabled()) {
+    if (EnvUtils.getPropertyAsBool(EXITABLE_READER_PROPERTY) || !queryLimits.isLimitsEnabled()) {
```
dsmiley: should this instead check if the existing reader is an instance of ExitableDirectoryReader?  It would have already been wrapped.

solr/core/src/test/org/apache/solr/search/TestQueryLimits.java
```diff
@@ -88,13 +89,12 @@ public void testQueryLimits() throws Exception {
...
           "should have partial results for expr " + matchingExpr,
           rsp.getHeader().get("partialResults"));
-      if (matchingExpr.contains(".")) {
-        assertEquals(matchingExpr, limit.trippedBy);
-      } else {
-        assertTrue(
-            "expected result to start with " + matchingExpr + " but was " + limit.trippedBy,
-            limit.trippedBy.startsWith(matchingExpr));
-      }
+      assertFalse("should have trippedBy info", limit.getTrippedBy().isEmpty());
```
dsmiley, line 92: I suspect you intended to put the trip info in the assert message string.

dsmiley (commented):
that's some impressive testing... I didn't know we had the ability to assert who exactly tripped the limit.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
sigram: I wonder in what scenario this would happen, realistically ... and whether we should care here. We need to make sure that the way we wrapped the original reader uses our implementation of the QueryTimeout. We have no way to determine that if the input is an EDR of unknown origin.

solr/core/src/test/org/apache/solr/search/TestQueryLimits.java
sigram: That's in the next assertion, here it's just checking that it's not empty - if it is then there's nothing to print.

sigram:
The helper class `CallerSpecificQueryLimit` helped me to preserve my sanity while working on SOLR-17172, when trying to figure out why a limit was reached (or the opposite, as was often the case). The way it works is not really specific to the QueryLimits api, so if you think it could be useful for some other tests then it's easy to generalize it.

dsmiley (approved):

PR #3397, opened 2025-06-20
Jira/solr 17789

bct-timo-crabbe:
https://issues.apache.org/jira/browse/SOLR-17789

# Description

Solr nodes do not pass full authorization details to other nodes.
Steps to reproduce:

1. Deploy cluster with more then one node
2. Use a authentication plugin where roles are supplied externally (like JWTAuth).
3. Add a private collection with lower number of replicas then the number of nodes in the cluster
4. Send request to node that does not holds a replica of the collection to force forwarding.

This results in a return code 403.

# Solution

Add the current user's Security Principal on the HttpClientContext in the sendRemoteQuery method (`solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java`) like the executeMethod method on the HttPSolrClient (`solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpSolrClient.java`)

# Tests

testInternodeAuthorization on the jwt-auth plugin (`solr\modules\jwt-auth\src\test\org\apache\solr\security\jwt\JWTAuthPluginIntegrationTest.java`).

This test will setup a test cluster with three nodes, create a private collection with only two replicas and will query every node for the documents in the collection.

# Checklist

Please review the following and check all that apply:

- [x] I have reviewed the guidelines for [How to Contribute](https://github.com/apache/solr/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.
- [x] I have created a Jira issue and added the issue ID to my pull request title.
- [x] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended, not available for branches on forks living under an organisation)
- [ ] I have developed this patch against the `main` branch.
 - [x] I have developed this patch against the `branch_9x` branch.
- [x] I have run `./gradlew check`.
 - Failed on benchmarking tests 
- [x] I have added tests for my changes.
- [ ] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

PR #3447, opened 2025-08-01
SOLR-17286: Remote Proxy via Jetty HttpClient

dsmiley:
No longer Apache HttpClient.

https://issues.apache.org/jira/browse/SOLR-17286

### DIFF
--- a/solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
+++ b/solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
@@ -37,15 +37,11 @@ import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 import java.io.EOFException;
 import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
 import java.lang.invoke.MethodHandles;
 import java.nio.charset.StandardCharsets;
-import java.security.Principal;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
@@ -60,26 +56,10 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Supplier;
 import net.jcip.annotations.ThreadSafe;
-import org.apache.http.Header;
-import org.apache.http.HeaderIterator;
-import org.apache.http.HttpEntity;
-import org.apache.http.HttpEntityEnclosingRequest;
-import org.apache.http.HttpResponse;
-import org.apache.http.client.methods.HttpDelete;
-import org.apache.http.client.methods.HttpEntityEnclosingRequestBase;
-import org.apache.http.client.methods.HttpGet;
-import org.apache.http.client.methods.HttpHead;
-import org.apache.http.client.methods.HttpOptions;
-import org.apache.http.client.methods.HttpPost;
-import org.apache.http.client.methods.HttpPut;
-import org.apache.http.client.methods.HttpRequestBase;
-import org.apache.http.client.protocol.HttpClientContext;
-import org.apache.http.entity.InputStreamEntity;
 import org.apache.solr.api.ApiBag;
 import org.apache.solr.api.V2HttpCall;
 import org.apache.solr.client.api.util.SolrVersion;
 import org.apache.solr.client.solrj.impl.CloudSolrClient;
-import org.apache.solr.client.solrj.impl.HttpClientUtil;
 import org.apache.solr.common.SolrException;
 import org.apache.solr.common.SolrException.ErrorCode;
 import org.apache.solr.common.cloud.Aliases;
@@ -130,6 +110,7 @@ import org.apache.solr.update.processor.DistributingUpdateProcessorFactory;
 import org.apache.solr.util.RTimerTree;
 import org.apache.solr.util.tracing.TraceUtils;
 import org.apache.zookeeper.KeeperException;
+import org.eclipse.jetty.client.HttpClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MarkerFactory;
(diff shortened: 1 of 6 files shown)


solr/core/src/java/org/apache/solr/servlet/HttpSolrProxy.java
```diff
@@ -0,0 +1,148 @@
...
+        .getHeaderNames()
+        .asIterator()
+        .forEachRemaining(
+            headerName -> {
+              HttpHeader knownHeader = HttpHeader.CACHE.get(headerName); // maybe null
+              if (!HOP_BY_HOP_HEADERS.contains(knownHeader)) {
+                // NOCOMMIT see SOLR-7274 SPNEGO support:
+                // && HttpHeader.HOST != knownHeader
+                // && HttpHeader.AUTHORIZATION != knownHeader
+                // && HttpHeader.ACCEPT != knownHeader) {
```
dsmiley: @anshumg, 10 years ago in [SOLR-17286](https://issues.apache.org/jira/browse/SOLR-17286),  you committed this particular exception list of 3 headers to not copy on the grounds of SPNEGO.  In the comments, it was actually @chatman who pointed this out.  I don't understand it; my thinking is that a proxy should be a proxy and pass everything through except for the hop-by-hop header list according to RFC.  If I restore these exceptions, then the wonderful test recently added will fail: https://github.com/apache/solr/blob/eaceca8911ef06c89a293f2437a6a560f34b1a43/solr/modules/jwt-auth/src/test/org/apache/solr/security/jwt/JWTAuthPluginIntegrationTest.java#L301
(see PR #3397 that added it)

solr/core/src/java/org/apache/solr/servlet/HttpSolrProxy.java
dsmiley, line 1: I refactored basically all of the remote proxy logic out of HttpSolrCall into this new class.

solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
```diff
@@ -736,107 +709,21 @@ void destroy() {
     }
   }
 
-  // TODO using Http2Client
-  private void remoteQuery(String coreUrl, HttpServletResponse resp) throws IOException {
-    HttpRequestBase method;
-    HttpEntity httpEntity = null;
+  protected void sendRemoteQuery() throws IOException {
```
dsmiley: TODO I want to rename this to "doRemoteProxy"

solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
```diff
@@ -736,107 +709,21 @@ void destroy() {
...
+    SolrRequestInfo.setRequestInfo(new SolrRequestInfo(req, new SolrQueryResponse(), action));
+    mustClearSolrRequestInfo = true;
+
+    HttpClient httpClient = cores.getDefaultHttpSolrClient().getHttpClient();
```
dsmiley: FYI using the underlying Jetty `HttpClient`, one layer deeper than the `Http2SolrClient`, thus skips over the authentication plugin interception stuff since it's not registered at quite that depth.  I abandoned a previous PR that attempted to lower that logic down here.

solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
```diff
@@ -736,107 +709,21 @@ void destroy() {
...
-      Principal userPrincipal = req.getUserPrincipal();
-      if (userPrincipal != null) {
-        // Normally the context contains a static userToken to enable reuse resources. However, if a
-        // personal Principal object exists, we use that instead, also as a means to transfer
-        // authentication information to Auth plugins that wish to intercept the request later
-        if (log.isDebugEnabled()) {
-          log.debug("Forwarding principal {}", userPrincipal);
-        }
-        httpClientRequestContext.setUserToken(userPrincipal);
-      }
```
dsmiley, line 799: FYI @bct-timo-crabbe the logic you just added didn't have a role to play when I switched over to Jetty HttpClient, as best as I could tell.  I found that letting the "Authorization" header through enabled your wonderful test to pass.

solr/solrj/src/java/org/apache/solr/client/solrj/impl/Http2SolrClient.java
dsmiley, line 1: Exposing the underlying HttpClient.  This means it was pointless to have the getter to getProtocolHandlers specifically.

solr/core/src/java/org/apache/solr/servlet/HttpSolrCall.java
dsmiley: Out of curiosity, on main (not this PR), I locally removed the 3 header exceptions and also removed the code above that you added to formerly propagate the principal.  Your new test fails... not sure why.

dsmiley:
Jetty's [ProxyServlet](https://github.com/jetty/jetty.project/blob/jetty-12.0.x/jetty-ee10/jetty-ee10-proxy/src/main/java/org/eclipse/jetty/ee10/proxy/AbstractProxyServlet.java#L499) is quite nice.  It's a bit more thorough... I learned that the "Connection" header can list additional headers to remove, which I also confirmed from MDN/RFC.

dsmiley:
If it turns out that we don't need the proxy to be AuthenticationPlugin aware, that it just needs to be a proxy (the current case of my PR), then I should just replace most of the code here with Jetty's ProxyServlet.  There's a test failure right now that is a bug in my proxy that I strongly suspect would be fixed by switching to ProxyServlet.

solr/benchmark/src/java/org/apache/solr/bench/Docs.java
```diff
@@ -110,11 +110,7 @@ public Iterator<SolrInputDocument> preGenerate(int numDocs) throws InterruptedEx
       executorService.execute(() -> docs.add(Docs.this.inputDocument()));
     }
 
-    executorService.shutdown();
-    boolean result = executorService.awaitTermination(10, TimeUnit.MINUTES);
-    if (!result) {
-      throw new RuntimeException("Timeout waiting for doc adds to finish");
-    }
+    ExecutorUtil.shutdownAndAwaitTermination(executorService);
```
dsmiley, line 113: technically unrelated but it caught my eye

solr/benchmark/src/test/org/apache/solr/bench/MiniClusterBenchStateTest.java
dsmiley, line 1: This test indirectly/unintentionally exercises the remote proxy functionality.  Go fig.  Any way, when it fails, it wasn't cleaning up so I made the test better formed.  It doesn't fail now, BTW.

iamsanjay:
@dsmiley 
org.apache.solr.security.hadoop.TestRuleBasedAuthorizationWithKerberos.testReadsAltUser 
failing on 9x

https://ci-builds.apache.org/job/Solr/job/Solr-Test-9.x/8536/

PR #3451, opened 2025-08-05
[SOLR-17837] omit PULL nodes from preferredLeader prop distribution

liangkaiwen:
https://issues.apache.org/jira/browse/SOLR-17837

# Description

For BALANCESHARDUNIQUE admin collection api setting preferredLeader property, PULL nodes should be omitted from consideration as they can never be elected leader

# Solution

* In ExclusiveSliceProperty, omit PULL type nodes when property to be set is preferredLeader

# Tests

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/ExclusiveSliceProperty.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ExclusiveSliceProperty.java
@@ -129,6 +129,10 @@ class ExclusiveSliceProperty {
     return replica.getState() == Replica.State.ACTIVE;
   }
 
+  private boolean canBeLeader(Replica replica) {
+    return replica.getType() == Replica.Type.NRT || replica.getType() == Replica.Type.TLOG;
+  }
+
   // Collect a list of all the nodes that _can_ host the indicated property. Along the way, also
   // collect any of the replicas on that node that _already_ host the property as well as any slices
   // that do _not_ have the property hosted.
@@ -149,6 +153,12 @@ class ExclusiveSliceProperty {
             // Note, we won't be committing this to ZK until later.
             removeProp(slice, replica.getName());
           }
+          continue;
+        }
+        if (SliceMutator.PREFERRED_LEADER_PROP.equals(property)
+            && !canBeLeader(
+                replica)) { // omit replicas that cannot potentially be leader from preferredLeader
+          // candidates
           continue;
         }
         allHosts.add(replica.getNodeName());


solr/core/src/java/org/apache/solr/cloud/ExclusiveSliceProperty.java
```diff
@@ -151,6 +155,12 @@ private boolean collectCurrentPropStats() {
           }
           continue;
         }
+        if (SliceMutator.PREFERRED_LEADER_PROP.equals(property)
+            && !canBeLeader(
+                replica)) { // omit replicas that cannot potentially be leader from preferredLeader
```
HoustonPutman:
```suggestion
            // omit replicas that cannot potentially be leader from preferredLeader
            && replica.getType().leaderEligible) { 
```

solr/core/src/java/org/apache/solr/cloud/ExclusiveSliceProperty.java
```diff
@@ -129,6 +129,10 @@ private boolean isActive(Replica replica) {
     return replica.getState() == Replica.State.ACTIVE;
   }
 
+  private boolean canBeLeader(Replica replica) {
+    return replica.getType() == Replica.Type.NRT || replica.getType() == Replica.Type.TLOG;
+  }
+
```
HoustonPutman:
```suggestion
```

liangkaiwen:
Currently working on tests. Getting a thread leak in the test whenever I add pull replicas, but not sure why :/

liangkaiwen:
> There's an easier way to check if a replica can be leader. But otherwise, this looks like a good change. Can you make a Changelog entry as well for 9.10?

Done. Let me know if I need to fix the format of the changelog

solr/CHANGES.txt
```diff
@@ -4,6 +4,27 @@ This file lists Solr's raw release notes with details of every change to Solr.
...
+Bug Fixes
+---------------------
+* SOLR-17837: Disallow PULL replicas from being counted/marked for "preferredLeader" property by BALANCESHARDUNIQUE
+
+Dependency Upgrades
+---------------------
+
+Other Changes
+---------------------
+
```
HoustonPutman: The section already exists below 10.0.0. Just add your line there.

liangkaiwen: Oops, my main branch was out of date. Updated

HoustonPutman (approved):

liangkaiwen:
It's unclear why this test is failing (looks like the test cloud is somehow repeatedly unable to elect a leader within 4s). Running the same tests locally on my own machine do not yield the same failure

liangkaiwen:
It has been noted by one of my teammates that `AbstractFullDistribZkTestBase` has been replaced in favor of `SolrCloudTestCase` and sometimes thread leaks can occur

@HoustonPutman do you think it is worth rewriting the test class entirely or can we get this change in even with the sporadic test failures?

liangkaiwen:
@HoustonPutman can this be merged?

iamsanjay:
This test failure might be related to this commit, 

https://develocity.apache.org/scans/tests?search.rootProjectNames=solr-root&search.timeZoneId=Asia%2FCalcutta&tests.container=org.apache.solr.cloud.api.collections.TestReplicaProperties


```
  - org.apache.solr.cloud.api.collections.TestReplicaProperties (:solr:core)
    Test history: https://develocity.apache.org/scans/tests?search.rootProjectNames=solr-root&tests.container=org.apache.solr.cloud.api.collections.TestReplicaProperties
    Test output: /home/jenkins/jenkins-agent/workspace/Solr/Solr-Test-main/solr/core/build/test-results/test/outputs/OUTPUT-org.apache.solr.cloud.api.collections.TestReplicaProperties.txt
    Reproduce with: ./gradlew :solr:core:test --tests "org.apache.solr.cloud.api.collections.TestReplicaProperties" -Ptests.haltonfailure=false "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=62B45FC2B6ED55E3 -Ptests.multiplier=2 -Ptests.useSecurityManager=true -Ptests.badapples=false -Ptests.file.encoding=US-ASCII
```

HoustonPutman:
Yeah, that's right. Though it looks like the cloud clients dead server checker can't be killed for some reason. Not sure why this would do that. But @liangkaiwen what's the reason for adding so many replicas to the test? It seems like it is slowing everything down a good amount.

liangkaiwen:
The test passes on my local machine, but agreed it is quite slow (18.4s to completion).

As for the replica numbers, the number of tlog nodes is equivalent to the max possible of tlog nodes(4) in the test prior to my change with the only change there being removing the random factor. For the pull nodes, the previous value before my change was a globally hard-coded 2, however I bumped this up to 4 so that there is more likelihood the test will catch any issue where atleast 1 pull node gets assigned leader incorrectly.

It was noted by a colleague of mine that the test base class here is outdated and might potentially be contributing to the flakiness? If it would be better, I could also reduce the number of replicas to 2 tlog and 2 pull? Let me know if you would like me to proceed with either or both of these ideas

PR #3453, opened 2025-08-06
SOLR-17806: Migrate ZkContainer metrics to OTEL

mlbiscoc:

Migrate ZkContainer metrics and remove Dropwizard initializations.

```
# TYPE solr_zk_child_fetches_total counter
solr_zk_child_fetches_total{otel_scope_name="org.apache.solr"} 16.0
# HELP solr_zk_cumulative_children_fetched_total Total cumulative children fetched count
# TYPE solr_zk_cumulative_children_fetched_total counter
solr_zk_cumulative_children_fetched_total{otel_scope_name="org.apache.solr"} 7.0
# HELP solr_zk_cumulative_multi_ops_total Total cumulative multi-operations count
# TYPE solr_zk_cumulative_multi_ops_total counter
solr_zk_cumulative_multi_ops_total{otel_scope_name="org.apache.solr"} 8.0
# HELP solr_zk_ops_total Total number of ZooKeeper operations
# TYPE solr_zk_ops_total counter
solr_zk_ops_total{category="CONTAINER",ops="delete",otel_scope_name="org.apache.solr"} 0.0
solr_zk_ops_total{category="CONTAINER",ops="exists",otel_scope_name="org.apache.solr"} 218.0
solr_zk_ops_total{category="CONTAINER",ops="multi",otel_scope_name="org.apache.solr"} 3.0
solr_zk_ops_total{category="CONTAINER",ops="read",otel_scope_name="org.apache.solr"} 148.0
solr_zk_ops_total{category="CONTAINER",ops="write",otel_scope_name="org.apache.solr"} 5.0
# HELP solr_zk_read_bytes_total Total bytes read from ZooKeeper
# TYPE solr_zk_read_bytes_total counter
solr_zk_read_bytes_total{category="CONTAINER",otel_scope_name="org.apache.solr"} 458648.0
# HELP solr_zk_watches_fired_total Total number of ZooKeeper watches fired
# TYPE solr_zk_watches_fired_total counter
solr_zk_watches_fired_total{category="CONTAINER",otel_scope_name="org.apache.solr"} 5.0
# HELP solr_zk_written_bytes_total Total bytes written to ZooKeeper
# TYPE solr_zk_written_bytes_total counter
solr_zk_written_bytes_total{otel_scope_name="org.apache.solr"} 951.0
```

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/ZkContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/ZkContainer.java
@@ -38,7 +38,6 @@ import org.apache.solr.common.util.ExecutorUtil;
 import org.apache.solr.common.util.SolrNamedThreadFactory;
 import org.apache.solr.common.util.StrUtils;
 import org.apache.solr.logging.MDCLoggingContext;
-import org.apache.solr.metrics.MetricsMap;
 import org.apache.solr.metrics.SolrMetricProducer;
 import org.apache.solr.metrics.SolrMetricsContext;
 import org.apache.zookeeper.KeeperException;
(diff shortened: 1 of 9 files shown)


mlbiscoc:
There doesn't appear to be any tests around these ZkContainer metrics. If someone wants, I can create tests to assert these are correct, otherwise this PR was just a straight migration.

I also squeezed a few other changes that fixes tests when I was merging other PRs.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/SolrZKMetricsListener.java
```diff
@@ -16,40 +16,27 @@
...
 import org.apache.curator.drivers.OperationTrace;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.api.CuratorEvent;
 import org.apache.curator.framework.api.CuratorListener;
-import org.apache.solr.common.MapWriter;
-import org.apache.solr.common.annotation.JsonProperty;
-import org.apache.solr.common.util.ReflectMapWriter;
 
-public class SolrZKMetricsListener extends AdvancedTracerDriver
-    implements ReflectMapWriter, CuratorListener {
```
dsmiley, line 32: I wonder why RefflectMapWriter was implemented (hence `JsonProperty` annotations)

mlbiscoc:
> Can you post an excerpt (prometheus format) on what these looks like?

The sample is in the PR description.

solr/core/src/java/org/apache/solr/core/ZkContainer.java
```diff
@@ -141,18 +140,82 @@ public void initZooKeeper(final CoreContainer cc, CloudConfig config) {
...
+
+                ctx.observableLongCounter(
+                    "solr_zk_watches_fired",
+                    "Total number of ZooKeeper watches fired",
+                    measurement -> {
+                      measurement.record(metricsListener.getWatchesFired(), attributes);
+                    });
+
+                ctx.observableLongCounter(
+                    "solr_zk_bytes_written_total",
```
dsmiley: the excerpt in the PR description shows `solr_zk_bytes_written_bytes_total` instead of what you have here.  What's here looks better, since the PR description shows "bytes" repeated.  Similar for the read metric.

dsmiley:
Doh!  :palm-to-face: indeed you have the excerpt in the PR description, a good/obvious location.  I looked everywhere _else_ (JIRA & bottom of PR).

solr/core/src/java/org/apache/solr/core/ZkContainer.java
mlbiscoc: I fixed it and updated the PR description with the correct naming in [ee09699](https://github.com/apache/solr/pull/3453/commits/ee09699eddd7a35ef2811939ced9a821b7ee22af).

Something I think isn't obvious though is how OTELs prometheus metric reader names things after and how OTEL sets units as it's metadata. 

For example, I just fixed up the metric for zk bytes read to just name `solr_zk_read` but the prometheus output does `solr_zk_read_bytes_total`. Thats because the metric reader does some sanitizing and tries to follow prometheus data model and format with the unit metadata having `byte` and counters always ending with `total` even though I didn't add it.

What I kind of regret was the way I setup the `unit` parameter because you can see I set bytes as "By". Thats because OTEL only follows [UCUM](https://ucum.org/ucum) for its units otherwise the metric throws an exception. I am going to go back with another PR to fix the units so that devs can just use a common set of constants UCUM units instead of passing in a string.

dsmiley: ah; cool.

Merge away!

PR #3458, opened 2025-08-08
SOLR-17806: Recreate OTEL metric registries on core rename/swapping

mlbiscoc:

OTEL metrics from Meter providers are immutable. On core renaming and swapping, we delete the corresponding registry and recreate it my reinitializing the metric producers with the new core name and labels.

### DIFF
--- a/solr/core/src/java/org/apache/solr/metrics/SolrCoreMetricManager.java
+++ b/solr/core/src/java/org/apache/solr/metrics/SolrCoreMetricManager.java
@@ -23,6 +23,8 @@ import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.common.Attributes;
 import java.io.Closeable;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 import org.apache.solr.cloud.CloudDescriptor;
 import org.apache.solr.common.util.Utils;
 import org.apache.solr.core.CoreContainer;
@@ -41,7 +43,6 @@ public class SolrCoreMetricManager implements Closeable {
   public static final AttributeKey<String> CORE_ATTR = AttributeKey.stringKey("core");
   public static final AttributeKey<String> SHARD_ATTR = AttributeKey.stringKey("shard");
   public static final AttributeKey<String> REPLICA_ATTR = AttributeKey.stringKey("replica");
-  public static final AttributeKey<String> SCOPE_ATTR = AttributeKey.stringKey("scope");
 
   private final SolrCore core;
   private SolrMetricsContext solrMetricsContext;
@@ -51,6 +52,11 @@ public class SolrCoreMetricManager implements Closeable {
   private String replicaName;
   private String leaderRegistryName;
   private boolean cloudMode;
+
+  // Track all metric producers registered for this core so we can re-initialize them during rename
+  private final List<MetricProducerInfo> registeredProducers = new ArrayList<>();
+
+  private record MetricProducerInfo(SolrMetricProducer producer, String scope) {}
 
   /**
    * Constructs a metric manager.
@@ -105,30 +111,35 @@ public class SolrCoreMetricManager implements Closeable {
   }
 
   /**
-   * Make sure that metrics already collected that correspond to the old core name are carried over
-   * and will be used under the new core name. This method also reloads reporters so that they use
-   * the new core name.
+   * Re-register all metric producers associated with this core. This recreates the metric registry
+   * resetting its state
    */
-  // NOCOMMIT SOLR-17458: Update for core renaming
-  public void afterCoreRename() {
-    assert core.getCoreDescriptor().getCloudDescriptor() == null;
-    String oldRegistryName = solrMetricsContext.getRegistryName();
-    String oldLeaderRegistryName = leaderRegistryName;
-    String newRegistryName =
-        createRegistryName(cloudMode, collectionName, shardName, replicaName, core.getName());
-    leaderRegistryName = createLeaderRegistryName(cloudMode, collectionName, shardName);
-    if (oldRegistryName.equals(newRegistryName)) {
-      return;
-    }
-    // close old reporters
-    metricManager.closeReporters(oldRegistryName, solrMetricsContext.getTag());
-    if (oldLeaderRegistryName != null) {
-      metricManager.closeReporters(oldLeaderRegistryName, solrMetricsContext.getTag());
-    }
-    solrMetricsContext =
-        new SolrMetricsContext(metricManager, newRegistryName, solrMetricsContext.getTag());
-    // load reporters again, using the new core name
-    loadReporters();
+  public void reregisterCoreMetrics() {
+    this.solrMetricsContext =
+        new SolrMetricsContext(
+            metricManager,
+            createRegistryName(cloudMode, collectionName, shardName, replicaName, core.getName()),
+            solrMetricsContext.getTag());
+    metricManager.removeRegistry(solrMetricsContext.getRegistryName());
+    if (leaderRegistryName != null) metricManager.removeRegistry(leaderRegistryName);
+
+    var attributesBuilder =
+        Attributes.builder()
+            .put(CORE_ATTR, core.getName())
+            .put(COLLECTION_ATTR, collectionName)
+            .put(SHARD_ATTR, shardName)
+            .put(REPLICA_ATTR, replicaName);
+
+    core.initializeMetrics(solrMetricsContext, attributesBuilder.build(), core.getName());
+
+    registeredProducers.forEach(
+        metricProducer -> {
+          var producerAttributes = attributesBuilder.build().toBuilder();
+          if (metricProducer.scope().startsWith("/"))
+            producerAttributes.put(HANDLER_ATTR, metricProducer.scope);
+          metricProducer.producer.initializeMetrics(
+              solrMetricsContext, producerAttributes.build(), metricProducer.scope);
+        });
   }
 
   /**
@@ -147,11 +158,12 @@ public class SolrCoreMetricManager implements Closeable {
               + producer);
     }
 
-    // NOCOMMIT SOLR-17458: These attributes may not work for standalone mode and maybe make the
-    // scope attribute optional
+    // Track this producer for potential re-initialization during core rename
+    registeredProducers.add(new MetricProducerInfo(producer, scope));
+
     var attributesBuilder =
         Attributes.builder()
-            .put(CORE_ATTR, core.getCoreDescriptor().getName())
+            .put(CORE_ATTR, core.getName())
             .put(COLLECTION_ATTR, collectionName)
             .put(SHARD_ATTR, shardName)
             .put(REPLICA_ATTR, replicaName);
@@ -179,6 +191,8 @@ public class SolrCoreMetricManager implements Closeable {
     }
     metricManager.unregisterGauges(
         solrMetricsContext.getRegistryName(), solrMetricsContext.getTag());
+
+    registeredProducers.clear();
   }
 
   public SolrMetricsContext getSolrMetricsContext() {
--- a/solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
+++ b/solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
@@ -629,14 +629,7 @@ public class SolrMetricManager {
    * @param name registry name
    * @return true if this name points to a registry that already exists, false otherwise
    */
-  // TODO SOLR-17458: We may not need
   public boolean hasRegistry(String name) {
-    Set<String> names = registryNames();
-    name = enforcePrefix(name);
-    return names.contains(name);
-  }
-
-  public boolean hasMeterProvider(String name) {
     return meterProviderAndReaders.containsKey(enforcePrefix(name));
   }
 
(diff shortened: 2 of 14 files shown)


solr/core/src/java/org/apache/solr/metrics/SolrCoreMetricManager.java
```diff
@@ -52,6 +53,11 @@ public class SolrCoreMetricManager implements Closeable {
   private String leaderRegistryName;
   private boolean cloudMode;
 
+  // Track all metric producers registered for this core so we can re-initialize them during rename
+  private final List<MetricProducerInfo> registeredProducers = new ArrayList<>();
```
mlbiscoc, line 58: In Dropwizard, the "core registries" only needed to be renamed on the registries itself which worked because nothing had labels. Not possible in OTEL. We need to track our metric producers so we can re-register and reinitialize its metrics with the correct core name for labels.

solr/core/src/java/org/apache/solr/metrics/SolrCoreMetricManager.java
```diff
@@ -147,11 +158,12 @@ public void registerMetricProducer(String scope, SolrMetricProducer producer) {
               + producer);
     }
 
-    // NOCOMMIT SOLR-17458: These attributes may not work for standalone mode and maybe make the
```
mlbiscoc, line 150: Turns out you try putting a `null` value for a label it just gets ignored so this always worked in standalone mode.

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
```diff
@@ -755,68 +755,15 @@ private static MetricRegistry getOrCreateRegistry(
...
-   * Swap registries. This is useful eg. during {@link org.apache.solr.core.SolrCore} rename or swap
-   * operations. NOTE: this operation is not supported for shared registries.
-   *
-   * @param registry1 source registry
-   * @param registry2 target registry. Note: when used after core rename the target registry doesn't
-   *     exist, so the swap operation will only rename the existing registry without creating an
-   *     empty one under the previous name.
-   */
-  // NOCOMMIT SOLR-17458: Don't think we need
-  public void swapRegistries(String registry1, String registry2) {
```
mlbiscoc, line 793: No such thing as "swapping or renaming registries" just recreating now.

solr/core/src/test/org/apache/solr/metrics/SolrMetricTestUtils.java
```diff
@@ -193,4 +230,86 @@ public static HistogramSnapshot.HistogramDataPointSnapshot getHistogramDatapoint
     return getDatapoint(
         core, metricName, labels, HistogramSnapshot.HistogramDataPointSnapshot.class);
   }
+
+  public static CounterSnapshot.CounterDataPointSnapshot getStandaloneSelectRequestsDatapoint(
```
mlbiscoc: I found myself very often using `select` and `update` requests as easy ways of asserting these registries were getting updated values. Just created these util methods to help. Will probably go back at the other tests that checked these and just call this instead of constantly recreating the same set of labels.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/NodesSysPropsCacher.java
```diff
@@ -74,6 +74,9 @@ public Map<String, Object> getSysProps(String nodeName, Collection<String> tags)
     return result;
   }
 
+  // NOCOMMIT: These properties were fetched from the /admin/metrics endpoint. These properties were
+  // stored as strings instead of numeric values. This is not possible in OTEL metrics. Need to
+  // revisit this later.
   private Map<String, Object> fetchProps(String nodeName, Collection<String> tags) {
```
mlbiscoc, line 80: Not sure if these properties can be retrieved from somewhere else but it cannot be from the /admin/metrics endpoint. Metrics are only numeric values that can be aggregated so it didn't really make sense that the metrics had strings as values. Is there a some endpoint available to get these system props instead?

solr/core/src/java/org/apache/solr/metrics/SolrCoreMetricManager.java
dsmiley: I suggest improving this comment to say "**core** rename".  "rename" alone is ambiguous.

solr/core/src/java/org/apache/solr/metrics/SolrCoreMetricManager.java
```diff
@@ -105,30 +111,35 @@ public void loadReporters() {
...
-    // close old reporters
-    metricManager.closeReporters(oldRegistryName, solrMetricsContext.getTag());
-    if (oldLeaderRegistryName != null) {
-      metricManager.closeReporters(oldLeaderRegistryName, solrMetricsContext.getTag());
-    }
-    solrMetricsContext =
-        new SolrMetricsContext(metricManager, newRegistryName, solrMetricsContext.getTag());
-    // load reporters again, using the new core name
-    loadReporters();
+  public void reregisterCoreMetrics() {
```
dsmiley, line 118: I would have hoped that removing then adding metrics allows us to use the very same code we already have for those separate things.

solr/core/src/test/org/apache/solr/metrics/SolrMetricManagerTest.java
```diff
@@ -64,48 +64,6 @@ public void setUp() throws Exception {
     this.reader = metricManager.getPrometheusMetricReader(METER_PROVIDER_NAME);
   }
 
-  // NOCOMMIT: We might not be supported core swapping in 10. Maybe remove this test
```
dsmiley, line 67: well we need _some_ test for the changes you did

solr/core/src/test/org/apache/solr/metrics/SolrMetricTestUtils.java
```diff
@@ -68,6 +69,42 @@ public static Map<String, Counter> getRandomMetrics(Random random, boolean shoul
...
 
+  /**
+   * Generate random OpenTelemetry metric names for testing Prometheus metrics. Returns a map of
+   * metric names to their expected increment values.
+   */
+  public static Map<String, Long> getRandomPrometheusMetrics(Random random) {
+    return getRandomPrometheusMetrics(random, random.nextBoolean());
+  }
+
+  public static Map<String, Long> getRandomPrometheusMetrics(
```
dsmiley: this looks useless.  Pass a false and it always return null.  Who would explicitly call it as such?  If it only exists for the random case above then inline it (which IntelliJ can do in a single action).

solr/core/src/test/org/apache/solr/metrics/SolrMetricTestUtils.java
```diff
@@ -68,6 +69,42 @@ public static Map<String, Counter> getRandomMetrics(Random random, boolean shoul
...
+      Random random, Map<String, Long> existing) {
+    HashMap<String, Long> metrics = new HashMap<>();
+    ArrayList<String> existingKeys = new ArrayList<>(existing.keySet());
+
+    int numMetrics = TestUtil.nextInt(random, 1, MAX_ITERATIONS);
+    for (int i = 0; i < numMetrics; ++i) {
+      boolean shouldReplaceMetric = !existing.isEmpty() && random.nextBoolean();
+      String name =
+          shouldReplaceMetric
+              ? existingKeys.get(TestUtil.nextInt(random, 0, existingKeys.size() - 1))
```
dsmiley, line 93: if existing is empty; this would probably fail

solr/core/src/test/org/apache/solr/metrics/SolrMetricTestUtils.java
```diff
@@ -68,6 +69,42 @@ public static Map<String, Counter> getRandomMetrics(Random random, boolean shoul
...
+              : TestUtil.randomSimpleString(random, 5, 10)
+                  + SUFFIX; // must be simple string for JMX publishing
+
+      Long incrementValue = Math.abs(random.nextLong() % 1000) + 1;
```
dsmiley: TextUtil.nextLong

solr/core/src/test/org/apache/solr/metrics/SolrMetricTestUtils.java
```diff
@@ -68,6 +69,42 @@ public static Map<String, Counter> getRandomMetrics(Random random, boolean shoul
...
+      Random random, boolean shouldDefineMetrics) {
+    return shouldDefineMetrics
+        ? getRandomPrometheusMetricsWithReplacements(random, new HashMap<>())
+        : null;
+  }
+
+  public static Map<String, Long> getRandomPrometheusMetricsWithReplacements(
+      Random random, Map<String, Long> existing) {
+    HashMap<String, Long> metrics = new HashMap<>();
+    ArrayList<String> existingKeys = new ArrayList<>(existing.keySet());
```
dsmiley: I suggest `List.copyOf(...)` here

solr/core/src/test/org/apache/solr/metrics/SolrMetricTestUtils.java
dsmiley: I'd use `new` in place of `get` as applicable.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/NodesSysPropsCacher.java
dsmiley: This is unfortunate.  CC @sigram 

For system properties in particular, http://localhost:8983/solr/admin/info/properties which you probably recall visualized from Solr's admin UI.  This change (with remedy) ought to be mentioned in the upgrade guide so people know.

dsmiley (commented):
Lots of great work here!

This unfortunate change of behavior should be mentioned in the upgrade guide `major-changes-in-solr-10.adoc`.  There should be a dedicated OTEL section there; we'll have lots to say.

By the way, there were nasty/insiduous memory leaks relating to metrics for cores that closed and/or core swapping.  They were fixed; you can find them in JIRA.  But as our metrics infrastructure is redone, the risk resurfaces.  I suppose we should create a JIRA issue for some sort of assessment that it hasn't re-occurred.  I'm unlikely to identify the risks as a code reviewer.  One aspect, if I recall, were gauge lambdas that had references to the core or searcher (e.g. to return a disk size) but the gauge didn't actually get cleaned  up.

solr/core/src/test/org/apache/solr/metrics/SolrMetricManagerTest.java
mlbiscoc: SolrMetricManager doesn't have a public swap registries method anymore to unit test. I tested the changes at a higher level you can see in `TestCoreAdmin` which does an initial metrics check -> swap and check registries reset to 0 -> select and check the registries are still updating according to correct cores.

solr/core/src/test/org/apache/solr/metrics/SolrMetricTestUtils.java
mlbiscoc: `!existing.isEmpty()` is checked right above so it shouldn't fail.

solr/core/src/test/org/apache/solr/metrics/SolrMetricTestUtils.java
mlbiscoc: Did you mean TestUtil.nextLong? Updated to that.

solr/core/src/java/org/apache/solr/metrics/SolrCoreMetricManager.java
mlbiscoc: We can't remove metrics from an existing registry after creation so we have to just completely tear it down and recreate it.

mlbiscoc:
> there were nasty/insiduous memory leaks relating to metrics for cores that closed and/or core swapping

Will make a Jira to double check this. I'll take a look at the previous jira that found this and what happened. Maybe will do some JFR testing to confirm and investigate.

solr/core/src/java/org/apache/solr/metrics/SolrCoreMetricManager.java
dsmiley: (pending)

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/NodesSysPropsCacher.java
dsmiley: Please insert this recommendation into the TODO comment, then we move on.

solr/core/src/java/org/apache/solr/metrics/SolrCoreMetricManager.java
```diff
@@ -105,30 +112,35 @@ public void loadReporters() {
...
+    this.solrMetricsContext =
+        new SolrMetricsContext(
+            metricManager,
+            createRegistryName(cloudMode, collectionName, shardName, replicaName, core.getName()),
+            solrMetricsContext.getTag());
+    metricManager.removeRegistry(solrMetricsContext.getRegistryName());
+    if (leaderRegistryName != null) metricManager.removeRegistry(leaderRegistryName);
+
+    var attributesBuilder =
```
dsmiley: you should just build this; have it be the attributes itself.  You're building in a loop below but it's redundant.

solr/core/src/java/org/apache/solr/metrics/SolrCoreMetricManager.java
```diff
@@ -105,30 +112,35 @@ public void loadReporters() {
...
+        Attributes.builder()
+            .put(CORE_ATTR, core.getName())
```
dsmiley, line 133: my point about the duplication, for example, can be seen in the attributes construction.  This is duplicated over in `registerMetricProcducer`.  But it's not a lot of duplication.  It could be ameliorated with comments on both sides.

mlbiscoc:
Anything else here? I would like to merge this because it makes some changes to Solr `SolrMetricsIntegrationTest` that was a bad apple and helps unblock `CoreContainer` [metrics migration and its tests](https://github.com/apache/solr/pull/3509).

dsmiley (approved):
I'm approving but I'd prefer if the duplication was minimally addressed with a comment on both sides of the duplication to help us keep in sync.

PR #3466, opened 2025-08-11
SOLR-17806: Use OtelUnit to create Otel insturments with units as metadata

mlbiscoc:

Created an `OtelUnit` as an enum. Otel instruments can be created with a unit type as metadata but only works with specific strings that follow [UCUM](https://ucum.org/ucum). This wasn't obvious as passing in a string to created an OtelUnit enum the constrains unit types to UCUM instead of constantly trying to pass strings.


dsmiley (approved):
Nice simple PR

PR #3469, opened 2025-08-12
SOLR-14393: Remove HIGHLIGHTER, QUERYPARSER, SPELLCHECKER metrics

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-14393

Clean up HIGHLIGHTER, QUERYPARSER, SPELLCHECKER metrics that were initialized from SolrInfoBean.


solr/core/src/java/org/apache/solr/handler/component/HighlightComponent.java
```diff
@@ -240,20 +240,11 @@ public void finishStage(ResponseBuilder rb) {
     }
   }
 
-  ////////////////////////////////////////////
-  ///  SolrInfoBean
```
dsmiley, line 244: fine; we generally don't use such method organizational headings.  It may be is only useful in larger classes... granted growth happens organically.

solr/core/src/java/org/apache/solr/handler/component/SpellCheckComponent.java
```diff
@@ -875,17 +875,7 @@ public Map<String, SolrSpellChecker> getSpellCheckers() {
     return Collections.unmodifiableMap(spellCheckers);
   }
 
-  // ///////////////////////////////////////////
-  // / SolrInfoBean
-  // //////////////////////////////////////////
-
-  @Override
   public String getDescription() {
```
dsmiley, line 879: can getDescription be removed if we don't implement SolrInfoBean?

dsmiley (approved):
Nice work

solr/core/src/java/org/apache/solr/handler/component/SpellCheckComponent.java
mlbiscoc: Not for some of these because they implement SearchComponent which has getDescription as an abstract method and does implement SolrInfoBean. These are used for `SolrInfoMBeanHandler` which I still have yet to address on this jira [SOLR-17853](https://issues.apache.org/jira/browse/SOLR-17853). Maybe we can remove it if we figure out what to do there.

dsmiley: What caught my attention is that you removed the `@Override`, which we mandate by our "precommit" checks.  Thus I was led to believe SpellCheckComponent no longer overrides this method.  So I wondered why is it staying around.

So I suspect you haven't been running precommit on all these PRs.  Try `./gradlew precommit -Pvalidation.sourcePatterns.failOnError=false`

mlbiscoc: Yeah I haven't really. Luckily there aren't many precommit errors besides the `NOCOMMIT` comments i've thrown around. I'll make a small PR to address these and then do precommit going forward.

PR #3471, opened 2025-08-12
SOLR-17860: Support PULL replicas in DocBasedVersionConstraintsProcessor

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17860

Only leader eligible replicas need to have an update log in Cloud mode, since only leader cores will do the logic in the URP. In non-cloud mode we will keep the same logic as before.


gerlowskija (commented):
The general fix looks reasonable to me, though ideally we'd add some tests for this as well.

PR #3476, opened 2025-08-14
Custom embedding model

renatoh:


renatoh:
Hi @alessandrobenedetti 
I am curious, if you think any of these customization could have the potential to be merged. It is all pretty rough at the moment and some refinement had to be done.
I have added:
- Using custom endpoint for vectorization
- Doing lazy vectorization by comparing content of new fields with the existing value
- Combining multiple fields into one vector
Thanks for your input!

solr/modules/llm/src/java/org/apache/solr/llm/textvectorisation/model/CustomModel.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.apache.http.HttpEntity;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.entity.StringEntity;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.util.EntityUtils;
```
dsmiley: No; let's please not use Apache HttpClient
https://issues.apache.org/jira/browse/SOLR-16367

solr/modules/llm/src/java/org/apache/solr/llm/textvectorisation/update/processor/TextToVectorUpdateProcessorFactory.java
```diff
@@ -70,13 +70,11 @@ public void init(final NamedList<?> args) {
...
+    if (!latestSchema.hasExplicitField(inputField)) {
       throw new SolrException(
           SolrException.ErrorCode.SERVER_ERROR, "undefined field: \"" + inputField + "\"");
     }
-
-    if (!latestSchema.isDynamicField(outputField) && !latestSchema.hasExplicitField(outputField)) {
+    if (!latestSchema.hasExplicitField(outputField)) {
       throw new SolrException(
           SolrException.ErrorCode.SERVER_ERROR, "undefined field: \"" + outputField + "\"");
     }
```
dsmiley: I don't know why these checks are here _at all_.  `org.apache.solr.schema.IndexSchema#getField` (called immediately below) is going to throw an exception with a specific error message about the field being undefined.

dsmiley (commented):
Is there a standard / convention API for doing an embedding over HTTP?

solr/modules/llm/src/java/org/apache/solr/llm/textvectorisation/model/CustomModel.java
dsmiley: I recommend the HttpClient in the JDK instead.  Modern Java.

renatoh:
> Is there a standard / convention API for doing an embedding over HTTP?

i don't think there is. with that approach endpoint and input-text field can be configured in the model and it expects an array of floats back.

renatoh:
> Is there a standard / convention API for doing an embedding over HTTP?
@dsmiley 
I don't think there is, as far as I understand that is the reason why Solr uses langchain4j. This lib has integration to the different provides like huggingface, cohere,etc.

renatoh:
@epugh Hi Eric,
That is the PR I've mention, basically It does this:

- Using custom endpoint for vector embedding
- Doing lazy vectorization by comparing content of new fields with the existing value
- Combining multiple fields into one vector

Would be good to get your feedback on that one.

Thanks
Renato

PR #3477, opened 2025-08-14
Solr 17863 fix per segment fingerprint cache race

kotman12:
---------

https://issues.apache.org/jira/browse/SOLR-17863

# Description

[SOLR-16713](https://issues.apache.org/jira/browse/SOLR-16713) introduced a race condition into the perSegmentFingerprintCache as it changed it from the [thread-safe guava implementation to the thread-UNSAFE WeakHashMap](https://github.com/apache/solr/commit/375ac647ad507638686848c549c30e6c077dca1b#diff-3bf44f923643744a743460bb0f64301618dd7529a602d945fe2f7f193d4cdde0). This flew under the radar for some time because this map previously had very little contention between threads. However after [SOLR-17756](https://issues.apache.org/jira/browse/SOLR-17756) we turbo-charged the race after parallelizing at a much more granular level upstream of this call. The net effect is that WeakHashMap:put can get stuck in an infinite-loop (probably due to a cycle in the underlying linked-list that it modifies):

# Solution

# Tests

It is hard to recreate this reliably. I've been able to reproduce by increasing the amount of segments written by `SolrIndexFingerprintTest` to ~500 although this does take some time. 

# Checklist

- [X ] I have added tests for my changes.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -21,6 +21,8 @@ import static org.apache.solr.handler.admin.MetricsHandler.PROMETHEUS_METRICS_WT
 
 import com.codahale.metrics.Counter;
 import com.codahale.metrics.Timer;
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 import java.io.Closeable;
 import java.io.FileNotFoundException;
 import java.io.IOException;
@@ -54,7 +56,6 @@ import java.util.Set;
 import java.util.Spliterator;
 import java.util.Spliterators;
 import java.util.UUID;
-import java.util.WeakHashMap;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
@@ -66,9 +67,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
 import java.util.stream.StreamSupport;
-
-import com.github.benmanes.caffeine.cache.Cache;
-import com.github.benmanes.caffeine.cache.Caffeine;
 import org.apache.commons.io.file.PathUtils;
 import org.apache.lucene.codecs.Codec;
 import org.apache.lucene.index.DirectoryReader;
@@ -274,9 +272,7 @@ public class SolrCore implements SolrInfoBean, Closeable {
   }
 
   private final Cache<IndexReader.CacheKey, IndexFingerprint> perSegmentFingerprintCache =
-      Caffeine.newBuilder()
-          .weakKeys()
-          .build();
+      Caffeine.newBuilder().weakKeys().build();
 
   public long getStartNanoTime() {
     return startNanoTime;
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -66,6 +66,9 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
 import java.util.stream.StreamSupport;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 import org.apache.commons.io.file.PathUtils;
 import org.apache.lucene.codecs.Codec;
 import org.apache.lucene.index.DirectoryReader;
@@ -270,8 +273,10 @@ public class SolrCore implements SolrInfoBean, Closeable {
     return startTime;
   }
 
-  private final Map<IndexReader.CacheKey, IndexFingerprint> perSegmentFingerprintCache =
-      new WeakHashMap<>();
+  private final Cache<IndexReader.CacheKey, IndexFingerprint> perSegmentFingerprintCache =
+      Caffeine.newBuilder()
+          .weakKeys()
+          .build();
 
   public long getStartNanoTime() {
     return startNanoTime;
@@ -2157,7 +2162,7 @@ public class SolrCore implements SolrInfoBean, Closeable {
     }
 
     IndexFingerprint f = null;
-    f = perSegmentFingerprintCache.get(cacheHelper.getKey());
+    f = perSegmentFingerprintCache.getIfPresent(cacheHelper.getKey());
     // fingerprint is either not cached or if we want fingerprint only up to a version less than
     // maxVersionEncountered in the segment, or documents were deleted from segment for which
     // fingerprint was cached
@@ -2197,8 +2202,8 @@ public class SolrCore implements SolrInfoBean, Closeable {
     }
     if (log.isDebugEnabled()) {
       log.debug(
-          "Cache Size: {}, Segments Size:{}",
-          perSegmentFingerprintCache.size(),
+          "Approximate perSegmentFingerprintCache Size: {}, Segments Size:{}",
+          perSegmentFingerprintCache.estimatedSize(),
           searcher.getTopReaderContext().leaves().size());
     }
     return f;
--- a/solr/core/src/test-files/solr/collection1/conf/solrconfig-tiny-segments.xml
+++ b/solr/core/src/test-files/solr/collection1/conf/solrconfig-tiny-segments.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" ?>
+
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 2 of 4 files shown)


kotman12:
A general question I have is how can I make more than 1 core available to the test? When I run in intellij `Runtime.getRuntime().availableProcessors()` resolves to `1` despite my machine having more than 1 cpu available. I do see `-XX:ActiveProcessorCount=1` as a suggested setting in `gradle.properties` but this doesn't seem to do anything (same can be said about the apparent ActiveProcessorCount override in `turbocharge-jvm-opts.gradle`). I assume when this runs on the build server EXECUTOR_MAX_CPU_THREADS may resolve to values greater than 1?

solr/core/src/test/org/apache/solr/update/SolrIndexFingerprintTest.java
```diff
@@ -28,29 +29,21 @@ public class SolrIndexFingerprintTest extends SolrTestCaseJ4 {
...
-    assertU(adoc("id", "107"));
-    assertU(adoc("id", "108"));
-    assertU(adoc("id", "109"));
-    assertU(commit());
+    // Create a set of 500 segments (to catch race conditions, i.e.SOLR-17863)
+    IntStream.range(0, 500)
+        .forEach(
+            i -> {
+              assertU(adoc("id", "" + i));
+              assertU(commit());
```
risdenk, line 46: Do you have to commit after every document?

kotman12: to reliably make 500 segments .. I think so? To tickle this race condition you need lots of leaves. Is there a better way?

kotman12: (or maybe it's not worth the test)

risdenk: Thats fine just didn't see the comment originally. All good.

risdenk (approved):

solr/core/src/test/org/apache/solr/update/SolrIndexFingerprintTest.java
```diff
@@ -28,29 +29,21 @@ public class SolrIndexFingerprintTest extends SolrTestCaseJ4 {
...
-    assertU(adoc("id", "104"));
-    assertU(adoc("id", "105"));
-    assertU(adoc("id", "106"));
-    assertU(commit());
-
-    assertU(adoc("id", "107"));
-    assertU(adoc("id", "108"));
-    assertU(adoc("id", "109"));
-    assertU(commit());
+    // Create a set of 500 segments (to catch race conditions, i.e.SOLR-17863)
```
dsmiley: Let's not repeat a constant in the code with your comment so that we don't have to synchronize them.  I suggest the word "many". 

In this situation, I recommend looking to see if org.apache.lucene.tests.util.LuceneTestCase.RANDOM_MULTIPLIER is 1 and if so return 3 but otherwise return 500.  Thus a nightly test run pushes it harder.  But maybe 500 is fast?  How long does this test take at 500?

solr/core/src/test-files/solr/collection1/conf/solrconfig-tiny-segments.xml
```diff
@@ -0,0 +1,32 @@
...
+
+<config>
+  <luceneMatchVersion>${tests.luceneMatchVersion:LATEST}</luceneMatchVersion>
+  <directoryFactory name="DirectoryFactory" class="${solr.directoryFactory:solr.MockDirectoryFactory}"/>
+  <schemaFactory class="ClassicIndexSchemaFactory"/>
+
+  <indexConfig>
+    <useCompoundFile>${useCompoundFile:false}</useCompoundFile>
+    <mergePolicyFactory class="org.apache.solr.index.TieredMergePolicyFactory">
+      <double name="maxMergedSegmentMB">0.0001</double>
```
dsmiley: if your goal is lots of segments, the textbook Lucene way to do this is NoMergePolicy.  It explicitly conveys the goal whereas this configuration is less apparent as to the intent.

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -270,8 +271,8 @@ public Date getStartTimeStamp() {
     return startTime;
   }
 
-  private final Map<IndexReader.CacheKey, IndexFingerprint> perSegmentFingerprintCache =
-      new WeakHashMap<>();
+  private final Cache<IndexReader.CacheKey, IndexFingerprint> perSegmentFingerprintCache =
```
dsmiley: I know this is strictly unrelated but I **hate** that we have a field defined here, shoved inbetween two methods (whose methods are slightly related and neither have *anything* to do with this field).  Sloppy as hell; it grates at my sensibilities.  I don't even like it in this class TBH but whatever.  Can you please move it?  Maybe below ConfigSet.

dsmiley (changes requested):

kotman12:
> A general question I have is how can I make more than 1 core available to the test? When I run in intellij `Runtime.getRuntime().availableProcessors()` resolves to `1` despite my machine having more than 1 cpu available. I do see `-XX:ActiveProcessorCount=1` as a suggested setting in `gradle.properties` but this doesn't seem to do anything (same can be said about the apparent ActiveProcessorCount override in `turbocharge-jvm-opts.gradle`). I assume when this runs on the build server EXECUTOR_MAX_CPU_THREADS may resolve to values greater than 1?

For posterity, the way you override this (as of the writing of this comment at least) is by updating `tests.jvmargs` to some `-XX:ActiveProcessorCount=N` in the `defaults-tests.gradle` file. I must have overlooked this in my first string search of this property.

solr/core/src/test/org/apache/solr/update/SolrIndexFingerprintTest.java
kotman12: It takes ~20s on my machine with 500 segments. I think nightly is appropriate and have implemented your initial suggestion. I've also addressed the other comments, please have another look when you get a chance 😄

solr/core/src/test/org/apache/solr/update/SolrIndexFingerprintTest.java
```diff
@@ -29,16 +29,17 @@ public class SolrIndexFingerprintTest extends SolrTestCaseJ4 {
 
   @BeforeClass
   public static void beforeTests() throws Exception {
-    initCore("solrconfig-tiny-segments.xml", "schema.xml");
+    initCore("solrconfig-nomergepolicyfactory.xml", "schema.xml");
```
dsmiley, line 32: fantastic to use an existing one :-)

iamsanjay:
@kotman12 I believe this is related
```
Reproduce with: ./gradlew :solr:core:test --tests "org.apache.solr.update.SolrIndexFingerprintTest.testSequentialVsParallelFingerprint" -Ptests.haltonfailure=false "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 -XX:ReservedCodeCacheSize=120m" -Ptests.seed=937823B9A5449C7B -Ptests.multiplier=2 -Ptests.useSecurityManager=true -Ptests.badapples=false -Ptests.file.encoding=UTF-8
```

```
Caused by: java.nio.file.FileSystemException: /home/jenkins/jenkins-agent/workspace/Solr/Solr-Test-main/build/tmp/tests-tmp/org.apache.solr.update.SolrIndexFingerprintTest_937823B9A5449C7B-001/index-NIOFSDirectory-001/_5o.fdx: Too many open files
        at org.apache.lucene.tests.mockfile.HandleLimitFS.onOpen(HandleLimitFS.java:67)
        at org.apache.lucene.tests.mockfile.HandleTrackingFS.callOpenHook(HandleTrackingFS.java:82)
        at org.apache.lucene.tests.mockfile.HandleTrackingFS.newFileChannel(HandleTrackingFS.java:202)
        at org.apache.lucene.tests.mockfile.HandleTrackingFS.newFileChannel(HandleTrackingFS.java:171)
        at java.base/java.nio.channels.FileChannel.open(FileChannel.java:309)
        at java.base/java.nio.channels.FileChannel.open(FileChannel.java:369)
```

kotman12:
@iamsanjay hmm .. appears so .. but there is a good reason there are so many segments. I wonder if this could be mitigated either by decreasing the number of segments (which unfortunately decreases the efficacy of the test) or by using a `RAMDirectory`? WDYT?

EDIT:

OTOH I find it a bit suspicious that these segment files aren't getting closed 🤔  .. Yea it's true this test spins up 500 segments but the fingerprint calculation only needs a handle to MAX_PARALLELISM number of files at any given time (at least in theory). This max parallelism should be bounded at cpu cores so maybe this test is flagging a real problem? 500 files isn't really that crazy for an index to have so I am not sure how much of this is the environment this test is running (constrained limits and/or lots of other tests running in parallel with handles to files) vs potentially a real resource leak? Will have to give it  another look..

PR #3482, opened 2025-08-18
SOLR-17806: Migrate SolrIndexSearcher metrics to OTEL

iamsanjay:

```
solr_core_registered_at_ms{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",shard="shard1"} 1.755533626708E12
solr_core_warmup_time_ms{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",shard="shard1"} 0.0
solr_searcher_deleted_docs{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",shard="shard1"} 0.0
solr_searcher_full_sort_count{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",shard="shard1"} 0.0
solr_searcher_index_commit_size_bytes{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",shard="shard1"} 9162.0
solr_searcher_index_version{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",shard="shard1"} 11.0
solr_searcher_live_docs_cache{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",shard="shard1",type="hits"} 0.0
solr_searcher_live_docs_cache{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",shard="shard1",type="inserts"} 0.0
solr_searcher_live_docs_cache{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",shard="shard1",type="naive_hits"} 0.0
solr_searcher_max_doc{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",shard="shard1"} 3.0
solr_searcher_name{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",shard="shard1"} 1.0
solr_searcher_num_docs{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",shard="shard1"} 3.0
solr_searcher_reader{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",reader="UninvertingDirectoryReader()",shard="shard1"} 1.0
solr_searcher_reader{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",reader="UninvertingDirectoryReader(_0(9.12.1):C1:[diagnostics={lucene.version=9.12.1, os.version=14.3, os.arch=aarch64, java.vendor=Oracle Corporation, os=Mac OS X, java.runtime.version=21.0.3+7-LTS-152, timestamp=1755533590814, source=flush}]:[attributes={Lucene90StoredFieldsFormat.mode=BEST_SPEED}] :id=d2wucl0ltql7gbnhuq5a5czzm _1(9.12.1):C1:[diagnostics={lucene.version=9.12.1, os.version=14.3, os.arch=aarch64, java.vendor=Oracle Corporation, os=Mac OS X, java.runtime.version=21.0.3+7-LTS-152, timestamp=1755533612458, source=flush}]:[attributes={Lucene90StoredFieldsFormat.mode=BEST_SPEED}] :id=d2wucl0ltql7gbnhuq5a5czzq _2(9.12.1):C1:[diagnostics={lucene.version=9.12.1, os.version=14.3, os.arch=aarch64, java.vendor=Oracle Corporation, os=Mac OS X, java.runtime.version=21.0.3+7-LTS-152, timestamp=1755533615759, source=flush}]:[attributes={Lucene90StoredFieldsFormat.mode=BEST_SPEED}] :id=d2wucl0ltql7gbnhuq5a5czzt)",shard="shard1"} 1.0
solr_searcher_reader{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",reader="UninvertingDirectoryReader(_0(9.12.1):C1:[diagnostics={lucene.version=9.12.1, os.version=14.3, os.arch=aarch64, os=Mac OS X, java.vendor=Oracle Corporation, java.runtime.version=21.0.3+7-LTS-152, timestamp=1755533590814, source=flush}]:[attributes={Lucene90StoredFieldsFormat.mode=BEST_SPEED}] :id=d2wucl0ltql7gbnhuq5a5czzm _1(9.12.1):C1:[diagnostics={lucene.version=9.12.1, os.version=14.3, os.arch=aarch64, os=Mac OS X, java.vendor=Oracle Corporation, java.runtime.version=21.0.3+7-LTS-152, timestamp=1755533612458, source=flush}]:[attributes={Lucene90StoredFieldsFormat.mode=BEST_SPEED}] :id=d2wucl0ltql7gbnhuq5a5czzq _2(9.12.1):C1:[diagnostics={lucene.version=9.12.1, os.version=14.3, os.arch=aarch64, os=Mac OS X, java.vendor=Oracle Corporation, java.runtime.version=21.0.3+7-LTS-152, timestamp=1755533615759, source=flush}]:[attributes={Lucene90StoredFieldsFormat.mode=BEST_SPEED}] :id=d2wucl0ltql7gbnhuq5a5czzt)",shard="shard1"} 1.0
solr_searcher_reader{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",reader="UninvertingDirectoryReader(_0(9.12.1):C1:[diagnostics={lucene.version=9.12.1, os.version=14.3, os.arch=aarch64, os=Mac OS X, java.vendor=Oracle Corporation, java.runtime.version=21.0.3+7-LTS-152, timestamp=1755533590814, source=flush}]:[attributes={Lucene90StoredFieldsFormat.mode=BEST_SPEED}] :id=d2wucl0ltql7gbnhuq5a5czzm _1(9.12.1):C1:[diagnostics={lucene.version=9.12.1, os.version=14.3, os.arch=aarch64, os=Mac OS X, java.vendor=Oracle Corporation, java.runtime.version=21.0.3+7-LTS-152, timestamp=1755533612458, source=flush}]:[attributes={Lucene90StoredFieldsFormat.mode=BEST_SPEED}] :id=d2wucl0ltql7gbnhuq5a5czzq)",shard="shard1"} 1.0
solr_searcher_reader{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",reader="UninvertingDirectoryReader(_0(9.12.1):C1:[diagnostics={lucene.version=9.12.1, os.version=14.3, os.arch=aarch64, os=Mac OS X, java.vendor=Oracle Corporation, java.runtime.version=21.0.3+7-LTS-152, timestamp=1755533590814, source=flush}]:[attributes={Lucene90StoredFieldsFormat.mode=BEST_SPEED}] :id=d2wucl0ltql7gbnhuq5a5czzm)",shard="shard1"} 1.0
solr_searcher_reader_dir{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",reader_dir="NRTCachingDirectory(MMapDirectory@/Users/sanjaydutt/Documents/solr/solr/packaging/build/solr-10.0.0-SNAPSHOT/server/solr/demo_shard1_replica_n1/data/index lockFactory=org.apache.lucene.store.NativeFSLockFactory@3b6705cc; maxCacheMB=48.0 maxMergeSizeMB=4.0)",shard="shard1"} 1.0
solr_searcher_skip_sort_count{category="SEARCHER",collection="demo",core="demo_shard1_replica_n1",otel_scope_name="org.apache.solr",shard="shard1"} 0.0
```


solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2592,83 +2607,145 @@ public SolrMetricsContext getSolrMetricsContext() {
...
-    parentContext.gauge(
-        rgauge(parentContext.nullNumber(), () -> reader.getVersion()),
-        true,
-        "indexVersion",
-        Category.SEARCHER.toString(),
-        scope);
+      SolrMetricsContext solrMetricsContext, Attributes attributes, String scope) {
+    var baseAttributes =
+        attributes.toBuilder()
+            .remove(AttributeKey.stringKey("scope"))
```
mlbiscoc: You shouldn't have to do this remove anymore

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2592,83 +2607,145 @@ public SolrMetricsContext getSolrMetricsContext() {
   // TODO SOLR-17458: Migrate to Otel
```
mlbiscoc: Remove this TODO

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2592,83 +2607,145 @@ public SolrMetricsContext getSolrMetricsContext() {
...
-        Category.SEARCHER.toString(),
-        scope);
+      SolrMetricsContext solrMetricsContext, Attributes attributes, String scope) {
+    var baseAttributes =
+        attributes.toBuilder()
+            .remove(AttributeKey.stringKey("scope"))
+            .put(CATEGORY_ATTR, Category.SEARCHER.toString())
+            .build();
+    solrMetricsContext.observableLongGauge(
+        "solr_searcher_name", "Searcher name (as attribute)", obs -> obs.record(1, baseAttributes));
```
mlbiscoc: Don't create metrics where value is in an attribute. The gauge then has a value of 1 that doesn't make any sense. Metrics are only numerics that can be aggregated. If the metric was a string as the value, just drop it completely and we need to figure out where this information should be instead. Also if you put unique names in the attributes, you blow up cardinality.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2592,83 +2607,145 @@ public SolrMetricsContext getSolrMetricsContext() {
...
+    solrMetricsContext.observableLongGauge(
+        "solr_searcher_caching_enabled",
+        "1 if caching enabled",
+        obs -> obs.record(cachingEnabled ? 1 : 0, baseAttributes));
+
+    // warmupTime (ms)
+    solrMetricsContext.observableLongGauge(
+        "solr_searcher_warmup_time_ms",
+        "Searcher warmup time (ms)",
+        obs -> obs.record(warmupTime, baseAttributes));
```
mlbiscoc: For OTEL, you put the unit as metadata. Pass `OtelUnit.MILLISECONDS` and remove `_ms` from the name. Prometheus metric reader will append milliseconds in the name itself as a way to follow prometheus format.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2592,83 +2607,145 @@ public SolrMetricsContext getSolrMetricsContext() {
...
+    solrMetricsContext.observableLongGauge(
+        "solr_searcher_registered_at_ms",
+        "Searcher registered time (ms)",
+        obs -> obs.record(registerTime.getTime(), baseAttributes));
+
+    // fullSortCount
+    solrMetricsContext.observableLongGauge(
+        "solr_searcher_full_sort_count",
+        "Total full sorts",
+        obs -> obs.record(fullSortCount.sum(), baseAttributes));
```
mlbiscoc: For these metrics, don't use a gauge. We can see that fullSortCount is basically a counter backed by LongAdder. We should either switch LongAdder and create an `AttributedLongCounter` or use an `observableLongCounter` and callback on `fullSortCount.sum()`. Remove `_count` from the name as counter type implies that already and prometheus format will append `_total` as a suffix.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2592,83 +2607,145 @@ public SolrMetricsContext getSolrMetricsContext() {
...
+    // fullSortCount
+    solrMetricsContext.observableLongGauge(
```
mlbiscoc: Did you accidentally dupe this?

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2592,83 +2607,145 @@ public SolrMetricsContext getSolrMetricsContext() {
...
+        "solr_searcher_full_sort_count",
+        "Total full sorts",
```
mlbiscoc: Can we have better descriptions of this and possibly others? "Total full sorts" alone doesn't tell use much. I am assuming this metric is counting the total number of full sorts it applied on a result set from a searchers doc list?

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2592,83 +2607,145 @@ public SolrMetricsContext getSolrMetricsContext() {
...
+    // reader identity (strings → attributes on a dummy value 1)
+    solrMetricsContext.observableLongGauge(
+        "solr_searcher_reader",
+        "Reader identity (as attribute)",
+        obs -> {
+          try {
+            obs.record(
+                1,
+                baseAttributes.toBuilder()
+                    .put(AttributeKey.stringKey("reader"), String.valueOf(reader))
```
mlbiscoc: Remove this reader attribute. Its not useful as an attribute and you can't really aggregate on something so large and unique of a string value of an object.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2592,83 +2607,145 @@ public SolrMetricsContext getSolrMetricsContext() {
...
           }
-        },
-        true,
-        "indexCommitSize",
-        Category.SEARCHER.toString(),
-        scope);
+        });
     // statsCache metrics
-    parentContext.gauge(
+    solrMetricsContext.gauge(
```
mlbiscoc: We should migrate these metrics as well.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2592,83 +2607,145 @@ public SolrMetricsContext getSolrMetricsContext() {
...
+        "solr_searcher_skip_sort_count",
+        "Total skip sorts",
+        obs -> obs.record(skipSortCount.sum(), baseAttributes));
+    solrMetricsContext.observableLongGauge(
+        "solr_searcher_live_docs_cache",
+        "LiveDocs cache metrics",
+        obs -> {
+          obs.record(
+              liveDocsInsertsCount.sum(),
+              baseAttributes.toBuilder().put("type", "inserts").build());
```
mlbiscoc: Use `TYPE_ATTR`. I don't think you can put in a "type" key just as a string.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2589,95 +2602,145 @@ public SolrMetricsContext getSolrMetricsContext() {
...
-        new MetricsMap(
-            map -> {
-              statsCache.getCacheMetrics().getSnapshot(map::putNoEx);
-              map.put("statsCacheImpl", statsCache.getClass().getSimpleName());
-            }),
-        true,
-        "statsCache",
-        Category.CACHE.toString(),
-        scope);
+    solrMetricsContext.observableLongGauge(
```
mlbiscoc: Looks like these LongAdder values can reset to 0. Maybe worth noting that in the description when these go to 0 or how.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2589,95 +2602,145 @@ public SolrMetricsContext getSolrMetricsContext() {
...
+        "solr_searcher_stats_cache",
+        "Operation counts for the searcher stats cache, reported per operation type",
+        obs -> {
+          var cacheMetrics = statsCache.getCacheMetrics();
+          obs.record(
+              cacheMetrics.lookups.sum(),
+              cacheBaseAttribute.toBuilder().put(TYPE_ATTR, "lookups").build());
+          obs.record(
+              cacheMetrics.missingGlobalFieldStats.sum(),
+              cacheBaseAttribute.toBuilder().put(TYPE_ATTR, "missing_global_field_stats").build());
```
mlbiscoc: You can drop `_stats` suffix from this attribute name.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2589,95 +2602,145 @@ public SolrMetricsContext getSolrMetricsContext() {
...
-        Category.SEARCHER.toString(),
-        scope);
+      SolrMetricsContext solrMetricsContext, Attributes attributes, String scope) {
+    var baseAttributes =
+        attributes.toBuilder().put(CATEGORY_ATTR, Category.SEARCHER.toString()).build();
+    // caching (boolean -> 0/1)
+    solrMetricsContext.observableLongGauge(
+        "solr_searcher_caching_enabled",
+        "1 if caching enabled",
+        obs -> obs.record(cachingEnabled ? 1 : 0, baseAttributes));
```
dsmiley: As this is effectively constant (cachingEnabled doesn't change for this searcher), do we actually need a gauge at all vs a counter that's initialized correctly?  I view gauges as slightly more heavyweight.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2589,95 +2602,145 @@ public SolrMetricsContext getSolrMetricsContext() {
...
+        "solr_searcher_warmup_time",
+        "Searcher warmup time (ms)",
+        obs -> obs.record(warmupTime, baseAttributes),
+        OtelUnit.MILLISECONDS);
+    // registeredAt (ms)
+    solrMetricsContext.observableLongGauge(
+        "solr_searcher_registered_at",
+        "Searcher registered time (ms)",
+        obs -> obs.record(registerTime.getTime(), baseAttributes),
+        OtelUnit.MILLISECONDS);
```
dsmiley: @mlbiscoc not sure if there's a better way to communicate an instant in time that something happened.  A timestamp, basically.  OtelUnit.MILLISECONDS, I suspect, is intended to convey a duration.

mlbiscoc:
> The SolrIndexSearcher is created often; on every commit. Will reregistering metrics simply overwrite the previously registered metrics without issue?

Thats a good question that I haven't actually ran into... But observable instruments are the only OTEL instruments that are `closable` without having to teardown the registry. So maybe what we should do instead is keep a reference of these then in the `close()` method, teardown and close all these observable instruments when the searcher closes to make sure they are closed instead of betting on an overwrite.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
mlbiscoc: The idea of `OtelUnit` was supposed to be the unit of the value but for timestamps and instant its unique. I found this [interesting discussion](https://github.com/open-telemetry/opentelemetry-specification/discussions/4342#discussioncomment-11634776) when trying to google this question. Basically they recommend that as a metric, don't record timestamps and use other signals like logs to record a timestamp. So maybe this metric shouldn't exist.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
mlbiscoc: Observable instruments are slightly more heavy weight. If these never change, then you are correct we can just change it to a normal gauge where `cachingEnabled` is set once.

mlbiscoc:
I merged main into this branch to fix the conflicts and cleaned up some of the metrics as well as the broken tests in [c9de7da](https://github.com/apache/solr/pull/3482/commits/c9de7da79c7048ca15473c0cefd0147666252382) and [1c01c9a](https://github.com/apache/solr/pull/3482/commits/1c01c9a9ac1835734cf4c34f183dc393df4f3da3). Tests should be passing now. Would like to merge this in at some point.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2605,95 +2636,152 @@ public SolrMetricsContext getSolrMetricsContext() {
...
+            obs -> {
+              try {
+                obs.record(reader.maxDoc(), baseAttributes);
+              } catch (Exception ignore) {
+              }
+            }));
+
+    toClose.add(
+        solrMetricsContext.observableLongGauge(
+            "solr_searcher_deleted_docs",
```
dsmiley: For these index statistics, I wonder if the name should have have index in it like "solr_searcher_index_deleted_docs"?
Just a thought.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2605,95 +2636,152 @@ public SolrMetricsContext getSolrMetricsContext() {
...
+              try {
+                obs.record(reader.numDocs(), baseAttributes);
+              } catch (Exception ignore) {
+                // replacement for nullNumber
+              }
+            }));
+
+    toClose.add(
+        solrMetricsContext.observableLongGauge(
+            "solr_searcher_max_doc",
```
dsmiley: not plural?

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
dsmiley: And technically we only need 2 of these 3 metrics, as the 3rd is computable from the 2 we keep.  But I'm sympathetic to keeping 3.  If we drop one, I suggest it be "num_docs" and rename "max_docs" as "docs" with just its comment clarifying.

mlbiscoc: > For these index statistics, I wonder if the name should have have index in it like "solr_searcher_index_deleted_docs"?

Don't have a strong preference here but I'm fine with keeping it off.

> And technically we only need 2 of these 3 metrics, as the 3rd is computable from the 2 we keep. But I'm sympathetic to keeping 3. If we drop one, I suggest it be "num_docs" and rename "max_docs" as "docs" with just its comment clarifying.

Fair point. I'm for dropping one of the metrics but why remove num_docs (live docs) over the deleted_docs? I feel like a user would care more about live_docs more on a dashboard. Majority of users probably look at live_docs over deleted so having it natively as a metric makes more sense to me.

dsmiley: > Majority of users probably look at live_docs

True -- +1
In my experience you plot them.  Having total docs & live docs would visually "show" the deleted as the difference.  So we can drop the deleted metric.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2605,95 +2636,152 @@ public SolrMetricsContext getSolrMetricsContext() {
...
-        scope);
+      SolrMetricsContext solrMetricsContext, Attributes attributes, String scope) {
+    var baseAttributes =
+        attributes.toBuilder().put(CATEGORY_ATTR, Category.SEARCHER.toString()).build();
+    // caching (boolean -> 0/1) - synchronous gauge
+    cachingEnabledGauge =
+        new AttributedLongGauge(
+            solrMetricsContext.longGauge("solr_searcher_caching_enabled", "1 if caching enabled"),
+            baseAttributes);
+    cachingEnabledGauge.set(cachingEnabled ? 1L : 0L);
```
dsmiley: I'm a bit confused here.  *The* IndexSearcher (used for nearly everything) _will_ have caches.  There is a so-called "realtime searcher" one can obtain from the SolrCore that is often pointing to the same searcher but in the off chance it doesn't, that one will not have caches.  I don't think we need/want metrics on that instance.  Do we have/publish metrics on both somehow?

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2605,95 +2636,152 @@ public SolrMetricsContext getSolrMetricsContext() {
...
+    warmupTimer =
+        new AttributedLongTimer(
+            solrMetricsContext.longHistogram(
+                "solr_searcher_warmup_time", "Searcher warmup time (ms)", OtelUnit.MILLISECONDS),
+            baseAttributes);
+
+    // fullSortCount
+    toClose.add(
+        solrMetricsContext.observableLongCounter(
+            "solr_searcher_full_sort",
```
dsmiley: looking at this and a couple other metrics, they read weirdly to me as it's not an innate characteristic of the searcher itself.  This metric counts how many queries have this "full sort" characteristic.  Shouldn't the metric name reflect that, like "solr_searcher_queries_full_sort" (or put "queries" last as we are counting them).

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2605,95 +2636,152 @@ public SolrMetricsContext getSolrMetricsContext() {
...
+                  liveDocsHitCount.sum(),
+                  baseAttributes.toBuilder().put(TYPE_ATTR, "hits").build());
+              obs.record(
+                  liveDocsNaiveCacheHitCount.sum(),
+                  baseAttributes.toBuilder().put(TYPE_ATTR, "naive_hits").build());
+            }));
+    // reader stats (numeric)
+    toClose.add(
+        solrMetricsContext.observableLongGauge(
+            "solr_searcher_num_docs",
```
dsmiley: not an innate characteristic of the searcher.  It's of the index that the searcher is viewing.  Thus arguably should have an "index" intermediate name.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2605,95 +2636,152 @@ public SolrMetricsContext getSolrMetricsContext() {
...
-              statsCache.getCacheMetrics().getSnapshot(map::putNoEx);
-              map.put("statsCacheImpl", statsCache.getClass().getSimpleName());
-            }),
-        true,
-        "statsCache",
-        Category.CACHE.toString(),
-        scope);
+    toClose.add(
+        solrMetricsContext.observableLongGauge(
+            "solr_searcher_stats_cache",
```
dsmiley: The `StatsCache` class should have been named `TermStatsCache`.  Its current name is *way* too generic looking!  _What_ stats?!  (CC @sigram )
While changing the name of that is definitely out of scope, what **is** in scope here is thinking of clear metric names we expose.  I propose that its metric name incorporate "termstats" instead of just "stats".

dsmiley: Also WDYT of moving this initialization to StatsCache (perhaps a static method) so that SolrIndexSearcher can be less distracted by details of that cache.  We do similarly for CaffeineCache.  SolrIndexSearcher is a huge class.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
mlbiscoc: > A "realtime" searcher will currently never become "registered" (since it currently lacks caching).

I don't think realTimeSearcher will publish metrics because its never registered according to this comment and metrics only initialize on registering of a searcher.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
mlbiscoc: Added index

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
mlbiscoc: > Also WDYT of moving this initialization to StatsCache (perhaps a static method) so that SolrIndexSearcher can be less distracted by details of that cache. We do similarly for CaffeineCache. SolrIndexSearcher is a huge class.

I like this idea and makes much more sense. Went ahead and moved it out.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
mlbiscoc:
```
A "realtime" searcher will currently never become "registered" (since it currently lacks caching)
```

I saw this comment so it looks like it won't publish metrics on realtime searchers since metrics only initialize when it registers.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
mlbiscoc: Added queries to the name but also split out full and sort as an attribute as `type` if it was a full sort or skipped.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
mlbiscoc: Dropped deleted

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
mlbiscoc: I like this idea so I moved it out and implemented StatsCache as a SolrInfoBean

mlbiscoc:
I refreshed my page by accident and thought I lost all my review comments. I rewrote them thinking they were gone but I guess they were still there...

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -2655,19 +2656,17 @@ public void initializeMetrics(
...
-            "Number of queries that required building a full DocList with sorting",
-            obs -> obs.record(fullSortCount.sum(), baseAttributes)));
-
-    // skipSortCount
-    toClose.add(
-        solrMetricsContext.observableLongCounter(
-            "solr_searcher_skip_sort",
-            "Number of queries where Solr skipped building a full sorted DocList and instead returned results directly from the DocSet",
-            obs -> obs.record(skipSortCount.sum(), baseAttributes)));
+            "solr_searcher_queries_sort",
```
dsmiley: shouldn't the "sort" aspect entirely be an attribute?  The metric would count queries and the attributes would characterize multiple natures of the query -- sort is one and we may add more (there are many characteristics to a query beyond the sort!  docSet?  multiThreaded?  cursorMark?  hasPostFilter?  hasReRank? ... I could go on)

dsmiley: I also support dropping this matter for now (no metric)

dsmiley (commented):
So glad you moved the metrics to StatsCache; thanks!

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
mlbiscoc: I don't think I've ever used this metric and maybe we can improve on this class later with better metric collection. I'm good with removing it.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
dsmiley: consequently lets drop "cachingEnabled" metric; it's useless / distracting

dsmiley (approved):

PR #3483, opened 2025-08-19
Exclude renovate PRs from being marked as 'stale'

janhoy:
Making renovate PRs stale after 60 days just creates message noise with not much value. This is because it does not serve to nag an author to tend to her PR, and some of us tend to periodically list all open renovate PRs some time before a release to do a batch of upgrades.

This PR adds the `exempt-stale` label to all renovate PRs.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -4,6 +4,7 @@
   "enabled": true,
   "dependencyDashboard": false,
   "enabledManagers": ["gradle", "github-actions"],
+  "labels": ["exempt-stale"],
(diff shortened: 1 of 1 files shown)


janhoy:
We got a storm of such emails this summer as the github action was auto paused (needs a bump every 3 months or so due to no commits to repo). Normally, a renovate PR will actually have some edits by renovate itself, as they are edited to bump new bugfix versions or auto rebase, so the problem is normally not that visible.

epugh (approved):
Thanks for your attention to this!   Definitly cutting down on the storms is nice!

cpoerschke (approved):

PR #3495, opened 2025-08-21
SOLR-17631: fix TestStressThreadBackup w.r.t. CheckIndex.Level

cpoerschke:
#3053 follow-up as per https://github.com/apache/solr/pull/3053#discussion_r2288536568

https://issues.apache.org/jira/browse/SOLR-17631

### DIFF
--- a/solr/core/src/test/org/apache/solr/handler/TestStressThreadBackup.java
+++ b/solr/core/src/test/org/apache/solr/handler/TestStressThreadBackup.java
@@ -31,6 +31,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import org.apache.lucene.index.CheckIndex;
 import org.apache.lucene.index.DirectoryReader;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.store.Directory;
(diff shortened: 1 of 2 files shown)


solr/core/src/test/org/apache/solr/handler/TestStressThreadBackup.java
```diff
@@ -334,7 +335,7 @@ private void validateBackup(final Path backup) throws IOException {
     final int numRealDocsExpected = Integer.parseInt(m.group());
 
     try (Directory dir = FSDirectory.open(backup)) {
-      TestUtil.checkIndex(dir, 0, true, true, null);
+      TestUtil.checkIndex(dir, CheckIndex.Level.MIN_LEVEL_FOR_SLOW_CHECKS, true, true, null);
```
dsmiley, line 338: why this choice?  I would expect that merely checksum (level 1) is fine for backups

cpoerschke: Hadn't given it too much thought, the Lucene PR for the signature change replaced `true, true, true` with `CheckIndex.Level.MIN_LEVEL_FOR_SLOW_CHECKS, true, true` albeit it in a test, and if the highest level is also the most comprehensive and thus slowest that may align with stress testing, perhaps.

If changing to level 1, any preferences w.r.t. `MIN_LEVEL_FOR_CHECKSUM_CHECKS` vs. `MIN_VALUE` vs. `DEFAULT_VALUE` (or I guess technically `1` is on the table too)?

```suggestion
      TestUtil.checkIndex(dir, CheckIndex.Level.DEFAULT_VALUE, true, true, null);
```

dsmiley: Keep it actually; you make a good point that this is a stress test.

dsmiley (approved):

PR #3496, opened 2025-08-21
SOLR-17870 - UnsupportedOperationException Fix - Return Partial Results collations when query Limit exceeded in SpellCheckCollator 

puneeetsharma:
UnsupportedOperationException Fix - Return Partial Results collations when query Limit exceeded in SpellCheckCollator

https://issues.apache.org/jira/browse/SOLR-17870

# Description

When a search query with spellcheck enabled (spellcheck=true) exceeds the specified timeAllowed, Solr throws a java.lang.UnsupportedOperationException. This occurs because in the timeout scenario, the SpellCheckCollator returns an immutable list. The SpellCheckComponent later attempts to sort this list of collations, which is an unsupported operation on an immutable list and results in the exception.

# Solution

The fix modifies the SpellCheckCollator to return a new mutable collections with partial results instead of an immutable list when the query time limit is exceeded. This allows the SpellCheckComponent to successfully sort the collations list without throwing an UnsupportedOperationException, enabling Solr to gracefully handle the timeout and return a partial response.

# Tests

To validate the fix, a unit test was added for the SpellCheckComponent. This test simulates a timeout condition by executing a spellcheck query with a very short timeAllowed value. The test asserts that the query completes successfully without throwing an UnsupportedOperationException and that the response is handled correctly.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/spelling/SpellCheckCollator.java
+++ b/solr/core/src/java/org/apache/solr/spelling/SpellCheckCollator.java
@@ -96,7 +96,7 @@ public class SpellCheckCollator {
     while (tryNo < maxTries && collNo < maxCollations && possibilityIter.hasNext()) {
 
       if (queryLimits.maybeExitWithPartialResults("SpellCheck collator")) {
-        return List.of();
+        return new ArrayList<>();
       }
 
       PossibilityIterator.RankedSpellPossibility possibility = possibilityIter.next();
--- a/solr/core/src/java/org/apache/solr/spelling/SpellCheckCollator.java
+++ b/solr/core/src/java/org/apache/solr/spelling/SpellCheckCollator.java
@@ -96,7 +96,7 @@ public class SpellCheckCollator {
     while (tryNo < maxTries && collNo < maxCollations && possibilityIter.hasNext()) {
 
       if (queryLimits.maybeExitWithPartialResults("SpellCheck collator")) {
-        return new ArrayList<>();
+        return collations;
       }
 
       PossibilityIterator.RankedSpellPossibility possibility = possibilityIter.next();
--- a/solr/core/src/test/org/apache/solr/spelling/SpellCheckCollatorTest.java
+++ b/solr/core/src/test/org/apache/solr/spelling/SpellCheckCollatorTest.java
@@ -16,6 +16,7 @@
  */
 package org.apache.solr.spelling;
 
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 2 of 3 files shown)


solr/core/src/java/org/apache/solr/spelling/SpellCheckCollator.java
```diff
@@ -96,7 +96,7 @@ public List<SpellCheckCollation> collate(
     while (tryNo < maxTries && collNo < maxCollations && possibilityIter.hasNext()) {
 
       if (queryLimits.maybeExitWithPartialResults("SpellCheck collator")) {
-        return List.of();
+        return new ArrayList<>();
```
cpoerschke: Might returning the `collations` so far (if any) be an alternative?
```suggestion
        return collations;
```

puneeetsharma: Yes seems good. I’ve added these changes. Returning collations will provide partial spell-check results when the query limit is exceeded.

solr/core/src/test/org/apache/solr/spelling/SpellCheckCollatorTest.java
```diff
@@ -756,4 +762,106 @@ public void testWithCursorMark() {
     List<?> collations = (List<?>) collationList.getAll("collation");
     assertEquals(1, collations.size());
   }
+
+  @Test
+  @SuppressWarnings("deprecation")
```
cpoerschke, line 768: Thanks for expanding the test coverage here! Unfortunately I'm not familiar enough with the spell check functionality to meaningfully review it, sorry. However, am curious w.r.t. what the deprecated stuff is and if/how perhaps it could be avoided, for this new test addition only.

puneeetsharma: Good question —the @SuppressWarnings("deprecation") is on this one new test method because it's the only place in the class that constructs org.apache.solr.spelling.Token directly (new Token() at lines 773/781), which is needed to build a SpellingResult via result.add(token, ...) — that's the only signature SpellingResult.add() exposes.
That Token class itself is @Deprecated (legacy pre-TokenStream API, with a TODO Refactor the spellchecker API to use TokenStreams properly, rather than this hack comment in its source) — it's not something I introduced, just unavoidable when directly building a SpellingResult for a unit test. Avoiding the suppression would mean reworking SpellingResult/SpellCheckCollator to stop taking Token, which is well beyond the scope of this partial-results fix.

sigram:
Hi @puneeetsharma - sorry that this fell through the cracks... The fix looks good, thank you for the unit test.

The PR needs a changelog entry (`./gradlew writeChangelog` and then edit + commit the resulting file)

puneeetsharma:
@sigram Added a changelog entry.

sigram:
@puneeetsharma this doesn't compile anymore on main because `LocalSolrQueryRequest` has been removed, use `SolrQueryRequestBase` instead (see SOLR-18038).

puneeetsharma:
Thanks @sigram , fixed — swapped LocalSolrQueryRequest for SolrQueryRequestBase per SOLR-18038. I verified the change by compiling and running SpellCheckCollatorTest against a merge of this branch with current main (my branch was based on an older main that still had the removed class, which is why it compiled locally). All tests pass.

sigram:
@puneeetsharma PR looks good now, the build is green, so it goes in - thank you for your contribution and for your patience ;)

dsmiley:
@sigram why mark this with Milestone 10.1... surely a 9.x user would appreciate it?  Don't over-think whether or not there will be a next 9.x release; my backport process will do it without you lifting a finger.  The main barrier to consider is if you think this change is likely to be difficult to backport for some reason.  This was a one-liner.

I'm changing the milestone.

PR #3499, opened 2025-08-22
SOLR-17854: Support filters to /admin/metrics endpoint for Prometheus metrics

mlbiscoc:
https://issues.apache.org/jira/browse/SOLR-17854

Add filtering for the `/admin/metrics` endpoint on Prometheus. You can filter by metric name and pre-defined set of labels (category, core, collection, shard and replica)


solr/core/src/java/org/apache/solr/handler/admin/MetricsHandler.java
```diff
@@ -69,6 +77,16 @@ public class MetricsHandler extends RequestHandlerBase implements PermissionName
...
   public static final String EXPR_PARAM = "expr";
   public static final String TYPE_PARAM = "type";
+  // Prometheus filtering parameters
+  public static final String CATEGORY_PARAM = "category";
+  public static final String CORE_PARAM = "core";
+  public static final String COLLECTION_PARAM = "collection";
+  public static final String SHARD_PARAM = "shard";
+  public static final String REPLICA_PARAM = "replica";
+  public static final String METRIC_NAME_PARAM = "name";
+  private final Set<String> labelFilterKeys =
```
dsmiley: not static?

solr/core/src/java/org/apache/solr/handler/admin/MetricsHandler.java
```diff
@@ -151,6 +169,73 @@ private void handleRequest(SolrParams params, BiConsumer<String, Object> consume
     consumer.accept("metrics", response);
   }
 
+  private void handlePrometheusRequest(SolrParams params, BiConsumer<String, Object> consumer) {
+    Set<String> metricNames = metricNamesFilter(params);
+    Map<String, Set<String>> labelFilters = labelFilters(params);
+
+    if (!metricNames.isEmpty() || !labelFilters.isEmpty()) {
```
dsmiley: logic seems wrong.  A comment would help clarify the intention.  I think such a comment here would be "no filtering".  But then the boolean expression should be: `metricNames.isEmpty() && labelFilters.isEmpty()`

solr/core/src/java/org/apache/solr/handler/admin/MetricsHandler.java
```diff
@@ -151,6 +169,73 @@ private void handleRequest(SolrParams params, BiConsumer<String, Object> consume
...
+
+    // Merge all filtered snapshots and return the merged result
+    MetricSnapshots mergedSnapshots = mergeSnapshots(allSnapshots);
+    consumer.accept("metrics", mergedSnapshots);
+  }
+
+  private Set<String> metricNamesFilter(SolrParams params) {
+    String[] metricNameParams = params.getParams(METRIC_NAME_PARAM);
+    if (metricNameParams == null || metricNameParams.length == 0) {
+      return Collections.emptySet();
```
dsmiley: nit: nowadays use `Set.of()`

solr/core/src/java/org/apache/solr/handler/admin/MetricsHandler.java
```diff
@@ -151,6 +169,73 @@ private void handleRequest(SolrParams params, BiConsumer<String, Object> consume
...
+    }
+
+    Set<String> metricNames = new HashSet<>();
+    for (String param : metricNameParams) {
+      if (param != null && !param.trim().isEmpty()) {
```
dsmiley: A null value is impossible.  And no empty check is necessary; splitSmart handles that.  Thus you can keep this simple and drop the condition.

solr/core/src/java/org/apache/solr/handler/admin/MetricsHandler.java
```diff
@@ -151,6 +169,73 @@ private void handleRequest(SolrParams params, BiConsumer<String, Object> consume
...
+    consumer.accept("metrics", mergedSnapshots);
+  }
+
+  private Set<String> metricNamesFilter(SolrParams params) {
+    String[] metricNameParams = params.getParams(METRIC_NAME_PARAM);
+    if (metricNameParams == null || metricNameParams.length == 0) {
+      return Collections.emptySet();
+    }
+
+    Set<String> metricNames = new HashSet<>();
```
dsmiley: I recommend instead collecting to a List then finally calling `Set.copyOf(list)`.  These immutable implementations are very efficient; I try to use them whenever I need an immutable List, Set, Map, especially if it's going to be evaluated/used multiple times.

solr/core/src/java/org/apache/solr/handler/admin/MetricsHandler.java
```diff
@@ -151,6 +169,73 @@ private void handleRequest(SolrParams params, BiConsumer<String, Object> consume
...
+
+    return labelFilters;
+  }
+
+  private Set<String> labelFilterValues(SolrParams params, String paramName) {
+    String[] filterLabels = params.getParams(paramName);
+    Set<String> labelValues = new HashSet<>();
+    if (filterLabels != null) {
+      for (String labels : filterLabels) {
+        if (labels != null && !labels.trim().isEmpty()) {
```
dsmiley: same feedback as earlier.
Maybe should have a general method readParamAsSet(params, name, char delim)

solr/core/src/java/org/apache/solr/handler/admin/MetricsHandler.java
```diff
@@ -499,9 +584,87 @@ private List<MetricType> parseMetricTypes(SolrParams params) {
...
   }
 
-  private String getCoreNameFromRegistry(String registryName) {
-    String coreName = registryName.substring(registryName.indexOf('.') + 1);
-    return coreName.replace(".", "_");
+  /**
+   * Merge a collection of individual {@link MetricSnapshot} instances into one {@link
+   * MetricSnapshots}. This is necessary because we create a {@link
+   * io.opentelemetry.sdk.metrics.SdkMeterProvider} per Solr core resulting in duplicate metric
+   * names across cores which is an illegal format if not under the same prometheus grouping.
```
dsmiley: > if not under

Don't you mean the inverse, "if under" ?

solr/core/src/java/org/apache/solr/handler/admin/MetricsHandler.java
```diff
@@ -499,9 +584,87 @@ private List<MetricType> parseMetricTypes(SolrParams params) {
...
+   */
+  private MetricSnapshots mergeSnapshots(List<MetricSnapshot> snapshots) {
```
dsmiley, line 581: deja vu... did you take this from elsewhere?

solr/core/src/java/org/apache/solr/metrics/SolrMetricManager.java
```diff
@@ -1661,17 +1661,17 @@ public MetricsConfig getMetricsConfig() {
     return metricsConfig;
   }
 
-  /** Get a shallow copied map of {@link PrometheusMetricReader}. */
-  public Map<String, PrometheusMetricReader> getPrometheusMetricReaders() {
+  /** Get a shallow copied map of {@link FilterablePrometheusMetricReader}. */
```
dsmiley, line 1643: nit: just because the precise implementation is filterable, doesn't mean you need to be so precise in the javadoc.  Whatever though.

solr/core/src/test/org/apache/solr/handler/admin/MetricsHandlerTest.java
```diff
@@ -17,724 +17,242 @@
...
-      h.getCoreContainer().getMetricManager().registry("solr.jetty").remove("solrtest_foo:bar");
-    }
   }
 
-  // NOCOMMIT: This test does a bunch of /admin/metrics calls with various params, with various
-  // filters and parameters. We have not migrated all the metrics to otel yet or even created any
-  // filters. Once that is done, we should revisit this test and assert the prometheus response.
   @Test
-  @BadApple(bugUrl = "https://issues.apache.org/jira/browse/SOLR-17458")
-  public void test() throws Exception {
```
mlbiscoc, line 93: Many of these tests that I removed tested all kind of different metrics and their values as well as filtering. I felt it was too much. We already test many of these metrics in other tests specific to their classes and `TestPrometheusResponseWriter` does a kind of higher level testing. So I dropped these and just did filter tests.

solr/core/src/java/org/apache/solr/handler/admin/MetricsHandler.java
mlbiscoc: Yeah I wrote this a while back. This was originally in `PrometheusResponseWriter` and I just moved it into `MetricHandler` which I think is a better place for this.

mlbiscoc:
> I really felt the label matching needed optimization so I did so. I started writing how it might work but then I wanted to get in on some algorithm fun.

Thanks for helping out on this!

solr/core/src/test/org/apache/solr/metrics/otel/FilterablePrometheusMetricReaderTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+  Labels actualLabels = Labels.of("key1", "value1", "key2", "value2", "key3", "value3");
+
+  @Test
+  public void testFilterMatchingLabel() {
+    SortedMap<String, Set<String>> requiredLabels =
+        new TreeMap<>() {
+          {
+            put("key1", Set.of("value1"));
+          }
```
dsmiley: Please don't use this style (subclass with anonymous constructor).  Just do `new TreeMap<>(Map.of(...))`

solr/core/src/test/org/apache/solr/handler/admin/MetricsHandlerTest.java
```diff
@@ -17,724 +17,242 @@
...
-        (k, v) -> {
-          assertTrue(v instanceof MapWriter);
-          Map<String, Object> map = new HashMap<>();
-          ((MapWriter) v).toMap(map);
-          assertTrue(map.size() > 2);
-        });
+    var metrics = (MetricSnapshots) resp.getValues().get("metrics");
+    assertEquals(2, metrics.size());
+    assertEquals(expectedRequestsMetricName, metrics.get(0).getMetadata().getPrometheusName());
+    assertEquals(expectedSearcherMetricName, metrics.get(1).getMetadata().getPrometheusName());
```
dsmiley: is the order sorted, and thus we can know the order without it changing?

solr/core/src/java/org/apache/solr/handler/admin/MetricsHandler.java
```diff
@@ -151,6 +171,58 @@ private void handleRequest(SolrParams params, BiConsumer<String, Object> consume
...
+    return labelFilters;
+  }
+
+  private Set<String> readParamsAsSet(SolrParams params, String paramName) {
+    String[] paramValues = params.getParams(paramName);
+    if (paramValues == null || paramValues.length == 0) {
+      return Set.of();
+    }
+
+    Set<String> paramSet = new HashSet<>();
```
dsmiley: This can merely be a List.  `Set.copyOf` will do the deduplication _again_.  It accepts a Collection.

solr/core/src/test/org/apache/solr/handler/admin/MetricsHandlerTest.java
mlbiscoc: I don't think this is actually true and I'm not going to bet on it. Changed up the test.

mlbiscoc:
Anything else here? Would like to merge as I have a branch ready for review for [SOLR-17853](https://issues.apache.org/jira/browse/SOLR-17853) that uses this filtering.

dsmiley (approved):
thanks

PR #3500, opened 2025-08-23
SOLR-17864: Migrate System Properties to modern equivalents.

epugh:

authentication.plugin=solr.security.auth.plugin
basicauth=solr.security.auth.basicauth.credentials
cloud.solr.client.max.stale.retries=solr.solrj.cloud.max.stale.retries
solr.hidden.sys.props=solr.responses.hidden.sys.props


solr/core/src/java/org/apache/solr/core/NodeConfig.java
```diff
@@ -861,25 +861,17 @@ public NodeConfigBuilder setHiddenSysProps(String hiddenSysProps) {
...
     /**
-     * Finds list of hiddenSysProps requested by system property or environment variable or the
-     * default
+     * Finds list of hiddenSysProps requested in priority of solr.xml, system property or the
+     * default set
      *
-     * @return set of raw hidden sysProps, may be regex
+     * @return set of raw hidden system properties, may be regex
      */
     private Set<String> resolveHiddenSysPropsFromSysPropOrEnvOrDefault(String hiddenSysProps) {
```
dsmiley: shouldn't this use EnvUtils?

solr/core/src/java/org/apache/solr/handler/configsets/ConfigSetAPIBase.java
```diff
@@ -96,10 +96,10 @@ protected void runConfigSetCommand(
   }
 
   protected void ensureConfigSetUploadEnabled() {
-    if (!"true".equals(System.getProperty("configset.upload.enabled", "true"))) {
+    if (!"true".equals(System.getProperty("solr.configset.upload.enabled", "true"))) {
```
dsmiley: shouldn't this use EnvUtils?

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java
```diff
@@ -89,7 +89,7 @@ public abstract class CloudSolrClient extends SolrClient {
 
   // no of times collection state to be reloaded if stale state error is received
   private static final int MAX_STALE_RETRIES =
-      Integer.parseInt(System.getProperty("cloudSolrClientMaxStaleRetries", "5"));
+      Integer.parseInt(System.getProperty("solr.solrj.cloud.max.stale.retries", "5"));
```
dsmiley, line 92: shouldn't this use EnvUtils?

solr/core/src/java/org/apache/solr/core/NodeConfig.java
epugh: I kind of wondered about this...   So, what I discovered about EnvUtils is that it, on Solr startup, merges the environment variables and system properties into one set, and puts them all in System.getProperties.   I actually did some work early on to migrated to using the EnvUtils equivalents of System.getProperty, and then backed away as "too much", but maybe I should have done that.  Also, maybe using EnvUtils would let me change this method to `resolveHiddenSysPropsFromEnvUtilsOrDefault` as a name???

solr/core/src/java/org/apache/solr/handler/configsets/ConfigSetAPIBase.java
epugh: It could, but it turns out it doesn't have too!   I almost added a JIRA to remind me to add System.getProperty to forbidden apis....

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java
epugh: It could, and indeed it would be shorter...   Do you think I should just start using EnvUtils everywhere instead of these equivalents?

epugh:
Pending the discussion on using `EnvUtils`, I think this PR is ready.  (and honestly, we could save the EnvUtils changes for another PR more globally!).   I've run the tests locally, and had Crave work as well.  Right now Crave seems a bit unhappy....  I'll kick off another attmept...

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java
dsmiley: Yes; use EnvUtils to read any property that starts with "solr".  Consistency.  Your PRs here are the perfect time to do this at the same time as renames.  I don't suggest you necessarily touch additional lines, but at least update the lines you need to touch for sys prop name changes.
CC @janhoy

solr/core/src/java/org/apache/solr/core/NodeConfig.java
dsmiley: >  Also, maybe using EnvUtils would let me change this method to resolveHiddenSysPropsFromEnvUtilsOrDefault as a name???

Yes exactly!  This is a large part of the very point of EnvUtils!  Jan introduced EnvUtils but didn't convert our codebase to use it.

solr/core/src/java/org/apache/solr/handler/configsets/ConfigSetAPIBase.java
dsmiley: forbidden apis is too much.  There are some sys props that aren't related to EnvUtils like testing/internal ones that don't start with "solr".

solr/solrj/src/java/org/apache/solr/client/solrj/impl/CloudSolrClient.java
epugh: Done

solr/core/src/java/org/apache/solr/handler/configsets/ConfigSetAPIBase.java
epugh: okay.

PR #3503, opened 2025-08-23
SOLR-17864: Flip solr.hideStackTrace to .enabled equivalent.

epugh:
Tricky one because I didn't want to change the solr.xml name AS well, so now in solr.xml we have solr.hideStackTrace, but system is solr.responses.stacktrace.enabled.

Why is this in solr.xml?

I am going to leave this one  for just this one, unless I find other solr.xml that are also system variable equivalents..    Then maybe this grows into a whole "why do we define these in solr.xml and in system variables?" disucssion!

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/NodeConfig.java
+++ b/solr/core/src/java/org/apache/solr/core/NodeConfig.java
@@ -618,7 +618,8 @@ public class NodeConfig {
     private String defaultZkHost;
     private Set<Path> allowPaths = Collections.emptySet();
     private List<String> allowUrls = Collections.emptyList();
-    private boolean hideStackTrace = Boolean.getBoolean("solr.hideStackTrace");
+    private boolean hideStackTrace =
+        !(Boolean.parseBoolean(System.getProperty("solr.responses.stacktrace.enabled", "true")));
 
     private final Path solrHome;
     private final String nodeName;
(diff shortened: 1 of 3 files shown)


solr/core/src/java/org/apache/solr/core/NodeConfig.java
```diff
@@ -618,7 +618,8 @@ public static class NodeConfigBuilder {
     private String defaultZkHost;
     private Set<Path> allowPaths = Collections.emptySet();
     private List<String> allowUrls = Collections.emptyList();
-    private boolean hideStackTrace = Boolean.getBoolean("solr.hideStackTrace");
+    private boolean hideStackTrace =
+        !(Boolean.parseBoolean(System.getProperty("solr.responses.stacktrace.enabled", "true")));
```
dsmiley, line 622: EnvUtils?

solr/solrj/src/resources/DeprecatedSystemPropertyMappings.properties
```diff
@@ -19,3 +19,4 @@ solr.enable.remote.streaming=solr.requests.streaming.remote.enabled
 solr.enable.stream.body=solr.requests.streaming.body.enabled
 solr.delete.unknown.cores=solr.cloud.startup.delete.unknown.cores.enabled
 solr.config.set.forbidden.file.types=solr.configset.forbidden.file.types
+solr.hide.stack.trace=solr.responses.stacktrace.enabled
```
dsmiley, line 22: it'd be neat if there was a way to reflect that we inverted a boolean.  Like use an exclamation in the beginning.

solr/core/src/java/org/apache/solr/core/NodeConfig.java
epugh: Sure!

solr/solrj/src/resources/DeprecatedSystemPropertyMappings.properties
epugh: Not a bad idea..    Maybe if the line starts with a `!` then you know while processing that it's an inversion.   That would simplify the rule logic in EnvUtils.java..    This is a great idea.

PR #3504, opened 2025-08-23
SOLR-17864: Migrate System Properties to modern style

epugh:

disable.config.edit=solr.configset.edit.enabled
disable.v2.api=solr.api.v2.enabled

This was pretty straightforward.  We may want to make another rule in EnvUtils for flipping properties that start with `disabled`, for now I hard coded the two.

### DIFF
--- a/solr/solrj/src/java/org/apache/solr/common/util/EnvUtils.java
+++ b/solr/solrj/src/java/org/apache/solr/common/util/EnvUtils.java
@@ -233,7 +233,8 @@ public class EnvUtils {
               key);
           setProperty(key, String.valueOf(!Boolean.getBoolean(deprecatedKey)));
         } else if (deprecatedKey.equals("disable.config.edit")
-            || deprecatedKey.equals("disable.v2.api") || (deprecatedKey.equals("solr.hide.stack.trace")) {
+            || deprecatedKey.equals("disable.v2.api")
+            || deprecatedKey.equals("solr.hide.stack.trace")) {
           log.warn(
               "Converting from legacy system property {} to modern .enabled equivalent {} by flipping the boolean property value.",
               deprecatedKey,


solr/core/src/java/org/apache/solr/handler/SolrConfigHandler.java
```diff
@@ -102,9 +103,9 @@
 public class SolrConfigHandler extends RequestHandlerBase
     implements SolrCoreAware, PermissionNameProvider {
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
-  public static final String CONFIGSET_EDITING_DISABLED_ARG = "disable.configEdit";
-  public static final boolean configEditing_disabled =
-      Boolean.getBoolean(CONFIGSET_EDITING_DISABLED_ARG);
+  public static final String CONFIGSET_EDITING_ENABLED_ARG = "solr.configset.edit.enabled";
```
dsmiley: I'm not sure we can equate "config editing" with "configset editing".  Maybe.  This flag controls the [Config API ](https://solr.apache.org/guide/solr/latest/configuration-guide/config-api.html) which is conceptually similar to but not identical to configset.  It's a separate area of the code and documentation, so I think it's setting should be reflective of that.  Perhaps `solr.api.config.immutable` or `solr.configset.configapi.immutable`

In my example there, I suggest `immutable` instead of `edit` as I think that a more general word and is in fact how we display a message to the user below...

solr/core/src/java/org/apache/solr/handler/SolrConfigHandler.java
```diff
@@ -132,10 +133,10 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp) throw
...
     Command command = new Command(req, rsp, httpMethod);
     if ("POST".equals(httpMethod)) {
-      if (configEditing_disabled || isImmutableConfigSet) {
+      if (!configEditingEnabled || isImmutableConfigSet) {
         final String reason =
-            configEditing_disabled
-                ? "due to " + CONFIGSET_EDITING_DISABLED_ARG
+            !configEditingEnabled
+                ? "due to " + CONFIGSET_EDITING_ENABLED_ARG
                 : "because ConfigSet is immutable";
```
dsmiley: to my point -- even "immutable" is the word chosen to communicate this to the user

solr/solr-ref-guide/modules/configuration-guide/pages/configsets-api.adoc
```diff
@@ -88,7 +88,7 @@ The output will look like:
 Upload a configset, which is sent as a zipped file.
 A single, non-zipped file can also be uploaded with the `filePath` parameter.
 
-This functionality is enabled by default, but can be disabled via a runtime parameter `-Dconfigset.upload.enabled=false`.
+This functionality is enabled by default, but can be disabled via a runtime parameter `-Dsolr.configset.upload.enabled=false`.
```
dsmiley, line 91: Maybe `solr.configset.immutable` would be a more universal way to communicate.... immutability?  Instead of "updload", "edit", ....

Oh, I see this flag only controls upload while it's possible to delete and clone.  @gerlowskija , shouldn't the underlying protection be about immutability generally instead of merely upload specifically?  Just upload seems like half of a security measure.

solr/solrj/src/java/org/apache/solr/common/util/EnvUtils.java
```diff
@@ -232,6 +232,14 @@ static synchronized void init(
               deprecatedKey,
               key);
           setProperty(key, String.valueOf(!Boolean.getBoolean(deprecatedKey)));
+        } else if (deprecatedKey.equals("disable.config.edit")
+            || deprecatedKey.equals("disable.v2.api")) {
```
dsmiley: why do these need to be called out specifically as a special case here?  I thought you had a general mechanism, wherein you need to only list these names in a file.

solr/core/src/java/org/apache/solr/handler/SolrConfigHandler.java
epugh: I think you are totally right that `config` and `configset` are two different things...      As far as the name goes, we don't have the term `immutable` used in other names, so I don't know that I want to add that.   And I'd like it to end in .enabled.   

`solr.config.edit.enabled`?  or maybe `solr.api.config.enabled`?

epugh: Okay, I just went with `solr.api.config.edit.enabled`

solr/solr-ref-guide/modules/configuration-guide/pages/configsets-api.adoc
epugh: @gerlowskija if this is something we need, maybe create a JIRA for it, as it seems like it's own block of work.?

solr/solrj/src/java/org/apache/solr/common/util/EnvUtils.java
epugh: I did, but it only worked with this pattern:

```
 if (key.endsWith(".enabled") && deprecatedKey.endsWith(".disabled")) {
```

I was thinking that I would just collect up a few of these, and then later go back and do something fancier.   I didn't have a sense of if it would be one or two, or a whole set.  Tackle it in another PR however.

Also, all of this get's deleted in a few months when we cut `branch_10` and start Solr 11!

epugh:
@dsmiley  if the name for the config editing works for you, then this one I think is ready to merge.

dsmiley:
Hoping for @gerlowskija's input as well.  Let's not merge a PR on the weekend that was opened on the weekend!

epugh:
> Hoping for @gerlowskija's input as well. Let's not merge a PR on the weekend that was opened on the weekend!

Fair point...   I'll keep that in mind, and adjust to what @gerlowskija says!

PR #3509, opened 2025-08-25
SOLR-17806: Convert CoreContainer metrics to OTEL

jkmuriithi:

# Description

Replace Dropwizard instruments within `CoreContainer.java` with their OpenTelemetry equivalents.

# Solution

Replaced `SolrMetricsContext` method calls.

Screenshot 2025-08-26 at 12 19 21 PM

# Tests

Test failures match the originating branch.

# Checklist

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/CoreContainer.java
+++ b/solr/core/src/java/org/apache/solr/core/CoreContainer.java
@@ -1038,7 +1038,7 @@ public class CoreContainer {
 
     solrMetricsContext.observableLongGauge(
         "solr_disk_space",
-        "Disk metrics for Solr's data home directory",
+        String.format("Disk metrics for Solr's data home directory (%s)", dataHome.toString()),
         measurement -> {
           try {
             var fileStore = Files.getFileStore(dataHome);


solr/solrj-zookeeper/src/java/org/apache/solr/client/solrj/impl/NodeValueFetcher.java
```diff
@@ -38,6 +38,8 @@
  * This class is responsible for fetching metrics and other attributes from a given node in Solr
  * cluster. This is a helper class that is used by {@link SolrClientNodeStateProvider}
  */
+// NOCOMMIT: Need to removed hardcoded references to Dropwizard metrics for OTEL conversion, and
+// probably change enum structure to be more compatible with OTEL naming
 public class NodeValueFetcher {
```
mlbiscoc, line 43: Will need to look into what this does and how it can be migrated.

mlbiscoc (commented):
Also need to migrate `testCoreContainerMetrics` that checked these container metrics

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1000,6 +1002,26 @@ private void loadInternal() {
 
     // initialize gauges for reporting the number of cores and disk total/free
 
+    Attributes containerAttrs =
```
dsmiley, line 1009: Can you please move metric initialization of SolrCores out of CoreContainer and into SolrCores.initializeMetrics where it belongs?

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1000,6 +1002,26 @@ private void loadInternal() {
...
+        Attributes.builder().put(CATEGORY_ATTR, SolrInfoBean.Category.CONTAINER.toString()).build();
+
+    solrMetricsContext.observableLongGauge(
+        "solr_cores_loaded",
```
dsmiley: I forget... are we namespacing all metrics under major areas (as we are doing for system props / env) lately?  If so, a name like "solr_node_cores_ ..." would be better.  If not... ok.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1018,8 +1040,52 @@ private void loadInternal() {
         "unloaded",
         SolrInfoBean.Category.CONTAINER.toString(),
         "cores");
+
     Path dataHome =
         cfg.getSolrDataHome() != null ? cfg.getSolrDataHome() : cfg.getCoreRootDirectory();
+
+    solrMetricsContext.observableLongGauge(
+        "solr_cores_filesystem_disk_space",
+        "Disk metrics for Solr's data home directory",
```
dsmiley: The description should reflect what we're using.  Maybe it could also have the path?

An aside... I kind of question wether Solr needs a "data home" differentiated from a "core root directory" but should a debate is out of scope.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1018,8 +1040,52 @@ private void loadInternal() {
...
+        measurement -> {
+          try {
+            measurement.record(
+                Files.getFileStore(dataHome).getTotalSpace(),
```
dsmiley: fileStore can be extracted so its used once

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1018,8 +1040,52 @@ private void loadInternal() {
         "unloaded",
         SolrInfoBean.Category.CONTAINER.toString(),
         "cores");
+
     Path dataHome =
         cfg.getSolrDataHome() != null ? cfg.getSolrDataHome() : cfg.getCoreRootDirectory();
+
+    solrMetricsContext.observableLongGauge(
+        "solr_cores_filesystem_disk_space",
```
dsmiley: I suggest calling this "solr_disk_space", (maybe "node" namespaced too).  
Practically speaking, I think this is the only disk space metric that matters.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1018,8 +1040,52 @@ private void loadInternal() {
...
+          } catch (IOException e) {
+            throw new SolrException(
+                ErrorCode.SERVER_ERROR,
+                "Error retrieving disk space information for data home directory" + dataHome,
+                e);
+          }
+        },
+        OtelUnit.BYTES);
+    solrMetricsContext.observableLongGauge(
+        "solr_cores_root_disk_space",
```
dsmiley: I suggest dropping this as redundant with "data".  If someone defined "data" (which is maybe rare and perhaps should be unsupported), then this different directory would be insignificant, only holding metadata.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
jkmuriithi: Done. Had to make some other changes as well to make SolrCores a `SolrInfoBean`.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
jkmuriithi: When aggregating, do we need metrics to have consistent descriptions? Or is it OK if each replica reports a different description under the same metric name? @mlbiscoc I'm not too familiar with the specifics of OTEL so maybe you could chime in?

mlbiscoc: Descriptions to my knowledge are not used in anyway for aggregation. So if they get picked up from 2 Solr nodes, each with a different description, that is ok. Prometheus backend or whatever backend they use should just looks at metric names and attributes. Descriptions are only there for user understanding. Also helps that updating descriptions are always backwards compatible to change, so we can update this later on if we need improvement.

jkmuriithi: Good to know. Updated the description with the path

PR #3510, opened 2025-08-26
SOLR-17879: Fail to start if it's major version is smaller than the cluster

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17879

The test is marked Ignored because the failure occurs in a way that does not close objects that the ZkController created midway through it's construction (like SolrZkClient).  So there's a suite failure but I can see the test otherwise passes.  Maybe I should just remove it.

Maybe the BATS tests can work with docker?  Or do we add to the docker module's tests, which only run on certain buildds.  Regardless, imagine this is backported to Solr 9, and we later add a test that runs a branch_9x snapshot (published where?) and then runs the current docker version, then we can test the behavior here.  Such a thing would need to be a separate PR.

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/ZkController.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ZkController.java
@@ -600,29 +600,31 @@ public class ZkController implements Closeable {
   }
 
   /**
-   * Checks version compatibility with other live nodes in the cluster. Refuses to start if there's
-   * a major version difference between the current Solr version and other live nodes in the cluster.
+   * Checks version compatibility with other nodes in the cluster. Refuses to start if there's a
+   * major version difference between our Solr version and other nodes in the cluster. Note: uses
+   * live nodes.
    */
-  private void checkVersionCompatibility() throws InterruptedException {
-    try {
-      Optional<SolrVersion> lowestVersion = zkStateReader.fetchLowestSolrVersion();
-      if (lowestVersion.isPresent()) {
-        SolrVersion currentVersion = SolrVersion.LATEST;
-        SolrVersion clusterVersion = lowestVersion.get();
-        
-        if (currentVersion.getMajorVersion() < clusterVersion.getMajorVersion()) {
-          log.warn("Current Solr version {} is older than cluster version {}", 
-                   currentVersion, clusterVersion);
-          String message = String.format(Locale.ROOT,
-              "Cannot start Solr due to major version incompatibility. Current Solr version: %s, "
-              + "lowest version in cluster: %s. All nodes in a cluster must run the same major version.",
-              currentVersion, clusterVersion);
+  private void checkClusterVersionCompatibility() throws InterruptedException, KeeperException {
+    Optional<SolrVersion> lowestVersion = zkStateReader.fetchLowestSolrVersion();
+    if (lowestVersion.isPresent()) {
+      SolrVersion ourVersion = SolrVersion.LATEST;
+      SolrVersion clusterVersion = lowestVersion.get();
+
+      if (ourVersion.lessThan(clusterVersion)) {
+        log.warn(
+            "Our Solr version {} is older than cluster version {}", ourVersion, clusterVersion);
+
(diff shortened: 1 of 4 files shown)


dsmiley:
It's been 4 days; I'd like to merge Tuesday.
I should add an upgrade note about this new limitation.  There's probably another page to update about upgrades generally as well.

dsmiley:
@HoustonPutman I updated this to consider the minor version as well, as discussed in JIRA.  I ran the test manually when un-ignored.
If I don't hear more, I'll merge Friday night EST.

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
```diff
@@ -872,38 +873,40 @@ public void removeLiveNodesListener(LiveNodesListener listener) {
...
+    SolrVersion lowest = null;
     // the last version to not specify its version in live nodes
     final SolrVersion UNSPECIFIED_VERSION = SolrVersion.valueOf("9.9.0");
+
     for (String nodeName : liveNodeNames) {
       String path = LIVE_NODES_ZKNODE + "/" + nodeName;
       byte[] data = zkClient.getData(path, null, null, true);
       if (data == null || data.length == 0) {
-        return UNSPECIFIED_VERSION;
+        return Optional.of(UNSPECIFIED_VERSION);
```
HoustonPutman, line 894: Since this is an optional now, why not just return empty instead of 9.9.0? (Same for the similar thing below)

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -602,6 +602,49 @@ private void checkNoOldClusterstate(final SolrZkClient zkClient) throws Interrup
...
+      if (ourVersion.lessThan(clusterVersion)) {
+        log.warn(
+            "Our Solr version {} is older than cluster version {}", ourVersion, clusterVersion);
+
+        // Check major version compatibility
+        if (ourVersion.getMajorVersion() < clusterVersion.getMajorVersion()) {
+          String message =
+              String.format(
+                  Locale.ROOT,
+                  "Refusing to start Solr to avoid lowering the lowest major version of nodes in the cluster. "
```
HoustonPutman:
```suggestion
                  "Refusing to start Solr, since our version is lower than the lowest version currently running in the cluster. "
```

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -602,6 +602,49 @@ private void checkNoOldClusterstate(final SolrZkClient zkClient) throws Interrup
...
+          throw new SolrException(ErrorCode.INVALID_STATE, message);
+        }
+
+        // Check minor version compatibility within the same major version
+        if (ourVersion.getMajorVersion() == clusterVersion.getMajorVersion()
+            && ourVersion.getMinorVersion() < clusterVersion.getMinorVersion()) {
+          String message =
+              String.format(
+                  Locale.ROOT,
+                  "Refusing to start Solr to avoid lowering the lowest minor version of nodes in the cluster. "
```
HoustonPutman:
```suggestion
                  "Refusing to start Solr, since our version is lower than the lowest version currently running in the cluster. "
```

solr/solrj-zookeeper/src/java/org/apache/solr/common/cloud/ZkStateReader.java
dsmiley: I was very deliberate about what to do here, and the javadocs should spell this out.  If there's no data for a particular node, then we know an older version of Solr is a member of this cluster.  An empty return from the method here conveys something different -- that there are no members of the cluster yet.

HoustonPutman: RTFM, fair enough haha.

HoustonPutman (approved):

PR #3511, opened 2025-08-26
New Collections UI 

iamsanjay:
Summary: First cut of Collections screen. 

Includes

- List Collections (left pane) with health dots
- Reload (no-arg; uses selected)
- Create Collection (basic: name, configSet, shards, replicas)
- Delete Collection (confirm dialog)
- Shard details panel
- Replica list per shard
- Add Replica (dialog)
- Delete Replica (inline)


iamsanjay:
@malliaridis Working on test cases now. JIRA is missing I believe. Can you confirm?

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/theme/Theme.kt
```diff
@@ -52,8 +52,8 @@ fun SolrTheme(
     content: @Composable () -> Unit,
 ) {
     val colorScheme = when {
-        useDarkTheme -> darkScheme
-        else -> lightScheme
```
malliaridis, line 56: I believe this change was not intended to be committed?

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/CollectionsComponent.kt
```diff
@@ -0,0 +1,35 @@
...
+    val model: StateFlow<Model>
+
+    val labels: Flow<CollectionsStore.Label>
+
+    data class Model(
+        val collectionsList: CollectionsList = CollectionsList(),
+        val selectedCollectionData: CollectionData? = null,
+        val liveNodesData: List<String> = emptyList(),
+        val mutating: Boolean = false,
+        val configSets: ListConfigSets? = null,
```
malliaridis, line 24: I would prefer a clear separation of concerns, so that UI components do not use any API data classes or namings.

This way we can test each element separately and in combination, and do not depend on the API's structure or models. Having a clear separation of internal domain models with API data models would, I believe, make minification easier as well.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/CollectionsComponent.kt
malliaridis, line 1: In general, I would try to add doc comments to all elements of an interface. First thing would be, what does the component interface contain? What is its role and what can be done with it?

Then of course what data does the model hold? Documentation of the deta model fields is here important. What do the fields include, what not?

Last but not least what interaction functions are available? What effect do they have or how are they used?

What I have also seen here is that you have created one "ultimate" component, which includes and covers a large amount of functionality, instead of breaking it down to separate components for easier testing and smaller files (better overview). THe ultimate component has as an effect that you pass down through several UI layers a function and / or a state, adding unnecessarily parameters / properties to UI components that have nothing to do with it. Sometimes it is not avoidable, sure, but in this case it can be avoided by using different compoenents.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/CollectionsComponent.kt
```diff
@@ -0,0 +1,35 @@
...
+    )
+
+    fun selectCollection(name: String)
+    fun fetchLiveNodesData()
```
malliaridis, line 28: Ideally, UI components would not define an initial fetch function. Instead, the stores would use a [Bootstrapper](https://arkivanov.github.io/MVIKotlin/store.html#Bootstrapper:~:text=they%20are%20connected%3A-,Bootstrapper,-This%20component%20bootstraps) to initiate the data fetching on loading / before displaying.

A fetch function (I would prefer in the UI the word get or load) would make mroe sense in cases where the user wants to reload something, and in that case it would be "reloadXYZ".

So the functions in a UI component should be actual inputs of the user, rather than initialization functions. This keeps the interface clean and the code footprint smaller.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/data/ClusterStatusModels.kt
malliaridis, line 1: Some of these interfaces like the response header are used widely in the API data models, so it would make more sense to move them to their own files. Even though it is intriguing adding multiple classes in the same file, I (personally) consider it a bad practice and should only be done in case some data classes are always served together (e.g. nested fields).

Ideally, we would also add doc comments to each field and what information it holds, practically we have for that only our API spec, which is not accessible during deveopment.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/data/ClusterStatusModels.kt
```diff
@@ -0,0 +1,82 @@
...
+) {
+    fun toCollectionData(name: String) = CollectionData(
+        name = name,
+        configName = configName.orEmpty(),
+        replicationFactor = replicationFactor ?: 0,
+        routerName = router?.name.orEmpty(),
+        router = router?.name.orEmpty(),
+        shards = shards,
+        shardCount = shards.size,
+    )
```
malliaridis, line 52: Would it perhaps make more sense to have this mapping function inside Mappers.kt, so that we keep logic out of data classes and all the mappings inside the same Mapper file?

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/data/ClusterStatusModels.kt
```diff
@@ -0,0 +1,82 @@
...
+    val state: String? = null,
+    // “true”/“false” are strings in your sample JSON
+    val leader: String? = null,
+    @SerialName("force_set_state") val forceSetState: String? = null,
+    @SerialName("base_url") val baseUrl: String? = null,
+)
+
+internal fun ClusterStatusResponse.collectionInfo(name: String): CollectionInfo? = cluster.collections[name]
+
+internal fun ClusterStatusResponse.toCollectionDataOrNull(name: String): CollectionData? = collectionInfo(name)?.toCollectionData(name)
```
malliaridis, line 82: Same applies for these. Perhaps it would make more sense to have them in the Mappers.kt (only if they are UI-specific though).

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/data/ZkTreeData.kt
```diff
@@ -0,0 +1,28 @@
...
+
+@Serializable
+data class ZkTree(
+    val tree: List<ZkNode> = emptyList(),
+)
+
+fun ZkTree.liveNodes(): List<String> {
+    val root = tree.firstOrNull { it.text == "/live_nodes" } ?: return emptyList()
+    return root.children.map { it.text }
+}
```
malliaridis, line 14: As mentioned before, it may make sense, at least for UI-specific logic, to move them to separate Mappers.kt file.

In the case it is not UI-specific, it would either make sense to move it to the API layer (where the response is transformed to internal domain model) or at least into the data class, so extension functions are not incorrectly used (may have some impact in compilation and generated code, see generated classes for reference).

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/CollectionsComponent.kt
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.solr.ui.components.collections.store.CollectionsStore
+
+/**
+ * Component interface that represents the collection section.
+ */
+interface CollectionsComponent {
+
+    val model: StateFlow<Model>
+
+    val labels: Flow<CollectionsStore.Label>
```
malliaridis, line 17: This element adds a direct dependency to the `CollectionStore`, which we in general want to avoid, mainly for testing purposes, but also for loose coupling (design goal).

Only in the component implementation (`DefaultCollectionsComponent`) you define which store you wanna use (implementation detail) and how you consume its labels (if it has any). With this dependency here, you are already require a `CollectionsStore` implementation with `CollectionStore.Label` labels to be used.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/data/CollectionsList.kt
```diff
@@ -0,0 +1,10 @@
+package org.apache.solr.ui.components.collections.data
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class CollectionsList(
+    @SerialName("collections")
+    val collectionsList: List<String> = emptyList(),
```
malliaridis, line 9: Why not using the same name as in the response here, and instead use an explicit `SerialName` annotation? Same of course applies to other serializable classes that are used as data transfer objects / data classes between API and frontend.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/integration/DefaultCollectionsComponent.kt
```diff
@@ -0,0 +1,74 @@
...
+    override fun selectCollection(name: String) {
+        store.accept(CollectionsStore.Intent.FetchCollectionData(name))
+    }
+
+    override fun fetchLiveNodesData() {
+        store.accept(CollectionsStore.Intent.FetchLiveNodesData)
+    }
+
+    override fun addReplica(nodeName: String?, shardName: String, type: String) {
+        println("addReplica: $nodeName, $shardName, $type")
```
malliaridis, line 50: I believe this is left-over from debugging or not implemented yet?

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/integration/DefaultCollectionsComponent.kt
```diff
@@ -0,0 +1,74 @@
...
+    override fun deleteReplica(shardName: String, replicaName: String) {
+        store.accept(CollectionsStore.Intent.DeleteReplica(shardName, replicaName))
+    }
+
+    override fun deleteCollection(name: String) {
+        store.accept(CollectionsStore.Intent.DeleteCollection(name))
+    }
+
+    override fun reloadCollection(name: String) {
+        println("reloadCollection: $name")
```
malliaridis, line 63: Same for this `println` statement

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/integration/DefaultCollectionsComponent.kt
malliaridis, line 1: Since you are using only one store in this component implementation, you may consider importing `
import org.apache.solr.ui.components.collections.store.CollectionsStore.Intent` and reference to intents as `Intent.DeleteReplica` instead of `CollectionStore.Intent.DeleteReplica` to make the code a bit more readable.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/integration/HttpCollectionsStoreClient.kt
```diff
@@ -0,0 +1,153 @@
...
+        shardName: String,
+        type: String,
+    ): Result<JsonObject> {
+        val response = httpClient.post("api/collections/$collectionName/shards/$shardName/replicas") {
+            contentType(ContentType.Application.Json)
+            setBody(
+                buildJsonObject {
+                    nodeName?.let { put("node", it) }
+                    put("type", type)
+                },
```
malliaridis, line 69: I would instead prefer a well-defined and documented serializable data class here, something like `AddReplicaRequestBody`. This way we do not have to explicitly tell here that we are using JSON, but rather let the client's serialization strategy handle the serialization for us.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/integration/HttpCollectionsStoreClient.kt
```diff
@@ -0,0 +1,153 @@
...
+            )
+        }
+        return when {
+            response.status.isSuccess() -> {
+                println(response.body<String>())
```
malliaridis, line 74: Just another `println` here

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/integration/HttpCollectionsStoreClient.kt
```diff
@@ -0,0 +1,153 @@
...
+    ): Result<JsonObject> {
+        val response = httpClient.post("api/collections") {
+            contentType(ContentType.Application.Json)
+            setBody(
+                buildJsonObject {
+                    put("name", name)
+                    put("config", configSet)
+                    put("numShards", numShards)
+                    put("replicationFactor", replicas)
+                },
```
malliaridis, line 145: Same rule for the serialization strategy here, use here a serializable data class too.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/store/CollectionsStore.kt
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    data class State(
+        val collectionsList: CollectionsList = CollectionsList(),
+        val selectedCollectionData: CollectionData? = null,
+        val liveNodesData: List<String> = emptyList(),
+        val mutating: Boolean = false,
+        val selected: String? = null,
+        val configSets: ListConfigSets? = null,
+    )
```
malliaridis, line 77: I would also like some doc-comments here, some fields are not directly obvious, like what is mutating and what is the different between `selected` string and `selectedCollectionData`? Why do we need both?

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/store/CollectionsStore.kt
malliaridis, line 1: Nicely dcoumented the intents, I like that! 👍

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/store/CollectionsStoreProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+            initialState = State(),
+            bootstrapper = SimpleBootstrapper(Action.FetchInitialCollectionsData),
+            executorFactory = ::ExecutorImpl,
+            reducer = ReducerImpl,
+        ) {}
+
+    private sealed interface Action {
+
+        /**
+         * Action used for initiating the initial fetch of environment data.
```
malliaridis, line 46: I believe with "environment data" you mean here collection data (small copy-paste mistake)?

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/store/CollectionsStoreProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+            when (intent) {
+                is Intent.FetchCollectionList -> {
+                    fetchCollectionList()
+                }
+                is Intent.FetchCollectionData -> {
+                    fetchCollectionData(intent.name)
+                }
+                is Intent.FetchLiveNodesData -> {
+                    fetchLiveNodesData()
+                }
```
malliaridis, line 78: Does the linter apply the wrapping in `{}` or was it a preference of yours here?

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/store/CollectionsStoreProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+                is Intent.AddReplica -> {
+                    val col = state().selected
+                    if (col == null) {
+                        publish(CollectionsStore.Label.Error("No collection selected"))
```
malliaridis, line 82: Generally well-handled here, as using labels is the way to go. However, we would have issues localizing these messages later in the UI, so you would hav eto use well-defined errors that are later mapped to error messages instead. This would allow us to localize later even these texts, instead of hardcoding them here.

This of course applies for all messages that are hardcoded and in english.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/store/CollectionsStoreProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+                // TODO Add error handling
+            }
+        }
+        private fun fetchCollectionData(name: String) {
+            scope.launch {
+                withContext(ioContext) {
+                    client.getClusterStatus(name)
+                }.onSuccess {
+                    dispatch(Message.CollectionSelected(it, name))
+                    dispatch(Message.SelectedChanged(name))
```
malliaridis, line 126: Why do you distinguish between `CollectionSelected` and `SelectedChanged`? Does it maybe make sense to keep only `CollectionSelected` and update both state values with it (as it contains already all information we need)?

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/store/CollectionsStoreProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+                val s = state().selectedCollectionData?.shards[shard]
+                val allActive = s?.replicas?.values?.all { it.state.equals("ACTIVE", true) } == true
+                val hasActiveLeader = s?.replicas?.values?.any { it.leader.equals("true", true) && it.state.equals("ACTIVE", true) } == true
+                if (allActive && hasActiveLeader) return // GREEN enough
+
+                delayMs = (delayMs * 2).coerceAtMost(maxDelayMs) // backoff
+            }
+            // still not green → let UI show “Recovering…”; user can press Reload
+        }
+    }
```
malliaridis, line 255: Don't get me wrong on this one, I like your solution. However, I would probably look into a more sustainable solution, perhaps with a v2 API change request, like server-side events, that would support updating this value without polling?

Alternatively, we may have an automatic reload functionality elsewhere which could eventually update the state of shards anyway, so we would not have to manually poll here.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/store/CollectionsStoreProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+        suspend fun getCollections(): Result<CollectionsList>
+        suspend fun getClusterStatus(collectionName: String): Result<ClusterStatusResponse>
+        suspend fun getZkTree(): Result<ZkTree>
+        suspend fun addReplica(collectionName: String, nodeName: String?, shardName: String, type: String): Result<JsonObject>
+        suspend fun deleteReplica(collectionName: String, shardName: String, replicaName: String): Result<JsonObject>
+        suspend fun deleteCollection(collectionName: String): Result<JsonObject>
+        suspend fun reloadCollection(collectionName: String): Result<JsonObject>
+        suspend fun fetchConfigSets(): Result<ListConfigSets>
+        suspend fun createCollection(name: String, numShards: Int, replicas: Int, configSet: String): Result<JsonObject>
+    }
```
malliaridis, line 301: Consider documenting interface functions, what data is expected and what responses may be returned (like errors). Also avoid in all cost using `JsonObject` as response, as we do not want to limit the implementation to JSON only. This is an implementation detail not relevant in the interface.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/main/integration/DefaultMainComponent.kt
```diff
@@ -125,8 +135,7 @@ class DefaultMainComponent internal constructor(
         //     NavigationComponent.Child.Configsets(configsetsComponent(componentContext))
 
         // TODO Uncomment once Collections available
```
malliaridis, line 137: You may also remove this TODO here :)

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/shards/replica/ReplicaListItem.kt
```diff
@@ -0,0 +1,69 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import org.apache.solr.ui.components.collections.data.Replica
+
+@Composable
+fun ReplicaListItem(
+    replicaName: String,
+    rep: Replica,
```
malliaridis, line 27: I would pass only a single argument `Replica` that contains already all replica data.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/shards/replica/ReplicaListItem.kt
```diff
@@ -0,0 +1,69 @@
...
+    onDelete: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    val isLeader = rep.leader?.equals("true", ignoreCase = true) == true
+    val secondary = listOfNotNull(rep.type, rep.state)
+        .filter { it.isNotBlank() }
+        .joinToString(" • ")
```
malliaridis, line 34: With the right domain model, you would not have to execute this logic on every rendering of the component (improves slightly the performance and keeps UI elements clean of "business logic").

So to improve this, `Replica` should already have mapped whether the replica is a leader or not and store it in `isLeader` for example. This would also apply for `secondary` (not sure at this point of my review what exactly it does).

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/shards/replica/ReplicaListItem.kt
```diff
@@ -0,0 +1,69 @@
...
+                }
+            }
+        },
+        supportingContent = {
+            Column {
+                if (secondary.isNotEmpty()) Text(secondary)
+                // optional 3rd line info
+                val node = rep.nodeName ?: rep.baseUrl
+                if (!node.isNullOrBlank()) {
+                    Text(node, style = MaterialTheme.typography.bodySmall, color = MaterialTheme.colorScheme.onSurfaceVariant)
```
malliaridis, line 53: Adding some wrapping would make this a bit easier to read. And consider using either for all params names, or for none (less preferred).

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/shards/replica/ReplicaListItem.kt
```diff
@@ -0,0 +1,69 @@
...
+                }
+            }
+        },
+        trailingContent = {
+            OutlinedButton(
```
malliaridis, line 58: `OutlinedButton` are correct, they are just not styled according to our designs (if I remember correct, the ribbon / click effect is rounded). For this reason, I would recommend to define a custom component that overrides the defaults automatically. This way, we can re-use this custom UI component and avoid repeating the `contentPadding` for example.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/shards/replica/ReplicaListItem.kt
```diff
@@ -0,0 +1,69 @@
...
+                onClick = onDelete,
+                contentPadding = PaddingValues(horizontal = 10.dp, vertical = 6.dp),
```
malliaridis, line 60: Prefer using the 8-point grid system

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/shards/replica/ReplicaListItem.kt
```diff
@@ -0,0 +1,69 @@
...
+            ) {
+                Icon(Icons.Outlined.Delete, contentDescription = null)
+                Spacer(Modifier.width(6.dp))
```
malliaridis, line 63: Prefer using the 8-point grid system

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionsContent.kt
```diff
@@ -0,0 +1,155 @@
...
+    // one-off events
+    LaunchedEffect(component) {
+        component.labels.collect { label ->
+            when (label) {
+                is CollectionsStore.Label.ReplicaAdded -> snack.showSnackbar("Replica added to ${label.shard}")
+                is CollectionsStore.Label.ReplicaDeleted -> snack.showSnackbar("Replica ${label.replica} deleted")
+                is CollectionsStore.Label.Error -> snack.showSnackbar(label.message)
+                is CollectionsStore.Label.CollectionDeleted -> snack.showSnackbar("Collection ${label.name} deleted")
+                is CollectionsStore.Label.CollectionReloaded -> snack.showSnackbar("Collection ${label.name} reloaded")
+                is CollectionsStore.Label.CollectionCreated -> snack.showSnackbar("Collection ${label.name} created")
```
malliaridis, line 59: We should use string resources here instead, to avoid hardcoded localized texts.

Since the logic is the same for each label and we only have different messages displayed, we could unify it in the componenent, provide the necessary mapper function (from store label to string via string resource), and then simplify the logic here by showing the snackbar for each event fired.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionsContent.kt
```diff
@@ -0,0 +1,155 @@
...
+            }
+        }
+    }
+    Scaffold(snackbarHost = { SnackbarHost(snack) }) {
```
malliaridis, line 63: This scaffold may be moved to higher UI layer (like `MainContent` or so) to allow other composables to make use of it (and avoid overlapping of snackbars). This would have a single scaffold / snackbar responsible for all snackbar messages.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionsContent.kt
```diff
@@ -0,0 +1,155 @@
...
+import org.apache.solr.ui.views.collections.shards.ShardsMaterialCards
+
+@Composable
+fun CollectionsContent(
+    component: CollectionsComponent,
+    modifier: Modifier = Modifier,
+) {
+    val model by component.model.collectAsState()
+    val snack = remember { SnackbarHostState() }
+    var showCreateDialog by remember { mutableStateOf(false) }
```
malliaridis, line 48: Since this is a UI state value, we would instead move it to the `CollectionsComponent.model` and consume and update the state via the component.This way logic stays in the UI component, not the composable, and is saved it is not the composable that determines when to execute specific logic, but the UI component (and partly the store). 

This may seem a bit of redundant, but it ensures consistency across all UI behavior and structure. It may be opinionated though.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionsContent.kt
```diff
@@ -0,0 +1,155 @@
...
+        CreateCollectionDialog(
+            visible = showCreateDialog,
+            configSets = model.configSets?.names ?: emptyList(),
+            onFetchConfigSets = { component.fetchConfigSets() },
+            onDismiss = { showCreateDialog = false },
+            onConfirm = { name, numShards, replicas, configSet ->
+                component.createCollection(name, numShards, replicas, configSet)
+                showCreateDialog = false
+            },
+        )
```
malliaridis, line 74: Since this is a dialog that may be provided in other contexts as well, it would make sense to define its own component, and add it as child component (via ChildSlot) of `CollectionsComponent`.

This would also take some responsibility away from the all-mighty `CollectionsComponent`.

Additional dialogs that would server other purposes would then be added to the same slot as well (since only one dialog can exist simultaneously).

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionsContent.kt
```diff
@@ -0,0 +1,155 @@
...
+                color = MaterialTheme.colorScheme.outlineVariant,
+                thickness = 1.dp,
+            )
+
+            Spacer(Modifier.width(16.dp))
+
+            // Right: details panel
+            Surface(tonalElevation = 0.dp, shape = MaterialTheme.shapes.medium, modifier = Modifier.weight(1f)) {
+                Column(Modifier.fillMaxWidth().padding(12.dp)) {
+                    val d = model.selectedCollectionData
```
malliaridis, line 107: Avoid such variables or general words like `data` in code, as it makes code unreadable / unnecessary complicated down the path. E.g. here `data` is read in `d` and later used as `details` param.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionsContent.kt
```diff
@@ -0,0 +1,155 @@
...
+                        Spacer(Modifier.height(16.dp))
+
+                        ShardsMaterialCards(
+                            shards = d.shards,
+                            resetKey = d.name,
+                            liveNodes = model.liveNodesData,
+                            onFetchLiveNodes = { component.fetchLiveNodesData() },
+                            onAddReplica = { shardName, node, type -> component.addReplica(node, shardName, type) },
+                            onDeleteReplica = { shardName, replica -> component.deleteReplica(shardName, replica) },
+                        )
```
malliaridis, line 130: This can be a nice component on its own.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionsContent.kt
```diff
@@ -0,0 +1,155 @@
...
+                        Spacer(Modifier.height(8.dp))
+                        HorizontalDivider()
+                    } else {
+                        CollectionDetailsCard(
+                            details = d,
+                            isLoading = false, // TODO wire a real loading state if needed
+                            onDelete = { name -> component.deleteCollection(name) },
+                            onReload = { name -> component.reloadCollection(name) },
+                            modifier = Modifier.fillMaxWidth(),
+                        )
```
malliaridis, line 119: This could be a separate component as well.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionsContent.kt
```diff
@@ -0,0 +1,155 @@
...
+private fun AddCollectionButton(
+    onAdd: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    OutlinedButton(onClick = onAdd, modifier = modifier) {
+        Icon(Icons.Outlined.Add, contentDescription = null)
+        Spacer(Modifier.width(8.dp))
+        Text("Add Collection")
+    }
+}
```
malliaridis, line 148: Consider using a shared composable instead, and of course string resources.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionDetailsCard.kt
```diff
@@ -0,0 +1,119 @@
...
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import org.apache.solr.ui.components.collections.data.CollectionData
+
+@Composable
+fun CollectionDetailsCard(
```
malliaridis, line 36: Would be useful to document the parameters, like what string is passed to `onDelete` and `onReload`?

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionDetailsCard.kt
```diff
@@ -0,0 +1,119 @@
...
+    details: CollectionData,
+    isLoading: Boolean,
+    onDelete: (String) -> Unit,
+    onReload: (String) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    var showDelete by remember { mutableStateOf(false) }
```
malliaridis, line 43: This value can be moved to UI state (model) if a component is introduced here.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionDetailsCard.kt
```diff
@@ -0,0 +1,119 @@
...
+        modifier = modifier,
+        shape = MaterialTheme.shapes.medium,
+        colors = CardDefaults.outlinedCardColors(),
+    ) {
+        // Header
+        Row(
+            modifier = Modifier.fillMaxWidth().padding(horizontal = 12.dp, vertical = 10.dp),
+            verticalAlignment = Alignment.CenterVertically,
+        ) {
+            Text("Collection: ${details.name}", style = MaterialTheme.typography.headlineSmall, modifier = Modifier.weight(1f))
```
malliaridis, line 55: Use for all parameters names, keeps code conventions in tact. Also prefer writing parameters in new lines when more than 2 are passed (typical Compose conventions).

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionDetailsCard.kt
```diff
@@ -0,0 +1,119 @@
...
+private fun KVItem(label: String, value: String) {
+    ListItem(
+        headlineContent = { Text(label) },
+        trailingContent = { Text(value, style = MaterialTheme.typography.bodyLarge) },
+        modifier = Modifier.padding(horizontal = 8.dp),
+    )
+}
+
+@Composable
+private fun StatPill(text: String) {
```
malliaridis, line 106: What exactly is a "pill"? This term seems less common in the UI. From content it looks like a chip / badge / label?

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionDetailsCard.kt
```diff
@@ -0,0 +1,119 @@
...
+        KVItem("Shard count", details.shardCount.toString())
+        KVItem("configName", details.configName)
+        KVItem("replicationFactor", details.replicationFactor.toString())
+        KVItem("router", details.routerName)
+        Spacer(Modifier.height(4.dp))
+    }
+}
+
+@Composable
+private fun KVItem(label: String, value: String) {
```
malliaridis, line 97: Avoid shortening words for better readability. So prefer written out form as `KeyValueItem` instead.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionDetailsCard.kt
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+
+        HorizontalDivider()
+
+        // Key/Value lines
+        KVItem("Shard count", details.shardCount.toString())
+        KVItem("configName", details.configName)
+        KVItem("replicationFactor", details.replicationFactor.toString())
+        KVItem("router", details.routerName)
```
malliaridis, line 91: You may replace strings with string resources to allow later localizations.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionDetailsCard.kt
```diff
@@ -0,0 +1,119 @@
...
+                Text("Delete")
+            }
+            DeleteCollectionDialog(
+                open = showDelete,
+                collectionName = details.name,
+                onDismiss = { showDelete = false },
+                onConfirm = { onDelete(details.name) },
+            )
+            Spacer(Modifier.width(8.dp))
+            OutlinedButton(onClick = { onReload(details.name) }, enabled = !isLoading, contentPadding = PaddingValues(horizontal = 12.dp, vertical = 6.dp)) {
```
malliaridis, line 78: Prefer params in separate lines when many.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CollectionsListCard.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.solr.ui.views.theme.backgroundLight
+
+@Composable
+fun CollectionsListCard(
+    names: List<String>,
+    selected: String?,
+    onSelect: (String) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    LazyColumn(
```
malliaridis, line 24: You can directly assign a composable when only one composable function is in the body of the function, like below:

```kotlin
fun FunctionName(
  // ...
) = LazyColumn(
  // ...
) { ... }
```

This reduces indentation and makes function look more like aliases.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/shards/replica/AddReplicaDialog.kt
```diff
@@ -0,0 +1,108 @@
...
+@Composable
+fun AddReplicaDialog(
+    visible: Boolean,
+    shardName: String,
+    liveNodes: List<String>,
+    onFetchLiveNodes: () -> Unit,
+    onDismiss: () -> Unit,
+    onConfirm: (node: String?, type: String) -> Unit,
+) {
+    if (!visible) return
```
malliaridis, line 37: I would move the visibility parameter out of this function and only call the composable if `visible` is true. It should not be responsibility of the component / composable itself if it is displayed, but the responsibility of the parent.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/shards/replica/AddReplicaDialog.kt
```diff
@@ -0,0 +1,108 @@
...
+    // keep most recent lambda reference across recompositions
+    val fetchLiveNodes by rememberUpdatedState(onFetchLiveNodes)
+    // fetch nodes when the dialog opens
+    LaunchedEffect(visible) {
+        if (visible) fetchLiveNodes()
+    }
+    var nodeExpanded by remember { mutableStateOf(false) }
+    var typeExpanded by remember { mutableStateOf(false) }
+    var selectedNode by remember { mutableStateOf<String?>(null) } // null => no specified node
+    var selectedType by remember { mutableStateOf(ReplicaType.NRT) }
```
malliaridis, line 48: All these belong to a UI state (model) and could make up a neat component.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/shards/replica/AddReplicaDialog.kt
```diff
@@ -0,0 +1,108 @@
...
+    onConfirm: (node: String?, type: String) -> Unit,
+) {
+    if (!visible) return
+
+    // keep most recent lambda reference across recompositions
+    val fetchLiveNodes by rememberUpdatedState(onFetchLiveNodes)
+    // fetch nodes when the dialog opens
+    LaunchedEffect(visible) {
+        if (visible) fetchLiveNodes()
+    }
```
malliaridis, line 44: This logic should not be called in the composable, but rather the UI component.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/shards/replica/AddReplicaDialog.kt
```diff
@@ -0,0 +1,108 @@
...
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberUpdatedState
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+
+enum class ReplicaType(val solr: String) { NRT("NRT"), TLOG("TLOG"), PULL("PULL") }
```
malliaridis, line 25: I would not store enum class in composable files. Write them out instead and you may also remove the redundant solr variable if they exactly match the enum values (enums can be looked up by a string using here `ReplicaType.valueOf(string)`.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/shards/HealthDot.kt
```diff
@@ -0,0 +1,31 @@
...
+
+@Composable
+fun HealthDot(status: String?, modifier: Modifier = Modifier, size: Dp = 10.dp) {
+    val s = status?.trim()?.lowercase()
+    val color = when (s) {
+        "green", "healthy", "good" -> HealthGreen
+        "yellow", "degraded", "recovering" -> HealthYellow
+        "red", "down", "failed" -> HealthRed
+        else -> MaterialTheme.colorScheme.outlineVariant
+    }
```
malliaridis, line 25: I would move this logic out to a mapper function and pass a Status enum here instead, so that UI stays clean of business logic.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/shards/HealthDot.kt
```diff
@@ -0,0 +1,31 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.drawscope.Stroke
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+
+private val HealthGreen = Color(0xFF2E7D32) // md green 700
+private val HealthYellow = Color(0xFFF9A825) // md amber 800
+private val HealthRed = Color(0xFFC62828) // md red 800
```
malliaridis, line 15: Color values should be part of the theme, so that switching between light and dark theme would have optimized colors based on the theme mode currently active.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CreateCollectionDialog.kt
```diff
@@ -0,0 +1,120 @@
...
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun CreateCollectionDialog(
+    visible: Boolean,
+    configSets: List<String>,
+    onFetchConfigSets: () -> Unit,
+    onDismiss: () -> Unit,
+    onConfirm: (name: String, numShards: Int, replicas: Int, configSet: String) -> Unit,
+) {
+    if (!visible) return
```
malliaridis, line 39: Remove visible parameter and let the parent / caller of this composable decide wether the composable should be rendered / called or not.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CreateCollectionDialog.kt
```diff
@@ -0,0 +1,120 @@
...
+
+    val fetch by rememberUpdatedState(onFetchConfigSets)
+    LaunchedEffect(visible) {
+        if (visible) fetch()
+    }
```
malliaridis, line 44: Launched effect should be part of the UI component. `LaunchedEffect` in Composables is recommended with our current stack usually when using animations, since data fetching should be handled by UI components and stores (via Bootstrapper).

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/CreateCollectionDialog.kt
```diff
@@ -0,0 +1,120 @@
...
+    var shardsText by remember { mutableStateOf("1") }
+    var replicasText by remember { mutableStateOf("1") }
+
+    var expanded by remember { mutableStateOf(false) }
+    var selectedConfig by remember { mutableStateOf<String?>(null) }
+
+    val shards = shardsText.toIntOrNull() ?: -1
+    val replicas = replicasText.toIntOrNull() ?: -1
+
+    val canCreate = name.isNotBlank() && shards >= 1 && replicas >= 1 && !selectedConfig.isNullOrBlank()
```
malliaridis, line 56: All these values would fit well into a component with a store, where the mapper would contain part of the logic. For example:

```kotlin
// Store state data class
data class State(
    selected: String? = null,
    name: String = "",
    shards: Int = 1,
    replicas: Int = 1,
    expanded: Boolean = false,
)

// Component model data class
data class Model(
    name: String = "",
    shardsText = "1", // may also be just int
    replicasText = "1", // may also be just int
    canCreate: Boolean = false,
)

// Mapper function
State.toModel(): Model = Model(
    name = it.name,
    shardsText = it.shards.toString(),
    replicasText = it.replicas.toString(),
    canCreate = it.name.isNotBlank() &&
        it.shards >= 1 &&
        it.replicas >= 1 &&
        !it.selected.isNullOrBlank()
)
```

I hope this clarifies a bit the roles of stores, components and mapper functions.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/DeleteCollectionDialog.kt
```diff
@@ -0,0 +1,83 @@
...
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun DeleteCollectionDialog(
+    open: Boolean,
+    collectionName: String,
+    onDismiss: () -> Unit,
+    onConfirm: () -> Unit,
+) {
+    if (!open) return
```
malliaridis, line 34: May be responsibility of parent, so parameter becomes redundant.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/DeleteCollectionDialog.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    AlertDialog(
+        onDismissRequest = onDismiss,
+        confirmButton = {
+            Button(
+                enabled = matches,
+                colors = ButtonDefaults.buttonColors(containerColor = MaterialTheme.colorScheme.error),
+                onClick = {
+                    onConfirm()
+                    onDismiss()
```
malliaridis, line 47: `onConfirm` should ideally dismiss dialog at the end, but only if successful. So calling both in a row may be a bit problematic in case of errors.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/collections/shards/ShardsMaterialCards.kt
```diff
@@ -0,0 +1,104 @@
...
+            entry.key.replace("shard", "").toIntOrNull() ?: Int.MAX_VALUE
+        }
+    }
+
+    var expanded by remember(resetKey) { mutableStateOf<String?>(null) }
+
+    // optional guard if a shard disappears
+    LaunchedEffect(shards, expanded) {
+        if (expanded != null && expanded !in shards.keys) expanded = null
+    }
```
malliaridis, line 63: This one seems a bit tricky, like a workaround for a problem that may be solved differently? 🤔

malliaridis (commented):
Impresive how much you were able to re-use and adapt to the existing format without being familiar with this stack and without any special briefing. Good job! You also made use of the difficult parts like `LaunchedEffect` and `CompositionLocalProvider` 👏 

I have some general remarks that I noticed during the review, so I will go a bit here into detail (see also review comments). Take them more like my personal preferences and practices that I try to apply everywhere, so if you see things different let us talk about it:

_**Large Component with many responsibilities**_

I think the most obvious part here is in my opinion the large `CollectionsComponent`. It holds many responsibilities and is very "large". I believe that if you would split it into multiple components and have one component with multiple child slots, it would become much easier both for reviewing, and later testing. I have added a couple comments in places where I believe introducing a component would make sense.

_**Distinguishing between domain and data models**_

Generally I started using the term "data models" and the package `data` for data classes that are API specific, and domain models and the package `domain` for UI-only data classes. The reason I strongly distinguish them is to create a clear abstraction between API and domain models, so that on API changes or migrations, we would not have to update the entire UI that uses the altered models. The clear abstraction of those does also allow us to define and implement the UI as we please and write separate tests for UI.

Rule of thumb that I try to follow here: Never expect an API / data model in a store client, and never return one in the result.

_**Usage of string resources**_

Although this was discussed before as "unnecessary", I want to use string resources wherever localized texts are displayed. I believe and I think that we will at some point be able to translate the app and provide it in other languages, and therefore it would be helpful if we already have used everywhere string resources, rather than hardcoded english texts / strings. So a few comments reference this switch to string resources.

_**8-point Grid System**_

In UI it is common practice to follow the 8-point grid system, especially in the Material theme. The 8-point grid system expect values like spacings, paddings, margins, box sizes etc. to all use a multiply of 8 points. So value like 6 or 10 are not allowed, and would be instead 8 and 16. In some cases multiplies of 4 are also acceptable. 

You can find a few articles talking about the benefits of it online if you are new in UI development.

_**Avoid state and logic in composables**_

Compose comes with many built-in features that make some of the libraries we use "redundant". However, we use the libraries for structuring our code and have clear definitions and boundaries. I believe that you have read a lot of examples and documentation from the official libraries of Google and JetBrains, so it is expected to follow practices from there.

However, we explicitly use `Component` interfaces to define the inputs, outputs and actions of composables (with some exceptions for very simple composables), so that we do not execute any logic in the UI / inside the composables. Some of my comments already mention such cases, and I personally recommend to use in these cases `Component`s instead. Theoretically, you would have to write UI tests that test exactly this logic in UI, but that becomes redundant and is moved to integration tests when we move the business logic to component implementations.

_**More doc comments**_

In many cases parameters and properties like function parameters with arguments, booleans, optionals and simple strings may not be very obvious what they store or how they are used. You added nice doc comments to the intents in the store, and I would like such comments in Composables, interfaces and interface properties / functions. You often can avoid adding doc comments to implementations of interfaces, so it would be enough when documenting at the right place (interfaces in this case).

---

Last but not least: Let us talk. I believe that your learnings and experience with my learnings and experience would add up well together and we can do great things once we are in sync. 😄

malliaridis:
> @malliaridis Working on test cases now. JIRA is missing I believe. Can you confirm?

And about that. Yes. [The designs](https://www.figma.com/design/VdbEfcWQ8mirFNquBzbPk2/Apache-Solr-Admin-UI-v2-Concept?node-id=14-3470&t=eixJKaZsYD5vsT3s-1) for the collections were not finished yet, so I couldn't specify the requirements and behavior yet.

malliaridis (commented):
Two more additions after running and testing the UI:
- I think it would make sense to create the designs together, and adjust he look-and-feel of the screen to the ones that match the overall app design.
- We would also cover responsiveness of the design, which currently does not scale properly if the window / screen is smaller than expected.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/theme/Theme.kt
iamsanjay: I like light theme. Is there anyway we could change this via UI?

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/CollectionsComponent.kt
iamsanjay: ohh I see for instance, fetchCollectionList is a bootstrap because we need list at the beginning. And fetchLiveNodesData is more like getLiveNodesData.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/data/ClusterStatusModels.kt
iamsanjay: +1 to separate file for data classes and common data classes. 
Comments are sometimes become hard-to-manage, I'd like comment where it does not make sense to dev, but for instance if the field name is replicationFactor then we know what it represents. Same goes for any functions such as getCollectionData/reloadCollection.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/collections/CollectionsComponent.kt
iamsanjay: 1. We can move Labels out from CollectionStore 
2. Or, create another set of Labels and use mappers to map

PR #3512, opened 2025-08-26
SOLR-17875: Rationalize bootstrap_conf, bootstrap_confdir, and -Dcollection.configName in start up scripts

epugh:
https://issues.apache.org/jira/browse/SOLR-17875

# Description

Migrate the bootstrap_conf and bootstrap_confdir properties, and deal with collection.configName.   Get some bats tests in to prove they work.

Investigate ConfigSetService and CreateCollectionCmd about their use of the system properties.

# Solution

# Tests

new bats tests

### DIFF
--- a/solr/packaging/test/test_start_solr.bats
+++ b/solr/packaging/test/test_start_solr.bats
@@ -93,3 +93,17 @@ teardown() {
   solr start --jettyconfig "--module=server"
   solr assert --started http://localhost:${SOLR_PORT} --timeout 5000
 }
+
+@test "bootstrap configset using bootstrap_confdir and collection.configName" {
+  local confdir_path="${SOLR_TIP}/server/solr/configsets/sample_techproducts_configs/conf"
+  
+  # Verify the source configset directory exists
+  test -d "${confdir_path}"
+  
+  # Start Solr with bootstrap_confdir pointing to techproducts configset
+  solr start -Dbootstrap_confdir="${confdir_path}" -Dcollection.configName=techproducts
+  solr assert --started http://localhost:${SOLR_PORT} --timeout 5000
+  
+  # Verify the techproducts configset was uploaded
+  config_exists "techproducts"
+}
--- a/solr/packaging/test/test_start_solr.bats
+++ b/solr/packaging/test/test_start_solr.bats
@@ -101,9 +101,15 @@ teardown() {
   test -d "${confdir_path}"
   
   # Start Solr with bootstrap_confdir pointing to techproducts configset
+  # bootstrap_confdir: path to config directory to upload
+  # collection.configName: name to give the uploaded configset (default: "configuration1")
+  # This uploads the sample_techproducts_configs/conf directory as "techproducts" configset
   solr start -Dbootstrap_confdir="${confdir_path}" -Dcollection.configName=techproducts
   solr assert --started http://localhost:${SOLR_PORT} --timeout 5000
   
-  # Verify the techproducts configset was uploaded
+  # Allow time for the configset upload to complete
+  sleep 1
+  
+  # Verify the techproducts configset was uploaded to ZooKeeper
   config_exists "techproducts"
 }


solr/bin/solr.cmd
```diff
@@ -936,10 +936,6 @@ IF "%SOLR_MODE%"=="solrcloud" (
     set "CLOUD_MODE_OPTS=!CLOUD_MODE_OPTS! -Dsolr.solrxml.required=true"
   )
 
-  IF EXIST "%SOLR_HOME%\collection1\core.properties" set "CLOUD_MODE_OPTS=!CLOUD_MODE_OPTS! -Dbootstrap_confdir=./solr/collection1/conf -Dcollection.configName=myconf -DnumShards=1"
```
epugh, line 939: @malliaridis I think you made the last change in here.   Can you look at this?  Basically, I believe this hardcoded "collection1" logic is no longer needed, nor is it desirable to have a magic collection.   I believe this can be removed from both of our start scripts...

dsmiley: +1 definitely don't want to see "collection1" in our startup scripts

solr/packaging/test/test_start_solr.bats
```diff
@@ -93,3 +93,17 @@ teardown() {
...
+@test "bootstrap configset using bootstrap_confdir and collection.configName" {
+  local confdir_path="${SOLR_TIP}/server/solr/configsets/sample_techproducts_configs/conf"
+  test -d "${confdir_path}"
+  
+  # This uploads the sample_techproducts_configs/conf directory as "techproducts_bootstrapped" configset
+  solr start -Dbootstrap_confdir="${confdir_path}" -Dcollection.configName=techproducts_bootstrapped
+  solr assert --started http://localhost:${SOLR_PORT} --timeout 5000
+  
+  # Allow time for the configset upload to complete
+  sleep 1
```
dsmiley: Solr does not upload these asynchronously.  If Solr has started, the configset should be there

solr/packaging/test/test_start_solr.bats
```diff
@@ -93,3 +93,17 @@ teardown() {
   solr start --jettyconfig "--module=server"
   solr assert --started http://localhost:${SOLR_PORT} --timeout 5000
 }
+
+@test "bootstrap configset using bootstrap_confdir and collection.configName" {
+  local confdir_path="${SOLR_TIP}/server/solr/configsets/sample_techproducts_configs/conf"
+  test -d "${confdir_path}"
+  
+  # This uploads the sample_techproducts_configs/conf directory as "techproducts_bootstrapped" configset
+  solr start -Dbootstrap_confdir="${confdir_path}" -Dcollection.configName=techproducts_bootstrapped
```
dsmiley: I was hoping you would rename these given the new system property / environment var naming scheme

solr/bin/solr.cmd
malliaridis: I tried to get some information from the git history. This addition is quite old, and by looking up "collection1" in our project, we have a lot of tests using this keyword / collection name. I am not sure if any of the tests may still need it.

I am all for removing it, but we should make sure that all these references are not making automatic use of it by having a file present in cloud mode and by using the CLI script.

epugh: i will dig a bit more....  thanks @malliaridis!

solr/packaging/test/test_start_solr.bats
epugh: that will be in the next pr ;-)

epugh: err next push!

epugh:
I think this one is ready!

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
```diff
@@ -85,22 +84,17 @@ private static ConfigSetService instantiate(CoreContainer coreContainer) {
...
-    // bootstrap _default conf, bootstrap_confdir and bootstrap_conf if provided via system property
+    // bootstrap _default conf and solr.configset.bootstrap.confdir if specified.
     try {
       // _default conf
       bootstrapDefaultConf();
 
-      // bootstrap_confdir
-      String confDir = System.getProperty("bootstrap_confdir");
+      // solr.configset.bootstrap.confdir
+      String confDir = System.getProperty("solr.configset.bootstrap.confdir");
```
dsmiley: no EnvUtils?  It would get us ENV var usage which is highly desirable for these

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
```diff
@@ -122,15 +116,19 @@ private void bootstrapDefaultConf() throws IOException {
   }
 
   private void bootstrapConfDir(String confDir) throws IOException {
-    Path configPath = Path.of(confDir);
+    if (!confDir.endsWith("conf")) {
```
dsmiley, line 120: Why mandate this?  People should be able to put their conf on disk wherever they want.

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
```diff
@@ -122,15 +116,19 @@ private void bootstrapDefaultConf() throws IOException {
...
     if (!Files.isDirectory(configPath)) {
       throw new IllegalArgumentException(
-          "bootstrap_confdir must be a directory of configuration files, configPath: "
+          "solr.configset.bootstrap.confdir must be a directory of configuration files, configPath: "
               + configPath);
     }
-    String confName =
-        System.getProperty(
-            ZkController.COLLECTION_PARAM_PREFIX + ZkController.CONFIGNAME_PROP, "configuration1");
+    String confName = System.getProperty("solr.collection.config.name", "configuration1");
```
dsmiley: again; EnvUtils

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
```diff
@@ -164,6 +162,32 @@ public static Path getDefaultConfigDirPath() {
...
+   * @param pathStr The path string to resolve
+   * @return The resolved Path object
+   * @see SolrDispatchFilter#SOLR_INSTALL_DIR_ATTRIBUTE
+   */
+  public static Path resolvePathWithSolrInstallDir(String pathStr) {
+    Path path = Path.of(pathStr);
+
+    // Convert to absolute path if it's relative and solr.install.dir is set
+    if (!path.isAbsolute()) {
+      String installDir = System.getProperty(SolrDispatchFilter.SOLR_INSTALL_DIR_ATTRIBUTE);
```
dsmiley: Can we mandate that this system property is set?  In the context of use of this utility method, I think/hope that's reasonable.

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
```diff
@@ -164,6 +162,32 @@ public static Path getDefaultConfigDirPath() {
...
   }
 
+  /**
+   * Resolves a path string into a Path object, handling both absolute and relative paths. If the
+   * path is relative and solr.install.dir system property is set, the path is resolved against the
+   * Solr installation directory. Otherwise, it's converted to an absolute path based on the current
+   * working directory.
+   *
+   * @param pathStr The path string to resolve
+   * @return The resolved Path object
```
dsmiley, line 172: you can say it's absolute and normalized

solr/packaging/test/test_start_solr.bats
epugh: I just figured out what you mean by this comment...  thanks!

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
epugh: I accidentally uploaded the parent directly as a configset, which obviously wound't work.   So I thought this woiuld add some protection from uploading the wworng directlry.   If you think it shouldn't be kept, then happy to remove it.

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
epugh: right!

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
epugh: Yep, makes sense.

solr/core/src/java/org/apache/solr/core/ConfigSetService.java
dsmiley: I can understand your intent but please remote the limitation.

What would be helpful is to detect if the directory provided itself contains a "conf" directory.  If it does, use that.  I think I've seen that kind of logic elsewhere but I could be wrong.

solr/solrj/src/resources/DeprecatedSystemPropertyMappings.properties
```diff
@@ -28,3 +28,5 @@ basicauth=solr.security.auth.basicauth.credentials
 cloud.solr.client.max.stale.retries=solr.solrj.cloud.max.stale.retries
 configset.upload.enabled=solr.configset.upload.enabled
 solr.hidden.sys.props=solr.responses.hidden.sys.props
+bootstrap_confdir=solr.configset.bootstrap.confdir
+collection.config.name=solr.collection.config.name
```
dsmiley, line 32: The actual system property that was used is "collection.configName" not "collection.config.name".

epugh: The entries in this file are post the normalization process that converts camel case to dot seperated.

epugh:
I have struggled to get a clean crave run, so if it runs locally I may just go ahead and merge and monitor the Jenkins job.   Need to keep pushing to get these done before Solr ten!

PR #3513, opened 2025-08-26
SOLR-17631: use Lucene102Codec instead of Lucene101Codec

cpoerschke:
#3053 follow-up, ref. https://github.com/apache/solr/pull/3053/files#r2050654278 thread.

https://issues.apache.org/jira/browse/SOLR-17631


cpoerschke:
oops silly me, there is no `Lucene102Codec` ... https://github.com/apache/lucene/tree/releases/lucene/10.2.2/lucene/core/src/java/org/apache/lucene/codecs/lucene102

PR #3517, opened 2025-08-27
Close OTEL Observable instruments

mlbiscoc:
No JIRA. Observable instruments `Gauge/Counters` closeable unlike synchronous instruments which aren't. They will close in the case of the `Meter Providers` closing but we should close them as well when we can.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/SolrCore.java
+++ b/solr/core/src/java/org/apache/solr/core/SolrCore.java
@@ -24,6 +24,7 @@ import static org.apache.solr.metrics.SolrCoreMetricManager.CORE_ATTR;
 import static org.apache.solr.metrics.SolrCoreMetricManager.SHARD_ATTR;
 
 import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.metrics.ObservableLongGauge;
 import java.io.Closeable;
 import java.io.FileNotFoundException;
 import java.io.IOException;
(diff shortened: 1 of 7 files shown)


solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -261,6 +262,12 @@ public class SolrCore implements SolrInfoBean, Closeable {
   private AttributedLongCounter newSearcherOtherErrorsCounter;
   private AttributedLongTimer newSearcherTimer;
   private AttributedLongTimer newSearcherWarmupTimer;
+  private ObservableLongGauge refCounter;
+  private ObservableLongGauge diskSpaceGauge;
+  private ObservableLongGauge indexSizeGauge;
+  private ObservableLongGauge segmentCounter;
+  private ObservableLongGauge coreStartTimeGauge;
+  private ObservableLongGauge coreIsLeader;
```
dsmiley: If these are only here to close; maybe we just have a `List<AutoCloseable> closeables`.  I've thought the same for many tests, for that matter.

dsmiley: And I encourage you to take code from Lucene's IOUtils to make this easy and add to Solr's IOUtils.

solr/core/src/java/org/apache/solr/core/SolrCore.java
```diff
@@ -1450,18 +1445,20 @@ public void initializeMetrics(
...
               "solr_core_is_leader",
               "Indicates whether this Solr core is currently the leader",
               (observableLongMeasurement -> {
                 observableLongMeasurement.record(
                     (coreDescriptor.getCloudDescriptor().isLeader()) ? 1 : 0,
                     baseGaugeCoreAttributes);
-              }));
+              })));
+
+    this.toClose = List.copyOf(observables);
```
dsmiley: why copy it?  I could see maybe using an unmodifiable wrapper but at least right now there's no point.

PR #3518, opened 2025-08-27
QueryRequest should only be for SearchHandler

dsmiley:
Internal refactoring.  There are many places where QueryRequest is being used for requests that are *not* to a SearchHandler.  That is an abuse/misuse.  It explains some portion of the use of "qt" that can be sidestepped via explicit path to `GenericSolrRequest`.

This PR also adds javadoc to QueryRequest to make it's intended use explicit.

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
+++ b/solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
@@ -89,11 +89,12 @@ import org.apache.lucene.store.IOContext;
 import org.apache.lucene.store.IndexInput;
 import org.apache.lucene.store.IndexOutput;
 import org.apache.solr.client.api.model.FileMetaData;
+import org.apache.solr.client.solrj.SolrRequest;
 import org.apache.solr.client.solrj.SolrServerException;
 import org.apache.solr.client.solrj.impl.Http2SolrClient;
 import org.apache.solr.client.solrj.impl.HttpClientUtil;
 import org.apache.solr.client.solrj.impl.InputStreamResponseParser;
-import org.apache.solr.client.solrj.request.QueryRequest;
+import org.apache.solr.client.solrj.request.GenericSolrRequest;
 import org.apache.solr.cloud.CloudDescriptor;
 import org.apache.solr.cloud.ZkController;
 import org.apache.solr.common.SolrException;
@@ -345,8 +346,13 @@ public class IndexFetcher {
     ModifiableSolrParams params = new ModifiableSolrParams();
     params.set(COMMAND, CMD_INDEX_VERSION);
     params.set(CommonParams.WT, JAVABIN);
-    params.set(CommonParams.QT, ReplicationHandler.PATH);
-    QueryRequest req = new QueryRequest(params);
+    var req =
+        new GenericSolrRequest(
+                SolrRequest.METHOD.GET,
+                ReplicationHandler.PATH,
+                SolrRequest.SolrRequestType.ADMIN,
+                params)
+            .setRequiresCollection(true);
     try {
       return solrClient.requestWithBaseUrl(leaderBaseUrl, leaderCoreName, req).getResponse();
     } catch (SolrServerException e) {
@@ -364,8 +370,13 @@ public class IndexFetcher {
     params.set(COMMAND, CMD_GET_FILE_LIST);
     params.set(GENERATION, String.valueOf(gen));
     params.set(CommonParams.WT, JAVABIN);
-    params.set(CommonParams.QT, ReplicationHandler.PATH);
-    QueryRequest req = new QueryRequest(params);
+    var req =
+        new GenericSolrRequest(
+                SolrRequest.METHOD.GET,
+                ReplicationHandler.PATH,
+                SolrRequest.SolrRequestType.ADMIN,
+                params)
+            .setRequiresCollection(true);
     try {
       NamedList<?> response =
           solrClient.requestWithBaseUrl(leaderBaseUrl, leaderCoreName, req).getResponse();
@@ -1876,7 +1887,6 @@ public class IndexFetcher {
       // the method is command=filecontent
       params.set(COMMAND, CMD_GET_FILE);
       params.set(GENERATION, Long.toString(indexGen));
-      params.set(CommonParams.QT, ReplicationHandler.PATH);
       // add the version to download. This is used to reserve the download
       params.set(solrParamOutput, fileName);
       if (useInternalCompression) {
@@ -1896,9 +1906,14 @@ public class IndexFetcher {
 
       NamedList<?> response;
       InputStream is = null;
-      // TODO use shardhandler
       try {
-        QueryRequest req = new QueryRequest(params);
+        var req =
+            new GenericSolrRequest(
+                    SolrRequest.METHOD.GET,
+                    ReplicationHandler.PATH,
+                    SolrRequest.SolrRequestType.ADMIN,
+                    params)
+                .setRequiresCollection(true);
         req.setResponseParser(new InputStreamResponseParser(FILE_STREAM));
         if (useExternalCompression) req.addHeader("Accept-Encoding", "gzip");
         response = solrClient.requestWithBaseUrl(leaderBaseUrl, leaderCoreName, req).getResponse();
@@ -2042,10 +2057,14 @@ public class IndexFetcher {
     ModifiableSolrParams params = new ModifiableSolrParams();
     params.set(COMMAND, CMD_DETAILS);
     params.set("follower", false);
-    params.set(CommonParams.QT, ReplicationHandler.PATH);
 
-    QueryRequest request = new QueryRequest(params);
-    // TODO use shardhandler
+    var request =
+        new GenericSolrRequest(
+                SolrRequest.METHOD.GET,
+                ReplicationHandler.PATH,
+                SolrRequest.SolrRequestType.ADMIN,
+                params)
+            .setRequiresCollection(true);
     return solrClient.requestWithBaseUrl(leaderBaseUrl, leaderCoreName, request).getResponse();
   }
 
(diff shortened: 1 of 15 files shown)


solr/solrj/src/java/org/apache/solr/client/solrj/request/QueryRequest.java
```diff
@@ -23,6 +23,9 @@
 import org.apache.solr.common.util.NamedList;
 
 /**
+ * For use with Solr's {@code SearchHandler}, generally at "/select". For other handlers, try {@link
+ * GenericSolrRequest}.
+ *
```
cpoerschke, line 28: Wondering if one might expect the same for `QueryResponse` (except there is no `GenericSolrResponse` currently it seems).

solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
```diff
@@ -2042,10 +2057,14 @@ NamedList<Object> getDetails() throws IOException, SolrServerException {
...
 
-    QueryRequest request = new QueryRequest(params);
-    // TODO use shardhandler
+    var request =
+        new GenericSolrRequest(
+                SolrRequest.METHOD.GET,
+                ReplicationHandler.PATH,
+                SolrRequest.SolrRequestType.ADMIN,
+                params)
+            .setRequiresCollection(true);
```
cpoerschke: subjective: same/similar lines four times, perhaps a small private static utility function would be nice

solr/core/src/java/org/apache/solr/update/PeerSyncWithLeader.java
```diff
@@ -331,10 +329,12 @@ private boolean handleUpdates(
...
   }
 
-  private NamedList<Object> request(ModifiableSolrParams params, String onFail) {
+  private NamedList<Object> doRtgRequest(ModifiableSolrParams params, String onFail) {
     try {
-      QueryRequest request = new QueryRequest(params, SolrRequest.METHOD.POST);
-      QueryResponse rsp = clientToLeader.requestWithBaseUrl(leaderBaseUrl, coreName, request);
+      var request =
+          new GenericSolrRequest(
+              SolrRequest.METHOD.GET, "/get", SolrRequest.SolrRequestType.QUERY, params).setRequiresCollection(true);
```
cpoerschke: precommit
```suggestion
                  SolrRequest.METHOD.GET, "/get", SolrRequest.SolrRequestType.QUERY, params)
              .setRequiresCollection(true);
```

solr/solrj/src/java/org/apache/solr/client/solrj/request/QueryRequest.java
dsmiley: Yes; I could clarify in the final commit message that this is about the pairing of both QueryRequest/QueryResponse.  QueryResponse is filled with things pertaining to SearchHandler that has no business being used for the places in this PR.

solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
dsmiley: The whole thing is verbose everywhere honestly; I'd rather see a wholistic solution than tackling it just here.

dsmiley:
I fixed the tests; this one should be ready to go.  Will merge Tuesday, or sooner if I get an approving review.

solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
gerlowskija: > I'd rather see a holistic solution than tackling it just here

This feels a bit like "letting the 'perfect' be the enemy of the 'good'" to me.  Sure, there's refactoring needed elsewhere, but if we ignore DRY while we wait for some other PR to do that refactor then the code will suffer in the interim (albeit in a tiny, small way).

If you're open to Christine's suggestion but just don't want to spend the time on it personally, lmk and I'm happy to collaborate on your branch here.  (If so, tag me or ping in Slack)

solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
```diff
@@ -345,8 +346,13 @@ public NamedList<Object> getLatestVersion() throws IOException {
     ModifiableSolrParams params = new ModifiableSolrParams();
     params.set(COMMAND, CMD_INDEX_VERSION);
     params.set(CommonParams.WT, JAVABIN);
-    params.set(CommonParams.QT, ReplicationHandler.PATH);
-    QueryRequest req = new QueryRequest(params);
+    var req =
```
gerlowskija: [Q] SolrJ has a SolrRequest for the v2 equivalent of this API: ReplicationApi.FetchIndexVersion.  If this code is targeting 10.0 only, wdyt of using that here instead of GSR?

solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
```diff
@@ -364,8 +370,13 @@ private void fetchFileList(long gen) throws IOException {
     params.set(COMMAND, CMD_GET_FILE_LIST);
     params.set(GENERATION, String.valueOf(gen));
     params.set(CommonParams.WT, JAVABIN);
-    params.set(CommonParams.QT, ReplicationHandler.PATH);
-    QueryRequest req = new QueryRequest(params);
+    var req =
```
gerlowskija: [Q] SolrJ has a SolrRequest for the v2 equivalent of this API: ReplicationApi.FetchFileList. If this code is targeting 10.0 only, wdyt of using that here instead of GSR?

solr/core/src/test/org/apache/solr/handler/TestReplicationHandler.java
```diff
@@ -220,9 +222,10 @@ private NamedList<Object> getIndexVersion(SolrClient s) throws Exception {
     ModifiableSolrParams params = new ModifiableSolrParams();
     params.set("command", "indexversion");
     params.set("_trace", "getIndexVersion");
-    params.set("qt", ReplicationHandler.PATH);
-    QueryRequest req = new QueryRequest(params);
-
+    var req =
```
gerlowskija: Ditto my comment above about there being v2 coverage for this API in SolrJ: ReplicationApi.FetchIndexVersion

solr/core/src/test/org/apache/solr/handler/TestReplicationHandler.java
```diff
@@ -234,8 +237,9 @@ private void reloadCore(JettySolrRunner jettySolrRunner, String core) throws Exc
     ModifiableSolrParams params = new ModifiableSolrParams();
     params.set("action", "reload");
     params.set("core", core);
-    params.set("qt", "/admin/cores");
-    QueryRequest req = new QueryRequest(params);
+    var req =
```
gerlowskija, line 235: [Q] SolrJ has coverage for the v2 version of this API: CoresApi.ReloadCore.  If this PR is intended for 10.0 only, wdyt about using that here instead of GSR?

solr/core/src/test/org/apache/solr/handler/admin/ShowFileRequestHandlerTest.java
```diff
@@ -51,8 +53,13 @@ public static void beforeTest() throws Exception {
 
   public void test404ViaHttp() {
     SolrClient client = getSolrClient();
-    QueryRequest request = new QueryRequest(params("file", "does-not-exist-404.txt"));
-    request.setPath("/admin/file");
+    var request =
```
gerlowskija: Ditto to Christine's comment above about the duplication in these GSR instantiations.

IMO this matters less in a test code, where a reader only look at a single test and want things to be more explicit.  But worth considering a refactor...

gerlowskija (commented):
Looks great overall - I'll be really excited to see the 'qt' stuff disappear.

In a few places, you're using GSR where we actually have v2 request objects in SolrJ.  If we're open to using those, I think that'd be a better alternative.  Tried to flag some of those "in line".

dsmiley:
I would very much welcome your improvement proposals to this PR, Jason :-).  Thanks for the review.  I'm a bit over-extended.

solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
gerlowskija: Tried switching to FetchIndexVersion and the change looks great, but ultimately decided against it until we agree on a path forward for [SOLR-17549](https://issues.apache.org/jira/browse/SOLR-17549).

solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
gerlowskija: ditto, re: held off for now until we get a fix for [SOLR-17549](https://issues.apache.org/jira/browse/SOLR-17549)

solr/core/src/java/org/apache/solr/handler/IndexFetcher.java
gerlowskija: Refactored for DRY

solr/core/src/test/org/apache/solr/handler/TestReplicationHandler.java
gerlowskija: [0] I *did* switch to FetchIndexVersion here, since there's code explicitly checking the status code at each call site.

solr/core/src/test/org/apache/solr/handler/TestReplicationHandler.java
gerlowskija: Tried switching to CoresApi.ReloadCore, but it looks like there's a bug in that implementation.  I've filed a ticket to follow up outside of this PR: https://issues.apache.org/jira/browse/SOLR-17907

solr/core/src/test/org/apache/solr/handler/TestReplicationHandler.java
```diff
@@ -217,19 +219,12 @@ private NamedList<Object> getDetails(SolrClient s) throws Exception {
...
-    params.set("_trace", "getIndexVersion");
-    var req =
-        new GenericSolrRequest(
-                SolrRequest.METHOD.GET, ReplicationHandler.PATH, SolrRequestType.ADMIN, params)
-            .setRequiresCollection(true);
-    NamedList<Object> res = s.request(req);
-    assertReplicationResponseSucceeded(res);
-
-    return res;
+    final var req = new ReplicationApi.FetchIndexVersion(coreName);
```
dsmiley, line 224: nice

PR #3519, opened 2025-08-27
SOLR-17806: Migrate thread pool metrics to OTEL

jkmuriithi:

# Description

With Dropwizard, `ExecutorService`'s were instrumented via calls to `MetricUtils.instrumentedExecutorService` (source code [here](https://github.com/dropwizard/metrics/blob/560ac4b38f5dd6172bfe1f68640951906dd5e0e7/metrics-core/src/main/java/com/codahale/metrics/InstrumentedExecutorService.java)), which in turn created an instance of `com.codahale.metrics.InstrumentedExecutorService`. This gave us task/thread counts and task duration information for various Solr thread pools.

Screenshot example of the existing metrics:
Screenshot 2025-08-27 at 3 45 55 PM

# Solution

Create `OtelInstrumentedExecutorService` to act as a drop-in replacement for `InstrumentedExecutorService`. Replace the `MetricUtils` method calls.

Screenshot example of the new metrics:
Screenshot 2025-08-27 at 3 55 03 PM
# Tests

The previous `MetricUtils` static method didn't seem to have any tests. Since this PR introduces a new class to replace the functionality of that method, I think it makes sense to add some unit tests. I'll come back to this PR to add those.

# Checklist


solr/core/src/java/org/apache/solr/util/stats/OtelInstrumentedExecutorService.java
```diff
@@ -0,0 +1,276 @@
...
+import org.apache.solr.metrics.otel.instruments.AttributedLongTimer;
+import org.apache.solr.metrics.otel.instruments.AttributedLongTimer.MetricTimer;
+import org.apache.solr.metrics.otel.instruments.AttributedLongUpDownCounter;
+
+/**
+ * OTEL instrumentation wrapper around {@link ExecutorService}. Based on {@link
+ * com.codahale.metrics.InstrumentedExecutorService}.
+ */
+public class OtelInstrumentedExecutorService implements ExecutorService {
+  public static final AttributeKey<String> NAME_ATTR = AttributeKey.stringKey("executor_name");
```
mlbiscoc: Maybe call it `EXECUTOR_NAME` since just `NAME` is too generic, I can imagine an actual `NAME` attribute existing one day.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -853,14 +853,11 @@ private void loadInternal() {
 
     // NOCOMMIT: Do we need this for OTEL or is this specific for reporters?
```
mlbiscoc: We can start removing these NOCOMMIT comments

solr/core/src/java/org/apache/solr/handler/component/HttpShardHandlerFactory.java
```diff
@@ -445,10 +445,7 @@ public void initializeMetrics(
     // TODO SOLR-17458: Add Otel
```
mlbiscoc: Clean up this TODO

solr/core/src/java/org/apache/solr/util/stats/OtelInstrumentedExecutorService.java
```diff
@@ -0,0 +1,283 @@
...
+
+    // Each metric type needs a separate name to avoid obscuring other types
+    this.submitted =
+        new AttributedLongCounter(
+            ctx.longCounter("solr_executor_tasks", "Number of ExecutorService tasks"),
+            attrs.toBuilder().put(TYPE_ATTR, "submitted").build());
+    this.completed =
+        new AttributedLongCounter(
+            ctx.longCounter("solr_executor_tasks", "Number of ExecutorService tasks"),
+            attrs.toBuilder().put(TYPE_ATTR, "completed").build());
```
mlbiscoc: Initialize the `ctx.longCounter` once then just pass it in

solr/core/src/java/org/apache/solr/util/stats/OtelInstrumentedExecutorService.java
```diff
@@ -0,0 +1,283 @@
...
+    this.running =
+        new AttributedLongUpDownCounter(
+            ctx.longUpDownCounter(
+                "solr_executor_tasks_active", "Number of running ExecutorService tasks"),
+            attrs.toBuilder().put(TYPE_ATTR, "running").build());
```
mlbiscoc: I'm not sure if this running should be an up/down counter or a gauge but I think its ok. I only every used them for cumulative sums which can also be negative but I could be wrong here.

Regardless just put running in the name. No point in having the type attribute if its the only metric with the `type=running as there is nothing to aggregate on it for this metric.

solr/core/src/java/org/apache/solr/util/stats/OtelInstrumentedExecutorService.java
```diff
@@ -0,0 +1,283 @@
...
+                "Timing of ExecutorService tasks",
+                OtelUnit.MILLISECONDS),
+            attrs.toBuilder().put(TYPE_ATTR, "idle").build());
+    this.duration =
+        new AttributedLongTimer(
+            ctx.longHistogram(
+                "solr_executor_task_times",
+                "Timing of ExecutorService tasks",
+                OtelUnit.MILLISECONDS),
+            attrs.toBuilder().put(TYPE_ATTR, "duration").build());
```
mlbiscoc: Same as the comment on initializing once.

solr/core/src/java/org/apache/solr/util/stats/OtelInstrumentedExecutorService.java
```diff
@@ -0,0 +1,283 @@
...
+          "solr_executor_fork_join_pool_threads",
+          "Fork join pool thread counts",
+          measurement -> {
+            measurement.record(
+                forkJoinPool.getActiveThreadCount(),
+                attrs.toBuilder().put(TYPE_ATTR, "active").build());
+            measurement.record(
+                forkJoinPool.getRunningThreadCount(),
+                attrs.toBuilder().put(TYPE_ATTR, "running").build());
+          });
```
mlbiscoc: I just pushed a [commit](https://github.com/apache/solr/commit/5cc10d5b7241b1c1195a3088a8709c556e25f0b8) to start closing these observables. We should close these on `shutdown()`

mlbiscoc (commented):
Cool, thanks @jkmuriithi!

solr/core/src/java/org/apache/solr/util/stats/OtelInstrumentedExecutorService.java
dsmiley: I'm confused by this feedback: shouldn't each AttributedLongCounter have its own longCounter since they are counting different things?

solr/core/src/java/org/apache/solr/util/stats/OtelInstrumentedExecutorService.java
```diff
@@ -0,0 +1,283 @@
...
+    }
+  }
+
+  /** {@inheritDoc} */
```
dsmiley: This is always pointless.  `@inheritDoc` is useful if you want to _add_ information beforehand as a precursor to the inherited/boilerplate information.

solr/core/src/java/org/apache/solr/core/CoreContainer.java
```diff
@@ -1132,14 +1129,13 @@ private void loadInternal() {
 
     // setup executor to load cores in parallel
     ExecutorService coreLoadExecutor =
-        MetricUtils.instrumentedExecutorService(
+        new OtelInstrumentedExecutorService(
```
dsmiley: I think we shouldn't be wrapping the ExecutorService if metrics are disabled.  For example they should be disabled by default in tests.  Having a factory method allows the conditional disablement to exist there.

solr/core/src/java/org/apache/solr/util/stats/OtelInstrumentedExecutorService.java
mlbiscoc: The `submitted` and `completed` metric are sharing a single metric name `solr_executor_tasks` and description that is differentiated by attributes. Some have their own longCounter but some share names just with different attributes. The way it is done here technically works. But the reason I am saying initialize it before is so there is one instance of its creation and lets say we want to change the description of the metric, we only have to change the one instance.

```
var tasksCounter =
    ctx.longCounter("solr_executor_tasks", "Number of ExecutorService tasks");
    this.submitted =
        new AttributedLongCounter(
            tasksCounter,
            attrs.toBuilder().put(TYPE_ATTR, "submitted").build());
    this.completed =
        new AttributedLongCounter(
           tasksCounter,
            attrs.toBuilder().put(TYPE_ATTR, "completed").build());
```

dsmiley: I understand that the name & description is common for multiple fully attributed / distinguished metrics.  But doesn't the "LongCounter" store the underlying value, and thus wouldn't sharing it with two different AttributedLongCounters thus share the same LongCounter instance (the backing value for the current _value_), and thus be totally invalid?

mlbiscoc: Yes BUT only if the metric name has no differing attributes on the metric so then it would just be a single time series. Counter instrument doesn’t hold a single backing value otherwise. In this case, there are 2 distinct metrics with 2 unique attributes (`type`) differentiated from each under under same metric name. The AttributedLongCounter is initialized with 2 different sets of attributes thus creating 2 separate time series on the same metric name.

Thats why in tests we get in a psuedo structure:
```
MetricSnapshot {
   Snapshot {
      Metric name: solr_executor_tasks <- Shared metric name
      Datapoints: { <- Unique time series
         solr_executor_tasks{type=submitted} 0
         solr_executor_tasks{type=completed} 0
      }
   }
}
```

[See here](https://github.com/apache/solr/blob/7ffda2ce9418671eea008d5423b5d16c479c0c20/solr/core/src/java/org/apache/solr/metrics/otel/instruments/AttributedLongCounter.java#L37) the counter adds not to just the metric itself but the unique sets of attributes as well that was binded to the AttributedLongCounter.

dsmiley: I admit I don't get it despite your attempts to explain it.  But no matter; let's proceed with your recommendation anyway.  Once the change you suggest is made, presumably a test here will show that the values between two such AttributedLongCounters are different despite sharing the same LongCounter (let's ensure this is so if not).  I'd love to debug this test and dive in and see for myself.

mlbiscoc: No worries, my explanation there might have had too much fluff there but let me try one last time I swear :). The `LongCounter`  came from an `SDKMeterProvider` under the hood and
my guess of the implementation of that `LongCounter` from the `SDKMeterProvider` in pseudocode might look something like this:
```
ConcurrentMap<Attributes, LongAdder> store = new ConcurrentHashMap<>();

void add(long n, Attributes attrs) {
  store.computeIfAbsent(attrs, a -> new LongAdder()).add(n);
}
```

I also did it for the DirectUpdateHandler2 already [here](https://github.com/apache/solr/blob/b79d939983740020af496e622cec721e6674c89c/solr/core/src/java/org/apache/solr/update/DirectUpdateHandler2.java#L247) with a bunch of tests making sure they [have their own values](https://github.com/apache/solr/blob/b79d939983740020af496e622cec721e6674c89c/solr/core/src/test/org/apache/solr/update/DirectUpdateHandlerTest.java#L228). But still, I agree we should have a test the different values for this PR.

dsmiley: Ah I see.  It's not an obvious implementation.  The _appearance_ of LongUpDownCounter or whatever suggests to me it actually has a value.  But kinda sorta... the wrapped Attributed thing _over_ it provides the attributes to the recording method on LongUpDownCounter (e.g. inc by 1) call and then internally it does what you say.  It's not obvious at all.  Did you see javadocs/sample code that suggest to you that the LongUpDownCounter instance (or similar) could be reused?

mlbiscoc: So here were [some snippets](https://opentelemetry.io/docs/languages/java/api/#counter) from the OTEL java documentation. The point of attributed wrappers is so we are not populating the classes with shared instrument members and attribute initializations and figuring out how to link them together at recording. The attributed classes wrap them together for us and make it look very similar to Dropwizard. But I did some digging in the OTEL spec and found this:

> If the attribute names and types are provided during the [counter creation](https://opentelemetry.io/docs/specs/otel/metrics/api/#counter-creation), the [OpenTelemetry API](https://opentelemetry.io/docs/specs/otel/overview/#api) authors MAY allow attribute values to be passed in using a more efficient way (e.g. strong typed struct allocated on the callstack, tuple). The API MUST allow callers to provide flexible attributes at invocation time rather than having to register all the possible attribute names during the instrument creation.

Attributes MAY be passed but the API MUST have attributes passed at invocation time rather than insturment creation. Digging even further I found they actually did allow for whats called `Bound` insturments to link attributes at creation which looks close to what my attributed wrappers did until they removed it in 1.10. See this [PR](https://github.com/open-telemetry/opentelemetry-java/pull/3928). So attributes at invocation time is the only way you can do it right now.

dsmiley:
As I browse Solr 10 metrics endpoint for the first time, I'm seeing these metrics standout as seeming misplaced to me, from a naming standpoint.  For metrics on a Solr core, I expect the metrics name to start with `solr_core_`.  If not at a core level, I think `solr_node_ would be the right prefix.  We do similarly already for caches.  What follows the `solr_` prefix should be a very restricted list matching the enum `org.apache.solr.core.SolrInfoBean.Group`

PR #3521, opened 2025-08-28
SOLR-17847: Introduce Navigation to Configsets

iamsanjay:
https://issues.apache.org/jira/browse/SOLR-17847

### DIFF
--- a/solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsNavBar.kt
+++ b/solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsNavBar.kt
@@ -32,6 +32,7 @@ import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
 import org.apache.solr.ui.generated.resources.Res
 import org.apache.solr.ui.generated.resources.configsets_index_query
+import org.apache.solr.ui.generated.resources.configsets_request_handlers
 import org.apache.solr.ui.generated.resources.configsets_search_components
 import org.apache.solr.ui.generated.resources.configsets_update_configuration
 import org.apache.solr.ui.generated.resources.files
@@ -75,6 +76,6 @@ private fun tabLabelRes(item: ConfigsetsTab) = when (item) {
     ConfigsetsTab.Schema -> Res.string.schema
     ConfigsetsTab.UpdateConfig -> Res.string.configsets_update_configuration
     ConfigsetsTab.IndexQuery -> Res.string.configsets_index_query
-    ConfigsetsTab.Handlers -> Res.string.configsets_index_query
+    ConfigsetsTab.Handlers -> Res.string.configsets_request_handlers
     ConfigsetsTab.SearchComponents -> Res.string.configsets_search_components
 }


solr/ui/src/commonMain/composeResources/values/strings.xml
```diff
@@ -50,6 +50,16 @@
...
 
+  <!-- Configsets (nav) -->
+  <string name="configsets_overview">Overview</string>
+  <string name="configsets_libraries">Libraries</string>
+  <string name="configsets_files">Files</string>
+  <string name="configsets_schema">Schema</string>
+  <string name="configsets_update_configuration">Update Configuration</string>
+  <string name="configsets_index_query">Index / Query</string>
+  <string name="configsets_request_handlers">Request Handlers / Dispatchers</string>
+  <string name="configsets_search_components">Search Components</string>
```
malliaridis: I believe that we could use general terms here, without prefixing it with `configsets`, because we will likely have similar words (like "overview", "files", "schema") in other places too, so reusing them would be beneficial (smaller files, less duplication).

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/configsets/ConfigsetsComponent.kt
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.solr.ui.components.configsets.overview.OverviewComponent
+import org.apache.solr.ui.views.navigation.configsets.ConfigsetsTab
+
+/**
+ * Contract for the Configsets feature.
+ *
+ * Responsibilities:
+ * - Owns tab/navigation state within the Configsets screen.
+ * - Loads and exposes the list of available configsets.
+ * - Persists the currently selected tab and configset.
```
malliaridis: Point 1 and 3 is the same I believe? Point 2 is responsibility eventually out of scope (not necessarily though).

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/configsets/ConfigsetsComponent.kt
```diff
@@ -0,0 +1,87 @@
...
+ * - [onSelectTab]: update the active tab.
+ * - [onSelectConfigset]: change the active configset.
+ *
+ * Notes:
+ * - Scope: feature-local only; app-level navigation is owned by [MainComponent].
+ * - State restoration is implementation-defined (e.g., MVIKotlin Store / Decompose StateKeeper).
+ *
+ * See also:
+ * - Decompose (component lifecycle & composition)
+ * - MVIKotlin (unidirectional data flow)
```
malliaridis: No need to repeat yourself here, because readers can look into the interface. When updating model or actions, you would also have to update documentation here (can be avoided). Nor do you have to provide information that is "common" knowledge or out of scope of this configset ("see also" points). You can reduce it to only what this component is all about, "managing the navigation and content / layout of the configset screen".

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/configsets/overview/OverviewComponent.kt
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.solr.ui.components.configsets.overview
+
+interface OverviewComponent
```
malliaridis: I would go with `ConfigsetOverviewComponent` to avoid naming conflicts later on when other sub-sections have an "overview" component.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/configsets/ConfigsetsComponent.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Child stack that holds the navigation state.
+     */
+    val childStack: Value<ChildStack<*, Child>>
+
+    /**
+     * All possible navigation targets (children) within the Configsets feature.
+     *
+     * Each child wraps its own component, which holds the state and logic for that screen.
```
malliaridis: That is correct, but we don't have to mention that here (part of common knowledge).

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/configsets/ConfigsetsComponent.kt
```diff
@@ -0,0 +1,87 @@
...
+    sealed interface Child {
+        data class Overview(val component: OverviewComponent) : Child
+    }
+
+    /**
+     * UI model for the Configsets screen.
+     *
+     * @property selectedTab Active tab.
+     * @property configSets List of available configsets (domain model).
+     * @property selectedConfigset Name of the active configset (non-null when available).
```
malliaridis: Is it perhaps better to allow null values, so that
- `null` means no selection (no reference, no configset)
- `string` means selection (actual configset exists)

Of course an empty string is also not a valid configset, but from reading the field I believe `selectedConfigset: String? = null` would be easier to handle.

Since in Kotlin you have explicit nullability, you can handle it differently like with `value?.let { ... }` (a Kotlin feature useful for executing code in a scoped block only if a value is not null).

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/configsets/data/ListConfigsets.kt
```diff
@@ -0,0 +1,29 @@
...
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/**
+ * List of config sets.
+ */
+@Serializable
+data class ListConfigsets(
+    @SerialName("configSets")
+    val names: List<String> = emptyList(),
```
malliaridis: Is there a special reason you chose to explicitly define a serial name, rather than using `val configSets`?

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/configsets/integration/DefaultConfigsetsConponent.kt
```diff
@@ -0,0 +1,127 @@
...
+import org.apache.solr.ui.views.navigation.configsets.ConfigsetsTab
+
+/**
+ * Default implementation of the [ConfigsetsComponent].
+ */
+class DefaultConfigsetsConponent internal constructor(
+    componentContext: AppComponentContext,
+    storeFactory: StoreFactory,
+    httpClient: HttpClient,
+    destination: String? = null,
```
malliaridis: I would go with an enum here, so that we limit the possible destinations to only known values. A string can be anything, but an enum cannot.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/configsets/integration/DefaultConfigsetsConponent.kt
```diff
@@ -0,0 +1,127 @@
...
+    ): Child = when (configuration) {
+        Configuration.Overview -> Child.Overview(overviewComponent(componentContext))
+    }
+
+    /**
+     * Calculates the initial stack based on the destination provided.
+     */
+    private fun calculateInitialStack(destination: String?): List<Configuration> = listOf(
+        when (destination) {
+            "Overview" -> Configuration.Overview
```
malliaridis: Consider enums (limited possibilities) instead of strings, this way, no else case is needed (thanks to type-safety).

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/configsets/integration/DefaultConfigsetsConponent.kt
```diff
@@ -0,0 +1,127 @@
...
+            mainContext = mainScope.coroutineContext,
+            ioContext = ioScope.coroutineContext,
+        ).provide()
+    }
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    override val model = store.stateFlow.map(mainScope, configsetsStateToModel)
+
+    override fun onSelectTab(tab: ConfigsetsTab) {
+        store.accept(ConfigsetsStore.Intent.SelectTab(tab))
```
malliaridis: Feel free to add `import 
import org.apache.solr.ui.components.configsets.store.ConfigsetsStore.Intent` to reduce the chained words referencing or using Intents of the component's store.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/configsets/integration/HttpEnvironmentStoreClient.kt
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Client implementation of the [ConfigsetsStoreProvider.Client] that makes use
+ * of a preconfigured HTTP client for accessing the Solr API.
+ *
+ * @property httpClient HTTP client to use for accessing the API. The client has to be
+ * configured with a default request that includes the host, port and schema. The client
+ * should also include the necessary authentication data if authentication / authorization
+ * is enabled.
+ */
+class HttpEnvironmentStoreClient(
```
malliaridis: I believe you meant here `HttpConfigStoreClient`?

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/configsets/store/ConfigsetsStore.kt
```diff
@@ -0,0 +1,44 @@
...
+    sealed interface Intent {
+        /**
+         * Intent for selecting tab.
+         */
+        data class SelectTab(val tab: ConfigsetsTab) : Intent
+
+        /**
+         * Intent for fetching configsets from the solr.
+         */
+        data object FetchConfigSets : Intent
```
malliaridis: Can the user manually fetch the configsets? Because intents describe a user's intent / wish to do something (on request). If not, consider using only `Action. FetchInitialConfigsets` in `ConfigsetsStoreProvider` (for initial fetch) and removing this.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/configsets/store/ConfigsetsStore.kt
```diff
@@ -0,0 +1,44 @@
...
+
+        data class SelectConfigSet(val configSetName: String) : Intent
+    }
+    data class State(
+        val selectedTab: ConfigsetsTab = ConfigsetsTab.Overview,
+        val selectedConfigset: String = "",
+        val configSets: ListConfigsets = ListConfigsets(),
```
malliaridis, line 45: You are exposing here an API model too far into the app. It would be better to use an internal domain model instead, like `List<ConfigSet>` where Configset may be a data model with just a `name` property for now (later we may consider using IDs, display names or just names, and other values).

Because we are free to use whatever models we want in the frontend, we don't have to keep the API structure across the app, and it will also be easier to migrate later to v2 endpoints (in case v1 is used) if we just map it as soon as we retrieve the data (so at the store client implementation).

Also, using a List directly will allow us later to directly use the field for iterating over it (`configSets.forEach { configset -> ... }`)

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/components/configsets/store/ConfigsetsStoreProvider.kt
```diff
@@ -0,0 +1,110 @@
...
+            is Message.SelectedConfigSetChanged -> copy(selectedConfigset = msg.configsetName)
+        }
+    }
+
+    /**
+     * Client interface for fetching configsets information.
+     */
+    interface Client {
+        /** To fetch a list of configsets. */
+        suspend fun fetchConfigSets(): Result<ListConfigsets>
```
malliaridis, line 110: So I would recommend not to return a `ListConfigsets` (API model), but rather a `List<Configset>` (domain model) here.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsContent.kt
```diff
@@ -0,0 +1,80 @@
...
+        selectConfigset = { s: String -> component.onSelectConfigset(s) },
+        availableConfigsets = model.configSets,
+        content = { tab, _ ->
+            when (tab) {
+                ConfigsetsTab.Overview -> {
+                    when (val child = currentChild) {
+                        is Child.Overview ->
+                            OverviewContent(component = child.component)
+                        else ->
+                            Box(Modifier.fillMaxSize(), contentAlignment = Alignment.Center) {
```
malliaridis: Always go with or without names for the parameters, avoid mixing it. Consider also writing params in new lines if more than 2.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsNavBarComponent.kt
```diff
@@ -0,0 +1,146 @@
...
+import org.apache.solr.ui.generated.resources.configsets_overview
+import org.apache.solr.ui.generated.resources.configsets_request_handlers
+import org.apache.solr.ui.generated.resources.configsets_schema
+import org.apache.solr.ui.generated.resources.configsets_search_components
+import org.apache.solr.ui.generated.resources.configsets_update_configuration
+import org.apache.solr.ui.views.navigation.configsets.ConfigsetsTab
+import org.jetbrains.compose.resources.stringResource
+
+@Composable
+fun ConfigsetsNavBarComponent(
```
malliaridis: Since it is a plain composable function, avoid using the kinda "reserved" suffix "Component". This avoids confusion with the component interfaces and implementations. Just name it `ConfigsetsNavBar` for example.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsNavBarComponent.kt
```diff
@@ -0,0 +1,146 @@
...
+    selectedTab: ConfigsetsTab,
+    selectTab: (ConfigsetsTab) -> Unit,
+    selectedConfigSet: String,
+    selectConfigset: (String) -> Unit,
+    availableConfigsets: ListConfigsets,
+    modifier: Modifier = Modifier,
+    content: @Composable (tab: ConfigsetsTab, configset: String) -> Unit = { tab, _ ->
+        Box(Modifier.fillMaxSize(), contentAlignment = Alignment.Center) { Text(stringResource(tabLabelRes(tab))) }
+    },
```
malliaridis: This is now the case where I would ask you "why does a navigation bar that is used for navigation need to know configsets? And why does it have a content? What can I put into it, and where is it added?"

By definition, I would expect the navigation bar to be only the elements on the top that I can choose from, so the navigation items. You are now providing the content as well, which kinda confuses the user of this composable.

At the same time, I am wondering, does the navigation bar need to know about the configsets? I am just navigating through tabs, am I not?

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsNavBarComponent.kt
```diff
@@ -0,0 +1,146 @@
...
+) {
+    if (availableConfigsets.names.isEmpty()) {
+        Box(Modifier.fillMaxSize(), contentAlignment = Alignment.Center) {
+            Text("No configsets available")
+        }
+        return
+    }
+
+    var expanded by remember { mutableStateOf(false) }
```
malliaridis: Another UI state that should be added to a component's state. If you do not add it, you won't be able later to load additional information on demand once the user expands it. 🤔

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsNavBarComponent.kt
```diff
@@ -0,0 +1,146 @@
...
+                            text = { Text(name) },
+                            onClick = {
+                                selectConfigset(name)
+                                expanded = false
+                            },
+                        )
+                    }
+                }
+            }
+        }
```
malliaridis: This block seems like it does not belong here 🤔

malliaridis (changes requested):
So now I am going into the fine details, because you already adapted great to my previous requests. 😄 

I have added a few thoughts on what could be added and what should eventually be removed / replaced (mainly for maintainability). Most of the comments are just small improvements, others are just to point out some potential redundancy.

Overall though the changes are almost ready to be merged. Do you plan to add some tests too? You can likely use the acceptance criteria as test descriptions, I would just rephrase them to the typical "given/when/then" phrases. I can also assist you with that if you want.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsNavBarComponent.kt
iamsanjay: Lets move the configset dropdown and content outside of this component to ConfigsetsComponent

malliaridis: With "[move out] to ConfigsetsComponent" you mean to `ConfigsetsContent` (composable)? If so, then yes, it makes sense to me. Otherwise you have to elaborate. :s

iamsanjay:
We   have two ways of looking at the config

- the Configsets API (list/upload/delete, names only) and
- the collection-scoped /solr/{collection}/config endpoints (actual config: handlers, params, etc.).

The solrconfig details on the first screen aren’t exposed by the Configsets API; unless we fetch and parse files ourselves, we have to go via a collection. That suggests we’re missing the collection context—in practice, or my personal experience so far, edits happen from the collection side. And the whole thought came to my mind when we discussed Configset object, because configsets api cannot share much info about config unless we go via collection scoped endpoints.

epugh:
In another PR it was noted that we may be missing the appropriate api endpoints for config sets!

iamsanjay:
I believe even ConfigsetsNavBar can be a Solr component(composable) because that is something that we will have for other screens as well.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsNavBar.kt
```diff
@@ -43,38 +43,32 @@ import androidx.compose.ui.Alignment
...
-        Box(Modifier.fillMaxSize(), contentAlignment = Alignment.Center) { Text(stringResource(tabLabelRes(tab))) }
-    },
 ) {
-    if (availableConfigsets.names.isEmpty()) {
+    if (availableConfigsets.isEmpty()) {
         Box(Modifier.fillMaxSize(), contentAlignment = Alignment.Center) {
             Text("No configsets available")
         }
         return
     }
```
malliaridis: I would normally move this logic to the parent (`ConfigsetsContent`), so that the NavBar does not have any reference to information it doesnt use. Render the `ConfigsetsNavBar` only if the availableConfigsets is not empty (in parent), otherwise render this box. And you may still add to the `ConfigsetComponent.Model` a state value like `hasConfigsets: boolean` that you can directly use (and populate during mapping of state to model), so that you don't need to run a function on a collection in the UI (even if it is a cheap operation here).

Most important though is that I would not hide the navigation elements (UI/UX). Navigation elements in general should stay where they are to maintain recognizability (users should see the screen and directly know where they are by the static elements like navigation options). They should also know what they can do here, even for a fresh installation. Most you could do (and I would not even do that) is to disable the navigation options. But instead, I would support in each section the case where no configset is selected yet, so that user can still navigate through the tabs and see what the content is, before actually taking any actions or loading any content.

When no configset is selected, the tab may display the sections with small hints (like "Here you can manage the configsets files. Select a configset to get started." for the Files tab). But that is out of scope of this PR.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsDropdown.kt
```diff
@@ -0,0 +1,64 @@
...
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun ConfigsetsDropdown(
+    selectedConfigSet: String,
+    selectConfigset: (String) -> Unit,
+    availableConfigsets: List<Configset>,
+    modifier: Modifier = Modifier,
+) {
+    var expanded by remember { mutableStateOf(false) }
```
malliaridis: I would likely want to add this to the component state, so that when the user clicks on it, we can perform some tasks later, like "silent" refresh in the background (if that makes sense at some point). So the more is stored in the component, the more we are in control of user interactions.

But we may also leave it as it is for now, and only update it later.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsDropdown.kt
```diff
@@ -0,0 +1,64 @@
...
+        ExposedDropdownMenuBox(
+            expanded = expanded,
+            onExpandedChange = { expanded = !expanded },
+            modifier = Modifier.widthIn(min = 256.dp).weight(1f),
+        ) {
+            OutlinedTextField(
+                value = selectedConfigSet,
+                onValueChange = {},
+                readOnly = true,
+                label = { Text("Configset") },
```
malliaridis: I believe there are still some values that can be localized?

solr/ui/src/commonTest/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsContentTest.kt
```diff
@@ -0,0 +1,128 @@
...
+class ConfigsetsContentTest {
+
+    @Test
+    fun emptyList_showsPlaceholder() = runComposeUiTest {
+        val comp = TestConfigsetsComponent(emptyList())
+
+        setContent { ConfigsetsContent(component = comp) }
+
+        // Placeholder text from the TextField
+        onNodeWithText("No configsets available").assertExists()
```
malliaridis, line 46: When we later introduce localization, these tests may fail becasue the text may be localized during tests and not found via `onNodeWithText`. Therefore, it is recommended to use `Modifier.testTag(...)` in the composables to check if specific composables are present.

solr/ui/src/commonTest/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsContentTest.kt
```diff
@@ -0,0 +1,128 @@
...
+
+        // Field click shouldn’t open the menu (anchor is disabled in our impl)
+        onNodeWithText("No configsets available").performClick()
+        runOnIdle { assertFalse(comp.model.value.expanded) }
```
malliaridis, line 50: To keep it simple with less nesting, prefer `advanceUntilIdle()`

solr/ui/src/commonTest/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsContentTest.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    private val _model = MutableStateFlow(
+        ConfigsetsComponent.Model(
+            selectedTab = ConfigsetsTab.Overview,
+            configsets = names.map(::Configset),
+            selectedConfigset = selected,
+            expanded = expanded,
+        ),
+    )
+    override val model: StateFlow<ConfigsetsComponent.Model> = _model
```
malliaridis, line 104: You may simplify your test component and your access in tests to your model values by not using a mutable state flow, but rather static fields where you can directly access. See for example [TestBasicAuthComponent](https://github.com/apache/solr/blob/main/solr/ui/src/commonTest/kotlin/org/apache/solr/ui/views/auth/TestBasicAuthComponent.kt). Of course the way you did it isn't wrong or bad, it is just less readable when accessing a field with `comp.model.value.selectedTab` instead of just `comp.selectedTab`. And you are close to start writing components that you have to test again due to the logic behind setting a field value and the relationship between fields. Imagine the case where you have `isValidEmail: Boolean` and `email: String` as state in your component. What would be the logic in your test component?

solr/ui/src/commonTest/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsContentTest.kt
```diff
@@ -0,0 +1,128 @@
...
+        // Placeholder text from the TextField
+        onNodeWithText("No configsets available").assertExists()
+
+        // Field click shouldn’t open the menu (anchor is disabled in our impl)
+        onNodeWithText("No configsets available").performClick()
+        runOnIdle { assertFalse(comp.model.value.expanded) }
+    }
+
+    @Test
+    fun selectItem_updatesSelection_andClosesMenu() = runComposeUiTest {
```
malliaridis, line 54: You are mixing here UI test with integration test of your test component (instead of your actual component). It is hard to explain this, but let me try to break it down anyway.

The statement "select item updates selection and closes menu" is a test that tests the behavior of your test component, and not testing the UI itself. Pure UI tests (that use the test component) would be:
- WHEN dropdown clicked THEN setMenuExpanded is called
- WHEN dropdown item clicked THEN onSelectConfigset called

Possible integration tests (tests that use the actual implementation of ConfigsetsComponent) would be:
- GIVEN expanded is true AND some configsets WHEN configset selected THEN dropdown collapses
- GIVEN expanded is true AND some configsets WHEN configset selected THEN selectedConfigset changes
- GIVEN expanded is false AND some configsets WHEN expand menu THEN expanded is true
- GIVEN expanded is false AND no configsets WHEN expand menu THEN expanded is false

Btw. the interface function `setMenuExpanded` could be renamed to `onExpandMenu` for better readability, both in tests and in code.

I believe with these test examples it is a bit clearer where to draw the lines between simple UI tests, that test composable functions with test components, and actual integration tests, that use actual implementations. One of the root causes why it is easy to confuse these is that we are providing both business logic and UI in the same package. If we would use different packages, it would be easier to distinguish them. business logic does not know UI composables, and UI composables do not know implementation details of business logic. We may consider splitting the UI module into two modules in the future (if that makes sense to have two modules for the UI).

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsDropdown.kt
iamsanjay: Yes this one is already fixed.

solr/ui/src/commonMain/kotlin/org/apache/solr/ui/views/configsets/ConfigsetsDropdown.kt
iamsanjay: I believe it is. You may have reviewed the older commit.

PR #3522, opened 2025-08-28
Fix reference in collapse and expand docs

benediktwerner:
The [Collapse and Expand Results documentation](https://solr.apache.org/guide/solr/latest/query-guide/collapse-and-expand-results.html#collapsing-query-parser) references itself below the `collectElevatedDocsWhenCollapsing` parameter which doesn't really make any sense. I adjusted it to reference the [Query Elevation Component](https://solr.apache.org/guide/solr/latest/query-guide/query-elevation-component.html) which I assume was the intention.


epugh:
Thank you for the fix and sorry it took so long!

PR #3524, opened 2025-08-29
SOLR-17877: Introduce cluster property overseerEnabled, replacing dual booleans

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17877

This PR removes the dual booleans.  They weren't advertised before thus no backwards compatibility consideration is needed.  The ability to toggle just the distributed cluster state update part is not needed.  Now there's an "overseerEnabled" cluster property.  There's also an EnvUtil fallback on "solr.cloud.overseer.enabled".

Upgrade guide notes should be added to at least advise someone who divined the secret setting.

A follow-on change will change the default setting to consider the cluster version.

Tests pass (ran locally).

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/Overseer.java
+++ b/solr/core/src/java/org/apache/solr/cloud/Overseer.java
@@ -712,7 +712,7 @@ public class Overseer implements SolrCloseable {
     this.stats = new Stats();
     this.config = config;
     this.distributedClusterStateUpdater =
-        new DistributedClusterStateUpdater(config.getDistributedClusterStateUpdates());
+        new DistributedClusterStateUpdater(reader.getClusterProperty(ZkStateReader.DISTRIBUTED_CLUSTER_STATE_UPDATES, false));
 
     this.solrMetricsContext =
         new SolrMetricsContext(
--- a/solr/core/src/java/org/apache/solr/cloud/Overseer.java
+++ b/solr/core/src/java/org/apache/solr/cloud/Overseer.java
@@ -712,7 +712,7 @@ public class Overseer implements SolrCloseable {
     this.stats = new Stats();
     this.config = config;
     this.distributedClusterStateUpdater =
-        new DistributedClusterStateUpdater(reader.getClusterProperty(ZkStateReader.DISTRIBUTED_CLUSTER_STATE_UPDATES, false));
+        new DistributedClusterStateUpdater(!"true".equals(String.valueOf(reader.getClusterProperty(ZkStateReader.OVERSEER_ENABLED, "true"))));
 
     this.solrMetricsContext =
         new SolrMetricsContext(
--- a/solr/core/src/java/org/apache/solr/cloud/Overseer.java
+++ b/solr/core/src/java/org/apache/solr/cloud/Overseer.java
@@ -711,8 +711,7 @@ public class Overseer implements SolrCloseable {
     this.zkController = zkController;
     this.stats = new Stats();
     this.config = config;
-    this.distributedClusterStateUpdater =
-        new DistributedClusterStateUpdater(!"true".equals(String.valueOf(reader.getClusterProperty(ZkStateReader.OVERSEER_ENABLED, "true"))));
+    this.distributedClusterStateUpdater = zkController.getDistributedClusterStateUpdater();
 
     this.solrMetricsContext =
         new SolrMetricsContext(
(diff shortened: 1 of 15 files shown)


solr/core/src/java/org/apache/solr/cloud/api/collections/DistributedCollectionCommandContext.java
dsmiley, line 1: no reason to create a new DistributedClusterStateUpdater here; we can take the existing one in ZkController

solr/core/src/java/org/apache/solr/cloud/Overseer.java
dsmiley, line 1: again; no reason to create another DistributedClusterStateUpdater here

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
```diff
@@ -1134,39 +1128,18 @@ public Builder useOtherCollectionConfigSetExecution() {
...
-     *
-     * @param distributedCollectionConfigSetApi When {@code true}, Collection and Config Set API
-     *     commands are executed in a distributed way by nodes. When {@code false}, they are
-     *     executed by Overseer.
-     * @param distributedClusterStateUpdates When {@code true}, cluster state updates are handled in
-     *     a distributed way by nodes. When {@code false}, cluster state updates are handled by
-     *     Overseer.
-     *     <p>If {@code distributedCollectionConfigSetApi} is {@code true} then this parameter must
-     *     be {@code true}.
+     * @param overseerEnabled When {@code true}, Collection and Config Set API commands are executed
```
dsmiley: I need to flip that explanation

dsmiley (commented):
CHANGES.txt would be an Improvement:
* Added "overseerEnabled" cluster property (or sys prop "solr.cloud.overseer.enabled") and removed the unadvertised pair of solr.xml "distributed" booleans that formerly configured this.

I intend to defer the upgrade guide notes for the related issue that flips the default.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -390,6 +382,26 @@ public ZkController(
...
+    // Now that zkStateReader is available, read the overseer enabled setting
+    // When overseerEnabled is false, both distributed features should be enabled
+    Object clusterProperty = zkStateReader.getClusterProperty(ZkStateReader.OVERSEER_ENABLED, null);
+    boolean overseerEnabled;
+    if (clusterProperty != null) {
+      overseerEnabled = "true".equals(String.valueOf(clusterProperty));
+    } else {
+      overseerEnabled = EnvUtils.getPropertyAsBool("solr.cloud.overseer.enabled", true);
+    }
+    boolean useDistributedUpdates = !overseerEnabled;
```
psalagnac: We should not have two different flags anymore.
(I would fully remove them and use `!overseerEnabled` instead.)

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -390,6 +382,26 @@ public ZkController(
               if (cc != null) cc.securityNodeChanged();
             });
 
+    // Now that zkStateReader is available, read the overseer enabled setting
+    // When overseerEnabled is false, both distributed features should be enabled
+    Object clusterProperty = zkStateReader.getClusterProperty(ZkStateReader.OVERSEER_ENABLED, null);
```
psalagnac: Can't we store the property directly as a boolean instead of a string?

dsmiley:
Thanks for the review!  I addressed the comments, including deciding to incorporate upgrade notes into this PR instead of deferring to another.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -389,6 +389,12 @@ public ZkController(
     if (overseerEnabled == null) {
       overseerEnabled = EnvUtils.getPropertyAsBool("solr.cloud.overseer.enabled", true);
     }
+    if (overseerEnabled) {
+      log.info("The Overseer is enabled.  It will process all cluster commands & state updates.");
+    } else {
+      log.info(
+          "The Overseer is disabled.  Cluster commands & state updates will happen on any/all nodes.");
+    }
```
dsmiley, line 397: @psalagnac I updated this PR to log in one spot and to not log in the constructors of the distributed components.

dsmiley:
I plan to merge this tonight (9pm EST)

PR #3418, opened 2025-07-04
SOLR-17319 : Combined Query Feature for Multi Query Execution

ercsonusharma:
https://issues.apache.org/jira/browse/SOLR-17319

# Description

This feature aims to execute multiple queries of multiple kinds across multiple shards of a collection and combine their result basis an algorithm (like Reciprocal Rank Fusion). It also help resolve the issues being discussed w.r.t the [ previous PR](https://github.com/apache/solr/pull/2489), mainly around across shard documents merging.  It provides more flexibility in terms of querying extending JSON Query DSL  ultimately enabling Hybrid Search in a pure way solving the shortcomings. 

# Solution

* Extended the QueryComponent to create new CombinedQueryComponent and ResponseBuilder to create new CombinedQueryResponseBuilder supports multiple response builders to hold the state and execute multiple queries.
* In JSON Query DSL, a parameter is added to identity Combined Query request and basis that it invokes the new CombinedQueryComponent
* CombinedQueryComponent have multiple response builders assigned for each query. These queries are first executed at the SolrSearchIndexer level and combined them using RRF for now.
* At Shard level also, the responses for the multiple queries are merged.

# Tests

* Added tests for testing the RRF logic independently.
* Added tests for across search index and distributed requests.
* Added tests to assert existing behaviour of search handler's QueryComponent as well as for the newly added CombinedQueryComponent basis the flag in json query DSL.

# Checklist

Please review the following and check all that apply:

- [x] I have reviewed the guidelines for [How to Contribute](https://github.com/apache/solr/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.
- [ ] I have created a Jira issue and added the issue ID to my pull request title.
- [x] I have given Solr maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended, not available for branches on forks living under an organisation)
- [x] I have developed this patch against the `main` branch.
- [ ] I have run `./gradlew check`.
- [x] I have added tests for my changes.
- [ ] I have added documentation for the [Reference Guide](https://github.com/apache/solr/tree/main/solr/solr-ref-guide)

PR #3565, opened 2025-08-29
[exploratory] abstract QueryComponent.mergeIds' ShardDoc processing

cpoerschke:


solr/core/src/java/org/apache/solr/handler/component/QueryComponent.java
```diff
@@ -905,7 +906,61 @@ protected boolean addFL(StringBuilder fl, String field, boolean additionalAdded)
     return true;
   }
 
+  protected abstract static class ShardDocQueue {
```
dsmiley, line 909: If you do an interface, you could have `ShardFieldSortedHitQueue` implement that interface.  Then, `DefaultShardDocQueue` could be removed.

cpoerschke: I like that idea, thanks for the suggestion.

Though on trying it out, then it didn't quite work out since `ShardFieldSortedHitQueue` would need to implement the `push` as either `add` or `insertWithOverflow` and making that choice doesn't seem particularly intuitive: https://github.com/apache/lucene/blob/releases/lucene/10.2.2/lucene/core/src/java/org/apache/lucene/util/PriorityQueue.java#L150-L184

However, making `DefaultShardDocQueue` into an anonymous class seems a good alternative, both locally for `QueryComponent` and if a deriving class needed to maintain a container alongside of `ShardFieldSortedHitQueue` or instead of it.

dsmiley (approved):

cpoerschke:
merged into https://github.com/apache/solr/pull/3418/commits branch

PR #3567, opened 2025-08-29
Force errorprone to max v2.31.0 on branch_9x

janhoy:
To avoid triggering JDK17 requirement. It will select v2.31 or else fail hard during version resolution.

See #3537 where I have incorporated the same changes in addition to some other tweaks. So likely the tweak will be merged from that PR and not this.

### DIFF
--- a/gradle/validation/error-prone.gradle
+++ b/gradle/validation/error-prone.gradle
@@ -35,6 +35,18 @@ if (skipReason) {
   }
 }
 
+configurations.all {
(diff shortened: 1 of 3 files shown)


gradle/validation/error-prone.gradle
dsmiley, line 1: I think your comment next to the version number in `versions.props` is enough, without needing this.  But it's fine.

dsmiley (approved):
thanks

PR #3585, opened 2025-08-31
Fix failing MiniClusterBenchStateTest

janhoy:
Draft PR that skips the code that makes configset path absolute, which does not fly with security manager.

### DIFF
--- a/solr/core/src/java/org/apache/solr/core/ConfigSetService.java
+++ b/solr/core/src/java/org/apache/solr/core/ConfigSetService.java
@@ -171,10 +171,10 @@ public abstract class ConfigSetService {
     Path path = Path.of(pathStr);
 
     // Convert to absolute path if it's relative
-//    if (!path.isAbsolute()) {
-//      String installDir = EnvUtils.getProperty(SolrDispatchFilter.SOLR_INSTALL_DIR_ATTRIBUTE);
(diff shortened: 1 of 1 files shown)


dsmiley:
Would a root cause fix be a SecurityManager `solr-tests.policy` change to ensure that the install path is readable?

janhoy:
> Would a root cause fix be a SecurityManager `solr-tests.policy` change to ensure that the install path is readable?

Why is this location even configurable, if we always want the same folder within the install path? Is it only for tests? If so, could the test set the sysprop as absolute from the beginning?

@epugh What was the reasoning behind the code change to convert to absolute path? What is the right fix in your eyes?

But if your suggestion @dsmiley is to add a new SM rule to allow all sub paths of `solr.install.dir`, then I suppose that would work. But it would also potentially open up for attackers modifying jar files in places we don't want?

epugh:
I _feel_ like what i saw was that using a RELATIVE path didn't work.  Only a fully qualified path worked when calling from the bin/solr script.   So, to support a RELATIVE path, then I added this.   I will take a fresh look and see if I can't do a smaller change.

janhoy:
Remember than when running tests locally, the security manager is not automatically enabled. It *is* enabled when running `bin/solr` from a shell though.

PR #3602, opened 2025-09-02
SOLR-17884: SolrJ, minimize org.apache.http usage

dsmiley:
https://issues.apache.org/jira/browse/SOLR-17884

Primarily 2 changes as seen in both commits.

This will go to branch_9x.


solr/solrj/src/java/org/apache/solr/client/solrj/impl/HttpClientUtil.java
```diff
@@ -73,42 +78,19 @@
  * @deprecated Used to configure the Apache HTTP client. Please use the Http2 client
  */
 @Deprecated(since = "9.0")
-public class HttpClientUtil {
+public class HttpClientUtil implements SolrHttpConstants {
```
dsmiley, line 81: one of the primary changes, showing we inherit constants from the extracted interface.

solr/solrj/src/java/org/apache/solr/common/util/Utils.java
dsmiley, line 1: don't want org.apache.http references here.  Moved to HttpClientUtil, which is all about that package.

dsmiley:
Planning to merge this one this weekend.  It's been up for review for awhile.

dsmiley:
Once I bring this to 9.x, I'd be tempted to create a hello-world gradle project with SolrJ, excluding Apache HttpClient, and then see if the JDK & Jetty based clients can talk to Solr without issue.

janhoy:
> Once I bring this to 9.x, I'd be tempted to create a hello-world gradle project with SolrJ, excluding Apache HttpClient, and then see if the JDK & Jetty based clients can talk to Solr without issue.

Yea, I'm particularly interested in showing how the JDK client can be used with neither apache or jetty deps on client side.

dsmiley:
For experimentation, I played with the Solr MCP server: https://github.com/adityamparikh/solr-mcp-server/pull/10
And discovered what worked and doesn't with respect to dependency exclusion.  I didn't actually change to use some local snapshot build, thus validating this change better.  I forgot but I have done a variety of find-usages one and two levels deep here for both Apache & Jetty dependencies.

I'll merge this Monday afternoon, subject to any further feedback.

PR #3607, opened 2025-09-02
SOLR-17806: Migrate Overseer and Blockcache metrics to OTEL

mlbiscoc:

Migrate Overseer and block cache metrics. Block cache doesn't seem to be used at the moment since the removal of HDFS module but still migrated it in case this ever does get used.

Overseer metrics
```
# HELP solr_overseer_collection_work_queue_size Size of overseer's collection work queue
# TYPE solr_overseer_collection_work_queue_size gauge
solr_overseer_collection_work_queue_size{category="OVERSEER",otel_scope_name="org.apache.solr"} 0.0
# HELP solr_overseer_state_update_queue_size Size of overseer's update queue
# TYPE solr_overseer_state_update_queue_size gauge
solr_overseer_state_update_queue_size{category="OVERSEER",otel_scope_name="org.apache.solr"} 0.0
```


dsmiley (approved):
+1 Just kind of a cursory review honestly.  In tests I trust.

When I see SolrInfoBean, which you newly implemented, it looks like old stuff and I question if we need it.  What is gained by doing so vs SolrMetricProducer?

mlbiscoc:
> When I see SolrInfoBean, which you newly implemented, it looks like old stuff and I question if we need it. What is gained by doing so vs SolrMetricProducer?

I had the same thought recently. I don't think we need it anymore honestly but we're a bit deep to just spend the time migrating off of it with more pressing changes for OTEL. I think we can just do clean up in a 10.x release as I don't think there is any backwards incompatibility after removal.

PR #3608, opened 2025-09-02
SOLR-17891: Reorganize the DeprecatedSystemPropertyMappings file

epugh:
https://issues.apache.org/jira/browse/SOLR-17891

Reorganize the mapping file to put the modern property on the left and the deprecated one on the left.

Sort the file by the highlevel categories to make working with the file easier.

Lastly, introduce a new=!old syntax with the `!` to indicate the values should be inverted when reading in legacy properties.


epugh:
This is ready.  I think the mapping file is easier to read.

dsmiley (commented):
Is there a bats (or other) test for a legacy property?

epugh:
> Is there a bats (or other) test for a legacy property?

There is; in the `test_start_solr.bats` file that verifies the warning shows up in the logs.

dsmiley (approved):

PR #3610, opened 2025-09-03
Remove some dead code from distributed updater

psalagnac:
Simple code cleanup. No functional change

@murblanc is the initial author of most the this change (I'm not adding an entry in change log, so just mentioning this here).


cpoerschke (approved):
Maybe also mention in commit message/title that some logging-related warnings in the class are addressed too.

solr/core/src/java/org/apache/solr/cloud/DistributedClusterStateUpdater.java
```diff
@@ -94,11 +83,9 @@ public DistributedClusterStateUpdater(boolean useDistributedStateUpdate) {
     this.useDistributedStateUpdate = useDistributedStateUpdate;
     if (log.isInfoEnabled()) {
       log.info(
-          "Creating DistributedClusterStateUpdater with useDistributedStateUpdate="
-              + useDistributedStateUpdate
-              + ". Solr will be using "
-              + (useDistributedStateUpdate ? "distributed" : "Overseer based")
-              + " cluster state updates."); // nowarn
+          "Creating DistributedClusterStateUpdater with useDistributedStateUpdate={}. Solr will be using {} cluster state updates.",
```
dsmiley: I suggest we log nothing _here_.  Instead log in ZkController initialization to say whether the Overseer is enabled or disabled.  Don't bother doing so in this PR; it better belongs in #3524 in ZkController when `overseerEnabled` is determined.  I'll do that.

dsmiley:
> murblanc is the initial author of most the this change (I'm not adding an entry in change log, so just mentioning this here).

In the commit message at the end, after a blank line separation, add this line:

    Co-authored-by: Ilan Ginzburg <iginzburg@salesforce.com>

murblanc (approved):

murblanc:
> > murblanc is the initial author of most the this change (I'm not adding an entry in change log, so just mentioning this here).
> 
> In the commit message at the end, after a blank line separation, add this line:
> 
> ```
> Co-authored-by: Ilan Ginzburg <iginzburg@salesforce.com>
> ```

[Edit: Pierre points to me that I wrote this dead code removal. I totally forgot about it, so please ignore the comment below]

The cleanup is Pierre's own work, so no need to update the commit message.
He meant I wrote most of the original code (and left over dead code).

psalagnac:
Thanks for the review.
I reverted the logging changes since they don't really belong to this PR. I may open another one that fixes the logger calls with wrong pattern.

PR #3614, opened 2025-09-04
SOLR-17893: Use caching to remove Remote Proxy bottleneck

HoustonPutman:
https://issues.apache.org/jira/browse/SOLR-17893

- Both `clusterState.getCollectionOrNull()` calls in HttpSolrCall should allow for cached values. There's no reason to fetch live versions every time here.
- The `lastUpdateTime` needs to be updated after getting the stats. If there is no change, that is an "update". We don't want to check every call after 2 seconds, if the ClusterState hasn't changed. We want to check once every 2 seconds.

### DIFF
--- a/solr/CHANGES.txt
+++ b/solr/CHANGES.txt
@@ -228,6 +228,8 @@ Improvements
 ---------------------
 * SOLR-17860: DocBasedVersionConstraintsProcessorFactory now supports PULL replicas. (Houston Putman)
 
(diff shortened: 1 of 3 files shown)


dsmiley (approved):
LGTM!
I remember looking at this many months ago, thinking along the lines of what you did but took no action.  I wasn't looking at it in the context of a perf problem to solve, so I let it be.

PR #3617, opened 2025-09-04
SOLR-17864: Migrate more system properties

epugh:

solr.dns.prevent.reverse.lookup
solr.auth.jwt.allow.outbound.http
solr.authorization.superuser
solr.disableFingerprint

### DIFF
--- a/solr/core/src/java/org/apache/solr/handler/admin/SystemInfoHandler.java
+++ b/solr/core/src/java/org/apache/solr/handler/admin/SystemInfoHandler.java
@@ -77,8 +77,8 @@ public class SystemInfoHandler extends RequestHandlerBase {
    *
    * @see #initHostname
    */
-  private static final String PREVENT_REVERSE_DNS_OF_LOCALHOST_SYSPROP =
-      "solr.admin.handler.systeminfo.dns.prevent.reverse.lookup.enabled";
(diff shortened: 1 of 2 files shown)


epugh:
I think this batch is ready for review!

solr/core/src/java/org/apache/solr/handler/admin/SystemInfoHandler.java
```diff
@@ -77,7 +78,7 @@ public class SystemInfoHandler extends RequestHandlerBase {
    * @see #initHostname
    */
   private static final String PREVENT_REVERSE_DNS_OF_LOCALHOST_SYSPROP =
-      "solr.dns.prevent.reverse.lookup";
+      "solr.admin.handler.systeminfo.dns.prevent.reverse.lookup.enabled";
```
dsmiley: The "enabled" suffix exposes an inversion of expressing what we're going to do.  Arguably boolean settings should express what we *will* do, which aligns with your opinionated stance on using "enabled" instead of "disabled".  Thus we could drop the "prevent" there and flip the meaning.

epugh: Love it!

PR #3620, opened 2025-09-04
[NO JIRA]: Fix BasicDistributedZkTest test failure

mlbiscoc:
No Jira. BasicDistributedZkTest has been failing since I migrated `DirectUpdateHandler2`. This is because it would call the /admin/metrics handler and look for the Dropwizard/Json format of these metrics. Migrated it to look for prometheus format instead.


dsmiley (commented):
I feel like we need a StringResponseParser for some tests like this.  (out of scope of this PR; just thinking out loud)

PR #3622, opened 2025-09-05
remove some stray backticks in Solr Ref Guide

cpoerschke:
Stumbled across one of them for real and then

```
git grep -n "\`" *.adoc | grep -v "\`.*\`"
```

prospected for more with small #2401 deja vu sense.


epugh (approved):
Nice find!

dsmiley (approved):

PR #3625, opened 2025-09-06
SOLR-17864: Migrate System Properties (part ocho)

epugh:

solr.prs.default
enable.update.log
zk.server.conf.dir
zk.server.data.dir
zk.run
solr.zk.embedded.host
zk.run.only

### DIFF
--- a/solr/bin/solr.in.cmd
+++ b/solr/bin/solr.in.cmd
@@ -127,7 +127,7 @@ REM environments where security is not a concern, 0.0.0.0 can be used to allow
 REM Solr to accept connections on all network interfaces.
 REM set SOLR_JETTY_HOST=127.0.0.1
 REM Sets the network interface the Embedded ZK binds to.
-REM set SOLR_ZOOKEEPER_SERVER_EMBEDDED_HOST=127.0.0.1
+REM set SOLR_ZOOKEEPER_EMBEDDED_HOST=127.0.0.1
 
 REM Restrict access to solr by IP address.
 REM Specify a comma-separated list of addresses or networks, for example:
--- a/solr/solrj/src/resources/DeprecatedSystemPropertyMappings.properties
+++ b/solr/solrj/src/resources/DeprecatedSystemPropertyMappings.properties
@@ -60,4 +60,4 @@ solr.zookeeper.chroot.create=create.zk.chroot
 solr.zookeeper.server.confdir=zk.server.conf.dir
 solr.zookeeper.server.datadir=zk.server.data.dir
 solr.zookeeper.server.enabled=zk.run
-solr.zookeeper.server.embedded.host=solr.zk.embedded.host
+solr.zookeeper.embedded.host=solr.zk.embedded.host
--- a/solr/bin/solr.in.sh
+++ b/solr/bin/solr.in.sh
@@ -152,7 +152,7 @@
 # Solr to accept connections on all network interfaces.
 #SOLR_JETTY_HOST="127.0.0.1"
 # Sets the network interface the Embedded ZK binds to.
-#SOLR_ZOOKEEPER_SERVER_EMBEDDED_HOST="127.0.0.1"
+#SOLR_ZOOKEEPER_EMBEDDED_HOST="127.0.0.1"
 
 # Enables HTTPS. It is implictly true if you set SOLR_SSL_KEY_STORE. Use this config
 # to enable https module with custom jetty configuration.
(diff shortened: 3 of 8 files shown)


epugh:
@gerlowskija remember when we hacked around with `zkRun` together in Pittsburgh?    I did some refactoring when it went from a string to a boolean, and I'd love another set of eyes...

solr/bin/solr.in.cmd
```diff
@@ -127,7 +127,7 @@ REM environments where security is not a concern, 0.0.0.0 can be used to allow
 REM Solr to accept connections on all network interfaces.
 REM set SOLR_JETTY_HOST=127.0.0.1
 REM Sets the network interface the Embedded ZK binds to.
-REM set SOLR_ZK_EMBEDDED_HOST=127.0.0.1
+REM set SOLR_ZOOKEEPER_SERVER_EMBEDDED_HOST=127.0.0.1
```
dsmiley: why put the word "embedded" in there?  "server" implies embedded; it refers to the ZK server that Solr is running itself, embedded.
I could see swapping SERVER for EMBEDDED though... it would be clearer.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -255,10 +255,11 @@ public String toString() {
   private final Map<String, String> replicasInLeaderInitiatedRecovery = new HashMap<>();
 
   // This is an expert and unsupported development mode that does not create
```
dsmiley: IMO we should drop this entirely.  I've never heard of it and doesn't really sound useful TBH.  Why run Solr that only runs ZK (that is confusing BTW!) when one could run ZK via Docker?  Docker makes running anything easy in a developer workstation scenario, which is what this feature here is for.

CC @markrmiller

solr/solrj/src/resources/DeprecatedSystemPropertyMappings.properties
```diff
@@ -55,3 +57,8 @@ solr.ui.enabled=!solr.admin.ui.disabled
 solr.ui.experimental.enabled=!solr.admin.ui.experimental.disabled
 
 solr.zookeeper.chroot.create=create.zk.chroot
+solr.zookeeper.server.confdir=zk.server.conf.dir
+solr.zookeeper.server.datadir=zk.server.data.dir
+solr.zookeeper.server.enabled=zk.run
+solr.zookeeper.server.embedded.host=solr.zk.embedded.host
```
dsmiley: ah; only this one did you put "embedded".  again; don't need both "server" and "embedded"; pick one consistently for the set of running and embedded ZK server.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
epugh: I will take a stab at removing, but if we want to restore it'll be the matter of rolling back a single commit!

epugh: Done in #1225833196dea24f413abbc3cc3fbef3d5b769fb

solr/solrj/src/resources/DeprecatedSystemPropertyMappings.properties
epugh: I guess i can see an argument for ".server" being for zookeeper server related topics, and then ".embedded" for settings specfic to an embedded ZK...

solr/bin/solr.in.cmd
epugh: We do want to keep this environmental name the same as the system property, so SOLR_ZOOKEEPER_EMBEDDED_HOST so we get the magic translation and don't need to pass it in expliiclty to Solr...

dsmiley: I understand that; they need to be consistent for EnvUtils to work.
I'm discussing what the name should be (which will look a certain way when expressed as an env var vs system property).

solr/solrj/src/resources/DeprecatedSystemPropertyMappings.properties
dsmiley: That is the same thing.  There is no distinction.  Solr only runs the server embedded; Solr doesn't run ZK separately.  A user probably will but not Solr.  These "server" props listed (like confdir) only applies to embedding ZK.  In theory we might decide to fork a process to run another JVM for ZK... that would then not be "embedded".  I could go either way on standardizing on "server" or "embedded" but I definitely propose we pick one.

epugh: I think I missed a rename in the mapping file...

epugh: Never mind, I just checked and the mapping file has the correct names.

PR #3632, opened 2025-09-07
Rename CloudHttp2SolrClient.withInternalClientBuilder

dsmiley:
to withHttpClientBuilder
(kept with deprecations)

Boring enough that I'd prefer to avoid JIRA/CHANGES.txt


dsmiley (commented):
Will merge this Tuesday if no further feedback.

dsmiley:
@jdyer1 do you agree with this change?

PR #3634, opened 2025-09-08
SOLR-17859: Restore the use of TimeLimitingBulkScorer.

sigram:
During the upgrade to Lucene 10 due to the Lucene API changes the `SolrIndexSearcher.search(List<LeafReaderContext>, Weight, Collector)` became obsolete and was commented out.

However, this removed Solr's ability to trigger the use of `TimeLimitingBulkScorer` in Lucene when `QueryLimits` are in use. This PR fixes it and adds a unit test to make sure that this scorer is in fact used when limits are active.

### DIFF
--- a/solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
+++ b/solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
@@ -57,6 +57,7 @@ import org.apache.lucene.search.BooleanClause.Occur;
 import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.search.CollectionStatistics;
 import org.apache.lucene.search.Collector;
+import org.apache.lucene.search.CollectorManager;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.Explanation;
 import org.apache.lucene.search.FieldDoc;
@@ -326,7 +327,7 @@ public class SolrIndexSearcher extends IndexSearcher implements Closeable, SolrI
 
     try {
       try {
-        super.search(query, collector);
+        search(query, collector);
       } finally {
         // The complete() method can use the collectors, so this needs to be surrounded by the same
         // catch logic that limit collecting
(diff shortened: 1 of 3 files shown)


solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -782,32 +783,31 @@ public QueryResult search(QueryResult qr, QueryCommand cmd) throws IOException {
...
   }
 
-  /*@Override
-  TBD . This is overridden to exit the query when limits are reached
-  protected void search(List<LeafReaderContext> leaves, Weight weight, Collector collector)
+  @Override
+  public void search(Query query, Collector collector)
       throws IOException {
     QueryLimits queryLimits = QueryLimits.getCurrentLimits();
-    if (EnvUtils.getPropertyAsBool(EXITABLE_READER_PROPERTY, Boolean.FALSE)
```
sigram, line 790: This condition is most likely obsolete here - if it's true then we're using the `ExitableDirectoryReader` anyway, regardless whether the query uses limits or not.

solr/core/src/java/org/apache/solr/search/SolrIndexSearcher.java
```diff
@@ -782,32 +783,31 @@ public QueryResult search(QueryResult qr, QueryCommand cmd) throws IOException {
...
-      super.search(leaves, weight, collector);
+      super.search(query, collector);
     } else {
       // Timeout enabled!  This impl is maybe a hack.  Use Lucene's IndexSearcher timeout.
       // But only some queries have it so don't use on "this" (SolrIndexSearcher), not to mention
       //   that timedOut() might race with concurrent queries (dubious design).
       // So we need to make a new IndexSearcher instead of using "this".
-      new IndexSearcher(reader) { // cheap, actually!
+      // Make sure to reuse the existing executor.
+      new IndexSearcher(reader, core.getCoreContainer().getIndexSearcherExecutor()) { // cheap, actually!
```
sigram: Previously this code would silently drop the parent searcher's executor, make sure we re-use it here.

solr/test-framework/src/java/org/apache/solr/search/CallerSpecificQueryLimit.java
```diff
@@ -53,10 +54,10 @@ public class CallerSpecificQueryLimit implements QueryLimit {
   // className -> set of method names
   private final Map<String, Set<String>> interestingCallers = new HashMap<>();
   // expr -> initial count
-  private final Map<String, Integer> maxCounts = new HashMap<>();
+  private final Map<String, Integer> maxCounts = new ConcurrentHashMap<>();
```
sigram, line 56: Turns out an ordinary HashMap doesn't like multi-threaded modifications.

dsmiley:
+1 thanks

chatman:
Thanks @sigram!

PR #3638, opened 2025-09-08
Fix a tiny bitrot in cloud.sh with command script params

gus-asf:
Just a couple updates to track the most current bin/solr command line options.


epugh (approved):
LGTM.   I thought I was on top of it, but clearly not, thanks for patching the rot!

PR #3640, opened 2025-09-09
SOLR-17445 Update Docker base image to eclipse-temurin:21-jre-noble

janhoy:
https://issues.apache.org/jira/browse/SOLR-17445

Bump base OS version to latest LTS (Noble) before 10.x release.


janhoy:
Merged this on lazy consensus, speak up if you find any flaws after the fact :)

PR #3643, opened 2025-09-10
SOLR-17293: overseerEnabled=false by default

dsmiley:
Changed some test usages of MiniSolrCloudCluster to use SolrCloudTestCase in order to get coverage of both modes instead of just one.

And tweaked some ZkController ctor order sequence

https://issues.apache.org/jira/browse/SOLR-17293

### DIFF
--- a/solr/core/src/java/org/apache/solr/cloud/ZkController.java
+++ b/solr/core/src/java/org/apache/solr/cloud/ZkController.java
@@ -74,6 +74,7 @@ import org.apache.solr.common.AlreadyClosedException;
 import org.apache.solr.common.MapWriter;
 import org.apache.solr.common.SolrException;
 import org.apache.solr.common.SolrException.ErrorCode;
+import org.apache.solr.common.cloud.ClusterProperties;
 import org.apache.solr.common.cloud.ClusterState;
 import org.apache.solr.common.cloud.DefaultZkACLProvider;
 import org.apache.solr.common.cloud.DefaultZkCredentialsInjector;
@@ -385,27 +386,23 @@ public class ZkController implements Closeable {
             });
 
     zkStateReader.createClusterStateWatchersAndUpdate(); // and reads cluster properties
+    // HENCEFORTH can read cluster props...
 
-    // note: Can't read cluster properties until createClusterState ^ is called
-    final String urlSchemeFromClusterProp =
-        zkStateReader.getClusterProperty(ZkStateReader.URL_SCHEME, ZkStateReader.HTTP);
-    // this must happen after zkStateReader has initialized the cluster props
-    this.baseURL = URLUtil.getBaseUrlForNodeName(this.nodeName, urlSchemeFromClusterProp);
+    this.baseURL =
+        URLUtil.getBaseUrlForNodeName(
+            this.nodeName,
+            zkStateReader.getClusterProperty(ZkStateReader.URL_SCHEME, ZkStateReader.HTTP));
 
     checkForExistingEphemeralNode(); // removes our live node if present
 
-    // Now that zkStateReader is available, read OVERSEER_ENABLED & live nodes.
+    // note: reads live nodes
     final Optional<SolrVersion> lowestNodeVersion = zkStateReader.fetchLowestSolrVersion();
-    boolean defaultOverseerEnabled =
-        lowestNodeVersion.orElse(SolrVersion.LATEST).lessThan(SolrVersion.forIntegers(10, 0, 0));
-    final boolean overseerEnabled =
-        Boolean.parseBoolean(
-            String.valueOf(
-                zkStateReader.getClusterProperty(
-                    ZkStateReader.OVERSEER_ENABLED,
-                    EnvUtils.getPropertyAsBool(
-                        "solr.cloud.overseer.enabled", defaultOverseerEnabled))));
 
+    if (lowestNodeVersion.isPresent()) {
+      checkClusterVersionCompatibility(lowestNodeVersion.get());
+    }
+
+    final boolean overseerEnabled = isOverseerEnabled(lowestNodeVersion);
     if (overseerEnabled) {
       log.info("The Overseer is enabled.  It will process all cluster commands & state updates.");
     } else {
@@ -418,10 +415,6 @@ public class ZkController implements Closeable {
         !overseerEnabled
             ? Optional.of(new DistributedCollectionConfigSetCommandRunner(cc, zkClient))
             : Optional.empty();
-
-    if (lowestNodeVersion.isPresent()) {
-      checkClusterVersionCompatibility(lowestNodeVersion.get());
-    }
 
     registerLiveNodesListener();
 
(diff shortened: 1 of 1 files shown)


solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -392,13 +392,19 @@ public ZkController(
     // this must happen after zkStateReader has initialized the cluster props
     this.baseURL = URLUtil.getBaseUrlForNodeName(this.nodeName, urlSchemeFromClusterProp);
 
-    // Now that zkStateReader is available, read OVERSEER_ENABLED.
+    checkForExistingEphemeralNode(); // removes our live node if present
```
dsmiley, line 396: Moved from init() to occur a little sooner because we're about to read live nodes (to get cluster version) right after.  Just being thorough; seems this check is an edge case.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -413,7 +419,18 @@ public ZkController(
             ? Optional.of(new DistributedCollectionConfigSetCommandRunner(cc, zkClient))
             : Optional.empty();
 
-    init();
+    if (lowestNodeVersion.isPresent()) {
+      checkClusterVersionCompatibility(lowestNodeVersion.get());
+    }
```
dsmiley: Brought this out of init() because I want to pass lowestNodeVersion to it.  This was the catalyst of doing a bit of reorganization that you'll see in this PR here.  I hope it's not "too much" in the same PR that has an objective to flip a default.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -413,7 +419,18 @@ public ZkController(
...
+
+    registerLiveNodesListener();
+
+    startOverseer();
```
dsmiley, line 421: You'll see init() has been reduced to only starting the overseer.  So now we call it that.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -423,6 +440,10 @@ public ZkController(
     this.overseerCollectionQueue = overseer.getCollectionQueue(zkClient);
     this.overseerConfigSetQueue = overseer.getConfigSetQueue(zkClient);
     this.sysPropsCacher = new NodesSysPropsCacher(cc.getDefaultHttpSolrClient(), zkStateReader);
+
+    // Do this last to signal we're up.
+    createEphemeralLiveNode();
+
```
dsmiley, line 440: Moved this to be even more "last" than it was :-)

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -1078,47 +1097,25 @@ private static void repairSecurityJson(SolrZkClient zkClient)
...
-      // Do this last to signal we're up.
-      createEphemeralLiveNode();
-    } catch (InterruptedException e) {
-      // Restore the interrupted status
-      Thread.currentThread().interrupt();
-      log.error("", e);
-      throw new ZooKeeperException(SolrException.ErrorCode.SERVER_ERROR, "", e);
-    } catch (KeeperException e) {
-      log.error("", e);
-      throw new ZooKeeperException(SolrException.ErrorCode.SERVER_ERROR, "", e);
```
dsmiley, line 1121: This exception stuff is actually redundant.  ZkContainer calls the constructor of ZkController and it's got the same stuff.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -423,6 +440,10 @@ public ZkController(
     this.overseerCollectionQueue = overseer.getCollectionQueue(zkClient);
     this.overseerConfigSetQueue = overseer.getConfigSetQueue(zkClient);
```
dsmiley, line 434: It's tempting to continue moving things around... like these overseer queues should probably be set up in startOverseer.

solr/test-framework/src/java/org/apache/solr/cloud/MiniSolrCloudCluster.java
```diff
@@ -1031,7 +1031,7 @@ public static class Builder {
 
     private boolean trackJettyMetrics;
     private boolean overseerEnabled =
-        EnvUtils.getPropertyAsBool("solr.cloud.overseer.enabled", true);
+        EnvUtils.getPropertyAsBool("solr.cloud.overseer.enabled", false);
```
dsmiley, line 1034: It'd be tempting to put the randomized default here.  I dunno... it might couple this utility base classes that a 3rd party plugin writer may not want to use.  Granted this may already be the case; we don't have a real way of testing that.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
HoustonPutman: Ahhh very good catch. This will definitely happen, so not just a weird edge case.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
```diff
@@ -392,13 +392,19 @@ public ZkController(
     // this must happen after zkStateReader has initialized the cluster props
     this.baseURL = URLUtil.getBaseUrlForNodeName(this.nodeName, urlSchemeFromClusterProp);
 
-    // Now that zkStateReader is available, read OVERSEER_ENABLED.
+    checkForExistingEphemeralNode(); // removes our live node if present
+
+    // Now that zkStateReader is available, read OVERSEER_ENABLED & live nodes.
+    final Optional<SolrVersion> lowestNodeVersion = zkStateReader.fetchLowestSolrVersion();
+    boolean defaultOverseerEnabled =
```
HoustonPutman: So I have a relatively big question here. So let's say a cloud is rolling upgraded from 9 to 10.

All nodes after the upgrade will be using the overseer, except for the last one upgraded to 10 right? That one will be using the distributed cluster state manager. This seems like an issue, but I have a feeling I'm missing something.

solr/core/src/java/org/apache/solr/cloud/ZkController.java
HoustonPutman: Would be nice to have a separate refactoring PR. Good restraint haha

HoustonPutman (commented):
Irrespective of the change itself, I like the code here and the changes made.

My big question is really around whether or not this will be safe during a rolling upgrade.

I kind of hate to suggest this because we want to keep the liveNode information lean, but maybe that can be used to safeguard starting Solr in 2 different cluster management modes. (overseer vs distributed). So if a node is started without the setting, then it uses what all other nodes are using. And if it's the first live node, then it will use the default as provided here. If the node has a setting given, it will be checked against the existing live nodes and if it doesn't match, we will throw and error and shut down. Thoughts?

solr/core/src/java/org/apache/solr/cloud/ZkController.java
dsmiley: Woah, great catch!  Code review FTW.  Damn; we need multi-version Solr testing.  Until then, at least manually -- which I should do using maybe Docker.

As for a solution, perhaps a Solr node observing an older version from lowestNodeVersion at startup (here) should explicitly set the `overseerEnabled`=false cluster property so that the choice is made durably without flipping accidentally.  This means someone doing a rolling upgrade will have a cluster that is with an Overseer henceforth.  But someone creating a fresh cluster or who upgrades via entire restart will get distributed mode (henceforth).  I'm glad I added the cluster downgrade prevention check as well.

PR #3645, opened 2025-09-10
SOLR-17641: Disable the Security Manager on JDK24+ (branch_9x)

janhoy:
This is the backport to branch_9x

(cherry picked from commit 390e11d1bafad162180df444c9633b737d1b2228)

https://issues.apache.org/jira/browse/SOLR-17641


janhoy:
Unfortunately the github action will not be able to run the bin-solr tests on JDK24 due to too old gradle. But we'll get a validation that Solr still runs on JDK11 with JSM enabled..

PR #3116, opened 2025-01-24
Update google.errorprone to v2.41.0

solrbot:
This PR contains the following updates:

| Package | Type | Update | Change |
|---|---|---|---|
| [com.google.errorprone:error_prone_core](https://errorprone.info) ([source](https://redirect.github.com/google/error-prone)) | dependencies | minor | `2.31.0` -> `2.41.0` |
| [com.google.errorprone:error_prone_annotations](https://errorprone.info) ([source](https://redirect.github.com/google/error-prone)) | dependencies | minor | `2.31.0` -> `2.41.0` |

---

### Release Notes

<details>
<summary>google/error-prone (com.google.errorprone:error_prone_core)</summary>

### [`v2.41.0`](https://redirect.github.com/google/error-prone/releases/tag/v2.41.0): Error Prone 2.41.0

New checks:

- [EffectivelyPrivate](https://errorprone.info/bugpattern/EffectivelyPrivate): Detect declarations that have `public` or `protected` modifiers, but are effectively private

Changes:

- Skip BooleanLiteral findings if the target type is boxed ([#&#8203;5134](https://redirect.github.com/google/error-prone/issues/5134))

Full changelog: <https://github.com/google/error-prone/compare/v2.40.0...v2.41.0>

### [`v2.40.0`](https://redirect.github.com/google/error-prone/releases/tag/v2.40.0): Error Prone 2.40.0

Changes:

- Bug fixes and improvements
- Releases (including snapshots) have migrated from [OSSRH to the Central Publisher Portal](https://central.sonatype.org/pages/ossrh-eol/#process-to-migrate)

Full changelog: <https://github.com/google/error-prone/compare/v2.39.0...v2.40.0>

### [`v2.39.0`](https://redirect.github.com/google/error-prone/releases/tag/v2.39.0): Error Prone 2.39.0

Changes:

- Temporarily downgrade to Guava 33.4.0 ([#&#8203;5108](https://redirect.github.com/google/error-prone/issues/5108))

Checks:

- [`BooleanLiteral`](https://errorprone.info/bugpattern/BooleanLiteral): Prefer `true` to `Boolean.TRUE`
- [`ExpensiveLenientFormatString`](https://errorprone.info/bugpattern/ExpensiveLenientFormatString): Renamed from `PreconditionsExpensiveString`, detects unnecessary calls to `String.format` in the arguments of lenient formatting methods.
[...]

PR #3647, opened 2025-09-10
Update gradle to 8.14.3

janhoy:
Supports Java 24, so we can then add matrix of JDK versions to run the BATS tests for.
I suppose we need to go to gradle 9 to get JDK 25 support eventually, but this is a stepping stone.


janhoy:
Enabled BATS run on JDK24 but our build is not yet compatible...

dsmiley:
Question: did you take anything from Lucene's same upgrade?  We share the same build infra... granted it grows increasingly divergent.

janhoy:
Nope, just a quick attempt. Makes sense to peek at their pr if they already upgraded.

janhoy:
See https://github.com/apache/lucene/pull/14591 for Lucene's upgrade PR.

janhoy:
Looks like we need to upgrade errorProne to be able to compile Solr under JDK24?

```
> Task :solr:api:compileJava
error: An unhandled exception was thrown by the Error Prone static analysis plugin.
     Please report this at https://github.com/google/error-prone/issues/new and include the following:
  
     error-prone version: 2.31.0
     BugPattern: (see stack trace)
     Stack Trace:
     java.lang.NoSuchFieldError: Class com.sun.tools.javac.code.TypeTag does not have member field 'com.sun.tools.javac.code.TypeTag UNKNOWN'
  	at com.google.errorprone.util.ASTHelpers.<clinit>(ASTHelpers.java:1301)
  	at com.google.errorprone.bugpatterns.PackageInfo.matchCompilationUnit(PackageInfo.java:39)
  	at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:449)
```

malliaridis:
> Looks like we need to upgrade errorProne to be able to compile Solr under JDK24?

We are 10 minor versions behind and a lot of new checks have been added. Would be a good idea to update it first.

But for backporting we should take [this information](https://github.com/google/error-prone/releases/tag/v2.32.0) into account (we might not be able to backport to 9.x).

janhoy:
Yes, I'm aware that 9.x needs to stay on 2.31. I have started to adapt #3116 with the missing rules..

.github/workflows/bin-solr-test.yml
```diff
@@ -18,6 +18,10 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 40
 
+    strategy:
```
dsmiley, line 21: IMO we should leave multi-JDK testing to Jenkins, and keep these GHA CI for PRs where we don't need to test multi-JDK.  The likelihood of breaking another JDK is small, and wouldn't be specific to the bats tests.

